PR #657, opened 2013-10-09
Stream/Trace.plot() improvements

megies:
Improve plotting:

 - better x-tick formatting
   - ticks get updated automatically
   - ticks are now correct (float seconds were cut off previously and tick locations were fixed rather than aligned on even values)
   - format of tick labels gets set automatically depending on distance between ticks (dynamically!)
 - share x axes of subplots (for interactive zooming)
 - Y tick labels at well-spaced values
 - warning text when zooming into a minmax-type plot

Speed of plotting is not decreased.

Here is a script to run some plots:
```python
from obspy import read

st = read()
for tr in st:
    tr.stats.location = "00"
st += read()

for type in (None, "relative"):
    for method in ("fast", "full"):
        for samp_rate in [0.0001, 0.001, 0.1, 1, 100, 10000]:
            print "Type: %8s, method: %6s, samp_rate: %.4f" % (type, method, samp_rate)
            for tr in st:
                tr.stats.sampling_rate = samp_rate
            st.plot(type=type, method=method)
```

This PR will also enable better custom plotting adjustments, e.g. use `plt.axvspan(num2date(my_timestamp), num2date(my_timestamp2))`


megies:
(Finally) good to merge from my point of view, please review.. Long overdue to make the preview plots have a sensible x-axis.

QuLogic:
> ticks get updated automatically

I like this PR already...

obspy/imaging/scripts/scan.py
```diff
@@ -244,9 +244,12 @@ def main(argv=None):
         matplotlib.use("agg")
     global date2num
```
QuLogic: Is there a reason why this should be global?

obspy/imaging/waveform.py
```diff
@@ -37,6 +40,10 @@
 from obspy.core.util import (FlinnEngdahl, createEmptyDataChunk,
                              locations2degrees)
 from obspy.core.util.decorator import deprecated_keywords
+from obspy.imaging.util import ObsPyAutoDateFormatter
+
+
+MINMAX_ZOOMLEVEL_WARNING_TEXT = "Warning: Zooming into MinMax Plot!"
```
QuLogic: So? This just restates what the user is already doing. Why is it a problem?

obspy/imaging/waveform.py
```diff
@@ -712,12 +721,20 @@ def __plotStraight(self, trace, ax, *args, **kwargs):  # @UnusedVariable
             # set starttime and calculate endtime
             trace.stats.starttime = self.starttime
         trace.data = np.require(trace.data, np.float64) * calib
+        if self.type == 'relative':
+            start = 0
```
QuLogic: `relative` plots do not always start at 0.

obspy/imaging/waveform.py
```diff
@@ -803,7 +820,15 @@ def __plotMinMax(self, trace, ax, *args, **kwargs):  # @UnusedVariable
                            minmax[:, 0].min(),
                            minmax[:, 1].max()])
         # Finally plot the data.
-        x_values = np.repeat(np.arange(self.width), 2)
+        if self.type == 'relative':
+            start = 0
```
QuLogic: Same note here.

obspy/imaging/waveform.py
```diff
@@ -1357,11 +1365,34 @@ def __setupFigure(self):
...
+        if xlim[1] - xlim[0] < self._initial_xrange:
+            dangerous = True
+        else:
+            dangerous = False
+        if dangerous and not self._minmax_plot_xrange_dangerous:
+            self._add_zoomlevel_warning_text()
+        elif self._minmax_plot_xrange_dangerous and not dangerous:
+            self._remove_zoomlevel_warning_text()
+        self._minmax_plot_xrange_dangerous = dangerous
+        return
```
QuLogic: Superfluous `return`?

obspy/imaging/waveform.py
```diff
@@ -1357,11 +1365,34 @@ def __setupFigure(self):
...
+        ax = self.fig.axes[0]
+        ax.text(0.95, 0.9, MINMAX_ZOOMLEVEL_WARNING_TEXT, color="r",
+                ha="right", va="top", transform=ax.transAxes)
 
-    def __getX(self, dx):
-        return dx * 1.0 / self.width
+    def _remove_zoomlevel_warning_text(self):
+        ax = self.fig.axes[0]
+        for text in list(ax.texts):
+            if text.get_text() == MINMAX_ZOOMLEVEL_WARNING_TEXT:
```
QuLogic: Could you not just save the result of `ax.text`?

QuLogic:
It looks pretty good in general, definitely an improvement, but I have some comments/questions about the images, too:
 * Whether the first and/or last ticks are shown seems kind of random.
 * `obspy/imaging/tests/images/plot.png`: The interval is 5 seconds and the end of the plot should be 1 second after the last, but the width of that 1 second only seems to be about a third of the width between each of the other ticks.
 * `obspy/imaging/tests/images/waveform_10_traces{,_huge,_tiny}.png`: Title is overlapping graph.
 * `obspy/imaging/tests/images/waveform_1_trace.png`: If you Onion Skin this one, you see the trace and the tick positions do not change at all, yet the tick labels go up by one. (Probably happens on others, but it's easiest to see here, since layout doesn't change.) Is this actually correct? 
 * `obspy/imaging/tests/images/waveform_reftime_relative.png`: The x-axis is 100% wrong, since it should not start at 0.

obspy/imaging/scripts/scan.py
megies: If there is one I can't see it right now. ;)
Anyway, that's some historic thing I guess and not related to this PR.

obspy/imaging/waveform.py
megies: That's right, didn't think of it..

obspy/imaging/scripts/scan.py
QuLogic: Yea, I couldn't either; just curious...

megies: I suppose it's some left over from trying to avoid importing matplotlib at top.

obspy/imaging/waveform.py
megies: Well, you're right supposing the user explicitely requested a fast/minmax plot.

But.. the fast/minmax option is used automatically for high npts and I am fairly sure that most people are not aware of this fact. I assume most people in that case would recognize the fact that the strange looks when zooming in are artifacts from the plotting, but if only one person would be held up for ten minutes wondering about the seismogram I think the warning message is already justified..

obspy/imaging/waveform.py
megies: Hmm, OK, never really used the `relative` option. I'll try to adjust correctly. In any case there should be a test to pin this behavior..

obspy/imaging/waveform.py
megies: I'll remove it..

obspy/imaging/waveform.py
QuLogic: No, I'm not against the warning. I just mean it should indicate why zooming can't or shouldn't be done a bit more clearly, even if it's just to say something as simple as it's unsupported or something.

obspy/imaging/waveform.py
QuLogic: There is; you replaced the image ;)

obspy/imaging/waveform.py
megies: Ok, I see. I thought it's enough and the text can't be too long. Maybe I'll try to be more clear..

obspy/imaging/waveform.py
megies: Myup.. actually only just now noticed after fixing the PR in that regard.. ;-)
Gonna push the fix in a couple of minutes..

megies:
> Whether the first and/or last ticks are shown seems kind of random.

Tick placement is totally up to matplotlib now. I'm fully aware that it might not be perfect for the initial plot state in many cases. But IMHO utterly outweighed by..
 - removing all the tick placement magic
 - having dynamically updating ticks
 - having correct tick labels *for any plot for the first time ever* (anything below seconds was simply stripped up to now)

> obspy/imaging/tests/images/plot.png: The interval is 5 seconds and the end of the plot should be 1 second after the last, but the width of that 1 second only seems to be about a third of the width between each of the other ticks.

Hmm.. I don't get your point, I think. A third of 5 secs is 1.66 and it's fitting well. The end of trace is `15.875`. Doing the plot of those files interactively the x axis is just fine. Actually, I dare say the x axis on the old image is just plain wrong, as wrong as for any old image that is on the order of seconds or even less (due to the tick labeling magic in combination with the strange calculation of x values on magically merged traces). 

> obspy/imaging/tests/images/waveform_10_traces{,_huge,_tiny}.png: Title is overlapping graph.

Well.. that's systematic..

> obspy/imaging/tests/images/waveform_1_trace.png: If you Onion Skin this one, you see the trace and the tick positions do not change at all, yet the tick labels go up by one. (Probably happens on others, but it's easiest to see here, since layout doesn't change.) Is this actually correct?

Again, the old tick labelling was just plain wrong in the sub-second range..

> obspy/imaging/tests/images/waveform_reftime_relative.png: The x-axis is 100% wrong, since it should not start at 0. I guess those parts of the PR predate the reftime addition.

Hmm.. indeed. Fixed now. Also totally cleaned up the messy handling of values on x-axis.

obspy/imaging/util.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+"""
+Waveform plotting utilities.
+
+:copyright:
+    The ObsPy Development Team (devs@obspy.org)
+:license:
+    GNU General Public License (GPL)
```
QuLogic: You meant LGPL, I assume?

QuLogic:
> Tick placement is totally up to matplotlib now. I'm fully aware that it might not be perfect for the initial plot state in many cases. But IMHO utterly outweighed by..

Agreed, it's certainly a lot better for interactive plots, and for static plots, people will probably tweak it.

> >    obspy/imaging/tests/images/plot.png: The interval is 5 seconds and the end of the plot should be 1 second after the last, but the width of that 1 second only seems to be about a third of the width between each of the other ticks.
> 
> Hmm.. I don't get your point, I think. A third of 5 secs is 1.66 and it's fitting well. The end of trace is 15.875.

Ah, okay, that's probably important. Is there a reason the title doesn't show the full time? I assumed it was actually what it said, and that's likely where the confusion arose. I think you pointed out earlier that the original simply truncated the actual tick location, which is now fixed, so why not in the title?

>     obspy/imaging/tests/images/waveform_10_traces{,_huge,_tiny}.png: Title is overlapping graph.
> 
> Well.. that's systematic..

I'm just wondering if the `subplots_adjust` might have caused it?

>     obspy/imaging/tests/images/waveform_1_trace.png: If you Onion Skin this one, you see the trace and the tick positions do not change at all, yet the tick labels go up by one. (Probably happens on others, but it's easiest to see here, since layout doesn't change.) Is this actually correct?
> 
> Again, the old tick labelling was just plain wrong in the sub-second range..

Alright, that's a perfectly valid reason.

obspy/imaging/waveform.py
```diff
@@ -22,21 +22,36 @@
...
+
+
+MINMAX_ZOOMLEVEL_WARNING_TEXT = "Warning: Zooming into MinMax Plot!"
+SECONDS_PER_DAY = 3600.0 * 24.0
+
+datelocator_warning_msg = (
+    "AutoDateLocator was unable to pick an appropriate interval for this date "
+    "range. It may be necessary to add an interval value to the "
+    "AutoDateLocator's intervald dictionary.")
+warnings.filterwarnings("ignore", datelocator_warning_msg, UserWarning,
```
QuLogic: Does this filter only affect this module, or all code?

obspy/imaging/util.py
megies: Thanks, that was before the License change and I must have overlooked it during updating to current master..

megies:
> [..] you pointed out earlier that the original simply truncated the actual tick location, which is now fixed, so why not in the title?

Absolutely, I didn't notice the times were truncated in the title, too. I'll change it.

> I'm just wondering if the subplots_adjust might have caused it?

I didn't touch the `subplot_adjust` (at least not til your comment ;)), but I removed the manually set X/Y coordinates in the title placement. I guess I'll revert that couple of lines to switch back to the suptitle placement magic..

> Does this filter only affect this module, or all code?

Hmm, yeah, you're right. I should've handled this in a context manager in the first place. I'll change it.

megies:
Should be good to go now..

megies:
I'll merge this now, anything that pops up can be fixed in master..

PR #726, opened 2014-01-30
added feature SSxM

ThomasLecocq:
Added obspy.signal.ssxm : method to compute Real-time Amplitude/Energy Measurement and
Spectral Seismic Amplitude/Energy Measurement (known as RSAM, RSEM, SSAM, SSEM).

I tried to be Py3 compatible (only for indexing, I don't know the rest).

### DIFF
--- a/obspy/signal/ssxm.py
+++ b/obspy/signal/ssxm.py
@@ -71,8 +71,8 @@ def ssxm(data, fs, id, starttime, rule='30S', bands=None, corners=4,
     for band in bands:
         if band != [0, 0]:
             tmp = s.copy()
-            tmp.data = bandpass(tmp, band[0], band[1], fs, corners=corners,
-                                zerophase=zerophase)
+            tmp[:] = bandpass(tmp, band[0], band[1], fs, corners=corners,
+                              zerophase=zerophase)
         else:
             tmp = s.copy()
 


krischer:
@ThomasLecocq Any update on this? I think the resampling from pandas can be easily replaced by some custom functionality and then we can get rid of pandas in this PR.

ThomasLecocq:
damn racists, pandas should be a dependency of obspy anyway ;)

but, yeah, ok, I'll think about it... (I really "think" pandas, so it's quite hard for me to move away from it, but I'll try)...

obspy/signal/ssxm.py
```diff
@@ -0,0 +1,118 @@
...
+#-------------------------------------------------------------------
+# Filename: ssxm.py
+#  Purpose: RSAM, RSEM, SSAM and SSEM calculations
+#   Author: Thomas Lecocq & Corentin Caudron
+#    Email: thomas.lecocq@seismology.be
+#
+# Copyright (C) 2012-2014 T. Lecocq, C. Caudron
+#-------------------------------------------------------------------
+"""
+Real-time Amplitude/Energy Measurement and
```
ThomasLecocq: indeed...

krischer:
@ThomasLecocq There is some interest in this. Can you finish this soonish? I guess only a couple things are missing:

- [ ] Address the comments from @QuLogic 
- [ ] Add a couple of tests.
- [ ] Replace the pandas calls.

ThomasLecocq:
I know you guys hate pandas, but, pffr... I'll see what I can do.

krischer:
;-)  No...i really like pandas and personally use it as much as I can but it is a dependency we can avoid and and in this case we should. Especially because pandas is not a pure Python dependency.

If you add some tests we can help replacing it!

ThomasLecocq:
Yop, 

OK, fair enough... Well... One thing that would be reallllly useful in the "replace" scheme would be the idea we had about the "windowed Stream"... Then it would be quite straightforward.

So, before addressing the items you listed, I suppose I should 

- [ ] rebase the PR on the current master ?

ThomasLecocq:
working on slide first

krischer:
No stop. I'm almost done with it ;-)

krischer:
I'm just at a workshop right now so it might take me a bit longer.

ThomasLecocq:
```


```

ThomasLecocq:
@krischer : do you add it at the trace level too ?

ThomasLecocq:
actually, I'm quite reversing the whole process.. to reduce the number of lines of `signal.ssxm` to the minimum :)

ssxm

krischer:
Yes. Mine has some additional features but the basics are the same. Should be ready sometime today.

ThomasLecocq:
> mine has addition features

tsss :dart: 

Excellent. I base my current work on mine and I'll adapt to yours. I'm working in `master`. Added `Stream.ssxm` and `Trace.ssxm` to mimic the other methods (`split` or else)

ThomasLecocq:
Brainstorming:

Currently, the `ssxm` method can return either the mean amplitude (ssAm), the mean energy (ssEm) and/or the score at percentiles `[p1,p2,...,pn]` of the amplitude and/or energy.

In the revised `ssxm` I'm implementing, the filtering is done before calling the function, so the filter/bandpass stuff is removed already. Imagine now I want the SSAM, SSEM, and the percentile 10 of each of my input traces from a stream. What would be the parameters to pass? Then how to name the returned traces? Hack the "component" of the trace id ?

so the signature of the method would be:
```python

def ssxm(window_length, amplitude=True, energy=False, percentiles=[])
```

for a 10 second windowing:
```python
#SSAM only: 
st.ssxm(window_length=10)
#SSEM only: 
st.ssxm(window_length=10, amplitude=False, energy=True)
#both:
st.ssxm(window_length=10, energy=True)

#SSAM and percentile 10
st.ssxm(window_length=10, percentiles=[10,])

#SSAM and percentile 10 and 50
st.ssxm(window_length=10, percentiles=[10,50])
```

ThomasLecocq:
runs smoooooth ...
ssxm 2

ThomasLecocq:
docs, tests & tutorials are going to be funny to do... classic...

krischer:
Should it really be attached directly to the stream and trace objects? I am not super familiar with the method but I don't think it is used enough to justify attaching it directly to the main objects.

What about

```python
from obspy.signal.ssxm import ssxm

#SSAM only: 
ssxm(st, window_length=10)
#SSEM only: 
ssxm(st, window_length=10, amplitude=False, energy=True)
#both:
ssxm(st, window_length=10, energy=True)

#SSAM and percentile 10
ssxm(st, window_length=10, percentiles=[10,])

#SSAM and percentile 10 and 50
ssxm(st, window_length=10, percentiles=[10,50])
```
And it could returned special trace objects.

```python
class SSXMTrace(obspy.coreTrace):
    ...
```

ThomasLecocq:
yeah, could be also, dunno.
currently it returns a regular Stream with Traces

ThomasLecocq:
on the contrary, I'd say that a "rolling_apply" à la `pandas` would be interesting to have in the core objects...

ThomasLecocq:
@krischer : ok modified the obspy.signal.ssxm.ssxm method to accept a Stream object and to process it from there. Once your slide methods will be availbable, I'll rebase on those and it will be good to go !

ThomasLecocq:
Guys: about the doc:

- add the docstring for the `ssxm` method in `obspy.signal.ssxm`
- add examples in `obspy.signal.__init__` too ?

krischer:
Put all docs and examples only in `obspy.signal.ssxm`; `obspy.signal.__init__` is already too full...we should probably clean it up.

ThomasLecocq:
okay ! 

I'll be cleaning things up, most probably starting fresh once the slide is merged, and most probably directly calling it `obspy.signal.noise`

QuLogic:
This PR has been moved from 0.11.0 to a future release pending updates from the original contributor.

PR #752, opened 2014-03-05
geo_interface implementation

markcwill:
This is an implementation of the `__geo_interface__` attribute proposed here:
https://gist.github.com/sgillies/2217756 for the obspy classes that have coordinates (namely, `Origin`, `Station`, plus `Event`). They return a dictionary of valid GeoJSON, which can be used by other programs. Useful for making JSON to kick out for mapping or a web app. The PyPi module `geojson` already supports this:

``` python
>>> from obspy.core.event import *
>>> import geojson
>>> o = Origin(latitude=44.5, longitude=123.4, depth=12.0, time=UTCDateTime())
>>> geojson.dumps(o)
'{"geometry": {"type": "Point", "id": "smi:local/069f5b1a-20b7-4950-9a90-bda084794efb", "coordinates": [123.4, 44.5, 12.0]}, "type": "Feature", "properties": {"updated": null, "time": 1394041177.414326}}'
```

Tests may need to be cleaned up a bit, and this could also be added to `Catalog` and `Network` by looping over its children.

### DIFF
--- a/obspy/core/event.py
+++ b/obspy/core/event.py
@@ -1754,14 +1754,14 @@ class Origin(__Origin):
             coords.append(self.depth)
         if isinstance(self.time, UTCDateTime):
             time = str(self.time)
-            coords.append(self.time.timestamp)
+            #coords.append(self.time.timestamp)  # float time as 4th coord
 
         if self.creation_info and self.creation_info.creation_time is not None:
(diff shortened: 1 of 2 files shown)


QuLogic:
This PR has been moved from 0.11.0 to a future release pending updates from the original contributor.

PR #773, opened 2014-04-10
read support for PDAS waveform files

megies:
PDAS manual mentions 4 different types of PDAS files, so far only data file is supported.


megies:
Rebased..

QuLogic:
Huh, you did that so fast that Travis broke...

megies:
;-)
..it passed before the final rebase so I didn't care to wait for Travis again. Should be fine.

PR #781, opened 2014-04-28
Kinemetrics EVT data format support.

ThomasLecocq:
Added support for the format + tests + reference PDF.


obspy/kinemetrics/evt.py
```diff
@@ -0,0 +1,411 @@
...
+        numchan = param[3]
+        num = (samplerate / 10) * numbyte * numchan
+        data = [[] for _ in range(numchan)]
+        if length != num:
+            raise EvtBadDataError("Bad data length")
+        for j in range(samplerate // 10):
+            for k in range(numchan):
+                i = (j * numchan) + k
+                if numbyte == 2:
+                    val = unpack(b">i", buff[i * 2:(i * 2) + 2] + b'\0\0')[0] \
```
QuLogic: Is this right? The method takes an `endian` argument, but doesn't seem to use it.

ThomasLecocq: @vcoeur61 : can you check this ?

vcoeur61: See my previous mail.
We have no 2 or 4 bytes files. Impossible to test.
The original code is based on kinemetrics C code and should work.
The code modified by Eliott Sales de Andrade : i am not sure, because 
the code is not // to 3 bytes code.
Perhaps ask to kinemetrics to have a 2bytes and 4bytes example code ?

Henri

megies:
Thanks @QuLogic for getting this Py3 ready. I'll merge it right now, anything that might be broken for 2/4byte files can be considered a minor bug and if necessary fixed later on.

mbyt:
@QuLogic thanks for all the Py3 work, I am really impressed. When I worked on the Py3 stuff a year ago I started https://github.com/obspy/obspy/wiki/Python-3-Pitfalls.

Could you go through it and check if it is correct? If not or non sufficient please modify. This should really helps all of us, which do not hack around in Py3 regularly.

PR #841, opened 2014-07-14
Ambient Seismic Noise in ObsPy

krischer:
This PR serves the purpose to discuss and implement what people want/expect from ObsPy in terms of handling ambient seismic noise data. I know a lot of people already have a bunch of code for this using ObsPy.

We would like to make the core of this functionality, mainly the processing and correlation handling, part of ObsPy. The goal is to have a stable and well tested toolbox for the basics of noise processing so people don't need to keep reimplementing those things.

I am not a noise person, so I don't know what exactly you require but the ideas right now are as follows:
- A set of preprocessing functions like 1-bit normalization, spectral whitening, ...
- A number of different correlation functions
- Objects to deal with CCs and CC stacks including serialization/deserialization

As a first step I propose to implement this in a `obspy.noise` submodule; we can always refactor this later on.

I'll contact a bunch of people whose github handles I don't know via email; please forward this to other people who might be interested as well.

@ThomasLecocq, @trichter, @lermert, @petrrr, ...

Please write your opinions and feel free to send PRs/directly push to this branch!

Cheers!

### DIFF
--- a/obspy/noise/correlation.py
+++ b/obspy/noise/correlation.py
@@ -1,4 +1,3 @@
-import h5py
 import numpy as np
 
 from obspy.noise.correlation_functions import phase_xcorr
@@ -42,7 +41,7 @@ class CorrelationStack(object):
         self.__locked = True
 
     def write(self, filename):
-        f = h5py.File('myfile.hdf5', 'r')
+        pass
 
 
 def correlate_trace(trace_a, trace_b, max_lag, correlation_type, **kwargs):
--- a/obspy/noise/correlation.py
+++ b/obspy/noise/correlation.py
@@ -1,0 +1,63 @@
+import h5py
+import numpy as np
+
+from obspy.noise.correlation_functions import phase_xcorr
+
+
+class Correlation(object):
+    def __init__(self, trace_a, trace_b, correlation, max_lag,
+                 correlation_type, correlation_options=None):
+        self.trace_a = trace_a
+        self.trace_b = trace_b
+        self.correlation = correlation
+        self.max_lag = max_lag
+        self.correlation_type = correlation_type
+        self.correlation_options = correlation_options \
+            if correlation_options else {}
+
+    def __add__(self, other):
+        corr = CorrelationStack()
+        corr += self
+        corr += other
+        return corr
+
+
+class CorrelationStack(object):
+    def __init__(self, correlations):
+        self.__correlations = correlations
+        self.__locked = False
+
+    def __add__(self, other):
+        if self.__locked:
+            msg = "Stack already locked."
+            raise ValueError(msg)
+        self.__correlations.append(other)
+
+    def stack(self):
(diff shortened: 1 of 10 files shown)


megies:

trichter:
Just want to say hello to this thread as I noticed on gitter that @ThomasLecocq  wants to spend some time working on this topic. +1 for adding basic noise processing in ObsPy, starting e.g. with whitening and time normalization. When it comes to cross-correlation, there are probably a lot of different codes out there, even an existing implementation in obspy.  [Here](https://gist.github.com/trichter/9607bc5c5fa792a0d6c3307355fb58d1) I put just another implementation based on scipy functions.

ThomasLecocq:
@trichter cool you're jumping in! I'll try to make a work plan by this weekend and I'll update you here ! Cheers!

krischer:
Regarding cross correlation: We can definitely (and probably should) incorporate various different ones. This'll just be handled with the plugin interface we use for all kinds of other things (data formats, tapers, filters, ...).

It makes a lot of sense to include many different variants (as long as they don't add new dependencies) into ObsPy.

megies:
Most important with new variants/implementations is to provide an usage example and/or point out the differences/ pros/cons.. (like the pick alignment stuff we do with the one time-domain xcorr we have in obspy) otherwise piling up lots of implementations might confuse users.

trichter:
I agree with @megies, that new cross correlation functions should be properly motivated. Unlike deconvolution cross correlation is a clearly defined operation. (Of course I am just talking about "normal" cross correlation.) The only thing which can be done different is padding and normalization. Maybe we can agree on a good practice to do these things?

I tried to compare output of Obspy's xcorr (xcorro) with the xcorr of the gist. Results:
* both seem to be equal for input arrays of the same length when data is demeaned beforehand
* xcorro still has problems with large lag times
* xcorro sometimes has other problems I do not understand
* xcorro seems to have problems if len(a) > len(b)

For len(a) < len(b) peaks are at different positions. Output of gist xcorr is produced by aligning the input signals at their middle for zero lag time (
```
-aaa-
bbbbb
```
) which is in line with scipy's correlate. Moreover, there is a statement in the C code of xcorro, that some adjustments made are not valid for len(a) != len(b). Instead of fixing the bugs in xcorro I'd suggest to use the scipy version of correlate with appropriate zero padding (if necessary different options for zero padding) to guarantee (a) the correct length and (b) the correct alignment of the output.

[link](https://nbviewer.jupyter.org/gist/trichter/12cdb0e2b0c4713812c517923805ce1c) to comparison

d-chambers:
@ThomasLecocq 

Since MSNoise serves the ambient noise needs for the community should we close this?

ThomasLecocq:
Yep, msnoise for full-workflow stuff, and @trichter 's obspy codes already in the core for custom/sandboxing library! Good to close.

Issue #857, opened 2014-08-05
Trace.resample not setting stats.sampling_rate correctly, if number of points is not an integer multiple of the decimation factor


kaeufl:
Maybe I'm missing something here, but it seems to me as if in Trace.resample(sampling_rate) the Trace.stats.sampling_rate is not properly set in the case that 

``` python
len(tr.data) % (tr.stats.sampling_rate / float(sampling_rate)) != 0.0
```

From the docs of scipy.signal.resample:

> The first sample of the returned vector is the same as the first
> sample of the input vector.  The spacing between samples is changed
> from dx to:
> 
>    dx \* len(x) / num

That is, as far as I understand the docs, the signal x is treated as having fixed length and resampled at num new points. If the condition len(tr.data) % factor != 0.0 is not met, then the actual sampling rate might therefore be different from the desired sampling rate.

However, in Trace.resample (https://github.com/obspy/obspy/blob/0.9.2/obspy/core/trace.py#L1468),
Trace.stats.sampling_rate is set to with whatever argument the method has been called, as far as I can see.

Thanks for clarification, if I'm wrong. Otherwise, I'd be happy to write a pull request.

s-g-mo:
I'm not sure if this has already been addressed, but something similar is occurring with the tr.decimate() function. When using this function in v.1.0.0 the new sampling rate isn't updated in the trace metadata as well. I've tried to work around it by decimating, then forcing the trace to update the stats with the new sampling rate using tr.stats.update. This does work for the current instance, however, when writing the trace out as a .SAC, the old sampling rate is written into the stats.

jkmacc-LANL:
@s-g-mo That's a bug in `io.sac`.  I'm staring right at it :disappointed:  Could you file a separate issue, and I'll take care of it?

s-g-mo:
Definitely :)

PR #859, opened 2014-08-06
Added the ability to interpolate in the frequency domain to Trace.resample(). Should fix issue #857

kaeufl:
Ok, so now here's the promised pull request. The new code uses the numpy.interp() function to interpolate the spectra in the frequency domain. I guess that's roughly what you were suggesting, Lion, right?

A few (in fact two) more thoughts:
1) I'm not sure to what extent the strict_length option makes sense. The endtime of a trace may also change if len(self.data) % factor == 0.0 holds, since the endtime is calculated based on (npts - 1)*delta. 
2) I've seen that in the current master branch there's a new signal.interpolation module, which also contains a weighted_average_slopes() function, of which I'm assuming that it does something similar to SAC's "interpolate" command. Maybe this would in general be the better solution for resample(), since this windowing + filtering business seems a little messy to me. On the other hand the "weighted average slopes" method might be messy too, so not sure...

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1580,73 +1580,92 @@ class TraceTestCase(unittest.TestCase):
         resolved.
         """
         starttime = UTCDateTime("1970-01-01T00:00:00.000000Z")
+        tr0 = Trace(np.sin(np.linspace(0, 2*np.pi, 10)),
+                    {'sampling_rate': 1.0,
+                     'starttime': starttime})
         # downsample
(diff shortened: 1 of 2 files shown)


martinvandriel:
This is just a note, that this is implemented in Instaseis by now and as far as I understand the only reason not to have moved it to obspy is the implementation in fortran.

https://github.com/krischer/instaseis/blob/master/instaseis/lanczos.py
https://github.com/krischer/instaseis/blob/master/instaseis/src/lanczos.f90

PR #865, opened 2014-08-18
obspy.fdsn complains about optional parameters not supported by server

megies:
When retrieving station inventory or waveform data from an fdsnws that does not support the optional parameters "includeavailability"/"updatedafter" for fdsnws-station and "minimumlength"/"longestonly" for fdsnws-dataselect, I get warnings like:

"The 'station' service at 'http://localhost:8080/fdsnws/1/station' cannot deal with the following required parameters: includeavailability, updatedafter
They will not be available for any requests. Any attempt to use them will result in an error."
etc.

However, these parameters are not *required* but *optional*.

Not a big issue other than a bit annoying, but perhaps it can be avoided. How about replacing the warning by an error issued only in case these parameters are actually used?


QuLogic:
I added the optional parameters so the warning shouldn't be printed. Optional parameters are printed in a separate section from "additional" parameters.

I also fixed some outdated test results that I noticed on the doc builder test results.

PR #885, opened 2014-10-08
Update spectral_estimation.py

susanacustodio:
The changes made to this code allow: 1) to store all PSD values, if wanted, so that we can later plot PPSD spectrograms; and 2) store PSD values at chosen frequencies, so that we can plot time series of PSD values at the chosen frequencies.

### DIFF
--- a/obspy/signal/spectral_estimation.py
+++ b/obspy/signal/spectral_estimation.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# -----------------------------------------------------------------------------
+#------------------------------------------------------------------------------
 # Filename: spectral_estimation.py
 #  Purpose: Various Routines Related to Spectral Estimation
 #   Author: Tobias Megies
 #    Email: tobias.megies@geophysik.uni-muenchen.de
 #
 # Copyright (C) 2011-2012 Tobias Megies
-# -----------------------------------------------------------------------------
+#------------------------------------------------------------------------------
 """
 Various Routines Related to Spectral Estimation
 
@@ -17,9 +17,6 @@ Various Routines Related to Spectral Estimation
     GNU Lesser General Public License, Version 3
     (http://www.gnu.org/copyleft/lesser.html)
 """
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
-from future.builtins import *  # NOQA
 
 import os
 import warnings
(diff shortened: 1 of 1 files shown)


susanacustodio:
I'm sorry but I never used GitHub before, so I am not quite sure how it works.
I understand that you are telling me that my changes were made on a very old spectral_estimation.py code, and that I should instead do my changes on a newer version of spectral_estimation.py. Is this correct. Can you send me or direct me to a newer version of spectral_estimation.py?
Many thanks.

krischer:
Heyhey,

you can always find the most recent version here: https://github.com/obspy/obspy/blob/master/obspy/signal/spectral_estimation.py

The easiest way for you might be to just manually apply the changes to the new version and have someone in your institute help you push it to your ObsPy fork. Or have someone help you rebase your fork to the latest master. Alternatively you could also just send me the updated version and I'll create a new PR for you.

Thanks!

susanacustodio:
Hi Lion,

Thanks! I’ll probably send you the file… that’s the easiest! :-)

S.

susanacustodio:
Hi Lion,

Here’s my modifications on top of the most recent spectral_estimation.py version (which you sent me last Friday, 30/Jan/2015).

Here’s what it does:

1. In saves the PSD amplitudes for chosen frequencies at each time step (that you specify when you call PPSD in the variable ‘store_freqs'). So, to call the function using my modified code you’d write:

       ppsd = PPSD(tr.stats, paz, skip_on_gaps=True, store_freqs=[.01, .03, .06, .14, .25, 3], write_psd=False)

2. In addition, if you specify ‘write_psd=True’ in the command above, the code also outputs all the PSDs that it computes to txt files. It outputs these in text format (in an optimisation, this should probably be stored in binary format) into a folder whose name is hard-wired as o-PSDtxt. If this folder does not exist, the code fails (this also needs to be worked…).

Now, onde you have this, you can then plot the time evolution of PSDs (power spectral densities) at the frequencies you chose in ’store-freqs’ and you can also plot ‘spectrograms’ (not real spectrograms, because they are based on the PSD amplitudes rather than actual FFT amplitudes) based on the PSDs, for any time frame of your analysis period. For this, use files noise5-plotpsd-SP.py, noise5-plotpsd-SP-time.py and noise7-plotspecgram.py attached.

You can find a folder containing an example of how this runs here:

https://dl.dropboxusercontent.com/u/9273555/noise-QC-for-ObsPy.zip

Hope this works now! Let me know if you need any more info.

Susana

PR #895, opened 2014-10-17
Matplotlib compatibility cleanup

QuLogic:
Removes a few cases of backwards-compatibility for versions of Matplotlib that we no longer support.


QuLogic:
As soon as CI passes, this should be okay.

krischer:
Looks good.

PR #899, opened 2014-10-23
Import 'B' response type (Hz) and converted it to 'A' (Laplace)

marcopovitch:
Using this patch permits to work with Poles and Zeros given in Hz (B response type in Seed documentation).

* The poles, zeros  are converted in Laplace (rad/s) - seems to be de default behavior of rdseed - 
* The normalisation factor is also scaled. 
* It applies to Parser and also to the arclink client.
* paz['response_type'] was added in the arclink client.

### DIFF
--- a/misc/docs/source/credits/CONTRIBUTORS.txt
+++ b/misc/docs/source/credits/CONTRIBUTORS.txt
@@ -9,6 +9,7 @@ Peter Danecek
 Martin van Driel
 Fabian Engels
 Sven Egdorf
+Marc Grunberg
 Conny Hammer
 Sebastian Heimann
 Lukas Heiniger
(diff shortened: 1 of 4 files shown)


megies:
Hmm.. seems this misses to test a converted PAZ against the expected values.. I'll merge it anyway (build passes on current master: https://travis-ci.org/megies/obspy/builds/48236353) to have it in 0.10.0, assuming that you've been using it for your own purposes. @marcopovitch, can you please add a test afterwards? I'll open a ticket for it..

marcopovitch:
ok I will add a test for this.

ThomasLecocq:
Thank you ! Gogogogogo ! If you do this fast you'll get access to some BE stations !

PR #900, opened 2014-10-23
read/write support for NonLinLoc formats

megies:
This PR is for..
 - writing NonLinLoc phase files from Catalog/Event (to use in NonLinLoc location runs)
 - reading NonLinLoc hypocenter-phase files into Catalog/Event files (to use internally or output to QuakeML)


megies:
Rebased..

QuLogic:
Actually, one more question: it looks like you've mocked out the `pyproj` call, so it actually necessary to add it to the test requirements? I know there's a doctest mentioning it, but we can just skip it since it doesn't actually test anything.

megies:
@QuLogic Uhm.. not sure what you mean.. but it sounds like it's no problem anymore?

QuLogic:
@megies Yep, that's what I meant. Though I didn't realize at first that the `my_conversion` function was used later on after the text. But since you've added the skip anyway, can we drop `pyproj` from our requires?

krischer:
I would leave it in the test requires as it is right now as it is required to run all tests. Or do you mean the requirements for the different linux packages?

megies:
>> But since you've added the skip anyway, can we drop pyproj from our requires?

> I would leave it in the test requires as it is right now as it is required to run all tests.

I think it should not be used anymore anywhere else, so I'll remove it from the requires.

PR #902, opened 2014-10-25
Cartopy plotting

QuLogic:
This is an experimental look at using [Cartopy](http://scitools.org.uk/cartopy/) instead of Basemap. Cartopy provides a more object oriented interface that integrates more seamlessly in matplotlib. For example, the projection is just a subclass of the regular Axes so you can do away with the transform yourself, plus the caller easily has access to the projection without having to create a new object.

This is certainly not ready for merging for a few reasons. ~~It completely replaces the existing function, so there's no fallback to Basemap, and I don't think it totally takes advantage of everything Cartopy provides. also, Cartopy doesn't yet support Python 3 (SciTools/cartopy#499).~~

The results are pretty close (where things work):
 * Default event plot: catalog-cartopy1 <br>
    is roughly the same, though I didn't add lakes or rivers. Detail is pretty close otherwise (small grid line style differences, less detail in the Arctic, and more detail in Antartica) but events are plotted in the same place.
 * Orthographic event plot: catalog-cartopy2 <br>
    is generally the same (though there's a small shift and the top of the longitude grid lines is different). (BTW, the ocean colour in this plot is uggly.)
 * Local plot: catalog-cartopy3 <br>
    ~~Cartopy doesn't appear to have this projection right now, so it's a bit worse off, unfortunately. It also doesn't seem to support the grid labels on the outside.~~


QuLogic:
I've rebased this one to take into account the slightly changed map plot settings, and to use the newly available projections in the latest Cartopy. The local plots are not quite right yet. I think it's because the Basemap version is calculated in the plotting coordinates, while the Cartopy one is calculated in lat/long, and the transform doesn't preserve the aspect ratio.

QuLogic:
Local plot should now have the correct aspect ratio; the only difference now between Basemap and Cartopy is the lack of grid labels.

QuLogic:
Besides the grid labels, I think this is pretty much done. Obviously, there are still issues with Cartopy at some rotations, but I don't think that should hold up merging. The 0.11.0 release won't be for a while, anyway.

megies:
:+1: Since this is providing an alternative way and is not interfering with the old and stable (and admittedly pretty much stale) basemap way (as long as I see), I don't see a reason why not to merge this now. (changelog!)

QuLogic:
> (changelog missing!)

Of course! Should be good now.

PR #905, opened 2014-11-06
write support for CNV event format

megies:
This PR is for writing CNV event/pick files from Catalog/Event objects (for use with VELEST program).

Still needs tests..


megies:
Added a minimal test and rebase, build passing. Merging..

PR #904, opened 2014-11-19
c_sac_taper optimization and documentation clarifications

QuLogic:
It is unclear what a "taper similar to sac" means. If it's, for example, [this taper command](http://seismolab.gso.uri.edu/~savage/sac/commands/taper.html), there are three types defined there. And the parameters to the ObsPy function do not directly correspond to this one. Also, why is it prefixed with `c_`? It's not optimized in C or anything.

(It could also do with some optimization, but I could do that myself.)


QuLogic:
Ping? Only needs the plot, really...

PR #915, opened 2014-11-19
Stop using 'jet' colour map by default

QuLogic:
I've been meaning to post this before, and [this post](http://www.climate-lab-book.ac.uk/2014/end-of-the-rainbow/) just reminded me of it. Matplotlib has been [taking its time](https://github.com/matplotlib/matplotlib/issues/875), but they have to come up with a default colour map for a wide variety of cases. Since we have so few plots, I don't see why we can't choose a sane default for them ourselves. (And waiting for matplotlib will leave out all the users with older versions anyway.)

So far, I have found the following plots that could use fixing:
 - [ ] Tutorial:
  - [x] [9 Beamforming Analysis](http://docs.obspy.org/tutorial/code_snippets/beamforming_fk_analysis.html)
   - The first scatter plot is default; second polar plot is hot_r.
  - [x] [11 Plotting Spectrograms](http://docs.obspy.org/tutorial/code_snippets/plotting_spectrograms.html)
   - Uses default.
  - [x] [23 Hierarchical Clustering](http://docs.obspy.org/tutorial/code_snippets/hierarchical_clustering.html)
   - Uses default.
   - I also don't like that the dissimilarity matrix doesn't match the colours on the dendrogram, though I don't know how easy it is to match them.
  - [ ] [24 Visualizing Probabilistic Power Spectral Density](http://docs.obspy.org/tutorial/code_snippets/probabilistic_power_spectral_density.html)
   - This seems to be a made-up colour map, though it seems pretty similar to rainbow.
  - [x] [25 Array Response Function](http://docs.obspy.org/tutorial/code_snippets/array_response_function.html)
   - Very obviously the default, and it doesn't look good.
  - [x] [26 Continuous Wavelet Transform](http://docs.obspy.org/tutorial/code_snippets/continuous_wavelet_transform.html)
   - All plots use the default.
  - [ ] [27 Time Frequency Misfit](http://docs.obspy.org/tutorial/code_snippets/time_frequency_misfit.html#plot-the-time-frequency-representation)
   - Tfr seems to be a different colour map from the rest, hand made, and not even the same as the original citations.
    - TfMisfits could maybe be replaced with RdYlBu_r, since it's divergent.
    - I'm not sure why TfGofs uses a reversed colour map compared to the others.
 - [ ] Basemaps:
  - [ ] [Catalog plots](http://docs.obspy.org/master/packages/autogen/obspy.core.event.Catalog.plot.html)
   - The examples use depth-colouring red-yellow-green, which is not good for colour-blindness.
   - Colouring by date will use the default colour map according to docs, though I don't think this is true.
  - [x] [Inventory plots](http://docs.obspy.org/master/packages/autogen/obspy.station.inventory.Inventory.plot.html)
   - Uses jet by default, though it tries to turn it into categories.
   - Changed to use the "Paired" colour map from ColorBrewer. This makes more sense since it's really a categorical designation. Additionally, the single colouring uses the colour from Paired with lut=1, instead of blue.
  - [x] [Network plots](http://docs.obspy.org/master/packages/autogen/obspy.station.network.Network.plot.html)
   - Might not be needed for anything.
   - The default colour is changed like the previous one. No colour map involved.

This PR is not nearly done, so suggestions are welcome.


QuLogic:
Rebased. Feel free to comment some more.

megies:
Just built docs pages based on the current state: http://www.geophysik.uni-muenchen.de/~megies/obspy-docs-colormap-change

Trying to summarize unresolved points:
 - my main concern: having a colormap for scatter plots which runs through whitish colors is problematic because some data points might not be visible (http://www.geophysik.uni-muenchen.de/~megies/obspy-docs-colormap-change/_images/beamforming_fk_analysis_11.png). what makes it even worse is that those data points are the points associated with high data values..
 - @krischer was criticizing the colormap for quake depth (@krischer maybe you have some links to plots at 3rd party seismo sites that show the general usage of the colors you mentioned for event depth?)

These two points aside it looks really good to me.

(We should change one of the three catalog example plots to use origin time for color coding..)

krischer:
I agree with the scatterplots concerns of @megies.

Regarding the colormap for earthquake depth. There are a couple examples at the USGS, e.g. this one: http://earthquake.usgs.gov/earthquakes/world/seismicity_maps/ The colorscale here is red for 0-69 km depth, green for 70-299 km. depth and blue for 300+ km depth.

One more question: If we have a new colormap, should we have like an "ObsPy Colormap" that is recognizable? Just so people looking at a plot can tell that it comes from ObsPy. There are a couple of nice ones here: http://nbviewer.ipython.org/github/pierre-haessig/lab-colormap-creator/blob/master/test_plots.ipynb

I'll take the liberty and put 4 up for voting. They all have end members that more or less correspond to the Python colors and go through green for some hue variation.

1. screen shot 2015-04-27 at 16 56 35
2. screen shot 2015-04-27 at 16 56 27
3. screen shot 2015-04-27 at 16 56 38
4. screen shot 2015-04-27 at 16 56 42

My vote goes to the second option. Other opinions?

That's only for the sequential colormaps. For the diverging we still need something but the currently used on is fine I guess (except the egg-shell color of the zero value which we should fix as discussed above). Its also only for the TF misfit.

QuLogic:
> The colorscale here is red for 0-69 km depth, green for 70-299 km. depth and blue for 300+ km depth.

Right, so I wasn't against Red/Green if it's semi-standard, but should we make it segmented like this, or continuous?

For the sequential examples, I will check on a couple other screens tomorrow, but the first two appear to have a kink at the green point, where it seems a bit disjoint. The fourth one also seems to have a little kink right in the middle. So right now, the third seems smoothest. BTW, did you try them out in greyscale?

endolith:
> the first two appear to have a kink at the green point, where it seems a bit disjoint. The fourth one also seems to have a little kink right in the middle. So right now, the third seems smoothest.

These are points where the curve takes a sharp angle through the colorspace:

index

If you're creating a new colormap from scratch, you should make sure the curve is smooth, probably working in Lch space.

http://graphicdesign.stackexchange.com/q/16246/10296

https://gist.github.com/endolith/74275dc8fa2bb9a78266

krischer:
@endlith Thanks for the explanation. Makes a lot of sense.

So should we choose the third option which as correctly pointed out by @QuLogic looks smoothest. It also is a smooth curve in Lab color space which is not Lch but I guess its good enough for us.


screen shot 2015-04-27 at 16 56 38

Looks also good in B&W.
screen shot 2015-04-27 at 16 56 38

screen shot 2015-04-29 at 11 52 54

megies:
That would make the following plot..
figure_1

krischer:
Cool. I did not expect it to work with this particular plot but that looks good enough for me.

endolith:
>So should we choose the third option which as correctly pointed out by @QuLogic looks smoothest? 

Well I was saying they could *all* be made smooth by using a spline or something to connect the 4 points instead of linear interpolation, or just choosing more than 4 points along a smooth curve by hand.

> It also is a smooth curve in Lab color space which is not Lch but I guess its good enough for us.

Lch is the same space as Lab, just using cylindrical coordinates instead of rectangular.  (lightness, chroma, hue) I guess it would change the interpolation though?

megies:
Here's how plots would look like (complete docs build) if we replace the sequential colormap with the whitish endmember by the above one with yellow endmember:  https://www.geophysik.uni-muenchen.de/~megies/obspy-docs-colormap-change_2/gallery.html

QuLogic:
Well, it looks pretty good in general. But now we come to the most difficult problem... what do we call it?

megies:
> Well, it looks pretty good in general. But now we come to the most difficult problem... what do we call it?

Ok, I didn't want to "overwrite" changes in this PR right away, but I guess that makes things more complicated, so I add the changes for the above docs build for more transparency in the discussion..

QuLogic:
OK, there are three things left now to confirm:
 * Tutorial:
  *  24 Visualizing Probabilistic Power Spectral Density
   * I assume we aren't changing this one as it matches IRIS and others.
  *  27 Time Frequency Misfit
   * These all use the ObsPy defaults now; just want to be sure about that.
 * Basemaps:
  * Catalog plots
   * I think you'd like Red/Green/Blue for depth at least, but any specific categories (depth splits where?)?

> (We should change one of the three catalog example plots to use origin time for color coding..)

I changed the second plot to do so.

megies:
> 24 Visualizing Probabilistic Power Spectral Density

I think it should stay with the old default which comes from the original McNamara paper/software and is well known in the community. We could add a `cmap` kwarg to the plot call though, to make it accessible.

QuLogic:
Matplotlib has a [new evaluation of colour maps](https://bids.github.io/colormap/) going on. For comparison, here's what this one looks like:
obspy-sequential
There are a couple of large(ish) jumps in perceptual delta, but they aren't too evident to me here. It's pretty nicely equivalent in colour-blind mode, too. This one is similar to matplotlib's option D...

megies:
Nice!
Well, we do have some jumps in there.. but compared to the [graphs for jet](https://bids.github.io/colormap/images/screenshots/jet.png) they are much smaller, although I have no idea what the units are or mean. I'd say both what we have now and this option D would work for us (in case matplotlib settles on option D -- which could take quite a while I guess).

QuLogic:
Based on [the SciPy 2015 video](https://www.youtube.com/watch?v=xAoljeRJ3lU), I think matplotlib did settle on their option D.

megies:
Here's our docs built with..
 - [our own current favorite](https://www.geophysik.uni-muenchen.de/%7Emegies/obspy-docs-colormap-change_2/gallery.html)
 - [matplotlib's option D (i.e. mpl's likely new default)](https://www.geophysik.uni-muenchen.de/%7Emegies/obspy-docs-colormap-change_optionD/gallery.html)

I think the final decision for the sequential colormap is between the two (I don't think mpl will mind if we hardcode-copy their colormap).

The only other issue left is whether we want the divergent colormap to have white or this yellowish color at zero level.

QuLogic:
So I can see a faint, but pretty evident green path around most smooth gradients. The Array Response Function (Tutorial Ch. 25) and TF Misfits (Tutorial Ch. 27) show it most prominently. Those lines go away with viridis.

krischer:
While I also find our own current favorite aesthetically a tiny bit more pleasing (I can really only tell them apart if viewed right next to each other...) I would vote for the viridis colormap as its apparently just a strictly better colormap.

Also should be we maybe invert the colorscale for the polar fk plot? Then it will be mostly light yellow with the high values a dark blue. The grid should be more visible this way.

endolith:
> There are a couple of large(ish) jumps in perceptual delta, but they aren't too evident to me here.

Compare https://cloud.githubusercontent.com/assets/302469/8269047/42abb430-1766-11e5-8b04-eaa162cf134c.png and http://bids.github.io/colormap/images/screenshots/option_d.png  The obspy one has a halo of green around the mountain top that looks a little out of place, while the mpl one is more smoothly blended from blue to green to yellow.  Their [viscm tool](https://github.com/matplotlib/viscm) has an editor that will automatically spline the colormap in their preferred colorspace (supposed to be better than Lab), can you make a similar map to this one but with smooth curves?

QuLogic:
> Also should be we maybe invert the colorscale for the polar fk plot? Then it will be mostly light yellow with the high values a dark blue. The grid should be more visible this way.

I'm not sure this looks as good as you expect it might:
figure_1

> The obspy one has a halo of green around the mountain top that looks a little out of place

By itself, it's hard to see. If you switch between the two images you linked, it becomes more obvious. It also seemed more obvious in the examples I mentioned.

I would be fine with just using viridis directly.

megies:
> Also should be we maybe invert the colorscale for the polar fk plot?

Totally -1 from my side. Light colors should represent high values. Why invert this for one particular plot? The grid is still visible in my opinion, and in areas with dark blue colors (i.e. no energy) there is not that much need to see the grid. It is well visible in the interesting areas, I would say.


Alright, looks like we have no votes against settling on matplotlibs Option D. Should we wait for an official statement from matplotlib that they settled for option D?

QuLogic:
They've already decided; it's in the video and [here](http://matplotlib.org/style_changes.html).

QuLogic:
There's the updated results. I'd probably squash before merging to leave out the extra binary diffs.

megies:
Ok, great! I rebased on current master and squashed a couple of minor commits getting rid of some binary diffs.

Only major open point left (the event depth discussion died out I think..?) is whether to go for `RdYlBu` or `RdBu`:

time_frequency_misfit_ex5_00 hiresx time_frequency_misfit_ex5_01 hiresx

time_frequency_misfit_ex5_00 hires
time_frequency_misfit_ex5_01 hires


I tend to agree with @krischer, that the yellow looks a bit old fashioned and a whitish color for zero levels might be a bit more intuitive..

QuLogic:
The original custom colour map was pure white in the middle, which was the main concern I had. It made it seem as if there were nothing there on that one plot (not pictured above.) But `RdBu` has a slight tint to it which mitigates that problem, and I agree it looks nicer.

krischer:
I tried to use the tool the matplotlib guys used to design viridis to generate a perfect diverging colormap but I failed. Is this possible with the tool? Maybe I am using it wrong. Otherwise: lets merge this :-)

megies:
> I tried to use the tool the matplotlib guys used to design viridis to generate a perfect diverging colormap but I failed.

Yeah, just quickly tried too, but couldnt get the tool in edit mode to start with `RdBu` and gave up after a few minutes..

megies:
Hmm.. for the divergent, should we go with red for high values like in the earlier state with yellowish zero-level? Right now I only can think of seismic velocities, and isn't fast usually cold/blue?

QuLogic:
You have any other papers or software that generates the same plots to use as reference? We should probably choose the least surprising one...

krischer:
The paper for the tf misfits uses red for positive and blue for negative values.

screen shot 2015-07-22 at 09 49 40

megies:
Ok, so definitively red for positive values in TF misfits. But question is still whether in general we want red for positive?
Either way, we can add a `obspy_divergent_r` colormap with reversed colors.

megies:
?Good to merge, I guess

QuLogic:
:ship:

ThomasLecocq:
Wawouuuuuuuuuuuuuuu !! Good jobs guys !!!

PR #916, opened 2014-11-20
PPSD: add a switch for hydrophone data

megies:
... and unify with special handling of rotational data (from ringlaser).

I was asked whether we could add support for hydrophones (omitting the differentiation in processing), so here goes. There could be better / more canonical ways to do this.. any other suggestions welcome. One downside of unifying the handling of ring laser data could be problems when loading an old pickle. I'm trying to avoid those problems but this should be tested..

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -85,6 +85,11 @@ master:
      - new option `PPSD.plot(..., cumulative=True)` for a cumulative plot of
        the histogram, i.e. a non-exceedence percentage visualization, similar
        to the `percentile` option.
+     - changes to special handling of rotational: now handled by kwarg
+       `special_handling="ringlaser"` (kwarg `is_rotational_data` is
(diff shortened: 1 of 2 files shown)


QuLogic:
Should you not add a `deprecated_keyword` decorator for `is_rotational_data`?

megies:
Need to get this inline with #1108...

megies:
> Should you not add a deprecated_keyword decorator for is_rotational_data?

It's handled in a more custom fashion inside the routine as the decorator can't map from `is_rotational_data=True` to `special_handling=ringlaser`.

megies:
@sstaehler, since this was originally prompted by you.. can you maybe comment on whether this still meets the hydrophone data needs (there was a change to the response removal with #1108)?

megies:
Don't have test data, so including without test. We can add some if people using this report back.

Issue #576, opened 2013-05-24
obspy.signal dependence on matplotlib


claudiodsf:
Currently `obspy.signal` depends on matplotlib (and imports matplotlib modules), due to `spectral_estimation.py`.

This creates some problem for people, like me, who need some specific module from `obspy.signal` (like, in my case `obspy.signal.invsim.cosTaper`), but, at the same time, need to customize matplotlib configuration.

For instance, I don't want to open a graphical window, so I need to do:

``` python
import matplotlib
matplotlib.use('agg')
```

This fails with the following warning:

```
/usr/local/lib/python2.7/site-packages/matplotlib/__init__.py:1042: UserWarning:  This call to matplotlib.use() has no effect
because the the backend has already been chosen;
matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.
```

Since matplotlib is already imported by `obspy.signal`.

There is a workaround to that, which is to unload and reload matplotlib (http://stackoverflow.com/questions/3285193/how-to-switch-backends-in-matplotlib-python).

However, it would be better, if possible to split out matplotlib dependence from `obspy.signal`.
In principle, the only module expected to depend on matplotlib should be `obspy.imaging`.

PR #929, opened 2015-01-19
Image comparison

krischer:
Work in progress...don't merge yet!

This PR replaces the import of matplotlib's `compare_images()` function with a modified copy of the version from the latest matplotlib release. This restores some sanity to the imaging tests. The output of different matplotlib versions might differ slightly but now we at least compare the pictures using the same metric!

Previously we had custom baseline tolerances for the different matplotlib versions while still having version dependent relative tolerances in quite a lot of tests! Totally crazy to be honest ;-)

This furthermore means that `nose` is no longer required and the `HAS_COMPARE_IMAGES` test skipping condition is now obsolete thus I removed it.

Also `obspy.signal.spectral_estimation` no longer has the nice failure mode when matplotlib is not present. Matplotlib has been a hard depedency of ObsPy for quite a while now.


mbyt:
+1 for one dependency (nose) less, great.

Even with adding a copied version of compare_images the overall LOC statistic of this Pull request are: "174 additions and 274 deletions." Respect!

QuLogic:
Have you considered turning it into a decorator? That should give you the autoskip behaviour when matplotlib isn't installed. It'd be a bit of a pain converting all the `with`s though.

krischer:
Turning what into a decorator? I'm not entirely sure I understand.

Other than that I would totally consider matplotlib a hard dependency of ObsPy and tests just will not run without it.

QuLogic:
The `ImageComparison` context. But that's just a bunch of tedious re-factoring of tests, so it's understandable if you don't want to do it. The only real benefit is adding skips for missing matplotlib. But I'm quite fine with it being a hard dependency for tests.

krischer:
I see. Yea that would be a bit nicer in most circumstances but you are right: too much work without any real gain...

obspy/core/util/testing.py
```diff
@@ -113,23 +114,113 @@ def add_doctests(testsuite, module_name):
...
+    This enable ObsPy to have the same image comparison metrics across
+    matplotlib versions. Futhermore nose is no longer a test dependency of
+    ObsPy.
+
+    Only works with png files in contrast to the matplotlib version.
+
+    Compare two "image" files checking differences within a tolerance.
+
+    Parameters
+    ----------
```
QuLogic: This is using the Google style of docstring, I guess. So it'll probably cause a bunch of warnings in Sphinx.

obspy/core/util/testing.py
```diff
@@ -113,23 +114,113 @@ def add_doctests(testsuite, module_name):
...
+    tol : float
+        The tolerance (a color value difference, where 255 is the
+        maximal difference).  The test fails if the average pixel
+        difference is greater than this value.
+
+    Example
+    -------
+    img1 = "./baseline/plot.png"
+    img2 = "./output/plot.png"
+    compare_images( img1, img2, 0.001 ):
```
QuLogic: No trailing colon?

obspy/core/util/testing.py
```diff
@@ -113,23 +114,113 @@ def add_doctests(testsuite, module_name):
...
+    """
+    from matplotlib import _png
```
QuLogic: Is `matplotlib._png` going to be stable? We could probably still use Pillow otherwise.

krischer: I actually ended up writing my own png writing function to work across matplotlib function...I'll push tomorrow.

obspy/core/util/testing.py
```diff
@@ -113,23 +114,113 @@ def add_doctests(testsuite, module_name):
...
+    actualImage = np.uint8(_png.read_png(actual) * 255.0)
+    expectedImage = expectedImage[:, :, :3]
+    actualImage = actualImage[:, :, :3]
+
+    # convert to signed integers, so that the images can be subtracted without
+    # overflow
+    expectedImage = expectedImage.astype(np.int16)
+    actualImage = actualImage.astype(np.int16)
+
+    "Calculate the per-pixel errors, then compute the root mean square error."
```
QuLogic: I guess you merged the function, so this docstring should be a comment now.

krischer: totally...my local version is actually already very different as lots of issues started to show up.

obspy/core/util/testing.py
```diff
@@ -192,7 +283,7 @@ def __init__(self, image_path, image_name, reltol=1, *args, **kwargs):
         self.keep_only_failed = "OBSPY_KEEP_ONLY_FAILED_IMAGES" in os.environ
         self.output_path = os.path.join(image_path, "testrun")
         self.diff_filename = "-failed-diff.".join(self.name.rsplit(".", 1))
-        self.tol = get_matplotlib_defaul_tolerance() * reltol
+        self.tol = reltol * 3.0
```
QuLogic: So what required the increased tolerance?

obspy/imaging/tests/test_scan.py
```diff
@@ -109,8 +91,8 @@ def test_multipleSamplingrates(self):
                         fp.flush()
                         fp.seek(0)
                         files.append(fp.name)
-                    with ImageComparison(self.path, 'scan_mult_sampl.png',
-                                         reltol=reltol) as ic:
+                    with ImageComparison(self.path, 'scan_mult_sampl.png') \
```
QuLogic: Might as well correct the file name to `scan_multi_sampling` or something, since it's too long for here anyway.

obspy/imaging/waveform.py
```diff
@@ -38,9 +38,6 @@
 import warnings
 
 
-MATPLOTLIB_VERSION = getMatplotlibVersion()
```
QuLogic: Huh, was this unused?

QuLogic:
BTW, since we have our own comparison function now, maybe we can fix this one bug that bothers me. When you pass `--keep-images`, the actual and diff lines show the temporary `/tmp/obspy-*.png` path. It would be better if they showed the kept location `obspy/*/tests/images/testrun/*.png`

megies:
+1 for cleaning up the image comparisons. thanks for the work @krischer.

obspy/core/util/testing.py
krischer: An RMSE of 3.0 essentially enables it to run across 1.3.x and 1.4.x as far as I can tell. For matplotlib versions smaller then 1.3, the tolerance is a lot higher. This assures the images are correct for newer matplotlib versions and for older ones it tests that the images get created and are not outrageously different.

I think this strike a good balance between test coverage and maintainability.

obspy/imaging/waveform.py
krischer: Not sure actually. Maybe flake8 has some exception for unused all uppercase names?

QuLogic:
There's one more mention of `nose` in `obspy/core/scripts/runtests.py`.

krischer:
So this turned out to be much longer then expected...in summary this pull request:

* Adds a couple more Docker images
* All tests pass on all Docker images
* Adds a custom `compare_images()` function
* The minimal supported numpy version is not officially 1.4.0

Ready to review in my opinion. I'll merge tomorrow if no objections arise.

misc/docker_tests/base_images/fedora_20/Dockerfile
```diff
@@ -2,8 +2,11 @@ FROM fedora:20
 
 MAINTAINER Lion Krischer
 
-RUN rpm -Uvh http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-2.noarch.rpm
+RUN rpm -Uvh http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
```
QuLogic: So, this is weird. I don't know why you'd enable EPEL repos on Fedora. I don't think anyone would normally do so. Usually, they'd be out of date.

misc/docker_tests/base_images/fedora_21/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM fedora:21
+
+MAINTAINER Lion Krischer
+
+RUN rpm -Uvh http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
```
QuLogic: Same thing here.

misc/docker_tests/base_images/fedora_21/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+# Can fail on occasion.
+RUN yum -y upgrade || true
+RUN yum install -y numpy scipy gcc-gfortran python-matplotlib python-sqlalchemy python-lxml python-mock python-basemap python-basemap-data python-tornado python-pip
+RUN pip install flake8 suds-jurko future
+# Fedora has a slightly messed up basemap installation.
+RUN mkdir -p /usr/lib64/python2.7/site-packages/mpl_toolkits/basemap/data/
+RUN mv /usr/share/basemap/* /usr/lib64/python2.7/site-packages/mpl_toolkits/basemap/data/
```
QuLogic: The fix for this went into testing on the 15, so it should be generally available around the 22/23 and we can remove this bit then.

obspy/core/tests/test_event.py
```diff
@@ -7,13 +7,16 @@
...
-from obspy.core.util.testing import ImageComparison, HAS_COMPARE_IMAGE
+from obspy.core.util.base import getBasemapVersion
+from obspy.core.util.testing import ImageComparison
 from obspy.core.util.decorator import skipIf
 import os
 import sys
 import unittest
 import warnings
 
+BASEMAP_VERSION = getBasemapVersion()
```
QuLogic: Huh, shouldn't more newlines be required around this? Also, it should go after the below import. In fact, it could replace that import, couldn't it?

obspy/core/util/testing.py
QuLogic: Ah, I see this "fixes" the station response plots, but those are legitimately broken (just not in ObsPy.)

obspy/imaging/maps.py
```diff
@@ -144,7 +151,6 @@ def plot_basemap(lons, lats, size, color, labels=None,
     if show_colorbar:
         map_ax = fig.add_axes([ax_x0, 0.13, ax_width, 0.77])
         cm_ax = fig.add_axes([ax_x0, 0.05, ax_width, 0.05])
-        plt.sca(map_ax)
```
QuLogic: I remember looking for these and _not_ doing this one for some reason, but I don't seem to recall what it was.

obspy/imaging/scripts/scan.py
```diff
@@ -73,23 +73,23 @@ def compressStartend(x, stop_iteration):
 
 
 def parse_file_to_dict(data_dict, samp_int_dict, file, counter, format=None,
-                       verbose=False, ignore_links=False):
+                       verbose=False, quiet=False, ignore_links=False):
     from matplotlib.dates import date2num
     if ignore_links and os.path.islink(file):
-        print("Ignoring symlink: %s" % (file))
+        if verbose or not quiet:
```
QuLogic: This changes default behaviour, no?

QuLogic:
> BTW, since we have our own comparison function now, maybe we can fix this one bug that bothers me. When you pass `--keep-images`, the actual and diff lines show the temporary `/tmp/obspy-*.png` path. It would be better if they showed the kept location `obspy/*/tests/images/testrun/*.png`

Were you able to fix this one too?

QuLogic:
Also, these dependency changes should probably go in the change log.

misc/docker_tests/base_images/fedora_20/Dockerfile
krischer: I am entirely unfamiliar with the redhat/centos/fedora ecosystem. I think I had to use it for centos to get the Python packges. So do I only need it for CentOS and not for Fedora?

obspy/imaging/maps.py
krischer: There was an issue with the map being plotted inside the colorbar axes when just replacing `plt.sca()` in all occurences. This has been fixed by manually assigning the axes of the map. The plots seem to end up correct.

obspy/imaging/scripts/scan.py
krischer: No. By default, both `verbose` and `quiet` are False. Thus the conditional evaluates to True and the default behavior is retained. The `verbose` flag furthermore overwrites the `quiet` flag.

krischer:
> BTW, since we have our own comparison function now, maybe we can fix this one bug that bothers me. When you pass --keep-images, the actual and diff lines show the temporary /tmp/obspy-*.png path. It would be better if they showed the kept location obspy/*/tests/images/testrun/*.png
> 
> Were you able to fix this one too?

I forgot about that. I'll try my best...this would definitely be useful!

misc/docker_tests/base_images/fedora_20/Dockerfile
QuLogic: CentOS is like a really long LTS. It doesn't make sense to install packages from it on Fedora, which is bleeding-edge.

obspy/imaging/scripts/scan.py
QuLogic: Oh, I misread somehow. For some reason, I thought it originally had a `if verbose:` there.

krischer:
Now raises a beautiful error message with correct paths to the images.

```python
FAIL: test_Beachball (__main__.BeachballTestCase)
  File "test_beachball.py", line 79, in test_Beachball
    Beachball(data_, facecolor="yellow", outfile=ic.name)
  File "/Users/lion/workspace/code/obspy/obspy/core/util/testing.py", line 398, in __exit__
    raise ImageComparisonException("\n".join([msg, upload_msg]))
obspy.core.util.testing.ImageComparisonException: Image comparision failed.
	Expected:  /Users/lion/workspace/code/obspy/obspy/imaging/tests/images/bb_sumatra_mt.png
	Actual:    /Users/lion/workspace/code/obspy/obspy/imaging/tests/images/testrun/bb_sumatra_mt.png
	Diff:      /Users/lion/workspace/code/obspy/obspy/imaging/tests/images/testrun/bb_sumatra_mt-failed-diff.png
	RMS:       130.00817205083686
	Tolerance: 3.0
```

obspy/imaging/maps.py
QuLogic: That might have been it. But if it works, it works!

QuLogic:
It looks like matplotlib is required for `obspy.signal` now, but it doesn't seem to be mentioned in the changelog. Also, I don't know if people from #576 are okay with it.

claudiodsf:
@QuLogic I'm not really happy to see matplotlib dependence into `obspy.signal`. As it was discussed in #576, in principle the only module requiring matplotlib should be `obspy.imaging`. 
But I can live with it, since I have not so much time to propose an alternative :wink:

megies:
Pleaaaaaase.. not the matplotlib thingy again.. ;)
The matplotlib dependency has been discussed for years. De facto matplotlib has been a dependency for ages. The dependency was finally made clear two years ago (!!), see setup.py and e7f3659a.

PR #930, opened 2015-01-20
Releases

aringler-usgs:
I have switched the PPSD routine to do the response removal in the frequency domain.  This fixes the problem of the wrong amplitude at very long periods (1000 s and more) when doing PPSDs.  I have modified the PPSD tests to be tested on the same data, but using this modified algorithm. 

Please let me know if I did any of this wrong or you would like me to do this some other way.

Thanks!
Adam


krischer:
Hi Adam,

can you please open a new pull requests towards the ObsPy `master` branch? Thanks! Then we can properly review your changes.

You sent a pull `request` to the releases branch which only works if you also forked from the `releases` branch. (Just select the master branch when choosing where to send the pull requests)

All the best,

Lion

Issue #743, opened 2014-02-26
test CatchOutput fails on MacOSX with Python 3.3


petrrr:
On Mac OS X 10.8 using the Macports Python 3.3.4 interpreter this test fails. 

Interestingly, this test seems to introduces some special characters into the report, this causes the reporter to fail on submission as well. (The later is documented here: https://github.com/obspy/reporter/issues/8)

```
======================================================================
FAIL: test_CatchOutput (obspy.core.tests.test_util_misc.UtilMiscTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/obspy/core/tests/test_util_misc.py", line 78, in test_CatchOutput
    self.assertEqual(out.stdout, "abc\nghi\njkl\n")
AssertionError: 'abc\n\x1b[?1034hdef\nghi\njkl\n' != 'abc\nghi\njkl\n'
  abc
- def
  ghi
  jkl
```

PR #932, opened 2015-01-20
Fix CatchOutput context manager

QuLogic:
`CatchOutput` corrupts some file I/O. This didn't seem to be a problem before, but matplotlib 1.4.2 changed the way they did I/O to fonts, and it seems to have triggered this bug. I have added a test case that should work similarly to what matplotlib does.

As far as I can tell, this change works just fine. Because of the `finally` clause, I _think_ there shouldn't be any new leaks. Please test though.

### DIFF
--- a/obspy/core/tests/test_util_misc.py
+++ b/obspy/core/tests/test_util_misc.py
@@ -11,6 +11,7 @@ from obspy.core.util.decorator import skipIf
 import os
 import platform
 import sys
+import tempfile
 import unittest
 
 
@@ -84,6 +85,18 @@ class UtilMiscTestCase(unittest.TestCase):
             else:
                 self.assertEqual(out.stdout, b"abc\nghi\njkl\n")
                 self.assertEqual(out.stderr, b"123\n456\n")
+
+    def test_CatchOutput_IO(self):
+        with CatchOutput() as out:
+            fn = tempfile.TemporaryFile(prefix='obspy')
+
+        try:
+            fn.write(b'abc')
+            fn.seek(0)
+            fn.read(3)
+            fn.close()
+        except OSError as e:
+            self.fail('CatchOutput has broken file I/O!\n' + str(e))
 
     def test_no_obspy_imports(self):
         """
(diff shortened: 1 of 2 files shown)


QuLogic:
Here's what I did for a test:
```
>>> import os
>>> os.getpid()
12345
>>> while True:
...:     with CatchOutput():
...:         pass
...:     
```

And in another shell, run `watch 'lsof -p <pid> | wc -l'`. Every two seconds or so it prints out the count of open files, which is ~100 on my system. So it seems like there are no leaks.

QuLogic:
Or something that should test most of `CatchOutput`'s code paths:
```python
while True:
    try:
        with CatchOutput():
            if random.randint(0, 1):
                print('x')
            if random.randint(0, 1):
                print('x', file=sys.stderr)
            if random.randint(0, 1):
                raise NotImplementedError('exception')
    except NotImplementedError:
        pass
```

krischer:
Can you try to print something in C, like the other `CatchOutput` manager test?

QuLogic:
OK, really comprehensive, eh?
```python
while True:
    try:
        with CatchOutput():
            if random.randint(0, 1):
                print('x')
            if random.randint(0, 1):
                print('x', file=sys.stderr)
            if random.randint(0, 1):
                os.system('echo "abc"')
            if random.randint(0, 1):
                libc.printf(b"def\n")
            if random.randint(0, 1):
                os.system('echo "123" 1>&2')
            if random.randint(0, 1):
                print("jkl", file=sys.stdout)
            if random.randint(0, 1):
                raise NotImplementedError('exception')
    except NotImplementedError:
        pass
```

Open files are about 98-104, but not leaking. Also, no printing occurs.

krischer:
Great!

krischer:
Does not work on OSX and Python 3:

```python
FAIL: test_CatchOutput (obspy.core.tests.test_util_misc.UtilMiscTestCase)
  File "/Users/lion/workspace/code/obspy/obspy/core/tests/test_util_misc.py", line 85, in test_CatchOutput
    self.assertEqual(out.stdout, b"abc\nghi\njkl\n")
AssertionError: b'abc\n\x1b[?1034hdef\nghi\njkl\n' != b'abc\nghi\njkl\n'

```

QuLogic:
*Sigh*, according to #743, OSX was always broken?

Issue #926, opened 2014-12-15
Unclear starttime behavior using section plot and sac files


Permian:
I've created a section plot using sacfiles, where the nzhour, etc. (or kztime) are equal to the origin time. The origin marker was set to zero.
It turned out that the starttimes used by obspy are different from the sac times, i.e. shifted by the sac header variable 'B'. Therefore, the travel times do not increase with epicentral distance. 

The correction for the starttime resulted in two problems (see full script below):

1.When I use the trim command
   `tr.trim(tr.stats.starttime-tr.stats.sac.b)`
   then the starttimes are equal to the origin time and the waves increase correctly with the distance. 
   BUT when I overwrite the trace starttimes using
  `tr.stats.starttime=tr.stats.starttime-tr.stats.sac.b`
   the traces are still shifted by the B value in the plot although the starttimes were correctly set.
   Why are the overwritten starttimes not recognized by section plot?
   What is the difference using trim command?   

2.Consequently, when I set the B variable to zero in the sac headers the starttime issue should be avoided. But this is not the case!
Indeed, the obspy starttimes are correct (=origin time) BUT the traces in the plot are still shifted by a time shift that equals the previous B value (which was set to zero)! 
How this can be?? This would mean that the B variable is also set somewhere else in the sac header and used by obspy?! 
I already have looked into the source codes of sacio and waveform plotting but I couldn't find an explanation for this curious starttime behaviors.

```javascript
st=read('*BHZ*sac')

print "starttime before correction"
for tr in st:
     print tr.stats.station, tr.stats.starttime, tr.stats.sac.b

## epicenter
eq_lat = st[0].stats.sac.evla
eq_lon = st[0].stats.sac.evlo
evdp = st[0].stats.sac.evdp
mag = st[0].stats.sac.mag
evtime = UTCDateTime('2008-06-13,23:43:46.79')  ## origin time


for tr in st:
        tr.stats.coordinates = ({'latitude': tr.stats.sac.stla, 'longitude': tr.stats.sac.stlo})
        tr.stats.distance = gps2DistAzimuth(tr.stats.coordinates.latitude,tr.stats.coordinates.longitude,eq_lat,eq_lon)[0]
        tr.trim(tr.stats.starttime-tr.stats.sac.b)   ### works
        #tr.stats.starttime=tr.stats.starttime-tr.stats.sac.b   ### does not work, ignored by section plot
        

print "starttime after correction"
for tr in st:
[...]

Issue #913, opened 2014-11-19
Record Section Plot Question/Bug?


krischer:
From the mailing list (just reformatted a bit): http://lists.swapbytes.de/archives/obspy-users/2014-November/001581.html

Dear All,

Having three issues with stream plot `st.plot(type='section')` in FC20 linux, with obspy0.9.2

```python
>>> import obspy
>>> print obspy.__version__
0.9.2
```

1. The traces are assumed to have the same start time. (There is no relative shifting between traces, as in `plot(type='normal')`. Easy to handle if known, but If this is a feature, not a bug, it would be good to state this in the docs?

2. scaling/normalization: in `plot(type='normal')` the parameter is `equal_scale`. `equal_scale` works fine, but In `plot(type='section')` the parameter is norm. (Why not use `equal_scale` in section?) But I can't get norm to do anything, despite:

```python
def __sectNormalizeTraces(self):
        """
        This helper function normalizes the traces
        """
        self._tr_normfac = np.ones(self._tr_num)
        if self.sect_norm_method == 'trace':
            # Normalize against each traces' maximum
            for tr in range(self._tr_num):
                self._tr_normfac[tr] = np.abs(self._tr_data[tr]).max()
        elif self.sect_norm_method == 'stream':
            # Normalize the whole stream
            self._tr_normfac.fill(self._tr_max_count_glob)
        else:
            msg = 'Define a normalisation method. Valid normalisations' +\
                'are \'trace\', \'stream\'. See documentation.'
            raise ValueError(msg)

        self._plot_init = False
```

but given the fact that I don't get msg even if I say: 

```python
plot(type='section', norm='bla') 
```

makes me think that it is not used at all. I don't understand this because it appears to be called here:

```python
def plotSection(self, *args, **kwargs):
        """
        Plots multiple waveforms as a record section on a single plot.
        """
        # Initialise data and plot
        self.__sectInitTraces()
         self.__sectInitPlot()
....
```

where:

```python
def __sectInitPlot(self):
        """
        Function initialises plot all the illustration is done by
        self.plotSection()
        """
        ax = self.fig.gca()
        # Calculate normalizing factor
        self.__sectNormalizeTraces()
        # Calculate scaling factor
        self.__sectScaleTraces()
        # ax.plot() prefered over containers
        for _tr in range(self._tr_num):
            # Scale, normalize and shift traces by offset
            # for plotting
            ax.plot(self._tr_data[_tr] / self._tr_normfac[_tr]
[...]

PR #933, opened 2015-01-21
Section plot updates

QuLogic:
This provides a couple updates to the section plot, namely:
 * Enable the `timeshift` option unconditionally. I don't know why anyone would want their traces to not be aligned to the same time. (See #913 and #926.)
 * Add a `reftime` option like in the relative plot, which allows for shifting the 0 point.


mbyt:
looks good to me

PR #934, opened 2015-01-22
Fix CatchOutput context manager, take 2

QuLogic:
I simplified `CatchOutput` by using `tempfile.TemporaryFile` instead of manual file management based on http://bugs.python.org/msg184312. Also should have sort-of fixed the bug with `def` disappearing on Python 3. The real fix is to use `PYTHONUNBUFFERED`, but that's external and cannot be tweaked in a script.

This works on Linux, obviously. But I guess we should see what's up with OSX and Windows, this time.


QuLogic:
No leaks, btw, using the same test as the last time.

mbyt:
Let's give it a try.

PR #935, opened 2015-01-22
Fix dtypes to SciPy's FFT when resampling.

### LINKED ISSUE #923: Endianness Issue with SciPy's FFT
SciPy's FFT cannot deal with big endian floats. We'll just have to enforce certain flags when passing to it.

Code snippet to reproduce:

```python
import obspy
import numpy as np

tr = obspy.read()[0]
tr.data = np.require(tr.data, dtype=">f4")
tr.resample(10)
```

Came up on the mailing list: http://lists.swapbytes.de/archives/obspy-users/2014-December/001602.html

QuLogic:
SciPy only likes native byte orders. ~~There are some other uses of `fftpack` though; I haven't had a chance to look into those.~~

Fixes #923, at least.

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1500,6 +1500,14 @@ class TraceTestCase(unittest.TestCase):
         self.assertEqual(tr_3.stats.sampling_rate, 10.0)
         self.assertEqual(tr_3.stats.starttime, tr.stats.starttime)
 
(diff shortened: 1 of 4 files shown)


QuLogic:
The other `fftpack` calls should also be okay now. I also slipped in an optimization of the `sonogram` function.

mbyt:
Looks good to me! And again thanks @QuLogic .

Issue #918, opened 2014-11-21
Possible memory leak in SeedLinkConnection.collect()


mbankdmt:
I'm seeing memory usage growth when using SeedLinkConnection.collect() in a loop.

Linux or Windows
obspy 0.9.2
numpy 1.8.2
python 2.7.6

Can anyone confirm?

QuLogic:
There was a [recent mailing list post](http://lists.swapbytes.de/archives/obspy-users/2015-January/001619.html) that claims this leak is caused by `range`. That being said, the development version will use the generator version of `range`, and this will likely fix the issue (assuming that is the cause).

mbyt:
@mbankdmt could you comment on the status of this ticket?

Based on the above comment, the memory could be due to range (why ever???). However, this is replaced in current master --- therefore, does this problem still persist in master?

mbankdmt:
I just cloned the latest master and did a run of the above code snippet against it. On my system the memory leak is still showing up.

mbyt:
Just tried your script, and I think I can reproduce the issue. I got the following output:
```
...
gc: collectable <list 0x7f1ce7ebb9e0>
gc: collectable <list 0x7f1ce7ebb878>
gc: collectable <list 0x7f1ce7eb7b00>
Unreachable objects: 25
Additional memory usage after 10 packets: 0 bytes, 0 bytes/packet
Additional memory usage after 20 packets: 0 bytes, 0 bytes/packet
...
Additional memory usage after 150 packets: 4096 bytes, 27 bytes/packet
Additional memory usage after 160 packets: 24576 bytes, 153 bytes/packet
Additional memory usage after 170 packets: 49152 bytes, 289 bytes/packet
Additional memory usage after 180 packets: 65536 bytes, 364 bytes/packet
Additional memory usage after 190 packets: 90112 bytes, 474 bytes/packet
...
```

mbyt:
With #940, the output is now:
```
...
gc: collectable <list 0x7f9752d4ebd8>
gc: collectable <list 0x7f9752aff1b8>
gc: collectable <list 0x7f9752affb90>
Unreachable objects: 33
Additional memory usage after 10 packets: 12288 bytes, 1228 bytes/packet
Additional memory usage after 20 packets: 12288 bytes, 614 bytes/packet
...
Additional memory usage after 190 packets: 16384 bytes, 86 bytes/packet
Additional memory usage after 200 packets: 16384 bytes, 81 bytes/packet
...
```

[...]

PR #940, opened 2015-01-25
Fix seedlink leak

QuLogic:
I think this fixes the leak in #918. At least, the example given no longer shows additional memory usage after ~29000 packets.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -66,6 +66,7 @@
      correctly (see #807)
    * New submodule `easyseedlink` providing an easier way to create
      SeedLink clients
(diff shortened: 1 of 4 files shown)


mbyt:
Great work @QuLogic. How did you manage to locate it? Valgrind?

QuLogic:
No, I originally looked for buffer-ish things like strings, but they didn't seem to have a problem. As I was looking, I noticed the allocation in C land, and looking up `msr_init` pointed in the direction of additional un-freed blockette data.

megies:
Awesome work. That one's been lurking around for ages. Big thanks! :)

Issue #824, opened 2014-06-18
ObsPy Packages


krischer:
There are a couple of efforts going on to package ObsPy for various operating systems. This issue serves to purpose of getting overview of and maybe coordinating the efforts a bit more. We should also put some work into highlighting the packages on the wiki page to make it easier for users.

---

https://linuxlifecycle.com/

---
#### Anaconda

Anaconda packages are built and deployed by pull requests to `conda-forge/obspy-feedstock` (make a new branch for a PR in `obspy/obspy-feedstock` for a new release). Some missing packages (Linux 32bit, ..) are built in docker images by @krischer.

#### Ubuntu/Debian/Raspbian Packages

@megies builds packages for those: https://github.com/obspy/obspy/wiki/Installation-on-Linux-via-Apt-Repository
 - [Release life cycle / EOL schedule](https://wiki.debian.org/DebianReleases#Production_Releases)
 - [Package version lookup](https://www.debian.org/distrib/packages) (for dependency versions etc.)

#### Raspbian

@megies offers packages for the hard-float Raspbian wheezy and jessie (see above)

#### Fedora/RHEL/CentOS

@qulogic has some ongoing efforts: https://github.com/QuLogic/rpmbuild-specs

Fedora:
 - [Current releases](https://fedoraproject.org/wiki/Releases#Current_Supported_Releases)
 - [EOL releases](https://fedoraproject.org/wiki/End_of_life)
 - [package search](https://apps.fedoraproject.org/packages) (for dependency versions etc.)

#### ArchLinux

~~@gauteh builds arch linux packages: https://aur.archlinux.org/packages/python-obspy-git/~~
https://aur.archlinux.org/packages/python-obspy/

#### FreeBSD

Wen Heping maintains FreeBSD packages: https://github.com/obspy/obspy/wiki/Installation-on-FreeBSD
I am not sure if he has a github account but I'll email him with a link to this issue.

#### Windows

@barsch Builds packages for windows: https://pypi.python.org/pypi/obspy/

#### MacPorts (OSX)

@petrrr maintains macports packages: https://github.com/obspy/obspy/wiki/Installation-on-Mac-OS-X-using-Macports

#### NetBSD:

 Kamel Derouiche builds NetBSD packages: http://pkgsrc.se/wip/py-obspy
These seem to be outdated. I'll query him about the current status and send him a link to this issue.

#### OpenSUSE:

```

---
```

---

---

Did I forget anything?
### Potential future plans
[...]

PR #827, opened 2015-01-25
Licensing clean up

megies:
This is to keep track of all licensing details discussed in #824. The following list keeps track of anything that is currently not LGPLv3.

 - [x] replacing libtau with obspy/taupy ~~**libtau** (no license at all? consent to relicense by B. Kennett, see https://github.com/obspy/obspy/blob/master/obspy/taup/LICENSE.txt)~~
    - [ ] ~~license for `obspy.taup` needs to be clarified in https://github.com/obspy/obspy/blob/master/obspy/taup/README.txt~~
    - [ ] ~~include as is in `obspy.taup` under LGPLv3?~~
 - [x] **evalresp** (no license information at all?)
    - [x] contacted Ilya Dricker and Eric Thomas (both ISTI) - tjm@iris ... does not exists anymore
       - Ilya gave his Ok
    - [x] contacted Robert Casey (IRIS)
    - [x] got feedback from Robert Casey and Tim Ahern (IRIS)
    - [x] evalresp contains evr_spline.c which comes from GNU plotutils and that is GPL. evr_spline is based on #892 on a LGPL SPLINE library --- needs merge.
    - [x] include as is in `obspy.signal` under LGPLv3?
 - [x] **obspy.imaging** (reason was GMT beachball script)
    - [x] [GMT now LGPLv3](https://github.com/obspy/obspy/issues/824#issuecomment-46743006)
    - [x] [all authors of matlab script consented to LGPLv3 for derived python script](https://github.com/obspy/obspy/issues/824#issuecomment-46743006)
    - [x] make everything in `obspy.imaging` LGPLv3
 - [x] **obspy.earthworm**
    - [x] [contacted V. Kress](https://github.com/obspy/obspy/issues/824#issuecomment-46743006)
    - [x] wait for reply
    - [x] relicense under LGPLv3?
 - [x] **GSE_UTI** (`obspy.gse2`) (no license information at all?!)
    - [x] contacted S. Stange about licensing issues
    - [x] got OK for LGPL

Anything I missed?

Finally, if everything is clear..
 - [x] move LGPLv3 license information to `obspy` top level package folder?


megies:
Can be merged after removal of `obspy.taup` and replacement with `obspy/taupy`.

QuLogic:
BTW, I heard this on the iris page, but you need both GPL and LGPL license texts since LGPLv3 is an extension of GPLv3.

krischer:
Also written down here: https://www.gnu.org/licenses/gpl-howto.html

So we should probably do that.

krischer:
Done in 66985a33bf2f35188a721e7deb8f9d673c6191db. Does this also mean we can remove the license files in all submodules? I think so.

megies:
> Done in 66985a3

This should have been done in and now is conflicting with this PR.. :P Anyway, not a big thing to fix.

> Does this also mean we can remove the license files in all submodules? I think so.

Already done in this PR.

PR #944, opened 2015-01-25
Replace obspy.taup Fortran code with TauPy

mbyt:
This PR replaces the use of the iaspei-tau library with a Python port of the Java TauP (http://www.seis.sc.edu/taup/) library. This means that ObsPy is finally free of Fortran code easing installation across all platforms tremendously (e.g. we are finally able to deliver conda packages). Additionally we gain quite a bit functionality like raypaths and pierce points.

~~Still to resolve: Licensing issues~~


megies:
Urrrggghs. The new `obspy.taup` code base is a mess.. but! It will make us `Fortran`-free:copyright: which cost us uncounted hours (and then some) for Win/Mac during packaging sprints in the past.

I've started some refactoring (for module naming and some methods..) but far from done. Anybody has a good tool to refactor everything according to our coding style.. go go go!

@QuLogic, please have mercy on our souls.. please apply any coding style changes you want to add directly in here.. working on comments on the code base won't be an option for this naming hell..

We're gonna try to get this ready for releasing until end of week.. to have at least the exposed routines clean.

krischer:
So Travis passes and the old obspy.taup interface correctly works. This is still really really slow compared to the old approach and we really have to optimize this!

Otherwise it also still lacks documentation of the new interface.

crotwell:
Just for official record keeping, I hereby grant my permission for your python port of my TauP Toolkit to be released with ObsPy under the LGPL license.

Philip Crotwell

mbyt:
@crotwell thanks so much!

QuLogic:
Alright, so I've done a bunch of optimizations, but I don't see much that can be done for the evaluation side of things (aside from maybe changing large parts of the structure/algorithm.) We might have to cut that full test suite into some smaller more focused tests.

krischer:
@qulogic Thanks for all this work on converting it to numpy!

Right not its still much too slow unfortunately. On my system its a factor of 500 slower than the old Fortran version. This is too much and will actually break existing applications. It can definitely be slower than the old version but not by this much. 

I guess there is no way around writing this function and everything that happens underneath it in C:

https://github.com/obspy/obspy/blob/3a4f604a109edba3b48dc295f65c67d852bf74f9/obspy/taup/tau_branch.py#L81-108

Just converting some of the underlying function to C will not cut it I think as its not only the calculations but also the amount of times all these functions are called. It think it should be possible but not really pretty. I already tried this a bit last night but it turned out harder than expected. I hope to find some time this week to work on this.

krischer:
I converted some functions to C and thus grabbed some low hanging fruits. There is still quite a lot of optimization potential. Nonetheless this now is "only" 70 times slower than the old Fortran version which just always calculates all phases. But if one chooses the new interface and selects the desired phases, it is only around 7 times slower which is totally ok for now.

Right now the models **cannot** be built during the installation anymore...The models now require the shared tau library which naturally only exists after the setup procedure. I chose not to workaround this issue as we should just figure out a way to properly (de)serialize the models to some defined data structure.

To still run this right now, comment out the section where the models are built at the very end of the `setup.py` file, install (with `-e`) once, uncomment the line, and install once more.

To do before we can merge this:

- [ ] Proper (de)serialization of models.
- [ ] Add a the `getTravelTimes()` function.
- [ ] Document the `obspy.taup.tau.TauPyModel` class, the new public interface.
- [ ] Make the aforementioned interface a bit nicer.
- [ ] Maybe add some plotting routines?

krischer:
A snippet that is worthwhile to optimize as it is still a fairly likely use case:

```python
from obspy.taup import getTravelTimes

for _i in range(180):
    getTravelTimes(float(_i), 25)
```

output4

QuLogic:
I made some patches using `mock` so that CI will finish, but we'll have to remove that eventually.

krischer:
I just noticed something else is wrong:

```python
In [1]: from obspy.taup import TauPyModel
In [2]: m = TauPyModel()
In [3]: print m.get_travel_times(source_depth_in_km=15, distance_in_degree=6,
   ...:                          phase_list=["P", "Pn", "Sn"])
7 arrivals
	Pn phase arrival at 14.118 seconds
	Sn phase arrival at 25.979 seconds
	P phase arrival at 88.209 seconds
	P phase arrival at 104.238 seconds
	P phase arrival at 104.267 seconds
	P phase arrival at 114.865 seconds
	P phase arrival at 114.874 seconds
```

The `Pn` and `Sn` phases arrive **much much** too fast. Might be due to the one check I removed in the C version. I'll look into it later. Also we need to add some tests for these phases probably by comparing to the output of the Java TauP program.

crotwell:
Just a guess, but I suspect what you are seeing is the downward and upward legs of Pn and Sn, but not the refracted part below the moho. The beginning of the Pn phase at about .6 deg gives 14 seconds travel time, which matches your values above. One way to check would be to see if the time stays at about 14 seconds regardless of distance.

megies:
> The only issue is that with this commit ObsPy depends on the Python and NumPy headers and thus -dev packages

@krischer, I think we do depend on Python headers for building ObsPy.. at least I put it in `Build-Depends` for the Debian packaging at some point.

mbyt:
> @krischer, I think we do depend on Python headers for building ObsPy.. at least I put it in
> Build-Depends for the Debian packaging at some point.
We are using ctypes everywhere to access shared C libraries directly. We do not use the Python API anywhere. Therefore, Python.h shouldn't be required (and I didn't find it when grepping for it).

mbyt:
On my machine the obspy.taup.tests ran on 2015-02-02 on 175min. Today it's only 5min. Factor 35 speed up --- I am amazed guys!

QuLogic:
> I just noticed something else is wrong:

Could they be related to the comments I made on 4345c21ba? The original code looked wrong to me, but maybe I did not fix them correctly. I also recall some strangeness in handling head waves somewhere, but don't recall at the moment.

QuLogic:
@crotwell could be right, though it does change a little bit:
```
In [11]: for dist in range(1, 22): print(dist, m.get_travel_times(source_depth_in_km=15, distance_in_degree=dist, phase_list=['Pn']))
1 1 arrivals
	Pn phase arrival at 14.031 seconds
2 1 arrivals
	Pn phase arrival at 14.048 seconds
3 1 arrivals
	Pn phase arrival at 14.066 seconds
4 1 arrivals
	Pn phase arrival at 14.083 seconds
5 1 arrivals
	Pn phase arrival at 14.101 seconds
6 1 arrivals
	Pn phase arrival at 14.118 seconds
7 1 arrivals
	Pn phase arrival at 14.135 seconds
8 1 arrivals
	Pn phase arrival at 14.153 seconds
9 1 arrivals
	Pn phase arrival at 14.170 seconds
10 1 arrivals
	Pn phase arrival at 14.188 seconds
11 1 arrivals
	Pn phase arrival at 14.205 seconds
12 1 arrivals
	Pn phase arrival at 14.223 seconds
13 1 arrivals
	Pn phase arrival at 14.240 seconds
14 1 arrivals
	Pn phase arrival at 14.258 seconds
15 1 arrivals
	Pn phase arrival at 14.275 seconds
16 1 arrivals
	Pn phase arrival at 14.293 seconds
17 1 arrivals
	Pn phase arrival at 14.310 seconds
18 1 arrivals
	Pn phase arrival at 14.327 seconds
19 1 arrivals
	Pn phase arrival at 14.345 seconds
20 1 arrivals
	Pn phase arrival at 14.362 seconds
21 0 arrivals
```

mbyt:
Went back to the old taupy days 746073e17d9d108c30e61d5b1711e7cd1988a035. Even there the problem persisist:
```
python -c "from taupy.tau import TauPyModel; m=TauPyModel(); print(m.get_travel_times(source_depth_in_km=15, distance_in_degree=6,phase_list=['P', 'Pn', 'Sn']))"
```

QuLogic:
All the funny phases that @krischer mentioned on gitter:
```
In [3]: print(m.get_travel_times(source_depth_in_km=15, distance_in_degree=10,
   ...:                          phase_list="P,S,PP,PPP,PPPP,S,SS,SS,SSSS,Sn,Pn".split(',')))
29 arrivals
	P phase arrival at 143.085 seconds
	Pn phase arrival at 143.262 seconds
	PP phase arrival at 150.746 seconds
	PPP phase arrival at 158.287 seconds
	PPPP phase arrival at 165.811 seconds
	PP phase arrival at 175.579 seconds
	PP phase arrival at 175.716 seconds
	PPP phase arrival at 178.739 seconds
	PPP phase arrival at 179.285 seconds
	PPPP phase arrival at 181.879 seconds
	PPPP phase arrival at 183.130 seconds
	PP phase arrival at 191.605 seconds
	PP phase arrival at 191.703 seconds
	PPP phase arrival at 192.251 seconds
	PPPP phase arrival at 193.069 seconds
	S phase arrival at 256.113 seconds
	Sn phase arrival at 256.688 seconds
	SS phase arrival at 268.828 seconds
	SSSS phase arrival at 293.443 seconds
	SS phase arrival at 304.161 seconds
	SS phase arrival at 304.400 seconds
	SSSS phase arrival at 314.860 seconds
	SSSS phase arrival at 316.941 seconds
	SS phase arrival at 330.914 seconds
	SSSS phase arrival at 333.267 seconds
	PPPP phase arrival at 3075.691 seconds
	PPPP phase arrival at 3169.104 seconds
	SSSS phase arrival at 5645.054 seconds
	SSSS phase arrival at 5828.966 seconds
```

PR #949, opened 2015-02-03
Taupy np

mbyt:
Try to unify the taupy work, try merging taupy-np into taupy.


mbyt:
@QuLogic sorry for jumping ahead... Just to many branches for my brain :\.

PR #950, opened 2015-02-04
Streaming to/from file-like objects for obspy.sac

krischer:
This PR adds support for reading/writing from/to file-like objects and open files to `obspy.sac`. This enables things like

```python
In [1]: import io

In [2]: import obspy

In [3]: with io.BytesIO() as fh:
   ...:     obspy.read()[:1].write(fh, format="sac")
```

Also works with sacxy.

```python
In [4]: with io.BytesIO() as fh:
   ...:     obspy.read()[:1].write(fh, format="sacxy")
```

It will fail if the stream object to be written has more than one Trace as the SAC format does not support that use case.

```python
In [5]: with io.BytesIO() as fh:
   ...:     obspy.read().write(fh, format="sac")
...
... in writeSAC(stream, filename, byteorder, **kwargs)
    422     if is_bytes_buffer(filename):
    423         if len(stream) > 1:
--> 424             raise ValueError("If writing to a file-like object in the SAC "
    425                              "format, the Stream object must only contain "
    426                              "one Trace")

ValueError: If writing to a file-like object in the SAC format, the Stream object can only contain one Trace.
```

It will only work with file-like objects in bytes mode. On Python 2 it additionally works with `StringIO.StringIO` and files opened in text mode with the built-in `open()` method but I'd consider this bad style.

To this end I added two new functions to `obspy.core.compatibility`: `is_text_buffer()` and `is_bytes_buffer()`. These check if the passed object can read/write text or bytes respectively. I think this is a fairly solid and stable way of doing these things across Python versions and we should add that to other formats as well.

The scheme to read from filenames or file-like object thus looks like this:

```python
    if is_bytes_buffer(filename):
        return _readSAC(buf=filename, headonly=headonly,
                        debug_headers=debug_headers, fsize=fsize, **kwargs)
    elif isinstance(filename, (str, bytes)):
        with open(filename, "rb") as fh:
            return _readSAC(buf=fh, headonly=headonly,
                            debug_headers=debug_headers, fsize=fsize, **kwargs)
    else:
        raise ValueError("Cannot open '%s'." % filename)
```

This should be able to handle everything thrown at it in the correct fashion, the underlying `_readSAC()` function only works with file-like objects. Once again, we should really add something like this to the other formats as well.

A side effect of this PR is that sacio now completely works with file-like objects but that should not affect too many people. I did not put any effort into reorganizing `obspy.sac` as one of the next things on our list is apparently to rewrite `obspy.sac`.

I wrote like 20 new tests so I am fairly sure it works as intended :-)

### DIFF
--- a/obspy/sac/core.py
+++ b/obspy/sac/core.py
@@ -24,8 +24,8 @@ def isSAC(filename):
     """
     Checks whether a file is a SAC file or not.
 
-    :type filename: str or file-like object
     :param filename: SAC file to be checked.
+    :type filename: str, open file, or file-like object
     :rtype: bool
     :return: ``True`` if a SAC file.
 
@@ -46,8 +46,8 @@ def _isSAC(buf):
     """
     Checks whether a file-like object contains a SAC file or not.
 
-    :type buf: file-like object or open file.
     :param buf: SAC file to be checked.
+    :type buf: file-like object or open file.
     :rtype: bool
     :return: ``True`` if a SAC file.
     """
@@ -113,8 +113,8 @@ def isSACXY(filename):
     """
     Checks whether a file is alphanumeric SAC file or not.
 
-    :type filename: str or file-like object.
     :param filename: Alphanumeric SAC file to be checked.
+    :type filename: str, open file, or file-like object
     :rtype: bool
     :return: ``True`` if a alphanumeric SAC file.
 
@@ -135,15 +135,11 @@ def _isSACXY(buf):
     """
     Checks whether a file is alphanumeric SAC file or not.
 
-    :type buf: file-like object oropen file
     :param buf: Alphanumeric SAC file to be checked.
+    :type buf: file-like object or open file
     :rtype: bool
     :return: ``True`` if a alphanumeric SAC file.
     """
-    # First find out if it is a text or a binary file. This should
-    # always be true if a file is a text-file and only true for a
-    # binary file in rare occasions (Recipe 173220 found on
-    # http://code.activestate.com/
     cur_pos = buf.tell()
     try:
         try:
@@ -173,18 +169,18 @@ def readSACXY(filename, headonly=False, debug_headers=False,
         This function should NOT be called directly, it registers via the
         ObsPy :func:`~obspy.core.stream.read` function, call this instead.
 
-    :type filename: filename of file-like object.
     :param filename: Alphanumeric SAC file to be read.
-    :type headonly: bool, optional
+    :type filename: str, open file, or file-like object
     :param headonly: If set to True, read only the head. This is most useful
         for scanning available data in huge (temporary) data sets.
-    :type debug_headers: bool, optional
+    :type headonly: bool
     :param debug_headers: Extracts also the SAC headers ``'nzyear', 'nzjday',
         'nzhour', 'nzmin', 'nzsec', 'nzmsec', 'delta', 'scale', 'npts',
         'knetwk', 'kstnm', 'kcmpnm'`` which are usually directly mapped to the
         :class:`~obspy.core.stream.Stream` object if set to ``True``. Those
         values are not synchronized with the Stream object itself and won't
         be used during writing of a SAC file! Defaults to ``False``.
+    :type debug_headers: bool
     :rtype: :class:`~obspy.core.stream.Stream`
     :return: A ObsPy Stream object.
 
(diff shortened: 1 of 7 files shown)


obspy/sac/core.py
```diff
@@ -100,27 +122,46 @@ def isSACXY(filename):
...
+def _isSACXY(buf):
+    """
+    Checks whether a file is alphanumeric SAC file or not.
+
+    :type buf: file-like object oropen file
+    :param buf: Alphanumeric SAC file to be checked.
+    :rtype: bool
+    :return: ``True`` if a alphanumeric SAC file.
+    """
     # First find out if it is a text or a binary file. This should
```
QuLogic: Remove comment?

obspy/sac/core.py
```diff
@@ -132,7 +173,7 @@ def readSACXY(filename, headonly=False, debug_headers=False,
         This function should NOT be called directly, it registers via the
         ObsPy :func:`~obspy.core.stream.read` function, call this instead.
 
-    :type filename: str
+    :type filename: filename of file-like object.
```
QuLogic: This is not the `type`.

obspy/sac/core.py
```diff
@@ -206,7 +314,7 @@ def readSAC(filename, headonly=False, debug_headers=False, fsize=True,
         This function should NOT be called directly, it registers via the
         ObsPy :func:`~obspy.core.stream.read` function, call this instead.
 
-    :type filename: str
+    :type filename: filename, open file, or file-like object.
```
QuLogic: Again, not the `type`.

obspy/sac/core.py
```diff
@@ -256,8 +403,9 @@ def writeSAC(stream, filename, byteorder="<", **kwargs):  # @UnusedVariable
 
     :type stream: :class:`~obspy.core.stream.Stream`
     :param stream: The ObsPy Stream object to write.
-    :type filename: str
-    :param filename: Name of file to write.
+    :type filename: str or file-like object
+    :param filename: Name of file to write. If it is not a filename, it only
```
QuLogic: There are two 'it's in this sentence, but they do not refer to the same thing. Should try and avoid that.

obspy/sac/sacio.py
```diff
@@ -578,66 +542,55 @@ def ReadSacHeader(self, fname):
...
 
-        :param f: filename (SAC binary).
+        :param fh: open file or file-like buffer
 
         >>> from obspy.sac import SacIO # doctest: +SKIP
-        >>> tr = SacIO('test.sac') # doctest: +SKIP
-        >>> tr.WriteSacBinary('test2.sac') # doctest: +SKIP
-        >>> u = SacIO('test2.sac') # doctest: +SKIP
+        >>> with open('test.sac', 'rb') as fh:
+        ...     tr = SacIO('test.sac') # doctest: +SKIP
```
QuLogic: `fh` is not used in this context.

obspy/sac/sacio.py
```diff
@@ -578,66 +542,55 @@ def ReadSacHeader(self, fname):
...
+        >>> with open('test2.sac', 'wb') as fh:
+        ...     tr.WriteSacBinary('test2.sac') # doctest: +SKIP
```
QuLogic: `fh` is not used in this context either.

obspy/sac/sacio.py
```diff
@@ -852,77 +794,67 @@ def readTrace(self, trace):
...
         >>> from obspy.sac import SacIO # doctest: +SKIP
-        >>> tr = SacIO('test.sac') # doctest: +SKIP
-        >>> tr.WriteSacXY('test2.sac') # doctest: +SKIP
-        >>> tr.IsValidXYSacFile('test2.sac') # doctest: +SKIP
+        >>> with open('test.sac', 'rb') as fh:
+        ...     tr = SacIO(fh) # doctest: +SKIP
+        >>> with open('test2.sac', 'wb') as fh:
+        ...     tr.WriteSacXY(fh) # doctest: +SKIP
+        >>> with open('test2.sac', 'wb') as fh:
+        ...     tr.IsValidXYSacFile('test2.sac') # doctest: +SKIP
```
QuLogic: `fh` is not used in this context.

krischer:
@QuLogic thanks a lot for the review!

Travis error on 2.6 is unrelated. Can I merge this?

QuLogic:
I haven't had a chance to check all the tests, but otherwise, it looks good.

obspy/sac/tests/test_core.py
```diff
@@ -478,6 +485,262 @@ def test_read_with_fsize(self):
...
+        del tr.stats.sac.depmen
+        del tr2.stats.sac.depmen
+        np.testing.assert_array_equal(tr.data, tr2.data)
+        self.assertEqual(tr, tr2)
+
+    def test_read_write_open_file(self):
+        st = readSAC(self.file)
+
+        with NamedTemporaryFile() as tf_out:
+            with open(tf_out.name, "wb") as fh:
```
QuLogic: Is it really necessary to re-open the file? `tf_out` should already have the file opened read/write.

krischer: Good point. Didn't work so far but it does now.

krischer:
`NamedTemporaryFile` now inherits from `io.BufferedIoBase` and behaves like a real binary file for all practical purposes.

PR #951, opened 2015-02-04
Pass on control over arclink request forwarding...

yannikbehr:
...to getStations and getNetworks

### DIFF
--- a/obspy/arclink/client.py
+++ b/obspy/arclink/client.py
@@ -97,7 +97,7 @@ class Client(object):
     * IPGP:  eida.ipgp.fr:18001
     * USP:   seisrequest.iag.usp.br:18001
     """
-    #: Delay in seconds between each status request
+    # : Delay in seconds between each status request
     status_delay = 0.5
 
     def __init__(self, host="webdc.eu", port=18002, user=None,
@@ -1307,7 +1307,7 @@ class Client(object):
 
         return data
 
-    def getNetworks(self, starttime, endtime):
+    def getNetworks(self, starttime, endtime, route=False):
         """
         Returns a dictionary of available networks within the given time span.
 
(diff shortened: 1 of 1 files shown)


megies:
Looks good to merge.. just, here's one more: https://github.com/obspy/obspy/blob/arclink_request_forwarding/obspy/arclink/client.py#L438

Shouldn't that one be passed on from parent as well?

yannikbehr:
Looks like it but the default for  'self.saveWaveform' differs from that for 'self.getInventory'. Not sure if there's a reason for it?

megies:
Not sure.. could be intentional for Inventories.. but then again I don't see a real reason to have a discrepancy there. Maybe somebody else can shed light upon this?

megies:
> 1095 problems were fixed (including 343 errors and 155 code smells).

Aha. Interesting. Nice work.. ;)

yannikbehr:
@krischer Lion, do you know more about this?

krischer:
@yannikbehr about the landscape stuff or arclink?

No clue who did the landscape stuff but we should maybe tell it to not parse the autogenerated code from our docs where most of the problems come from ;-)

I'm also not sure on the Arclink issue. Does not every arclink node have a copy of the master table and could thus provide the inventory information? Therefore there might be no need to actually route? Nonetheless I would also pass the argument to the `getInventory()` method - I don't think it can do any harm.

@barsch You are probably the ObsPy Arclink expert ;-) Any opinion?

barsch:
I added landscape months ago for the "health" badge on https://github.com/obspy/obspy/wiki - however until now it never messed with the commit messages - seems like they changed that yesterday - hell the even spam my mailbox now with health messages :(

concerning arclink: seems like it should be ```route=route``` there too

yannikbehr:
I would've merged this but I have no idea what the failed Travis CI build means. Is it save to merge?

barsch:
IMHO it is fine to merge as the failed test is known to fail on some machines due to timeout and not related to the fix above.

megies:
I'm gonna merge `releases` right away into `master` so that this doesn't get left out of 0.10..

QuLogic:
That failing test is corrected on `master`.

PR #954, opened 2015-02-09
Sort imports

QuLogic:
NOTE: We should probably merge this quickly or there will be many conflicts.

This is based on running the [`isort` tool](https://github.com/timothycrosley/isort) over the source (`isort -f future -p obspy -ot -ac -rc .`), but I had to make a few manual changes to comply with PEP8 and the SciPy conventions. You can compare [the original result with the manual tweaks here](https://github.com/QuLogic/obspy/compare/QuLogic:isort-orig...QuLogic:isort-orig2). It's mostly:
 * Correcting two blank lines after imports. Sometimes `isort` only adds 1 and sometimes adds 4!
 * Move standard library imports from `future` to be between standard library imports and third-party imports. It would normally put these at the end, and this does trigger E402 in the new pep8, so we may want to undo this one.
 * Make `future.builtins` the absolute first import. Normally `from future import standard_library` would sort first using standard `isort`. I just did this one because I think `future.builtins` is more important.
 * Sort NumPy/matplotlib according to SciPy conventions instead of alphabetical.

So, the layout is something like this (with unnecessary bits left out when possible) now:
```
from __future__ import ...
from future.builtins import *
from future import standard_library  # If necessary
from future... # Any others

import stdlib  # io, sys, etc.
from stdlib import stuff  # also io, sys, etc., but specific elements from them

    import py2_py3_compat_standard_library  # Pulled out of future

import third_party
from third_party import ...
import numpy as np
import matplotlib.something...
from matplotlib... import something
import scipy.integrate  # Or others...

import obspy
import obspy.core
from obspy.imaging import ...
```

PS, I did not bother to fix taup; I'll just make the changes on the PR.


QuLogic:
Ooops, just realized `isort`skips the `__init__.py` by default. This is good in one or two cases, but I'd like to apply it there as well.

krischer:
Looks good. I just checked a couple of files but I like the result. I'll wait until you ran the script also over the `__init__.py` files and then I'll merge.

QuLogic:
OK, I think that should do it. BTW, what do you think about:
```
from something_long import (something, other_thing,
                            wrapped_and_aligned_thing)
```
vs.
```
from something_long import (something, other_thing,
    indented_thing)
```

The first is the default out of `isort`, but the latter is more in line with what we've been doing. It's just an option away if we like that better.

krischer:
I would take the first one. The second one is invalid according to PEP8 I believe.

megies:
:+1: for sorting the imports :)

> The second one is invalid according to PEP8 I believe.

Think so too.. I guess it's the same as with function calls, where the first line after the bracket has to be empty if you want to go with a single indent on follow up lines.

QuLogic:
> The second one is invalid according to PEP8 I believe.

Oh, possibly. I never really checked what the real output was, or whether it passed PEP8. I just know there are some options.

megies:
Btw.. I guess checking the import sort order in our test suite is too much work? Or maybe it's a too high demand on contributions? Otherwise over time imports will get jumbled again..

QuLogic:
Well, `isort` does have a check mode, but it depends on `pies` for 2/3 support and that currently conflicts with `future` due to [configparser stuff](PythonCharmers/python-future#118). Plus, it places `with standard_library` at the end and I'd prefer to place that with the stdlib (even if it triggers PEP8's new E402).

PR #955, opened 2015-02-09
Modify git version to match PEP440.

QuLogic:
The revision number (after last tag) is now the "dev" version and the commit hash is the "local version" as defined in PEP440.

This fixes a warning on Travis, though I don't know if many people install development versions with `pip` instead of `setup.py.

### DIFF
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -65,14 +65,22 @@ def call_git_describe(abbrev=4):
                   cwd=OBSPY_ROOT, stdout=PIPE, stderr=PIPE)
 
         p.stderr.close()
-        line = p.stdout.readline().decode()
+        line = p.stdout.readline().decode().strip()
         p.stdout.close()
 
         # (this line prevents official releases)
         # should work again now, see #482 and obspy/obspy@b437f31
         if "-" not in line and "." not in line:
-            line = "0.0.0-g%s" % line
-        return line.strip()
+            version = "0.0.0.dev0+g%s" % line
+        else:
+            parts = line.split('-')
+            version = parts[0]
+            try:
+                version += '.dev' + parts[1]
+                version += '+' + '-'.join(parts[2:])
+            except IndexError:
+                pass
+        return version
     except:
         return None
 
@@ -105,7 +113,7 @@ def get_git_version(abbrev=4):
 
     # If we still don't have anything, that's an error.
     if version is None:
-        return '0.0.0-tar/zipball'
+        return '0.0.0+tar/zipball'
 
     # If the current version is different from what's in the
     # RELEASE-VERSION file, update the file to be current.
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -117,7 +117,7 @@ def call_git_describe(abbrev=4):
         try:
             version += '.dev+' + parts[1]
             if remote_tracking_branch is not None:
-                version += '-' + remote_tracking_branch
+                version += '.' + remote_tracking_branch
         # IndexError means we are at a release version tag cleanly,
         # add nothing additional
         except IndexError:
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -109,13 +109,12 @@ def call_git_describe(abbrev=4):
     # (this line prevents official releases)
     # should work again now, see #482 and obspy/obspy@b437f31
     if "-" not in line and "." not in line:
-        version = "0.0.0.dev0+.g%s" % line
+        version = "0.0.0.dev+.g%s" % line
     else:
-        parts = line.split('-')
+        parts = line.split('-', 1)
         version = parts[0]
         try:
-            version += '.dev' + parts[1]
-            version += '+' + '-'.join(parts[2:])
+            version += '.dev+' + parts[1]
             if remote_tracking_branch is not None:
                 version += "." + remote_tracking_branch
         # IndexError means we are at a release version tag cleanly,
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -83,7 +83,7 @@ def call_git_describe(abbrev=4):
         p.stdout.close()
         remote_info = [line_ for line_ in remote_info
                        if line_.startswith('*')][0]
-        remote_info = re.sub(r".*? \[(.*?)\] .*", r"\1", remote_info)
+        remote_info = re.sub(r".*? \[([^ :]*).*?\] .*", r"\1", remote_info)
         remote, branch = remote_info.split("/")
         # find out real name of remote
         p = Popen(['git', 'remote', '-v'],
@@ -102,7 +102,8 @@ def call_git_describe(abbrev=4):
         else:
             remote = None
         if remote is not None:
-            remote_tracking_branch = "%s/%s" % (remote, branch)
+            remote_tracking_branch = re.sub(r'[^A-Za-z0-9._-]', r'_',
+                                            '%s-%s' % (remote, branch))
     except (OSError, ValueError):
         pass
 
@@ -116,7 +117,7 @@ def call_git_describe(abbrev=4):
         try:
             version += '.dev+' + parts[1]
             if remote_tracking_branch is not None:
-                version += "." + remote_tracking_branch
+                version += '-' + remote_tracking_branch
         # IndexError means we are at a release version tag cleanly,
         # add nothing additional
         except IndexError:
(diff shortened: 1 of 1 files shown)


megies:
According to this section: https://www.python.org/dev/peps/pep-0440/#dvcs-based-version-labels

`As hashes cannot be ordered reliably such versions are not permitted in the public version field. `

This would mean that including the "git commits since" count in the `.dev` public part is problematic as well, because the "commits since" count does not guarantee that you're comparing the count on the very same development line (e.g. if you work on a branch).

So maybe instead of `0.9.2.dev1615+g2e886b7400` it should be `0.9.2.dev+1615.g2e886b7400`?

QuLogic:
Yes, that's true. I assumed we wouldn't make a habit of released `dev` versions like this, so this is mostly to make Travis quieter.

megies:
Never noticed Travis complaining, I have to admit. I find the PEP compliant versioning a bit uglier, as it looks just a bit like a garbled up `git describe`. But I do like sticking to PEP so I'm fine with a change.. :)

QuLogic:
> Never noticed Travis complaining, I have to admit.

It's hidden in the `pip install .` step, but I figure if anyone uses that, it'd be nice to quiet it.

megies:
Hmm, I regularly use `pip install .` and `pip install -e .`.. never noticed a warning there.. anyway..

Anyway.. if we change the git based version numbering.. we could add information about the remote tracking branch in the version number. Something like the following could be instructive e.g. in test report version numbers  to quickly see what a test/version belongs to etc.

e.g.
```
0.9.2.dev.1390.ga92f69d55f-dirty.aringler-usgs/releases
0.9.2.dev.1615.g2e886b7400-dirty.QuLogic/pep440
```

Opinions?

QuLogic:
Makes sense. There's still a bug though; if you aren't in sync with your tracking branch, then `git branch -vv` outputs ` ahead/behind: #`, and I don't think the spaces would be allowed. In fact, I'm not sure the `/` is even allowed.

QuLogic:
Oops, misread the since tag for `check_call` as the tag for `check_output`. Guess I'll have to revert that one.

QuLogic:
I'd also rather use git plumbing instead of porcelain, but I can't seem to get `git for-each-ref --format='%(upstream)' commit` to actually print the upstream branch.

megies:
Actually it seems to be contradicting itself.. https://www.python.org/dev/peps/pep-0440/#local-version-segments allows `-` and `_` in local versioning (although only as separators). So we can keep those two I would say, as replacing it also kinda makes it look like a separator anyway.

Ok.. so we make it like this?
`0.9.2.dev.1390.ga92f69d55f-dirty.aringler-usgs.releases`

megies:
Btw.: You moved the "number of commits since" next to `.dev` again, was that on intention? Like I said I think this number is non-canonical and should not be in the local identifier according to that PEP440.

QuLogic:
pip was [complaining about `0.9.2.dev.1628.g7005c`](https://travis-ci.org/obspy/obspy/jobs/50703843#L449), though I'm not sure if it was fixed by moving the number after the `dev`, or by changing the `.` to a `+`.

QuLogic:
> allows - and _ in local versioning

Yes, you're right about that. pip does still [print a warning about it](https://travis-ci.org/obspy/obspy/jobs/51128585#L294), though:
```
/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/setuptools/dist.py:282:
 UserWarning: Normalizing '0.9.2.dev0+1947-g197a' to '0.9.2.dev0+1947.g197a'
```

QuLogic:
As we discussed, I dropped the `0` after `dev`, stuck with `-` between sections, and replaced invalid characters with `_` (originally I suggested `.`, but then thought that would conflict with our other usage of it.)

megies:
Ok so right now we're at
```
0.9.2.dev+1947-g62fe6e-QuLogic-pep440
```
And the following warning about normalizing the version number (which we can ignore, normalizing works just fine, and the local version part can't be compared anyway):
```
UserWarning: Normalizing '0.9.2.dev+1947-g62fe6e-QuLogic-pep440' to '0.9.2.dev0+1947.g62fe6e.qulogic.pep440'
```

Are you OK with using a `.` in between the git describe part, and the remote tracking branch info? I think it could serve as a bit more of a logic/visible separation there.

That would make it:
```
0.9.2.dev+1947-g62fe6ef78c-dirty.QuLogic-pep440  # remotely tracked branch, local changes
0.9.2.dev+1948-g849895d313                       # no remote, clean local state
0.9.2.dev+1948-g849895d313.megies-pep440b        # remotely tracked branch, clean local state
0.0.0+archive                                    # tar/zipball installation
```

QuLogic:
Yep, looks good.

megies:
@barsch, we need to adapt `tests.obspy.org`, the links to github are broken with the changed version numbers.. didn't find where to change it (its not in the `reporter`, right?)

screenshot from 2015-02-18 23 06 18

petrrr:
I'd have a related question here. 

I started providing a devel version of Obspy through Macports, for people needing the most recent features or fixes.

Because I did not what to end up with a 0.0.0 version, but cannot obtain the git describe string during build, I construct the string manually. This way for example the version reference to tests make more sense. I am not convinced this is the best way to deal with this, as it involves manually edit the change count.

Any thoughts of a better approach?

QuLogic:
> I started providing a devel version of Obspy through Macports, for people needing the most recent features or fixes.

All the more reason to make a release soon, eh, @krischer?
 
> cannot obtain the git describe string during build

Hmm, why not?

QuLogic:
@megies I think the test page linking is [done here](https://github.com/obspy/reporter/blob/master/src/reporter/core/models.py#L78)

megies:
I discussed with @QuLogic and since pip stores the version number in the *strict* normalized form internally, we decided that we bow to strict PEP440 in order to avoid a mismatch between *our* version number and the version number used by pip..

I'm rebasing to current master, so that appveyor has a chance to pass..

megies:
Merge this?

QuLogic:
I guess the server copy will need to be refreshed with your changes at some point still? But other than that, I think this can be merged.

mbyt:
> I guess the server copy will need to be refreshed

Alright, I think it is updated when I merged this, let's see if travis passes in master.

mbyt:
@megies did not work out :\ see https://travis-ci.org/obspy/obspy/jobs/52944728
```
Processing /home/travis/build/obspy/obspy
File "<string>", line 20, in <module>
File "/tmp/pip-PGlcTg-build/setup.py", line 770, in <module>
setupPackage()
File "/tmp/pip-PGlcTg-build/setup.py", line 696, in setupPackage
version=get_git_version(),
File "/tmp/pip-PGlcTg-build/obspy/core/util/version.py", line 147, in get_git_version
version = call_git_describe(abbrev)
File "/tmp/pip-PGlcTg-build/obspy/core/util/version.py", line 95, in call_git_describe
if line_.startswith(remote)][0].split()[1]
Complete output from command python setup.py egg_info:
File "<string>", line 20, in <module>
File "/tmp/pip-PGlcTg-build/setup.py", line 770, in <module>
setupPackage()
File "/tmp/pip-PGlcTg-build/setup.py", line 696, in setupPackage
version=get_git_version(),
File "/tmp/pip-PGlcTg-build/obspy/core/util/version.py", line 147, in get_git_version
version = call_git_describe(abbrev)
File "/tmp/pip-PGlcTg-build/obspy/core/util/version.py", line 95, in call_git_describe
if line_.startswith(remote)][0].split()[1]
```

megies:
huh? all the Travis builds passed earlier?!?

megies:
Maybe it's a Travis issue? It's been extremely laggy today, hardly being able to display build outputs for me..

mbyt:
Might be, so let's ping it tomorrow and see what happens.

megies:
Btw.. here's a successful build directly based on this, adding remote tracking info in Travis (to get branch info in tests.obspy.org).. https://travis-ci.org/megies/obspy/builds/52924310 and corresponding test reports, e.g. http://tests.obspy.org/21274/

megies:
Working again.. after just adding a commit that prints some git info in Travis. Guess it was just your sporadic Travis hick-up.

QuLogic:
We should probably make the check for remotes a bit more robust, though. I mean, if someone just creates a local branch, there should be no remote. I haven't tested it, but I guess that might fail, too?

megies:
Maybe you're right. We should probably just except everything here https://github.com/obspy/obspy/blob/master/obspy/core/util/version.py#L107.

QuLogic:
It's fixed now, though I just realized I got the commit message backwards...

PR #956, opened 2015-02-10
Fix mlabday epoch in array_processing.

### LINKED ISSUE #870: Array analysis matplotlib date conversion bug
From the mailinglist: http://lists.swapbytes.de/archives/obspy-users/2014-August/001478.html

> "Unix Time" to "Rata Die" conversion that takes place here 
> https://github.com/obspy/obspy/blob/master/obspy/signal/array_analysis.py#L1018 
> is wrong because "Rata Die" epoch starts at 1, not at zero. Add one and it is correct. The comment above that line is wrong. Should be days not hours. 

> Below is a python command line session that hopefully explains why this is wrong 

```python
>>> from time import time 
>>> from datetime import datetime 
>>> from matplotlib.dates import date2num 
>>> t = time() 
>>> t / (24. * 3600) +719162 - date2num(datetime.fromtimestamp(t)) 
-1.0 
```

QuLogic:
The numerical date used in matplotlib is days from 0001-01-01 + 1. Fixes #870.

Also, add date formatting on the beamforming example.


mbyt:
thx

PR #959, opened 2015-02-17
Plots for TauP raypaths

krischer:
This PR adds some basic plotting capabilities to `obspy.taup`. It offers two different types of plots: `spherical` and `cartesian` ones, it defaults to `spherical`.

Usage:

```python
>>> from obspy.taup import TauPyModel
>>> m = TauPyModel()
>>> m.get_ray_paths(500, 130).plot()
```

rays

```python
>>> m.get_ray_paths(500, 130).plot(plot_type="cartesian")
```

rays2

@crotwell: Sometimes the paths don't end at the rights points. Is this a consequence of the model not being sampled fine enough, or some other issue? Two examples:

```python
>>> m.get_ray_paths(10, 5, phase_list=["P"]).plot(plot_type="cartesian", plot_all=False)
```

problem

```python
>>> m.get_ray_paths(0, 10, phase_list=["PPPPPP"]).plot(plot_type="spherical", plot_all=True)
```
problem2


megies:
Looks great! Legend maybe needs extra room?

krischer:
Rewriting the `obspy.taup` docs is next on my list.

The problem with the legend more or less disappears if plotted interactively and one can just resize the window. I also guess that in most real use cases people will plot way less phases.

crotwell:
Regarding the end points not lining up, is this something that happens with my java TauP? If so then it is probably a bug and I should look into it. If not, then I guess it is a artifact of the translation. Checking that is probably worth doing in general.

The only other thought I have that might help is have you checked for paths that go the "wrong" way around the earth? For example of PPPPPP, there is a path that goes clockwise and another one that goes counterclockwise. If you don't "flip" the long way around path when plotting it will not end at the correct location. In your example one ray travels 350 degrees to end up 10 degrees away from the start, another travels 370 degrees, starting in the opposite direction, and also ends up at the same end point. THis is the purpose of the "Purist Distance" column in the output of taup_time.

krischer:
@crotwell: Thanks for your help with this. I think you are correct regarding the `PPPPPP` plot. I'll fix it. This one also works with your Java TauP.

The issue with the local `P` phases persists in your Java code though. Sorry..the plots are not as pretty. Code to produce them is at the end (your Java TauP version does not appear to be able to plot in a zoomed in fashion).

untitled

untitled

```python
import matplotlib.pylab as plt
import numpy as np
import subprocess

taup_path = "/Users/lion/Downloads/TauP-2.1.2/bin/taup_path"

# Local plot.
out = subprocess.Popen(
        [taup_path + " -mod iasp91 -pH 'P' -h 10 -deg 5 -o stdout"],
        shell=True,
        stdout=subprocess.PIPE).communicate()[0].decode()

plt.figure(1)

plt.gca().invert_yaxis()
phases = []
current_phase = None
for line in out.strip().split("\n"):
    line = line.strip()
    if line.startswith(">"):
        if current_phase:
            phases.append(current_phase)
        current_phase = []
        continue
    dist, depth = map(float, line.split())
    current_phase.append((dist, 6371 - depth))
if current_phase:
    phases.append(current_phase)


for phase in phases:
    plt.plot([_i[0] for _i in phase], [_i[1] for _i in phase])


# Spherical PPPPPP plot
out = subprocess.Popen(
        [taup_path + " -mod iasp91 -pH 'PPPPPP' -h 0 -deg 10 -o stdout"],
        shell=True,
        stdout=subprocess.PIPE).communicate()[0].decode()

phases = []
current_phase = None
for line in out.strip().split("\n"):
    line = line.strip()
    if line.startswith(">"):
        if current_phase:
            phases.append(current_phase)
        current_phase = []
        continue
    dist, depth = map(float, line.split())
    current_phase.append((np.deg2rad(dist), depth))
if current_phase:
    phases.append(current_phase)

plt.figure(2)
ax = plt.subplot(111, polar=True)
ax.set_theta_zero_location('N')
ax.set_theta_direction(-1)
ax.set_xticks([])
ax.set_yticks([])

for phase in phases:
    ax.plot([_i[0] for _i in phase], [_i[1] for _i in phase])

ax.xaxis.set_major_formatter(plt.NullFormatter())
ax.yaxis.set_major_formatter(plt.NullFormatter())
ax.set_rmin(0)
ax.set_rmax(6371)

plt.show()
```

krischer:
@crotwell I did find some other (a bit obscure) phases where it does not quite add up. The issue is identical in your Java version and our Python port. On the left is always the Python version, on the right the Java version.

untitled

crotwell:
I'll check into these. I suspect this is the case of rounding errors for phases that traverse a lot of layers. It might be that I need to up the sampling in this case. Thanks for letting me know.

obspy/taup/tau.py
```diff
@@ -4,6 +4,12 @@
                         unicode_literals)
 from future.builtins import *  # NOQA
 
+import copy
+
+import matplotlib.cbook
+import matplotlib.pyplot as plt
+import numpy as np
```
QuLogic: NumPy, matplotlib, then cbook.

obspy/taup/tau.py
```diff
@@ -11,14 +17,35 @@
 from .taup_time import TauP_Time
 
 
+# From colorbrewer2.org
```
QuLogic: Which set?

obspy/taup/tau.py
```diff
@@ -31,6 +58,131 @@ def __str__(self):
...
 
+    def plot(self, plot_type="spherical", plot_all=True, ax=None, show=True):
+        """
+        Plot the ray paths if any have been calculated.
+
+        :param plot_type: Either ``"spherical"`` or ``"cartesian"``.
+            A spherical plot is always global whereas a cartesian one can
+            also be local.
+        :type plot_type: str
+        :param plot_all: By default only rays that match the requested
```
QuLogic: I'm confused by the name here. It's plot _all_, and the description is "by default only something...", but the default is `plot_all=True`, so it's either named the opposite of what it does, or the default is wrong.

obspy/taup/tau.py
krischer: This is still a bit confusing to me. I like the general structure future -> stdlib -> third party -> obspy but I think each block in itself should be alphabetically ordered. The numpy -> matplotlib -> scipy order and all at the end of the third party block makes little sense to me...but I value consistency more than this so I guess I'll change it ;-)

obspy/taup/tau.py
```diff
@@ -31,6 +58,131 @@ def __str__(self):
...
+        """
+        arrivals = []
+        for _i in self:
+            if _i.path is None:
+                continue
+            dist = _i.purist_distance % 360.0
+            if abs(dist - self.distance) / dist > 1E-5:
+                if plot_all is False:
+                    continue
+                # Mirror on axis.
```
QuLogic: Should we just do this lower down in the stack to avoid confusion?

obspy/taup/tau.py
krischer: Ah true...i changed the default behavior in some commit and forgot to adopt the doc. thanks.

obspy/taup/tau.py
QuLogic: For scientific packages, it's SciPy conventions, which unfortunately are not alphabetical.

obspy/taup/tau.py
krischer: I also thought about this and am not really sure as we'd then end up with negative distances which might confuse even more people.

obspy/taup/tau.py
krischer: Is there any reason why matplotlib is before scipy? Chronologically plotting comes after the analysis ;-)

QuLogic: Probably because I was thinking of SciPy and forgot they'd have put scipy in the first-party location. :stuck_out_tongue:

obspy/taup/tau.py
QuLogic: You should be able to get these from matplotlib. From #915:
```
cmap = plt.get_cmap('Paired', lut=6)
colors = ['#%02x%02x%02x' % tuple(col * 255 for col in cmap(i)[:3])
          for i in range(6)]
```

QuLogic:
I was wondering if there might some use to an alternative labelling method, e.g., labelling arrivals at their receiver:
phases

Say you asked for P at x° and PP at 2x° (you can do that, right?), then the first leg would overlap and you wouldn't see one of them.

This might be a bit tricky if you have many arrivals near each other.

krischer:
This would of course be very nice but kind of impossible (or at least very hard to do) as all of our phases have the same receiver as of now. The plotting is trivial to extend to allow arrivals at multiple receivers/distances but it would be pretty hard to actually get nicely separated single arrivals which are needed for your proposed plot.

A plot like this would still be awesome for the tutorial and the gallery but it would probably need to be a one-off plot. The differentiated P and S legs though we should still do :-)

megies:
In the cartesian plot, if there are arrivals coming in from the opposite direction, maybe there should be a second receiver triangle at `-= 360` degrees?

QuLogic:
I added some simple plots to the tutorial as well as one approximating the above. I tried to fit in complicated phases like `SKPPKP`, but contrary to that above picture, those are really long distance (>180 or even 270), and they'd make the plot rather messy.

travel_time_body_waves1

Note: You can see that `pPcP` is one of those phases that does not end at the right location.

misc/docs/source/tutorial/code_snippets/travel_time_body_waves.py
```diff
@@ -0,0 +1,81 @@
...
+    arrivals.plot(plot_type='spherical', plot_all=phase in PLOT_ALL,
+                  legend=False, label_arrivals=True,
+                  show=False, ax=ax)
+
+# Annotate regions
+ax_right.text(0, 0, 'Solid\ninner\ncore',
+              horizontalalignment='center', verticalalignment='center',
+              bbox=dict(facecolor='white', edgecolor='none', alpha=0.7))
+ocr = (model.model.radiusOfEarth -
+       (model.model.sMod.vMod.iocbDepth + model.model.sMod.vMod.cmbDepth) / 2)
```
QuLogic: So this is kind of ugly. Maybe we want to add some shortcuts.

obspy/taup/tau.py
```diff
@@ -11,14 +18,70 @@
...
+    "#ff7f00",
+    "#6a3d9a",
+    "#a6cee3",
+    "#b2df8a",
+    "#fb9a99",
+    "#fdbf6f",
+    "#cab2d6"]
+
+
+class SmartPolarText(matplotlib.text.Text):
```
QuLogic: Might want to make this private or put somewhere else.

obspy/taup/tau.py
```diff
@@ -31,6 +94,177 @@ def __str__(self):
...
+            not shown.
+        :type plot_all: bool
+        :param legend: If boolean, specify whether or not to show the legend
+            (at the default location.) If a str, specify the location of the
+            legend. If you are plotting a single phase, you may consider using
+            the ``label_arrivals`` argument.
+        :type legend: bool or str
+        :param label_arrivals: Label the arrivals with their respective phase
+            names. This setting is only useful if you are plotting a single
+            phase as otherwise the names be large and possibly overlap or clip.
```
QuLogic: I missed a verb here; please add in 'may' or 'could' next time you update.

obspy/taup/tau.py
```diff
@@ -31,6 +94,177 @@ def __str__(self):
...
+                        lw=2.0)
+            ax.set_yticks(radius - discons)
+            ax.xaxis.set_major_formatter(plt.NullFormatter())
+            ax.yaxis.set_major_formatter(plt.NullFormatter())
+            # Pretty earthquake marker.
+            ax.plot([0], [radius - arrivals[0].source_depth],
+                    marker="*", color="#FEF215", markersize=20, zorder=10,
+                    markeredgewidth=1.5, markeredgecolor="0.3", clip_on=False)
+
+            # Pretty station marker.
```
QuLogic: FYI, I only change to annotations on the spherical side. The Cartesian wasn't going to run into the same problem.

ThomasLecocq:
just a note, the direct upward arrival "p" doesn't exist, right ? It's capital "P", then it can be named "pP" !?

claudiodsf:
@ThomasLecocq No, the direct upward P arrival is "p". "pP" is the arrival which first goes upwards and finally arrives at teleseismic distance as a P-wave. You can see a "pP" arrival at ~315° on @QuLogic plot.

ThomasLecocq:
humm... http://pasadena.wr.usgs.gov/office/hough/Storchak.pdf not sure about that...

claudiodsf:
At teleseismic distance, "P" is a mantle phase. I don't think that, at global scale, you can call a crustal P-wave just "P".  You can use though, like in the paper you're referring to, more specific notation like "Pg", "Pb" or "Pn".

krischer:
To clear up the confusion: The TauP phase name convention is a bit different from the IASPEI one. Its documented in the TauP manual and we should probably also document it in `obspy.taup`. 

http://www.seis.sc.edu/downloads/TauP/taup.pdf

ThomasLecocq:
> we should probably also document it in obspy.taup. 

yeah :-)

claudiodsf:
@QuLogic Why is that ray path green? What is your coloring criterion?

krischer:
@claudiodsf Its just green because there are two SKKS arrivals at the same distance.

claudiodsf:
@krischer Ah ok. I still think, though, that we should differentiate ray paths by colors, like in your first plots.

krischer:
Probably. Its still the case with the automatic plots.

The one from @qulogic is just to show off the taup interface in the tutorial and gallery :-) But we can still use different colors in there. I would also really like to distinguish P and S legs like in the famous Stein picture somewhere further up this thread. But that, as said, requires some changes further down the stack which will happen I guess.

QuLogic:
Yes, this example is just for the tutorial/gallery. The plotting code can't really handle multiple distances, so it is essentially many single plots over top of each other, and they end up the same colour. The green is the secondary arrival; there are secondary arrivals for `PPP` and `SSS`, but they bounce very near to other phase arrivals, so I found them confusing and left them out.

obspy/taup/tau.py
```diff
@@ -11,14 +18,70 @@
...
-    Class that is returned by the interface methods.
+    List of arrivals returned by the
+    methods of the :class:`~obspy.taup.tau.TauPyModel` class.
+
+    :param arrivals: List of arrivals.
+    :param model: The model used to calculate the arrivals.
+    :param distance: The requested distance.
     """
-    __slots__ = ["arrivals"]
+    __slots__ = ["model", "distance"]
```
QuLogic: So is it necessary to hold on to the distance here? Could you use `arrival.distance` vs. `arrival.purist_distance` instead? That way, the multiple arrival scenario in the tutorial could just be `+=` and a single `plot`.

mbyt:
Really, really nice plots! Now I finally understand what's taup good for :o).

krischer:
Travis and appveyor failures are unrelated. Can we merge this?

QuLogic:
The body wave example seems to have broken now.

QuLogic:
The only issue I have here is that the imports means you effectively need `matplotlib` installed correctly during installation. IIRC, depending on how you install, runtime dependencies might be installed after the package is built.

PR #963, opened 2015-02-20
Equal-area map projections

QuLogic:
It was mentioned in #902 that azimuthal equidistant maybe didn't make sense. @krischer suggested using Mollweide for global and Albers equal area for local plots. As far as I know, this hasn't been in a release, so it would be best to merge before 0.10.

The only question is what standard parallels to choose for the local plot. The defaults actually make it look very similar to azimuthal equidistant for the two station tests.


megies:
> As far as I know, this hasn't been in a release, so it would be best to merge before 0.10.

The event plots have been there I think, but yes I agree, any work on the plots should be merged before 0.10.0.

QuLogic:
> The event plots have been there I think

Hrm, yes, you're right; I'll add a changelog entry, then.

QuLogic:
Two-ish weeks without any other complaints, so...

PR #964, opened 2015-02-25
Fixing erroneous event type in USGS catalog

yannikbehr:
replacing underscore with space in multi-word event types


krischer:
Can you reapply these patches on the current master? ObsPy 0.10.0 will be released in a couple of days and I really doubt we'll ever release a 0.9.3. Otherwise there is a small PEP8 error, but the rest looks good.

You will have to open a new PR as AFAIK there it no way to change the branch to merge into.

QuLogic:
> Otherwise there is a small PEP8 error, but the rest looks good.

> AssertionError: 80 != 0

I think this build is actually catching the new PEP8 release and triggering all the new errors and warnings. It's unlikely that @yannikbehr introduced 80 errors with this PR, and in fact it's possible there are no new issues at all.

Anyway, all the more reason to target `master`.

krischer:
Haha :-) Good point. I didn't even look at the error messages. I just saw the number 80 and somehow assumed it was related to line length...

PR #965, opened 2015-02-26
Underscore issue in USGS catalog event types

yannikbehr:


krischer:
The USGS event file contained two `creationInfo` elements in the `eventParameters`. This is not allowed according to the QuakeML standard so I removed it. Is this an actual file from the USGS?

yannikbehr:
No, I stitched together (badly, obviously) two USGS events that caused problems. Thanks for fixing that.

PR #969, opened 2015-03-06
SEISAN Waveform import bug

EPAnt:
It seems there is a bug when importing SEISAN waveform files. This can be checked when importing the test files in obspy/seisan/tests . Basicaly, the last sample of the data array usually has an unusually high value, resulting in a strange jerk/pulse in the waveform file. Because of this I usually just do
```
for el in st: el.data=el.data = el.data[:-1]
```
to delete the last sample...

I provided two waveform files, one in SEISAN, and another in MSEED which was converted from the SEISAN waveform using SEISAN/wavetool . However, I have tested the test waveforms already provided in the repository obspy/seisan/tests/data and the results is the same, altough the amplitude of the waveforms does obsfuscate this bug sometimes...


EPAnt:
Added these two waveform files per T. Megies request, to aid in solving the "last sample pulse in SEISAN waveform files" bug.

mbyt:
I continued this work in #971 (forked from here), as I do not have push permission to this branch.

PR #971, opened 2015-03-06
Seisan waveforms misaligned / wrong by one offset

### LINKED ISSUE #970: SEISAN Waveform import bug
It seems there is a bug when importing SEISAN waveform files. This can be checked when importing the test files in ```obspy/seisan/tests``` . Basicaly, the last sample of the data array usually has an unusually high value, resulting in a strange jerk/pulse in the waveform file. Because of this I usually just do
``` python
for el in st: el.data=el.data = el.data[:-1]
```
to delete the last sample...

I provided two waveform files, one in SEISAN, and another in MSEED which was converted from the SEISAN waveform using SEISAN/wavetool . However, I have tested the test waveforms already provided in the repository ```obspy/seisan/tests/data``` and the results is the same, altough the amplitude of the waveforms does obsfuscate this bug sometimes...

mbyt:
Resolves #970. supersedes #969.

Original #969 description:

----
It seems there is a bug when importing SEISAN waveform files. This can be checked when importing the test files in obspy/seisan/tests . Basicaly, the last sample of the data array usually has an unusually high value, resulting in a strange jerk/pulse in the waveform file. Because of this I usually just do
```python
for el in st:
    el.data = el.data[:-1]
```
to delete the last sample...

I provided two waveform files, one in SEISAN, and another in MSEED which was converted from the SEISAN waveform using SEISAN/wavetool . However, I have tested the test waveforms already provided in the repository obspy/seisan/tests/data and the results is the same, altough the amplitude of the waveforms does obsfuscate this bug sometimes...


obspy/core/tests/test_waveform_plugins.py
```diff
@@ -174,7 +178,10 @@ def test_isFormat(self):
                     filelist.extend([os.path.join(directory, _i) for _i in
                                      files])
                 for file in filelist:
+                    if any([n in file for n in KNOWN_FALSE]):
+                        continue
                     if isFormat(file) is True:  # pragma: no cover
+                        print(file)
```
QuLogic: Do we need to print it? Doesn't it go into `false_positives` and print from there?

megies:
Looks good to merge for me.

PR #972, opened 2015-03-09
TauPy docs

QuLogic:
A lot of tedious work, so I probably made some typos. It's not totally complete, but hopefully is all the interesting stuff?


obspy/taup/helper_classes.py
```diff
@@ -68,3 +68,57 @@ def __init__(self, topDepth=None, botDepth=None, ray_param=-1):
...
 )
+
+
+class Arrival(object):
+    """
+    Convenience class for storing parameters associated with a phase arrival.
+
+    :ivar phase: :class:`~obspy.taup.seismic_phase.SeismicPhase` that
+        generated this arrival
+    :ivar distance: Actual distance in degrees
```
krischer: Just for my own understanding: `ivar` instead of `param` is used here to document that it is also an instance attribute?

obspy/taup/helper_classes.py
```diff
@@ -68,3 +68,57 @@ def __init__(self, topDepth=None, botDepth=None, ray_param=-1):
...
+    :ivar time: Travel time in seconds
+    :ivar purist_dist: Purist angular distance (great circle) in radians
+    :ivar ray_param: Ray parameter in seconds per radians
+    :ivar name: Phase name
+    :ivar purist_name: Phase name changed for true depths
+    :ivar source_depth: Source depth in kilometers
+    :ivar incident_angle:
```
QuLogic: Oops, forgot to fill these two in.

obspy/taup/helper_classes.py
QuLogic: Yes, I assume most people will get the `Arrival` object as a result and will almost never need to create one.

QuLogic: The only annoyance is they appear as "Variables", not as "Attributes" like the two `@property` methods do.

megies:
Gonna merge this tomorrow..

krischer:
Merging now to get closer to a release. Please fix any (very likely) typos in master.

PR #973, opened 2015-03-09
Simplistic seedlink client

megies:
.. that implements the usual `Client.get_waveform(net, sta, loc, cha, start, end)` routine for quick and easy requesting of finite time windows via seedlink.

Although this might be adding to the confusion in `obspy.seedlink` I think this might be useful as it covers a different use case (simple requests of finite time windows) than the existing client structures (oriented at more complex use cases, especially working on continuous data streams).

e.g.
```python
>>> from obspy.seedlink import Client
>>> client = Client('rtserver.ipgp.fr')
>>> t = UTCDateTime() - 3600
>>> client.get_waveform("G", "FDF", "00", "BHZ", t, t + 5)
<obspy.core.stream.Stream object at 0x...>
```

Would still need more docs..

### DIFF
--- a/obspy/seedlink/basic_client.py
+++ b/obspy/seedlink/basic_client.py
@@ -1,0 +1,148 @@
+# -*- coding: utf-8 -*-
+"""
+SeedLink request client for ObsPy.
+
+:copyright:
+    The ObsPy Development Team (devs@obspy.org)
+:license:
+    GNU Lesser General Public License, Version 3
+    (http://www.gnu.org/copyleft/lesser.html)
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA @UnusedWildImport
+
+import warnings
+
+from obspy import Stream
+from obspy.seedlink.slclient import SLClient, SLPacket
+from obspy.seedlink.client.seedlinkconnection import SeedLinkConnection
+
+
+class Client(object):
+    """
+    SeedLink request client.
+
+    This client is intended for requests of specific, finite time windows.
+    To work with continuous realtime data streams please see
+    :class:`~obspy.seedlink.slclient.SLClient` and
+    :class:`~obspy.seedlink.easyseedlink.EasySeedLinkClient`.
+
+    :type server: str
+    :param server: Server name or IP address to connect to (e.g.
+        "localhost", "rtserver.ipgp.fr")
+    :type port: int
+    :param port: Port at which the seedlink server is operating (default is
+        `18000`).
+    :type timeout: float
+    :param timeout: Network timeout for low-level network connection in
+        seconds.
+    :type debug: bool
+    :param debug: Switches on debugging output.
+    """
+    def __init__(self, server, port=18000, timeout=20, debug=False):
+        """
+        Initializes the SeedLink request client.
+        """
+        self.timeout = timeout
+        self.debug = debug
+        self._slclient = SLClient(loglevel=debug and "DEBUG" or "CRITICAL")
+        self._server_url = "%s:%i" % (server, port)
+
+    def _connect(self):
+        """
+        Open new connection to seedlink server.
+        """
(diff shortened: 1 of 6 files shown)


obspy/seedlink/basic_client.py
```diff
@@ -0,0 +1,148 @@
...
+from obspy.seedlink.client.seedlinkconnection import SeedLinkConnection
+
+
+class Client(object):
+    """
+    SeedLink request client.
+
+    This client is intended for requests of specific, finite time windows.
+    To work with continuous realtime data streams please see
+    :class:`obspy.seedlink.slclient.SLClient` and
```
QuLogic: No `~` on this one?

obspy/seedlink/basic_client.py
```diff
@@ -0,0 +1,148 @@
...
+    :class:`~obspy.seedlink.easyseedlink.EasySeedLinkClient`.
+
+    :type server: str
+    :param server: Server name or IP Adress to connect to (e.g.
```
QuLogic: s/Adress/Address/

obspy/seedlink/tests/test_basic_client.py
```diff
@@ -0,0 +1,38 @@
...
+        self.client = Client("rtserver.ipgp.fr")
+
+    def test_get_waveform(self):
+        t = UTCDateTime() - 3600
+        for request in [["G", "FDF", "00", "LHN", t, t + 20],
+                        ["G", "CLF", "00", "BHZ", t, t + 10]]:
+            st = self.client.get_waveform(*request)
+            self.assertTrue(len(st) > 0)
+            for tr in st:
+                self.assertTrue(tr.id == ".".join(request[:4]))
```
QuLogic: This sort of line is why I asked about `assertEqual(s)` on gitter, @megies. I think string comparison should work fine, right?

megies: Hm, so it's about cosmetics and not about incompatibilities / unittest backwards compat etc., right? I.e. that `assertEqual(a, b)` is a bit simpler to read than `assertTrue(a == b)`? Fair point, I can change it.

QuLogic: Well, it's not all about cosmetics, as I recall some cases where `assertEqual` didn't work. IIRC, it was with custom type in 2.6, but strings should work, for sure.

petrrr:
Hi, this client is great. Just testing it!
What about adding support for wildcards and lists of channels?

megies:
> What about adding support for wildcards and lists of channels? Should I open a new ticket?

Feel free towork on it. I am fairly sure that you'd need to do some more magic internally, i.e. query what's available from the server. I don't think just passing wildcards to the server works. But I'm not sure and for me it wasn't top-priority. Usually with realtime servers you know beforehand what data you want exactly (at least that's the way for me)

CTrabant:
FWIW, all standard SeedLink servers support wildcarding of location and channel identifiers (* and ? wildcards).  The IRIS ringserver SeedLink implementation further supports those wildcards for network and station identifiers.  There might be some proprietary implementations (e.g. hardware manufacturers) that do not support wildcards, but I would not worry about those much.

megies:
Thanks for the info Chad! So, @petrr feel free to work on it..

megies:
I've checked some wildcard requests with our current implementation and found `?` to work well in loc/cha codes but not `*` for multiple characters. Also I wasn't able to get a wildcard request in station/network to run. This is likely to be influenced by how it's implemented in our underlying low-level seedlink client (`SLCLient`), so I'm not saying it's not working on server side.. ;)

I've updated our docs accordingly, that's as much as we get without fiddling in the deeper implementation I think. Thanks again @CTrabant for pointing this out.

@petrrr, I was not able to fetch data from IRIS seedlink server, maybe you can double check? Thanks!

CTrabant:
Ah, right, only the ? wildcard is allowed for location/channel.  Sorry and thanks for double checking.

PR #975, opened 2015-03-11
do not round float result of subtraction of two UTCDateTime objects to set precision attribute

megies:
I just ran into a strange floating point precision issue when subtracting two UTCDateTime objects and doing further calculations on the result. I am not sure but I am inclined to think that this is a bug, as this results in loss of precision. In all other cases the precision attribute is exclusively used to influence the boolean values of comparisons but never to actually truncate the precision..

Opinions?

Here's a comparison of subtracting two UTCDateTime objects vs. subtracting their timestamps..
```python
import numpy as np
from obspy import UTCDateTime

t = UTCDateTime(0.1)

errors = []

for t2_timestamp in np.linspace(0.2, 0.8, 10000):
    t2 = UTCDateTime(t2_timestamp)

    error = abs((t2.timestamp - t.timestamp) - (t2 - t)) 
    errors.append(error)

plt.hist(errors)
plt.show()
```

figure_1


mbyt:
@megies could you check the change on the test_trigger.py ?

megies:
> test_trigger.py ?

uhm.. what `test_trigger.py`?

barsch:
probably http://tests.obspy.org/67641/ - those are also failing for me if I run the signal test suite using this branch

PR #977, opened 2015-03-11
upgrading to libmseed 2.14

### LINKED ISSUE #976: libmseed 2.14 released
Version 2.14 of libmseed has been released:

Changes from the 2.13 version (currently in master):
https://seiscode.iris.washington.edu/news/31

The closing of a memory leak that could be encountered during errors in packing miniSEED records is the most significant change.  There is also supported added for minimalistic handling of leap seconds defined in a standard leap second file.

mbyt:
Pure upgrade to newest libmseed version 2.14, closes #976

Does not integrate any handling of leap seconds mentioned in #976.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -85,6 +85,7 @@
          with non-zero exit code.
        - Faster reading of a single record header
        - Added option "-a" to print information of all records
+   * upgrade to libmseed 2.14
  - obspy.ndk:
    * New submodule able to read NDK files from the Global CMT project.
  - obspy.neries:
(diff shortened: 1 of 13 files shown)


megies:
@mbyt, to avoid having issue/PR separated, you can use the github API to make an issue into a PR: https://github.com/obspy/obspy/wiki/Interesting-Reading-on-Git,-GitHub

But of course, sometimes one does not want ot hijack somebody else's issue.. ;)

mbyt:
yes, I'll do that next time. Avoids some indirection and probably also some confusion...

PR #978, opened 2015-03-11
Re-fix mseed infinite loop / Update to libmseed 2.15

QuLogic:
Apparently, the test file I originally put in the test was fixed by 2.14 somehow, even though the infinite loop wasn't actually fixed.

I replaced the test file with the one in the [original bug report](https://seiscode.iris.washington.edu/issues/617) and it does trigger the infinite loop, so I've re-applied the patch.


CTrabant:
Dang, I meant to include this fix in the latest release.  I will fix this immediately and create another release.

QuLogic:
@CTrabant Can you please also include the full text of the LGPL license in the source tarball? Term 1 states:

> You may copy and distribute verbatim copies of the Library's complete source code as you receive it, in any medium, provided that ... and **distribute a copy of this License** along with the Library.

Which, as I read it, means you need to include the full text of the license, not just the header in each file or the couple of paragraphs in the README.

QuLogic:
I'm going to hold off on merging this one until the new libmseed instead.

CTrabant:
Nope on the LGPL license.  Too pedantic for me.

CTrabant:
New libmseed 2.15 release:

thanks again for the patch.

QuLogic:
@CTrabant Looks good and everything passed in the new 2.15.

PR #980, opened 2015-03-12
Handling of misaligned traces in merge operations

megies:
This PR is intending to improve and make clearer the handling of misaligned traces (regarding their sampling point, i.e with sampling points shifted on a sub-sample-interval scale).

E.g...
```
Trace 1:  ----|---------|---------|---------|---------|--
Trace 2:  ---|---------|---------|---------|---------|---
```

Before, two traces with their sampling points misaligned by even as much as half the sampling interval were simply considered to be sampled at the same points in merge operations. For some applications that might cause problems.

With this PR, before anything else in merging, it is checked whether the sub-sample shift is to be considered acceptable (very conservative default of 1% of sampling interval, can be controlled via kwargs) and then the traces are aligned on the first trace's sampling points (nb: not by any signal processing but simply shifting the time axis).

This PR also now is able to merge directly adjacent traces when they have a slight misalignment, which is what triggered it in the first place as this happens quite frequently in real data.

e.g.
```
Trace 1:  ----|---------|---------|        (|)
Trace 2:                                     |---------|---
```

### DIFF
--- a/obspy/core/tests/test_stream.py
+++ b/obspy/core/tests/test_stream.py
@@ -2167,6 +2167,88 @@ class StreamTestCase(unittest.TestCase):
         st2.integrate(method='cumtrapz')
         self.assertEqual(st1, st2)
 
+    def test_misaligned_traces(self):
+        """
+        Tests the option to correct misaligned traces in `Stream._cleanup()`,
+        which is the first thing done in any `Stream.merge()` operation.
+
+        We create a simple trace and then prepare other traces that are shifted
+        on sub-sample scale. With these shifted traces create different traces
+        that are a) contained, b) overlapping, c) directly adjacent and d)
+        separated by a "real" gap. In two loops we test a wide variety of
+        combinations of actual misalignment ratio (percentage of sampling
+        interval) and to-be-fixed misalignment threshold in cleanup().
+        """
+        samp_rate = 1.23
+        delta = 1.0 / samp_rate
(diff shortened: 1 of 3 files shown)


obspy/core/tests/test_stream.py
```diff
@@ -2167,6 +2144,81 @@ def test_integrate_args(self):
         st2.integrate(method='cumtrapz')
         self.assertEqual(st1, st2)
 
+    def test_misaligned_traces(self):
+        """
+        Tests the option to correct misaligned traces in `Stream._cleanup()`.
+
+        We create a simple trace and then prepare other traces that are shifte
```
QuLogic: This sentence is truncated.

obspy/core/tests/test_stream.py
```diff
@@ -2167,6 +2144,81 @@ def test_integrate_args(self):
...
+            tr_gap3.stats.starttime += delta
+            tr_gap4 = tr_adjacent4.copy()
+            tr_gap4.stats.starttime += delta
+            traces_gap = [tr_gap1, tr_gap2, tr_gap3, tr_gap4]
+            for to_be_fixed_misalignmnt_ratio in np.linspace(0.011, 0.49, 5):
+                # first test traces with normal gap, these should never change
+                for trx in traces_gap:
+                    self.assertFalse(_gets_merged(
+                        trx, to_be_fixed_misalignmnt_ratio))
+                # now lets check the cases were depending on parameters there
```
QuLogic: where?

PR #982, opened 2015-03-13
updating meta.yaml for conda packaging

mbyt:
work in progress for conda packaging.

Repo: https://binstar.org/obspy/obspy

- [x] does it make sense to open a binstar obspy account? `<update>`done`</update>`
- [x] symlinks make problems `<update>`only a problem when using the path specifier and not directly downloading from github, therefore OK`</update>`


krischer:
I already created an ObsPy organization on binstar and also a more or less finished meta.yaml. Lets merge ours!

Issue #958, opened 2015-02-15
future and mock are build dependency


krischer:
~~After merging #944 `future` is now a build dependency~~ EDIT: After merging #984 this is not longer the case.

We need to either bootstrap this or workaround it somehow. The best solution would of course be to have a non-pickled model serialization routine.

QuLogic:
And so is `mock` now.

The main thing to get this done is to rewrite the `TauBranch` class somehow. There are (2, N) instances in the model. Each instance contains a few scalars, but _also_ some arrays, which are unfortunately different lengths. So while we can turn all the scalars into (2, N) arrays, we can't turn the others into (2, N, M) since M is not fixed. Once we have only arrays, we just need to use NumPy to save it.

krischer:
Could we just serialize everything to JSON? I know its slow but it is very cross platform and the models only need to be read once.

QuLogic:
Can we do that directly? Or do we still need to encoder/decoder class to do stuff? If it's the latter, we might as well try implementing [`object.__getstate__()`](https://docs.python.org/2/library/pickle.html#object.__getstate__) in some way to get pickling through NumPy to work.

krischer:
No it would require a custom encoder/decoder. My reasoning would just be that its a more standard format then any of the filetypes numpy can serialize to.

mbyt:
numpy arrays can directly be serialized to to binary using `ndarray.tofile()`. That's absolutely cross platform, same interface as C, however, the flexible size @QuLogic mentioned is a no go for this approach :\.

krischer:
Its unfortunately not cross platform, see http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tofile.html

```
This is a convenience function for quick storage of array data. Information on
endianness and precision is lost, so this method is not a good choice for files
intended to archive data or transport data between machines with different
endianness. Some of these problems can be overcome by outputting the data
as text files, at the expense of speed and file size.
```

It probably unlikely to matter for any of the platforms of interest to us but if possible we might as well support other architectures.

mbyt:
To my understanding, this is only the case, if you did not specify the endianess in the dtype you are serializing or not using this dtype for deserialization (`np.fromfile(.., dtype=dtype`). However, I never worked on a big endian architecture with numpy available .... so I cannot be sure ...

krischer:
[...]

PR #983, opened 2015-03-13
TauModel serialization to json 

megies:
Serialization of `TauModel` working on Py2, only missing the `.sMod` `SlownessModel` bit right now..


QuLogic:
I'm not super-keen on text-serialization; I don't suppose there's any way to adapt this to NumPy's saving method?

ThomasLecocq:
:trollface: why not using pandas ? :-) (no link, just kidding (and it's friday night))

QuLogic:
I haven't read it all yet, but maybe something like `np.savez_compressed(fn, **data)` would work, since you've done the work of pulling out the relevant key/values.

megies:
Yeah, I admit it's a bit ugly.. Considered `np.savez`, but I figured we can't use it because..
 - we always have arrays after deserializing.. even attributes that were originally a single float or a single bool (might not be very problematic)
 - `TauModel` has an attribute `.tauBranches` which is a 2D array of type `TauBranch`.. no chance for a simple `savez` here

megies:
> :trollface: why not using pandas ? :-) (no link, just kidding (and it's friday night))

:trollface: sadly, no belgian beer for this sprint. but our brew has 95 by the [`THE BROS`](http://www.beeradvocate.com/beer/profile/820/2560/). cheers.

QuLogic:
> we always have arrays after deserializing.. even attributes that were originally a single float or a single bool (might not be very problematic)

I _think_ it should be okay because they'd be 0-D, but might be trouble for any `isinstance` checks.

> TauModel has an attribute .tauBranches which is a 2D array of type TauBranch.. no chance for a simple savez here (show stopper)

@mbyt had some ideas in #958 for that; not sure how simple they are.

megies:
> @mbyt had some ideas in #958 for that; not sure how simple they are.

I don't see anything in there about handling the object type of `TauBranch`.
It's a complex class and I don't see how it can be handled just by numpy standard routines.. or am I missing some magic here?

QuLogic:
https://github.com/obspy/obspy/issues/958#issuecomment-74437364 - not _super_ simple, I'd guess.

megies:
> #958 (comment) - not super simple, I'd guess.

Sorry, i see the structured array part but I still son't see the handling arrays of arbitrary object types..

QuLogic:
Ah, you'd basically have to pivot them (obligatory pandas reference?) Go from array of `TauBranch.x` (single float) to `np.array` (for all branches) and from `TauBranch.a` (array) to 2D `np.array` (for all branches).

obspy/taup/slowness_model.py
```diff
@@ -634,7 +637,7 @@ def depthInFluid(self, depth):
             otherwise.
         :rtype: :class:`~numpy.ndarray` (dtype = :class:`bool`)
         """
-        ret = np.zeros(shape=depth.shape, dtype=np.bool_)
+        ret = False
```
QuLogic: I'm not sure about this?

QuLogic:
Hmm, ~430K per model (vs 256K via pickle), but we don't need one for each Python version, so I guess that's a net win?

megies:
> Hmm, ~430K per model (vs 256K via pickle), but we don't need one for each Python version, so I guess that's a net win?

Plus we can still gzip it, which makes it ~180K per model. But there's still a strange error, have to find out what is wrong with the de/serialized models..

megies:
Hmm.. Actually I've been investigating the numpy way for serialization a bit now. I think I got a way layed out that should work pretty well.

Here's an example snippet I played with that shows how to serialize an object `A` that has an 2D array attribute of class `B` which c be represented with a structured array:

```python
import random
import numpy as np


class A(object):
    def __init__(self, seed=None):
        if seed is None:
            return
        np.random.seed(seed)
        self.a = np.random.randint(0, 100, 1)
        self.b = np.random.randint(0, 100, 10)
        my_bs = [B(np.random.randint(0, 100, 1))
                 for x in np.random.randint(0, 100, 10)]
        self.my_bs = np.array(my_bs).reshape((2, 5))

    def __str__(self):
        info = ", ".join(map(str, [self.a, self.b]))
        info += "\n" + "=" * 40 + "\n"
        info += "\n".join([str(b) for b in self.my_bs.flatten()])
        return info

    def serialize(self):
        i, j = self.my_bs.shape
        bs = np.vstack([np.hstack([self.my_bs[i_][j_]._to_array()
                                   for j_ in range(j)])
                        for i_ in range(i)])
        np.savez("/tmp/pickling/a.npz", a=self.a, b=self.b, my_bs=bs)

    @staticmethod
    def deserialize():
        arr = np.load("/tmp/pickling/a.npz")
        a = A()
        a.a = arr['a']
        a.b = arr['b']
        i, j = arr['my_bs'].shape
        a.my_bs = np.array([[B._from_array(arr['my_bs'][i_][j_])
                             for j_ in range(j)]
                            for i_ in range(i)])
        return a


class B(object):
    def __init__(self, seed=None):
        if seed is None:
            return
        np.random.seed(seed)
        self.x = np.random.randint(0, 100, 1)
        self.y = np.random.randint(0, 100, 10)

    def __str__(self):
        return ", ".join(map(str, [self.x, self.y]))

    def _to_array(self):
        arr = np.empty(shape=(), dtype=[('x', 'i4'),
                                        ('y', 'i4', self.y.shape)])
        arr['x'] = self.x
        arr['y'] = self.y
        return arr

    @staticmethod
    def _from_array(arr):
        b = B()
        b.x = arr['x']
        b.y = arr['y']
        return b
#
#    def serialize(self):
#        np.save("/tmp/pickling/b.npy", self._to_array())
#
#    @staticmethod
#    def deserialize():
#        return B._from_array(np.load("/tmp/pickling/b.npy"))


if __name__ == "__main__":
    b = B(random.randint(1, 1000))
    print b
    print
    a = A(random.randint(1, 1000))
    print a
    print
    a.serialize()
    a_ = A.deserialize()
    print a_
```

So I think I'll serialize `TauModel` using `np.savez`, the above 2D array of `TauBranch`es was what worried me the most before..

mbyt:
Great example, makes sense to me. Even easier to understand the the `split(').reshape()` magic :).

> But there's still a strange error, have to find out what is wrong with the de/serialized models..

Does this resolve this strange error of the 4 extra values?

megies:
Closing in favor of #984.

PR #984, opened 2015-03-15
TauModel serialization to (compressed) numpy npz

megies:
Platform/architecture independent serialization of `TauModel` in numpy's compressed npz format and shipping models prebuilt now. Supersedes #983.


megies:
Some things to mention:
 - serialization will fail if `TauModel.tauBranches` do not have the same length in their array-type attributes (`dist`, `tau`, `time`), no idea if we can assume this fact for all scenarios
 - simple scalar attributes (e.g. of type `float`) currently end up as 0-D numpy arrays of the respective type. this could be changed but it does not seem to be a problem..
 - on my Python2 before this change `TauModel.sMod.vMod.modelName` used to be unicode, now it's `str`

obspy/taup/helper_classes.py
```diff
@@ -63,6 +63,29 @@ def __init__(self, topDepth=None, botDepth=None, ray_param=-1):
...
+    def _from_array(arr):
+        """
+        Create instance object from a structured array used in serialization.
+        """
+        depth_range = DepthRange()
+        depth_range.topDepth = arr[0]
+        depth_range.botDepth = arr[1]
+        depth_range.ray_param = arr[2]
+        for key, value in zip(['topDepth', 'botDepth', 'ray_param'], arr):
+            setattr(depth_range, key, value)
```
QuLogic: This is being done twice, isn't it?

obspy/taup/slowness_model.py
```diff
@@ -634,7 +637,7 @@ def depthInFluid(self, depth):
             otherwise.
         :rtype: :class:`~numpy.ndarray` (dtype = :class:`bool`)
         """
-        ret = np.zeros(shape=depth.shape, dtype=np.bool_)
+        ret = False
```
QuLogic: I still don't know why this is being changed.

obspy/taup/tau_model.py
```diff
@@ -413,3 +371,206 @@ def getBranchDepths(self):
...
+            TauModel
+            ========
+            cmbBranch <type 'int'>
+            cmbDepth <type 'float'>
+            debug <type 'bool'>
+            iocbBranch <type 'int'>
+            iocbDepth <type 'float'>
+            mohoBranch <type 'int'>
+            mohoDepth <type 'float'>
+            noDisconDepths <type 'list'> (of float!?)
```
QuLogic: Yea, apparently. It seems to get changed only when splits occur, so I'm not sure they can be predicted.

obspy/taup/helper_classes.py
megies: Right, I changed my mind while writing this.. I'll fix it

obspy/taup/tau_model.py
```diff
@@ -413,3 +371,206 @@ def getBranchDepths(self):
...
+
+    @staticmethod
+    def deserialize(filename):
+        """
+        Deserialize model from numpy npz binary file.
+        """
+        with np.load(filename) as npz:
+            model = TauModel(sMod=None, skip_calc=True)
+            # a) handle simple contents
+            complex_contents = [
```
QuLogic: simple or complex? These two lines _seem_ contradictory (though not actually.) Maybe just move the comment below this variable.

obspy/taup/slowness_model.py
megies: When going through the insides for serialization I think I found this always is a single valued array.. but yeah, I did not really make sure, so I'll revert this change.

obspy/taup/tau_model.py
megies: Doesn't matter if the size can be predicted or not in this case.

obspy/taup/tau_model.py
```diff
@@ -413,3 +371,206 @@ def getBranchDepths(self):
...
+        # a) handle simple contents
+        keys = ['cmbBranch', 'cmbDepth', 'debug', 'iocbBranch', 'iocbDepth',
+                'mohoBranch', 'mohoDepth', 'noDisconDepths', 'radiusOfEarth',
+                'ray_params', 'sourceBranch', 'source_depth', 'spherical']
+        arrays = dict([(key, getattr(self, key)) for key in keys])
+        # b) handle .tauBranches
+        i, j = self.tauBranches.shape
+        # the following only works if all branches have the same length in
+        # their array-type attributes, e.g. branch.tau has same length for all
+        # branches. this seems to be the case at least for the default models
```
QuLogic: I thought this was the main concern from the previous tickets. Maybe we're just lucky? Did you try regenerating the models after undoing the change to `depthInFluid`?

QuLogic:
Two minor documentation bits: a) how does one use a custom model (I don't think that's mentioned yet?) and b) how do we update the existing models if we change stuff.

obspy/taup/tau_model.py
QuLogic: Oh, actually, `splitBranch` is called when the model is loaded and a source depth is set. So as long as the results don't change there...

obspy/taup/slowness_model.py
QuLogic: `coarseSample` below here should be calling it with an array of layer depths, I think.

obspy/taup/tau_model.py
megies: I mean, I could also store each branch separately in serialization, it's just a bit less elegant. But I guess I should do it nevertheless.. safer beats nice looks I guess?

megies:
> Two minor documentation bits:a) how does one use a custom model (I don't think that's mentioned yet?) and b) how do we update the existing models if we change stuff.

Should be handled with 549262c and 0a4a9b73435e69623ec2063bcf8da929243bd024.

All build bots pass, so this is good to merge for me. (and then let's release as soon as possible :))

megies:
Squashed another change in there (https://github.com/obspy/obspy/commit/39739c59d5abd408aa6e93917ccd629451e2b573#diff-e2b8f78af649663610fccfa0a5c7dbedL530), rebased and force-pushed.

megies:
Rebased and force-pushed once more..

krischer:
Works fine on OSX with Python 2.7 and 3.4 :-)

megies:
..and even appveyor passing, so I'll merge this now. :cherry_blossom:

mbyt:
@megies great work!

Now we can really profit from the removal of the fortran code --- without the disadvantage of gaining a complicated and hard to maintain and debug partly broken setup.py / installation!

PR #992, opened 2015-03-19
Mopad encoding

QuLogic:
Instead of assuming bytes is UTF-8 and str was correctly decoded, use the same encoding as stdout.

### DIFF
--- a/obspy/imaging/tests/test_mopad_script.py
+++ b/obspy/imaging/tests/test_mopad_script.py
@@ -5,10 +5,12 @@ The obspy-mopad script test suite.
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
+from future.utils import PY3
 from future import standard_library
 
 import io
 import os
+import sys
 import unittest
 from itertools import product
 with standard_library.hooks():
@@ -45,14 +47,17 @@ class MopadTestCase(unittest.TestCase):
 Fault plane 1: strike =  77°, dip =  89°, slip-rake = -141°
 Fault plane 2: strike = 346°, dip =  51°, slip-rake =   -1°
 '''
-        expected = expected.encode('utf-8')
 
         result = out.stdout[:-1]
-        if result.startswith(b'b'):
-            # Total mojibake when the system is configured strangely. Mostly
-            # only occurs when building packages, so that LANG=C, and Python
-            # picks ASCII for output encoding.
-            expected = str(expected).encode('ascii')
+        try:
+            expected = expected.encode(sys.stdout.encoding)
+        except:
+            expected = expected.encode('utf-8')
+            if PY3:
+                # This may look like Py3k does extra stuff, but it's actually
+                # saner!
+                expected = repr(expected)
+                expected = expected.encode('ascii')
 
         self.assertEqual(expected, result)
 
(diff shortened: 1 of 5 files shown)


obspy/imaging/tests/test_mopad_script.py
```diff
@@ -45,14 +47,17 @@ def test_script_convert_type_sdr(self):
...
-            expected = str(expected).encode('ascii')
+        try:
+            expected = expected.encode(sys.stdout.encoding)
+        except:
+            expected = expected.encode('utf-8')
+            if PY3:
+                # This may look like Py3k does extra stuff, but it's actually
+                # saner!
+                expected = repr(expected)
+                expected = expected.encode('ascii')
```
mbyt: @QuLogic I just do not get the idea. If I enter the `if PY3`, `expected` is a byte string, the `repr` of a byte str is normal string containing something like `"b'blah_blub'"`. Then encoding this normal string results in the byte string `b"b'blah_blub'"` which is different from a simple byte string obtained via `out.stdout[:-1]`. What am I missing?

QuLogic: `expected` is initially Unicode. If it cannot be converted to stdout encoding, then it's unilaterally converted to a `bytes` containing UTF-8.

Python 2 is "helpful" and converts it back to a regular string.

Python 3, on the other hand, treats `bytes` just like any other object and prints out the `repr`. `repr(bytes)` is a `str`, but stdout is `bytes` (i.e., `fdopen(1, 'wb')`), so it must be encoded again into the stdout encoding. Technically speaking, I probably shouldn't have used ascii here, but the stdout encoding again...

mbyt:
@QuLogic unfortunately, I still do not understand it. I retyped the lower branch in ipython3:
```python
In [1]: expected = "Fault plane 1: strike =  77°"

In [2]: expected = expected.encode('utf-8')

In [3]: expected
Out[3]: b'Fault plane 1: strike =  77\xc2\xb0'

In [4]: expected = repr(expected)

In [5]: expected
Out[5]: "b'Fault plane 1: strike =  77\\xc2\\xb0'"

In [6]: expected = expected.encode('ascii')

In [7]: expected
Out[7]: b"b'Fault plane 1: strike =  77\\xc2\\xb0'"
```
So what happens, is that you will get a byte string containing another byte string. The `b"b'` is the problem from my point of view. In contrast result is just a byte string, thus has a single `b"`.

What am I missing?

QuLogic:
I think you are confusing what's printed by IPython with what the actual value is. When it's a `bytes` object, it will print out the `repr`, so you get double 'b'.

The result is a `bytes` and it starts with `b'`; if you printed the `repr`, you'd get double 'b' there too. The expected is the same, but when printed in IPython it prints an additional `b"`.

mbyt:
To my understanding this has nothing to do with ipython:
```
(python3.4) ~$ python -c "print(repr('normal string'.encode()).encode())"
b"b'normal string'"
```
Same again, "double encoded" byte string, which we compare in the above code with a byte string.

QuLogic:
Yes, sorry, nothing to do with IPython; Python 3 in general prints out the `repr` of `bytes` object.

mbyt:
Just as summary, clarified the problem on gitter:

Ah, now I got it. It's mopad's try/except and not python3. I was missing these patterns in mopad
```python
        try:
            print(aa)
        except:
            print(aa.encode('utf-8'))
```

PR #994, opened 2015-03-20
move all files used in the codebase from "submodule"/docs to "submodule"/data

megies:
so that we can just leave out "submodule"/docs in released zipball
generally. also much cleaner.


krischer:
Can I just manually do this, as I am currently doing the restructuring?
Would be a bit painful to merge this in. Otherwise the idea is good.

megies:
See gitter, our 0.10.0 pypi zip is broken after all, missing files, so we have to do a 0.10.1 with old codebase structure. Want me to merge this into `releases` instead, so that you can do it manually on `master`?

krischer:
Ah ok. Oh well 0.10.1 is also a nice number ;-) Yes, if possible just merge this into `releases`.

megies:
Will have to merge it back from releases into master anyway.. but yeah.. ;)

megies:
Merging this manually into `releases` now.

PR #842, opened 2015-03-20
Reorganization of ObsPy submodule structure

krischer:
As mentioned a lot of times (recently in https://github.com/obspy/obspy/pull/841#issuecomment-48897862), we should think about reorganizing the internal submodule structure.

 - as always, this would have to handled via `DeprectionWarning`s and redirecting imports for one full major release cycle (like we did for `obspy.signal.psd` when renaming to `obspy.signal.spectral_estimation`; e.g. if we make the switch for 0.10.0 all imports need to be redirected until at 0.11.0).
 - move everything that has no routines intended to be used directly to ...
   - `obspy.plugins`?
   - `obspy.formats`?
   - `obspy.io`?
 - things that should be moved..
   - obspy.station to obspy.core.station ??
   - obspy.station.stationxml to ... ?
 - ...

more comments...?

current plan:
```
obspy
├── core/
│   ├── trace.py
│   ├── stream.py
│   ├── utcdatetime.py
│   ├── event.py
│   ├── station/
│   │   ├── inventory.py
│   │   ├── network.py
│   │   └── ... 
│   └── util/
│       ├── geodetics
│       └── ...
├── io/
│   ├── ah
│   ├── cnv 
│   ├── css 
│   ├── datamark
│   ├── gse2
│   ├── json
│   ├── kinemetrics
│   ├── mseed
│   ├── ndk 
│   ├── nlloc
│   ├── pdas
│   ├── pde 
│   ├── quakeml
│   ├── sac 
│   ├── seg2
│   ├── segy
│   ├── seisan
│   ├── sh
│   ├── stationxml
│   ├── wav 
│   ├── y
│   ├── xseed
│   └── zmap
├── clients
│   ├── arclink
│   ├── earthworm
│   ├── fdsn
│   ├── iris
│   ├── neic
│   ├── neries
│   ├── seedlink
│   └── seishub
├── db
├── imaging
├── lib
├── realtime
├── signal
└── taup (rename?)
```


krischer:
While there is still some stuff missing (I think I can finish it tomorrow) this is more or less ready to merge and if we really want it we should merge is ASAP and fix any issue in the master.

I like the new structure but this will effectively break all existing applications. So we need to reroute the imports for at least one release and raise lots of deprecation warnings. This makes this really a bit ugly so I am not entirely sure we want this change so please chime in anyone!

As a transition strategy I propose (if we choose to do it):

* Merge now.
* We all leave our applications/scripts depending on ObsPy as they are and reroute imports internally with deprecations warnings until they work again. This should more or less assure that stuff will not break if people update to 0.11 as I guess we are our own power users.
* After 0.11 is released we remove all the rerouted imports and update our own applications/scripts.
* ObsPy 0.12 will have no more rerouted imports.

mbyt:
Lot's of work happening here.

What I do not understand is why there are so many changes in the "files changed" tab of this pull request: +-10k. I thought, most of it should be moves only. What am I missing?

It's a big interface change. This together with the interface change #952 deserves a 1.0.0 ("interface change"."new feature"."bug fix") release from my point of view :D.

markcwill:
@mbyt You beat me to it! I was typing this:

@krischer I like it and think it's an improvement, but IMHO a "breaking" change is something that might warrant a major version number bump. So a 0.11.x with warnings would be OK, but not sure about going to 0.12 and breaking backwards compat. I realize it's a bigger deal going from 0.x-1.0 than 1.x-2.x, or something.

With that said, I guess that would raise the possibility of making a 1.x.x-dev branch instead of merging to master?

mbyt:
@markcwill I am unsure about a new branch for 1.x.x, we just could use master, as all bug fixing etc happens on releases, see ee https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model. What am I missing?

QuLogic:
Since this 75%? done, I'd hate to have to throw it away and wait.

> With that said, I guess that would raise the possibility of making a 1.x.x-dev branch instead of merging to master?

I feel like that's just going to cause a whole lot of merge trouble. 1.0 is just a number, why don't we just switch to it for the next release? ObsPy is used fairly widely and seems pretty stable (developmently and in use), so why not?

QuLogic:
Something else interesting is to look at what might happen to existing PRs:
 * 3 PRs by me; no problem there.
 * 5 PRs by @megies; I assume this will also not be a problem.
 * 4 PRs adding completely new files or formats. Probably there would be some minor merge issues due to the plugin stuff, and the locations would need to be changed, of course.
 * 4 PRs with local changes to a single file. There might be some issues here, but at least they'll be isolated.
 * 2 PRs changing multiple files; might be a bit more troublesome, but at least there are few of them.

krischer:
@mbyt I also don't understand the +- 10,000. Last time I checked github showed += 500ish; maybe something to do with it no longer being able to auto merge?

megies:
I agree with @mbyt and @markcwill, a major bump to 1.0 would be one option to avoid the overhead of rerouting imports for one release cycle and just make a hard cut. Actually, I think many people just ignore `DeprecationWarnings` and handle the issue only when the programs stop running ("Ah, a deprecation warning. I should handle it.. I'll do it later this week.").

> With that said, I guess that would raise the possibility of making a 1.x.x-dev branch instead of merging to master?

@markcwill, that part I don't understand (agreeing with @mbyt's and @QuLogic's follow-up comment). master is master, if we decide to make a hard cut, master will end up producing 1.0, if we decide on the rerouted import /deprecation warning, master will produce 0.11.0 eventually. Either way, merging this into master should be done as soon as possible.

@QuLogic, I think we can handle the broken PR's pretty easily. Just one or two hours of manual merging around, nothing compared to the work that went into taup, quakeml or stationxml. Also, we probably never had less pending PRs than right now (with so many cleaned up and made ready for 0.10).

So, I could live with both, either hard-cut/major bump or import reroute/deprecation warnings/minor bump.

One last thought, since this would be a really major change, we should all wreck our brains if we should change something else as well along with it, something that might only be important in the future but not right now. (ehrm, that sounds pretty strange probably. don't know how to word it better.. ;)

QuLogic:
> we should all wreck our brains if we should change something else as well along with it

*cough* colour maps *cough*

QuLogic:
Also, `git diff` prints out this, so possibly GitHub is limited in its rename determination as well:
```
warning: only found copies from modified paths due to too many files.
warning: you may want to set your diff.renameLimit variable to at least 1524 and retry the command.
```

QuLogic:
I finished up the network tests and docs, though I didn't change the tutorial yet. If we go the deprecation route, those would probably be handy for testing. The docs were changed rather bluntly, so they might require some new top-level pages or the like.

markcwill:
@megies, @krischer, @QuLogic sorry, brain fart, I think I forgot you guys are merging `master` -> `releases`, not `develop` -> `master`.

My point was, I don't quite see a major version bump as "just a number". Probably not a big deal for this one, but IMO there should be a clearly stated policy if say, bug fixes will not be ported to a `0.x` release once a `1.0.0` happens, and the delineation should be made as clearly and loudly as possible.

CTrabant:
> My point was, I don't quite see a major version bump as "just a number". Probably not a big deal for this one, but IMO there should be a clearly stated policy if say, bug fixes will not be ported to a 0.x release once a 1.0.0 happens, and the delineation should be made as clearly and loudly as possible.

I second this notion.  While there should not be too much meaning packed into version numbers, changing major version numbers at major changes can be very useful.  Fast forward a few years, keeping in mind production environments that do not change quickly, and it will be very easy to remember when ObsPy change to Python3 only based solely on major version number.  Imagine if what we know as Python 3 was versioned Python 0.12.45.3, it would be basically useless.

QuLogic:
I think you may have misread my "just a number" comment; I meant we _shouldn't be afraid_ to make the jump to 1.0 if the changes are major enough to warrant it (which I think means we both agree.)

Of course, there are also some connotations with a "1.0" that aren't there with a "2.0", namely stability, etc., but I think (except for this change), ObsPy really is quite stable at the moment.

mbyt:
We divagated from the actual topic into a discussion of version numbers ;o).

Back to the topic: I'd vote for merging this, let's resolve the remainder in master --- and the next release is not in the near future --- so there remains lot's of time to discuss....

krischer:
Give me like one more day and then we can merge this. I really want the import deprecations to work and I think I can get it working.

I want that my code that depends on ObsPy runs on the current master and the latest stable. Otherwise it just makes my (and I assume others as well) life harder and honestly we should not merge anything that makes things more cumbersome in the long run for us.

krischer:
Turned out easier (and cleaner) than expected :-) Sometimes Python is quite nice to use. Feel free to merge as soon as CI passes.

Essentially the imports of the previous version now still work but raise a deprecation warning.

megies:
I just noticed that `obspy/core/scripts` has been moved to `obspy/scripts`. However all other `scripts` folders remain as subfolders of the respective submodules, namely:
```
obspy/imaging/scripts
obspy/db/scripts
obspy/io/mseed/scripts
obspy/io/xseed/scripts
```
@krischer, what's the rationale here? Isn't that introducing kind of an inconsistency?

PR #989, opened 2015-03-18
getWaveform for earthworm client fails

mbyt:
This looks like an regression introduced with Py3 compatibility.
Here the backtrace:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/Users/petr/work/test-nrt/test-ew.py in <module>()
      7 client = Client("hew2.int.ingv.it", 16222, timeout=30.0)
      8 
----> 9 response = client.availability("IV", "*", channel="*")
     10 
     11 t = obspy.UTCDateTime()

/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/obspy/earthworm/client.py in availability(self, network, station, location, channel)
    219         # get overview of all available data, winston wave servers can not
    220         # restrict the query via network, station etc. so we do that manually
--> 221         response = getMenu(self.host, self.port, timeout=self.timeout)
    222         # reorder items and convert time info to UTCDateTime
    223         response = [(x[3], x[1], x[4], x[2], UTCDateTime(x[5]),

/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/obspy/earthworm/waveserver.py in getMenu(server, port, scnl, timeout)
    202     sock = sendSockReq(server, port,
    203                        getstr.encode('ascii', 'strict'),
--> 204                        timeout=timeout)
    205     r = getSockCharLine(sock, timeout=timeout)
    206     sock.close()

/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/obspy/earthworm/waveserver.py in sendSockReq(server, port, reqStr, timeout)
    138         s.send(reqStr)
    139     else:
--> 140         s.send(reqStr + '\n')
    141     return s
    142 

TypeError: can't concat bytes to str

```

PR #995, opened 2015-03-21
obspy.earthworm client times out on Py3.

QuLogic:
Okay this one is a follow up on  #989, which fixed Py2 behavior. Py3 still fails with a timeout:
```
petr% python3.4 ./test-ew.py
socket timeout in getSockCharLine()
socket timeout in getSockCharLine()
 ```
I also checked the public server I indicated earlier. It should be available on `pubavo1.wr.usgs.gov:16022`. A minimal connection test can be performed with telnet, the request `MENU: 0` should provide a sort of inventory. This is what I obtain:
```
petr% telnet pubavo1.wr.usgs.gov 16022
Trying 130.118.181.39...
Connected to pubavo1.wr.usgs.gov.
Escape character is '^]'.
MENU: 0
  239 ACH EHE AV 1421685777.065 1426869775.0500011 s4  76 ACH EHN AV 1421685777.065 1426869774.0500011 s4  74 ACH EHZ AV 1421685777.065 1
[...]
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,8 @@
 releases:
- - fixing python3 compatibility for seedlink.easyseedlink
+ - obspy.earthworm
+   * Fix Python3 compatibility problem
+ - obspy.seedlink:
+   * Fix Python3 compatibility for seedlink.easyseedlink
 
 0.10.1: (doi: 10.5281/zenodo.16248)
  - minor changes for correct distribution of official release
--- a/obspy/earthworm/client.py
+++ b/obspy/earthworm/client.py
@@ -79,9 +79,9 @@ class Client(object):
         >>> from obspy.earthworm import Client
         >>> client = Client("pubavo1.wr.usgs.gov", 16022)
         >>> dt = UTCDateTime() - 2000  # now - 2000 seconds
-        >>> st = client.getWaveform('AV', 'ACH', '--', 'EHE', dt, dt + 10)
+        >>> st = client.getWaveform('AV', 'ACH', '', 'EHE', dt, dt + 10)
         >>> st.plot()  # doctest: +SKIP
-        >>> st = client.getWaveform('AV', 'ACH', '--', 'EH*', dt, dt + 10)
+        >>> st = client.getWaveform('AV', 'ACH', '', 'EH*', dt, dt + 10)
         >>> st.plot()  # doctest: +SKIP
 
         .. plot::
@@ -90,9 +90,9 @@ class Client(object):
             from obspy import UTCDateTime
             client = Client("pubavo1.wr.usgs.gov", 16022, timeout=5)
             dt = UTCDateTime() - 2000  # now - 2000 seconds
-            st = client.getWaveform('AV', 'ACH', '--', 'EHE', dt, dt + 10)
+            st = client.getWaveform('AV', 'ACH', '', 'EHE', dt, dt + 10)
             st.plot()
-            st = client.getWaveform('AV', 'ACH', '--', 'EH*', dt, dt + 10)
+            st = client.getWaveform('AV', 'ACH', '', 'EH*', dt, dt + 10)
             st.plot()
         """
         # replace wildcards in last char of channel and fetch all 3 components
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by QuLogic** [comment]
@petrrr Try that one your internal server. I guess the public one is not completely standard, though.

**Comment by mbyt** [comment]
@QuLogic do you mind to give me push access to `obspy/QuLogic`.

Otherwise, I need to go for the complicated procedure of forking the pull request, do some changes and propose a new pull request, which I like to avoid...

**Comment by mbyt** [comment]
@petrrr should work now.

File: CHANGELOG.txt [hunk 1]
```diff
@@ -1,3 +1,7 @@
+development:
```
QuLogic: This is on releases, so it'd be better to mark it 0.10.x just to ease merging (if we don't make any more point releases, that is.)

petrrr:
Thanks a lot for your work. Unfortunately, I cannot confirm definitely this solved the issue we observed with our server until returning to the office.

petrrr:
Okay! I now can confirm that this also now works fine with our server. So it is probably good to merge. BTW: could this make its way into a bugfix release?

megies:
> could this make its way into a bugfix release?

We'll merge it into releases. But it depends on whether we decide it's worth to do a 0.10.2 or head straight for next major release.

megies:
Felt free to rebase and force-push for easy merging..

petrrr:
@megies, as far as I understand this PR is based on release and could be applied directly without rebase. It will need to be rebased to be applied to master with the new directory layout.

petrrr:
Closing this in favor of the rebased PR  #1007.

QuLogic:
I'll merge this when CI passes. The main change is already okay, it's just docs and polish that were added.

obspy/earthworm/client.py
```diff
@@ -77,22 +77,22 @@ def getWaveform(self, network, station, location, channel, starttime,
...
             from obspy import UTCDateTime
-            client = Client("pele.ess.washington.edu", 16017, timeout=5)
-            dt = UTCDateTime(2013, 1, 17) - 2000  # now - 2000 seconds
-            st = client.getWaveform('UW', 'TUCA', '', 'BHZ', dt, dt + 10)
+            client = Client("pubavo1.wr.usgs.gov", 16022, timeout=5)
+            dt = UTCDateTime() - 2000  # now - 2000 seconds
+            st = client.getWaveform('AV', 'ACH', '--', 'EHE', dt, dt + 10)
             st.plot()
-            st = client.getWaveform('UW', 'TUCA', '', 'BH*', dt, dt + 10)
+            st = client.getWaveform('AV', 'ACH', '--', 'EH*', dt, dt + 10)
```
krischer: I did not follow this PR in detail, but why is the `--` needed? All other clients treat (or at least should treat) an empty string as an empty location code. There was also a LONG thread on the FDSN mailing list about this issue a couple months back. Why not just internally map it to dashes?

QuLogic: It looks like `getWaveform` does handle empty strings, but `availability` doesn't (which is why I thought it was necessary.)

krischer: All of them should probably be made to handle empty strings IMHO. I don't mind if they also work with dashes, but at least the tests and docs should use empty strings. Then it is consistent with the other clients.

petrrr:
@krischer wrt the `--`: if I understood correctly how the client is implemented (and we do not want to touch this) it will be necessary. Basically the availability requests the whole "menu". If there is no locID, it will be set to `--`. Then the whole list is matched for the pattern defined by the parameters to the `availabilty()` method.

krischer:
so `--` internally is like a wildcard? That's a bit strange to me as `*` appears to be the wildcard for the channel codes.

Could we just have the following ruleset for the outwards facing `Client.getWaveform()` method:

1. `location=None`: Wildcard; download data with any location code.
2. `location=""`: Explicitly empty location code.
3. `location="--"`: Raise?

QuLogic:
> If there is no locID, it will be set to --.

Doesn't seem to be a wildcard to me. Sounds like it works like an empty location, just like you asked about.

krischer:
So this could just be replaced by an empty string and work the same?

QuLogic:
Seems to work: 9bcfc87. Only thing is the result would not be empty, but `--`.

PR #996, opened 2015-03-24
remove support for python 2.6

mbyt:
This is work in progress

* remove travis 2.6 entry
* py27 brings all these assert* which show more info in the test log. That is replace assertTrue with the assert* method where possible
* gzip.open can be used as context manager
* unittest class now ships `setUpClass`
...


obspy/core/tests/test_trace.py
```diff
@@ -1296,7 +1296,7 @@ def test_taper(self):
         tr.taper(max_percentage=0.05, type='cosine')
         for i in range(len(data)):
             self.assertTrue(tr.data[i] <= 1.)
```
QuLogic: There's `assertLessEqual` that could be applied here, too.

Issue #942, opened 2015-01-25
Add test for already merged #899 (arclink, convert response from type A/Hz to B/Laplace)


megies:
@marcopovitch can you please add a test for #899?

megies:
This really ought to be tested, @marcopovitch..

megies:
stale, closing.

Arclink is kinda obsolete anyways (although local arclink xml files might still be around)

PR #997, opened 2015-03-24
Make sure old imports still work

krischer:
This is essentially a continuation of #842 as it makes sure that packages relying on imports in ObsPy 0.10 still work with the latest master where we reorganized the structure.

With this PR, none of my codes relying on ObsPy require any changes and warnings are raised appropriately.

All of this can be removed as soon as 0.11 has been released.

### DIFF
--- a/obspy/__init__.py
+++ b/obspy/__init__.py
@@ -33,8 +33,6 @@ from future.builtins import *  # NOQA
 from future.utils import PY2, native_str
 
 import imp
-import importlib
-from types import ModuleType
 import warnings
 import sys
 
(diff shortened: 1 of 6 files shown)


krischer:
This fixes some issues with #942 and also makes "attribute access imports" like

```python
import obspy
obspy.xseed.Parser(filename)
```

work. These worked before and I fear they are fairly common so we have to provide a deprecation path.

krischer:
Only Python 2.6 CI fails which is expected and fixed in #996

megies:
@krischer `obspy.core.util` used to serve routines from (former submodule) `geodetics`, these imports are not redirected, so the following doesn't work in master:
`from obspy.core.util import gps2DistAzimuth`

krischer:
Good point. I'll add a deprecation path.

krischer:
Done in #999.

PR #999, opened 2015-03-26
The Big Rename

krischer:
As we just restructured the internal layout of ObsPy now seems to be a good time to refactor other things that we postponed too long.

Contents of this PR:

* snake case for all functions and methods across the code base => much more consistent code
* for some function I took some more liberties and renamed them to more descriptive names
* all clients should now have an actually consistent interface modeled after `obspy.clients.fdsn`
* I/O plugins are now private to further discourage their direct use
* Most outwards facing functions should have a deprecation path so existing codes will continue to work.
* `obspy.signal` used to import essentially everything into its top-level `__init__.py`; I removed most of these imports. I am aware this might break some existing codes - I already added some deprecations paths but we might have to add more.

As I had to perform most of these operations manually I might have missed some.


obspy/core/util/base.py
```diff
@@ -246,27 +247,27 @@ def _getOrderedEntryPoints(group, subgroup=None, order_list=[]):
 
 
 ENTRY_POINTS = {
```
QuLogic: Changing entry point names might be a bit troublesome for existing plugins.

obspy/clients/earthworm/waveserver.py
```diff
@@ -51,30 +51,30 @@ def getNumpyType(tpstr):
     return tp
 
 
-class tracebuf2:
+class TraceBuf2:
```
QuLogic: Might as well make it derive from `object`, too.

obspy/clients/seishub/client.py
```diff
@@ -654,12 +700,17 @@ def getList(self, network=None, station=None, **kwargs):
...
             if key not in ["self", "kwargs"]:
                 kwargs[key] = value
-        url = '/seismology/station/getList'
+        url = '/seismology/station/get_list'
         root = self.client._objectify(url, **kwargs)
         return [dict(((k, v.pyval) for k, v in node.__dict__.items()))
                 for node in root.getchildren()]
 
-    def getCoordinates(self, network, station, datetime, location=''):
+    @deprecated("'get_coordinates' has been renamed to 'get_coordinates'. Use "
```
QuLogic: I think you've got a small typo here...

obspy/clients/seishub/client.py
```diff
@@ -797,14 +853,19 @@ class _EventMapperClient(_BaseRESTClient):
...
 
-    def getList(self, limit=50, offset=None, localisation_method=None,
-                author=None, min_datetime=None, max_datetime=None,
-                first_pick=None, last_pick=None, min_latitude=None,
-                max_latitude=None, min_longitude=None, max_longitude=None,
-                min_magnitude=None, max_magnitude=None, min_depth=None,
-                max_depth=None, used_p=None, min_used_p=None, max_used_p=None,
-                used_s=None, min_used_s=None, max_used_s=None,
-                document_id=None, **kwargs):
+    @deprecated("'get_list' has been renamed to 'get_list'. Use "
```
QuLogic: And another one here.

QuLogic:
I don't think you should have renamed `skipIf`. It's part of `unittest`, and the local copy should be removed now that we don't need to support Python 2.6.

obspy/core/util/base.py
krischer: Good point. I'll revert that. Now that you mentioned it my plugins are not working anymore...

krischer:
> Are we not removing neries entirely?

I also remember something along this line. Are we? Then I'll remove it completely.


> I don't think you should have renamed skipIf. It's part of unittest, and the local copy should be removed now that we don't need to support Python 2.6.

Also good point. I was kind of runnig on autopilot with all the renames ;-) I'll change it to the `unittest` one.

krischer:
Any opinions on this? This is a fairly big change that potentially hurts backwards compatibility so I don't just wanna go and merge it.

QuLogic:
I haven't totally looked at the repercussions, but I agree with the sentiment.

megies:
:+1:  for unifying code formatting (finally)

It feels like we're stacking massive changes, maybe we should really think about 1.0 to make the massive changes clear for all..

megies:
Oh, and we should add a code formatting checker (for function names at least)?

krischer:
We could add `pylint` to the code formatting checks but it is super pedantic (although most things do have their merit). This is essentially what *landscape.io* is doing (amongst other things). Does anyone know when it runs, and when not? To me it just seems to randomly run checks sometimes...

krischer:
Regarding the version number: I think we should release 0.11 fairly soon (in a couple months) which still has all the deprecations. And then we'll do a 1.0 and on purpose break backwards compatibility where we deem necessary. Or should the next version just be 1.0 and we remove the deprecations right away?

mbyt:
Why not going for a 1.0 and ship the deprecations? 1.0 means interface changed, that's what we did. For me an interface change does not say anything about an additional smooth conversion path layer... I fully agree to avoid breaking code when easily possible --- and we have all these nice deprecations in place...

krischer:
No objections thus I'll merge this and we fix any issues in the master.

megies:
> obspy.signal used to import essentially everything into its top-level __init__.py; I removed most of these imports. I am aware this might break some existing codes - I already added some deprecations paths but we might have to add more.

it even breaks code in our docstrings...

```
/home/megies/git/obspy/obspy/core/trace.py:docstring of obspy.core.trace.Trace.simulate:76: WARNING: Exception occurred in plotting obspy-core-trace-Trace-simulate-1
 from /home/megies/git/obspy/misc/docs/source/packages/autogen/obspy.core.trace.Trace.simulate.rst:
  File "/home/megies/anaconda/lib/python2.7/site-packages/matplotlib/sphinxext/plot_directive.py", line 506, in run_code
    six.exec_(code, ns)
  File "/home/megies/anaconda/lib/python2.7/site-packages/six.py", line 624, in exec_
    exec("""exec _code_ in _globs_, _locs_""")
  File "<string>", line 2, in <module>
ImportError: cannot import name corn_freq_2_paz
```

megies:
Just noticed.. #999.. 1.5 x evil. Fits.

ThomasLecocq:
:+1:

PR #1001, opened 2015-03-26
Handle non-UTF8 IO a little better.

QuLogic:
Instead of printing out a bytes object, just replace the problematic character. It won't look as fancy, but at least you won't see strange escaped characters and quotes around it.

### DIFF
--- a/obspy/imaging/scripts/mopad.py
+++ b/obspy/imaging/scripts/mopad.py
@@ -4091,7 +4091,7 @@ def main(argv=None):
                 try:
                     print(decomp)
                 except:
-                    print(decomp.encode('utf-8'))
+                    print(decomp.replace('°', ' deg'))
                 return
             else:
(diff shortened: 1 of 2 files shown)


mbyt:
see also gitter discussion. An other possibility is to avoid the UTF-8 degree symbol completely...  Anyway it's certainly clearer now. Merging.

megies:
I'm a bit lost regarding the mopad/encoding related changes.. @mbyt @QuLogic can you check whether these are covered correctly in the changelog? One change was done before 0.10.1 I think? But this is clearly afterwards..?

QuLogic:
Yes, one change was done before 0.10.1. The change here is that if the console/terminal cannot print `°`, then it prints ` deg` instead of printing the `bytes` representation of everything.

megies:
Well, probably doesnt need a changelog entry by itself and we're good, I guess?

PR #1003, opened 2015-03-27
 remove support for python 2.6, part 2

mbyt:
Follow up of #996.

* Continue replacing assertTrue through assertIs, assertEqual, assertSmallerEqual solely for nicer / more informative error messages
* introducing dictionary comprehensions
* introducing set comprehensions
* using multiple contextmanager in one statement


obspy/core/tests/test_stream.py
```diff
@@ -1736,11 +1736,11 @@ def test_copy(self):
         """
         st = read()
         st2 = st.copy()
-        self.assertTrue(st == st2)
-        self.assertTrue(st2 == st)
+        self.assertEqual(st, st2)
+        self.assertEqual(st2, st)
         self.assertFalse(st is st2)
```
QuLogic: Can use `assertIsNot` for this one.

obspy/core/tests/test_stream.py
```diff
@@ -1932,10 +1932,10 @@ def test_issue540(self):
                 endtime=st[0].stats.endtime + 0.01, pad=True, fill_value=None)
         self.assertEqual(len(st[0]), 3002)
         self.assertTrue(isinstance(st[0].data, np.ma.masked_array))
-        self.assertTrue(st[0].data[0] is np.ma.masked)
+        self.assertIs(st[0].data[0], np.ma.masked)
         self.assertTrue(st[0].data[1] is not np.ma.masked)
```
QuLogic: And here `assertIsNot`, too.

QuLogic:
Note completely sure about changing `UTCDateTime` tests, since they're specifically about checking the individual comparison operators.

obspy/core/tests/test_stream.py
mbyt: And I didn't even look at the assertFalse yet.

These are very, very far from covering all assertTrue and all assertFalse. However, they should be a step in the correct direction. Idea is to get more informative log messages, such that we ideally do not even need to rerun the test or debug for seeing what's happening.

mbyt:
My understanding is that all these individual comparison operators are internally used by the unittest. The only thing that is changed that we get more informative error logs.

QuLogic:
So was additional stuff to do here (besides making it mergeable again)?

mbyt:
There are still tons of `self.assertTrue` which we could replace. However, I think it is a good enough for a first shot.

PR #1007, opened 2015-03-31
Fix earthworm clien behavior for Py3.

petrrr:
This new pull request is a follow up of PR #995


QuLogic:
I think you misread what @megies said in the other PR.

PR #1009, opened 2015-04-01
Use subprocess.check_output instead of Popen.

QuLogic:
Less fiddling with stdout and stderr that way.


megies:
Nice, didn't know / never used this `check_output` routine.. ah, new in 2.7, maybe that's why.

megies:
Hmm, fails on appveyor though..

QuLogic:
It seems that that failure always existed. It was just masked by the overly broad `except:` clause.

megies:
> It seems that that failure always existed. It was just masked by the overly broad except: clause.

OK, even better then..

PR #1018, opened 2015-04-16
replace invalid characters when getting a QuakeML URI from a Resource identifier

megies:
Currently has one test fail that is checking for an error when trying to get a QuakeML URI from a Resource Identifier with spaces in it.


krischer:
What about checking for invalid characters at the time the resource id attribute is set? Then the error is caught earlier and less magic is involved. I cannot think of a good reason to allow invalid characters in a resource identifier. Is there one?

obspy/core/event.py
```diff
@@ -756,10 +756,17 @@ def get_quakeml_uri(self, authority_id="local"):
...
         regex = r"^(smi|quakeml):[\w\d][\w\d\-\.\*\(\)_~']{2,}/[\w\d\-\." + \
                 r"\*\(\)_~'][\w\d\-\.\*\(\)\+\?_~'=,;#/&amp;]*$"
+        # if current string is a valid QuakeML resource ID just use it
         result = re.match(regex, str(id))
         if result is not None:
             return id
-        id = 'smi:%s/%s' % (authority_id, str(id))
+        # if not, we replace invalid characters and prepend a valid prefix
+        id_string, replace_count = \
+            re.subn(r"[^\w\d\-\.\*\(\)\+\?_~'=,;#/&amp;]", '_', str(id))
```
QuLogic: So I realize this was copied out of the spec, but I don't think `&amp;` was intended here or above.

megies:
> What about checking for invalid characters at the time the resource id attribute is set? Then the error is caught earlier and less magic is involved. I cannot think of a good reason to allow invalid characters in a resource identifier. Is there one?

Actually that's what I proposed almost two years ago.. 0_o  #559

megies:
This is now most rigorous, only allowing valid QuakeML URI whenever initializing a ResourceIdentifier. But I fear this might cause problems.. still not sure how to best approach this one..

krischer:
> But I fear this might cause problems.. still not sure how to best approach this one..

It seems so. 13 test failures from other event readers that set a resource id... Should we just fix that in the other modules? I still think its a good idea to force a correct id at init time.

> Actually that's what I proposed almost two years ago.. 0_o #559

So I assume you are also in favor of doing it ;-)

I can do it if everyone is okay with that. I think it will have only very minimal repercussions in existing scripts.

QuLogic:
> 13 test failures from other event readers that set a resource id... Should we just fix that in the other modules?

Probably a good idea to that either way.

megies:
> I can do it if everyone is okay with that. I think it will have only very minimal repercussions in existing scripts.

Go ahead, that's the only way to get an idea of how drastic a change this would be if we really do it..

megies:

megies:
This PR needs some love.. or a review.. or both..

`event.py` was refactored into `event/*.py` in the meantime..

d-chambers:
At the risk of sounding like a broken record, I think we should push this to 2.0. It has the potential introduce some breaking changes so is probably better suited for a major release no?

krischer:
> At the risk of sounding like a broken record, I think we should push this to 2.0. It has the potential introduce some breaking changes so is probably better suited for a major release no?

I agree and I also just changed the milestone. This might very well break some things for people who relied on the current behavior.

d-chambers:
Hey @megies,

This is tagged for 2.0. I would prefer not to do too much on it until we have removed the future and compatibility stuff to avoid too many merge conflicts. Any reason to work on this before then?

megies:
Oh, OK. Nah, fine with me.

PR #1025, opened 2015-04-21
Add possibility to read files with `tbz` suffix

petrrr:
Minor change to be able to read files with `tbz` suffix. This is handled the same way as all tarfiles and ending `tar.bz2` was already supported.  Also *simplified* the condition (not sure but I find this more readable). Maybe the constant tuple could/should be moved elsewhere.

### DIFF
--- a/obspy/core/util/decorator.py
+++ b/obspy/core/util/decorator.py
@@ -119,8 +119,8 @@ def uncompress_file(func):
             raise IOError(msg)
         # check if we got a compressed file or archive
         obj_list = []
-        if filename.endswith('.tar') or filename.endswith('.tgz') or \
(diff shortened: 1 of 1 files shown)


petrrr:
Commit 9f9407b intends to simplify the handling of tar and zipfiles. This assumes that we accept that files may be misnamed. 

If we prefer to enforce that names correspond to file types the original implementation needs to be fixed, as the check will be ignored anyway.

QuLogic:
It looks good to me. I don't even know why those imports needed to be in the `try`.

megies:
Could justify a short changelog entry, probably.. not too important though, I guess..?

PR #1012, opened 2015-04-25
Python 2 Unicode error with NEIC client

QuLogic:
Reported on [the mailing list](http://lists.swapbytes.de/archives/obspy-users/2015-April/001678.html):
```python
from obspy import UTCDateTime
from obspy.neic import Client
client = Client()
indivSta = client.getWaveform('HV', 'NPT', '', 'HHN',
                              UTCDateTime(2015, 3, 4, 22, 0),
                              UTCDateTime(2015, 3, 5, 2, 0))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/obspy/neic/client.py",
line 121, in getWaveform
    seedname = network.ljust(2, " ") + station.ljust(5, " ") + channel + \
TypeError: must be char, not unicode
```

### DIFF
--- a/obspy/neic/client.py
+++ b/obspy/neic/client.py
@@ -118,8 +118,7 @@ class Client(object):
             msg = "channel expression matches less than 3 characters " + \
                   "(use e.g. 'BHZ', 'BH?', 'BH[Z12]', 'B??')"
             raise Exception(msg)
-        seedname = network.ljust(2, " ") + station.ljust(5, " ") + channel + \
-            location.ljust(2, " ")
+        seedname = '%-2s%-5s%s%-2s' % (network, station, channel, location)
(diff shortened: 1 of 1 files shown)


QuLogic:
> Running all tests without the __future__ imports would have been the better way to port.

Just tried without `unicode_literals` (the only one that's relevant, I'd say), and the only failures are strings that are genuinely Unicode (i.e., input to `io.StringIO`). So we wouldn't actually trigger much else by doing this unless we added new tests first.

krischer:
@QuLogic  Thanks for testing. Then I guess we don't have to do it.

This is ready to merge otherwise, right?

QuLogic:
Yep.

PR #1029, opened 2015-04-25
Velocity reduction with degree offsets not handled correctly

QuLogic:
The documentation says `vred` is "Perform velocity reduction, in m/s.", and [the code](https://github.com/obspy/obspy/blob/master/obspy/imaging/waveform.py#L1211) uses a simple `offset / vred`. However, the units for `offset` changed if you use `dist_degree=True`.

I put in a simple `vred = kilometer2degrees(vred / 1e3)` to make it deg/s. Not sure if that makes sense, or whether we should use one of the more accurate GC algorithms instead.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,8 @@ releases:
    * Fix Python3 compatibility problem
  - obspy.imaging:
    * Fix flipped maps due to bug in Basemap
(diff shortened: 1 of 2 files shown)


QuLogic:
So, let me know if you think this okay, but **don't merge**. I'll open another PR with a few other section plot improvements.

PR #1030, opened 2015-04-25
Section plot fixes

QuLogic:
This supersedes #1029 to reduce the branching. These are all related to section plots.
 * Correct units for velocity reduction when using degree offsets (see #1029 for full explanation).
 * Allow overlaying multiple section plots by not modifying lines from previous plots.


krischer:
The accuracy of the approximation for the velocity reduction when using degree offsets should be ok. I hope noone makes any measurements on these plots ;-)

PR #1031, opened 2015-04-27
Adjust minium required versions for NumPy and SciPy

krischer:
With Python 2.6 support being history I think we can reasonably up the minimum version numbers for NumPy and SciPy.

I chose the ones shipping with Ubuntu 12.04 LTS which appears to be the earliest Python 2.7 shipping distro we support. We don't support CentOS 6 and Debian 6 anymore (as they depend on Python 2.6).

The new minimum version numbers are NumPy >= 1.6.1 and SciPy >= 0.9.0.


krischer:
Any comment on this?

CHANGELOG.txt
```diff
@@ -3,6 +3,8 @@ master:
    * Requirements have been increased to reflect latest distributions:
      * Removed support for Python 2.6.
      * matplotlib >= 1.1.0 is now required.
+	 * numpy >= 1.6.1 is now required.
+	 * scipy >= 0.9.0 is now required.
```
QuLogic: I guess this is a tab, but should be spaces...

QuLogic:
I don't think I've got anything against it; maybe should ask other packagers, though...

megies:
No problem for deb packages..

mbyt:
Based on the obspy 0.10 release notes, we even announced https://github.com/krischer/obspy/blob/adjust_min_versions/CHANGELOG.txt#L52 scipy >= 0.10.

As far as I remember, you checked the docker files back then, don't remember on which distro we based that. Might be debian wheezy.

Ubuntu 12.04 is really old, the new LTS 14.04 is now nearly one year available. I don't think we need to support that. People could still use conda on that boxes.

Whatever, I do not have any problem supporting scipy >= 0.9. Scipy 0.10 does not ship any live safer from my point of view.

krischer:
Sorry guys...I screwed up the rebasing by wanting to move too fast...so I had to force push to fix things. Should be good now. I was too lazy to retain the previous commits.

megies:
No problem.

PR #1026, opened 2015-04-27
Polarization analysis doesn't work

jwassermann:
Claudia Roeoesli contacted me because of problems with the new polarization module.

After polishing the code it seems that the polarization_analysis function is no longer working properly. Looking into the code, I've seen that the dictionary returned by the function only give back the last (npts) value rather than a vector. Also the time stamp is now missing. Should be easy to fix?!

Cheers Jo

### DIFF
--- a/obspy/signal/polarization.py
+++ b/obspy/signal/polarization.py
@@ -458,9 +458,16 @@ def polarization_analysis(stream, win_len, win_frac, frqlow, frqhigh, stime,
     :param method: the method to use. one of ``"pm"``, ``"flinn"`` or
         ``"vidale"``.
     :type method: str
-    :returns: Dictionary with posix timestamp, azimuth, incidence angle,
-        errors, rectilinearity, planarity, and/or ellipticity (the returned
-        values depend on the used method).
+    :rtype: dict
+    :returns: Dictionary with keys ``"timestamp"`` (POSIX timestamp, can be
+        used to initialize :class:`~obspy.core.utcdatetime.UTCDateTime`
+        objects), ``"azimuth"``, ``"incidence"`` (incidence angle) and
+        additional keys depending on used method: ``"azimuth_error"`` and
+        ``"incidence_error"`` (for method ``"pm"``), ``"rectilinearity"`` and
+        ``"planarity"`` (for methods ``"flinn"`` and ``"vidale"``) and
+        ``"ellipticity"`` (for method ``"flinn"``). Under each key a
+        :class:`~numpy.ndarray` is stored, giving the respective values
+        corresponding to the ``"timestamp"`` :class:`~numpy.ndarray`.
     """
     if method.lower() not in ["pm", "flinn", "vidale"]:
         msg = "Invalid method ('%s')" % method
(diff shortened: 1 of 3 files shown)


megies:
The tests had to be adapted because output is now arrays not scalars. Please double check @jwassermann, @krischer.

krischer:
Looks good to me as far as i can tell.

@jwassermann Can we merge this?

obspy/signal/polarization.py
```diff
@@ -359,12 +359,31 @@ def fit_func(beta, x):
...
+    az_error = 1.0 / ((1.0 ** 2 + az_slope ** 2) * azimuth) * az_error
+    # az_error = math.degrees(az_error)
+    in_error = 1.0 / ((1.0 ** 2 + in_slope ** 2) * incidence) * in_error
+    # in_error = math.degrees(in_error)
+
+    azimuth = math.degrees(azimuth)
+    incidence = math.degrees(incidence)
+
+    if azimuth < 0.0:
+        azimuth = 360.0 + azimuth
```
QuLogic: The only one that's not `+=`...

jwassermann:
Tobi and I fixed teh module and the corresonding tests - we will merge
it today...

Jo

On 06/05/15 19:17, Lion Krischer wrote:
>
> Looks good to me as far as i can tell.
>
> @jwassermann <https://github.com/jwassermann> Can we merge this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/obspy/obspy/pull/1026#issuecomment-99542097>.
>

-- 
-----------------------------------------------------------------
   Dr. Joachim Wassermann

   Geophysikalisches Observatorium   
   Ludwigshoehe 8                   
   82256 Fuerstenfeldbruck           
                                    
   Tel.: +49 89 2180-739-62          
   FAX.: +49 89 2180-739-70          

 email: j.wassermann@lmu.de
 www: www.erdbeben-in-bayern.de | www.geophysik.uni-muenchen.de
..................................................................

   Department fuer Geo- & Umweltwissenschaften
   Sektion Geophysik
   Theresienstrasse 41/IV
   80333 Muenchen

  Tel.: +49 089 2180 4237
  FAX.: +49 089 2180 4205
------------------------------------------------------------------

megies:
Right now I'm thinking that it is probably a good idea to change how the timing info is coded in the result by replacing the
```python
{"timestamp": np.ndarray([.....]),
...}
```
part by something like
```python
{"starttime": UTCDateTime(..),
 "delta": 0.05,
 ...}
```
That would save some memory and also would avoid possible rounding errors in very long timestamp arrays.

Actually.. an even more progressive approach would be to replace the `dict` result with one array of values for each "characteristic" of the result by a `Stream` object with several traces, specifying the result information type e.g. in `stats.channel`.

E.g. have a result stream like..
```
X Trace(s) in Stream:
BW.RJOB..azimuth     | 2009-08-24T00:20:03.000000Z - 2009-08-24T00:20:32.990000Z | 100.0 Hz, 3000 samples
BW.RJOB..incidence   | 2009-08-24T00:20:03.000000Z - 2009-08-24T00:20:32.990000Z | 100.0 Hz, 3000 samples
BW.RJOB..planarity   | 2009-08-24T00:20:03.000000Z - 2009-08-24T00:20:32.990000Z | 100.0 Hz, 3000 samples
BW.RJOB..ellipticity | 2009-08-24T00:20:03.000000Z - 2009-08-24T00:20:32.990000Z | 100.0 Hz, 3000 samples
...
```

But of course, both options would not go well with merging into releases, I guess, as it's changing the interface a lot. On the other hand this PR bug fix is already changing what comes out of the routine anyway..

obspy/signal/polarization.py
megies: Yeah, probably some not-so-nice looking code in here but this PR is more about the essential fix of the functionality. I would say we leave cosmetics to `master`.

krischer:
> On the other hand this PR bug fix is already changing what comes out of the routine anyway..

True but it actually fixes a serious bug so this should still go into releases I think

I also think its a good idea to have some more descriptive output objects but they should probably not be stream/trace objects as many operations just don't make sense and some functionality will randomly break with super long channels code.

Maybe we should finally implement a `BaseTrace` class...we could call it `TimeSeriesTrace`. It will then just be a basic container for arbitrary time series data with basic method like `filter()`, `detrend()`, ... that always make sense.

Normal `Trace` objects would inherit from it, as would the return objects of the polarization analysis. Naturally a topic for another PR.

megies:
> Normal Trace objects would inherit from it, as would the return objects of the polarization analysis. Naturally a topic for another PR.

Yeah, I totally agree. So let's just merge this basic bug fix..

PR #1033, opened 2015-04-27
Backport sosfilt from scipy 0.16.0.

### LINKED ISSUE #1028: A question about butterworth-bandpass filtering
Hi,

I just found a possible question about butterworth-bandpass filtering. I download a seismogram from IRIS and try to filter it with a butterworth-bandpass filter from 0.02hz to 0.2hz, but found the result is incorrect with some extremely large number after filtering. I guess it might be caused by some kind of unstability of filtering algorithm. Here is my code:

```python
from obspy.fdsn import Client
from obspy import UTCDateTime
from obspy.taup.taup import getTravelTimes
from obspy.core import AttribDict
client = Client('iris')
starttime = UTCDateTime('2014-08-24T10:20:44.0600')
p_arrival_time = 376.945875099  # p phase arrival time
st = client.get_waveforms(network='AK', station='BPAW', location='*', channel='BHZ', 
                          starttime=starttime+p_arrival_time-60, endtime=starttime+p_arrival_time+180)
st.remove_response(output='disp')  # remove the instrument response
st.detrend(type='demean')  # remove the mean
st.detrend(type='linear')  # remove the linear trend
st.taper(type='hann', max_percentage=0.05)  # taper
# st.resample(20)  # The decreasing sampling rate will make filtering work well
st.filter('bandpass',freqmin=0.02, freqmax=0.2, zerophase=True)  # filter
st.write('tmp.sac',format='sac')
```

BTW, when I decrease the sampling rate from original 50hz to 20hz, the filtering will be good. I wonder whether it might also be related to sampling rate.

Thx

QuLogic:
The second-order sections form for filters are much more stable than the
form using transfer functions.

Fixes #1028, but on master.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -15,6 +15,9 @@ master:
  - obspy.imaging:
    * Experimental support for Cartopy when plotting maps. Use the `method`
      argument to functions that plot maps to select between Basemap or Cartopy.
+ - obspy.signal:
+   * Switch to second-order sections for filters; backported from SciPy 0.16.0
+     (see #1028)
 
 releases:
  - obspy.core:
--- a/obspy/signal/_sosfilt.py
+++ b/obspy/signal/_sosfilt.py
@@ -1,0 +1,509 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------
+# Filename: _sosfilt.py
+#  Purpose: Backport of Second-Order Section Filtering from SciPy 0.16.0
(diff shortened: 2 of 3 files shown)


krischer:
Great. I hope our test suite is enough to guard against regressions but I think so.

mbyt:
Great work. Does it make sense to add a test, e.g. one of the example which improved https://github.com/obspy/obspy/pull/1028#issuecomment-96564165? Or is this covered elsewhere?

QuLogic:
Yes, that's probably a good idea. Especially since I never actually tried it out with SciPy 0.16.0... :stuck_out_tongue:

PR #1034, opened 2015-04-27
Section plot additions: orientation & colour

QuLogic:
Two additions to the section plots.
 * Add an orientation setting so that you can make the time axis horizontal.
 * Add an alternative colour setting (network/station/channel) so that the lines can be multi-coloured by that setting.

I'm not sure that overloading the `color` argument this way is the best method. For one, it means you can't specify the colours of the groups. Maybe a second argument that enabled colour-by-group would be a better option?


QuLogic:
I'm starting to think perhaps a `color_by='network'` would make more sense...

Travis failure must be me forgetting to update the plot when I rebased to master...

QuLogic:
Oh, I might have had an even better idea. Instead of a bunch of single options like this, maybe something more Pandas/seaborn-esque. Instead of overloading `color` and the like, how about two arguments `group_by` and `label_by`, which can (possibly) apply to all plot types.

For regular plots, grouping would be something like placing all channels of a station on the same subplot. For section plots, labelling would allow you to overlay multiple channels. Then color/linestyle/alpha can continue to work as they already do (or perhaps expanded a little, but still with the same meaning at least). I haven't thought through all the application, but I think these two arguments might be a bit more generic while still allowing for obvious meanings that apply to each type of plot.

megies:
I never really used the section plot, so I can't really say what's best for it. Change as you think is best, I guess.

Having a `group_by` kwarg for the regular preview plot sounds like something that could be useful.

krischer:
I really like the `group_by` and `label_by` idea! If you actually decide to do it, please feel free to refactor the monster `WaveformPlotting` object. I feel like the logic for plotting normal waveforms is pretty sound by now but the way we handle all these parameters and the different places they are documented is rather wild...

Some other structure, maybe multiple hierarchical classes or with mixins or even completely without classes would be much better and would make it easier to extend over time.

QuLogic:
Not sure why the image is failing; maybe a matplotlib 1.3.x thing? Also, why aren't the images uploading to imgur?

megies:
Images are uploaded alright but with the changes in 5c477df19ba1e423bb no information on the uploads is shown anymore if `OBSPY_KEEP_IMAGES` is not set as well.
Should be fixed with 178272ddc1f238.

I felt free to rebase and force push.

It seems the legend is missing in the plot on mpl 1.3.1.

megies:
Oh, actually pyimgur upload can't work because this branch is on your fork and the secure imgur client id env variable is not decoded on PRs coming from outside the main repo..

QuLogic:
Ah, didn't notice you fixed that! OK, looks good to me then. I'll try out the `group_by` and `label_by` stuff in a separate PR.

megies:
> OK, looks good to me then.

I didn't look at the details, only fixed the test fail, but if it looks good to you it's good to me. ;)

PR #1032, opened 2015-04-27
SacIO() fails with python 2.7

QuLogic:
When using python 2.7, the following code fails:
```python
from obspy.sac import SacIO
with open('test.sac') as fh:
    sac = SacIO(fh)
```
with the following error:
```python
TypeError: unsupported operand type(s) for -: 'NoneType' and 'long'
```

This is related to [this line of code](https://github.com/obspy/obspy/blob/releases/obspy/sac/sacio.py#L470), since `file.seek()` does not return anything in python 2.7, and therefore `length` is `None`.

Tests pass on python 2.7 probably because of: 
```python
from future.builtins import *
```

Following [this answer on Stack Overflow](http://stackoverflow.com/questions/2104080/how-to-check-file-size-in-python/19079887#19079887), I suggest (sorry, I have no time to make a PR):
```python
cur_pos = fh.tell()
fh.seek(0, os.SEEK_END)
size = fh.tell()
fh.seek(cur_pos, os.SEEK_SET)
```
(note also the use of `os.SEEK_END` and `os.SEEK_SET` for better code readability).

### DIFF
--- a/obspy/sac/sacio.py
+++ b/obspy/sac/sacio.py
@@ -25,6 +25,7 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 from future.utils import native_str
 
+import os
 import time
 import warnings
 
@@ -467,8 +468,9 @@ class SacIO(object):
         Test for a valid SAC file using arrays.
         """
         cur_pos = fh.tell()
-        length = fh.seek(0, 2)
-        fh.seek(cur_pos, 0)
+        fh.seek(0, os.SEEK_END)
+        length = fh.tell()
+        fh.seek(cur_pos, os.SEEK_SET)
         try:
             npts = self.GetHvalue('npts')
         except:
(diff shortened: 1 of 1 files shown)


claudiodsf:
Thanks for making this a PR!

With this modification, there is no more need to have:
```python
from future.builtins import *
```
in `test_sacio.py`

krischer:
The

```python
from future.builtins import *
```

line kind of forces Python 2 to behave more like Python 3 so we chose to have it everywhere.

Please note that we discourage the direct use of functions from the various I/O modules. In the next version these imports might not work anymore.

```python
from obspy.sac import SacIO
```

Out of curiosity: why don't you use the `Stream` and `Trace` objects?

claudiodsf:
@krischer So that I can find bugs like this :stuck_out_tongue_winking_eye: 

Seriously, it's for historical reasons. Also, I found it easier to use for setting time-related header fields (like `A` or `T0`), which are relative to reference time and not to trace beginning time (these two times are different if `B != 0`).

But I guess I will try to switch to `Trace`.

megies:
Probably no need to add a changelog for this, I guess..?

krischer:
No...the error is hidden in case one uses the normal `read()` function in any case I think.

krischer:
Or maybe add a changelog to justify 0.10.2 once we release?

PR #1041, opened 2015-05-05
client.fdsn: add shortcut for new EIDA node KOERI

petrrr:
Minor edit of the URL map to add a new EIDA node. Formally this is a new feature, so I branched from master, but I'd propose that it is ported back to releases as well.


QuLogic:
Sounds good.

megies:
+1 should definitively be in releases as well.

PR #1046, opened 2015-05-13
BugFix for Datamark

ThomasLecocq:
- error in reading datawide=4 files (' ' instead of b' ')
- error in reading datawide=0.5

both work, still need a test for datawide = 0.5


megies:
Duplicate of #1016.

PR #1047, opened 2015-05-13
station/stationxml: move DataAvailability from Channel to BaseNode

megies:
According to the [FDSN xml schema (the extension for data availability)](https://github.com/obspy/obspy/blob/master/obspy/io/stationxml/data/fdsn-station%2Bavailability-1.0.xsd) data availability is defined for all of network/station/channel and not only for channel. It thus should be moved to the parent class `BaseNode`.


QuLogic:
Looks good to me.

PR #1050, opened 2015-05-19
toying around with making obspy (base)map plots reusable

megies:
This is just to show that in principle we should be able to make ObsPy's basemap (preview) plots of events/stations reusable with very little changes (so far), inspired by this [ipython notebook](http://nbviewer.ipython.org/gist/QuLogic/fa595296657f585f12bc).

Most problematic would probably eventually get the correct handling/updating of colorbar (or probably would eventually get impossible, esp. with consecutive event plots).

Anyway, the following example works:
```python
plt.rcParams['figure.figsize'] = 12, 8
from obspy import UTCDateTime
from obspy.fdsn import Client
client = Client('IRIS')
minlat = 35
maxlat = 37.5

minlon = -100
maxlon = -95

starttime = UTCDateTime('2012-01-01T00:00:00')
endtime = UTCDateTime('2015-02-01T00:00:00')

minmag = 0.0
maxmag = 4.0

ok_ev = client.get_events(minlatitude=minlat, maxlatitude=maxlat,
                          minlongitude=minlon, maxlongitude=maxlon,
                          minmagnitude=minmag, maxmagnitude=maxmag,
                          starttime=starttime, endtime=endtime)

ok_sta = client.get_stations(network='TA', channel='BH?',
                             minlatitude=minlat, maxlatitude=maxlat,
                             minlongitude=minlon, maxlongitude=maxlon)

title = 'M<%d -- %s to %s -- %d total events' % (
    maxmag,
    starttime,
    endtime,
    len(ok_ev))

fig, bmap = ok_ev.plot(projection='local', show=False)
ok_sta.plot(bmap=bmap)
```

figure_1


krischer:
:+1: I would pursue this. Simple change but pretty useful. It does break an existing interface a bit but I don't think it matter much in this case.

QuLogic:
If you don't want to change the API, add a `return_bmap` keyword arg and only return it if passed.

megies:
Yes, although I'll probably just attach the `Basemap` object to the `Figure` instance for later reuse and revert the API return value change.

megies:
This now works without the API return value change..
So..
 - document in `.plot()` docstrings of Catalog/... and Inventory/... objects, and..
 - add example snippet to tutorial
.. and this is good?

megies:
```python
from obspy import read_events, read_inventory

catalog = read_events()
inventory = read_inventory()

fig = catalog.plot(show=False)
inventory.plot(fig=fig)
```

figure_1

obspy/imaging/maps.py
```diff
@@ -160,134 +163,142 @@ def plot_basemap(lons, lats, size, color, labels=None, projection='global',
...
-    bmap.fillcontinents(color=continent_fill_color,
-                        lake_color=water_fill_color)
-    # draw the edge of the bmap projection region (the projection limb)
-    bmap.drawmapboundary(fill_color=water_fill_color)
-    # draw lat/lon grid lines every 30 degrees.
-    bmap.drawmeridians(np.arange(-180, 180, 30))
-    bmap.drawparallels(np.arange(-90, 90, 30))
+        map_ax = fig.axes[0]
+        fig = map_ax.get_figure()
+        bmap = fig.bmap
```
QuLogic: Should the case that someone uses a completely new `Figure` without bmap be handled as well?

megies: I'd say probably not, this is a pretty specialized case and the docstring clearly states that a figure from a previous call to the routine is expected. However, todo: I/we need to also clearly state this fact when adding this to the Inventory/Catalog/... plotting routines.

QuLogic: I guess I just mean a more specific error might be appropriate? Instead of an `AttributeError`.

megies: Hmm yeah, why not..

megies: I'm adding some additional text to the error messages now (I prefer not to change the raised exception type):

```
  File "/home/megies/git/obspy/obspy/core/inventory/inventory.py", line 579, in plot
    title=None, show=False, **kwargs)
  File "/home/megies/git/obspy/obspy/imaging/maps.py", line 684, in plot_map
    return plot_basemap(*args, **kwargs)
  File "/home/megies/git/obspy/obspy/imaging/maps.py", line 302, in plot_basemap
    map_ax = fig.axes[0]
IndexError: list index out of range. Please provide a figure object from a previous call to the .plot() method of e.g. an Inventory or Catalog object.


  File "/home/megies/git/obspy/obspy/core/inventory/inventory.py", line 579, in plot
    title=None, show=False, **kwargs)
  File "/home/megies/git/obspy/obspy/imaging/maps.py", line 684, in plot_map
    return plot_basemap(*args, **kwargs)
  File "/home/megies/git/obspy/obspy/imaging/maps.py", line 307, in plot_basemap
    bmap = fig.bmap
AttributeError: 'Figure' object has no attribute 'bmap'. Please provide a figure object from a previous call to the .plot() method of e.g. an Inventory or Catalog object.
```

QuLogic: Yea, sounds good.

QuLogic:
Is it just the example left?

megies:
Added an image test as well, also documentation. Should be ready to merge, I think if the tests pass..

@QuLogic, I think the `plot_cartopy` part could be adapted like in 77871d3ba22d as well, stripping the manually making room for colorbar axes..

megies:
Rebased, to get rid of unrelated test fails..

megies:
Finally have given in.. worst thing they do is probably block it if there's too many uploads in short time, but probably we won't hit that mark..

megies:
Not sure why the axes/colorbar get different size ratios on my machine vs. travis. Same matplotlib version and same basemap version..

QuLogic:
I'm also seeing the different ratio here; is it necessary to make that change?

megies:
Rebased..

megies:
> I'm also seeing the different ratio here; is it necessary to make that change?

It's just that this manual spacing comes too late when reusing a plot without prior colorbar.. it already has the map axes set up. But what I can do is revert to the manual setting in general and only for the combined plot when running into problems (because no space was previously allocated for the colorbar) use matplotlib's option to steal from another axes..

PR #1055, opened 2015-05-19
obpsy.taup seems not to work with surface waves ('kmps')

QuLogic:
~~~~{.python}
model = TauPyModel(model="iasp91")
arrivals = model.get_travel_times(source_depth_in_km=55,
                                  distance_in_degree=67, phase_list=["2kmps"])
~~~~
returns an empty list. For a fixed velocity, I think get_travel_times should always contain a phase.


QuLogic:
OK, this should fix the arrivals, but I'm not entirely sure whether the result makes sense. We don't seem to have any tests for surface waves. @krischer, maybe you can try them out with the Java version to get some tests?

krischer:
Added a test. Works like a charm. Now this is actually a good reason to release `0.10.3` :-)

PR #1061, opened 2015-05-28
obspy.signal.rotate.rotate2ZNE seems buggy

krischer:
I am failing to understand the result of rotate2ZNE and I think the problem is in _dip_azimuth2ZSE_base_vector. If I build the matrix, I observe two problems: 

~~~~{.python}
#!/usr/bin/env python
import numpy as np
from obspy.signal.rotate import _dip_azimuth2ZSE_base_vector

dip_1, dip_2, dip_3, = 35.3, 35.3, 35.3
azimuth_1, azimuth_2, azimuth_3 = 0., 120., 240.

base_vector_1 = _dip_azimuth2ZSE_base_vector(dip_1, azimuth_1)
base_vector_2 = _dip_azimuth2ZSE_base_vector(dip_2, azimuth_2)
base_vector_3 = _dip_azimuth2ZSE_base_vector(dip_3, azimuth_3)

T = np.matrix([base_vector_1, base_vector_2, base_vector_3]).transpose()
print(T)
[[-0.57785762  0.28892881  0.28892881]
 [-0.81613759  0.4080688   0.4080688 ]
 [ 0.          0.8660254  -0.8660254 ]]

~~~~

according to Wieland, this is what the matrix should look like:

wieland

However, for the Z component (first line of T, last line in Wielands notation), the values are not all the same. Secondly, the matrix is not invertible, which it should be because the components are linearly independent;

~~~~{.python}
  File "/home/ex/anaconda/lib/python2.7/site-packages/numpy/linalg/linalg.py", line 90, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
numpy.linalg.linalg.LinAlgError: Singular matrix
~~~~

This is what I would need at the moment, because I want to compute UVW given NEZ.

Also, the tests seem to cover only very specific examples (orthogonal bases).

### DIFF
--- a/obspy/signal/rotate.py
+++ b/obspy/signal/rotate.py
@@ -164,7 +164,7 @@ def _dip_azimuth2ZSE_base_vector(dip, azimuth):
 
 
 def rotate2ZNE(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
-               azimuth_3, dip_3):
+               azimuth_3, dip_3, inverse=False):
     """
     Rotates an arbitrarily oriented three-component vector to ZNE.
 
@@ -186,9 +186,16 @@ def rotate2ZNE(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
     :param data_3: Data component 3.
     :param azimuth_3: The azimuth of component 3.
     :param dip_3: The dip of component 3.
+    :param inverse: If `True`, the data arrays will be converted from ZNE to
+        whatever coordinate system the azimuths and dips specify. In that
+        case data_1, data_2, data_3 have to be data arrays for Z, N,
+        and E and the dips and azimuths specify where to transform to.
+    :type inverse: bool
 
     :rtype: Tuple of three NumPy arrays.
-    :returns: The three rotated components, oriented in Z, N, and E.
+    :returns: The three rotated components, oriented in Z, N, and E if
+        `inverse` is `False`. Otherwise they will be oriented as specified
+        by the dips and azimuths.
 
     >>> # An input of ZNE yields an output of ZNE
     >>> rotate2ZNE(np.arange(3), 0, -90, np.arange(3) * 2, 0, 0, \
@@ -214,19 +221,22 @@ def rotate2ZNE(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
     # Build transformation matrix.
     T = np.matrix([base_vector_1, base_vector_2, base_vector_3]).transpose()
 
-    # Apply it.
-    z, s, e = np.dot(T, [data_1, data_2, data_3])
-    # Replace all negative zeros. These might confuse some further processing
-    # programs.
+    if not inverse:
+        x, y, z = np.dot(T, [data_1, data_2, data_3])
+        y *= -1
+    else:
+        x, y, z = np.dot(np.linalg.inv(T), [data_1, -data_2, data_3])
+
+    # Replace all negative zeros. These might confuse some further
+    # processing programs.
+    x = np.array(x).ravel()
+    x[x == -0.0] = 0
+    y = np.array(y).ravel()
+    y[y == -0.0] = 0
     z = np.array(z).ravel()
     z[z == -0.0] = 0
-    # Return a North pointing array.
-    n = -1.0 * np.array(s).ravel()
-    n[n == -0.0] = 0
-    e = np.array(e).ravel()
-    e[e == -0.0] = 0
 
-    return z, n, e
+    return x, y, z
 
 
 if __name__ == '__main__':
--- a/obspy/signal/rotate.py
+++ b/obspy/signal/rotate.py
@@ -159,31 +159,22 @@ def _dip_azimuth2ZSE_base_vector(dip, azimuth):
     dip = np.deg2rad(dip)
     azimuth = np.deg2rad(azimuth)
 
-    # Define the rotation axis for the dip.
-    c1 = 0.0
-    c2 = 0.0
-    c3 = -1.0
-    # Now the dip rotation matrix.
-    dip_rotation_matrix = np.cos(dip) * \
-        np.matrix(((1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (0.0, 0.0, 1.0))) + \
-        (1 - np.cos(dip)) * np.matrix(((c1 * c1, c1 * c2, c1 * c3),
-                                       (c2 * c1, c2 * c2, c2 * c3),
-                                       (c3 * c1, c3 * c2, c3 * c3))) + \
-        np.sin(dip) * np.matrix(((0, -c3, c2), (c3, 0, -c1), (-c2, c1, 0)))
-    # Do the same for the azimuth.
-    c1 = -1.0
-    c2 = 0.0
-    c3 = 0.0
-    azimuth_rotation_matrix = np.cos(azimuth) * \
-        np.matrix(((1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (0.0, 0.0, 1.0))) + \
-        (1 - np.cos(azimuth)) * np.matrix(((c1 * c1, c1 * c2, c1 * c3),
-                                           (c2 * c1, c2 * c2, c2 * c3),
-                                           (c3 * c1, c3 * c2, c3 * c3))) + \
-        np.sin(azimuth) * np.matrix(((0, -c3, c2), (c3, 0, -c1), (-c2, c1, 0)))
+    # We start with a northwards pointing vector. Thus the dip is just a
+    # rotation around the east axis with the negative angle.
+    dip_rot = np.matrix((
+        (np.cos(-dip), -np.sin(-dip), 0.0),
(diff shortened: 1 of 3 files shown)


martinvandriel:
here is a much easier implemtation to get the basis vectors and it matches the wieland solutions:

~~~~{.python}
#!/usr/bin/env python
import numpy as np


def _dip_azimuth2ENZ_base_vector(dip, azimuth):
    """
    Helper function to compute a unit vector described by azimuth and dip
    ENZ (East, North, Up) basis.
    """
    delta = np.deg2rad(dip)
    phi = np.deg2rad(azimuth)

    e = np.sin(phi) * np.cos(delta)
    n = np.cos(phi) * np.cos(delta)
    z = -np.sin(delta)

    return np.array([e, n, z])

dip = - (90. - np.rad2deg(np.arctan(2 ** 0.5)))

v1 = _dip_azimuth2ENZ_base_vector(dip, -90.)
v2 = _dip_azimuth2ENZ_base_vector(dip, 30.)
v3 = _dip_azimuth2ENZ_base_vector(dip, 150.)

# wielands reference solution
v1_ref = np.array([-2.,        0., 2 ** .5]) / 6 ** .5
v2_ref = np.array([ 1.,  3 ** 0.5, 2 ** .5]) / 6 ** .5
v3_ref = np.array([ 1., -3 ** 0.5, 2 ** .5]) / 6 ** .5

np.testing.assert_allclose(v1, v1_ref, atol=1e-7)
np.testing.assert_allclose(v2, v2_ref, atol=1e-7)
np.testing.assert_allclose(v3, v3_ref, atol=1e-7)
~~~~

krischer:
Cool! About time somebody tests that. I wrote that some years ago because somebody wanted it but never got around to really make sure it works as advertised.

I rewrote the `_dip_azimuth2ZSE_base_vector()` function as I honestly cannot understand my own code anymore...

Now it at least reproduces the formula in the book.

```python
#!/usr/bin/env python
import numpy as np
from obspy.signal.rotate import _dip_azimuth2ZSE_base_vector

dip_1, dip_2, dip_3, = -35.3, -35.3, -35.3
azimuth_1, azimuth_2, azimuth_3 = 0., 120., 240.

base_vector_1 = _dip_azimuth2ZSE_base_vector(dip_1, azimuth_1)
base_vector_2 = _dip_azimuth2ZSE_base_vector(dip_2, azimuth_2)
base_vector_3 = _dip_azimuth2ZSE_base_vector(dip_3, azimuth_3)

T = np.matrix([base_vector_1, base_vector_2, base_vector_3]).transpose()

# Matrix is in ZSE; print in XYZ  which appears to be south-east-up.
print(T[1])
print(T[2])
print(T[0])

print("=======")
print("Expected:")
print(1.0 / np.sqrt(6.0) * np.matrix((
    (-2, 1, 1),
    (0, np.sqrt(3), -np.sqrt(3)),
    (np.sqrt(2), np.sqrt(2), np.sqrt(2)))))

print("=======")

print(np.linalg.inv(T))
```

Results in 

```
[[-0.81613759  0.4080688   0.4080688 ]]
[[ 0.          0.70679589 -0.70679589]]
[[ 0.57785762  0.57785762  0.57785762]]
=======
Expected:
[[-0.81649658  0.40824829  0.40824829]
 [ 0.          0.70710678 -0.70710678]
 [ 0.57735027  0.57735027  0.57735027]]
=======
[[  5.76843359e-01  -8.16855730e-01  -3.40084515e-16]
 [  5.76843359e-01   4.08427865e-01   7.07417813e-01]
 [  5.76843359e-01   4.08427865e-01  -7.07417813e-01]]
```

martinvandriel:
Now that we are working on it, I would suggest to also offer the inverse transform by adding an optional boolean argument. In that case, just apply the inverse of matrix T (not sure if that equals the transpose in the non orthogonal case) to get UVW from ZNE.

krischer:
Sure. The interface is going to be a bit awkward in that case but I guess thats ok.

krischer:
The inverse transform appears to work now. It can do round trips with rotations. I still have no clue if the actual rotations work as advertised.

@martinvandriel If you have a real world test data set or something: Can you check if it actually does what its supposed to do?

megies:
changelog..

martinvandriel:
~~~~{.python}
    dip = -np.deg2rad(dip)
    azimuth = -np.deg2rad(azimuth)
~~~~
that cost me a few minutes...why would you do that?

krischer:
> that cost me a few minutes...why would you do that?

The SEED standard for some reason defines the dip and azimuth of instruments in a left-handed coordinate system. I agree its a bit confusing (maybe we should just do the calculations in a left-handed system...) but the docstring does state that it uses that convention. I think we should keep it because the common use for this function is to hook it up to SEED/StationXML files and autorotate borehole or other sensors with non-standard orientations.

martinvandriel:
I can't follow you how the coordinate system changes the sign of the angles, but I was more referring to the confusing redefinition of the same variable with a different meaning. Reminds me of a prominent error in a specfem paper, where this - is lost:
https://github.com/geodynamics/specfem3d/blob/master/src/shared/get_attenuation_model.f90#L354

why not just
~~~~{.python}
    dip = np.deg2rad(dip)
    azimuth = np.deg2rad(azimuth)
    return np.array([-np.sin(dip),
                     -np.cos(azimuth) * np.cos(dip),
                     np.sin(azimuth) * np.cos(dip)])
~~~~
appears to be much less confusing to me.

krischer:
Fair enough. I changed it.

> I can't follow you how the coordinate system changes the sign of the angles

The angles are in a left-handed coord system; a sign change results in the same rotation in a right handed one provided the rotation axes is the same. Now we just use the rotation matrices for a left handed system.

krischer:
Rebased on top of the releases branch and we added an additional and hopefully pretty good test. If everything passes this should be good to merge.

obspy/signal/rotate.py
```diff
@@ -155,39 +155,16 @@ def _dip_azimuth2ZSE_base_vector(dip, azimuth):
...
-                                           (c2 * c1, c2 * c2, c2 * c3),
-                                           (c3 * c1, c3 * c2, c3 * c3))) + \
-        np.sin(azimuth) * np.matrix(((0, -c3, c2), (c3, 0, -c1), (-c2, c1, 0)))
-
-    # Now simply rotate a north pointing unit vector with both matrixes.
-    temp = np.dot(azimuth_rotation_matrix, [[0.0], [-1.0], [0.0]])
-    return np.array(np.dot(dip_rotation_matrix, temp)).ravel()
+    return np.array([-np.sin(dip),
+                     -np.cos(azimuth) * np.cos(dip),
+                     np.sin(azimuth) * np.cos(dip)])
```
QuLogic: Well that's certainly simpler!

obspy/signal/rotate.py
```diff
@@ -237,19 +226,22 @@ def rotate2ZNE(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
...
 
-    # Apply it.
-    z, s, e = np.dot(T, [data_1, data_2, data_3])
-    # Replace all negative zeros. These might confuse some further processing
-    # programs.
+    if not inverse:
+        x, y, z = np.dot(T, [data_1, data_2, data_3])
+        y *= -1
+    else:
+        x, y, z = np.dot(np.linalg.inv(T), [data_1, -data_2, data_3])
```
QuLogic: Didn't @martinvandriel say the transpose was the inverse? That would be less work.

obspy/signal/tests/test_rotate.py
```diff
@@ -132,6 +133,77 @@ def test_rotate_NE_RT_NE(self):
...
+            v, 30, dip,
+            w, 150, dip)
+
+        fac = 1.0 / np.sqrt(6.0)
+
+        z_ref = np.array([fac * 3.0 * np.sqrt(2.0), 0.0, -fac * np.sqrt(2.0)])
+        n_ref = np.array([0.0, fac * 2.0 * np.sqrt(3.0), 0.0])
+        e_ref = np.array([0.0, 0.0, -4.0 * fac])
+
+        np.testing.assert_allclose(e, e_ref)
```
QuLogic: I assume this one shouldn't be here since there the same assert below.

obspy/signal/tests/test_rotate.py
```diff
@@ -132,6 +133,77 @@ def test_rotate_NE_RT_NE(self):
...
+
+        self.assertTrue(np.allclose(z, z_ref))
```
QuLogic: If you add the tolerances here like in the other tests, I suppose that Travis will pass.

obspy/signal/rotate.py
krischer: ;-)

obspy/signal/rotate.py
martinvandriel: Not in general. I think the condition might be that the axes are orthogonal to each other.

krischer: Yep. That would only be true for an orthonormal/rotation matrix which our matrix here in most cases where this method is useful is actually not.

krischer:
All tests pass. Thanks everyone for the discussions and the review!

PR #1065, opened 2015-06-01
readEvents not working with wildcarded filename

megies:
Currently `readEvents` seems broken due to the `map_example_filename` decorator, always returning `None` when reading multiple files.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 0.10.x:
   - obspy.core:
+    * Fix reading of multiple catalog files using globs (see #1065).
     * Fixed a bug when using
       `Trace.remove_response(..., water_level=None)`.
       With that setting that is supposed to not use any water level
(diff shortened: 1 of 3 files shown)


QuLogic:
It wasn't the decorator...

QuLogic:
Still failing, but I think that's because `Catalog` equality is too strict. It doesn't just check that the list of events are the same, but also that they are the same order. And `glob.glob` has no defined ordering.

krischer:
Why not just sort the filenames before reading? Then the order is deterministic and I don't see any downside to that approach.

QuLogic:
Sure, that works for the test, but the docstring for `Catalog.__eq__` says "True if both Catalogs contain the same events." and nothing about the order.

krischer:
I'd argue we should just change the documentation. While logically two event catalogs are identical if they contain the same events, the order does matter when working with catalog objects in practice.

megies:
Rebased to see what CI says.

megies:
Travis fails are unrelated, see PythonCharmers/python-future#162.

PR #1066, opened 2015-06-02
ESRI shapefile write support

megies:
Shapefiles are very handy for use in GIS applications.
Only works with `gdal` module installed, need to check how to setup Travis/Appveyor for it..

Output files are binaries, checking for binary equality seems to work, at least on the same machine.

Here's a look at the feature table of the resulting shapefile for a catalog object:

screenshot from 2015-06-02 11 06 53


ThomasLecocq:
:+1: 

I have a similar thing for outputting to KML using pyKML... could be integrated easily I assume...

megies:
Output to kml is pretty straight forward, actually (unless you're doing anything fancy, I guess).
There's an old routine for kml based on `lxml`, we should probably just outsource it as a plugin for `Catalog`..

Is your `pyKML` based stuff doing anything other than the mentioned routine?

Here's a kml example from the above routine:
```xml
<?xml version='1.0' encoding='ASCII'?>
<kml xmlns="http://www.opengis.net/kml/2.2">
  <Document>
    <name>SeisHub Event Locations</name>
    <Style id="earthquake">
      <IconStyle>
        <scale>0.5</scale>
        <Icon>
          <href>http://maps.google.com/mapfiles/kml/shapes/earthquake.png</href>
        </Icon>
        <hotSpot x="0.5" y="0" xunits="fraction" yunits="fraction"/>
      </IconStyle>
      <LabelStyle>
        <color>ff0000ff</color>
        <scale>0.8</scale>
      </LabelStyle>
    </Style>
    <Folder>
      <name>SeisHub Events (2015-06-02)</name>
      <open>1</open>
      <description>Fetched from: http://teide.geophysik.uni-muenchen.de:8080
Fetched at: 2015-06-02 13:47:09.934197

Search options:
limit=1
public=True</description>
      <Style>
        <ListStyle>
          <listItemType>check</listItemType>
          <bgColor>00ffffff</bgColor>
          <maxSnippetLines>5</maxSnippetLines>
        </ListStyle>
      </Style>
      <Placemark>
        <name>2011-01-27: 2.9</name>
        <styleUrl>#earthquake</styleUrl>
        <Style>
          <IconStyle>
            <scale>1.77611254191</scale>
          </IconStyle>
          <ListStyle>
            <maxSnippetLines>5</maxSnippetLines>
          </ListStyle>
        </Style>
        <Point>
          <coordinates>9.7538333333,47.6606666667,0</coordinates>
        </Point>
        <description>
resource_name: 5622.xml
public: False
datetime: 2011-01-27 01:24:16.769999
longitude: 9.75383333333
latitude: 47.6606666667
depth: 16060.0
magnitude: 2.89335769232
used_p: 7
used_s: 7</description>
      </Placemark>
    </Folder>
  </Document>
</kml>
```

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,181 @@
...
+                        unicode_literals)
+from future.builtins import *  # NOQA
+from future.utils import native_str
+
+from obspy import Catalog
+from obspy.core.event import Origin, Magnitude
+from obspy.core.inventory import Inventory
+
+
+try:
```
krischer: I generally prefer programs to not catch import errors and then reraise them with another description. Why not just let the original import error raise? At least to me its more clear whats going on then.

megies: I just thought I'd point out more clearly that this needs an additional module to be installed which is not a hard dependency of obspy and thus likely not yet installed. I think otherwise it's really surprising to run into an `ImportError` when using a package that otherwise handles all dependencies via the packaging system.

How about I reraise with the original message and append some more information to it?

ThomasLecocq:
no, nothing fancier...

megies:
> no, nothing fancier...

Ok, I'll open an issue to track this idea.. probably it won't hurt to have the simpler routine but without another dependency then..

obspy/io/shapefile/core.py
megies: The message now reads:
```
ImportError: No module named gdal. ObsPy's write support for
   shapefiles requires the 'gdal' module to be installed in addition
   to the general ObsPy dependencies.

That OK?
```

krischer: Works for me :-)

megies:
Here's how attribute table for shapefiles from an Inventory look like:

screenshot from 2015-06-02 17 32 24

CHANGELOG.txt
```diff
@@ -13,6 +13,7 @@ master:
      version they will fail. Pay attention to the deprecation warnings.
    * Support for additional event data formats:
      - CMTSOLUTION files used by many waveform solvers.
+     - ESRI shapefile write support, useful in GSI applications (see #1066)
```
QuLogic: GSI?

obspy/io/shapefile/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+"""
+obspy.io.shapefile - ESRI shapefile write support
+=================================================
+This module provides write support for the ESRI shapefile file format.
```
QuLogic: I know it's semi-unrelated, but the 'file' here seems unnecessary.

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def _write_shapefile(obj, filename, **kwargs):
+    """
+    Write :class:`~obspy.core.event.Catalog` object to a ESRI shapefile.
+
+    :type obj: :class:`~obspy.core.event.Catalog` or
+        :class:`~obspy.core.inventory.Inventory`
+    :param obj: ObsPy object for shapefile output
+    :type filename: str or file
```
QuLogic: I don't think this function accepts a `file`.

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+    :param filename: Filename to write to. According to ESRI shapefile
+        definition, multiple files with the following suffixes will be written:
+        ".shp", ".shx", ".dbj", ".prj".
```
QuLogic: Which one should be specified?

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    if not has_GDAL:
+        raise ImportError(IMPORTERROR_MSG)
+    if not filename.endswith(".shp"):
+        filename += ".shp"
+
+    driver = ogr.GetDriverByName(native_str("ESRI Shapefile"))
+
+    driver.DeleteDataSource(filename)
+    data_source = driver.CreateDataSource(filename)
```
QuLogic: Does this need to be closed?

QuLogic: Also, according to [these C++ docs](http://www.gdal.org/classOGRSFDriver.html), you may be using a deprecated function.

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+    # field name is 10 chars max
+    # ESRI shapefile attributes are stored in dbf files, which can not
+    # store datetimes, only dates, see:
+    # http://www.gdal.org/drv_shapefile.html
+    # use POSIX timestamp for exact origin time, set time of first pick
+    # for events with no origin
+    field_definitions = [
+        ["EventID", ogr.OFTString, 100, None],
+        ["OriginID", ogr.OFTString, 100, None],
+        ["Magnitu_ID", ogr.OFTString, 100, None],
```
QuLogic: Maybe just `MagID`, or is there some convention for this?

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+            point = ogr.Geometry(ogr.wkbPoint)
+            point.AddPoint(origin.longitude, origin.latitude)
+            feature.SetGeometry(point)
+
+        layer.CreateFeature(feature)
+        # Destroy the feature to free resources
+        feature.Destroy()
+
+    # Destroy the data source to free resources
+    data_source.Destroy()
```
QuLogic: Ah, I see it's destroyed here, okay. I think it makes more sense to destroy in `_write_shapefile`, though.

obspy/io/shapefile/core.py
```diff
@@ -0,0 +1,282 @@
...
+def _get_WGS84_spatial_reference():
+    # create the spatial reference
+    sr = osr.SpatialReference()
+    # Simpler and feels cleaner to initialize by EPSG code but that depends on
+    # a csv file shipping with GDAL and some GDAL environment paths being set
+    # correctly which was not the case out of the box in anaconda, so better
+    # hardcode this bit.
+    # sr.ImportFromEPSG(4326)
+    wgs84_wkt = \
+        """
```
QuLogic: Ah, you know, leading spaces are only trimmed in doc strings, so it wasn't necessary to place this opening on its own line (assuming that's why you did it.)

setup.py
```diff
@@ -295,10 +299,14 @@
         'readFormat = obspy.io.cmtsolution.core:_read_cmtsolution',
         'writeFormat = obspy.io.cmtsolution.core:_write_cmtsolution'
         ],
+    'obspy.plugin.event.SHAPEFILE': [
+        'writeFormat = obspy.io.shapefile.core:_write_shapefile',
+        ],
```
QuLogic: I think the indent on this closing bracket doesn't match the rest of the file.

krischer:
I just came across this: https://github.com/GeospatialPython/pyshp

Might be a better suited for this purpose as gdal is really hard to install sometimes whereas `pyshp` is pure Python.

obspy/io/shapefile/core.py
megies: Yep, thanks. Can't work with a file-like object, as it always has to be multiple files by shapefile format definition..  Overlooked this one mention of it after copy/paste.

obspy/io/shapefile/core.py
megies: I based the code on this example: pcjericks.github.io/py-gdalogr-cookbook/layers.html#create-a-new-shapefile-and-add-data

I think I read somewhere that the `Destroy()` calls are flushing the files, can't find it right now though. Maybe it could make sense to put the `Destroy()` calls in a try/finally but that's just guessing..

megies: No idea if it is also deprecated in Python, I guess it might fail to work at some point, but the python API is really hard to read so I would tend to rework it only when it starts to stop working for anybody (it'll show up in Travis pretty soon with any new gdal release I would expect.).

obspy/io/shapefile/core.py
megies: I don't care..

It's 10 char max and the automagic routines are stripping long names to 8 chars with underscore/numeric counters. But of course I didn't want to rip apart the "ID" part and thought I'd rather use what length is available to make it as clear as possible what this is referring to. But I guess I might as well change it..

obspy/io/shapefile/core.py
megies: Absolutely, I agree..

obspy/io/shapefile/core.py
megies: Just never know how to make long strings PEP8 compliant and this way didn't spring any alarm signs for me, the parser for this string doesn't seem to care about any spaces in between and neither do I. ;)

setup.py
megies: Turns out we have a pretty wild mix overall in there..

QuLogic: Oh, yea, you're right. I was just going by what I saw in this little diff.

megies:
I saw the pyshp and was weighing pros/cons of pyshp vs. gdal. I know gdal is much more of a monster but on the other hand it is *waaaay* more actively developed and maintained than pyshp (which saw it's last release 13 months ago). And since gdal is available prebuilt in anaconda I decided I'd go with the way larger and more vibrant project here.

krischer:
That's reason enough for me!

megies:
Squashed some minor commits and rebased on current master..

megies:
Tested output binaries are equal on Appveyor/Windows as well. Nice.

megies:
Good to merge I think..

PR #1068, opened 2015-06-02
obspy.mseed blockette 100 fix

krischer:
This PR fixes a bug in `obspy.mseed` that has been reported to me.

It probably never really resulted in a problem because usually

1. The sampling rate is not in the problematic range or
2. All traces have the same sampling rate

I nonetheless also added a test.


krischer:
Sorry...against the wrong branch...

PR #1069, opened 2015-06-02
obspy.mseed Blockette 100 fix

krischer:
This PR fixes a bug in `obspy.mseed` that has been reported to me.

It probably never really resulted in a problem because usually

1. The sampling rate is not in the problematic range or
2. all traces have the same sampling rate

I nonetheless also added a test.


megies:
PR looks good to me (didn't check the test).

krischer:
I wrote the test beforehand to trigger the problem so I'm pretty sure it tests the correct code path.

PR #1073, opened 2015-06-03
fixing xlimits for plotTfr + test

martinvandriel:
xlimits where not set in the tfr axes and there was no test either


QuLogic:
Assuming someone's good with the image, the code seems fine to me.

megies:
Change is only one line and only changing x limits, so really no big deal, I guess. Merging.

megies:
Actually it feels a bit like it maybe should be backported to `releases`? But maybe not worth the bother..?

PR #860, opened 2015-06-04
Sliding Window Generator on Stream and Trace objects.

krischer:
It would be really useful to have a generator on the `Stream` and `Trace` object yielding windows. I imagine something like the following:

```python
for windowed_st in st.slide(window_length=100, step=50):
    # windowed_st has to contain a copy of the data as to not modify the original data.
    windowed_st.taper()
    ...
```

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1622,11 +1622,13 @@ class Stream(object):
 
         Please keep in mind that it only returns a new view of the original
         data. Any modifications are applied to the original data as well. If
-        you don't want this you have to create a copy of the yielded windows.
+        you don't want this you have to create a copy of the yielded
+        windows. Also be aware that if you modify the original data and you
+        have overlapping windows, all following windows are affected as well.
 
         Not all yielded windows must have the same number of traces. The
-        algorithm will determine the maximal temporal extends by analysing
-        all Traces and then create windows based on these times.
+        algorithm will determine the maximal temporal extents by analysing
+        all Traces and then creates windows based on these times.
 
         .. rubric:: Example
 
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1434,8 +1434,9 @@ class Stream(object):
             given ``fill_value``. Defaults to ``False``.
         :type nearest_sample: bool, optional
         :param nearest_sample: If set to ``True``, the closest sample is
-            selected, if set to ``False``, the next sample containing the time
-            is selected. Defaults to ``True``.
+            selected, if set to ``False``, the outer (previous sample for a
+            start time border, next sample for an end time border) sample
+            containing the time is selected. Defaults to ``True``.
 
             Given the following trace containing 4 samples, "|" are the
             sample points, "A" is the requested starttime::
@@ -1567,8 +1568,9 @@ class Stream(object):
             Defaults to ``False``.
         :type nearest_sample: bool, optional
         :param nearest_sample: If set to ``True``, the closest sample is
-            selected, if set to ``False``, the next sample containing the time
-            is selected. Defaults to ``True``.
+            selected, if set to ``False``, the outer (previous sample for a
+            start time border, next sample for an end time border) sample
+            containing the time is selected. Defaults to ``True``.
 
             Given the following trace containing 4 samples, "|" are the
             sample points, "A" is the requested starttime::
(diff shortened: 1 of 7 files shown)


ThomasLecocq:
:+1:

krischer:
This turned out to be a bit more code to write than expected but should do the trick and it is tested quite well.

* I also added the `nearest_sample` argument from the `taper()` method to `Stream/Trace.slice()`.
* By default it does no processing and only returns a view of the original data. This is the fastest and most general. One can always create copies inside the loop.
* Parameters (all times are in seconds).
  * `window_length`: Window length in seconds.
  * `step`: Advancement between two windows. Can be negative.
  * `offset`: Offset from the first sample.
  * `include_partial_windows`: Also return smaller windows if there is no other way.
  * `nearest_sample`:  Same as on `Stream/Trace.trim()`.
 
Usage:

```python
for window in st.slice(window_length=5.0, step=10.0, offset=1.0,
                       include_partial_windows=True):
    # It returns a slice by default so if needed, create a copy.
    window = window.copy()
    ...
```

ThomasLecocq:
you mean, st.slide right ? :)

obspy/core/stream.py
```diff
@@ -1563,6 +1565,18 @@ def slice(self, starttime=None, endtime=None, keep_empty_traces=False):
         :type keep_empty_traces: bool, optional
         :param keep_empty_traces: Empty traces will be kept if set to ``True``.
             Defaults to ``False``.
+        :type nearest_sample: bool, optional
+        :param nearest_sample: If set to ``True``, the closest sample is
+            selected, if set to ``False``, the next sample containing the time
```
QuLogic: The word "next" is a bit confusing, because it indicates "after", but for the start time (as in the example below), it seems to mean "before".

obspy/core/stream.py
```diff
@@ -1592,12 +1606,92 @@ def slice(self, starttime=None, endtime=None, keep_empty_traces=False):
...
+              include_partial_windows=False, nearest_sample=True):
+        """
+        Generator yielding equal length sliding windows of the Stream.
+
+        Please keep in mind that it only returns a new view of the original
+        data. Any modifications are applied to the original data as well. If
+        you don't want this you have to create a copy of the yielded windows.
+
+        Not all yielded windows must have the same number of traces. The
+        algorithm will determine the maximal temporal extends by analysing
```
QuLogic: s/extends/extents/

obspy/core/stream.py
```diff
@@ -1592,12 +1606,92 @@ def slice(self, starttime=None, endtime=None, keep_empty_traces=False):
...
+        all Traces and then create windows based on these times.
```
QuLogic: s/create/creates/

obspy/core/stream.py
```diff
@@ -1592,12 +1606,92 @@ def slice(self, starttime=None, endtime=None, keep_empty_traces=False):
...
             new.append(sliced_trace)
         return new
 
+    def slide(self, window_length, step, offset=0,
+              include_partial_windows=False, nearest_sample=True):
+        """
+        Generator yielding equal length sliding windows of the Stream.
+
+        Please keep in mind that it only returns a new view of the original
+        data. Any modifications are applied to the original data as well. If
```
QuLogic: I know it's implied, but it might be good to explicitly state that it would also affect all following windows.

obspy/core/util/misc.py
```diff
@@ -490,6 +490,54 @@ def factorize_int(x):
...
+    :type step: float
+    :param offset: The offset of the first window in seconds relative to
+        the start time of the whole interval.
+    :type offset: float
+    :param include_partial_windows: Determines if windows that are
+        shorter then 99.9 % of the desired length are returned.
+    :type include_partial_windows: bool
+    """
+    if step > 0:
+        end = endtime.timestamp - 0.001 * step
```
QuLogic: Maybe you want to use `endtime.precision` or some such thing instead of `0.001`?

ThomasLecocq:
:+1: for a rapid merge of this... :-)

obspy/core/stream.py
krischer: Very important point. I added it.

obspy/core/util/misc.py
krischer: The endtime is only fed into the `np.arange` function call so if we use the precision value of the datetime objects here we might end up with numerical precision issues on the numpy side. I think using 0.1 % is pretty save.

krischer:
Thanks for the review @QuLogic. As always very helpful.

ThomasLecocq:
poum poum poum... :-) still not merged ? :)

krischer:
I also think its ready to merge.

PR #1079, opened 2015-06-12
obspy.fdsn.client.Client.get_waveforms() raises NotImplementedError

megies:
```python
Python 2.7.8 (default, Apr 14 2015, 16:15:56) 
[GCC 4.2.1 Compatible Apple LLVM 5.1 (clang-503.0.38)] on darwin
>>> from obspy.fdsn.client import Client
>>> from obspy.core.utcdatetime import UTCDateTime
>>> client = Client("IRIS")
>>> st = client.get_waveforms('N4', 'X58A', '*', 'HNZ', UTCDateTime(2015, 5, 1, 0, 0), UTCDateTime(2015, 2, 11, 0, 0))
  File "/opt/antelope/python2.7.8/lib/python2.7/site-packages/obspy/fdsn/client.py", line 594, in get_waveforms
    data_stream = self._download(url)
  File "/opt/antelope/python2.7.8/lib/python2.7/site-packages/obspy/fdsn/client.py", line 1188, in _download
    raise NotImplementedError(msg)
NotImplementedError: Bad request. Please contact the developers.
>>>
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,6 @@
 0.10.x:
+  - obspy.fdsn:
+    * More detailed error messages on failing requests (see #1079)
   - obspy.mseed:
     * Blockette 100 is now only written for Traces that need it. Previously
       it was written or not for all Traces, depending on whether the last
--- a/obspy/fdsn/client.py
+++ b/obspy/fdsn/client.py
@@ -1180,32 +1180,43 @@ class Client(object):
         code, data = download_url(
             url, headers=self.request_headers, debug=self.debug,
             return_string=return_string, data=data, timeout=self.timeout)
+        # get detailed server response message
+        if code != 200:
+            try:
+                server_info = "\n".join([
+                    line for line in data.read().splitlines() if line])
+            except:
+                server_info = None
+
         # No data.
         if code == 204:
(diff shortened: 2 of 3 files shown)


megies:
screenshot from 2015-06-12 11 48 20

megies:
http://tests.obspy.org/25751/
Single failing test is unrelated.

CTrabant:
> @CTrabant <https://github.com/CTrabant> Can it be relied upon that error messages start at stderr: and are a single line?
> 
> Doesn't matter much I think, as we also have to take into account that other FDSN servers might have completely different error info messages. I would propose to just add the complete server reply to our exception message for more user insight.
> 
The general form of the error responses is defined on page 3 of the spec: http://www.fdsn.org/webservices/FDSN-WS-Specifications-1.1.pdf.  The first line is:

Error <CODE>: <SIMPLE ERROR DESCRIPTION>

At a minimum, I suggest displaying that first line for the user.  You could choose to display the whole thing also.

megies:
> At a minimum, I suggest displaying that first line for the user. You could choose to display the whole thing also.

@CTrabant, as seen in screenshot, we're now (i.e. with next release) appending the complete server response to exception messages in FDSNWS requests.

PR #1081, opened 2015-06-18
read REFLEXW waveform data

megies:


QuLogic:
Looks like you're maybe just missing `future` imports?

obspy/io/reflexw/core.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
```
QuLogic: This file is not executable?

obspy/io/reflexw/core.py
```diff
@@ -0,0 +1,61 @@
...
+    ("timecollect", np.float64),
+    ("dummys", (np.float32, 8))])
+DATA_TRACE_HEADER_LENGTH = 154
+
+
+def _read_reflexw_datafile(filename, format_code):
+    """
+    :type filename: str
+    :param filename: Filename of REFLEXW data file.
+    :type filename: int
```
QuLogic: The name of the argument is incorrect here and below.

QuLogic:
I guess tests and documentation are missing too.

PR #1085, opened 2015-06-29
fixed calibration module for future use

jwassermann:


QuLogic:
Is this not tested? Maybe you should add one.

megies:
Added a test for this (fails on `releases`, should pass on this branch). This problem is similar to problems we experienced with `numpy.xxx(..., dtype="float")`, string handling inside numpy having problems with our `future` string handling.

PR #1087, opened 2015-06-30
setup.py: add setuptools dependency

petrrr:
There are several import from the `pkg_resources` module provided by the `setuptools` packages within the code. 
```
petr% grep pkg_ -r * |grep import
grep: misc/docs/source/tutorial/ObsPyTutorial.pdf: No such file or directory
misc/installer/salt-formula/obspy/dependencies.sls:{% from "obspy/map.jinja" import pkg_deps with context %}
obspy/core/event.py:from pkg_resources import load_entry_point
obspy/core/stream.py:from pkg_resources import load_entry_point
obspy/core/tests/test_waveform_plugins.py:from pkg_resources import load_entry_point
obspy/core/util/base.py:from pkg_resources import iter_entry_points, load_entry_point
obspy/station/inventory.py:from pkg_resources import load_entry_point
setup.py:        from pkg_resources import EntryPoint
```

Therefore adding `setuptools` as explicit requirement, its a "library dependency". 

However, given the fact that `setuptools` has to be installed anyway, we might want it to become an normal install time dependency and get ride of the try/except constructs in the `setup.py` file. Any opinions on this?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,7 @@ INSTALL_REQUIRES = [
     'scipy>=0.7.2',
     'matplotlib',
     'lxml',
+    'setuptools',
     'sqlalchemy']
 EXTRAS_REQUIRE = {
     'tests': ['flake8>=2', 'pyimgur'],


QuLogic:
> However, given the fact that setuptools has to be installed anyway, we might want it to become an normal install time dependency and get ride of the try/except constructs in the setup.py file.

Note: Dependencies for installation and dependencies for runtime are not quite the same. The former do not get actually installed, so you should specify them under both requires.

I have no objections, but I guess other packagers should chime in.

megies:
Don't really have an opinion on this. I never really thought about this anyway and never encountered any issues with this..

petrrr:
@QuLogic: Yes, I agree that install and runtime dependencies are different. 

My point here is that we need `setuptools` at runtime anyway. So why not getting rid of all the fuzz about `setuptools` yes/no/maybe [1] and just require it at install time as well. Most installations have pulled in `setuptools` at some point anyway, otherwise this bug here wouldn't have remained undetected for so long. So requiring straight away is not a big deal.

I'd assume, dependencies can be defined (or overwritten) on the level of package definition independently of `setup.py`. For Macports I now do this already. However, I usually rely on what is requires by setup.py, and these package definitions might be generated automatically. So having the requirements defined well, never hurts.

[1] https://github.com/obspy/obspy/blob/master/setup.py#L27-L30

PR #1088, opened 2015-07-02
PSD additions

megies:
Adding some psd routines written by @bmorg.

Still needs basic review, and most likely refactoring, docs, etc..

Also there's some monkey patching onto Trace/Stream in there which we need to rework.


megies:
Might need review and harmonization with #837, so punting to next major release cycle..

Issue #952, opened 2015-02-05
Base all waveform plugins on BytesIO instead of filenames


mbyt:
Currently we have something similar to that:
```
def isGSE2(filename):
def readGSE2(filename, headonly=False, verify_chksum=True,
def writeGSE2(stream, filename, inplace=False, **kwargs):  # @UnusedVariable
```
for all waveform plugins.

I definitely think support for BytesIO similar to #950 makes sense for all waveform plugins. However, I propose to open the file in core.stream.Stream.{read,write} and pass the file pointer to the waveform `{is,read,write}XXX`. Then we have the logic for opening and closing the files only once and we basically enforce that all new waveform plugins support BytesIO 

Surely, this will break code out there which directly uses  `{is,read,write}XXX`. However, the most common and the only recommended way (e.g. st.write("GSE2") stays the same.

krischer:
+1

Right now we also perform all kinds of magic in the main `read()` function which hides what is going on at times.

megies:
> Surely, this will break code out there which directly uses {is,read,write}XXX. However, the most common and the only recommended way (e.g. st.write("GSE2") stays the same.

To..
 - avoid breaking existing code and..
 - further discourage use of these methods..

..we could leave the old routines (e.g. `readMSEED()`) for one major release cycle, deprecate them and unregister them in `setup.py`.. and switch to new private methods (e.g. `_read_MSEED()`) that register in `setup.py` with the new behavior.

krischer:
I would honestly just make a cut with a new version (eventually ;-)). This is just too much work and most of these function actually state that they should **NOT** be called directly, e.g.:

http://docs.obspy.org/master/packages/autogen/obspy.mseed.core.readMSEED.html

So we never made any guarantees on the stability of these methods so they are fair game and can be changed however we desire :-)

megies:
Yeah sure.. it's a bit overhead but not too much I would say. And it might be a good spot to actually make them private functions (to be even clearer about them not being intended to be used). Anyway.. I'm fine with either way to do it. :)

mbyt:
> And it might be a good spot to actually make them private functions (to be even clearer about them not being intended to be used)

+1

QuLogic:
Same/related to #852.

megies:
Closing as a duplicate..

Issue #107, opened 2012-09-09
Confused by "starttime" when reading sac files


trac2github:
I'm confused by the meaning of "starttime" when reading sac files using obspy.sac or obspy.core .

Using, obspy.sac I do the follwoing, in order to read the file "seism.sac":

``` python
import obspy.sac
sac = obspy.sac.ReadSac("seism.sac")
print(sac.starttime)
```

While using, obspy.core, I do:

``` python
from obspy.core import read
tr = read("seism.sac")
print tr
```

In both cases I get the same starting time, which is not the time of the first sample, but rather the SAC reference time, which is defined by the variables NZJEAR, NZJDAY, NZHOUR, NZMIN, NZSEC, NZMSEC.

This time corresponds to the starting time (which for me is the time of the first sample) only if the header field "B" is equal to 0.

"B" is in fact the time of the first sample. Like all the other time markers (e.g. "A", "T0", "E") it is expressed in seconds respect to the reference time.

To test this, open sac and generate a sample seismogram using the following commands ("SAC>" is the sac prompt):

```
SAC> fg seism
SAC> w seism.sac
```

The seismogram generated has the following reference time: 1981-03-29 10:38:14.0

However B=9.4599991, so the time of the first sample (starttime) is 1981-03-29 10:38:23.46

You can verify this within SAC either visually, using a plotting command, or using the command "sync", which takes B=0 and adjusts the reference time accordingly:

```
SAC> fg seism
SAC> sync
SAC> w seism_sync.sac
```

PR #1092, opened 2015-07-09
Sac rewrite

jkmacc-LANL:
## Why a re-write?
The `sacio` module underlying ObsPy's SAC handling was a bit hard to follow, as it has a long inheritance, which made it hard to track down issues or make fixes.  This re-write attempts to make the SAC plugin easier to understand and maintain, as well as offer some potential improvements.

### Improve maintainability.
I've split out the header specification (`header.py`), the low-level array-based SAC file I/O (`arrayio.py`), and the object-oriented interface (`sactrace.py`), whereas it was previously all within one `sacio.py` module.  I hope that the flow of how each plugs into the other is clear, so that bug tracking is straight-forward, and that hacking on one aspect of SAC handling is not cluttered/distracted by another.

### Expand support for round-trip SAC file processing
This PR attempts to improve support for a common work flow: read one or more SAC files into ObsPy, do some processing, then (over)write them back as SAC files that look mostly like the originals.  Previously, ObsPy Traces written to SAC files wrote only files based on the first sample time (iztype 9/'ib').  In [`util.py:obspy_to_sac_header`](https://github.com/jkmacc-LANL/obspy/blob/sac-rewrite/obspy/io/sac/util.py#L237)  of this PR, if an old `tr.stats.sac` SAC header is found, the *iztype and reference "nz" times are used and kept*, and the "b" and "e" times of the Trace being written are adjusted according to this reference time.  This preserves the absolute reference of any relative time headers, like t0-t9, carried from the old SAC header into the new file.  This can only be done if SAC to Trace conversion preserves these "nz" time headers, which is possible with the current `debug_headers=True` flag.

## To Do:
1. [x] Testing integration.
2. ~~[ ] setup.py integration.~~
3. [x] Docs.

## Feedback
Comments and suggestions about the following would be greatly appreciated:

1. The design decisions and round-trip processing support mentioned above
2. License language and attribution  
    As much of the hard work of basic I/O came from the old `sacio`, I only wish to tack my name onto the author list for this module.  Please offer any suggestions to make that clear.
3. My git-fu.  
    This is the first time I've done anything that might be merged into a larger project, so any comments or suggestions on how to improving my use of git would be welcomed.

EDIT: shoutouts
@yannikbehr , @mbyt , @cja12, @claudiodsf


obspy/io/sac/header.py
```diff
@@ -0,0 +1,289 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.utils import native_str
```
QuLogic: Unused.

obspy/io/sac/sacpz.py
```diff
@@ -94,6 +100,295 @@ def _write_sacpz(inventory, file_or_file_object):
...
+    -31.62 0.00
+    0.00 0.00
+    0.00 0.00
+    '''
+
+    poles = []
+    zeros = []
+
+    if isinstance(paz_file, (str, native_str)):
+        paz_file = open(paz_file, 'r')
```
QuLogic: You need to use a `try`/`finally` to ensure the file is closed even if an error occurs.

obspy/io/sac/sacpz.py
```diff
@@ -94,6 +100,295 @@ def _write_sacpz(inventory, file_or_file_object):
...
+                        poles.append(complex(0, 0j))
+                    break
+                else:
+                    poles.append(complex(float(a[0]), float(a[1])))
+        if line.find('CONSTANT') != -1:
+            a = line.split()
+            # in the observatory this is the seismometer gain [muVolt/nm/s]
+            # the A0_normalization_factor is hardcoded to 1.0
+            constant = float(a[1])
+    paz_file.close()
```
QuLogic: I don't think you want to close the file if you didn't open it.

obspy/io/sac/LICENSE.txt
```diff
@@ -0,0 +1,16 @@
+Copyright 2015. Los Alamos National Security, LLC for pysac LA-CC-15-051. This
```
QuLogic: This is not pysac.

obspy/io/sac/LICENSE.txt
```diff
@@ -0,0 +1,16 @@
...
+material was produced under U.S. Government contract DE-AC52-06NA25396 for Los
+Alamos National Laboratory (LANL), which is operated by Los Alamos National
+Security, LLC for the U.S. Department of Energy. The U.S. Government has rights
+to use, reproduce, and distribute this software.  NEITHER THE GOVERNMENT NOR
+LOS ALAMOS NATIONAL SECURITY, LLC MAKES ANY WARRANTY, EXPRESS OR IMPLIED, OR
```
QuLogic: This sentence seems redundant to existing LGPL terms.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
+# -*- coding: utf-8 -*-
+"""
+Low-level array interface to the SAC file format.
+
+Functions in this module work directly with numpy arrays that mirror the SAC
```
QuLogic: s/numpy/NumPy/

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+
+These read/write routines are very literal; there is almost no value or type
+checking, except for byteorder and header/data array length.  File- and array-
+based checking routines are provided for additional checks where desired.
+
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.utils import native_str
+from future.builtins import *
```
QuLogic: `# noqa` is needed here.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+from .util import SacIOError, SacInvalidContentError
+from .util import is_valid_enum_int, is_same_byteorder, byteswap
+
+
+def init_header_arrays(*args):
+    """
+    Initialize arbitrary header arrays.
+
+    Parameters
+    ----------
```
QuLogic: Unfortunately, I don't think we use this NumPy style of docstring.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+
+    Returns
+    -------
+    arrays : list of numpy.ndarrays
+        The desired SAC header arrays.
+
+    """
+    if len(args) == 0:
+        args = ('float', 'int', 'str')
+    #http://stackoverflow.com/a/13052254/745557
```
QuLogic: What is the relevance of this link?

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+            # XXX: why not init native byte order?
+            hf = np.ndarray(70, dtype=native_str('<f4'))
+            hf.fill(HD.FNULL)
+            out.append(hf)
+        elif itype == 'int':
+            # null integer header array
+            hi = np.ndarray(40, dtype=native_str('<i4'))
+            hi.fill(HD.INULL)
+            # set logicals to 0, not -1234whatever
+            for hdr in HD.INTHDRS:
```
QuLogic: `for i, hdr in enumerate(HD.INTHDRS):` and remove the `index`ing below.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+                if hdr.startswith('l'):
+                    hi[HD.INTHDRS.index(hdr)] = 0
+            # TODO: initialize enumerated values to something?
+            # calculate distances be default
```
QuLogic: s/be/by/

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+        doesn't match header, or header arrays are incorrect length.
+
+    """
+    # TODO: rewrite using "with" statement instead of open/close management.
+    # checks: byte order, header array length, file size, npts matches data length
+    try:
+        f = open(source, 'rb')
+        is_file_name = True
+    except IOError:
+        raise SacIOError("No such file: " + source)
```
QuLogic: Is there really any benefit to wrapping this exception?

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    except TypeError:
+        # source is already a file-like object
+        f = source
+        is_file_name = False
+
+    isSPECIFIED = byteorder is not None
```
QuLogic: This is an odd name.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    # READ HEADER
+    # The sac header has 70 floats, 40 integers, then 192 bytes
+    #    in strings. Store them in array (and convert the char to a
+    #    list). That's a total of 632 bytes.
+    # --------------------------------------------------------------
+    hf = from_buffer(f.read(4 * 70), dtype=native_str(endian_str + 'f4'))
+    hi = from_buffer(f.read(4 * 40), dtype=native_str(endian_str + 'i4'))
+    hs = from_buffer(f.read(24 * 8), dtype=native_str('|S8'))
+
+    isVALID = is_valid_byteorder(hi)
```
QuLogic: This variable is kind of useless.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    if headonly:
+        data = None
+    else:
+        data = from_buffer(f.read(npts * 4), dtype=native_str(endian_str + 'f4'))
+
+        if len(data) != npts:
+            f.close()
+            raise SacIOError("Cannot read all data points")
+
+    if is_file_name:
```
QuLogic: This should be in a `try`/`finally`.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    try:
+        fh = open(source, 'rb')
+        is_file_name = True
+    except IOError:
+        raise SacIOError("No such file: " + source)
+    except TypeError:
+        fh = source
+        is_file_name = False
+
+    contents = fh.read()
```
QuLogic: If you're going to read the whole thing, you should close the file right here.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+
+    contents = [_i.rstrip(b"\n\r") for _i in contents.splitlines(True)]
```
QuLogic: Why keed the newlines if you're going to strip them?

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    if data is None:
+        # file exists, just modify it (don't start from scratch)
+        fmode = 'rb+'
+    else:
+        # start from scratch
+        fmode = 'wb+'
+
+    # TODO: use "with" statements (will always closes the file object?)
+    try:
+        f = open(dest, fmode)
```
QuLogic: You should open the file after doing all the checks below.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+        is_file_name = True
+    except IOError:
+        raise SacIOError("Cannot open file: " + dest)
+    except TypeError:
+        f = dest 
+        is_file_name = False
+
+    if data is None and (f.mode != 'rb+'):
```
QuLogic: Unnecessary parenthesis.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+        raise ValueError(msg)
+
+    # deal with desired byte order
+    if data is None:
+        assert hf.dtype.byteorder == hi.dtype.byteorder
+    else:
+        assert hf.dtype.byteorder == hi.dtype.byteorder == data.dtype.byteorder
+
+    if byteorder is None:
+        pass
```
QuLogic: There's not really any need for this empty block.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    try:
+        f = open(dest, fmode)
+        is_file_name = True
+    except IOError:
+        raise SacIOError("Cannot open file: " + dest)
+    except TypeError:
+        f = dest
+        is_file_name = False
+
+    if data is None and (f.mode != 'r+'):
```
QuLogic: Unnecessary parenthesis.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    Returns
+    -------
+    dict
+        The correctly-ordered SAC header values, as a dictionary.
+    nulls : bool
+        If True, return all values including nulls.
+
+    """
+    if nulls:
+        items = [(key, val) for (key, val) in zip(HD.FLOATHDRS, hf)] + \
```
QuLogic: This list comprehension is unnecessary. Use `list(zip(HD.FLOATHDRS, hf))` to create the list from the iterator.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+                [(key, val) for (key, val) in zip(HD.STRHDRS, hs) if val != HD.SNULL]
+
+    header = dict(items)
+
+    # here, we have to append the 2nd kevnm field into the first and remove
+    #   it from the dictionary.
+    # XXX: kevnm may be null when kevnm2 isn't
+    if 'kevnm2' in header:
+        if 'kevnm' in header:
+            header['kevnm'] += header.pop('kevnm2').decode()
```
QuLogic: I don't see you decoding `kevnm` too.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    Returns null hf, hi, hs arrays, optionally filled with values from a
+    dictionary.  No header checking.
+
+    """
+    hf, hi, hs = init_header_arrays()
+
+    # have to split kevnm into two fields
+    # TODO: add .lower() to hdr lookups, for safety
+    if header is not None:
+        for hdr, value in header.iteritems():
```
QuLogic: You need to replace `iteritems` with `items` for Python 3.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+            if hdr in HD.FLOATHDRS:
+                hf[HD.FLOATHDRS.index(hdr)] = value
+            elif hdr in HD.INTHDRS:
+                if not isinstance(value, (np.integer, int)):
+                    warnings.warn("Non-integers may be truncated.")
+                    print(" {}: {}".format(hdr, value))
```
QuLogic: This should be part of the warning, I think.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+            val = hi[HD.INTHDRS.index(hdr)]
+            if val not in (0, 1, HD.INULL):
+                msg = "Header '{}' must be {{{}, {}, {}}}."
+                raise SacInvalidContentError(msg.format(hdr, 0, 1, HD.INULL))
+
+    if 'data_hdrs' in tests:
+        try:
+            isMIN = hf[HD.FLOATHDRS.index('depmin')] == data.min()
+            isMAX = hf[HD.FLOATHDRS.index('depmax')] == data.max()
+            isMEN = hf[HD.FLOATHDRS.index('depmen')] == data.mean()
```
QuLogic: `isMEN`?

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+        # all header reference time fields are set
+        if not all([val != HD.INULL for val in [nzyear, nzjday, nzhour, nzmin, nzsec, nzmsec]]):
+            msg = "Null reference time values detected."
+            raise SacInvalidContentError(msg)
+
+        # reference time fields are reasonable values
+        nzjday_ok = 0 <= nzjday <= 366
+        nzhour_ok = 0 <= nzhour <= 24
+        nzmin_ok = 0 <= nzmin <= 59
+        nzsec_ok = 0 <= nzsec <= 59
```
QuLogic: Seconds can go up to 60, so I assume it's SAC that doesn't allow it.

obspy/io/sac/header.py
```diff
@@ -0,0 +1,289 @@
...
+from future.builtins import *
+
+MODULE_DOCSTRING = """
+SAC header specification, including documentation.
+
+Header names, order, and types, nulls, as well as allowed enumerated values, are
```
QuLogic: The "and" appears a bit early in this list.

obspy/io/sac/header.py
```diff
@@ -0,0 +1,289 @@
...
+# Module documentation string
+__doc__ = MODULE_DOCSTRING + HEADER_DOCSTRING
+
+# ------------ NULL VALUES ----------------------------------------------------
+FNULL = -12345.0
+INULL = -12345
+SNULL = b'-12345  '
+
+# ------------ HEADER NAMES, TYPES, ARRAY POSITIONS ---------------------------
+# these are useful b/c they can be use forwards or backwards, like:
```
QuLogic: used

obspy/io/sac/header.py
```diff
@@ -0,0 +1,289 @@
...
+             'inc': 83, 'io_': 84, 'il': 85, 'ir': 86, 'it': 87, 'iu': 88,
+             'ieq3': 89, 'ieq0': 90, 'iex0': 91, 'iqc': 92, 'iqb0': 93,
+             'igey': 94, 'ilit': 95, 'imet': 96, 'iodor': 97, 'ios': 103}
+
+# reverse look-up: you have the number, want the string
+ENUM_NAMES = dict((v,k) for k,v in ENUM_VALS.iteritems())
+
+
+# accepted values, by header
+ACCEP_VALS = {'iftype': ['itime', 'irlim', 'iamph', 'ixy'],
```
QuLogic: Why leave off the T?

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+header values in a user-friendly way. This includes some value-checking, native
+Python logicals (True, False) and nulls (None) instead of SAC's 0, 1, or -12345...
+
+SAC headers are implemented as properties, with appropriate getters and setters.
+
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.utils import native_str
+from future.builtins import *
```
QuLogic: Needs a `# noqa`.

obspy/io/sac/sacpz.py
megies: Or just use the `with open(paz_file, 'r') as paz_file:` block statement.

obspy/io/sac/LICENSE.txt
megies: If it's LGPL we don't need a License file in this submodule. The copyright information could be moved to the README.txt and/or `__init.__.py`, I think.

obspy/io/sac/arrayio.py
jkmacc-LANL: Discussion and timings for the fastest way to fill a NumPy array.  Add a note to this effect above the link?

obspy/io/sac/arrayio.py
jkmacc-LANL: `is_byteorder_specified` ?

obspy/io/sac/arrayio.py
jkmacc-LANL: This is an inherited snippet, and I don't know which edge case it was written for, it any.  Maybe I'll remove it in the current obspy master and see if any sacio/sac.core tests fail.

obspy/io/sac/arrayio.py
jkmacc-LANL: ? Please elaborate.
`try` assumes it's a file name, and should be opened. `except TypeError:` catches an already-open file object, so no opening should happen below it.

obspy/io/sac/arrayio.py
jkmacc-LANL: Only male variable names are allowed. :wink:
(Keeping parallel with the header name.)

obspy/io/sac/LICENSE.txt
jkmacc-LANL: I had to call it something when it was processed for release.  I'm fuzzy on what my obligations are, but I bet they'd be satisfied if I put everything new under `obspy/io/sac/pysac` (including the wonky redundant LICENSE.txt) so that it nominally retains that name.  Thoughts?

obspy/io/sac/arrayio.py
jkmacc-LANL: :disappointed:  I know reST makes pretty web pages, but docstrings are generally viewed as text, and reST is hideous to look at.

obspy/io/sac/LICENSE.txt
krischer: As long as we can keep distributing all of ObsPy under LGPL I am fine with it. Could you not just call it `obspy.io.sac (originally based on pysac)` or something along these lines?

obspy/io/sac/arrayio.py
krischer: True enough but for consistency with the rest of ObsPy, please use our normal docstring style. If someone is up to rewrite our documentation using the NumPy style that would be great!

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    if len(args) == 0:
+        args = ('float', 'int', 'str')
+    #http://stackoverflow.com/a/13052254/745557
+    out = []
+    for itype in args:
+        if itype == 'float':
+            # null float header array
+            # XXX: why not init native byte order?
+            hf = np.ndarray(70, dtype=native_str('<f4'))
+            hf.fill(HD.FNULL)
```
krischer: This is a bit obscure to read. Can you use `np.full()` which is only marginally slower but much easier to read in my opinion:

```python
In [2]: %timeit a = np.empty(1e6, dtype=np.float32); a.fill(5.0)
The slowest run took 4.91 times longer than the fastest. This could mean that an intermediate result is being cached
1000 loops, best of 3: 344 µs per loop

In [5]: %timeit a = np.full(1e6, fill_value=5.0, dtype=np.float32)
1000 loops, best of 3: 355 µs per loop
```

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    """
+    if len(args) == 0:
+        args = ('float', 'int', 'str')
+    #http://stackoverflow.com/a/13052254/745557
+    out = []
+    for itype in args:
+        if itype == 'float':
+            # null float header array
+            # XXX: why not init native byte order?
+            hf = np.ndarray(70, dtype=native_str('<f4'))
```
krischer: Picking up on the todo item: Why not just use the native byte order and convert to it in case it differs from what is stored in the SAC file?

Did anyone actually ever run ObsPy on a big endian architecture?

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    Raises
+    ------
+    ValueError
+        Unrecognized byte order.
+    SacIOError
+        File not found, incorrect specified byteorder, theoretical file size
+        doesn't match header, or header arrays are incorrect length.
+
+    """
+    # TODO: rewrite using "with" statement instead of open/close management.
```
krischer: FYI: Just using `with` is not enough if we want it to work with filenames, open files and `io.BytesIO` objects. In the long run we should implement #952. For now, I think what has been done here is a fairly simple solution to the problem: https://github.com/obspy/obspy/blob/master/obspy/io/cmtsolution/core.py

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+    -----
+    A user can/should not _create_ a header-only binary file.  Use mode 'wb+'
+    for data=None (headonly) writing.
+
+    """
+
+    # deal with file name versus File-like object, and file mode
+    if data is None:
+        # file exists, just modify it (don't start from scratch)
+        fmode = 'rb+'
```
krischer: Do we really need this functionality? What would that be used for? Modifying the header values in a very large amount of SAC files? I don't think the potential performance benefits justify the additional complexity. This might also easily produce corrupt SAC files. If there is a serious use case where this speedup actually matters please ignore my comment.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+
+    if data is None and (f.mode != 'rb+'):
+        msg = "File mode must be 'wb+' for data=None."
+        raise ValueError(msg)
+
+    # deal with desired byte order
+    if data is None:
+        assert hf.dtype.byteorder == hi.dtype.byteorder
+    else:
+        assert hf.dtype.byteorder == hi.dtype.byteorder == data.dtype.byteorder
```
krischer: Please replace the `assert` statements with a more descriptive exception.

obspy/io/sac/arrayio.py
krischer: An even simpler way would to just check if the date is actually valid:

```python
try:
    obspy.UTCDateTime(year=nzyear, julday=nzjday, hour=nzhour, minute=nzmin,
                     second=nzsec, microsecond=nzmsec)
except ValueError as e:
    raise SacInvalidContentError("Invalid reference time: %s" % str(e))
```

This also does not allow seconds of 60 but that is something we should deal with in the `UTCDateTime` class.

obspy/io/sac/arrayio.py
```diff
@@ -0,0 +1,637 @@
...
+        if not all([val != HD.INULL for val in [nzyear, nzjday, nzhour, nzmin, nzsec, nzmsec]]):
+            msg = "Null reference time values detected."
+            raise SacInvalidContentError(msg)
+
+        # reference time fields are reasonable values
+        nzjday_ok = 0 <= nzjday <= 366
+        nzhour_ok = 0 <= nzhour <= 24
+        nzmin_ok = 0 <= nzmin <= 59
+        nzsec_ok = 0 <= nzsec <= 59
+        nzmsec_ok = 0 <= nzsec <= 999
```
krischer: should be `nzmsec` here.

obspy/io/sac/LICENSE.txt
jkmacc-LANL: @krischer I think that would be fine.

obspy/io/sac/util.py
```diff
@@ -0,0 +1,357 @@
...
+    stats['npts'] = npts
+    stats['sampling_rate'] = np.float32(1.) / np.float32(delta)
+    stats['network'] = _clean_str(knetwk)
+    stats['station'] = _clean_str(kstnm)
+    stats['channel'] = _clean_str(kcmpnm)
+    stats['location'] = _clean_str(khole)
+    stats['calib'] = calib
+
+    # store _all_ provided SAC header values
+    stats['sac'] = sacheader.copy()
```
krischer: I'm not sure if this is done elsewhere: The values in the `stats.sac` attributes must be set to -12345 in the case they are not set. I agree that having `None` is nicer (as is done in the `SACTrace` class) but enough code that tests for -12345 is in existence and we don't really want to break that (I know I did that a fair number of times...).

krischer:
This is great! Thanks a lot. I did not yet fully read it but so far I like it.

I had enough people that wanted to have full control over all the `SAC` headers which always was a bit difficult with the old implementation so a seperete `SACTrace` class seems to be useful indeed.

Furthermore I think it would a good idea to have the `SACTrace` class only deal with file pointers and open `io.BytesIO` objects. Then all the file opening/closing and further issues can be handled further up the chain.

obspy/io/sac/arrayio.py
jkmacc-LANL: My apologies; I should've been clear that, of course, I will fix the docstrings.

obspy/io/sac/arrayio.py
jkmacc-LANL: This line is inherited.  I also like the idea of just initializing in native byte order, then doing any checks at the appropriate time elsewhere.

obspy/io/sac/arrayio.py
jkmacc-LANL: I realize that this functionality adds complexity, and isn't quite ready yet, but modifying lots SAC headers without touching the data is very common in my experience.  Providing a way to muck with the headers en masse suggests that ObsPy(.io.sac...) may be used to prep data for the many research codes that use ancillary information in SAC headers, instead of using an installation of SAC itself to do that.  Again, I know that file I/O is ObsPy's main interest in an `io.sac` module, but some people's main interest in ObsPy may be to have a little control over SAC files without the SAC program.  

As far as performance goes, I haven't tested it myself.  For small files, it's probably negligible.  But for longer files, I can see a potential performance and memory penalty for reading in all the data.  I assume there was a reason to program this functionality into the SAC program.

I won't push too hard for this, but I'd like to try to do it.

obspy/io/sac/util.py
jkmacc-LANL: I think this function, as it is currently used, will pass -12345 values to `stats.sac`.  I'll check.  Either way, the job of this function is to pass the correct values as ObsPy expects them, so I'll make sure it does.

obspy/io/sac/arrayio.py
krischer: As long as there is some actual use case for this I am fine with it. I am not sure why it was originally introduced.

One thing that just came to my mind is the possibility to mass-correct and potentially autofill SAC headers (you already have some functionality with the distance and azimuth calculations baked in).

jkmacc-LANL: Sometimes event or pick information needs to be in the header for processing, or the station height is in km but should be in meters.  SAC isn't a great format, but a lot of code still plugs into those headers.  Also, once a user uses ObsPy to pre-process their data, they're one step closer to just using ObsPy to *process* their data.

obspy/io/sac/arrayio.py
jkmacc-LANL: Agreed, but list(zip(..)) obfuscates the difference between this block and the next one, for which a list comprehension seems necessary.

obspy/io/sac/arrayio.py
jkmacc-LANL: Right, except warnings only warn once for the same problem.  The user will miss all other key-values that may have been truncated.

obspy/io/sac/arrayio.py
QuLogic: We have a big-endian arch at our cluster here. I'll see if I can get all the deps together to test (it's pretty annoying.)

obspy/io/sac/arrayio.py
QuLogic: Sure, but see @krischer's other comment.

obspy/io/sac/arrayio.py
QuLogic: Sounds good.

obspy/io/sac/arrayio.py
QuLogic: There are various checks for byte order and such below that could possibly raise an exception. Instead of wrapping it all in a `try` and indenting it, it's better to just open the file for the least amount of time as possible.

obspy/io/sac/arrayio.py
QuLogic: I don't see any reason why they need to be the same.

You could instead change the condition to `if nulls or val != HD.*NULL` and have only one loop.

obspy/io/sac/arrayio.py
QuLogic: Warnings with different messages are not ignored.

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+def _boolgetter(hdr):
+    def get_bool(self):
+        value = self._hi[HD.INTHDRS.index(hdr)]
+        return bool(value)
+    return get_bool
+
+def _boolsetter(hdr):
+    def set_bool(self, value):
+        if value not in (True, False, 1, 0):
+            msg = "Logical header values must be {True, False, 0, 1}"
```
QuLogic: This message has the numbers "backwards".

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        try:
+            value = func(self.data)
+        except TypeError:
+            # data=None (headonly=True)
+            try:
+                value = self._hf[HD.FLOATHDRS.index(hdr)]
+                null = HD.INULL
+            except ValueError:
+                # hdr is 'npts', the only integer
+                # Will this also trip if a data-centric header is misspelled?
```
QuLogic: Yes, it will.

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+
+            # get the float seconds between the old and new reftimes
+            shift = old_reftime - new_reftime
+
+            # shift the relative time headers
+            self._allt(np.float32(shift))
+
+        except AttributeError:
+            msg = "New reference time must be an obspy.UTCDateTime instance."
+            raise ValueError(msg)
```
QuLogic: Isn't that a `TypeError`?

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        # XXX: user can put non-SAC key:value pairs into the header.
+        header.update(kwargs)
+
+        # -------------------------- DATA ARRAY -------------------------------
+        if data is None:
+            # this is like "headonly=True"
+            pass
+        else:
+            if not isinstance(data, np.ndarray):
+                raise SacError("data needs to be a numpy.ndarray")
```
QuLogic: Is this not a `TypeError`?

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        ----------
+        hf, hi, hs : numpy.ndarray
+            Header arrays of dtype 'f4', int, '|S08', respectively.
+        data : numpy.ndarray
+            Corresponding seismogram array.
+
+        Examples
+        --------
+        >>> sac = SACTrace._from_arrays()
+        >>> print sac
```
QuLogic: `print(sac)`

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+
+        if hf is None:
+            hf = hf0
+        if hi is None:
+            hi = hi0
+        if hs is None:
+            hs = hs0
+
+        # get the default instance, but replace the arrays
+        # itializes arrays twice, but it beats converting empty arrays to a
```
QuLogic: initializes

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        Raises
+        ------
+        SacInvalidContentError
+            If any specified tests fail.
+        ValueError
+            'data_hdrs' is specified and data is None, empty array
+
+        Examples
+        --------
+        >>> sac = SACTrace.read(filename)
```
QuLogic: This doctest should be skipped, I think.

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+
+        """
+        # interpret hdrlist
+        if hdrlist == 'all':
+            hdrlist = sorted(self._header.keys())
+        elif hdrlist == 'picks':
+            hdrlist = sorted(('b', 'e', 'o', 'a', 'f', 't0', 't1', 't2', 't3',
+                              't4', 't5', 't6', 't7', 't8', 't9'))
+        elif hdrlist == 'special':
+            warnings.warn("Not implemented: use <arrow-up> in your interpreter")
```
QuLogic: I don't understand what this means.

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        """
+        Produce a print-friendly string of header values for __repr__ ,
+        .listhdr(), and .lh()
+
+        """
+        # interpret hdrlist
+        if hdrlist == 'all':
+            hdrlist = sorted(self._header.keys())
+        elif hdrlist == 'picks':
+            hdrlist = sorted(('b', 'e', 'o', 'a', 'f', 't0', 't1', 't2', 't3',
```
QuLogic: Why not pre-sort them? You only need to move 'a' and 'f'.

obspy/io/sac/sactrace.py
```diff
@@ -0,0 +1,1127 @@
...
+        else:
+            header_str.append("\tunrecognized iztype: {}".format(iztype))
+        #
+        # non-null headers
+        hdrfmt = "{:10.10s} = {}"
+        for hdr in hdrlist:
+            # XXX: non-null header values might have no property for getattr
+            try:
+                header_str.append(hdrfmt.format(hdr, getattr(self, hdr)))
+            except AttributeError as e:
```
QuLogic: `e` appears unused.

obspy/io/sac/tests/test_sactrace.py
```diff
@@ -0,0 +1,187 @@
+import os
+import unittest
+import datetime
+
+import numpy as np
+
+from future.utils import native_str
```
QuLogic: Unused. And you're missing the other `__future__` imports.

obspy/io/sac/tests/test_sactrace.py
```diff
@@ -0,0 +1,187 @@
...
+        sac = SACTrace.read(self.file, byteorder='little')
+        self.assertEqual(sac.npts, 100)
+        self.assertEqual(sac.kstnm, 'STA')
+        self.assertEqual(sac.delta, 1.0)
+        self.assertEqual(sac.kcmpnm, 'Q')
+        self.assertEqual(sac.reftime.datetime, datetime.datetime(1978, 7, 18, 8, 0))
+        self.assertEqual(sac.nvhdr, 6)
+        self.assertEqual(sac.b, 10.0)
+        self.assertAlmostEqual(sac.depmen, 9.0599059e-8)
+        np.testing.assert_array_almost_equal(self.testdata[0:10],
```
QuLogic: I forget what version of NumPy we updated master to use. Are we allowed to use this test, @krischer?

QuLogic: Ah, I see it's in the old code, so it's probably fine.

obspy/io/sac/util.py
```diff
@@ -0,0 +1,357 @@
+# -*- coding: utf-8 -*-
+"""
+SAC module helper functions and data.
+
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.utils import native_str
```
QuLogic: Unused.

obspy/io/sac/util.py
```diff
@@ -0,0 +1,357 @@
...
+from future.builtins import *
```
QuLogic: Needs a `# noqa`.

QuLogic:
There are many spacing issues; you should run `flake8` locally to find them more quickly (but they're also there on Travis.)

obspy/io/sac/arrayio.py
jkmacc-LANL: :+1

obspy/io/sac/arrayio.py
jkmacc-LANL: Aren't there already big endian files in sac/tests/data?

obspy/io/sac/sacpz.py
jkmacc-LANL: @QuLogic Please post a snippet.  This function is unchanged from current master, and I'm not sure where to put a `try / finally`.

QuLogic: You need to wrap everything during which the file is open:
```
try:
    # All file IO stuff.
finally:
    if was_opened_by_us:
        paz_file.close()
```

jkmacc-LANL:
@QuLogic Is longer than 80 columns OK if it's more readable?

krischer:
@jkmacc-LANL No...we enforce most of PEP8 through this test here: https://github.com/obspy/obspy/blob/master/obspy/core/tests/test_code_formatting.py

While sometimes a bit annoying consistency in our code base helps a lot in the long run.

jkmacc-LANL:
I saw this was tagged for 0.11.  Is that close?

krischer:
Not that close. I guess its at least a couple of months away. If this does not get ready for 0.11 we'll just move it to the next release ;-)

But it would be a good fit for 0.11 as we already did some major changes and refactorings for it.

jkmacc-LANL:
In trying to make the `io.sac` module tests pass, I noticed that some of the problems were because the new implementation is more strict about some things than the old one.  Specifically, some of the test data have inconsistent headers (e.g. `depmen != data.mean()`), and `SACTrace` doesn't like that.  Is this intentional, so that minor header issues are tolerated?   If not, is it possible to make appropriate changes in the test data, or should I just scale back the strictness of the new implementation?

obspy/io/sac/arrayio.py
QuLogic: If you're interested, I've found [this project](https://github.com/dadadel/pyment) which lets you nearly automatically convert. Unfortunately, there are some bugs with doctests, but hopefully they'll get fixed sometime soon.

megies:
For something like `depmen != data.mean()` I would probably issue a warning (and maybe update that header variable) but not raise an error. In general being a bit more strict is OK I think, slightly adapting tests (or test data) in that case should be fine. Of course adapting tests can be a bit delicate, so being verbose about test changes (e.g. commenting on them here) is really nice for review.

obspy/io/sac/arrayio.py
jkmacc-LANL: @QuLogic Many thanks!

obspy/io/sac/tests/test_core.py
```diff
@@ -316,7 +316,7 @@ def test_undefinedB(self):
             self.assertEqual(tr2.stats.starttime.timestamp, 269596810.0)
```
jkmacc-LANL: It is not clear to me what `test_undefinedB` is testing, and I think it may need to change.  This test seems to make sure that unsetting tr.stats.sac.b doesn't change the tr.stats.starttime with the trace is written and re-read in. It does this, however, by changing the SAC reference time ("nz" header fields).  You can see this in `self.assertEqual(sac_ref_time.timestamp, 269596800.0)` and `self.assertEqual(sac_ref_time2.timestamp, 269596810.0)`.  Changing the SAC reference time is not OK, because it invalidates all other relative time headers in the file (i.e. a, o, and t0-t9, which are commonly used to store a number of phase picks and windows).  Also, if the file is iztype "io", this is equivalent to changing an event origin time.  This test preserves tr.stats.starttime when "b" is unset only by invalidating the SAC file.  I think there is a better way.

When writing a SAC file from a Trace that was also read in from a SAC file, the old tr.stats.sac.b doesn't matter.  One only has to check that the old tr.stats.sac header has a valid reference time, compare it to tr.stats.starttime, and then write a new "b" (and "e") that respects this difference.  No `iztype` checking or changes to the SAC reference time is required, and any relative time headers therein are preserved.  

I would like to change this test such that it (1) makes sure tr.stats.starttime does not change even if "b" is unset, (2) makes sure the SAC reference time ("sac_ref_time") also does not change event if "b" is unset, and (3) allows "b" to become set when writing from a Trace object.

krischer: I am a bit too lazy to delve into this right now but I'll do it first thing week. Until then, here is the commit and the issue that led to that test: #107 and 450de3cdd1ac976bcfaec08b84308bae57ef8df3

Cheers!

megies: This was originally brought up by @claudiodsf, maybe he can comment on this?

jkmacc-LANL: Will he see your mention of this?

megies: I've sent him an email..

megies:
Strange how appveyor can pass when there's obviously still some problems in the tests..
E.g. [here](https://travis-ci.org/obspy/obspy/jobs/79226661#L740) SAC's empty field values should be mapped to empty strings.

Not many fails though, looks like. Almost getting there? :)

obspy/io/sac/tests/test_core.py
claudiodsf: Hello, sorry for late reply.

I don't remember having written that `test_undefinedB`. In any case I agree with @jkmacc-LANL on that SAC reference time should not be changed, at least not without recomputing al the time-relative header fields (B, A, O, T0-T9).

Concerning "B", I would suggest to always write it to disk, i.e. if a user unsets B, this is recomputed and written to disk on writing a Trace object.

jkmacc-LANL: Thanks, Claudio.  I've made the change these changes in commit 39ccc896c047f824977cec242098aa24e34ee982 .  "B" is always written (because a Trace always knows its first sample time), and the reference time, if set, is not changed.  The current `test_undefinedB` now reflects these expectations.

jkmacc-LANL:
@megies It's a lot of work passing these tests:-)  Clearly, though, I still don't know the difference between a `bytes` and a `str`.

megies:
I think the biggest problem we have with the Python2/3 change regarding strings is the additional layer of complexity to be able to support both using `future`..

megies:
Congratulations. :octocat: 
Looks like we can likely do final review and include this in the next release. :)

jkmacc-LANL:
:sweat_smile: Thanks to you, Lion, Elliot, and Claudio for the help!  I would still like to complete the suggestions everyone made, starting with sphinx docstrings, I guess.

megies:
That's the spirit. ;-)

jkmacc-LANL:
I reorganized a bit, which hid lots of still-relevant comments behind "outdated diffs".  I've collected them here, and I'll try to check them off as I go.

1. [ ] [try/finally](https://github.com/obspy/obspy/pull/1092#discussion-diff-34327165)
2. [ ] [try/finally](https://github.com/obspy/obspy/pull/1092#discussion-diff-34327793)
3. [ ] [strip newlines](https://github.com/obspy/obspy/pull/1092#discussion-diff-34327903)
4. [ ] [open file after checks](https://github.com/obspy/obspy/pull/1092#discussion-diff-34328262)
5. [ ]  [deal with file names and BytesIO objects](https://github.com/obspy/obspy/pull/1092#discussion-diff-34370244)
6. [ ] [maybe remove 'rb+' requirement](https://github.com/obspy/obspy/pull/1092#discussion-diff-34371319)
7. [ ]  [replace assert statement](https://github.com/obspy/obspy/pull/1092#discussion-diff-34371403)

obspy/io/sac/sacpz.py
jkmacc-LANL: The file IO stuff is integrated into a lot of the logic, so it would be a pretty large `try:` block.  That might outweigh the benefit of using `try/finally`.

QuLogic:
Any chance of adding more coal to that :steam_locomotive:?

jkmacc-LANL:
Sure, I'll try to re-stoke the :fire:

jkmacc-LANL:
Call it good for now?

obspy/io/sac/pysac/header.py
```diff
@@ -0,0 +1,308 @@
...
+
+HEADER_DOCSTRING = """
+============ ==== ===============================================
+Field Name   Type Description
+============ ==== ===============================================
+""" + HDR_DESC + \
+        "\n============ ==== ==============================================="
+
+# Module documentation string
+__doc__ = MODULE_DOCSTRING + HEADER_DOCSTRING
```
megies: This should be used in the `__init__.py` to be better visible to the user.. such that it appears on the API front page for SAC.

megies:
Some comments, mostly on general things (not using SAC, so didn't look into the new implementation in detail, I have to admit).

Can we please get rid of that `LICENSE.txt`? The note on LGPL is superfluous, as ObsPy in general is LGPL. The copyright/author notice mentioning LANL can of course stay but I'd prefer to have it as a `README.txt`, either a new one in `sac/pysac` or merged with `sac/README.txt`. That's were we put all the attributions across all the submodules whenever there's something that should be mentioned.

Also, on the structure, kind of the same comment as https://github.com/obspy/obspy/pull/1125#issuecomment-147672326. Do you feel like we need the separation of this new implementation by means of a subfolder (`/sac/pysac/*.py`)? I mean.. ObsPy is python per definition and everything in `obspy.sac` always used to be Python code.. Can we not just move all of these Python files in `/pysac` down one level into `sac`? Just reminding of the Python Zen.. `Flat is better than nested.` Or are you concerned that the authorship notice is not 100% clear on what it covers in that case?

One more thing, would be great to have a short guide on how to transition codes that used low-level functionality in `obspy.sac` that is gone now (I assume some codes will break? which is OK, since this is pretty low-level).. or at least a short Overview on how to use the low-level functionality (if it's not already there and I failed to see it..) see e.g. https://github.com/obspy/obspy/wiki/Migration-to-ObsPy-0.9.0 which was a (super-nice, @krischer :)) guide for 0.9 -> 0.10 transition..

jkmacc-LANL:
@megies Certainly, I can move the `LICENSE.txt` into a `pysac/README.txt` to avoid any license confusion.  Regarding the structure, I completely agree with your point, and I originally organized the files the way you suggested.  However, the nested structure solves two issues for me, both related to the mechanisms by which my employer open-sources software.  My hope was that, with a thoughtful `pysac/__init__.py`, imports in `sac/core.py` or anywhere else would always look flat.

I think a guide like the one you suggested is a great idea.  As an aside, those wiki pages tend to get lost as the number of pages grows and because of their flat organization.  Could you suggest a natural/prominent place to feature the link to this page once it's up?

megies:
> solves two issues for me, both related to the mechanisms by which my employer open-sources software.

OK then, if there is a valid reason to keep it nested, then that's totally fine. It can stay as is.

> Could you suggest a natural/prominent place to feature the link to this page once it's up?

The place to go is the submodule's API front page, i.e. `obspy/io/sac/__init__.py` which shows up as http://docs.obspy.org/master/packages/obspy.io.sac.html.

krischer:
Hmm...I would really like to flatten the structure. We spent a lot of effort to clean up the overall structure of ObsPy and we should try to keep it tidy and this additional nesting serves no practical purpose for ObsPy. And once it goes into a release we cannot really change it anymore...

That being said - if it really has to be in this submodule for whatever reason I'm still alright with it.

@jkmacc-LANL Weird question - I know we can legally do it but we don't want to get you into trouble: would it be ok if we move it after the merge? Then you have already fulfilled the obligations to your employer and the code is LGPL so everyone can do as they please with at as long as it retains the license and attributions.

jkmacc-LANL:
@krischer That's a good solution, and it wouldn't be a problem for me, except that I also need to host it in my team's [org repos](https://github.com/LANL-Seismoacoustics/pysac) (haven't posted code there yet), and it would be easier for maintenance if the two looked as similar as possible.   In the future, I intended to just update `obspy/io/sac/pysac` and more-or-less copy the `pysac` directory into this other location.  I didn't want the two locations to diverge.

Other suggestions are welcome, as this is one is probably not optimal.  This seems like some kind of git submodule/subtree relationship, but I didn't want to get into that.

megies:
> I also need to host it in my team's org repos (haven't posted code there yet), and it would be easier for maintenance if the two looked as similar as possible.

That's indeed a case we didn't have yet. I can imagine a couple of solutions, none of them ideal, most of them relying on `git submodule` (which would be a *new*, especially in terms of installation issues):
 * maintain your code in `pysac` folder in your `LANL-Seismoacoustics/pysac` repo -- include it as `git submodule` at `obspy/io/sac/pysac` (it still has obspy imports in it, so this feels kind of half-hearted)
 * move all of `obspy/io/sac` into your repo (but the plugins in `setup.py` would still need to be defined in obspy main repo in that case -- so that's also a bit ugly, although those plugin definitions can be kept fixed)
 * ... ugly syncing across repositories with lots of rebasing

QuLogic:
Well, there is a fourth option: release _all_ of ObsPy as a "friendly" fork, which never diverges except for some small sac bits, but they're always merged upstream. I don't know the constraints though, so that might not be acceptable.

jkmacc-LANL:
Well, I originally never intended to develop outside of `io/sac/pysac`.  Do you foresee issues with occasionally copying updated release versions over into that other repo?  Everything I read about submodules makes them sound like a death by a thousand cuts.

megies:
> Do you foresee issues with occasionally copying updated release versions over into that other repo?

Nope, not at all. That's easiest for *us*, actually, since we don't have to do anything.

jkmacc-LANL:
No, it isn't important the the folder structure looks the same between the projects.  I was just thinking it would be easier to copy things over if they looked the same.  Also, if `io.sac.core` always just imported from `pysac` instead of `sactrace`, or whatever, it would insulate ObsPy from any internal changes to PySAC.

megies:
> No, it isn't important that the the folder structure looks the same between the repos.

Makes it easier..

> I was just thinking it would be easier to copy things over if they looked the same.

I agree. But.. on the other hand, it's 4 files you need/want to copy. Should not be too hard too maintain, right? Like I said, files/filenames won't change a lot over time..
(unless another SAC-rewrite hero jumps in ;-)

> Also, if `io.sac.core` always just imported from `pysac` instead of `sactrace`, or whatever, it would insulate ObsPy from any internal changes to PySAC.

Imports don't care if it's from a file in a subfolder or in a file in the same folder.

If future changes in the code you contribute break the main `io.sac.core` functionality, that is plugged in to `read()`, that's a problem regardless of whether the code is in a subfolder or not..
Routines that should not be relied on to stay the same API-wise (or stay there at all) should be made private (i.e. prefixed with an underscore), they can change whatever they like as long as the main functionality in `io.sac.core` continues to do the job.

jkmacc-LANL:
I also realize that this exchange itself has taken longer than moving the files, but we'd also have two `README`'s and two `__init__.py`'s to combine, both of which contain slightly different things.  I agree that flat is better than nested, but I tend to think this is for things that users are intended to *use*.  ObsPy users are actually discourage from using anything under `io/sac`.  Just another take on it.

If you have a suggestion about the `README`s and `__init__.py`'s, I'll make the change.

megies:
> I also realize that this exchange itself has taken longer than moving the files,

Well, yes, sorry about the hassle, really.
Let me try to justify why we're putting that much consideration into this seemingly irrelevant detail..

Whatever decision on the structure is taken now, and makes it into the next release, we're stuck with it in the future and can not change it without either..
 * breaking peoples' codes (unhappy users), or
 * going to *major* efforts, implementing deprecation warnings, rerouting of imports, etc.. (unhappy developers, partially unhappy users -- they still have to adapt their codes at some point)

See #999 for a current example for the amount of work that's involved when
 * *a)* moving around things in the submodule structure and (e.g. `obspy.mseed` -> `obspy.io.mseed`)
 * *b)* renaming functions etc. to match our code style (e.g. `getWaveforms()` -> `get_waveforms()`)

*a)* is just coming from the fact that we have such a huge amount of new submodules, so we decided we need to reorganize. It wasn't really foreseeable some years ago. *b)* however could have been avoided had we always been rigorous with newly added code.

---

Anyway, @krischer, @QuLogic, what do you say, should we just keep the `pysac` folder to make the separation easier?

jkmacc-LANL:
I wasn't going to add any more to this PR, but I don't know what "This branch has conflicts that must be resolved" means.

megies:
There's a merge conflict with master, probably just some trivial thing like the changelog. Basically this branch needs to be `git rebase master`ed. Let me know if you're not feeling comfortable with doing the merging/rebasing, I can do it in that case (would need push rights to your fork then).

obspy/io/sac/sacpz.py
megies: > so it would be a pretty large try: block. That might outweigh the benefit of using try/finally.

I'd probably prefer the long block, over an potentially unclosed file I think..

jkmacc-LANL:
`rebase`-ing may be a bit over my pay grade :smile:  I can probably manage to figure out how to give you push rights, though.  ...OK, I just added you as a collaborator.  I'll remove you after this is all done, though, because I'll likely keep this fork around.

megies:
Did a rebase and will force-push right now. I pushed the current state of this PR as a a backup branch to your fork called [sac-rewrite_bkp_before_rebase](https://github.com/jkmacc-LANL/obspy/commits/sac-rewrite_bkp_before_rebase), just in case.

megies:
Ok, force-pushed. You can resync with your fork by e.g. doing the following. Make sure that `git diff` is showing nothing (or save any local diffs from your local `sac-rewrite` branch somewhere outside the repo) because local diffs will be overwritten.. :skull: 

```bash
git checkout sac-rewrite
git fetch --all
git reset --hard REMOTE_ALIAS_OF_YOUR_FORK/sac-rewrite   # if unsure, find out the alias of "jkmacc-LANL/obspy" doing a "git remote -vv"
```

megies:
Inadvertently smuggled in two unrelated commits, just corrected this and force-pushed again..

jkmacc-LANL:
Thanks, Tobias.  I'll just sync my local with this and look for the eventual merge and PR close.  :metal:

QuLogic:
There are a couple comments about that need to be addressed or just okayed, but it should be good to :ship: after that.

krischer:
:+1: Also look good to me! :-)

obspy/io/sac/util.py
QuLogic: Did you check this?

jkmacc-LANL: I did, but forgot to follow up- sorry.  Yes, unset headers return -12345 when `debug_headers=True`.  If `False`, they just don't show up.  There are some [tests](https://github.com/jkmacc-LANL/obspy/blob/master/obspy/io/sac/tests/test_core.py#L424) that ensure this, too.

krischer:
This was a long one. Thanks a lot for this!

claudiodsf:
:+1: Congrats on merging!

jkmacc-LANL:
Thanks, everyone!

megies:
@jkmacc-LANL, we need to replace some numpy functions not present in older numpy versions, see #1183.

jkmacc-LANL:
:confused: Is `np.fft.rfftfreq` in `io.sac` ?  That was the only function I saw mentioned in #1183, and I don't think I use it.

QuLogic:
It looks more like `np.full` is missing.

megies:
Sorry for not being clear enough, the fails can be seen in the Travis results of the PR I linked: https://travis-ci.org/obspy/obspy/jobs/92129050#L738

It's indeed mostly `np.full` calls.. (new in numpy 1.8 http://docs.scipy.org/doc/numpy/release.html#id17)

jkmacc-LANL:
@megies It looks like you took care of this, correct?  If so, thanks, and I appreciate the notification.

megies:
> @megies It looks like you took care of this, correct? If so, thanks, and I appreciate the notification.

Yes, was no big deal anyway.

PR #1093, opened 2015-07-10
Add support for time values with sub-second precision

3rdcycle:
This fixes the issue that was reported on obspy-users on 8 June 2015.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -29,6 +29,8 @@ master:
    * Upgrade to libmseed 2.16
  - obspy.io.shapefile:
    * New module for ESRI shapefile write support (see #1066)
+ - obspy.io.zmap
+   * Add support for time values with sub-second precision
  - obspy.signal:
    * Switch to second-order sections for filters; backported from SciPy 0.16.0
      (see #1028)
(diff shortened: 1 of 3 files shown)


CHANGELOG.txt
```diff
@@ -29,6 +29,8 @@ master:
    * Upgrade to libmseed 2.16
  - obspy.io.shapefile:
    * New module for ESRI shapefile write support (see #1066)
+ - obspy.io.zmap
+   * Add support for time values with sub-second precision
```
krischer: Can you add the issue number for future reference?

obspy/io/zmap/core.py
```diff
@@ -137,6 +137,8 @@ def _serialize(self, catalog):
             origin = ev.preferred_origin()
             if origin:
                 dec_year = self._decimal_year(origin.time)
+                dec_second = origin.time.second + \
+                    origin.time.microsecond / float(1e6)
```
krischer: super minor nitpick: no need to wrap this in `float()` as the exponential notation will always produce floating point numbers.

krischer:
Looks good to me. Thanks for that!

FYI for future bug fixes: can you please base them on the `releases` branch? Then they will also flow into the next minor release.  For this one we'll just cherry pick it onto the releases branch as the changes are so minor.

3rdcycle:
Ok, will use the releases branch next time. I don't think there's an issue number I could refer to since this has only been reported on the mailing list AFAIK.

krischer:
This issue has a number: #1093 ;-)

krischer:
Thanks. I'll take care that the changes also end up in the releases branch.

PR #1067, opened 2015-07-20
add kml output for Catalog/Inventory

megies:
Refactoring this routine that builds a kml file using lxml:
https://github.com/obspy/obspy/blob/master/obspy/clients/seishub/client.py#L1029

### DIFF
--- a/obspy/io/kml/core.py
+++ b/obspy/io/kml/core.py
@@ -22,17 +22,17 @@ from obspy import UTCDateTime
 
 def inventory_to_kml_string(
         inventory,
-        icon_url="http://maps.google.com/mapfiles/kml/shapes/triangle.png",
+        icon_url="https://maps.google.com/mapfiles/kml/shapes/triangle.png",
         icon_size=1.5, label_size=1.0, cmap="Paired", encoding="UTF-8",
         timespans=True, strip_far_future_end_times=True):
     """
-    Convert an :class:`~obspy.core.inventory.inventory.Inventory` to a kml
+    Convert an :class:`~obspy.core.inventory.inventory.Inventory` to a KML
     string representation.
 
     :type inventory: :class:`~obspy.core.inventory.inventory.Inventory`
     :param inventory: Input station metadata.
     :type icon_url: str
-    :param icon_url: Internet URL of icon to use for station (e.g. png image).
+    :param icon_url: Internet URL of icon to use for station (e.g. PNG image).
     :type icon_size: float
     :param icon_size: Icon size.
     :type label_size: float
@@ -41,7 +41,7 @@ def inventory_to_kml_string(
     :param encoding: Encoding used for XML string.
     :type timespans: bool
     :param timespans: Whether to add timespan information to the single station
-        elements in the kml or not. If timespans are used, the displayed
+        elements in the KML or not. If timespans are used, the displayed
         information in e.g. Google Earth will represent a snapshot in time,
         such that using the time slider different states of the inventory in
         time can be visualized. If timespans are not used, any station active
@@ -50,7 +50,7 @@ def inventory_to_kml_string(
     :param strip_far_future_end_times: Leave out likely fictitious end times of
         stations (more than twenty years after current time). Far future end
         times may produce time sliders with bad overall time span in third
-        party applications viewing the kml file.
+        party applications viewing the KML file.
     :rtype: byte string
     :return: Encoded byte string containing KML information of the station
         metadata.
@@ -106,7 +106,7 @@ def inventory_to_kml_string(
             SubElement(placemark, "color").text = color
             if sta.longitude is not None and sta.latitude is not None:
                 point = SubElement(placemark, "Point")
-                SubElement(point, "coordinates").text = "%.10f,%.10f,0" % \
+                SubElement(point, "coordinates").text = "%.6f,%.6f,0" % \
                     (sta.longitude, sta.latitude)
 
             SubElement(placemark, "description").text = str(sta)
@@ -141,17 +141,17 @@ def inventory_to_kml_string(
 
 def catalog_to_kml_string(
         catalog,
-        icon_url="http://maps.google.com/mapfiles/kml/shapes/earthquake.png",
+        icon_url="https://maps.google.com/mapfiles/kml/shapes/earthquake.png",
         label_func=None, icon_size_func=None, encoding="UTF-8",
         timestamps=True):
     """
-    Convert an :class:`~obspy.core.event.Catalog` to a kml string
+    Convert a :class:`~obspy.core.event.Catalog` to a KML string
     representation.
 
     :type catalog: :class:`~obspy.core.event.Catalog`
     :param catalog: Input catalog data.
     :type icon_url: str
-    :param icon_url: Internet URL of icon to use for events (e.g. png image).
+    :param icon_url: Internet URL of icon to use for events (e.g. PNG image).
     :type label_func: func
     :type label_func: Custom function to use for determining each event's
         label. User provided function is supposed to take an
@@ -159,13 +159,14 @@ def catalog_to_kml_string(
         empty labels use `label_func=lambda x: ""`.
     :type icon_size_func: func
     :type icon_size_func: Custom function to use for determining each
-        event's icon size. User provided function is supposed to take an
-        :class:`~obspy.core.event.Event` object as single argument.
+        event's icon size. User provided function should take an
+        :class:`~obspy.core.event.Event` object as single argument and return a
+        float.
     :type encoding: str
     :param encoding: Encoding used for XML string.
     :type timestamps: bool
     :param timestamps: Whether to add timestamp information to the event
-        elements in the kml or not. If timestamps are used, the displayed
+        elements in the KML or not. If timestamps are used, the displayed
         information in e.g. Google Earth will represent a snapshot in time,
         such that using the time slider different states of the catalog in time
         can be visualized. If timespans are not used, any event happening at
@@ -255,7 +256,7 @@ def catalog_to_kml_string(
         if origin:
             if origin.longitude is not None and origin.latitude is not None:
                 point = SubElement(placemark, "Point")
-                SubElement(point, "coordinates").text = "%.10f,%.10f,0" % \
+                SubElement(point, "coordinates").text = "%.6f,%.6f,0" % \
                     (origin.longitude, origin.latitude)
 
         SubElement(placemark, "description").text = str(event)
@@ -280,7 +281,7 @@ def _rgba_tuple_to_kml_color_code(rgba):
     except:
         r, g, b = rgba
         a = 1.0
-    return "".join([hex(int(x * 255))[-2:] for x in (a, b, g, r)])
+    return "".join(["%02x" % int(x * 255) for x in (a, b, g, r)])
 
 
 def _get_event_timestamp(event):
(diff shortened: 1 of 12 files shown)


megies:
Our example catalog written to kml format as seen in Google Earth..

screenshot from 2015-07-20 17 53 27

obspy/io/kml/core.py
```diff
@@ -0,0 +1,150 @@
...
+    :type label_func: func
+    :type label_func: Custom function to use for determining each event's
+        label. User provided function is supposed to take an
+        :class:`~obspy.core.event.Event` object as single argument, e.g. for
+        empty labels use `label_func=lambda x: ""`.
+    :type icon_size_func_func: func
+    :type icon_size_func_func: Custom function to use for determining each
+        event's icon size. User provided function is supposed to take an
+        :class:`~obspy.core.event.Event` object as single argument.
+    :rtype: str
```
QuLogic: Are you sure this is not `bytes`? Since lxml prefers to parse `bytes` and determine encoding from the XML, I'd assume it produces `bytes` as well.

megies:
And here's our example inventory data: 
screenshot from 2015-07-21 17 28 00

obspy/io/kml/core.py
megies: Well.. actually it's `str` in Python2 and `bytes` in Python3, both encoded though.

megies:
Here's two example movies showing the use of timestamps of events and timespans of stations when viewing the output kmls (TA stations, Mag 5+ events around US).

https://www.geophysik.uni-muenchen.de/~megies/TA_kml.ogv
https://www.geophysik.uni-muenchen.de/~megies/US_mag5.ogv

QuLogic:
That looks pretty cool. Those are some long-lasting earthquakes, though.

megies:
Well if I'd just show em for the rupture duration.. ;)

This time slider thing has a start/endtime, I think i had it at around three months for the quakes vid for them to last some time in the display.

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+from obspy import UTCDateTime
+
+
+def inventory_to_kml_string(
+        inventory,
+        icon_url="http://maps.google.com/mapfiles/kml/shapes/triangle.png",
+        icon_size=1.5, label_size=1.0, cmap="Paired", encoding="UTF-8",
+        timespans=True, strip_far_future_end_times=True):
+    """
+    Convert an :class:`~obspy.core.inventory.inventory.Inventory` to a kml
```
QuLogic: KML

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+    string representation.
+
+    :type inventory: :class:`~obspy.core.inventory.inventory.Inventory`
+    :param inventory: Input station metadata.
+    :type icon_url: str
+    :param icon_url: Internet URL of icon to use for station (e.g. png image).
```
QuLogic: PNG

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+    :type icon_size: float
+    :param icon_size: Icon size.
+    :type label_size: float
+    :param label_size: Label size.
+    :type encoding: str
+    :param encoding: Encoding used for XML string.
+    :type timespans: bool
+    :param timespans: Whether to add timespan information to the single station
+        elements in the kml or not. If timespans are used, the displayed
```
QuLogic: KML

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+        information in e.g. Google Earth will represent a snapshot in time,
+        such that using the time slider different states of the inventory in
+        time can be visualized. If timespans are not used, any station active
+        at any point in time is always shown.
+    :type strip_far_future_end_times: bool
+    :param strip_far_future_end_times: Leave out likely fictitious end times of
+        stations (more than twenty years after current time). Far future end
+        times may produce time sliders with bad overall time span in third
+        party applications viewing the kml file.
```
QuLogic: KML

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # add one marker per station code
+        for sta in net:
+            placemark = SubElement(folder, "Placemark")
+            SubElement(placemark, "name").text = ".".join((net.code, sta.code))
+            SubElement(placemark, "styleUrl").text = "#station_%i" % i
+            SubElement(placemark, "color").text = color
+            if sta.longitude is not None and sta.latitude is not None:
+                point = SubElement(placemark, "Point")
+                SubElement(point, "coordinates").text = "%.10f,%.10f,0" % \
```
QuLogic: 10 digits seems a bit much. IIRC, IRIS will only give you 6 digits.

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+
+from lxml.etree import Element, SubElement, tostring
+from matplotlib.cm import get_cmap
+
+from obspy import UTCDateTime
+
+
+def inventory_to_kml_string(
+        inventory,
+        icon_url="http://maps.google.com/mapfiles/kml/shapes/triangle.png",
```
QuLogic: Please use `https`; Google auto-redirects there anyway.

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+                        SubElement(timespan, "end").text = str(end)
+
+    # generate and return KML string
+    return tostring(kml, pretty_print=True, xml_declaration=True,
+                    encoding=encoding)
+
+
+def catalog_to_kml_string(
+        catalog,
+        icon_url="http://maps.google.com/mapfiles/kml/shapes/earthquake.png",
```
QuLogic: Please use `https`.

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+        label_func=None, icon_size_func=None, encoding="UTF-8",
+        timestamps=True):
+    """
+    Convert an :class:`~obspy.core.event.Catalog` to a kml string
```
QuLogic: an -> a. kml -> KML

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+    representation.
+
+    :type catalog: :class:`~obspy.core.event.Catalog`
+    :param catalog: Input catalog data.
+    :type icon_url: str
+    :param icon_url: Internet URL of icon to use for events (e.g. png image).
```
QuLogic: PNG

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+    :type label_func: func
+    :type label_func: Custom function to use for determining each event's
+        label. User provided function is supposed to take an
+        :class:`~obspy.core.event.Event` object as single argument, e.g. for
+        empty labels use `label_func=lambda x: ""`.
+    :type icon_size_func: func
+    :type icon_size_func: Custom function to use for determining each
+        event's icon size. User provided function is supposed to take an
```
QuLogic: is supposed to -> should.

Also, return value should be?

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+        :class:`~obspy.core.event.Event` object as single argument.
+    :type encoding: str
+    :param encoding: Encoding used for XML string.
+    :type timestamps: bool
+    :param timestamps: Whether to add timestamp information to the event
+        elements in the kml or not. If timestamps are used, the displayed
```
QuLogic: KML

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+        SubElement(placemark, "styleUrl").text = "#earthquake"
+        style = SubElement(placemark, "Style")
+        icon_style = SubElement(style, "IconStyle")
+        liststyle = SubElement(style, "ListStyle")
+        SubElement(liststyle, "maxSnippetLines").text = "5"
+        SubElement(icon_style, "scale").text = str(icon_size_func(event))
+        if origin:
+            if origin.longitude is not None and origin.latitude is not None:
+                point = SubElement(placemark, "Point")
+                SubElement(point, "coordinates").text = "%.10f,%.10f,0" % \
```
QuLogic: See other comment about # of digits.

obspy/io/kml/core.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    Convert tuple of (red, green, blue, alpha) float values (0.0-1.0) to KML
+    hex color code string "aabbggrr".
+    """
+    try:
+        r, g, b, a = rgba
+    except:
+        r, g, b = rgba
+        a = 1.0
+    return "".join([hex(int(x * 255))[-2:] for x in (a, b, g, r)])
```
QuLogic: To avoid the trimming, you could do `'%02x' % (...)`.

megies:
Only a basic test missing..

QuLogic:
Will need a rebase now, looks like.

megies:
> Will need a rebase now, looks like.

Just the usual, CHANGELOG needing a trivial manual merge. Rebased, force-pushed.

QuLogic:
Don't forget to add the test! And a couple fixes for the others.

megies:
Added test..

obspy/io/kml/tests/data/catalog.kml
```diff
@@ -0,0 +1,109 @@
...
+            <maxSnippetLines>5</maxSnippetLines>
+          </ListStyle>
+        </Style>
+        <Point>
+          <coordinates>79.689000,41.818000,0</coordinates>
+        </Point>
+        <description>Event:	2012-04-04T14:21:42.300000Z | +41.818,  +79.689 | 4.4 mb | manual
+
+	        resource_id: ResourceIdentifier(id="quakeml:eu.emsc/event/20120404_0000041")
+	         event_type: u'not reported'
```
QuLogic: At some point, we should look into removing these `u`s.

QuLogic: Oh, and I guess by "some point", I mean now, since it's breaking Python 3 tests.

QuLogic:
Looks fine except Python 3 is broken right now.

obspy/io/kml/tests/data/catalog.kml
megies: Mhm, should be fixed now, although that's a `core` issue and only pops up here because it wasn't covered by tests yet..

megies:
Only thing that's left is `str()` on numeric types having different length on Py2/3. Not sure how to best work around this in the test.. any good ideas?

QuLogic:
If you can guarantee that there are only ASCII characters in that place, then `native_str` will work fine. For numeric types, we ended up using an explicit length for Cartopy (i.e., `'{:.16f}'.format(0.1234)`).

megies:
> If you can guarantee that there are only ASCII characters in that place, then native_str will work fine. 

Currently this occurred with `event_type` and these are pre-defined with only ASCII characters. But I guess this is might also affect other fields with user determined content.. anyway, not strictly related to this issue, so has to be fixed in `master` if anybody runs into problems with it, I guess.

> For numeric types, we ended up using an explicit length for Cartopy (i.e., '{:.16f}'.format(0.1234)).

Hmm, yeah. Kinda wanted to avoid manually setting this, but so what I guess..

PR #1097, opened 2015-07-23
method to get simple (old style) PAZ dictionary from Response objects

megies:


obspy/core/inventory/response.py
```diff
@@ -1324,6 +1324,31 @@ def get_sacpz(self):
...
+        Get simple dictionary representation of poles, zeros and overall
+        sensitivity.
+
+        >>> from obspy import UTCDateTime, read_inventory
+        >>> t = UTCDateTime(2015, 1, 1)
+        >>> inv = read_inventory()
+        >>> resp = inv.get_response("BW.RJOB..EHZ", t)
+        >>> paz = resp.get_paz_dict()
+        >>> print(sorted(paz.keys()))
+        [u'poles', u'sensitivity', u'zeros']
```
QuLogic: I think you want to try and avoid these `u`s.

megies: I'm not doing anything to these, isn't that rather a problem on a lower level.. inside the inventory types?

obspy/core/inventory/response.py
```diff
@@ -1324,6 +1324,31 @@ def get_sacpz(self):
...
+        >>> print(sorted(paz.items()))  # doctest: +NORMALIZE_WHITESPACE
+        [(u'poles', [(-0.037004+0.037016j), (-0.037004-0.037016j),
+                     (-251.33+0j), (-131.04-467.29j), (-131.04+467.29j)]),
+         (u'sensitivity', 2516800000.0),
+         (u'zsseros', [0j, 0j])]
```
QuLogic: Also, what?

megies: Whooops, that's a relic of checking the doctest..

megies:
Two minor corrections, squashed and force-pushed..

CTrabant:
Keep in mind that poles and zeros along with the total channel sensitivity are insufficient to use for general instrument correction.  For one, the units need to be known.  For two, when calculating the response and performing spectral division the A0 (normalization factor) needs to be included; this is why the CONSTANT value in SAC P&Zs is a product of the total sensitivity and A0.

I'm not quite sure what the goal is here, so perhaps these comments are not appropriate.  Use as you wish.

PR #1101, opened 2015-07-30
Lanczos Resampling/Interpolation

krischer:
This PR adds Lanczos interpolation to ObsPy as has already been discussed a couple of times. Its an extended version of what we already had in Instaseis rewritten in C. I hooked it into `Stream.interpolate`/`Trace.interpolate` as it is a kind of interpolation method and it fits the existing API very well.

It is almost always better then the existing interpolation methods and should come in very handy for changing the sampling rate. Results are pretty nice:

```python
import matplotlib.pylab as plt
import numpy as np
import obspy

data = np.sin(np.linspace(0, 200 * np.pi, 250))

tr = obspy.Trace(data=data)

tr2 = tr.copy()
tr2.interpolate(method="lanczos", a=25, sampling_rate=100)

plt.figure(figsize=(18, 8))
plt.plot(tr.times(), tr.data, "o", label="samples")
plt.plot(tr2.times(), tr2.data, label="resampled")
plt.plot(np.linspace(0, 249, 10000),
         np.sin(np.linspace(0, 200 * np.pi, 10000)), "--",
         label="original", color="k")
plt.legend()
plt.xlim(30, 40)
plt.show()
```

screen shot 2015-07-31 at 00 40 55

I put quite some work into the documentation - pages of interest:

* http://www.geophysik.uni-muenchen.de/~krischer/doc_lanczos/packages/autogen/obspy.signal.interpolation.lanczos_interpolation.html
* http://www.geophysik.uni-muenchen.de/~krischer/doc_lanczos/packages/autogen/obspy.signal.interpolation.plot_lanczos_windows.html
* http://www.geophysik.uni-muenchen.de/~krischer/doc_lanczos/packages/autogen/obspy.core.trace.Trace.interpolate.html

It currently supports three different windows to taper the sinc function which represent different trade-offs between accuracy and "wiggliness". If somebody feels we need more please let me know.

obspy-signal-interpolation-plot_lanczos_windows-1

Furthermore the `interpolate()` method now has an additional argument: `time_shift` - a cheap way to shift seismograms by subsamples. Not as good as a Fourier domain shift but still pretty useful in many cases.

There is still some optimization potential in the C code which could speed it up so if anyone wants to have a go: Please feel free :-)

@martinvandriel Just pinging you to make you aware of this.

Should be ready for review!

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1815,6 +1815,70 @@ class TraceTestCase(unittest.TestCase):
             self.assertEqual(int_tr.stats.delta, 2.0)
             self.assertLessEqual(int_tr.stats.endtime, org_tr.stats.endtime)
 
+    def test_interpolation_time_shift(self):
+        """
+        Tests the time shift of the interpolation.
+        """
+        tr = read()[0]
+        tr.stats.sampling_rate = 1.0
+        tr.data = tr.data[:500]
+        tr.interpolate(method="lanczos", sampling_rate=10.0, a=20)
+        tr.stats.sampling_rate = 1.0
+        tr.data = tr.data[:500]
+        tr.stats.starttime = UTCDateTime(0)
+
+        org_tr = tr.copy()
+
+        # Now this does not do much for now but actually just shifts the
+        # samples.
+        tr.interpolate(method="lanczos", sampling_rate=1.0, a=1,
+                       time_shift=0.2)
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime + 0.2)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime + 0.2)
+        np.testing.assert_allclose(tr.data, org_tr.data, atol=1E-9)
+
+        tr.interpolate(method="lanczos", sampling_rate=1.0, a=1,
+                       time_shift=0.4)
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime + 0.6)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime + 0.6)
+        np.testing.assert_allclose(tr.data, org_tr.data, atol=1E-9)
+
+        tr.interpolate(method="lanczos", sampling_rate=1.0, a=1,
+                       time_shift=-0.6)
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime)
+        np.testing.assert_allclose(tr.data, org_tr.data, atol=1E-9)
+
+        # This becomes more interesting when also fixing the sample
+        # positions. Then one can shift by subsample accuracy while leaving
+        # the sample positions intact. Note that there naturally are some
+        # boundary effects and as the interpolation method does not deal
+        # with any kind of extrapolation you will loose the first or last
+        # samples.
+        # This is a fairly extreme example but of course there are errors
+        # when doing an interpolation - an shift using an FFT is more accurate.
+        tr.interpolate(method="lanczos", sampling_rate=1.0, a=50,
+                       starttime=tr.stats.starttime + tr.stats.delta,
+                       time_shift=0.2)
+        # The sample point did not change but we lost the first sample,
+        # as we shifted towards the future.
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime + 1.0)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime)
+        # The data naturally also changed.
+        self.assertRaises(AssertionError, np.testing.assert_allclose,
+                          tr.data, org_tr.data[1:], atol=1E-9)
+        # Shift back. This time we will loose the last sample.
+        tr.interpolate(method="lanczos", sampling_rate=1.0, a=50,
+                       starttime=tr.stats.starttime,
+                       time_shift=-0.2)
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime + 1.0)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime - 1.0)
+        # But the data (aside from edge effects - we are going forward and
+        # backwards again so they go twice as far!) should now again be the
+        # same as we started out with.
+        np.testing.assert_allclose(
+            tr.data[100:-100], org_tr.data[101:-101], atol=1E-9, rtol=1E-4)
+
     def test_interpolation_arguments(self):
         """
         Test case for the interpolation arguments.
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1853,10 +1853,10 @@ class TraceTestCase(unittest.TestCase):
         # positions. Then one can shift by subsample accuracy while leaving
         # the sample positions intact. Note that there naturally are some
         # boundary effects and as the interpolation method does not deal
-        # with any kind of extrapolation you will loose the first or last
+        # with any kind of extrapolation you will lose the first or last
         # samples.
         # This is a fairly extreme example but of course there are errors
-        # when doing an interpolation - an shift using an FFT is more accurate.
+        # when doing an interpolation - a shift using an FFT is more accurate.
         tr.interpolate(method="lanczos", sampling_rate=1.0, a=50,
                        starttime=tr.stats.starttime + tr.stats.delta,
                        time_shift=0.2)
@@ -1867,7 +1867,7 @@ class TraceTestCase(unittest.TestCase):
         # The data naturally also changed.
         self.assertRaises(AssertionError, np.testing.assert_allclose,
                           tr.data, org_tr.data[1:], atol=1E-9)
-        # Shift back. This time we will loose the last sample.
+        # Shift back. This time we will lose the last sample.
         tr.interpolate(method="lanczos", sampling_rate=1.0, a=50,
                        starttime=tr.stats.starttime,
                        time_shift=-0.2)
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2171,7 +2171,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
     @raise_if_masked
     @_add_processing_info
     def interpolate(self, sampling_rate, method="weighted_average_slopes",
-                    starttime=None, npts=None):
+                    starttime=None, npts=None, *args, **kwargs):
         """
         Interpolate the data using various interpolation techniques.
 
@@ -2274,7 +2274,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
                                   dt)) + 1
         self.data = np.atleast_1d(func(np.require(self.data, dtype=np.float64),
                                        old_start, old_dt, starttime, dt, npts,
-                                       type=method))
+                                       type=method, *args, **kwargs))
         self.stats.starttime = UTCDateTime(starttime)
         self.stats.delta = dt
 
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2171,7 +2171,8 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
     @raise_if_masked
     @_add_processing_info
     def interpolate(self, sampling_rate, method="weighted_average_slopes",
-                    starttime=None, npts=None, *args, **kwargs):
+                    starttime=None, npts=None, time_shift=0.0,
+                    *args, **kwargs):
         """
         Interpolate the data using various interpolation techniques.
 
@@ -2214,6 +2215,11 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         :param npts: The new number of samples. Will be set to the best
             fitting  number to retain the current end time of the trace if
             not given.
+        :type time_shift: float
+        :param time_shift: Interpolation also can shift the data with
+            subsample accuracy. The time shift is always given in seconds. A
+            positive shift means the data is shifted towards the future,
+            e.g. a positive time delta.
 
         .. rubric:: _`Usage Examples`
 
@@ -2251,32 +2257,44 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             raise ValueError("The time step must be positive.")
         dt = 1.0 / sampling_rate
 
-        if isinstance(method, int) or method in ["linear", "nearest", "zero",
-                                                 "slinear", "quadratic",
-                                                 "cubic"]:
-            func = _get_function_from_entry_point('interpolate',
-                                                  'interpolate_1d')
-        else:
-            func = _get_function_from_entry_point('interpolate', method)
-        old_start = self.stats.starttime.timestamp
-        old_dt = self.stats.delta
+        # We just shift the old start time. The interpolation will take care
+        # of the rest.
+        if time_shift:
+            self.stats.starttime += time_shift
 
-        if starttime is not None:
-            try:
-                starttime = starttime.timestamp
-            except AttributeError:
-                pass
-        else:
-            starttime = self.stats.starttime.timestamp
+        try:
+            if isinstance(method, int) or \
+                    method in ["linear", "nearest", "zero", "slinear",
+                               "quadratic", "cubic"]:
+                func = _get_function_from_entry_point('interpolate',
+                                                      'interpolate_1d')
+            else:
+                func = _get_function_from_entry_point('interpolate', method)
+            old_start = self.stats.starttime.timestamp
+            old_dt = self.stats.delta
 
-        if npts is None:
-            npts = int(math.floor((self.stats.endtime.timestamp - starttime) /
-                                  dt)) + 1
-        self.data = np.atleast_1d(func(np.require(self.data, dtype=np.float64),
-                                       old_start, old_dt, starttime, dt, npts,
-                                       type=method, *args, **kwargs))
-        self.stats.starttime = UTCDateTime(starttime)
-        self.stats.delta = dt
+            if starttime is not None:
+                try:
+                    starttime = starttime.timestamp
+                except AttributeError:
+                    pass
+            else:
+                starttime = self.stats.starttime.timestamp
+            endtime = self.stats.endtime.timestamp
+            if npts is None:
+                npts = int(math.floor((endtime - starttime) / dt)) + 1
+
+            self.data = np.atleast_1d(func(
+                np.require(self.data, dtype=np.float64), old_start, old_dt,
+                starttime, dt, npts, type=method, *args, **kwargs))
+            self.stats.starttime = UTCDateTime(starttime)
+            self.stats.delta = dt
+        except:
+            # Revert the start time change if something went wrong.
+            if time_shift:
+                self.stats.starttime -= time_shift
+            # re-raise last exception.
+            raise
 
         return self
 
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2219,7 +2219,10 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         :param time_shift: Interpolation also can shift the data with
             subsample accuracy. The time shift is always given in seconds. A
             positive shift means the data is shifted towards the future,
-            e.g. a positive time delta.
+            e.g. a positive time delta. Please note that a time shift in
+            the Fourier domain is always more accurate then this. When using
+            Lanczos interpolation with large values of ``a`` and away from the
+            boundaries this is nonetheless pretty good.
 
         .. rubric:: _`Usage Examples`
 
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2176,8 +2176,9 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         """
         Interpolate the data using various interpolation techniques.
 
-        Be careful when downsampling data and make sure to apply a proper
-        anti-aliasing lowpass filter in case its necessary.
+        Be careful when downsampling data and make sure to apply an appropriate
+        anti-aliasing lowpass filter before interpolating in case it's
+        necessary.
 
         .. note::
 
@@ -2203,7 +2204,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         * ``"lanczos"``: This offers the highest quality interpolation and
           should be chosen whenever possible. It is only due to legacy
           reasons that this is not the default method. The one downside it
-          has that it can be fairly expensive. See the
+          has is that it can be fairly expensive. See the
           :func:`~obspy.signal.interpolation.lanczos_interpolation` function
           for more details.
         * ``"weighted_average_slopes"``: This is the interpolation method used
@@ -2211,7 +2212,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
           :func:`~obspy.signal.interpolation.weighted_average_slopes` for
           more details.
         * ``"slinear"``, ``"quadratic"`` and ``"cubic"``: spline interpolation
-          of first, second or third order
+          of first, second or third order.
         * ``"linear"``: Linear interpolation.
         * ``"nearest"``: Nearest neighbour interpolation.
         * ``"zero"``: Last encountered value interpolation.
@@ -2238,7 +2239,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             subsample accuracy. The time shift is always given in seconds. A
             positive shift means the data is shifted towards the future,
             e.g. a positive time delta. Please note that a time shift in
-            the Fourier domain is always more accurate then this. When using
+            the Fourier domain is always more accurate than this. When using
             Lanczos interpolation with large values of ``a`` and away from the
             boundaries this is nonetheless pretty good.
 
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2176,14 +2176,14 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         """
         Interpolate the data using various interpolation techniques.
 
-        No filter, antialiasing, ... is applied so make sure the data is
-        suitable for the operation to be performed.
+        Be careful when downsampling data and make sure to apply a proper
+        anti-aliasing lowpass filter in case its necessary.
 
         .. note::
 
             The :class:`~Trace` object has three different methods to change
             the sampling rate of its data: :meth:`~.resample`,
-            :meth:`~.decimate`, and :meth:`~.interpolate`
+            :meth:`~.decimate`, and :meth:`~.interpolate`.
 
             Make sure to choose the most appropriate one for the problem at
             hand.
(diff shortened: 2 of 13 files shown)


QuLogic:
Windows disagrees with it... You need to also `#define _USE_MATH_DEFINES`, I think.

obspy/core/trace.py
```diff
@@ -2251,32 +2283,44 @@ def interpolate(self, sampling_rate, method="weighted_average_slopes",
...
+            endtime = self.stats.endtime.timestamp
+            if npts is None:
+                npts = int(math.floor((endtime - starttime) / dt)) + 1
+
+            self.data = np.atleast_1d(func(
+                np.require(self.data, dtype=np.float64), old_start, old_dt,
+                starttime, dt, npts, type=method, *args, **kwargs))
+            self.stats.starttime = UTCDateTime(starttime)
+            self.stats.delta = dt
+        except:
```
QuLogic: That's a lot of code to put in a `try`; what exactly are you trying to catch here?

QuLogic:
Is there a reason you don't accept the window via SciPy's windowing functions? I see it's calculated as needed, but there's a double loop, so pre-calculating it might be an easy optimization.

obspy/core/trace.py
krischer: Just resetting the starttime in case something fails. Otherwise it will be a bit confusing if the operation fails and the Trace object is still modified.

Its a cheap trick to get the time-shift capability essentially for free. I also reraise the exception so it is not really caught just delayed a bit.

krischer:
> Is there a reason you don't accept the window via SciPy's windowing functions? I see it's calculated as needed, but there's a double loop, so pre-calculating it might be an easy optimization.

Initially I did not want to go down that road so that it could eventually also support irregularly sampled data. Then I realized that the sampling theorem does no longer really hold with irregularly sampled data so I am not sure if sinc based interpolation is a good idea for that kind of data in any case.

Now it should be possible to precalculate the windows which should also speed things up a bit.

krischer:
No wait...that actually cannot work. If one dt does not cleanly divide the other dt the windows are sampled at a new position each time so it cannot be precalculated.

QuLogic:
Ah, that's too bad; scipy supports quite a few different windows.

obspy/core/tests/test_trace.py
```diff
@@ -1815,6 +1815,70 @@ def test_interpolate(self):
...
+                       time_shift=-0.6)
+        self.assertEqual(tr.stats.starttime, org_tr.stats.starttime)
+        self.assertEqual(tr.stats.endtime, org_tr.stats.endtime)
+        np.testing.assert_allclose(tr.data, org_tr.data, atol=1E-9)
+
+        # This becomes more interesting when also fixing the sample
+        # positions. Then one can shift by subsample accuracy while leaving
+        # the sample positions intact. Note that there naturally are some
+        # boundary effects and as the interpolation method does not deal
+        # with any kind of extrapolation you will loose the first or last
```
QuLogic: lose

obspy/core/tests/test_trace.py
```diff
@@ -1815,6 +1815,70 @@ def test_interpolate(self):
...
+        # samples.
+        # This is a fairly extreme example but of course there are errors
+        # when doing an interpolation - an shift using an FFT is more accurate.
```
QuLogic: a shift

obspy/core/trace.py
```diff
@@ -2171,18 +2171,19 @@ def split(self):
...
-                    starttime=None, npts=None):
+                    starttime=None, npts=None, time_shift=0.0,
+                    *args, **kwargs):
         """
         Interpolate the data using various interpolation techniques.
 
-        No filter, antialiasing, ... is applied so make sure the data is
-        suitable for the operation to be performed.
+        Be careful when downsampling data and make sure to apply a proper
+        anti-aliasing lowpass filter in case its necessary.
```
QuLogic: it's

obspy/core/trace.py
```diff
@@ -2194,18 +2195,36 @@ def interpolate(self, sampling_rate, method="weighted_average_slopes",
...
 
+        .. rubric:: _`Interpolation Methods:`
+
+        The chosen method is crucial and we will elaborate a bit about the
+        choices here:
+
+        * ``"lanczos"``: This offers the highest quality interpolation and
+          should be chosen whenever possible. It is only due to legacy
+          reasons that this is not the default method. The one downside it
+          has that it can be fairly expensive. See the
```
QuLogic: is that

obspy/core/trace.py
```diff
@@ -2194,18 +2195,36 @@ def interpolate(self, sampling_rate, method="weighted_average_slopes",
...
+          :func:`~obspy.signal.interpolation.lanczos_interpolation` function
+          for more details.
+        * ``"weighted_average_slopes"``: This is the interpolation method used
+          by SAC. Refer to
+          :func:`~obspy.signal.interpolation.weighted_average_slopes` for
+          more details.
+        * ``"slinear"``, ``"quadratic"`` and ``"cubic"``: spline interpolation
+          of first, second or third order
```
QuLogic: Missing period.

obspy/core/trace.py
```diff
@@ -2214,9 +2233,22 @@ def interpolate(self, sampling_rate, method="weighted_average_slopes",
         :param npts: The new number of samples. Will be set to the best
             fitting  number to retain the current end time of the trace if
             not given.
+        :type time_shift: float
+        :param time_shift: Interpolation can also shift the data with
+            subsample accuracy. The time shift is always given in seconds. A
+            positive shift means the data is shifted towards the future,
+            e.g. a positive time delta. Please note that a time shift in
+            the Fourier domain is always more accurate then this. When using
```
QuLogic: than

obspy/signal/headers.py
```diff
@@ -114,6 +114,43 @@
...
+    C.c_double,
+    # len_in
+    C.c_int,
+    # len_out,
+    C.c_int,
+    # a,
+    C.c_int,
+    # window
+    C.c_int]
+clibsignal.lanczos_resample.restype = C.c_void_p
```
QuLogic: It returns `None`, not a `void *`.

obspy/signal/headers.py
```diff
@@ -114,6 +114,43 @@
...
+                           flags=native_str('C_CONTIGUOUS')),
+    # int len
+    C.c_int,
+    # int a,
+    C.c_int,
+    # int return_type,
+    C.c_int,
+    # enum lanczos_window_type window
+    C.c_int]
+clibsignal.calculate_kernel.restype = C.c_void_p
```
QuLogic: Also returns `None`, not `void *`.

obspy/core/trace.py
```diff
@@ -2171,18 +2171,19 @@ def split(self):
...
     def interpolate(self, sampling_rate, method="weighted_average_slopes",
-                    starttime=None, npts=None):
+                    starttime=None, npts=None, time_shift=0.0,
+                    *args, **kwargs):
         """
         Interpolate the data using various interpolation techniques.
 
-        No filter, antialiasing, ... is applied so make sure the data is
-        suitable for the operation to be performed.
+        Be careful when downsampling data and make sure to apply a proper
```
QuLogic: I would say appropriate instead of proper here.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+
+def lanczos_interpolation(data, old_start, old_dt, new_start, new_dt, new_npts,
+                          a, window="lanczos", *args, **kwargs):
+    r"""
+    Function performing Lanczos resampling, see
+    http://en.wikipedia.org/wiki/Lanczos_resampling for details. Essentially a
+    finite support version of sinc resampling (the ideal reconstruction
+    filter). For large values of ``a`` it converges towards sinc resampling. If
+    used for downsampling, make sure to apply a proper anti-aliasing lowpass
```
QuLogic: I would say appropriate instead of proper here.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    filter first.
+
+    .. note::
+
+        In most cases you do not want to call this method directly but invoke
+        it via either the :meth:`obspy.core.stream.Stream.interpolate` or the
```
QuLogic: Remove "the" at the end of this line.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    :type new_start: float
+    :param new_start: The start of the interpolated array. Must be greater
+        or equal to the current start of the array.
+    :type new_dt: float
+    :param new_dt: The desired new time delta.
+    :type new_npts: int
+    :param new_npts: The new number of samples.
+    :type a: int
+    :param a: The width of the window in samples on either side. Runtimes
+        scales linearly with the value of ``a`` but the interpolation also get
```
QuLogic: Runtime scales, and gets.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+        better.
+    :type window: str
+    :param window: The window used to multiply the sinc function with. One
```
QuLogic: The `with` is a bit strange here. But you can replace `multiple` with `taper` and drop the `with`, right?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+        of ``"lanczos"``, ``"hanning"``, ``"blackman"``. The window determines
+        the trade-off between "sharpness" and the amplitude of the wiggles in
+        the pass and stop band. Please use the
+        :func:`~obspy.signal.interpolation.plot_lanczos_windows` function to
+        judge these for any given application.
+
+    Values of ``a`` >= 20 show good results even for data that has
+    energy close to the Nyquist frequency. If your data is way oversampled
```
QuLogic: `way` -> `extremely`

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+    Also be aware of any boundary effects. All values outside the data
+    range are assumed to be zero which matters when calculating interpolated
+    values at the boundaries. At each side the area with potential boundary
+    effects is ``a`` * ``old_dt``. If you want to avoid any boundary effects
+    you will have to remove these values.
+
+    **Mathematical Details:**
+
+    The :math:`sinc` function is defined as
```
QuLogic: This should be `\operatorname{sinc}`; does sphinx support that?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    kernel.
+
+    .. math::
+
+        \begin{align}
+            L(t) =
+            \begin{cases}
+                sinc(t)\, \cdot sinc(t/a)
+                    & \text{if } t \in [-a, a]
+                    \text{ and `window`} = \text{`lanczos`}\\
```
QuLogic: Since this is LaTeX, the second backtick should be an apostrophe.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+                sinc(t)\, \cdot \frac{1}{2}
+                (1 + \cos(\pi\, t/a))
+                    & \text{if } t \in [-a, a]
+                    \text{ and `window`} = \text{`hanning`}\\
+                sinc(t)\, \cdot \left( \frac{21}{50} + \frac{1}{2}
+                \cos(\pi\, t/a) + \frac{2}{25} \cos (2\pi\, t/a) \right)
+                    & \text{if } t \in [-a, a] \text{ and `window`} =
+                    \text{`blackman`}\\
+                0                     & \text{else,}
```
QuLogic: The comma is kind of strange here.

obspy/signal/interpolation.py
QuLogic: Also, maybe you wanted `\texttt{window}` instead?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+            \end{cases}
+        \end{align}
+
+
+    Finally interpolation is performed by convolving the discrete signal
+    :math:`s_i` with that this kernel and evaluating it at the new timesamples
```
QuLogic: that this?

QuLogic: Split `time samples`.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    :math:`t_j`:
+
+    .. math::
+
+        \begin{align}
+            S(t_j) = \sum_{i = \left \lfloor{x}\right \rfloor - a + 1}
```
QuLogic: What is `x`? The notation seems a bit off compared to the paper.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+        \end{align}
+
+    where :math:`\lfloor \cdot \rfloor` denotes the floor function. For more
+    details and justification please see [Burger2009]_ and [vanDriel2015]_.
+    """
+    _validate_parameters(data, old_start, old_dt, new_start, new_dt, new_npts)
+    dt_factor = float(new_dt) / old_dt
+    offset = new_start - old_start
+    if offset < 0:
+        raise ValueError("Cannot extrapolate.")
```
QuLogic: Cannot extrapolate what?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+    if a < 1:
+        raise ValueError("a must be at least 1.")
+
+    return_data = np.zeros(new_npts, dtype="float64")
```
QuLogic: Avoid strings for `dtype`, no?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+    :type x: :class:`numpy.ndarray`
+    :param x: The x values at which to calculate the kernel.
+    :type a: int
+    :param a: The width of the window in samples on either side.
+    :type window: str
+    :param window: The window used to multiply the sinc function with. One
+        of ``"lanczos"``, ``"hanning"``, ``"blackman"``.
+
+    Return a dictionary of arrays.
```
QuLogic: Needs some description, I think.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+    .. code-block:: python
+
+        from obspy.signal.interpolation import plot_lanczos_windows
+        plot_lanczos_windows(a=20)
+
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        plt.figure(figsize=(10, 15))
```
QuLogic: Maybe accept `kwargs` that get passed to the created figure below?

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+        plt.subplot(height, 2, 3 + 2 * _i + 1)
+        plt.title(key.capitalize() + " Response")
+        plt.plot([0.0, 0.5, 0.5, 1000], [1.0, 1.0, 0.0, 0.0], "--",
+                 color="0.1")
+        plt.plot(np.fft.rfftfreq(len(x), dx),
+                 arrays[key]["fft"])
+        plt.xlim(0.2, 0.8)
+        plt.ylim(-0.1, 1.1)
+
+    plt.suptitle("Different windows for sinc interpolation with an a of %i"
```
QuLogic: I think this might be clearer as `with a=%i`.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+    arrays = {}
+    for key in _LANCZOS_KERNEL_MAP.keys():
+        arrays[key] = calculate_lanczos_kernel(x, a, key)
+        arrays[key]["fft"] = \
+            np.abs(np.fft.rfft(arrays[key]["full_kernel"]) * dx)
+
+    height = len(_LANCZOS_KERNEL_MAP) + 1
+
+    plt.subplot(height, 2, 1)
```
QuLogic: I prefer `plt.subplots`.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+        plt.plot(x, arrays[key]["full_kernel"], label=key.capitalize())
+
+    plt.legend()
+    plt.xlim(-a, a)
+    plt.ylim(-0.3, 1.1)
+    plt.title("All Windows")
+
+    plt.subplot(height, 2, 2)
+    plt.plot([0.0, 0.5, 0.5, 1000], [1.0, 1.0, 0.0, 0.0], "--",
+             color="0.1", label="ideal")
```
QuLogic: If you're going to capitalize the key, this one should be `Ideal`.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    plt.ylim(-0.1, 1.1)
+    plt.legend()
+    plt.title("Frequency Response of All Windows")
+
+    for _i, key in enumerate(sorted(_LANCZOS_KERNEL_MAP.keys())):
+        plt.subplot(height, 2, 3 + 2 * _i)
+        plt.title(key.capitalize())
+        plt.plot(x, arrays[key]["full_kernel"], color="black", label="final")
+        plt.plot(x, arrays[key]["only_sinc"], "--", color="gray", label="sinc")
+        plt.plot(x, arrays[key]["only_taper"], color="red", label="taper")
```
QuLogic: Probably these labels should also be capitalized to match.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,255 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        plt.figure(figsize=(10, 15))
+        from obspy.signal.interpolation import plot_lanczos_windows
+        plot_lanczos_windows(a=20)
+    """
+    x_max = 1024.0 - 0.5
+    n = 2 ** 15
+    x = np.linspace(-x_max, x_max, n)
```
QuLogic: Would it be simpler to use `rfftfreq` for `x`, or is it not well sampled enough?

obspy/signal/interpolation.py
QuLogic: Also, `np.empty` might be a bit faster...

obspy/signal/src/lanczos_resampling.c
```diff
@@ -0,0 +1,167 @@
...
+
+enum lanczos_window_type {
+    LANCZOS = 0,
+    HANNING = 1,
+    BLACKMAN = 2,
+};
+
+
+/* Sinc function */
+double sinc(double x) {
```
QuLogic: As these are not used elsewhere, you should make them `static`.

obspy/signal/src/lanczos_resampling.c
```diff
@@ -0,0 +1,167 @@
...
+
+/* von Hann window or raised cosine window*/
+double hanning_kernel(double x, int a) {
+    return 0.5 * (1.0 + cos(x / (double)a * M_PI));
+}
+
+
+/* Blackman window
+ *
+ * Values are taken from http://mathworld.wolfram.com/BlackmanFunction.html
```
QuLogic: Also defined in [NumPy](http://docs.scipy.org/doc/numpy/reference/generated/numpy.blackman.html) and [SciPy](http://docs.scipy.org/doc/scipy-dev/reference/generated/scipy.signal.blackman.html).

obspy/signal/src/lanczos_resampling.c
```diff
@@ -0,0 +1,167 @@
...
+ */
+double blackman_kernel(double x, int a) {
+    return 21.0 / 50.0 +
+           0.5 * cos(x / (double)a * M_PI) +
```
QuLogic: You may have the sign wrong on this term, based on the above links.

QuLogic: According to Oppenheim and Schafer, it should be negative, but I couldn't look up the reference from the Wolfram page because the page numbers are completely off and the Blackman/Tukey papers are rather dense.

obspy/signal/interpolation.py
QuLogic: OK, I see it's commented as necessary in the C code later.

obspy/signal/src/lanczos_resampling.c
```diff
@@ -0,0 +1,167 @@
...
+        x = dt * idx + offset;
+        for (m=-a; m<=a; m++) {
+            i = (int)floor(x) - m;
+            if (i < 0 || i >= len_in) {
+                continue;
+            }
+            _x = x - i;
+            /* Apply kernel and sum up if within 'a' samples on either side. */
+            if (-a <= _x && _x <= a) {
+                if (window == LANCZOS) {
```
QuLogic: Moving this `if` outside both loops should speed things up a bit.

obspy/signal/src/lanczos_resampling.c
```diff
@@ -0,0 +1,167 @@
...
+void calculate_kernel(double *x, double *y, int len, int a,
+                      int return_type, enum lanczos_window_type window) {
+    int idx;
+    double value;
+
+    for (idx=0; idx<len; idx++) {
+        value = x[idx];
+
+        /* Sinc times window */
+        if (return_type == 0) {
```
QuLogic: Also can speed up a little by moving outside the loop.

obspy/signal/interpolation.py
krischer: No that would not work here. `x` are the values on which the window is evaluated on before the FFT is taken. It is actually used further down the road to get the frequencies for the FFT.

obspy/signal/src/lanczos_resampling.c
krischer: In the Oppenheim book they use a minus sign but inside the cosine its two pi instead of only one. So its really the same as this formula here ;-)

QuLogic: From plotting it, I'm not sure that's true directly. But I think the difference is in the domain, NumPy's being a "window" (0 -- M points), whereas this one is more like a convolution kernel (-M -- M in time.)

obspy/signal/interpolation.py
QuLogic: Ah, yes, I guess I just saw `rfftfreq` in a loop and had that in the back of my mind. Maybe the call can still be moved out of the loop though?

obspy/signal/src/lanczos_resampling.c
krischer: Yea your probably right. It somehow must be the same window because I test it against the numpy implementation and they use the same definition as in the oppenheimer book.

obspy/signal/interpolation.py
krischer: Its just passed to MathJax but it does work.

obspy/signal/interpolation.py
krischer: But then one has to deal with a list of axis objects. I'm not sure that is cleaner. I'll try it for the next plot I do ;-)

obspy/signal/interpolation.py
krischer: I only used barebones pyplot and don't allow any options on purpose. This is really a one-off plot and if somebody wants to build something on top of it they should start over.

obspy/signal/src/lanczos_resampling.c
krischer: This really only done for plotting so its not speed critical at all and does not justify the additional complexity IMHO.

obspy/signal/src/lanczos_resampling.c
krischer: I tried this but the speed improvements are marginal (~1% maybe even within the measurement inaccuracy). This also means that the whole nested loop has to be repeated for each taper which makes the code really ugly. Probably not worth it here.

I also tried using function pointers which (from a code style point of view) would be the cleanest approach but the performance impact is severe. I guess the compiler can then no longer inline the functions.

obspy/signal/interpolation.py
krischer: It could but performance is really no issue at all for this plot.

krischer:
Thanks a lot for the very thorough review @QuLogic! I acted on most suggestions so good to merge from my point of view!

obspy/signal/interpolation.py
QuLogic: Not a list, a 2D array, so that way it's nice and clear of the placement.

obspy/signal/src/lanczos_resampling.c
QuLogic: It should be able to inline if they're static. But I guess compilers are just too good and loop unrolling and such these days for this change to make much difference any more.

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,264 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+
+
+def lanczos_interpolation(data, old_start, old_dt, new_start, new_dt, new_npts,
+                          a, window="lanczos", *args, **kwargs):
+    r"""
+    Function performing Lanczos resampling, see
+    http://en.wikipedia.org/wiki/Lanczos_resampling for details. Essentially a
+    finite support version of sinc resampling (the ideal reconstruction
+    filter). For large values of ``a`` it converges towards sinc resampling. If
+    used for downsampling, make sure to apply a appropriate anti-aliasing
```
QuLogic: an appropriate

obspy/signal/interpolation.py
```diff
@@ -172,6 +173,264 @@ def weighted_average_slopes(data, old_start, old_dt, new_start, new_dt,
...
+    :param x: The x values at which to calculate the kernel.
+    :type a: int
+    :param a: The width of the window in samples on either side.
+    :type window: str
+    :param window: The window used to multiply the sinc function with. One
+        of ``"lanczos"``, ``"hanning"``, ``"blackman"``.
+
+    Returns a dictionary of arrays:
+
+    * ``"full_kernel"``: The tapered sinc function evaluated at samples x.
```
QuLogic: `x` should be in double-backticks, right?

QuLogic:
Just a couple minor typos, but :+1: otherwise.

krischer:
I am really bad at proof reading my own stuff o.O

Thanks again!

PR #1102, opened 2015-07-31
Bugfix for libmseed when writing negative timestamps with microsecond accuracy

krischer:
`libmseed` has a bug when writing timestamps smaller than 0 with microsecond accuracy (which involves Blockette 1001). This PR fixes that.

I also added a test and fixed a small issue with the record analyzer.

@CTrabant: 4072176f67da7f6429d62994562980f421191043 is a patch for `libmseed`.

### DIFF
--- a/obspy/mseed/src/libmseed/msrutils.c
+++ b/obspy/mseed/src/libmseed/msrutils.c
@@ -358,7 +358,16 @@ msr_normalize_header ( MSRecord *msr, flag verbose )
 	  sec = msr->starttime / (HPTMODULUS / 10000);
 	  usec = msr->starttime - (sec * (HPTMODULUS / 10000));
 	  usec /= (HPTMODULUS / 1000000);
-	  
+
+      /* Deal with negative timestamps and a negative offset.
+       *
+       * The value is rounded down when writing the fixed header
+       * thus the offset needs to be relative to the next smaller
+       * value representable by the BTIME definition. */
+      if (msr->starttime <= 0 && usec < 0) {
+        usec += 100;
+      }
+
 	  msr->Blkt1001->usec = (int8_t) usec;
 	  offset += sizeof (struct blkt_1001_s);
 	}
--- a/obspy/mseed/src/libmseed/msrutils.c
+++ b/obspy/mseed/src/libmseed/msrutils.c
@@ -7,7 +7,7 @@
  *   ORFEUS/EC-Project MEREDIAN
  *   IRIS Data Management Center
  *
- * modified: 2012.088
+ * modified: 2015.213
  ***************************************************************************/
 
 #include <stdio.h>
@@ -257,6 +257,8 @@ int
 msr_normalize_header ( MSRecord *msr, flag verbose )
 {
   struct blkt_link_s *cur_blkt;
+  hptime_t hptimems;
+  int8_t usecoffset;
   char seqnum[7];
   int offset = 0;
   int blktcnt = 0;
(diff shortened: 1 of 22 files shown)


krischer:
Travis error is unrelated. Seems to be some kind of issue with Python 2.6.

obspy/mseed/src/libmseed/msrutils.c
```diff
@@ -358,7 +358,16 @@ msr_normalize_header ( MSRecord *msr, flag verbose )
 	  sec = msr->starttime / (HPTMODULUS / 10000);
 	  usec = msr->starttime - (sec * (HPTMODULUS / 10000));
 	  usec /= (HPTMODULUS / 1000000);
-	  
+
+      /* Deal with negative timestamps and a negative offset.
```
QuLogic: The indent is wrong here (though you probably didn't notice because libmseed uses strange mixed tab/space indents.)

obspy/mseed/tests/test_mseed_special_issues.py
```diff
@@ -731,6 +731,35 @@ def test_writing_blockette_100(self):
...
+        Tests that reading and writing data with microsecond accuracy and
+        before 1970 works as expected.
+        """
+        # Test a couple of timestamps. Positive and negative ones.
+        timestamps = [123456.789123, -123456.789123, 1.123400, 1.123412,
+                      1.123449, 1.123450, 1.123499, -1.123400, -1.123412,
+                      -1.123449, -1.123450, -1.123451, -1.123499]
+
+        for timestamp in timestamps:
+            # For the sake of sanity check first for times after 1970.
```
QuLogic: Is that true? The `timestamps` list doesn't seem to be ordered that way.

krischer: Oops. True enough. Its a leftover from the first version of this test...

obspy/mseed/src/libmseed/msrutils.c
krischer: It looked right in the editor I used. I couldn't use my normal editor as it auto-formats a bunch of things so the changelog would have been huge...

QuLogic: If you set tabstops to 8 characters, it should be more obvious.

CTrabant:
> @CTrabant <https://github.com/CTrabant>: 4072176 <https://github.com/obspy/obspy/commit/4072176f67da7f6429d62994562980f421191043> is a patch for libmseed.
> 
Thanks for the patch.  I took a slightly different approach and used the opportunity to round to tenths of milliseconds and range the microsecond offset to that recommended by SEED:

https://seiscode.iris.washington.edu/news/48

PS. You can blame the goofy tab/space formatting on years of development in Emacs.  I've been hesitant to re-justify with consistent space usage because it would make a mess of difference listings, someday though, plus it looks fine in my Emacs :P

krischer:
@CTrabant Thanks a lot. That should do the trick!

@ all: We currently have libmseed 2.16m in the `master` and libmseed 2.15 in the `releases` branch. So we could either update both to libmseed 2.17 or only update the master and leave this patch here for the releases branch.

I would update both but if someone has a strong reason not to do this I'm also ok with it.

megies:
> So we could either update both to libmseed 2.17 or only update the master and leave this patch here for the releases branch.

I don't care too much which way it is, your call.

krischer:
I updated to the latest libmseed version. Will merge as soon as CI passes.

krischer:
Travis failure is again unrelated.

Issue #986, opened 2015-03-17
Some TauP paths do not end at the requested location


QuLogic:
Pulled this out of #959 for tracking purposes. Probably shouldn't hold up the release, though. @crotwell can comment here if he has any further input.

> I did find some other (a bit obscure) phases where it does not quite add up. The issue is identical in your Java version and our Python port. On the left is always the Python version, on the right the Java version.
![untitled](https://cloud.githubusercontent.com/assets/800487/6229800/abf51eb6-b6ba-11e4-8d3e-b43588f34e41.png)

crotwell:
I believe I have fixed this in TauP (java) with commits 

https://github.com/crotwell/TauP/compare/98dc60ccb272f2bf4ec69a32b0e546d3c2a2cd4d...9b737bd9d11a0c6fdf758051b709ad3d4e7ee33c

Basic issue is that the path is very sensitive to ray parameter for very long phases like PKPPKP.  The linear interpolation does a reasonable job of getting close to the time, but when taup calculates the path, it uses the interpolated arrival's ray parameter and sends it back through the model. But if the ray parameter is slightly wrong due to the linear interpolation in the original calculation, and the path is long, then the errors accumulate and the output can be off by a couple of degrees. The new method does a refinement calculation shooting rays that helps to get a much more accurate ray parameter, and hence more accurate path. Some tradeoff in speed, so I might make this optional before the release.

Other little bug fixes and improvements in there as well.

crotwell:
Also, prerelease of TauP with this fix in it is here:
http://www.seis.sc.edu/downloads/TauP/prerelease/TauP-2.3.0-SNAPSHOT.tgz

QuLogic:
Thanks for the update @crotwell. We'll try and look into incorporating the changes soon.

PR #1103, opened 2015-08-02
taup: Support for buried receivers

QuLogic:
This is a port of buried receivers from the TauP prerelease (see commit messages for specifics.) Once this is merged, it should be easier to apply the fixes noted in #986. This is currently targeting 0.11 since it's an enhancement, but if we want the above bugfixes, it might be necessary to backport to 0.10. 

One test is failing; I am not sure if I made a mistake or whether the tests should be updated (some commit messages in TauP indicated some tests _might_ have needed updating.) @krischer Are you able to run the prerelease as well to run some checks?

I added a test _plot_, but no numerical tests. @krischer If you can run the prerelease, maybe you can add some tests as well? Otherwise, I'll look into running the prerelease TauP eventually.


megies:
> it might be necessary to backport to 0.10.

Fine with me. Just do what is easiest, more important to not waste precious time trying to prevent a minor feature change leaking back.

krischer:
I replaced the part thats failing with the output from the prerelease. For some reason it suddenly produces way less phases (but that might just due to it being at the CMB and some phase naming reasons) but that is consistent with your changes.

There are unfortunately a couple more differences. The takeoff angle appears to be off 180 - x degrees. Maybe some sign error somewhere? The traveltime is also a tiny bit off compared to the output from the prerelease java version. The difference is very small and I cannot tell if this will get away by the time we manage to implement the remaining changes.

None of the command line utilities can set the receiver depth (or at least I did not find the settings). I'll give the jython console a try once I get around to it to generate some test data for the buried receivers.

QuLogic:
I think the `180 - x` thing might have been added in some later commits; I'll have a look for them. I think the command-line utilities should be able to take a `--stadepth` option.

QuLogic:
I made some changes so that takeoff and incident angle calculations are consistent with the prerelease, but some new tests failures arose. They will probably need to be updated.

QuLogic:
Ping? I'd like to add the other corrections.

krischer:
I added a simple test for a buried receiver. It (and others) currently fails due to the aforementioned reasons.

I'm not sure how to best proceed here. Merge this and live with a failing master branch for a while? Keep working on this branch? Make a new branch from this branch here?

I would prefer one of the latter two options.

QuLogic:
Oh, that's weird, I thought I had applied everything that could be related to angles. I guess we'll have to try implementing the rest and see how things go.

megies:
@QuLogic, @krischer, what's the status on this? Close to be ready or rather punt it to future release?

QuLogic:
We think it needs the long phase fixes; I'll work on that later tonight or tomorrow. I think it's nearly there now.

megies:
What's it with the failing tests? I guess those are these fixed in master by merging #1189?

krischer:
Should be fixed. #1189 had no test failures.

QuLogic:
Yea, we never really figured out why it failed here as the buried receiver and long phase stuff is all mixed together in the Java version.

megies:
Yeah, that's what I figured more or less.. :+1:

PR #1104, opened 2015-08-06
Bug in `Trace.remove_response(..., water_level=None)`

megies:
When water level was set to `None` (i.e. no water level used, solely relying on `pre_filt`) the instrument response spectrum was never inverted but multiplied in frequency domain anyway, thus resulting in a convolution rather a deconvolution, resulting in completely unexpected and unusable results.
Before the fix this was just simply circumvented by specifying a high water_level value (e.g. `water_level=600`) which effectively means no water level at all.

Thanks to @cja12 for letting me know!

Here is a notebook to visualize the issue (esp. see bug last three boxes):
https://gist.github.com/megies/73af368d7f625fb13b0d

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,12 @@
 0.10.x:
+  - obspy.core:
+    * Fixed a bug when using
+      `Trace.remove_response(..., water_level=None)`.
+      With that setting that is supposed to not use any water level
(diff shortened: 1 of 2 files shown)


megies:
Travis fail on Py2.6 is unrelated (see PythonCharmers/python-future#162).

PR #1105, opened 2015-08-06
plot what is going on in frequency domain during Trace.remove_response()

megies:
This adds a kwarg `plot=False` to `Trace.remove_response()`. It can be used to bring up a plot of what is going on in the frequency domain during instrument correction..
 - `pre_filt` data spectrum tapering
 - inversion of instrument spectrum, optionally with `water_level`
 - multiplication of inverted spectrum onto data spectrum
 - time domain comparison of raw/corrected data

Here's an example:
```python

st = read("http://examples.obspy.org/IU_ULN_2015-07-18T02.mseed")
tr = st[0]
inv = read_inventory("http://examples.obspy.org/IU_ULN.xml",
                     format="STATIONXML")

pre_filt = [0.0033, 0.005, 45, 50] 
output = "DISP"

tr.attach_response(inv)
tr.remove_response(pre_filt=pre_filt, output=output, water_level=60, plot=True)
```

figure_1

I know this should in principle go into master branch.. however, I think this could be extremely valuable to many obspy users/beginners and I think we should consider making an exception to get this out as soon as possible (to include it in tutorial etc.). Other opinions?

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2265,7 +2265,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
     @_add_processing_info
     def remove_response(self, output="VEL", water_level=60, pre_filt=None,
                         zero_mean=True, taper=True, taper_fraction=0.05,
-                        **kwargs):
+                        plot=False, **kwargs):
         """
         Deconvolve instrument response.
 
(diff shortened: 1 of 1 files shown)


megies:
Need to add changelog..

QuLogic:
Maybe [colour the scales](http://matplotlib.org/examples/api/two_scales.html)?

megies:
Since we're pushing for a major release, I think I'll instead base this on master.. continued in #1116.

PR #1106, opened 2015-08-08
Spectrum Plot

jrleeman:
This is just a rough sketch of a spectrum plot function. The need for this came up at a recent EarthScope workshop. Not exactly sure where this belongs, but opening the PR should help with that! I'll clean up the docstring and add some tests as well.

### DIFF
--- a/obspy/imaging/spectrum.py
+++ b/obspy/imaging/spectrum.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+
+"""
+Plotting the spectrum of seismograms
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA @UnusedWildImport
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy import fft
+
+def plot_spectrum(data, samp_rate, outfile=None, fmt=None,
+             axes=None, title=None, show=True):
+
+    """
+     Compute and plot the spectrum of the input data
+     """
+
+    # enforce float for samp_rate
+    samp_rate = float(samp_rate)
+
+    npts = len(data)
+
+    # Compute the FFT
+    npts = len(data)
+    k = np.arange(npts)
+    T = npts/samp_rate
+    frq = k/T  # two sides frequency range
+    frq = frq[range(npts/2)]  # one side frequency range
+
+    X = fft(data)/npts  # fft computing and normalization
+    X = X[range(npts/2)]
+
+    if not axes:
+        fig = plt.figure()
+        ax = fig.add_subplot(111)
+    else:
+        ax = axes
+
+    ax.loglog(frq, abs(X), color = 'k') # plotting the spectrum
+    ax.set_xlabel('Freq (Hz)')
+    ax.set_ylabel('|Y(freq)|')
+
+    # set correct way of axis, whitespace before and after with window
+    # length
+    ax.axis('tight')
+    ax.set_xlim(0, frq[-1])
+    ax.grid(False)
+
+    if axes:
+        return ax
+
+    if title:
+        ax.set_title(title)
+
+    if outfile:
+        if fmt:
+            fig.savefig(outfile, format=fmt)
+        else:
+            fig.savefig(outfile)
+    elif show:
+        plt.show()
+    else:
+        return fig


obspy/imaging/spectrum.py
```diff
@@ -0,0 +1,65 @@
...
+    npts = len(data)
+
+    # Compute the FFT
+    npts = len(data)
+    k = np.arange(npts)
+    T = npts/samp_rate
+    frq = k/T  # two sides frequency range
+    frq = frq[range(npts/2)]  # one side frequency range
+
+    X = fft(data)/npts  # fft computing and normalization
```
krischer: scipy's fftpack also has an `rfft()` function which is quicker for real valued input which can always assume for waveforms.

ThomasLecocq: just a side note, rfft returns an 2*N length array for a real N input, [0, 0j, ... ] ...

obspy/imaging/spectrum.py
```diff
@@ -0,0 +1,65 @@
...
+    samp_rate = float(samp_rate)
+
+    npts = len(data)
+
+    # Compute the FFT
+    npts = len(data)
+    k = np.arange(npts)
+    T = npts/samp_rate
+    frq = k/T  # two sides frequency range
+    frq = frq[range(npts/2)]  # one side frequency range
```
krischer: There are `fftfreq`/`rfftfreq` functions to calculate the frequency values in scipy's fft pack.

obspy/imaging/spectrum.py
krischer: Yea true...thats a bit awkward...one can also use `numpy.fft.rfft` but the scipy variant is faster.

ThomasLecocq: or, just slice the output
```python
X = rfft(data)[::2]
```
?

krischer: No that cannot work. The spectrum still has to be complex. If you really want to work with complex values you can add zeros after the first sample (and after the last if n is odd) and set the dtype to a complex array...

```python
In [3]: temp = scipy.fftpack.rfft(np.arange(10))

In [4]: i = temp[0]

In [5]: temp[0] = i

In [6]: rfft = np.concatenate([[i], temp, [0]])

In [7]: rfft.dtype = np.complex128

In [7]: rfft
Out[7]:
array([ 45.+45.j        ,  -5.+15.38841769j,  -5. +6.8819096j ,
        -5. +3.63271264j,  -5. +1.62459848j,  -5. +0.j        ])

In [8]: scipy.fftpack.fft(np.arange(10))
Out[8]:
array([ 45. +0.j        ,  -5.+15.38841769j,  -5. +6.8819096j ,
        -5. +3.63271264j,  -5. +1.62459848j,  -5. +0.j        ,
        -5. -1.62459848j,  -5. -3.63271264j,  -5. -6.8819096j ,
        -5.-15.38841769j])
```

krischer:
Good idea. We have been throwing around the idea of a "frequency domain trace" for a while but I guess that will not happen for a bit. Until then this is a very good idea. I guess we should then also have a `Stream/Trace.plot_spectrum()` method.

A couple comments:

* Should we enforce a taper on the data?
* It probably worth it to do the FFT with the next power of 2. We have a function for that: `obspy.signal.util.next_pow_2`
* SciPy has a couple of spectral density estimators - we could also use one of these.

jrleeman:
I wouldn't think that enforcing a taper is a good behavior since most users (especially from the SAC world) won't expect it. Probably should just hold ```plot_spectrum``` to do what it says only. I'll rework this a bit. Does this belong in imaging as it sits?

jrleeman:
I tried padding to powers of 2 and didn't see any measurable speed increase, but need to look into Numpy's FFT routine. Also need to look about it's normalization. I think @cja12 may have already looked at that?

shaharkadmiel:
Hi jrleeman, I believe you need to normalize `X` by 0.5*npts when using `numpy.fft.rfft`. If you test this on a simple superposition of some sine waves with different amplitudes and different frequencies you will see that the amplitude at the known frequencies in the frequency domain are wrong.

cja12:
Actually, I think for the spectral plot, you want to multiply by dt, which is the same as dividing by the samp_rate. 

Create a unit area "impulse" to test - make a zero array, then put a spike of 1/dt somewhere in the array. The area of that spike (really a triangle) is 1/2 (2dt) (1/dt) = 1. The spectral amplitude should be unity. If you bring that signal back to the time domain, you would need to multiply the spectrum by df and a factor of 2 for the realft tricks (dt * df = npts).

srate = 10.0
d = np.zeros(1024)
d[512] = 1/srate
plot_spectrum(d, srate, outfile=None, fmt=None, axes=None, title=None, show=True)

shaharkadmiel:
@cja12 , I see what you are saying about an impulse but when I test this approach with a superposition of sine waves at known frequencies and amplitudes (with or without added noise) I get much higher amplitudes than expected. However, when I divide by 0.5*npts (which is the same as FFT size) I get exactly the expected amplitudes.

```Python
def sine(amp, f, t):
    return amp * np.sin(f * 2*np.pi * t)

f1 = 0.5 ; amp1 = 20.
f2 = 1.  ; amp2 = 10.
f3 = 5. ; amp3 = 5.
f4 = 15. ; amp4 = 1.

tmax = 200
delta = 0.01

t = np.arange(0,tmax, delta)
npts = t.size
# signal = sine(amp, freq, time)
s0 = sine(amp1, f1, t)
s1 = sine(amp2, f2, t)
s2 = sine(amp3, f3, t)
s3 = sine(amp4, f4, t)
noise = 2*np.random.random(npts)-0.5
signal = (s0 + s1 + s2 + s3)# + noise)

fig, (ax1,ax2) = plt.subplots(1,2, figsize=(6,3), sharey=True)
freq = np.fft.rfftfreq(npts, delta)
amp = np.abs(np.fft.rfft(signal))*delta
ax1.loglog(freq, amp)
ax1.grid()
ax1.hlines([amp1,amp2,amp3,amp4],freq[1],freq[-1], linestyles='dashed')
ax1.set_xlim(freq[1],freq[-1])

freq = np.fft.rfftfreq(npts, delta)
amp = np.abs(np.fft.rfft(signal))/(0.5*npts)
ax2.loglog(freq, amp)

ax2.grid()
ax2.hlines([amp1,amp2,amp3,amp4],freq[1],freq[-1], linestyles='dashed')
ax2.set_xlim(freq[1],freq[-1])
```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAYEAAADSCAYAAACsNHzJAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAALEgAACxIB0t1+/AAAIABJREFUeJztnXm8HEW5979PAlkICRAgQEAW2S6giCwJEJCwGRFEQJCLIltAFu+V933Fy70oMQZl86roZTdgREEWBQVECcs5gPAKiIjshH1HdsJOyHP/qJqcPn26Z7pnerprZp7v59Of3qqrflNd09Vdy/OIqmIYhmH0JsOqFmAYhmFUh1UChmEYPUxXVwIiMrNqDUmYrnyEqqsqQs6PULWZrjoaurlPQERUVaVqHXFMVz5C1VUVIedHqNpMVzpd/SVgGIZh1KejKgEROaDednzdbNxZzjXSEpquenrK1pWmMzRdZRJKGcpyn0xX+LrqaRmCqnbMAvTX205YazNxZznXSEtouhroKVVXjnyqVFeZS54ylCc/8uZJlvtUxb0yXcU+C6JLR/UJiEjniDU6Fq2gjdbKttFu0sr1YmULARCR8cCFwBjgBlU9Juu1VfxBjd6hyoexlW2jXdQr11X1CRwKnKaqU4DNfKVQOCIytR3xtorpykeouqoi5PwIVZvpSqeSLwHgOuA+ERkOjALeq0iHYRhGT1NZn4CIrIzr5HocmKaqCzNco/bJnB0RRqvyTtU6OomqypiVbaOd1CtfhTYHiciNIjI5sj9NROaJyJsicrGIjPXHJ6rqM6q6NvAAsFOROgwQQYC3q9ZhGEbYFFIJ+If9WcBWgPpjywAXAEcDqwMCHOcvOUlENvTbbwNvFaEjQdfUdsTbKqYrH6HqqoqQ8yNUbaYrnaK+BKb4uN6MHNsVuENVL1XVl4ATgH/1504EzhCRPmCcqvYXpMNoEhH+D8zarmodhmGUTNaJChknMzwGTPLbPwROipwbBSwExmSMaybuqyK+TI2EmWr7yfugw6BPYVjW8ApXvxiK/gr3k8rczIL/JzNT0qn6t9t+9+6nlutCO4ZF5DFgb1W9TUTOBp5Q1e9Hzr8PrKqqzzcZv6p1nmVChGHAh8AwVRreZBEUeFaVlUvQ9jFV7ml3Os1gHcNGN1Jax3CM14DREREjcENSX21jmoMIob0tiZJ0SWydgbkj26JkKHeLsGzWwKHex6oIOT9C1Wa60mlnJfAgsGFkfz1gnqranAADOsx4oWF0K+1sDloOmAfsAdwNnAf8RVVntRC/fTJnRIThwAJguCoZ5mCgwHOqTCxBmwITVHmx3WnlxZqDjG6kkuYgdSOC9gfOxXUYP4cbIWSUQxPNQYZh9BqFVgKquoaq3hbZv9wfG6uq01X1gyLTa0QI7W1JhKoL5o6oWkES4eZXNYScH6FqM13pWLusEUHK/GpoOS0RrhJhzSLEGEavUqk/AREZBvxRVadlDG/tphkRYXHgfWBxVRZkCK/A86qsVII2BVZQ5Z8FxHOwKucUo8z6BIzupKohonURkfWBu3CjhgzDMIwKqLI56H5gU+DJdiUQQntbEqHqgmusT6ADCDk/QtVmutKprBJQx3vQeDar0RQ2OsgwjIYUXglkNSddBhqoYbpQdcGOpY7eykq4+VUNIedHGdpEGCFCrudIqHkWhK6agaFWF2AacBbOSFzNiNwywMu4CWPLAZcAp8SuuylHGklGkFINfJFuqKsHwo9QZxRusYzhVeG5kvSrgq7QevyqcFCh+en+EsX8J3L+f6xsZw5/ri9DoegJP3y9cl1kIZ4F/Ax4g4FKYH9gbiTMxsDzsetuzPNHyalpalG/r8ilDF2gI/3DdkTG8ApzXynn92tiJZA3v3w80wu+N1pRmcicbqjluixtoDe77ApLV6j55dNJza/CmoNUdYaqHoJ786+xIXBnZP8+YIKIjIlc96miNBiGYcQRYR6sNrpxyB6lDTXOYwx8CZwNfCt2/n1gxQzxzCT9Ezntk2cqg21oz+nd8CPV+RNYLGN4VfhtSfpVQVdoPf4+hZ2Kzs/U8wX+R2ampBNoWQot/J992c4aXhVWCkh/JeFTzxU+WSxmRO5k4H1V/bY/NwJ4FxitTVgTtQk12RFhFPAOMEqVhnntJ169oMqKJWhTYEVVXiggHpss1mOIcDOwpWq2kW++nKylyiNt1rU0MFGV+9qZTjNUOVmscHPSIqIJy8yUsHNyhp/ZLeFh1Dvu6Mh3s4UHuHSFMvQnhW0Q/5y0a2D67HblZ9lk1SYiU0Mue+0v2zdvGcmvDOEBVn643frhV68C9+aMf0679GQu20V+5vqviscYaA5aDudEZlu/fRUwo4W4NWf4qc2m1c6lDF2go32zy6iM4bXkjuGGTYKN8svHYx3DAS0lle1cHcOunOz85RJ0XZZHV5n3sl75auuXgFZsTlpDGIObQKi6YMeGNoaqINz8qoaQ8yNcbVfeWrWCJELIr8WKjlBV14jtXw5cXnQ6hmEYRut0tSlpCcAuRxIl6WrCbMQ1hb8UFEGo97EqQs6PcLXtMrlxmPIJIb+6uhIwcmOjU4wupXMedSKsK8LfSkuwjE6JOp0Vw4Cr83RupCwzU8LP7N3wS/jx0aMyhleFF0rSr4kdw/njVzWzEb0Y/s++DGUNrwprlaD/t5rWMZwnftADI7+vkPysV64rcyojzp/ARcBSqrpqxmtUbSx1JkQYA7wJjFHl7QzhFXhRlQklaFPceOrnCojH5gn0GE3OE1hHlXlt1nUpsHtWXXXiORA4t9V4BscZoFMZzJ9AgFyzeNUKkgg3v6oh5PwIV9uuk6pWkEQI+WX+BLoXe6s0DKMhhVQCktGHgIjMEJE+EflcEek2QgMYg5tEybpyVAY2T6ATCDk/wtV2+W1VK0gihPxqqRLwD/uzgK3wb/QisgxwAXA0sDruIXQcgKrOUtVtVfWKVtI1DKOnaeYr176MU2j1S2CKj+PNyLFdgTtU9VJ1M4ZPAP61Thxtaw4Kob0tiXDnCVxrfQIdQMj5Ea62z5fRJ5D7WRZCfrU0OUhVZwCIyA6Rw6k+BFT1rYQ4zJ9AONjbktEJWD9igbSjY3gszrsYAKr6LrDAH89MspW+wTWnt6aYup83fFn7tXbANqcn0A+svnXW62vHqs6f+H5afrnfd9K6Rabn1221Mtpq2VbV/lDuTXy/pLIN9Af3LIBLlst7fb2yXWx+pZfrQuYJSBt9CMTSsbHUGRFhHPA6ME6V+RnCK/CSKsuXoM3mCQSSbifS5DyBdVV5qM26fgvsYfME2uBDoFniNWEohKoLrh1RtYIkws2vagg5P8LVtpvZDkqhHZXA74GtRWRbEVkO1zF8fhvSMepjb5WGUS4d2VdReCWgFfsQiGnpryLdRoQ7T2CHMucJZNYV6n2sipDzI1xtv7N5AikUYjpYzYeAYRjlYV+5BdI59lWbIIT2tiRK0tXEH+Va8yfQAYScH+Fq281sB6XQ1ZWAAdhbk9F9dGTbe6hUUgmIyHgRmSsiN4vI8e1KJ4T2tiRC1VVyn0Bmws2vagg5P0rS1sSLze9uL17GEHJXTiHcy6q+BA4FTlPVKcBmIjK+Ih3djNlXMQyjIVVVAtcB14nIcGAU0JY5BCG0tyVRsq48toPK7BPIrCvU+1gVIedHSdqaaA7aw/oEUqikElDV24ClgAdws4nfqUJHl9PMW721tRqdgH2xFkjLlYA04UtARFZS1WdUdW1cRbBTqzqSCKG9LYmA5wl82D4ZzRPqfayKkPMjXG2X2jyBFJquBKQ1XwIni0jNtMTbwBDrooZhGCk088VqXw8ptPIl0IovgROBM0SkDxjXrtowhPa2JGyeQD5CvY9VEXJ+hKvtC5tVrSCJEPKr6T99K74EVPVeXCVitJ88lUGZb0v2ZmaUiHV3pVF0x3AhvgSgdZvr0WNZw5e1X6bNdfj4lOzhOyu/zJ9A9fcmvl+mr4x8/wVpe9mGS5bPe31X+BOQknwJ+PjM5npGRFgOeBFYVpVXMoRX4DVVlilBmwIfUeXpAuIxfwI9hgi3AFvk9Cewvir3t1nXJcCeve5PIBhfAhBGe1sSJesKdZ5AZkK9j1URcn6UpK2JN9e9bJ5ACkVXAuZLwDCMdhPqF1NHdjwUWgmE5EvA6+mvKu16hGtfxeYJdAIh50e42i6xeQIptPz5b74EgifUt6ZQdRnhY/MECqSrTUmH0N6WREm6mpknMLx4Ga0T6n2sipDzI9yyvZfNE0ihko5AEVkVuAw3nHS+qu5ahQ4jjtjbkmH0GFWNBlkT+ImqntfOREJob0siXNtB25s/gQ4g5PwoSVsTzUGXlOFPIDch3MuqmoPWBKaLyA0isl9FGrqd0P0JFDWSwr5eeo9Q77mNDsrB08BRwI7AwSKyQjsSCaG9LYlw5wlcZ30CHUDI+RGuti/aPIEUSqsEJGJKGpirqrer6vvAzcBHy9JhBEOob3OG0VOU5k8gZkp6johsJCICfAI307hwQmhvSyLceQLb2zyBDiDk/AhX28XWJ5BCaf4EYpwAnI6zAHaFqja0bWM0jb1xG92GzRMokEr8Cajq/aq6papuo6pntKChLiG0tyURqi7rE+gMQs6PcOcJ7F3GPIHclVMI97ISfwJGKYQ+OsgwSqRFc8ldjPkT6F6b67jWts1D9ScgreaX+30nmz+BJn9rB5dtbcKfgLT798PFE/Jeb/4E8qdlNtczIsJE4BlgJVWezxBegfmqjCtBmwKrqfJkAfEcosrsYpRVV8asbGenSX8CH1flnjbruhDY2/wJmD+BhoSqy/oEOoOQ8yNcbaX0CeQmhPwyfwLdj71dGt2Gte8XiPkTqICA5wksLF5G64R6H6si5PwIt2xfVMY8gdyVUwj30vwJGIbRC9gXcQrmT6ACStaVx3ZQmeUhs65Q72NVhJwf4WrbZ9OqFSQRQn51dSXQ49g8AcMwGlJJJSAix4kzJtcnIneKyG7tSCeE9rYkAvYnYLaDOoCQ8yNcbb/+a9UKkgghvypxKqOqxwKIMy53DnBFFTqMIdiXgGH0GFU3B80CfqSqbXkDDaG9LYmSdDXxQLd5Ap1AyPkRrraO6hModQhsFf4EdvH7ywOrqepfytLQo9jbvdFtNPGQNP/ZaVThT+BKH3R/nNnpthFCe1sS4Y6lLrVPILO+UO9jVYScH+GW7QusTyCFqvwJAOwAXNNs+oZhGEbrVOJPAEBVP6Oqr7eQfkNCbZ8Md57A9dYn0AGEnB8laWuiOehLndQnUCpNVwKqOkNVDwFejhxO9SeQN34pxpT0Ru02HxvwvjhTy9tunvV6uHNYQPoz6O2nF01J274r2/nKyn3rtlsfXLR8UfF1sinps4EnVPX7kfPvA6uqakNzxhnSMnO7GRFhNeBxYFVVnsoQXoG3VcldYTehTYE1VHm8gHjMlHSP0aQp6Y1UuavNui4A9inAlPQBwM871ZT0a8DoSMIjcHMRXi04HSM79mAxDCMV8ydQASXpauLhf73ZDuoAQs6PcLV92foEUjB/AoZhGD2M+ROogJLHUuf4ItiuzJmKmdMK9T5WRcj5Ea6288uYJ2D+BPy++RMwDCM0rG8shaptB7WVENrbkihZV555AkGWh1DvY1WEnB/hatvX+gRSCPJPbxRC6G8+oeszjJ6gKn8CK4vI9SJyq4gc2q50QmhvSyJUXbCd+RjuAELOj3C1/cpsB6VQ1ZfAfwDHApsDB4jI6AbhjebJ88Ztb+eG0TylmoAuiqoqgXVwNoYUN4povXYkEkJ7WxI2TyAfod7Hqgg5P8LVZn0CaVT1p38QmCIiSwCbAYtXpMMwjJ4gyPebICjNn4AMdipzPPDvwG+BF4BnW9WRRAjtbUmE62PY+gQ6gZDzI1xt591RtYIkQsiv0vwJxJzK7AIcAewGzFfVhgbOjNyE3r4fuj6ju+jI9voyqMqfwIPAZcANwH+3oKEuIbS3JRGqLvMn0BmEnB/hatvP+gRSqMSfgKrerKqTVXVzVb0uKX4xfwKF2FyHnfP4E5CA9GeyuW7+BHpzP78/gXtL8Cdw4YSqfl/j+NLLtfkT6FJEWBt4CFhLlUcyhFfgPVVGlaBNgTVVebSAeMyfQI/RpD+BTVT5W5t1/RLY1/wJmD8BwzCMjsH8CVRASbqasCJq8wQ6gZDzI1xt+29StYIkQsgv8ydgVIU1fRhGAJg/gQoIeJ5AkMPoQr2PVRFyfoSr7Rc2TyAF8yfQvdibtmEYDQmyDbgoQmhvSyJUXXB9kBVHuPlVDSHnR7jaDrB5Ail0dSVgAPZFYBhGHUqtBERkmIhcnbZfNCG0tyVRso/hHJTaJ5BZX6j3sSpCzo9wtc2xPoEUSqsERGR94C682WgR2SC6bwSBfTUYRo9R5pfA/cCmwJN+/77YfuGE0N6WRLjzBPqCrARCvY9VEXJ+hKvtgDLmCeT+kg4hv0qrBNTxHj6j4vuG0Q5EWFyEf0s595HY/pikcEY3IEG+4IRAU5WA5Pch8LmiBOchhPa2JELVBduWXiGLICL1y2GD/FpcpK5Too8B/xNJbzG/VuBJEbbw+x9jsEXcYAm3/ISnTaT2Jfxz6xNIIVclIM37ELiiSNFGJppoDqqEPYEPW7j+dOB9EdZoFFCEN4EPRNgtcvgWv767BQ2G0bHk/RJoxYdAjfjbZtvePkNob0siVF3QV2YfUa1yWjvxpPtCGO22M+XXof66DUW4WITNRFgiFqbW3DO9Cb3BEG75CVnbQWY7KIVcf/pWfAhE4vhUvf0aYv4ECrFJDntMyh6+CH8Cs38l4spDg/AL3fZpq6Wc3xf63067XoRV3O/rr51eyZ0780hgL+A2OOtU2H9TH/4zkfC7DOTPousXbafZXS+KIsp2r+/n9ydwzzrt1ge/7kh/Aqhq7gVnF2iS3z4b+Fbs/PvAis3E3SBdLTrObl1ANwBV0PUyhlfQDwpI907f7d8orTX89lF+fyToMpEw34nGAzoJVEA39OH/4deLFh9uRuTYOaCHxsMlLA8ObFdTxqxs5ypjtzQqY5Gw4u/tpBJ0/SKrrgbxHFBEPIPjJDW+Ij7/zYdA91CEo/nh9U4OdNSxugh34fqXwL1MvFLn0luBNYH1/f6IhLjnwqBO4oOAMzNoXidDGKMzCb1PrHKKqASC8iEQJf45FAol68pxj/uHPFjzIMJMYNnI/ga1h74IY/xInPH+9CRcufm831/Th9tsaLyL8iv6h056cO8IfLL5X9AZhFquIWRt0zeuWkESIeRXEZWA+RAIk5qbyPF1Q7WACB8T4ed+exjwHWBiJMg9wIci7IAbTHAL8II/Fx8QMNKvb/Prf/HxToQrTo9c02jU085N/BSj67F5Amm0XAloYD4EomgAY3CTKElX7Y18lUYBRWpNLFOjzTVZuBs4wG8P+oqIjMwRGDQks9ZctFYsrkVWHkVYi4ERZj+AXWqmRUbB4AlevUio5RpC1ja7rf6FmyWE/GrKn4CaD4FOoPZQbVgJAPdGtteAoQ7gRZgBfKA6tIL3M23jlcdb0SAJaR5SR88XItvRPgYby2+ETEdaP+hqU9IhtLcl0U5dImwnwkhgd+CvwEqx8yNEuEeET/j9VQfO9gNclBL1d4Hj/TUqwmqRc29CXZMLR+T6Eb5/wDNi8DBOI9RyDUFq8y8gX7V5Ail0dSXQyYiwjQi75gi/hAizgeuAr+G+BB4CVowF3cAvR4uwEQOjc2psKsJnRfilCPNFWMxXKnH6Y/tfzKo1A9GvhN0LjNcwjBilVAIS8RsgIuNFZK6I3Cwix7cz3aztbf5Bt3o7tQxGtxn8Bj5UD+4h+3sRfpISJj4U8zoGZsLWmuveBvYRYU0RlvQP8438uX1wk/winfhTaxt/APYFlgQ+AN6NpLtoiGcs/Z+m/Z7WmdowRC8RQjtyGuFqO9v6BFJoeyUgMT8CuOn9p6nqFGAzEWnb6JUseMNhtwD3xWzKtBrvGiKsnHB8GPAN4Esp143HPXg/wNlj+nf/lq8i7CPClj7oayIc469ZAtg8Ek3NamatieZhYD7wJ1wHfisUMZfAMMrGRgelUMaXQNyPwHXAdSIyHDfao23zCeq1t3kTw98G+oCfAVsDZ4qwT/50WEWE9UQ4WIQrRLgANzzydH9+jIgb8gisDX1jgK+L8ICIGwXjbd3sCHwWNzFqkion49r1a8MeLwBu9hYwlwS+7o/vlyItbjlxap2fsXS4be/9VQsIihDakdMIV9uhZcwf6Uh/Ak2NDsqDujnL79Xso6jqbSKyMq4p4nHgnaLS8k0V6wPTgE/DZRNEuBd4BnjWr5/B/e6f4sasb6zKU/76HYCrRVhClXP8seHAQtXkGyzCXrhK5DngAVzzyljge8BNvgP1e8CWInwTOBJevhk3Hv5D3AP+QuASYDXgv4AbVfm7T+Ju4NSUn7yCCE+RPgLoT8AC4JTY8Xtx/QKLUOV1kXefrhOXYXQi/gug++YJiLA9MFqVKyPHBDfnZhVVHvbH6vvJyGl/4kZgcmR/GjAPNzrkYmCsPz4D94b9uUjYm/x6YuTY/wA7t2L/AnQZ0L1AZ4M+BfoY6Jmgu4PuALo/6DGgp4JeBnqbtxVzAKgkxLc26OM+3FOgC0B/lGLjYyPQF0E3Tjl/CuijoA/49fOgr4Ee5c9vAPqw377d2zg5FfTrkTh+6I//OmLrZnbM9s39cRs6MR1xWzmPptjdGZbBzk6XL0PLWBlLVel24kI+20GL+Xu7ZQm6fh7VBToKdHgT8ezvNS/dINz8WHpLgh4Ufw6A3levfGUUxTTgLFx7cM1w3DI4a6J7AMvh3mRPqRNHrRL4JbCh3z4JmJo9c1DQ4aCTcQbGbgF9A/Qq0K+DrkvCg72JmzABdCvQ1XCVwkugoyLnBVeJvAD6xQbx7A46FvR4XOUyvhZX5KHbB3q9374WdI9IHJuAzgLdMfKwmg56eGT/jMj2iwk6RoAeEgnz3fjDLxI2gAdxlQupf5Z2LlYJ5Pp/dkolsAD01Cbi2S9SJh+uE26+DzM5Ev4Hke3DQL/UqFxn7RMo0o/AicAZItIHjNP8veP/BGbjmlxmABNU+awqP1XlQdWBZptm29tU+acqf1blCVXmAX9n8KzX83GdrzurcnGDeC5TZT7wY2AfVV4B2dyfXwiMw9nB2Rh4BNge12RVi+MOVWaocg0DfTgLcQbXarzEgHnvuQk63geeiBz6sV9HJ4nV8us+nCG3rYDa2OrTyc6GjYPkpb/4KDuYENqR0whX2+Fl2A6KNzkNJ+H/IMLYmje8lPzSyPaiOTMizBZhXxF2iqUXTSNqNuUbuD7GumSqBLRAPwKqeq+qTlHncezwtDTTbK7D1MNU+bgqR4EsqD1Q/TVxG9pF+ROYDUx3+4cegXtATgFZMmt8qrwIMnKoDXLZBJgDLAWH/Rh+chquMzjh98g2tQeiKh9G7OG/AqwDfR/CD55NSX8u7HeID/+6C3H+47EH7EawzJHAiqrcDDIOfvk74P8BKw+1vz/nN0Pt8cuyLPIUNv3woedD369jd70g0sp2u+3dd9N+fn8C/1i79fT7VcQ9VJPOw69XHLzfD37Geyz8G3D+ZXXS2yVaNt3wbpkK/dNxLSlX+f3as3ZcJPx6tfyBPy2y7uzTaN2fABX5EYjEn+kTsA2feaNwTUJrgM4F/WrB8X8U9BUyNGX5z7wDI9sKekCOtNaJXHsy6DzQwzJeG28+2TF+zId7BPQFv/1YwnULq2/2SVuqKmPVpNuJS87moMX9vZ1SQLoKemSd8+fFmoMU9FZcs/JiseN9fntp0E399g2gZyeUy99HrqstuxVVrlsZItozfgRUeRc3PPNMnDvEOQXH/yiuN18bBnZes+JNUG/nSOuhyO67qqytmsnmflJc15BsLmIzBkYfXRO7RoCv4GY111gqJYn7c2gRH7dhRKmNgKzr5yIH9fxhLwsgwlki/Kc/Ngk3CvLQWNhaWT0BuN1vf4pkK7i7inCs367N07ksh+a6tFIJBOtHoEbB7ZOzgU8Dx6lrY2+aJF2q2R7kqvxGdZBxNojM6M3Jgka6YkRnBa/k9UR1b+WPvaLKS/7YDX59Xi2QKuercjpuwMHWqrwRieM4v74Q96egxD6BpctKqBXCbXcvTVuWl6Ua3uTJ5WnDrBsiwqoii4ZZLxRhpAgbR86vI8I2DLS/f5WBCZs1dhAZYsIFuGD12IFlh4YBYJZf5/ntmWilEugpPwKq/AM3y/eXVWtJ4J9NXvdM4yADqHIk3sKoKs8nBHkpfkCV83F/jsOJOZVXZa4qf44cOlldJ7iosk+kIgE3FBngNEi0ZRTlwQbnE1Gt9ZUYgZP5yxeY7FbjPp4nARHOFuHLfrcPONJvfwgcBtzhZ/GPxJW3/lgUcWsBuzF4tv42ImwFEz8TC9eobBdeCeRsExvoE/D7u/pj84FzgMXb2xaY3q7Vi4tv77uhyWsngA5r4rozQe+IHdsH9L9BR7TwW7Yhed7GS/53jgedEPvtg/oiIscn1+KKhDmsUbtplWXMynausnK5v2dZ+9AGlZEM1yzlr5kfjwP0CNxw79r+2Jz9Tj+ObN8e2V4nZzyF9XX5P0pnICKqqtbu6/HmIy5VHWR/v6sQ4VPADRpr7/e/fSNcU9jFqotMYyswWdV5KPP7qCIirI2boDjNR3M4bgjeUQNhqiljVrazI8LjuNn1y6sO/fqMhNsTN38JWNQflRROgRVU3Re1COvhRjuCayJ8LRL8dQb3YY2DQc2ZefgHbRlSnYSQVr662pR0qG2nBesq7B4Gml83wcFp/gjeVjc35BORY30wqPMb/B9a3ZyPRf0p6jrEm+oUr5JA7xPQPm0ijPMmzlfFVQD3MdiVKf58tL/ukoHNfkSGeqXzBhsBfijCrr5CiDbJvAbOhaonPojhdppnwxDmwHR1JdAjdPU9VEXhnLRRQkM+Y1XZTnXQmxs4i6w17oqFfwRzWVkaIogIJ4oMdnbU4JrdgJtwJs5rkyGfB1aIBf0DsLgIy4qwRUJUT/r4VIS9RViWgVFqfwG29dt3xq47sI68dbP9ioAppw2PYcDVfntVnHXLPuDynPFkbtfrhcW3911RtY5r6oRhAAAMXklEQVQKf/tHM4S7DfT7kX3BmeHQWLhpbo22oqv539OedEFXB122pHuyEuhpDcLsGWmrnla7J7EwYyLbGzZo7x4JukqkTLSyfKe97fJVLqSWryr8CawF/ETdjOHMnrOMVIoa/9yJNOzQUmWSKt+K7CtDm4tQ5eqCtVWKCMNFOAr3Vvv/k5pCmox3mAjTvOOjONsAR6SlJcJBuCaaE3FDgPcT4fO4YZdHiLCUCBOBN0UWNbt8OSmuCBcBT4kwrk6Y3zWIo8bMjOG6iir8CXwUZ4LhBhFJs4NfCKG2nVqfQD7q6Gp2VMP5OF8WHUmj+yTC+jhHSTvj/ntnAjeKDPLdnCEdRoswWZzf6p+J0IczgX4lznAkImwksmh+xabQtwDoE+GZ2oNZhO+JsBzO78W3cGPevwtswcDkqNNwL4u1Ycuf9c01afbIrvDrz/t12vDeI4AvhND2nkx/1QLK9ycAPI0bjXEXcK2IXK2qL7RbR5dyDO7P3qvkGS++CP81EMykxiyIsAzOuOBn4A/bifA0Q/1kPItzjvR/gWOBs9UZKfyRCG8B/SJ8WpX7vZ+McarJM/y9gbMrcZMCX/XbF+GsBIzBvfHfipsMeIoI6wJbwD2zYdt7cO5JJwHX4h78j+N8VfxGlXdEmAesgZsBX2O1yPYFuGbjJDes7+B8WmfxRbKkKgu7z5tAgeRstyzCn8CwyLETgC2qbje1pfOWaLtxsfGi1fyewekyYDJ9BoNNph+JMy8+FfTLoN/E+a24BPRmnM+M1VLy7CugL+NMmr8H+gHebk1C2J+C/pEEe/g40+TP4fxjXISze/US6Dug432Yk0FnMmC751h/PtrerzhfG9H265dj+0dGti+N6VgrQ3v40T7ssdW3y1e5pJfrrAW0SH8C5+HGdwtwFTA+9D+oLb2zVFkJgE4EPdA/WF8GvRs3CW8HIv4sWktH18QZLByJc7Z0Vuz8RNCLfdqpTk1APwn6LzifGE+Bfht0bOT8Tv4BNMuv+0BfjcXxNZxjpUt9HAq6CugvIg+wTSPbExN0DM9YCSzeIFyXL6SW66wFdBbOheIbkUpgf2BuJMzGwPN14rjRr9fDNWHcABzezj8oORzWlPyHN12B6qq4EnjFVwAH4Ue8tDM/QFf2aY7x+2viPOUdBzo6Rzyb1CqAqDbQaTjHJ4/5h9GjdeI42GX9wLZf1ohsJ84QxlnCrYX5THIlwFS/fwDOquf9sXCNlrtBzy/+4dxXeSVQhT+B+1V1S1XdRlXPSEuzCJvrFOdPoCf2sfya6teV+BOA0Xuqsrcq54Ks1f7fKmvjXsj2dPu/Pg04VZVjQSZnjU+VO0A2GRr/iPdxz4y74b+OgQNPSosPTlon0kn6ZsQ+/svAut5XxTYp168Hf3p2sK+MPz4x1FfGiO1UmaPKE7DXifCtb+MMtu0w1LfEzy8evH/VkiA/8zuvwW9vqd73RZ598ydgiy2ZlqrKWHXp6u6gN+Jcs74IulTB8f8naGozcSTcAZEvgV0avf0nXL8Ezt/41v66yaDXgTZ8HkWuiS5jYvv3+rAKuhmuXyR+zal+/WTBb/LfqPxLIIWe8SdgGF3MlTj3prOBU7R4S6on4zzTNeJCWOQ2cZFpdNVsw4BVeVvdSKfF/P6tqmyvydZuh1we2z9Unbn2/4gc+7tfr6LK7TrUnPxCVf4N54p1I3/sJAas3zbLk6r8sMU46mL+BCrAdOUjVF1VUWR+qPIBzjz6+gz2F9EUcW2qLFRd5Ailno53VfmT323FX0fi5MkGefZ3nGFBgP/CVYgweO7B/gCqiebXHwEe8Oe/psorwHTg2ww4ofkVztIyDDKp3p+mqVZ5Ncy7VjF/AoZhnAR8TpX5VQvxtPIimbslQpU3Vfm6330+Uml9GAmzYOiVjMDZG/oksHn0hCrn+mse8vtfUeVgdb4yHmaRn4NBrMGAz4HazIbaV8pP/PrF7L8sG01XAqr6Eq52PBfXV/AcriIIBlXtr1pDEqYrH6Hqqoqi80OVl1SLmXRYkLbal8A6+dPnDhhqQiKHrvsi2+fjjMftlpLWB6r0qzK/TgX6JWLOlDx/daupC4CP+MrhcQYe/jVPfAuja1UmqDOJXTO5czgkGsvLTK4Zw6q6Rmz/cuDyVgQYhmHEqH0JvNLMxc1+0WjM34A63+JzmokrEsdbwMMJxxeK8DLOz/fTkVM1DUfjJuE+Gjteo/bFc5Eqr4qwDoNtYo0nY/51tRniUNuSTVc+QtVVFSHnR0Haal8CH9QNlYNA82wD2ORrsWMCrkNclVmq/Mofnw0cXwukA25Za18I0Q7oPt9JHnXdmkpXVwKGYXQktS+BpHb4rkGVF+Bv8dFYiVaOVLlXI9ZwI3yYcKzWj7AtcGsjHaW4lxSRYcAfVXVaw8D141E1F3xGG6mqjFnZHkCEFXCjY0YmDMXsakQ4D/hKvGkqJawSyyN/7HpVtvf7ywNLgzyUVr5K9ycgIt8TkT6/3CkiiZ0uhmH0LLWHWld/CaSQ+ZnsO5PjleSrRN7+VXkx1lTUfIItMMifgKp+W1W3xfVuz2PALnjhBNoOaLpyEqquqgg5PwrStgBc52kBcQHh5lmCrpa+BlUZr8oxea5peyWgjvcYOitvFvAjVU1q0zIMo3d5C2eWphe5q3GQYsnVJyAiNwLfVNVb/f404FSc44mrgOmqOl9EZuA6JX6kqlf4sDep6tZ+e3ngLFXdI5dYazc12oz1CRhVIoIAUuRXkIs3vXxl+hIQkWkichawFf6NXkSWwXn/ORpYHfcZcxyAqs5S50M4ralnf3+tYRiG4fFDQ9tuKiJK1uagKT7sm5FjuwJ3qOqlfvbwCaT7A42zA3BNZpVN0kHtgEFgujqDkPMjVG2mqw75zMIOMiX9Q+CkyLlRuIkLhbj9A2bivjpssaXqZWYRZdrKti2BLTNVlVYqgSr8CfTX205YazNxZznXSEtouhroKVVXjnyqVFeZS54ylCc/8uZJlvtUxb0yXcU+C6KL+RMwDMPoYTrNn8CcBtvxdbNxZznXSEt0O77OQ71rks6laUnSEV/nod41SefStCTpiK/zUO+apHPxY82kWQRzGmzH183G3ehcvf2k7fg6D/WuqacjTUt0O77OQ71r6ulI0xLdjq/zUO+apHNpWgaRd4joY8Deqnqb9yEwD9gDuBtn+vQvqjorc4RtJtRhd6YrH6HqqoqQ8yNUbaYrna72JwB8t2oBKZiufISqqypCzo9QtZmuFEoxIGcYhmGEiZmSNgzD6GGsEjAMw+hhrBIwDMPoYawSMAzD6GF6shIQkaVF5HoRuUlETq5aTxQRGSYiV1etA0BExonItSJym4gcXrWeOCHlVQiEXK4hrPsVctkuO596shIAvgKcp8609UQRWatqQTDUC1sAHAqco6qTgH1EZGTVgmoEmFchEGS5hiDvV5Blu4p86tVK4H6c/wNwllHHVaglyiAvbAEwCbjWb/8NWL9CLXFCy6sQCLVcQ3j3K9SyXXo+LVZWQiGhqtcCiMjXgHVU9W8VSwK8JSl4T0RCmbyxFM5GFMB8YGyFWgYRYF5VTqjlGoK8X0GW7SryqeO/BETkRhGZHNmfJiLzRORNEblYRMb64zO8c/tdRGQxEbkQ2AjYORBdn2uHjla0Aa8z8OcYB7wckLauJtRy3aS2ni/bQZfrrGZMQ1uAacBZOB8GNfPWy+Bu5h7AcsAlwCkJ1x4OfCc0XZE4bgpBG3AU8EVgONCHn2EegrZ251VVS6jlulVt7b5foZbtTijXnfwl0Iq3synATv4tpU9EPhaIrhrt+hTMq2027sHyF1wnWjs/UZvNt1CaF4oi1HLdqrYavVa2gy/XHdsnoKozAERkh8jhDYE7I/v3ARNEZIyqvhW5dt8QdUXi+FQg2l4Dtm2HlgK0veWva0teVUWo5bpVbZE4eqpsd0K57uQvgSTGAm/UdlT1XWAB1Xf6hKoLTFsnEHI+mLb8BKWr2yqBUL2dhaoLTFsnEHI+mLb8BKWr2yqBKrydZSFUXWDaOoGQ88G05ScoXd1WCfwe2FpEthXn+ewE4PyKNUG4usC0dQIh54Npy09QurqqEtBAvZ2FqgtMWycQcj6YtvyEpss8ixmGYfQwXfUlYBiGYeTDKgHDMIwexioBwzCMHsYqAcMwjB7GKgHDMIwexioBwzCMHsYqAcMwjB7GKgHDMIwexioBwzCMHsYqAcMwjB7mfwEehh9QZJPpZQAAAABJRU5ErkJggg==

jrleeman:
Hmm. I just compared with SAC and it matches @cja12. I'm not sure why your plots show that result though. Attached is the SAC spectrum, the one normalized by dividing by the sample rate, and one by 0.5*npts in that order. I'll dig into this some more, but thought that these were food for thought.

sac

cja_norm

kadmiel_norm

shaharkadmiel:
Hmmm indeed! What is the signal used for this fft ( @jrleeman )?
In order to compare the `fft` result of my signal (the one described above, superimposing four sine waves) with the result achieved with `SAC`, I wrote out a sac file:

```python
trace = obspy.Trace(signal)
trace.stats.delta = delta
trace.write('signal.sac', format='SAC')
```
read it with `SAC` and plotted it:
```
SAC> r ./signal.sac 
SAC> fft 
 DC level after DFT is -3.9727e-12
SAC> qdp off
SAC> ylim 1e-17 1e4
SAC> plotsp
```

The result is indeed different from what I get with numpy's `rfft` however the peaks are closer to what I get with @cja12 's approach.
I am also bothered by the fact that the results are so different with numpy's `rfft` noise slightly above 10**-13 but SAC's `fft` has only few drops to such small values.

image

jrleeman:
The signal I showed was just a .sac file I had convenient (@shaharkadmiel). I've lined out a bunch of thoughts and tried to prove which way is right to myself in the notebook at https://gist.github.com/jrleeman/84b4afcbed95e715ca0c . Thus far, I think that the norming @cja12 suggested is right, but looks to still have a factor of 2 missing? I also talked with @sanandak about the amplitude issue. We both feel that the amplitude on the spectra won't exactly match that that created the waveform, but haven't been able to formally show why. This is proving to be quite the educational experience and forcing understanding on me! It also proved that the MATLAB example uses the wrong normalization, but gets the right answer because of a fluke (in the notebook).

krischer:
Hi guys,

the units and normalization of spectral (density) estimates are a very common problem. Good and short reading: www.bksv.com/doc/bo0438.pdf

We can likely also avoid to reinvent the wheel here by using a thoroughly tested implementation we already have in ObsPy:

http://docs.obspy.org/master/packages/autogen/obspy.signal.spectral_estimation.html#module-obspy.signal.spectral_estimation

There are also a couple of spectral estimators in scipy we could readily use:
http://docs.scipy.org/doc/scipy/reference/signal.html#spectral-analysis

Cheers!

Lion

shaharkadmiel:
Hi @krischer,

`scipy.signal.periodogram` yields similar result to Numpy's `rfft`.
`matplotlib.mlab.psd` yields a nice and smooth spectrum resembling the overall shape of SAC's `fft`. I think I am going with `matplotlib.mlab.psd` padded to the next power of 2 (psd, next 2**).

```python
from matplotlib.mlab import psd
from scipy.signal import periodogram

fig, ax = plt.subplots()

amp, freq = psd(signal,NFFT=signal.size,Fs=1/delta)
ax.loglog(freq,amp, label='psd, signal.size')

amp, freq = psd(signal,NFFT=2**15,Fs=1/delta)
ax.loglog(freq,amp, label='psd, next 2**')

freq, amp = periodogram(signal,fs=1/delta)
ax.loglog(freq,amp, lw=2, zorder=0, label='periodogram')

ax.grid()

ax.axis('tight')
ax.legend(loc=(0.1,0.2))
```

image

cja12:
I am traveling and can explain more after I return home, but I think that the sine-wave test is more suitable for a power-density spectrum. @jrleeman is working on a spectrum.  We use power spectral densities when looking at "stationary" signals like noise. When we look at transient signals carrying information on seismic source processes (e.g. P-wave, S-wave, etc.), we want to estimate the spectrum. Also, when we want to convolve or deconvolve, we want to deal with spectra, not power-density spectra. So @jrleeman is correct to use an FFT, not a power-spectral density estimator. Follow the physical units, the spectrum must have units equal to [original units] x dt.

cja12:
One other item, when you do want to estimate the power-density spectrum for a stationary time series, I agree with Lion, use the built-in estimators.

shaharkadmiel:
Hi @cja12 , I agree with you. Then how come the result of SAC's fft is so different from Numpy's/Scipy's fft and so much closer to the PSD estimators?

I would think SAC, which is a tool oriented for seismological signals, transient signals, would produce the spectum, not the power density. Moreover, SAC has a built in Power Density Spectra estimator (https://ds.iris.edu/files/sac-manual/commands/spe.com/pds.html) different from the FFT command (https://ds.iris.edu/files/sac-manual/commands/fft.html).

cja12:
SAC's FFT is a simple spectrum, and I believe that SAC zero-pads the signal to the next power of two. The comparisons by @jrleeman show good consistency with SAC's FFT, small differences are likely from the zero-padding. And you're right, SAC includes some PSD estimators for approximately stationary time series like noise. 

Perhaps the difference between your sine-signal spectrum and SAC's is a result of the length of time series used in the computation. In your initial test, all the sine waves fit perfectly (zero at the beginning and at the end) in the 200 second sample. When they don't fit so well you get spectral leakage. To see the effects of spectral leakage change the duration of your sine wave signal to 200.01 seconds and compute again with python. Or try 205.0 to see the leakage increase for the 1/2 Hz signal. 

These computations make the periodogram look "unstable" - but the variance of the periodogram in all cases is actually equal to the value of the periodogram. Other PSD estimators based on classic windows, etc. reduce the variance while sacrificing some frequency resolution.

One final issue is that to get a PSD - the periodogram, you need to use the square of the amplitude of the FFT result (the power in PSD is the signal power, the square). I think you would want

    periodogram = np.square(np.abs(np.fft.rfft(signal))/npts)

But specific definitions of the PSD vary. You might, for example, need to include a factor of two for negative frequencies depending on the definition. For PSD's use one of the built-in functions, they have more options for reducing variance in the estimate.

Numerical Recipes (Chapter 13) is a very readable, practical introduction to these issues. You can find it online.

krischer:
@cja12 That all sounds very good and reasonable to me.

So I guess for now we should aim to create `plot_spectrum()` methods on the `Stream` and `Trace` objects and maybe `plot_psd()` methods which could just wrap `matplotlib.Axes.psd()` (then we don't have to deal with all the parameters that go into it but just delegate to matplotlib).

If someone wants to give it a try, we need:

* A `plot_spectrum()` function in `obspy.imaging` (extending the one by @jrleeman should work) that can plot any number of traces at once + tests.
* A `plot_psd()` function thats just a thin wrapper around `matplotlib.Axes.psd()` but also for multiple traces at once + tests.
* Methods on the `Stream/Trace` objects.

shaharkadmiel:
Thank you @cja12 for your detailed and reasoned explanation.

cja12:
Scaling (and frequency "packing") is tricky - I have to revisit every few years and/or every time I start using a new set of FFT routines.

jrleeman:
I had a start at writing some tests, but should probably look at adding the methods to ```Stream/Trace```. I think we should leave the arguments/returns of figure, axes, and the spectrum frequencies/amplitudes. It allows the user a lot of versatility and reduces them having to re-plot for their specific widget.  I'm supposing that the multiple trace support would just be introduced in the ```Stream``` method by calling ```plot_spectrum``` in ```imaging``` multiple times?

shaharkadmiel:
I thought perhaps to add the frequency and amplitude vectors as variables to the Trace object. Just like times() (actually a method) and data. That would mean that the user would access them through trace.freq and trace.amp where freq could be eithe an explicit array or a method to line times() is. Perhaps to would also be useful to add nfreq, df, and max_spect, like npts, delta, and max. 

Another option is to have a boolean argument ret_spect, like in np.linspace for example for the user to specify if it is favorable to return the spectrum or it is enough just to plot.

shaharkadmiel:
I thought perhaps to add the frequency and amplitude vectors as variables to the `Trace` object. Just like `times()` (actually a method) and `data`. That would mean that the user would access them through `trace.freq` and `trace.amp` where `freq` could be either an explicit array or a method like `times()` is. Perhaps to would also be useful to add `nfreq`, `df`, and `max_spect`, like `npts`, `delta`, and `max`. 

Another option is to have a boolean argument `ret_spect`, like `np.linspace` has `retstep` for example for the user to specify if it is favorable to return the spectrum or it is enough just to plot.

megies:
> I thought perhaps to add the frequency and amplitude vectors as variables to the Trace object. Just like times() (actually a method) and data.

Actually we plan to implement this by a new object called something like `FrequencyTrace`, because it won't fit 100% into our current time series oriented `Trace` object, see #837.

cja12:
Generating a plot without returning the spectrum is not a bad approach given that spectral objects are being developed for a future release. I motivated the plotting effort by noting that there was no simple command to get a plot of an amplitude spectrum. I expect that the FrequencyTrace or whatever it ends up being called will drive a revision or re-implementation of display functions.

megies:
> Generating a plot without returning the spectrum is not a bad approach [...]

Of course, I agree. I just wanted to mention this because I think all the derived data like frequency vector and power vector should probably not be attached to the trace. We want to at some point implement this by having a transformation routine that yields a frequency domain representation of the original trace (and vice versa). Having both representations in the same object brings redundancy and thus potentially inconsistencies (e.g. when time domain modifications happen to the trace afterwards).
And yes, we can refactor it at a later time, in any case.

Good to see such a lively discussion after this got prompted at the IRIS short course.

jrleeman:
I added a test outline and the return spectrum option as discussed earlier. Is the plan to implement this "as-is" with some improvement or to wait for the trace refactor?

krischer:
I guess the trace refactor is still a bit in the making so I'd personally be fine to merge this with some further work. This is a small feature and will be very easy to merge into the trace refactor once that is done.

Other opinions?

Khalilsqu:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.periodogram.html#scipy.signal.periodogram

periodogram is available in scipy

PR #1107, opened 2015-08-09
Future 0.15 problems

megies:
Here's  a fix for a problem in reading xseed files that starts popping up with new `future` release 0.15. Probably there's more changes needed prompted by that release..

### DIFF
--- a/obspy/core/util/decorator.py
+++ b/obspy/core/util/decorator.py
@@ -11,11 +11,12 @@ Decorator used in ObsPy.
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
-from future.utils import native_str
+from future.utils import native_str, native_bytes
 
 import functools
 import inspect
 import os
+import re
 import socket
 import unittest
 import warnings
@@ -279,10 +280,17 @@ def map_example_filename(arg_kwarg_name):
             # check kwargs
             if arg_kwarg_name in kwargs:
                 if isinstance(kwargs[arg_kwarg_name], (str, native_str)):
-                    if kwargs[arg_kwarg_name].startswith(prefix):
+                    arg_ = kwargs[arg_kwarg_name]
+                    if isinstance(arg_, (bytes, native_bytes)):
+                        # try to find encoding specified in document
+                        match = re.match(r'[^>]* encoding=(["\'])(.+?)\1',
+                                         arg_)
+                        # use detected encoding or default of UTF-8
+                        encoding = match and match.group(2) or "UTF-8"
+                        arg_ = arg_.decode(encoding)
+                    if arg_.startswith(prefix):
                         try:
-                            kwargs[arg_kwarg_name] = \
-                                getExampleFile(kwargs[arg_kwarg_name][9:])
+                            kwargs[arg_kwarg_name] = getExampleFile(arg_[9:])
                         # file not found by getExampleFile:
                         except IOError:
                             pass
@@ -295,11 +303,19 @@ def map_example_filename(arg_kwarg_name):
                 else:
                     if ind < len(args) and isinstance(args[ind], (str,
                                                                   native_str)):
+                        arg_ = args[ind]
+                        if isinstance(arg_, (bytes, native_bytes)):
+                            # try to find encoding specified in document
+                            match = re.match(r'[^>]* encoding=(["\'])(.+?)\1',
+                                             arg_)
+                            # use detected encoding or default of UTF-8
+                            encoding = match and match.group(2) or "UTF-8"
+                            arg_ = arg_.decode(encoding)
                         # need to check length of args from inspect
-                        if args[ind].startswith(prefix):
+                        if arg_.startswith(prefix):
                             try:
                                 args = list(args)
-                                args[ind] = getExampleFile(args[ind][9:])
+                                args[ind] = getExampleFile(arg_[9:])
                                 args = tuple(args)
                             # file not found by getExampleFile:
                             except IOError:
--- a/obspy/core/util/decorator.py
+++ b/obspy/core/util/decorator.py
@@ -11,7 +11,7 @@ Decorator used in ObsPy.
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
-from future.utils import native_str, native_bytes
+from future.utils import native_str
 
 import functools
 import inspect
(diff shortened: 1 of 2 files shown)


krischer:
That only works if `data` is a (I guess system encoding) encoded unicode string; in other cases this will fail. I'll have a look later on - the issue might be higher up the chain.

QuLogic:
What test triggered the problem? I've run with 0.15 without issue (and Travis seems to agree, I guess.)

megies:
Sorry, only just now got a chance to come up with a snippet to reproduce. It's a behavior change in between `future` `0.14` and `0.15`. I can only reproduce this when the request actually goes across network..

```python
                        unicode_literals)
from future import standard_library

import os

    import urllib.parse
    import urllib.request

from obspy import xseed  # NOQA


fname = "dataless.seed.BW_FURT.xml"

urls = ["file://" + os.path.join(os.path.dirname(xseed.__file__), "tests",
                                 "data", fname),
        "https://raw.githubusercontent.com/obspy/obspy/releases/"
        "obspy/xseed/tests/data/" + fname]

for addr in urls:
    print(addr)
    response = urllib.request.urlopen(addr)
    doc = response.read()
    print(xseed.Parser(doc))
```
```
file:///home/megies/anaconda/envs/stable/src/obspy/obspy/xseed/tests/data/dataless.seed.BW_FURT.xml
Networks:
	BW (BayernNetz)
Stations:
	BW.FURT (Furstenfeldbruck, Bavaria, BW-Net)
Channels:
	BW.FURT..EHE | 200.00 Hz | Lennartz LE-3D/1 seismometer | 2001-01-01 -  | Lat: 48.2, Lng: 11.3
	BW.FURT..EHN | 200.00 Hz | Lennartz LE-3D/1 seismometer | 2001-01-01 -  | Lat: 48.2, Lng: 11.3
	BW.FURT..EHZ | 200.00 Hz | Lennartz LE-3D/1 seismometer | 2001-01-01 -  | Lat: 48.2, Lng: 11.3
https://raw.githubusercontent.com/obspy/obspy/releases/obspy/xseed/tests/data/dataless.seed.BW_FURT.xml
  File "/tmp/bug_urlopen_parser.py", line 24, in <module>
    print(xseed.Parser(doc))
  File "/home/megies/anaconda/envs/stable/src/obspy/obspy/xseed/parser.py", line 107, in __init__
    self.read(data)
  File "/home/megies/anaconda/envs/stable/src/obspy/obspy/core/util/decorator.py", line 299, in new_func
    if args[ind].startswith(prefix):
  File "/home/megies/anaconda/envs/stable/lib/python2.7/site-packages/future/types/__init__.py", line 163, in wrapper
    raise TypeError(errmsg.format(mytype))
TypeError: argument can't be <type 'unicode'>
```

krischer:
Looks to a be a problem with the `map_example_filename()` decorator. If it is passed a bytestring it will test with string operations which fail. The best fix would probably be to not have the decorator do anything if it is being passed a bytestring but only be active for normal strings.

megies:
You're right it's the `.startswith("/path/to")` acting on the encoded bytestring that's exploding.

But still if that's changed we end up having to specify some encoding to the `Parser`. I guess `urllib` is using the encoding specified in the xml header.. but can we be sure? And why did this behavior change in `future`..

krischer:
That's very messy. We should just make sure that whatever gets to the parser is a bytestring/python 2 string or a file-like object in binary mode or otherwise raise an error. Its just too much pain to deal with unicode encodings and feeding that to lxml and so on.

krischer:
`urllib` will use whatever encoding is in the HTTP response (which the webserver serving the file might well take from the xml header - or just assume its utf-8).

krischer:
So this is really nasty as it essentially invalidates all the checks we have for strings or bytes across the codebase...

```python
In [12]: from future.builtins import *

In [13]: import future.types

In [14]: import future.utils

In [15]: xx = future.types.newbytes("asd")

In [16]: isinstance(xx, future.utils.native_str)
Out[16]: True

In [17]: isinstance(xx, future.utils.native_bytes)
Out[17]: True

In [18]: isinstance(xx, str)
Out[18]: False

In [19]: isinstance(xx, bytes)
Out[19]: True
```

we do quite a lot of checks like `isinstance(thingy, [str, future.utils.native_str])` which are then not valid anymore.

I think the best solution is to just provide a custom urlopen method that does not return the `newbytes` object as it confuses so much of our codebase.

krischer:
We should probably also add two extra test runners (one for Python 2 and one for Python 3) to travis that test against the latest repository version of `future`...

megies:
> We should probably also add two extra test runners (one for Python 2 and one for Python 3) to travis that test against the latest repository version of future...

:+1:

megies:
Another try to get this fixed in a proper way..

The actual problem is caught by `seishub` tests which are not run by CI (and test server is only internally available here..), so...

before
 - Py2: http://tests.obspy.org/31085/ (fail)
 - Py3: http://tests.obspy.org/31088/ (pass)

after
 - Py2: http://tests.obspy.org/31087/ (pass)
 - Py3: http://tests.obspy.org/31090/ (pass)

megies:
Travis fails are unrelated.
Also ran tests for `fdsn iris seishub seedlink` to see if there are negative sideeffects in those that are not covered by CI, looks good..
 - Py2: http://tests.obspy.org/31105/
 - Py3: http://tests.obspy.org/31111/

obspy/core/util/decorator.py
```diff
@@ -279,10 +280,17 @@ def new_func(*args, **kwargs):
             # check kwargs
             if arg_kwarg_name in kwargs:
                 if isinstance(kwargs[arg_kwarg_name], (str, native_str)):
-                    if kwargs[arg_kwarg_name].startswith(prefix):
+                    arg_ = kwargs[arg_kwarg_name]
+                    if isinstance(arg_, (bytes, native_bytes)):
+                        # try to find encoding specified in document
+                        match = re.match(r'[^>]* encoding=(["\'])(.+?)\1',
+                                         arg_)
```
krischer: I'm confused. What happens if the file is not an XML file and thus has no encoding specified?

megies: Well, then the behavior is unchanged.. this was prompted by a xml file getting returned encoded.

I don't know if we want to start trying to extract encoding information in all kinds of ways, unless we actually encounter a problem.. I could decode with default of UTF-8 in those cases?

krischer: I still don't get it. The `map_example_filename` decorator should really only map` /path/blub` to the actual file. Should any encoding and what not not be handled elsewhere?

megies: Ok, yeah, no need to decode actually, can just switch to ’re’ for the matching.. took a while to see whats failing and then a while to see how to best handle this.. ;)

megies:
Py2: http://tests.obspy.org/31122/
Py3: http://tests.obspy.org/31124/

Looks good to me, ready to merge when CI passes, I think.

krischer:
:+1: Looks good to me now!

PR #931, opened 2015-01-20
Change to response removal in PPSD

aringler-usgs:
I an purposing a bug fix to remove the response in the PPSD routine in the frequency domain.  This gives a better estimate of the very long-period noise levels.  It is also the same approach McNamara takes.

I have changed the three files that are used for testing the PPSD routine to reflect the new changes to the algorithm.  

Please let me know if there is anything else you want changed on my end.

Thanks,
Adam

PR #1108, opened 2015-08-10
PPSD response removal updates

QuLogic:
A continuation of #931, rebased with one minor fix commit.

CC @aringler-usgs, @megies, @krischer.

List of changes in this PR:
 - improved response removal algorithm (mostly better results in long periods), removed `water_level` kwarg
 - new `metadata` kwarg, deprecated `paz` and `parser` kwargs
 - metadata can now be provided as any of `Inventory` (StationXML) / `Parser` (SEED, XML-SEED) / filename of RESP file / PAZ dictionary
 - full response (i.e. all response stages including FIR filters etc.) is used in response removal now, whenever possible (i.e. in all cases but a PAZ dictionary)
 - change default colormap to new ObsPy default (matplotlib's viridis)
 - new `save_npz`/`load_npz` routines, should work much more stable across changing obspy/matplotlib/etc versions
 - new option `.plot(..., cumulative=True)` for a cumulative 2D plot, i.e. non-exceedence percentage


obspy/signal/spectral_estimation.py
```diff
@@ -617,6 +607,26 @@ def __process(self, tr):
         # working with the periods not frequencies later so reverse spectrum
         spec = spec[::-1]
 
+        # Here we remove the response using the same conventions
+        # since the power is squared we want to square the sensitivity
+        # we can also convert to acceleration if we have non-rotational data
+        if self.is_rotational_data:
+            # in case of rotational data just remove sensitivity
+            spec /= paz['sensitivity']**2
```
krischer: missing whitespaces around `**`. Not sure why its not caught by flake8.

obspy/signal/spectral_estimation.py
```diff
@@ -617,6 +607,26 @@ def __process(self, tr):
...
+                                    self.delta, nfft=self.nfft)
+            resp = resp[1:]
+            resp = resp[::-1]
+            # Now get the amplitude response (squared)
+            respamp = np.absolute(resp * np.conjugate(resp))
+            # Make omega with the same conventions as spec
+            w = 2.0 * math.pi * _freq[1:]
+            w = w[::-1]
+            # Here we do the response removal
+            spec = (w**2) * spec / respamp
```
krischer: same whitespace issue.

krischer:
Looks mostly good to me. In the long run this should really be hooked up to instrument removal facilities we already have.

obspy/signal/spectral_estimation.py
QuLogic: It's not really a requirement for PEP8; not sure we do that consistently either.

QuLogic:
@aringler-usgs Are you okay with this now?

aringler-usgs:
Hello Elliott,

Yes.  Do you need me to do something on my end?

Best,
Adam

On Wed, Aug 26, 2015 at 7:03 PM, Elliott Sales de Andrade <
notifications@github.com> wrote:

> @aringler-usgs <https://github.com/aringler-usgs> Are you okay with this
> now?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/obspy/obspy/pull/1108#issuecomment-135225266>.
>



-- 
Follow USGS Seismic Instrumentation on Twitter: @USGS_Seismic

QuLogic:
No, I think it's up to @megies and @krischer now.

megies:
I think I'll clean up the metadata handling as well, while we're making changes to PPSD (and also use all response stages).. should be within a few days, I hope.
I'll also add a test that checks against current state of results (to fix them).
After that we can make a final review.

ThomasLecocq:
then I can work on the writing of the `msnoise-ppsd` plugin :-)

https://fallmeeting.agu.org/2015/abstract/msnoise-not-only-dvv-a-framework-for-continuous-seismic-data-analysis/

megies:
OK, some additional major changes, since we're already working on PPSD:

 - metadata are now handled in a unified fashion using `metadata` kwarg,
   old kwargs `paz` and `parser` are deprecated (most old codes should
   still work, issuing a deprecation warning)
 - metadata are now also accepted as
    - an Inventory object (to enable metadata acquired from StationXML
      files or FDSN web services)
    - a filename/path of a local RESP file (since we can use RESP files via
     evalresp.. so why not enable it in PPSD)
 - full response information (i.e. all response stages) is now used in
   calculation of instrument response whenever possible (i.e. in all
   cases other than a PAZ dictionary)

I added a test that fixes the current results, so it'll fail if anything at all changes in there.

@aringler-usgs, maybe you can double check if this looks good for you.

I think the only question that remains is:
 - Do we need a kwarg to enable users to only use PAZ stage when using Inventory/RESP/SEED station information?

Here's a comparison of images from
 - [IRIS mustang](http://service.iris.edu/mustang/noise-pdf/1/query?target=IU.ANMO.00.LHZ.M&starttime=2010-01-01&endtime=2010-01-02&format=plot&plot.interpolation=none&plot.period.min=1&plot.period.max=1000&plot.power.min=-190&plot.power.max=-110)
 - obspy using full response (which is identical for RESP file, StationXML and dataless SEED)
 - obspy using PAZ dictionary only:

(legends are cropped to get the images scale equally, x is 1 to 1000 seconds, y is -190 to -110 dB)

iuanmo_iris
iuanmo_inv
iuanmo_paz


And here's some code to do these calculations for any types of metadata for IU.ANMO data in our test suite:
```python
import os

import obspy
from obspy.signal import PPSD
from obspy.io.xseed import Parser

directory = os.path.join(
    os.path.dirname(obspy.__file__), 'signal', 'tests', 'data')

# Read in ANMO data for one day
st = read(os.path.join(directory, 'IUANMO.seed'))

# Use a canned ANMO response which will stay static
paz = {'gain': 86298.5, 'zeros': [0, 0], 
       'poles': [-59.4313, -22.7121 + 27.1065j, -22.7121 + 27.1065j,
                 -0.0048004, -0.073199], 'sensitivity': 3.3554*10**9}
inv = read_inventory(os.path.join(directory, 'IUANMO.xml'))
parser = Parser(os.path.join(directory, 'IUANMO.dataless'))
resp = os.path.join(directory, 'IUANMO.resp')

# Make an empty PPSD and add the data
for metadata, suffix in zip([paz, inv, parser, resp],
                            ["paz", "inv", "parser", "resp"]):
    print(suffix)
    ppsd = PPSD(st[0].stats, metadata, db_bins=(-190, -110, 1.0))
    ppsd.add(st)
    ppsd.plot("/tmp/IUANMO_%s.png" % suffix,
              period_lim=(1, 1000), show_coverage=False)
```

megies:
And yet another major change in PPSD:
 - new save/load mechanism in numpy npz format, should work much better (aka. at all) in case python module versions have changed in between saving/loading the PPSD pickle files..

Please give it a try and comment..

megies:
Rebased onto current master!

aringler-usgs:
I wonder why the spectra drops below the NLNM right after the FIR filter correction?  Is there something going on with the interpolation?  This looks like an issue with the mustang version...  Or am I missing something?

untitled drawing

megies:
No idea, that's the IRIS mustang result..

megies:
I think we should change the colormap away from the PQLX default. I've been reluctant to switch away from PQLX default for years, but It's just really a bad colormap, and we can have an easy means for people to switch back to it, if necessary. Now is a good time to make a change, with all the changes in PPSD anyway and also all the colormap changes in ObsPy.

This is the comparison of PQLX vs. Viridis (which we use now for all sequential plots as a default).

Any strong opinions against switching to a new default?
PQLX colormap can be used simply by [these two lines of code](https://github.com/obspy/obspy/commit/45febc6416ea1e7f918ca5d2e06bc48e4be7ef92#diff-f91123410e50ce0ac26c7c3aefe942eaR3).

figure_1
figure_2

I've also added a concise list of all changes in this PR at the top..

megies:
...aaaaand, one more addition: new option for a cumulative plot, which nicely displays non-exceedence percentage. kind of like the percentiles just in a continuous (or well, semi-continuous, since I think that 5% color steps are pretty much ideal for visual interpretation) fashion:

```python
ppsd.plot(cumulative=True)
```

figure_8b

krischer:
> I think the only question that remains is:
> 
>     Do we need a kwarg to enable users to only use PAZ stage when using Inventory/RESP/SEED station information?

IMHO we don't need this. If the full response is given just use it.


I also really like the use of `viridis` - I feel like it is much easier to "read" then the old colormap. I never noticed that red and pink are on opposite ends of the scale on the PQLX colormap which is really misleading.

obspy/imaging/tests/test_ppsd.py
```diff
@@ -43,6 +43,20 @@ def test_ppsd_plot(self):
             plt.draw()
             fig.savefig(ic.name)
 
+    def test_ppsd_plot_cumulative(self):
+        """
+        """
```
krischer: empty docstring? ;-P

krischer:
Great work and big improvement to the previous state!

Looks good to merge from my point of view. Might require another rebase?

obspy/signal/spectral_estimation.py
```diff
@@ -355,17 +364,33 @@ def __init__(self, stats, paz=None, parser=None, skip_on_gaps=False,
...
+    def times(self):
+        return list(map(UTCDateTime, self._times_used))
+
+    @property
+    def times_used(self):
+        return list(map(UTCDateTime, self._times_used))
+
+    @property
+    def times_data(self):
+        return [[UTCDateTime(t1), UTCDateTime(t2)]
```
QuLogic: Maybe a tuple here?

obspy/signal/spectral_estimation.py
```diff
@@ -228,68 +243,71 @@ class PPSD():
...
+    The PPSD object supports saving to a numpy npz compressed binary file:
 
-    >>> ppsd.save("myfile.pkl.bz2", compress=True) # doctest: +SKIP
+    >>> ppsd.save_npz("myfile.npz") # doctest: +SKIP
 
     The saved PPSD can then be loaded again using the static method
-    :func:`~obspy.signal.spectral_estimation.PPSD.load`, e.g. to add more data
-    or plot it again:
+    :func:`~obspy.signal.spectral_estimation.PPSD.load_npz`, e.g. to add more
+    data afterwards or to simply plot the results again. Metadata has to be
```
QuLogic: s/has to/must/

obspy/signal/spectral_estimation.py
```diff
@@ -228,68 +243,71 @@ class PPSD():
...
 
     .. note::
 
-        It is safer (but a bit slower) to provide a
-        :class:`~obspy.io.xseed.parser.Parser` instance with information from
-        e.g. a Dataless SEED than to just provide a static PAZ dictionary.
+        When using metadata from an
+        :class:`~obspy.core.inventory.inventory.Inventory`,
+        a :class:`~obspy.io.xseed.parser.Parser` instance or from a RESP file
+        , information on metadata will be correctly picked for the respective
```
QuLogic: This comma probably shouldn't be on a separate line.

obspy/imaging/cm.py
```diff
@@ -40,15 +45,17 @@ def _get_cmap(name):
...
         name = name.rsplit(".npz", 1)[0]
     filename = os.path.join(directory, name + ".npz")
+    kwargs = lut and {"N": lut} or {}
     cmap = LinearSegmentedColormap(name=name,
-                                   segmentdata=dict(np.load(filename)))
+                                   segmentdata=dict(np.load(filename)),
+                                   **kwargs)
     return cmap
 
-
```
QuLogic: I don't think this line should have been removed.

megies:
Final rebase and squashed some low hanging micro-commits.. ;-)

Gonna merge when CI passes, any further polishing (if needed) can be done in master.

PR #1111, opened 2015-08-27
Bugfix for the Lanczos interpolation

krischer:
Fix + regression test for a bug in the Lanczos interpolation routine.

Really bad bug that resulted in a potentially quite significant time shift if the original sampling rate is not 1 Hz and the new and old start times are not identical. Most tests work with 1 Hz as its simple to reason about it so this bug was not caught...


krischer:
I'm merging this as I am fairly sure its correct and I need this.

megies:
> Most tests work with 1 Hz as its simple to reason about it so this bug was not caught...

Yeah, that's something that has caused bugs before..

PR #1113, opened 2015-09-07
Test on nightly Python builds on Travis

krischer:
Let's see if that works. If yes then we should be able to catch any issues with upcoming Python versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
   - 2.7
   - 3.3
   - 3.4
+  - "nightly"
 sudo: false
 install:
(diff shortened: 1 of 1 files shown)


QuLogic:
I didn't think conda provided nightly builds...

krischer:
Oops...It does not...I forgot we changed travis to conda a couple month ago - it would have worked if we would still use the travis python.

Still early here...sorry about that.

QuLogic:
I was thinking of it too, but I can only find build recipes and no hosted packages for 3.5 betas.

PR #1114, opened 2015-09-08
is_StationXML(): less rigorous (and faster) checking of file

megies:
Main reason for this change is to hopefully autodetect more StationXML files that have slight schema breaches but which we still can read into an Inventory object.


megies:
Travis fail is yet again caused by PythonCharmers/python-future#162.

PR #1115, opened 2015-09-08
more robust response plot (in presence of strange channels)

### LINKED ISSUE #1060: obspy.read_inventory doesn't work with stationXML
We enjoyed the Obspy presentation given today by Lion.  
I tried using the obspy.read_inventory on one of our stationXML files and it failed.
Here is the file: http://files.anss-sis.scsn.org/production/FDSNstationXML/CI/CIBAR.xml
Here is the error:
screen shot 2015-05-27 at 9 07 47 am

megies:
Closes #1060.

### DIFF
--- a/obspy/core/util/obspy_types.py
+++ b/obspy/core/util/obspy_types.py
@@ -411,6 +411,14 @@ class ComplexWithUncertainties(CustomComplex):
                                       upper_uncertainty=_upper)
 
 
+class ObsPyException(Exception):
+    pass
+
+
+class ZeroSamplingRate(ObsPyException):
+    pass
+
+
 if __name__ == '__main__':
     import doctest
     doctest.testmod(exclude_empty=True)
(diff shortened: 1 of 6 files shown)


megies:
Travis fails unrelated, see PythonCharmers/python-future#162.

obspy/core/util/obspy_types.py
```diff
@@ -411,6 +411,14 @@ def imag(self):
...
 
 
+class ObsPyException(Exception):
+    pass
+
+
+class ZeroSamplingRate(ObsPyException):
+    pass
+
+
```
krischer: Should maybe go in a new `obspy.core.exceptions` or even at a top level `obspy.exceptions` as people will import this if do this across the code base. Otherwise I am all for this change :-)

megies: Yes, considered other possibilities and I'm pretty indifferent where it ends up.. feel free to change.. I'm just not feeling much for making it top-level `obspy.exceptions`.

PR #1117, opened 2015-09-10
obspy.station.station.Station() does not require input parameter site

megies:
Hi 

I note that the class `obspy.station.station.Station()` by default sets the attribute site to None, however in the current stationXML schema (http://www.fdsn.org/xml/station/fdsn-station-1.0.xsd) the attribute Site of a station element is defined by:

        <xs:element name="Site" type="fsx:SiteType">
            <xs:annotation>
                <xs:documentation>These fields describe the location of the station
        		using geopolitical entities (country, city,
        		etc.).</xs:documentation>
            </xs:annotation>
        </xs:element>
And therefore probably should be considered mandatory. It seems that this is what is expected by  `obspy.station.stationxml.write_StationXML(...)` since the following code snippet will choke on this latter function rather than on creating an instance of the former class:

    #! /usr/bin/env python
    import sys,obspy 
    Sta = obspy.station.station.Station(code="STA",elevation=0,latitude=0,longitude=0)
    Net = obspy.station.network.Network(code="NET",stations=[Sta])
    Inv = obspy.station.inventory.Inventory(networks=[Net],source="test")
    obspy.station.stationxml.write_StationXML(Inv,sys.stdout)

Now personally I wouldn't mind not having to specify the site attribute but I guess then we should try to convince whoever is maintaining the stationXML schema (think it's Chad Trabant one shold talk to) to do the changes there as well as in `obspy.station.stationxml.write_StationXML(...)`

regP


megies:
Looks like you're right, we should make `site` mandatory, see attached patch.

obspy/station/station.py
```diff
@@ -47,8 +47,12 @@ def __init__(self, code, latitude, longitude, elevation, channels=None,
         :type longitude: :class:`~obspy.station.util.Longitude`
         :param longitude: The longitude of the station
         :param elevation: The elevation of the station in meter.
+        :type site: :class:`~obspy.station.util.Site` (or str)
         :param site: These fields describe the location of the station using
-            geopolitical entities (country, city, etc.).
+            geopolitical entities (country, city, etc.). If a plain `str` is
+            provided instead of a :class:`~obspy.station.util.Site` object, a
+            :class:`~obspy.station.util.Site` objet with this string as field
```
QuLogic: s/objet/object/

megies:
Hmm.. this PR changes the order of call args/kwargs and thus might break peoples' codes.. we want to push 1.1.0 ASAP, so I feel like we have to punt this to 1.1.1 unless we find a good solution soon..

krischer:
Rebased and force-pushed. A lot of tests fail.

krischer:
I'm actually not sure we want this - sure it is inconsistent with the StationXML standard but the Inventory object does not have to be limited to the StationXML capabilities. Also this will definitely break a lot of existing codes (see all the failing tests...).

I think careful documentation and nice error messages when writing to StationXML are another way to solve this issue.

If we merge this we can only have for major versions and not for bug-fix releases and even then I'd rather not have yet another breaking change to be honest.

megies:
Yeah, I agree that we should be able to work around this with clear error messages and it's more important to not change the API..

megies:
Changed base branch to `maintenance_1.1.x`. Should ideally rebase on this base branch, eventually, at least if we decide we want to go in this direction..

megies:
Not sure what we should do with this..? Punt to 1.2.0?

krischer:
As argued before I would not merge this as is. It will break existing codes without strong enough gains for us. We should rather enhance our error messages.

megies:
Hmm.. yeah, I'm torn on this one as well. Breaking codes is really bad, so you're probably right that having better error messages might be the less aggressive way to handle this.

megies:
I'm feeling a bit bad that we haven't handle this one yet, since @paitor reported these issues years ago in #1118, #1119 and maybe some more tickets..

markcwill:
What about making all the args to `Station.__init__` kwargs set to `None`, and then checking the ones that should be set in code? I've seen that in other projects with lots of input args. That way you can keep the order, and separate the `not None` validity checks to be either in init or as a separate method that gets called when QuakeML serializing? Basically remove all "required" args and if they really need to be required, raise your own error in `__init__`

I *think* this will work and not break API, haven't actually tried yet, no code b/c maybe should be different PR?

dsentinel:
I lean with @markcwill here. Especially in not requiring all fields stationxml requires at object init.
Often I want to create a empty object, and populate it's attributes later. I already have a fair bit of code that works around this by using some sort of valid defaults at initialization. I think it would be much cleaner to have these `None` so I can distinguish whether I have set it, or it is a value I made up so I could create the object at the time.
e.g. I create a channel with some common fields, make 3 copies, then adjust the individual fields; BH[Z12]

Is it considered that the internal representation (inventory objects) always represent valid StationXML?

I think validation is a  separate concern and should be a 'factored out' such that each object has a `validate()` and `inventory.validate` that walks and validates all children. Called by default at `write()` with a `write(validate=False)` option.

d-chambers:
> I think validation is a separate concern and should be a 'factored out' such that each object has a validate()

@dsentinel this is a very good idea (for both inventories and catalogs). Also, a mechanism to allow users to define custom validation functions would be useful. Each serialization method would potentially have its own validation function(s) true?

krischer:
I also agree with @markcwill, @dsentinel, and @d-chambers.

We cannot really enforce any additional checks in the `__init__()` methods. While these could be done in a way to not break API they would still break existing code and we only break backwards compatibility if we have to.

The `validate()` method would indeed have to be different for every file format as every format  has different requirements. We do not really want to associate our internal objects directly with a format (altough the the `Inventory` and `Catalog` objects are closely modelled after `StationXML` and `QuakeML`, respectively).

We could add an optional `validate_FORMAT()` function to each of the IO plugins, similar to the current `is_/read_/write_FORMAT()` functions.

krischer:
Yes - feel free to open a new issue to discuss more generic validation hooks in ObsPy. I would not merge this here but also not close it as OP still raises a couple of points (that could be dealt with with more generic validators).

megies:
Base branch will have to be changed either to `master` or to `maintenance_1.2.x` once 1.2.0 is released.

WayneCrawford:
Similar to https://github.com/obspy/obspy/pull/1120, which seems to have been fixed, there is no argument for ``external_references`` in ``Station().__init__()``.  Since it works in ``Channel``, should be easy to implement in ``Station``

Noticed in version 1.4.0

PR #1121, opened 2015-09-14
Support Python 3.5

QuLogic:
Since we don't have a `pyproj` wheel for 3.5, I installed it from conda, but then the 3.3 package doesn't like our pinned NumPy version. Should we just bump NumPy for all of 3.x (`pyproj` is available in conda for newer NumPy)?

The four remaining failures are due to changes in GDAL 2.0.0 which is the only available version for 3.5. I suspect there is some change to a version field in the `.dbf` file (other files from the test are unchanged). @megies How do you think we should proceed to fix the test? I am not really familiar with the shapefile format to decide.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,19 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   # Replace dep1 dep2 ... with your dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy==1.8.2 scipy==0.14.0 matplotlib==1.3.1 basemap==1.0.7 flake8 future lxml sqlalchemy mock nose gdal
+  - |
+      if [[ "${TRAVIS_PYTHON_VERSION:0:1}" == '3' ]]; then
+        NUMPY_VERSION=1.9.2
+        SCIPY_VERSION=0.16.0
+        MPL_VERSION=1.4.3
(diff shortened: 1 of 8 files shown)


megies:
> Since we don't have a pyproj wheel for 3.5, I installed it from conda, but then the 3.3 package doesn't like our pinned NumPy version. Should we just bump NumPy for all of 3.x (pyproj is available in conda for newer NumPy)?

Whatever works, I'd say. Looks like pyproj is available through conda for all our Travis bots, so you could try to just use it from anaconda.. ?

> @megies How do you think we should proceed to fix the test? I am not really familiar with the shapefile format to decide.

Actually I was surprised the tests work as they do because I just did some binary comparison of some of the output files. Like you said, my guess is the output is fine just some version fields changing the binary probably. I can have a look end of week, you can also go ahead and deactivate that offending test on Py35 for now and open a ticket as a reminder if you want.

QuLogic:
> Looks like pyproj is available through conda for all our Travis bots

That's because I special-cased `pyproj`. It's not available for NumPy 1.8 that I can tell (see [older build](https://travis-ci.org/obspy/obspy/jobs/80179710).)

megies:
Oh, we fixed numpy to 1.8, I see, missed that point before. Yeah, why not use 1.9 for Python 3.x in that case..

PR #1122, opened 2015-09-15
y: Ignore non ASCII characters in TAG_STATION_INFO

barsch:
see also
http://lists.swapbytes.de/archives/obspy-users/2015-September/001803.html


megies:
Should be backported to releases in principle, but probably new major release is next anyway. Needs changelog, otherwise looks good to me.

barsch:
its not a bug nor a feature - more a convience function for faulty Y files which shouldn't exists at all - IMHO no need to touch the changelog ;)

megies:
Ok, I get the point, fair enough.

Looking at the code, maybe the handling of the long ASCII NULL terminated string fields should be changed I think. According to the manual, invalid ASCII characters are allowed to appear in there (although only after the terminating ASCII NULL -- which isnt't in the faulty example). Anyway, an edge case I guess.. The `ignore` should work fine, unless we run into valid characters after the termination character which should be ignored as well. But yeah, real strange edge case.

barsch:
yes you are right - after the final ASCII 00 could be everything - decoding should be done after truncating!

megies:
Here's a proposal for slightly improved parsing according to documentation. Tests are not sensitive to these tag parsing issues though..

barsch:
feels a bit to convoluted - how about a custom unpack function with a "z" indicator for ASCIIZ and "s" for standard strings - see also http://stackoverflow.com/questions/11850950/unpacking-a-struct-ending-with-an-asciiz-string

megies:
Indeed, looks much better that way, gonna adapt this..

megies:
Implemented with a wrapped `unpack` with `z` type flag now. Rebased to get rid of my previous commit.

megies:
Fixed previous commit for Py3 and squashed/rebased.

megies:
Good to merge, I think.

obspy/io/y/core.py
```diff
@@ -256,10 +307,11 @@ def _read_y(filename, headonly=False, **kwargs):  # @UnusedVariable
                 # UCHAR Key[16]
                 #   Unique key that identifies this record in the database.
                 data = fh.read(next_tag)
-                parts = unpack(endian + b'd16s', data[8:])
+                parts = _unpack_with_asciiz_and_decode(
+                    endian + b'd16s', data[8:])
```
QuLogic: Shouldn't this be `z`? There's a NUL-strip just below here.

obspy/io/y/core.py
```diff
@@ -256,10 +307,11 @@ def _read_y(filename, headonly=False, **kwargs):  # @UnusedVariable
...
                 params = AttribDict()
                 params.load_date = parts[0]
-                params.key = parts[1].rstrip(b'\x00')
```
QuLogic: Odd that it wasn't `decode`d too.

obspy/io/y/core.py
```diff
@@ -308,10 +361,11 @@ def _read_y(filename, headonly=False, **kwargs):  # @UnusedVariable
                 # UCHAR Key[16]
                 #   Unique key that identifies this record in the database.
                 data = fh.read(next_tag)
-                parts = unpack(endian + b'd16s', data[8:])
+                parts = _unpack_with_asciiz_and_decode(
+                    endian + b'd16s', data[8:])
```
QuLogic: Again, not `z`?

obspy/io/y/core.py
```diff
@@ -322,9 +376,9 @@ def _read_y(filename, headonly=False, **kwargs):  # @UnusedVariable
                 #  PathName is the full name of the file which contains the
                 #  response information for this station.
                 data = fh.read(next_tag)
-                parts = unpack(b'260s', data[8:])
+                parts = _unpack_with_asciiz_and_decode(b'260s', data[8:])
```
QuLogic: Also a `z`?

obspy/io/y/core.py
megies: Unfortunately the [manual](https://github.com/obspy/obspy/blob/master/obspy/io/y/docs/NmxDataFormats_RefGuide_14602R4.pdf) is not very stringent. For many `char` fields it doesn't specify whether it's `BLANKPAD` or `ASCIIZ`. I decided to only touch things that have a clear statement in the manual whether it's  `BLANKPAD` or `ASCIIZ` (I also didn't want to touch more than neccessary since I think that header fields are more or less untested).

obspy/io/y/core.py
megies: Yeah. But again, everything that goes into `trace.stats.y` extra headers is untested.

obspy/io/y/core.py
megies: See above, probably yes, but lacks proper documentation.

obspy/io/y/core.py
megies: See above, probably yes, but lacks proper documentation.

obspy/io/y/core.py
QuLogic: I guess we'd need to have a real file in order to test that out, then?

megies: We have a real file. How the char header fields are expected to be treated is not stated in the manual though, so I wouldn't know how to define what ultimately is expected.

Anyway, I'd merge this now, so that this fix gets into `master`. Anybody who is using this format can work on (or at least trigger) any further polishing that wasn't done back when this was first added to obspy.

PR #1123, opened 2015-09-16
Pass correct types to mpl's date2num.

QuLogic:
The expected input is a `datetime.datetime` object. Passing in `UTCDateTime` used to work only because there were checks for `datetime.date`. The soon-to-be-release 1.5.0 swaps that test for `datetime.datetime`, which this is not, so conversions are treated as if no times were provided.

Testing with 1.5.0rc1 seems to be fine other than this one thing.


megies:
Should go into `releases`, probably? If we do a 0.10.3 anyway, that is..

krischer:
I'll manually cherry pick it onto the releases branch.

PR #1128, opened 2015-09-22
First successful read the docs build

megies:
Last aspect that needed fixing was that we rely on the `Makefile` structure for docs building (mostly downloading header bar from docs.obspy.org during building, but Read The Docs only runs a basic `sphinx-build` so anything needed during the build has to be in `conf.py`. [See this docs entry for introducing RTD-only tweaks into conf.py](https://read-the-docs.readthedocs.org/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs):

screenshot from 2015-09-22 14 28 21

[First successful RTD build](https://readthedocs.org/projects/obspy/builds/3342256/).. and [the output](http://obspy.readthedocs.org/en/rtd/):

screenshot from 2015-09-22 14 32 36

### DIFF
--- a/misc/docs/source/conf.py
+++ b/misc/docs/source/conf.py
@@ -13,6 +13,9 @@ import sphinx_bootstrap_theme
 
 import obspy
 
+
+READ_THE_DOCS = os.environ.get('READTHEDOCS', None) == 'True'
+
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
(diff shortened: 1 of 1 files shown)


krischer:
:tada:

QuLogic:
Cool. :confetti_ball:

megies:
Probably could use some polishing, but it works at least..

ThomasLecocq:
:+1: :-)

megies:
Once the setup (style sheets etc?) is finalized we can also backport the necessary changes to the past stable releases (with hard coding the respective fixed stable release version numbers) to get them on RTD..

PR #1130, opened 2015-09-24
PPSD: store individual psd pieces for customs stacks

megies:
Switch PPSD to store individual psd trace for every processed (by default) hour segment, allowing for customized stacks of data (e.g. by start/endtime or time of day).

Planned changes:
 - [ ] make time of day selection timezone aware (to correctly work across daylight saving time borders)
 - [ ] add plot for individual period bin (or range of periods) vs. time
 - [x] add restriction by weekday
 - [x] add restriction by month?
 - [ ] don't use static gaps as output from `st.getGaps()` but rather only save data intervals and calculate gaps from those (also would visualize gaps in between streams)
 - [x] frequency bin width can be set during PPSD init
 - [x] `plot()` can be switched to have axis in `Hz` instead of `s`

Example:
 - daytime
figure_day

 - nighttime
figure_night


megies:
Squashed some commits and rebased, force-pushed. If CI passes I think this can be merged. Other change plans are less important and can be done later, but keeping the individual psds should go into 0.11.0.

megies:
Squashed more fixes in and force-pushed.

megies:
Noticed that I didn't add tests... -_-
Gonna add them now..

obspy/signal/spectral_estimation.py
```diff
@@ -501,14 +619,17 @@ def __check_time_present(self, utcdatetime):
...
+                                    utcdatetime.timestamp)
+        index2 = bisect.bisect_right(self._times_processed,
                                      utcdatetime.timestamp + self.ppsd_length)
         if index1 != index2:
             return True
         else:
             return False
 
+    @deprecated("Support for old pickled PPSD objects will be dropped. "
+                "Please use new 'save_npz'/'load_npz' mechanism.")
```
QuLogic: This is an internal method. Did you place the deprecation here on purpose?

megies: Yeah, I agree, not needed here, already moved this tiny check into the deprecated `load` routine in soon to be merged PR: https://github.com/obspy/obspy/pull/1166/files#diff-9b9d51c398e8720552f8a521d3f6c69bL631

PR #1131, opened 2015-09-28
Mass Downloader for the FDSN Web Services

krischer:
Hi guys,

first of all I want to apologize that this is so big but I have been working on (an using) this for quite a while and it kind of grew and it would be a real pain to split it up in smaller pull requests.

This PR adds a couple of things:

* Support for requesting gzipped data for XML and text requests from the FDSN web services. This results in much smaller and thus faster requests. If a service supports it it will be used.
* Support for reading the `text` output of the FDSN `station` web services. It contains a limited amount of information compared to the full StationXML but once again is much faster.
* A number of stability improvements to `obspy.clients.fdsn`.
* Helper functionality to download large amount of data across a number of FDSN web services. This is the bulk of the PR and I'll write a bit more about it here.

It is pretty well documented and test coverage is also quite alright considering what is does (lots of mock tests and other tricks).

Rendered Documentation + Motivation (please read it for more information - I don't want to repeat everything here):

* https://www.geophysik.uni-muenchen.de/~krischer/_build/packages/autogen/obspy.clients.fdsn.mass_downloader.html#module-obspy.clients.fdsn.mass_downloader
* https://www.geophysik.uni-muenchen.de/~krischer/_build/packages/autogen/obspy.clients.fdsn.mass_downloader.domain.Domain.html#obspy.clients.fdsn.mass_downloader.domain.Domain
* https://www.geophysik.uni-muenchen.de/~krischer/_build/packages/autogen/obspy.clients.fdsn.mass_downloader.restrictions.Restrictions.html#obspy.clients.fdsn.mass_downloader.restrictions.Restrictions

An important point is that the whole thing is usable as a library so it can be embedded into larger projects quite easily. Thus I think it has a place in ObsPy.

The mass downloader solves lots of real world issue when attempting to download data across FDSN data centers. I realize there are plans from data centers and organizations to make these things cleaner and easier in the future but this can be used right now independent of data center politics and what not.

It tries to make things as easy as possible for users but also data centers. It downloads in parallel, merges, batches, and splits requests and data, has some QC, ...

Example usage:

```python
import obspy
from obspy.clients.fdsn.mass_downloader import CircularDomain, \
    Restrictions, MassDownloader

origin_time = obspy.UTCDateTime(2011, 3, 11, 5, 47, 32)

# Circular domain around the epicenter. This will download all data between
# 70 and 90 degrees distance from the epicenter. This module also offers
# rectangular and global domains. More complex domains can be defined by
# inheriting from the Domain class.
domain = CircularDomain(latitude=37.52, longitude=143.04,
                        minradius=70.0, maxradius=90.0)

restrictions = Restrictions(
    # Get data from 5 minutes before the event to one hours after the
    # event. This defines the temporal bounds of the waveform data.
    starttime=origin_time - 5 * 60,
    endtime=origin_time + 3600,
    # You might not want to deal with gaps in the data. If this settings is
    # True, any trace with a gap/overlap will be discarded.
    reject_channels_with_gaps=True,
    # And you might only want waveforms that have data for at least 95 % of
    # the requested time span. Any trace that is shorter than 95 % of the
    # desired total duration will be discarded.
    minimum_length=0.95,
    # No two stations should be closer than 10 km to each other. This is
    # useful to for example filter out stations that are part of different
    # networks but at the same physical station. Settings this to zero or
    # None will disable that filtering.
    minimum_interstation_distance_in_m=10E3,
    # Only HH or BH channels. If a station has HH channels, those will be
    # downloaded, otherwise the BH. Nothing will be downloaded if it has
    # neither. You can add more/less patterns if you like.
    channel_priorities=("HH[Z,N,E]", "BH[Z,N,E]"),
    # Locations codes are arbitrary and there is no rule which location is
    # best. Same logic as for the previous setting.
    location_priorities=("", "00", "10"))

# No specified providers will result in all known ones being queried.
mdl = MassDownloader()
# The data will be downloaded to ``./waveforms/`` ans ``./stations`` with
# autmatically chosen names.
mdl.download(domain, restrictions, mseed_storage="waveforms",
             stationxml_storage="stations")
```


krischer:
Appveyor fails for some reason I don't understand but it appears to be unrelated.

megies:
8000 lines? Epic.. Review should take until 0.14, I guess. :P

megies:
> Appveyor fails for some reason I don't understand but it appears to be unrelated.

Seem to happen randomly, these build errors. Restarted the build..

megies:
There's a fail on appveyor: https://ci.appveyor.com/project/obspy/obspy/build/1.0.888-master#L983

krischer:
Can somebody by chance check what the problem is on Windows? Might be some combination of line endings + unicode that screws up the csv reader...

megies:
> Can somebody by chance check what the problem is on Windows?

Maybe @ThomasLecocq.. ?

obspy/clients/fdsn/client.py
```diff
@@ -1519,11 +1545,24 @@ def download_url(url, timeout=10, headers={}, debug=False,
...
         return None, e
 
-    code = url_obj.getcode()
+    code = response.getcode()
+
+    # Unpack gzip if necessary.
+    if response.info().get("Content-Encoding") == "gzip":
+        if debug is True:
+            print("Uncompressing gzipped response for %s" % url)
+        # Cannot directly stream to gzip from urllib!
```
QuLogic: Works in Python 3; maybe something for `future`?

CHANGELOG.txt
```diff
@@ -23,6 +25,12 @@ master:
    * Trace.remove_response() now has `plot` option to show/output a plot of all
      individual steps of instrument response removal in frequency domain
      (see #1116).
+ - obspy.clients.fdsn:
+   * Now requests gzipped data for the XML files. Much smaller files!
+   * The station service can now also be used to download files in the text
+     format. This has limited information but is much faster.
+   * New mass downloader to assist in downloading data across a larg number
```
QuLogic: s/larg/large/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+Why Would You Want to Use This?
+-------------------------------
+
+Directly using the FDSN web services for example via the
+:mod:`obspy.clients.fdsn` client is fine for small amounts of data but quickly
+becomes cumbersome for larger data sets. Many data centers do provide tools to
+easily download larger amounts of data but that is usually only from one data
+center. Now most seismologist don't really care a lot where the data they
```
QuLogic: s/seismologist/seismologists/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+download originates from - they just want the data for their use case and
```
QuLogic: s/from//

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+The mass downloader module will acquire all waveforms and associated station
+information across all known FDSN web service implementations producing a
+**clean data set** ready for further use.  It works by
+
+1. figuring out what stations each provider offers,
+2. downloading MiniSEED and associated StationXML meta information in an
+   efficient and data center friendly manner, and
+3. deal with all the nasty real-world data issues like missing or incomplete
```
QuLogic: s/deal/dealing/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+--------------
+
+Before delving into the nitty-gritty details of how it works and why it does
+things in a certain way we'll demonstrate the usage of this module on two
+annotated examples. They can serve as templates for your own needs.
+
+Earthquake Data
+~~~~~~~~~~~~~~~
+
+The classic seismological data set is consisting of waveform recordings for a
```
QuLogic: s/is consisting/consists/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+    # inheriting from the Domain class.
+    domain = CircularDomain(latitude=37.52, longitude=143.04,
+                            minradius=70.0, maxradius=90.0)
+
+    restrictions = Restrictions(
+        # Get data from 5 minutes before the event to one hours after the
+        # event. This defines the temporal bounds of the waveform data.
+        starttime=origin_time - 5 * 60,
+        endtime=origin_time + 3600,
+        # You might not want to deal with gaps in the data. If this settings is
```
QuLogic: s/settings/setting/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+        # True, any trace with a gap/overlap will be discarded.
+        reject_channels_with_gaps=True,
+        # And you might only want waveforms that have data for at least 95 % of
+        # the requested time span. Any trace that is shorter than 95 % of the
+        # desired total duration will be discarded.
+        minimum_length=0.95,
+        # No two stations should be closer than 10 km to each other. This is
+        # useful to for example filter out stations that are part of different
+        # networks but at the same physical station. Settings this to zero or
```
QuLogic: s/Settings/Setting/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+        # None will disable that filtering.
+        minimum_interstation_distance_in_m=10E3,
+        # Only HH or BH channels. If a station has HH channels, those will be
+        # downloaded, otherwise the BH. Nothing will be downloaded if it has
+        # neither. You can add more/less patterns if you like.
+        channel_priorities=("HH[Z,N,E]", "BH[Z,N,E]"),
```
QuLogic: Is this pattern a regex or glob? Why bother with the commas?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+        # Locations codes are arbitrary and there is no rule which location is
```
QuLogic: s/Locations/Location/
s/rule which/rule as to which/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+        # best. Same logic as for the previous setting.
+        location_priorities=("", "00", "10"))
+
+    # No specified providers will result in all known ones being queried.
+    mdl = MassDownloader()
+    # The data will be downloaded to ``./waveforms/`` ans ``./stations`` with
```
QuLogic: s/ans/and/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+    # autmatically chosen names.
```
QuLogic: s/autmatically/automaticall/

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: Is `./stations` a single file? Why does it not end in a slash?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+:class:`~obspy.clients.fdsn.mass_downloader.restrictions.Restrictions` class.
+
+The :class:`~obspy.clients.fdsn.mass_downloader.domain` module currently
+defines three different domain types used to limit the geographical extent of
+the queried data:
+:class:`~obspy.clients.fdsn.mass_downloader.domain.RectangularDomain`,
+:class:`~obspy.clients.fdsn.mass_downloader.domain.CircularDomain`, and
+:class:`~obspy.clients.fdsn.mass_downloader.domain.GlobalDomain`. Subclassing
+:class:`~obspy.clients.fdsn.mass_downloader.domain.Domain` enables the
+construction of arbitrary complex domains. Please see the
```
QuLogic: s/arbitrary/arbitrarily/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+...     minimum_interstation_distance_in_m=1000,
+...     channel_priorities=["HH[Z,N,E]", "BH[Z,N,E]"],
+...     location_priorities=["", "00", "01"])
+
+
+Step 2: Storage Options
+~~~~~~~~~~~~~~~~~~~~~~~
+
+After determining what to download, the helpers must know where to store the to
+be downloaded data. This requires some flexibility in case this is integrated
```
QuLogic: "to be downloaded" is a bit odd to read. Maybe "requested" or "selected".

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+as a component into a bigger system. An example of this is a toolbox that has a
+database to manage its data. A major concern is to not download already
+existing data. In order to enable such a use case the download helpers can be
```
QuLogic: s/already existing/pre-existing/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+given functions that are evaluated when determining the filenames of the to be
+downloaded data.  Depending on the return value, the helper class will download
```
QuLogic: Same minor quibble about "to be downloaded". Something like "potential" might work here.

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+the whole, only parts, or even nothing, of that particular piece of data.
```
QuLogic: the whole, part, or even none, of ...

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+...     channel_priorities=["HH[Z,N,E]", "BH[Z,N,E]"],
+...     location_priorities=["", "00", "01"])
+
+
+Step 2: Storage Options
+~~~~~~~~~~~~~~~~~~~~~~~
+
+After determining what to download, the helpers must know where to store the to
+be downloaded data. This requires some flexibility in case this is integrated
+as a component into a bigger system. An example of this is a toolbox that has a
```
QuLogic: There are three "this" in the previous two sentences, and it's a bit unclear what "this" is. It should say "this <something>".

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+which is rather general but also quite long.
+
+**Option 2: String Template**
+
+For more control use the second possibility and provide a string containing
+``{network}``, ``{station}``, ``{location}``, ``{channel}``, ``{starttime}``,
+and ``{endtime}`` format specifiers. These values will be interpolated to
+acquire the final filename. The start and end times will be formatted with
+``strftime()`` with the specifier ``'%Y-%m-%dT%H-%M-%SZ'`` in an effort to
```
QuLogic: Is it really necessary to use the dashes? ISO8601 allows you to drop the colons.

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+filename.
+
+>>> def get_mseed_storage(network, station, location, channel, starttime,
+...                       endtime):
+...     # Returning True means that neither the data nor the StationXML file
+...     # will be downloaded.
+...     if is_in_db(network, station, location, channel, starttime, endtime):
+...         return True
+...     # If a string is returned the file will be saved in that location.
+...     return os.path.join(ROOT, "%s.%s.%s.%s.mseed." % (network, station,
```
QuLogic: Why does the filename end in a period?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+...                                                       location, channel))
+>>> mseed_storage = get_mseed_storage
+
+.. note::
+
+    No matter which approach is chosen, if a file already exists, it will not
+    be overwritten; it will be parsed and the download helper class will
```
QuLogic: How does this parsing work if the helper function only returns `True`?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+will write to ``"some_folder/NETWORK/STATION.xml"``, in this case for example
+to ``"some_folder/BW/FURT.xml"``.
+
+.. note::
+
+    If the StationXML file already exists, it will be opened to see what is in
+    the file. In case it does not contain all necessary channels, it will be
+    deleted and **only those channels needed in the current run will be
+    downloaded again**. Pass a custom function to the ``stationxml_path``
```
QuLogic: OK, not all will be downloaded, but which ones get saved in the resulting file?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+    argument if you require different behavior as documented in the following.
```
QuLogic: the following section.

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+**Option 3: Custom Function**
+
+As with the waveform data, the StationXML paths can also be set with the help
+of a function. The function in this case is a bit more complex then for the
```
QuLogic: s/then/than/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+waveform case. It has to return a dictionary with three keys:
+``"available_channels"``, ``"missing_channels"``, and ``"filename"``.
+``"available_channel"`` is a list of channels that are already available as
```
QuLogic: `available_channels`?

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+station information and that require no new download. Make sure to include all
+already available channels; this information is later used to discard
+MiniSEED files that have no corresponding station information.
+``"missing_channels"`` is a list of channels for that particular station that
+must be downloaded and ``"filename"`` determines where to save these. Please
+note that in this particular case the StationXML file will be overwritten if it
+already exists and the ``"missing_channels"`` will be downloaded to it,
```
QuLogic: "and _only_ the ..." right?

obspy/clients/fdsn/mass_downloader/domain.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Domain definitions for the the download helpers.
```
QuLogic: s/the the/the/

obspy/clients/fdsn/mass_downloader/domain.py
```diff
@@ -0,0 +1,177 @@
...
+from future.utils import with_metaclass
+
+from abc import ABCMeta, abstractmethod
+
+
+class Domain(with_metaclass(ABCMeta)):
+    """
+    Abstract base class defining a domain - subclass it to define a new domain.
+
+    Each subclass has to implement the :meth:`~.get_query_parameters`
```
QuLogic: s/has to/must/

obspy/clients/fdsn/mass_downloader/domain.py
```diff
@@ -0,0 +1,177 @@
...
+
+    .. code-block:: python
+
+        import fiona
+        import shapely.geometry
+
+        from obspy.clients.fdsn.mass_downloader import Domain
+
+
+        class Germany(Domain):
```
QuLogic: I was just thinking some kind of shapely domain might be useful, but I guess the example's right here.

obspy/clients/fdsn/mass_downloader/domain.py
```diff
@@ -0,0 +1,177 @@
...
+        Return the domain specific query parameters for the
+        :meth:`~obspy.clients.fdsn.client.Client.get_stations` method as a
+        dictionary. Possibilities keys for rectangular queries are
+
+        * ``minlatitude``
+        * ``maxlatitude``
+        * ``minlongitude``
+        * ``maxlongitude``
+
+        For cirular queries:
```
QuLogic: s/cirular/circular/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+        """
+        Returns true if any of the station's time intervals have status
+        "DOWNLOADED" or "EXISTS". Otherwise it returns False meaning it does
+        not have to be considered anymore.
+        """
+        status = []
+        for chan in self.channels:
+            for ti in chan.intervals:
+                status.append(ti.status)
+        status = list(set(status))
```
QuLogic: Why can't `status` just be a `set` in the first place?

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+            utils.safe_delete(self.stationxml_filename)
+
+    @property
+    def stationxml_filename(self):
+        return self._stationxml_filename
+
+    @stationxml_filename.setter
+    def stationxml_filename(self, value):
+        """
+        Setter creating the directory for the file if does not already exist.
```
QuLogic: s/if/if it/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+            "\t-> Has station information for channels:    {has}\n"
+            "\t-> Misses station information for channels: {miss}\n"
+            "\t{channels}"
+        ).format(
+            network=self.network,
+            station=self.station,
+            lat=self.latitude,
+            lng=self.longitude,
+            filename=self.stationxml_filename,
+            status="exist" if (self.stationxml_filename and os.path.exists(
```
QuLogic: s/exist/exists/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+            for channel in available_channels:
+                if channel not in channels_wanting_station_information:
+                    continue
+                self.have_station_information[channel] = \
+                    channels_wanting_station_information[channel]
+
+            self.stationxml_filename = storage["filename"]
+
+            # Raise a warning if something is missing, but do not raise an
+            # exception or halt the programm at this point.
```
QuLogic: s/programm/program/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                self.stationxml_status = STATUS.IGNORE
+
+    def prepare_mseed_download(self, mseed_storage):
+        """
+        Loop through all channels of the station and distribute filenames
+        and the current status of the channel.
+
+        A MiniSEED interval will be ignored, if the `mseed_storage` function
+        returns `True`.
+        Possible statuses after this the method execution are IGNORE, EXISTS,
```
QuLogic: s/this the //

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+        self.location = location
+        self.channel = channel
+        self.intervals = intervals
+
+    @property
+    def needs_station_file(self):
+        """
+        Determine if the channel requires any station information.
+
+        As soon as at the status of at least one interval is either
```
QuLogic: s/at //

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+    :type client: :class:`obspy.fdsn.client.Client`
+    :param client: An initialized FDSN client.
+    :type client_name: str
+    :param client_name: The name of the client. Only used for logging.
+    :type restrictions: :class:`~.restrictions.Restrictions`
+    :param restrictions: The non-domain related restrictions for the query.
+    :type domain: :class:`~.domain.Domain` subclass
+    :param domain: The domain definition.
+    :param mseed_storage: The MiniSEED storage settings.
+    :param stationxml_storage: THe StationXML storage settings.
```
QuLogic: s/THe/The/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+
+    def __bool__(self):
+        return bool(len(self))
+
+    def __str__(self):
+        avail_map = {
+            None: "Unknown reliability of availability information",
+            True: "Reliable availability information",
+            False: "Non-reliable availability information"
+
```
QuLogic: Is this newline supposed to go after the closing brace?

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                indexes_to_remove.append(most_common)
+                nns = list(itertools.filterfalse(
+                    lambda x: most_common in x, nns))
+
+            # Remove these indices this results in a set of stations we wish to
+            # keep.
+            remaining_stations.extend(
+                set([_i[1] for _i in itertools.filterfalse(
+                    lambda x: x[0] in indexes_to_remove,
+                    enumerate(stations))]))
```
QuLogic: Is `itertools` really necessary? This is equivalent to `[_i[1] for _i in enumerate(stations) if x[0] not in indexes_to_remove]`, isn't it (which seems simpler to me)?

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+            rejected_stations.extend(
+                set([_i[1] for _i in filter(
+                    lambda x: x[0] in indexes_to_remove,
+                    enumerate(stations))]))
```
QuLogic: Same as above with negated condition.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                    existing_stations)
+                # Now we have to get the distance to the closest existing
+                # station for all new stations.
+                distances = np.ma.array(existing_kd_tree.query(stations)[0][0])
+                if np.isinf(distances[0]):
+                    break
+                distances.mask = False
+
+                # Step one is to get rid of all stations that are closer
+                # then the minimum distance to any existing station.
```
QuLogic: s/then/than/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                    continue
+                station.have_station_information[c_id] = times
+
+            station.miss_station_information = still_missing
+            if still_missing:
+                station.stationxml_status = STATUS.DOWNLOAD_PARTIALLY_FAILED
+            else:
+                station.stationxml_status = STATUS.DOWNLOADED
+
+        # Now loop over all stations and that the status of the ones that
```
QuLogic: s/that/set/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                             self.client_name, filecount,
+                             download_size / 1024.0 ** 2,
+                             e_time - s_time,
+                             (download_size / 1024.0) / (e_time - s_time)))
+
+    def download_mseed(self, chunk_size_in_mb=25, threads_per_client=3):
+        """
+        Actually download MiniSEED data.
+
+        :param chunk_size_in_mb:
```
QuLogic: Needs description.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+        Should be run after the MiniSEED and StationXML downloads finished.
+        It will make sure that every MiniSEED file also has a corresponding
+        StationXML file.
+        """
+        for station in self.stations.values():
+            station.sanitize_downloads(logger=self.logger)
+
+    def _check_downloaded_data(self):
+        """
+        Read the downloaded data, set the proper status flags and a remove
```
QuLogic: s/a //

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+
+        Returns the downloaded_bytes and the discarded_bytes.
+        """
+        downloaded_bytes = 0
+        discarded_bytes = 0
+        for sta in self.stations.values():
+            for cha in sta.channels:
+                for interval in cha.intervals:
+                    # The status of the interval should not have changed if
+                    # it did not require doenloading in the first place.
```
QuLogic: s/doenloading/downloading/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                        self.logger.warning(
+                            "Empty file '%s'. Will be deleted." %
+                            interval.filename)
+                        utils.safe_delete(interval.filename)
+                        discarded_bytes += size
+                        interval.status = STATUS.DOWNLOAD_FAILED
+                        continue
+
+                    # If user did not want gappy files, remove them.
+                    if self.restrictions.reject_channels_with_gaps is True and\
```
QuLogic: This `is True` is not necessary.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                            len(st) > 1:
+                        self.logger.info("File '%s' contains %i traces. Will "
```
QuLogic: Message should mention something about gaps.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+        time_range = restrictions.minimum_length * (restrictions.endtime -
+                                                    restrictions.starttime)
+        channel_availability = []
+        for filename in filenames:
+            st = obspy.read(filename, format="MSEED", headonly=True)
+            if restrictions.reject_channels_with_gaps and len(st) > 1:
+                self.logger.warning("Channel %s has gap or overlap. Will be "
+                                    "removed." % st[0].id)
+                try:
+                    os.remove(filename)
```
QuLogic: Why not `utils.safe_delete`? (I haven't read what that actually does yet.)

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+
+        for station_id in station_ids:
+            try:
+                del self.stations[station_id]
+            except KeyError:
+                pass
+
+    def get_availability(self):
+        """
+        Queries the current client for information of what stations are
```
QuLogic: s/of/on/

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+            "endtime": self.restrictions.endtime,
+            # Request at the channel level.
+            "level": "channel"
+        }
+        # Add the domain specific query parameters.
+        arguments.update(self.domain.get_query_parameters())
+
+        # Check the capabilities of the service and see what is the most
+        # appropriate way of acquiring availability information. Some services
+        # right now require manual overwriting of what they claim to be
```
QuLogic: s/overwriting/overriding/

obspy/clients/fdsn/mass_downloader/mass_downloader.py
```diff
@@ -0,0 +1,329 @@
...
+        Launch the actual data download.
+
+        :param domain: The download domain.
+        :type domain: :class:`~.domain.Domain`
+        :param restrictions: Non-spatial downloading restrictions.
+        :type restrictions: :class:`~.restrictions.Restrictions`
+        :param mseed_storage: Where to store the waveform files. See
+            the :mod:`~obspy.clients.fdsn.mass_downloader` documentation for
+            more details.
+        :type mseed_storage: str or fct
```
QuLogic: s/fct/function/

obspy/clients/fdsn/mass_downloader/mass_downloader.py
```diff
@@ -0,0 +1,329 @@
...
+        :param stationxml_storage: Where to store the StationXML files. See
+            the :mod:`~obspy.clients.fdsn.mass_downloader` documentation for
+            more details.
+        :type stationxml_storage: str of fct
```
QuLogic: s/of fct/or function/

obspy/clients/fdsn/mass_downloader/mass_downloader.py
```diff
@@ -0,0 +1,329 @@
...
+        :param download_chunk_size_in_mb: MiniSEED data will be downloaded
+            in bulk chunks. This settings limits the chunk size. A higher
+            numbers means that less total download requests will be send,
```
QuLogic: s/numbers/number/

obspy/clients/fdsn/mass_downloader/mass_downloader.py
```diff
@@ -0,0 +1,329 @@
...
+            helper.prepare_stationxml_download()
+            helper.download_stationxml()
+
+            # Sanitize the downloaded things if desired. Assures that all
+            # waveform data also has the corresponding station information.
+            if restrictions.sanitize:
+                helper.sanitize_downloads()
+
+            # Filter afterwards if availability information is not reliable.
+            # This unfortunately results in already downloaded data to be
```
QuLogic: s/to be/being/

obspy/clients/fdsn/mass_downloader/mass_downloader.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # Catch warnings in the main thread. The catch_warnings() context
+        # manager does not reliably work when used in multiple threads.
+        p = ThreadPool(len(self.providers))
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            clients = p.map(_get_client, self.providers)
+        p.close()
+        for warning in w:
+            logger.debug("Warning during initializing one of the clients: " +
```
QuLogic: s/initializing/initialization of/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+class Restrictions(object):
+    """
+    Class storing non-domain restrictions for a query. This is best explained
+    with two examples. See the list below for a more detailed explanation
+    of the parameters. The first set of restrictions is useful for event
+    based earthquake set queries.
+
+    >>> import obspy
+    >>> restrictions = Restrictions(
+    ...     # Get data from 5 minutes before the event to one hours after the
```
QuLogic: s/hours/hour/

obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -0,0 +1,503 @@
...
+
+    # Circular domain around the epicenter. This will download all data between
+    # 70 and 90 degrees distance from the epicenter. This module also offers
+    # rectangular and global domains. More complex domains can be defined by
+    # inheriting from the Domain class.
+    domain = CircularDomain(latitude=37.52, longitude=143.04,
+                            minradius=70.0, maxradius=90.0)
+
+    restrictions = Restrictions(
+        # Get data from 5 minutes before the event to one hours after the
```
QuLogic: s/hours/hour/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+    ...     # event.
+    ...     starttime=obspy.UTCDateTime(2012, 1, 1),
+    ...     endtime=obspy.UTCDateTime(2012, 1, 2),
+    ...     # You might not want to deal with gaps in the data.
+    ...     reject_channels_with_gaps=True,
+    ...     # And you might only want waveform that have data for at least
```
QuLogic: s/waveform/waveforms/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+    ...     # 95 % of the requested time span.
+    ...     minimum_length=0.95,
+    ...     # No two stations should be closer than 10 km to each other.
+    ...     minimum_interstation_distance_in_m=10E3,
+    ...     # Only HH or BH channels. If a station has HH channels,
+    ...     # those will be downloaded, otherwise the BH. Nothing will be
+    ...     # downloaded if it has neither.
+    ...     channel_priorities=("HH[Z,N,E]", "BH[Z,N,E]"),
+    ...     # Locations codes are arbitrary and there is no rule which
```
QuLogic: s/Locations/Location/
s/which/as to which/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+    :param endtime: The endtime of the data.
+    :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param station_starttime: The start time of the station files. If not
+        given, the ``starttime`` argument will be used. This is useful when
+        trying to incorporate multiple waveform datasets with a central
+        station file archive as StationXML files can be downloaded once and
+        for the whole time span.
+    :type station_starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param station_endtime: The endtime of the station files. Analogous to
+        the the ``station_starttime`` argument.
```
QuLogic: s/the the/the/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+    ...     # be deleted afterwards. This is not desirable for large noise
+    ...     # data sets.
+    ...     sanitize=False,
+    ...     # Guard against the same station having different names.
+    ...     minimum_interstation_distance_in_m=100.0)
+
+
+    :param starttime: The start time of the data to be downloaded.
+    :type starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param endtime: The endtime of the data.
```
QuLogic: s/endtime/end time/

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -0,0 +1,200 @@
...
+    :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param station_starttime: The start time of the station files. If not
+        given, the ``starttime`` argument will be used. This is useful when
+        trying to incorporate multiple waveform datasets with a central
+        station file archive as StationXML files can be downloaded once and
+        for the whole time span.
+    :type station_starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param station_endtime: The endtime of the station files. Analogous to
```
QuLogic: s/ endtime / end time /

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+            # cases.
+            if b[4] <= cur_bulk[-1][5] + 2:
+                cur_bulk[-1][5] = b[5]
+                continue
+            cur_bulk.append(b)
+        bulk_channels[key] = cur_bulk
+    bulk = list(itertools.chain.from_iterable(bulk_channels.values()))
+
+    # Save first to a temporary file, then cut the file into separate files.
+    temp_filename = NamedTemporaryFile().name
```
QuLogic: This leaks a file handle; try running the test file independently.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+                    open_files[filename].write(fh.read(info["record_length"]))
+            finally:
+                for f in open_files.values():
+                    try:
+                        f.close()
+                    except:
+                        pass
+    finally:
+        try:
+            os.remove(temp_filename)
```
QuLogic: Why not just use `NamedTemporaryFile` in a context as it was intended?

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    def spherical2cartesian(data):
+        """
+        Converts a list of :class:`~obspy.fdsn.download_status.Station`
+        objects to an array of shape(len(list), 3) containing x/y/z in meters.
+        """
+        # Create three arrays containing lat/lng/elevation.
+        shape = len(data)
+        lat = np.array([_i.latitude for _i in data], dtype=np.float64)
+        lon = np.array([_i.longitude for _i in data], dtype=np.float64)
+        r = np.array([EARTH_RADIUS for _i in data], dtype=np.float64)
```
QuLogic: This is not elevation, as the comment says.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+        self.data = data
+        self.kd_tree = cKDTree(data=cart_data, leafsize=10)
+
+    def query(self, points):
+        points = self.spherical2cartesian(points)
+        d, i = self.kd_tree.query(points)
+        new_d, new_i = [], []
+
+        # Filter NaNs. Happens when not enough points are available.
+        for _d, _i in zip(d, i):
```
QuLogic: Aren't `d` and `i` NumPy arrays? Hence,
```python
m = np.isfinite(d)
d = d[m]
i = i[m]
```

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: OK, I see it's `fnmatch`, which has no special handling of commas, so this example is not working quite as it says.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    one or more channels are found it stops. Otherwise it will attempt to
+    retrieve channels matching the next pattern. And so on.
+
+    :type channels: list
+    :param channels: A list containing channel names.
+    :type priorities: list of unicode or None
+    :param priorities: The desired channels with descending priority. Channels
+    will be matched by fnmatch.fnmatch() so wildcards and sequences are
+    supported. The advisable form to request the three standard components
+    of a channel is "HH[Z,N,E]" to avoid getting e.g. rotated components.
```
QuLogic: These commas in the pattern should be removed, e.g.,
```python
In[4]: [x for x in string.printable if fnmatch.fnmatch('HH' + x, 'HH[Z,E,N]')]
Out[4]: ['E', 'N', 'Z', ',']
```

obspy/clients/fdsn/mass_downloader/download_helpers.py
QuLogic: OK, now I see it doesn't ignore errors, so carry on then...

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    :type station: str
+    :param channels: The channels. Each channel is a tuple of two strings:
+        location code and channel code.
+    :type channels: list of tuples
+    :param starttime: The start time.
+    :type starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :param endtime: The end time.
+    :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+
+    The rules are simple, if it is a function, network, station, channels,
```
QuLogic: This clause appears to continue on after "function", even though that's the end of the condition, so something like "if it is a function, then the network, station, ..." would be better.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    starttime, and endtime are passed as arguments.
```
QuLogic: s/starttime/start time/
s/endtime/end time/

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    Otherwise it is considered to be a folder name and the resulting
+    filename will be ``"FOLDER_NAME/NET.STA.xml"``
+
+    This function will either return a string or a dictionary with three
+    keys: ``"available_channels"``, ``"missing_channels"``, and ``"filename"``.
+    """
+    # Call if possible.
+    if callable(str_or_fct):
+        path = str_or_fct(network, station, channels, starttime, endtime)
+    # Check if its a format template.
```
QuLogic: s/its/it's/

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    elif ("{network}" in str_or_fct) and ("{station}" in str_or_fct):
+        path = str_or_fct.format(network=network, station=station)
+    # Otherwise assume its a path.
```
QuLogic: s/its/it's/

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    filename will be ``"FOLDER_NAME/NET.STA.xml"``
+
+    This function will either return a string or a dictionary with three
+    keys: ``"available_channels"``, ``"missing_channels"``, and ``"filename"``.
+    """
+    # Call if possible.
+    if callable(str_or_fct):
+        path = str_or_fct(network, station, channels, starttime, endtime)
+    # Check if its a format template.
+    elif ("{network}" in str_or_fct) and ("{station}" in str_or_fct):
```
QuLogic: Isn't this a bit strict? One could try to apply a formatting option to the substitutions.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+    If it is a string, and it contains ``"{network}"``,  ``"{station}"``,
+    ``"{location}"``, ``"{channel}"``, ``"{starttime}"``, and ``"{endtime}"``
+    formatting specifiers, ``str.format()`` is called.
+
+    Otherwise it is considered to be a folder name and the resulting
+    filename will be
+    ``"FOLDER_NAME/NET.STA.LOC.CHAN__STARTTIME__ENDTIME.mseed"``
+
+    In the last two cases, the times will be formatted with
+    ``"%Y-%m-%dT%H-%M-%SZ"``.
```
QuLogic: See earlier question about ISO8601.

obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -0,0 +1,534 @@
...
+        raise TypeError("'%s' is not a filepath." % str(path))
+
+
+def get_mseed_filename(str_or_fct, network, station, location, channel,
+                       starttime, endtime):
+    """
+    Helper function getting the filename of a MiniSEED file.
+
+    The rules are simple, if it is a function, network, station, location,
+    channel, starttime, and endtime are passed as arguments and the resulting
```
QuLogic: See comment about the phrasing in the previous function.

obspy/clients/fdsn/tests/test_client.py
```diff
@@ -847,7 +848,7 @@ def custom_side_effects(*args, **kwargs):
         # Some custom urls
         base_url_event = "http://other_url.com/beta/event_service/11"
         base_url_station = "http://some_url.com/beta2/station/7"
-        base_url_ds = "http://new.com/beta3/dataselect/8"
+        base_url_ds = "http://new_url.com/beta3/dataselect/8"
```
QuLogic: I did not notice these before, but they are valid (if unused at the moment) domain names. It might be better to use `example.{com,org,edu}` or `something.example` which are reserved.

obspy/core/util/base.py
```diff
@@ -36,8 +36,9 @@
                    'io.datamark', 'io.gse2', 'io.json', 'io.kinemetrics',
                    'io.mseed', 'io.ndk', 'io.nlloc', 'io.pdas', 'io.pde',
                    'io.quakeml', 'io.sac', 'io.seg2', 'io.segy', 'io.seisan',
-                   'io.sh', 'io.shapefile', 'io.stationxml', 'io.wav',
-                   'io.xseed', 'io.y', 'io.zmap', 'realtime', 'signal', 'taup']
+                   'io.sh', 'io.shapefile', 'io.stationxml', 'io.stationtxt',
```
QuLogic: Not properly sorted.

obspy/io/stationtxt/__init__.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+obspy.io.stationtxt - Read support for the  FDSNWS station text files
```
QuLogic: Extra space after "the".

obspy/io/stationtxt/core.py
```diff
@@ -0,0 +1,256 @@
...
+    except:
+        pass
+
+    if not first_line.startswith("#"):
+        return False
+    first_line = first_line.lstrip("#").strip()
+    if not first_line:
+        return False
+
+    # IRIS currently has a wrong header name. Just map it.
```
QuLogic: How do they have the wrong header when it's their own extension‽

obspy/io/stationtxt/core.py
```diff
@@ -0,0 +1,256 @@
...
+                warnings.warn(
+                    "Failed to parse channel %s.%s.%s.%s due to: %s" % (
+                        net, sta, loc, chan, str(e)),
+                    UserWarning)
+                continue
+            stations[(net, sta)].channels.append(channel)
+        inv.networks.extend(list(networks.values()))
+    else:
+        # Cannot really happen - just a safety measure.
+        raise NotImplementedError
```
QuLogic: Better to add `level` to the error message even if it's impossible.

obspy/io/stationxml/core.py
```diff
@@ -715,7 +734,7 @@ def _write_stationxml(inventory, file_or_file_object, validate=False,
         don't trust ObsPy. Defaults to False.
     """
     # Check if any of the channels has a data availability element. In that
-    # case the namespaces need to be adjusted.
+    # case the namespaces needs to be adjusted.
```
QuLogic: This sentence was correct before the change.

QuLogic:
> Can somebody by chance check what the problem is on Windows? Might be some combination of line endings + unicode that screws up the csv reader...

All your `open` calls in text mode do not specify an encoding, so it is probably not `utf-8` on Windows.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+        self._remove_failed_and_ignored_stations()
+
+    def _remove_failed_and_ignored_stations(self):
+        """
+        Removes all stations that have no time interval with either exists
+        or downloaded status.
+        """
+        to_be_removed_keys = []
+        for key, station in self.stations.items():
+            if station.has_existing_or_downloaded_time_intervals is True:
```
QuLogic: This is a `bool`, isn't it? So `is True` is not needed.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -0,0 +1,1140 @@
...
+                         "(%.2f seconds)" % (self.client_name, end - start))
+
+        # Get the time intervals from the restrictions.
+        intervals = [TimeInterval(start=_i[0], end=_i[1])
+                     for _i in self.restrictions]
+
+        for network in inv:
+            for station in network:
+                # Skip the station if it is not in the desired domain.
+                if needs_filtering is True and \
```
QuLogic: This `is True` is not needed.

QuLogic:
I was largely avoiding reading the actual code hoping that the tests would be sufficient, but then `test_mass_downloader.py` was too large for GitHub to show it! So, some points then:
 * `test_channel_priority_filtering` priorities need to remove the commas.
 * line 347, 2147, s/its/it's/
 * line 598, Nof?
 * line 709, s/Endtime/End time/
 * line 752, s/are/is/
 * line 972, 1003, s/property testing/property that tests/
 * line 993, s/Some/Same/
 * line 1014, s/per/by/
 * line 1095, tests don't return anything.
 * line 1144, s/Right/Right now/
 * line 1265, 1495, 1531, s/exists/exist/
 * line 1319, 1348, 1380, s/thats/that's/
 * in `test_prepare_stationxml_download_simple_cases`, `p.call_count` is only checked in one case but it's still reset; should it be checked more often?
 * Many closures are `return something; pass`. What's the point of the `pass`?
 * line 1661-1662, the comment seems irrelevant in this test.
 * line 1895, "Tests the helper object that downloads from a client."
 * line 2024, similar phrasing change.
 * line 2027, s/can/can be/
 * `test_download_stationxml` and `test_get_availability` don't seem to assert anything at all; I guess the mocking is enough?
 * line 2089, I don't know what this sentence is trying to say.
 * In `test_parse_miniseed_filenames`, can the `NamedTemporaryFile` not just be used as a context (as I asked elsewhere)?
 * Not many asserts in `test_download_method`, which I guess is also not possible.

krischer:
Thanks a lot for the review! I'll get to it today.

My email client crashed trying to sort all the emails into conversations ;-)

obspy/clients/fdsn/client.py
krischer: @edschofield

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Oh wow...I've been using `fnmatch` for year but was no aware of this...thankfully commas are rarely contained in what I want to match ;-)

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Its a folder. I'll clarify it.

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Done.

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Just a typo.

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Nothing will happen if it always returns `True`.

`True` means that the data exists but that the download helpers need no longer worry to also download the StationXML - from that point on it will essentially be ignored.

If it returns a string (filename) and the file exists it will be parsed and the corresponding StationXML data will be downloaded.

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Only those for which the current downloading run did download some MiniSEED files. Is that not clear from the sentence?

obspy/clients/fdsn/mass_downloader/__init__.py
krischer: Changed it. In my opinion that would still have been a valid sentence but I trust the native speaker ;-)

obspy/clients/fdsn/mass_downloader/domain.py
krischer: ;-)

obspy/clients/fdsn/mass_downloader/download_helpers.py
krischer: Yea I know - I still like to do this as Python's dynamic typing has bitten me too many times...

obspy/clients/fdsn/mass_downloader/utils.py
krischer: > One could try to apply a formatting option to the substitutions.

What do you mean by that?

The names must be unique for each network and station combination so I don't think its too strict.

obspy/clients/fdsn/tests/test_client.py
krischer: Good point.

obspy/io/stationtxt/core.py
krischer: Yea I also don't know how that happened. Maybe it changed between being an extension - now its an official output format of the web service.

IRIS fixed it after my report - I'm just leaving this in as I fear other services might model their own after the IRIS service and the line does not hurt.

obspy/clients/fdsn/mass_downloader/download_helpers.py
krischer: See my comment on the same issue above.

obspy/clients/fdsn/mass_downloader/download_helpers.py
krischer: see above.

krischer:
Thanks again for the very extensive review.

> All your open calls in text mode do not specify an encoding, so it is probably not utf-8 on Windows.

Good point. I hard-coded utf-8 which should be pretty save. Let's see what appveyor has to say.

> test_download_stationxml and test_get_availability don't seem to assert anything at all; I guess the mocking is enough?

It should be - it checks that the methods on the actual FDSN client receive the correct arguments and these methods are tested elsewhere.

> Not many asserts in test_download_method, which I guess is also not possible.

Not without actually downloading something or even more effort with the mocks. This test already mocks like 10 function at which point there are diminishing returns and the cognitive cost of the mocks becomes a bit too high.
Right now it at least makes sure that the test coverage is pretty high.

krischer:
Hmm...appveyor still fails for some reason :-(

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: `bash` would accept `{Z,N,E}` which might be what you're thinking of, but I think it's an extension to normal POSIX syntax and `fnmatch` doesn't accept it.

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: Okay, probably this note needs a small adjustment then, since there are actually 3 conditions (filename+nonexistent, filename+existent, True) but it seems to imply there are only two (existent+nonexistent).

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: I think it does make sense, but I'm envisioning a case where the file already contains channels/stations that will not be downloaded. My understanding is that those would be lost, but it _might_ be unclear to some.

obspy/clients/fdsn/mass_downloader/__init__.py
QuLogic: Technically yes; I just find it reads a little better this way.

obspy/clients/fdsn/mass_downloader/utils.py
QuLogic: Perhaps it's too much flexibility, but what I mean it someone might try something like: `'{network:_<2s}.{station:_<5s}'`

QuLogic:
If no-one on Windows comes along, you might have to add some sort of verbose assertion failure message. Or maybe focus on the bit that's probably Unicode-related.

QuLogic:
Oh, I think you missed a couple encodings in the tests too.

krischer:
Seems to have finally passed appveyor: http://tests.obspy.org/29243/

There is some unrelated error but the `stationtxt` test suite did run - at least on Python 2.7. I restarted the app veyor build ones again to hopefully also get Python 3 working.

krischer:
Yay. CI passed :-) I'm merging this before the master diverges again...

ThomasLecocq:
Yo... Sorry, I missed that one... You still need the windows test ?

krischer:
@ThomasLecocq No we figured that out. Thanks for asking though :-)

Issue #1098, opened 2015-07-27
formatter for structured archives


ThomasLecocq:
Hi guys,

following the discussion on gitter, and my current not-super-easy way to handle different archive structures, I was wondering we could achieve some brainstorming on the best way to achieve automatic formatting of Trace.stats to a known (or user defined) structure.

Currently, based on what I do with MSNoise, II can generalize to this, if the `format` = "SDS":

```python

known_structures = []
known_structures["SDS"] = ["YEAR","NET","STA","CHAN.TYPE",
                           "NET.STA.LOC.CHAN.TYPE.YEAR.JDAY"]

def to_structured_archive(stats, year, jday, structure="SDS", prefix=None):    
    tmp = os.path.join(known_structures[structure])
    file = tmp.replace('YEAR', "%04i"%year)
    file = file.replace('NET', stats.network)
    file = file.replace('STA', stats.station)
    file = file.replace('LOC', stats.location)
    file = file.replace('CHAN', stats.channel)
    file = file.replace('JDAY', "%03i"%jday)
    file = file.replace('TYPE', "D")
    if prefix:
        return os.path.join(prefix, file)
    else:
        return file
```

the `year` and `jday` param are passed because the `trace` could start before midnight (0.5s before, for example), but in SDS you would want this to be stored on the day, not the day before.

@megies suggested adding a `prefix` kwarg so that a full file path could be obtained from the method directly.

I should add a check if the structure passed is an array, then the structure is most-probably passed and not its nickname...

But, then, we would need to define an allowed vocabulary (STA, NET, etc)... Or... 

```python 
known_structures["SDS"] = ["{s.starttime.year}", "{s.network}"] 
```
could work ?

Anyway, let's beerstorm on that !

megies:
Having `stats` as first argument makes me expect that more information from it is used (time etc.), but you're only using the SEED ID. So why not just have a `seed_id` as first argument?
As long as there's only SDS I would keep it simple and just make a little helper function for getting SDS filename. Or do you have concrete plans for other folder structures?
Also, should avoid Python builtin names (`file`)..

ThomasLecocq:
Well, there is at least "BUD":
```python
data_structure['BUD'] = "NET/STA/STA.NET.LOC.CHAN.YEAR.DAY"
```

or the default format (Win) for Datamark dataloggers:

```python
data_structure['WIN'] = ["YY","YYMM","YYMMDD","YYMMDDHH","YYMMDDHH.MM"]
```

[...]

PR #1135, opened 2015-10-07
new client to read data from local SDS directory structure

megies:
Works in principle, but still some todos left, so not yet ready to merge, opening this PR for visibility of the branch..

### DIFF
--- a/obspy/clients/filesystem/sds.py
+++ b/obspy/clients/filesystem/sds.py
@@ -1,0 +1,194 @@
+# -*- coding: utf-8 -*-
+"""
+obspy.clients.filesystem.sds - read support for SeisComP Data Structure
+=======================================================================
+This module provides read support for data stored locally in a SeisComP Data
+Structure (SDS) directory structure.
+
+The directory and file layout of SDS is defined as:
+
+    <SDSdir>/YEAR/NET/STA/CHAN.TYPE/NET.STA.LOC.CHAN.TYPE.YEAR.DAY
+
+These fields are defined by SDS as follows:
+
+    SDSdir :  arbitrary base directory
+    YEAR   :  4 digit year
+    NET    :  Network code/identifier, up to 8 characters, no spaces
+    STA    :  Station code/identifier, up to 8 characters, no spaces
+    CHAN   :  Channel code/identifier, up to 8 characters, no spaces
+    TYPE   :  1 characters indicating the data type, recommended types are:
+               'D' - Waveform data
+               'E' - Detection data
+               'L' - Log data
+               'T' - Timing data
+               'C' - Calibration data
+               'R' - Response data
+               'O' - Opaque data
+    LOC    :  Location identifier, up to 8 characters, no spaces
+    DAY    :  3 digit day of year, padded with zeros
+
+See https://www.seiscomp3.org/wiki/doc/applications/slarchive/SDS.
+
+:copyright:
+    The ObsPy Development Team (devs@obspy.org)
+:license:
+    GNU Lesser General Public License, Version 3
+    (http://www.gnu.org/copyleft/lesser.html)
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
+
+import glob
+import os
+from datetime import timedelta
+
+from obspy import Stream, read
+from obspy.core.util.misc import BAND_CODE
+
+
+SDS_FMTSTR = os.path.join(
+    "{year}", "{network}", "{station}", "{channel}.{type}",
+    "{network}.{station}.{location}.{channel}.{type}.{year}.{doy:03d}")
+
+
+class Client(object):
+    """
+    Request client for SeisComP Data Structure archive on local filesystem.
+
+    For details see the :meth:`~obspy.clients.filesystem.sds.Client.__init__()`
+    method.
+    """
+    def __init__(self, sds_root, sds_type="D", format="MSEED",
+                 fileborder_seconds=30, fileborder_samples=5000):
+        """
+        Initialize a SDS local filesystem client.
+
+        >>> from obspy.clients.filesystem.sds import Client
+        >>> client = Client("/my/SDS/archive/root")  # doctest: +SKIP
+
+        :type sds_root: str
+        :param sds_root: Root directory of SDS archive.
+        :type sds_type: str
+        :param sds_type: SDS data type identifier, one single character. Types
+            recommended by the SDS definition are: 'D' - Waveform data,
+            'E' - Detection data, 'L' - Log data, 'T' - Timing data,
+            'C' - Calibration data, 'R' - Response data, 'O' - Opaque data. Can
+            also be wildcarded by setting to ``?`` or ``*``.
+        :type format: str
+        :param format: File format the data is stored in, see
+            :func:`~obspy.core.stream.read()` for a list of file formats
+            supported by ObsPy. Usually, SDS archives are stored in "MSEED"
+            format. Can be set to ``None`` for file format autodetection
+            (slowing down the reading).
+        :type fileborder_seconds: float
+        :param fileborder_seconds: Defines in which cases the client checks the
+            previous/next daily file for the requested data (data in SDS
+            archives usually spill over on the day break, at least for a few
+            seconds). For example setting ``fileborder_seconds=30`` means that
+            for a request with ``starttime`` at ``00:00:29`` UTC (or an
+            ``endtime`` at ``23:59:31``), the previous daily file is also
+            checked, if it contains matching data. The maximum of both
+            ``fileborder_seconds`` and ``fileborder_samples`` is used when
+            determining if previous/next day should be checked for data.
+        :type fileborder_samples: int
+        :param fileborder_samples: Similar to ``fileborder_seconds``. The given
+            number of samples is converted to seconds by mapping the band
+            code of the requested channel to sampling frequency. The maximum of
+            both ``fileborder_seconds`` and ``fileborder_samples`` is used when
+            determining if previous/next day should be checked for data.
+        """
+        if not os.path.isdir(sds_root):
+            msg = ("SDS root is not a local directory: " + sds_root)
+            raise IOError(msg)
+        self.sds_root = sds_root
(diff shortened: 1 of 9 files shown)


megies:
@ThomasLecocq, somewhat along a similar topic as #1098.

ThomasLecocq:
@megies Wünderbar !

obspy/io/dirtree/sds.py
```diff
@@ -0,0 +1,131 @@
...
+import os
+from datetime import timedelta
+
+from obspy import Stream, read
+from obspy.core.util.misc import BAND_CODE
+
+
+SDS_FMTSTR = os.path.join(
+    "{year}", "{network}", "{station}", "{channel}.{type}",
+    "{network}.{station}.{location}.{channel}.{type}.{year}.{doy:03d}")
```
krischer: This probably complicates things a bit but I think it should be possible to make this configurable. 

Instead of `doy` I would also use `julday`. Sounds less like a typo and is what we use elsewhere.

megies: Well, it's only used internally.. also, I've found that "julian day" is quite ambiguous, so I find "day of year" clearer..

megies:
Other opinions for the naming? I'm not 100% happy with `obspy.io.dirtree`..

ThomasLecocq:
archive_path ?

dirtree make me think the actual file name is not included ?

ThomasLecocq:
in pure python it's "tm_yday", so why not yday ?

obspy/io/dirtree/sds.py
krischer: > Well, it's only used internally

I think it might be useful to expose this. Maybe someone stores it with `network/station/year/...`. If you limit the temporal tags to `year` and `doy` there is also not too much additional logic.

krischer:
I actually think `dirtree` is quite a good name. Describes exactly what it is.

Even more true if my suggestion of the user-customizable structure is implememented ;-)

megies:
Only other thing I can think of is `io.filesystem` but that might be confusing as well.. `archive_path` doesn't feel like it's intuitively telling what it's about.. probably stay at `dirtree` i guess..

petrrr:
I now had the occasion to have a look at the details here. 

I am wondering if it would not make sense to give this a client like interface. This is what I thought would make most sense when I had a similar idea some time ago. One would instantiate an object with parameters like archive root and file layout (SDS, BUD, etc), and then would just "query" for the data of interest. 

Just a thought.

megies:
Yeah, I was also considering that approach. On the other hand it brings no real practical advantages, I think. There's really hardly any information that would be stored with the object..

obspy/io/dirtree/sds.py
megies: Well, for SDS the structure is defined by the manual. If somebody wants to generalize this, there's nobody keeping him or her from it, it could be outsourced to a more general case outside of `sds.py`.

megies:
> Yeah, I was also considering that approach. On the other hand it brings no real practical advantages, I think. There's really hardly any information that would be stored with the object..

Hmm, thinking more about it, I think I'll probably really should move it into `clients`. Kind of really makes more sense..

ThomasLecocq:
I do agree with @petrrr , it's actually how I use such a function... request data from my archive...

megies:
Refactored into `clients.filesystem.sds`. Otherwise should be good to go, I think. If everybody's happy with the new naming and CI passes I'll rebase/squash and merge later today maybe.

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+
+class Client(object):
+    """
+    Request client for SeisComP Data Structure archive on local filesystem.
+
+    For details see the :meth:`~obspy.clients.filesystem.sds.Client.__init__()`
+    method.
+    """
+    def __init__(self, sds_root, sds_type="D", format="MSEED",
+                 fileborder_seconds=30, fileborder_samples=5000):
```
krischer: The `fileborder_seconds` and `fileborder_samples` arguments are not yet documented.

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+            :func:`~obspy.core.stream.read()` for a list of file formats
+            supported by ObsPy. Usually, SDS archives are stored in "MSEED"
+            format. Can be set to `None` for file format autodetection (slowing
+            down the reading).
+        """
+        if not os.path.isdir(sds_root):
+            msg = ("SDS root is not a local directory: " + sds_root)
+            raise IOError(msg)
+        self.sds_root = sds_root
+        self.sds_type = sds_type
```
krischer: I don't think this needs to be set here. Its already an arguments on the `get_waveforms` method and could just have a default value in that method.

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+
+        >>> from obspy import UTCDateTime
+        >>> t = UTCDateTime("2015-10-12T12")
+        >>> st = client.get_waveforms("IU", "ANMO", "*", "HH?", t, t+30)
+        ... # doctest: +SKIP
+
+        :type seed_id: str
+        :param seed_id: SEED ID of data that should be read
+            (e.g. 'IU.ANMO.00.LHZ'). Wildcards '*' and '?' are supported
+            (e.g. 'IU.*.*.HH?').
```
krischer: Outdated docstring.

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+        :type starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+        :param starttime: Start of requested time window.
+        :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+        :param endtime: End of requested time window.
+        :type cleanup_merge: bool
+        :param cleanup_merge: Whether to perform a cleanup merge on the
+            resulting stream to merge seamless traces originating from
+            different files (:meth:`st.merge(-1)
+            <obspy.core.stream.Stream.merge>`)
```
krischer: It might be useful to be able to set the merge strategy here, e.g. `False` is not merge, any numerical value would correspond to one of the merge methods.

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+        :type sds_type: str
+        :param sds_type: Override SDS data type identifier that was specified
+            during client initialization.
+        :rtype: :class:`~obspy.core.stream.Stream`
+        """
+        if starttime > endtime:
```
krischer: Better use `>=` I guess.

obspy/clients/filesystem/sds.py
megies: I think of this as something you set during initialization as a default and that's hardly ever touched during `get_waveforms()`. I think it logically fits here.

obspy/clients/filesystem/sds.py
megies: Oops, that's right..

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,165 @@
...
+        # time window (plus day before and day after)
+        while t < t_max:
+            year_doy.add((t.year, t.julday))
+            t += timedelta(days=1)
+        year_doy.add((t_max.year, t_max.julday))
+
+        st = Stream()
+        full_paths = set()
+        for year, doy in year_doy:
+            sds_type = sds_type and sds_type or self.sds_type
```
krischer: Does not have to be in the loop and could probably be simplified to

```python
sds_type = sds_type or self.sds_type
```

obspy/clients/filesystem/sds.py
krischer: Fair enough. I just think it might be confusing if the same parameter can be set in two places with one overwriting the other but you might be right in that it enables a more convenient use of the client.

obspy/clients/filesystem/sds.py
megies: Good idea!

obspy/clients/filesystem/sds.py
megies: Ooops!

obspy/clients/filesystem/sds.py
```diff
@@ -0,0 +1,181 @@
...
+        :param format: File format the data is stored in, see
+            :func:`~obspy.core.stream.read()` for a list of file formats
+            supported by ObsPy. Usually, SDS archives are stored in "MSEED"
+            format. Can be set to `None` for file format autodetection (slowing
+            down the reading).
+        :type fileborder_seconds: float
+        :param fileborder_seconds: Defines in which cases the client checks the
+            previous/next daily file for the requested data (data in SDS
+            archives usually spill over on the day break, at least for a few
+            seconds). For example setting `fileborder_seconds=30` means that
```
krischer: RST requires double backticks, otherwise it renders as italic :-(

megies: Really? Oooops!

obspy/clients/filesystem/sds.py
megies: Yeah, kinda awkward. If I remove it in one place, though, I'd remove it from `get_waveforms`. Can always use a different client for different data type.. probably not much harm to leave it in, though, I think.

krischer: Yea leave it in. Might be worthwhile to do the same for the merge strategy as well.

megies: Hmm.. doesn''t work well for this one, we'd run into ambiguities with use of `None` ("no override" vs. "no merge"). I think it's fine, 99.5% of all cases people will want `merge=-1` and thus don't have to specify `merge=..` on a `get_waveforms()` call, I think.

krischer: :+1:

megies:
Squashed and force pushed. Should be good to go, if everybody's happy with the naming `obspy.clients.filesystem.sds`.

ThomasLecocq:
approved !

megies:
Gonna merge this, if anything comes up, let me know here we can still change it in master before next release.

PR #1144, opened 2015-10-15
Enable Travis OSX builds

QuLogic:
OS X builds appear to be publicly available on Travis now, so enable them. Only built on 3.5 right now since the builds start and run a bit slower than the Linux ones. Also, clean up a few things now that we're using conda, like the unnecessary wheelhouse.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,20 @@
 language: c
 
-env:
-  matrix:
-    - PYTHON_VERSION=2.7
-    - PYTHON_VERSION=3.3
-    - PYTHON_VERSION=3.4
-    - PYTHON_VERSION=3.5
-  global:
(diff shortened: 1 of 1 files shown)


krischer:
Nice.

I think adding Python 2.7 to the OS X tests might be worth the additional wait time. That's what most people still use. I agree that its ok to skip testing 3.3 and 3.4 on OS X.

megies:
Nice indeed. :+1: 

> I think adding Python 2.7 to the OS X tests might be worth the additional wait time. That's what most people still use. I agree that its ok to skip testing 3.3 and 3.4 on OS X.

I'd also add Py2.7, right now the slowest builds are both Linux/OSX Py3.5 anyway..

QuLogic:
In [one of the previous builds](https://travis-ci.org/obspy/obspy/builds/85446733), the OS X builds took about the same time regardless of Python version, so I decided to pick the latest. I'm not really sure why that is though.

krischer:
It seems to very random - I guess it just depends on the currently available virtual machines. The linux builds are also super quick on linux for the run you just linked.

megies:
Looks good to merge to me.

PR #1143, opened 2015-10-15
HTTP redirects are not followed by obspy.fdsn.get_waveforms_bulk()

krischer:
When a get_waveforms_bulk() request receives an HTTP redirect it raises an exception and fails.  Ideally a redirect would be followed to complete the request.

In one specific case, when undergoing maintenance at our main data center the IRIS DMC redirects all service requests with an HTTP 307 (Temporary Redirect) to our auxiliary data center.  During recent maintenance when this redirect was in place I tried get_waveforms_bulk() and got this:

```
In [11]: st = client.get_waveforms_bulk(bulk)
FDSNException                             Traceback (most recent call last)
<ipython-input-11-f03c235fa5f9> in <module>()
----> 1 st = client.get_waveforms_bulk(bulk)

.../anaconda3/lib/python3.4/site-packages/obspy-0.10.2-py3.4-macosx-10.5-x86_64.egg/obspy/fdsn/client.py in get_waveforms_bulk(self, bulk, quality, minimumlength, longestonly, filename, attach_response, **kwargs)
    770 
    771         data_stream = self._download(url,
--> 772                                      data=bulk.encode('ascii', 'strict'))
    773         data_stream.seek(0, 0)
    774         if filename:

.../anaconda3/lib/python3.4/site-packages/obspy-0.10.2-py3.4-macosx-10.5-x86_64.egg/obspy/fdsn/client.py in _download(self, url, return_string, data)
   1206         # Catch any non 200 codes.
   1207         elif code != 200:
-> 1208             raise FDSNException("Unknown HTTP code: %i" % code)
   1209         return data
   1210 

FDSNException: Unknown HTTP code: 307

In [12]: 
```
FWIW, the redirection worked just fine for get_waveform() requests.


krischer:
Once this PR is merged `obspy.clients.fdsn` will also follow redirects for POST requests.

@CTrabant Thanks for the redirection server! That helped a lot. Is that a permanent service, i.e. can we write tests against it?

This is not done automatically by Python's stdlib (and I guess neither for other libraries and other languages) for reasons of security and it is actually forbidden according to RFC 2616:

https://www.ietf.org/rfc/rfc2616.txt

```
   If the 307 status code is received in response to a request other
   than GET or HEAD, the user agent MUST NOT automatically redirect the
   request unless it can be confirmed by the user, since this might
   change the conditions under which the request was issued.
```

I guess the worst thing that can happen security wise in ObsPy is that somebody sends a "confidential" station list to some other server if the original one was hacked. That's a very acceptable risk in my opinion.

krischer:
I forgot - if someone wants to test this:

```python
from obspy import UTCDateTime

client = Client("IRIS",
                service_mappings={"dataselect": "http://ds.iris.edu/files/redirect/307/dataselect/1"})

t1 = UTCDateTime("2010-02-27T06:30:00.000")
t2 = t1 + 5
st = client.get_waveforms("IU", "ANMO", "00", "LHZ", t1, t2)
print(st)

print("=============")

t1 = UTCDateTime("2010-02-27T06:30:00.000")
t2 = t1 + 1
t3 = t1 + 3
bulk = [("IU", "ANMO", "*", "BHZ", t1, t2),
        ("IU", "AFI", "1?", "BHE", t1, t3),
        ("GR", "GRA1", "*", "BH*", t2, t3)]
st = client.get_waveforms_bulk(bulk)
print(st)
```

CTrabant:
> @CTrabant <https://github.com/CTrabant> Thanks for the redirection server! That helped a lot. Is that a permanent service, i.e. can we write tests against it?
> 
There are no plans or foreseeable reason to remove the redirections, so you should be able to count on them.  We'll need them for our own testing anyway.

megies:
@krischer, shouldn't this go into `releases`? Also it seems ready?

krischer:
Ah I forgot about this. I'll add some tests later today then we can merge it.

It will only be two commits so I'll cherry-pick them onto the releases branch after the merge. Really annoying that one cannot change the target branch of a PR on github...

megies:
> It will only be two commits so I'll cherry-pick them onto the releases branch after the merge. Really annoying that one cannot change the target branch of a PR on github...

Yep, sounds good.

megies:
> I guess the worst thing that can happen security wise in ObsPy is that somebody sends a "confidential" station list to some other server if the original one was hacked. That's a very acceptable risk in my opinion.

How about sending authentication credentials for access to restricted data? Is that affected by this scenario?

krischer:
> How about sending authentication credentials for access to restricted data? Is that affected by this scenario?

Hmm...yea that would affect it. Right now it just blindly redirects everything. @CTrabant What is the DMC's stance on this? You don't employ HTTPS so a man-in-the-middle attack would be very possible.

megies:
> Hmm...yea that would affect it. Right now it just blindly redirects everything. @CTrabant What is the DMC's stance on this? You don't employ HTTPS so a man-in-the-middle attack would be very possible.

If credentials are at stake we should be very careful..

QuLogic:
> shouldn't this go into releases?

I thought we were going to end up going for 0.11.0 anyway.

megies:
> I thought we were going to end up going for 0.11.0 anyway.

Yeah, but @krischer thought we'd might as well do a last bug fix 0.10.x release at the same time.. should probably not be a big deal..

CTrabant:
> On Oct 27, 2015, at 3:08 AM, Tobias Megies <notifications@github.com> wrote:
> 
> Hmm...yea that would affect it. Right now it just blindly redirects everything. @CTrabant <https://github.com/CTrabant> What is the DMC's stance on this? You don't employ HTTPS so a man-in-the-middle attack would be very possible.
> 
> If credentials are at stake we should be very careful..
> 
Yeah, good to be careful with credentials.

The DMC does not have a strong recommendation at the moment and support with common tools varies.  The Perl redirection we use in our Fetch clients will submit credentials to the redirected site.  wget will do the same, follow and authentication.  cURL will not, but it can be turned on with the --location-trusted option.  FWIW.

megies:
So, to be on the safe side.. I would propose to only follow the redirect when a request without credentials is done and otherwise print a meaningful message that a redirect was attempted but it is not followed due to credentials in the request.

Any other proposals?

QuLogic:
Sounds reasonable.

CTrabant:
> So, to be on the safe side.. I would propose to only follow the redirect when a request without credentials is done and otherwise print a meaningful message that a redirect was attempted but it is not followed due to credentials in the request.
> 
> Any other proposals?
> 
Sounds prudent.

I suggest offering a flag to set to allow redirections to be followed when authenticating, otherwise a user is left trying to figure out how to circumvent the failure.  The DMC's services were all redirected today for maintenance, so it is not uncommon.

megies:
> I suggest offering a flag to set to allow redirections to be followed when authenticating, otherwise a user is left trying to figure out how to circumvent the failure. The DMC's services were all redirected today for maintenance, so it is not uncommon.

Fair enough, if the default is `False` then it's up to the user to decide whether he/she wants their credentials following the redirect.

krischer:
The client now has an additional keyword argument `force_redirect`. It defaults to `False` and will raise an exceptions with an helpful error message if people use it together with credentials. If `True` it will always redirect.

screen shot 2015-11-18 at 14 34 32

This should do the trick. Ready to go from my point of view if it passes the review.

megies:
:+1: appveyor fail is unrelated, comes from a03403c76e46b397a0cfcdc38e

megies:
Test fails are unrelated.

megies:
Cherry-picked into `releases` (unfortunately we get those cherry-picked commits as duplicates in `master` :/)

krischer:
Hmm...does that work? If yes git is more impressive then I would have thought...

But do we really want this into `releases`? Its not a bug fix but rather a new feature I think.

megies:
Hmm.. without that new option it would have fit into releases, I think.. anyway.. pushed already, won't hurt anybody.

krischer:
True enough :-)

QuLogic:
On a related note, those temporary redirects do not correctly handle encryption. That is, if you go to `https://ds.iris.edu/files/redirect/307/...`, you get redirected to `http://service.iris.edu/fdsnws/...`. They shouldn't be stripping encryption.

Of course, these are test redirects so it's not that important, but just for future consideration...

QuLogic:
@CTrabant Actually, that is a bug that affects all redirects on the IRIS site, e.g., open https://www.iris.edu/ will redirect to http://www.iris.edu/hq/.

CTrabant:
> @CTrabant <https://github.com/CTrabant> Actually, that is a bug that affects all redirects on the IRIS site, e.g., open https://www.iris.edu/ <https://www.iris.edu/> will redirect to http://www.iris.edu/hq/ <http://www.iris.edu/hq/>.
> 
> On a related note, those temporary redirects do not correctly handle encryption. That is, if you go to https://ds.iris.edu/files/redirect/307/..., you get redirected to http://service.iris.edu/fdsnws/.... They shouldn't be stripping encryption
> 
Thanks for noting this.  We are looking into how to avoid the change of protocol when redirecting.  It's a fairly low priority task so it may not happen right away.

PR #1056, opened 2015-10-15
decorators mangling call syntax in API docs

megies:
We really need to fix the mangled call syntax in API docs for decorated function/method calls...

screenshot from 2015-05-20 15 50 41
screenshot from 2015-05-20 15 50 13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ install:
         flake8
         future
         lxml
+        decorator
         sqlalchemy
         mock
         nose
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ install:
   - "conda create -q --yes -n test python=%PYTHON_VERSION%"
   - "activate test"
   # Install default dependencies
-  - "conda install -q --yes pip numpy scipy matplotlib=1.3.1 lxml sqlalchemy flake8 mock nose gdal"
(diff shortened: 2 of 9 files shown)


megies:
This is a first shot at using `decorator` module, just to get some movement in here. If anybody has a better solution, step forward.. ;-)

One thing is not working yet, concerning `deprecated_keywords` decorator: It's erroring out if the length of the `args` in the call syntax has changed, which was working OK with the old `functools` approach. Couldn't work around this, yet..

Otherwise, the call syntax in docs/interactive help now is what is in the source.. for the first time in a looooong time..

Also concerns packaging, as this introduces a new dependency (it's packaged for Debian, so I'm guessing it's OK for other packages as well..?).

krischer:
> Otherwise, the call syntax in docs/interactive help now is what is in the source.. for the first time in a looooong time..

:tada: 


> Also concerns packaging, as this introduces a new dependency (it's packaged for Debian, so I'm guessing it's OK for other packages as well..?).

Its a very old and stable module so I don't think we have to be worried. It should be available everywhere.

megies:
Reverting the change for `deprecated_keywords` decorator, as the normal use of the new decorator module hickups when length of `args` changes in between..

Pretty sure this can be done with the lower-level functionality in `decorator` module but couldn't figure it out yet..

@obspy/packagers please remember to add `decorator` module as a new dependency.

PR #1145, opened 2015-10-16
Minor bug in triggering: last occurence of characteristic function falling gelow off threshold is missed

megies:
First discovered by @petrrr, by switching to numpy 1.10. Numpy 1.10 is now showing a warning when a boolean index array gets expanded to match the shape of the array that is being indexed.

See posts on numpy mailing list:
https://mail.scipy.org/pipermail/numpy-discussion/2015-August/073467.html

```
obspy/signal/tests/test_trigger.py /home/megies/git/obspy/obspy/signal/trigger.py:349: VisibleDeprecationWarning: boolean index did not match indexed array along dimension 0; dimension is 3395 but corresponding boolean dimension is 3394
  of.extend(ind2[np.diff(ind2) > 1].tolist())
```

The `np.diff()` can't make use of the last item, so the last time the trigger's characteristic function is falling below the off-threshold was not used. We have to add that last item by hand.

I'm not sure if this glitch actually had a negative impact right now, though. My feeling is that this only has an influence in extreme edge cases, namely when the last time the characteristic function falls below the off-threshold is actually closely preceeded by an exceedance of the on-threshold. Even then, with `max_len_delete=False` (default) my guess is that the length of that trigger might be wrong, and that only with `max_len_delete=True` this could actually lead to a missed trigger.. but the code is not easy to read..

(btw.. numpy is also working around `DeprecationWarning`s not shown by default on Python nowadays.. ;-))

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -28,6 +28,9 @@
   - obspy.signal:
     * Bug fixed within rotate.rotate2ZNE(). Additionally it can now also
       perform the inverse rotation (see #1061).
+    * Bug fixed in triggering. When using option `max_len_delete` and a trigger
+      occurred right before the end of data, one trigger was potentially lost
+      (see #1145 for details).
   - obspy.station:
     * Plotting responses across multiple channels is more robust now in
       presence of some strange channels (e.g. with zero sampling rate,
--- a/obspy/signal/tests/test_trigger.py
+++ b/obspy/signal/tests/test_trigger.py
@@ -112,7 +112,8 @@ class TriggerTestCase(unittest.TestCase):
         # check that max_len_delete drops the picks
         picks_del = triggerOnset(cft, 1.5, 1.0, max_len=50,
                                  max_len_delete=True)
-        np.testing.assert_array_equal(picks_del, on_of[np.array([0, 1, 5, 6])])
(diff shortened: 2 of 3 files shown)


megies:
The first commit keeps the old behavior and only fixes avoids the numpy warning.

The second commit fixes the last trigger-off occurence previously being missed. Actually in one test this issue lead to the last trigger being ignored with `max_len_delete=True`, as suspected in the issue description. The change in the second commit actually leads to the trigger being kept in some cases where it could also be dropped. At the end of the trace there can be an ambiguous scenario when:
 * `max_len_delete=True`
 * characteristic function goes above on-threshold shortly before end of data
 * subsequently, characteristic function does not go below off-threshold until end of the data
 * `end-of-data - on-time` of that trigger is less than "`max_len`" maximum time of triggers that should be kept
 * from the data it can not be told whether the ongoing trigger will be longer than `max_len` (should be discarded) or if it drops below off_threshold in less than `max_len` (should be kept)

Old behavior (image from inside the `test_triggerOnset` test case in `obspy/signal/tests/test_trigger.py`:
figure_x_bug

New behavior (retaining the last trigger):
figure_x_fixed



Anyway.. better to include one more trigger than to lose it.

PR #1146, opened 2015-10-16
.fdsn: add EMSC web service shortcut

petrrr:
This (trivial) PR just adds a shortcut for the EMSC web service.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -30,6 +30,8 @@ master:
      individual steps of instrument response removal in frequency domain
      (see #1116).
  - obspy.clients.fdsn:
+   * Replace FDSN webservice shortcut `NERIES` with `EMSC` and deprecate the 
+     `NERIES` shortcut, will be removed in a future release (see #1146). 
    * Now requests gzipped data for the XML files. Much smaller files!
    * The station service can now also be used to download files in the text
      format. This has limited information but is much faster.
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -182,6 +182,15 @@ class Client(object):
                    "Please change to 'Client('NIEP')'. This re-routing will "
                    "be removed in a future release.")
             warnings.warn(msg)
+        # Deprecate FDSN URL-shortcut 'NERIES' in favour of 'EMSC', see #1146.
+        # TODO: remove in 0.12.x or 1.x release
+        elif base_url.upper() == "NERIES":
+            base_url = "EMSC"
+            msg = ("FDSN short-URL 'NERIES' has been replaced by 'EMSC'. "
+                   "Please change to 'Client('EMSC')'. This re-routing will "
+                   "be removed in a future release.")
+            warnings.warn(msg)
+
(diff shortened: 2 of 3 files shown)


krischer:
Do we really need different names for the same endnode? I know we already have that ORFEUS but I think we should remove that as well.

megies:
Well the old name `"NERIES"` might not be super-ideal, but that all results from this strange `seismicportal` web page name..

petrrr:
@krischer: Not sure I understand your comment. 

Are you referring to the `ORFEUS` vs. `ODC` alias? Well, in the context of EIDA we usually use `ODC` (Orfeus data center) when referring to the EIDA node, or even `KNMI` (the hosting organization), but I find the later inappropriate. Orfeus is the European organization. If I would have to pick one I'd rather use `ODC`, but I found `ORFEUS` already around.

`NERIES`: Sorry, I completely missed that this was around and probably wouldn't have associated it with the EMCS catalog immediately. So okay, if you want to avoid aliases, we can close this PR. On the other hand, NERIES refers to a pretty old project (even NERA, the subsequent project is now finished for quite some time).

@megies: I agree that this `seismicportal` webpage, which basically is the product of various projects is misleading.

In any case, I leave it to you to decide ...

megies:
So, EMSC is what actually is behind seismicportal? I'm OK with adding the alias then, maybe we can deprecate the `NERIES` keyword at the same time..

megies:
Ping. @petrrr, would you mind adding a deprecation warning when the old short URL for seismicportal.eu is used? Or what's your view on this? I think having multiple short URLs for one endpoint might get confusing..

megies:
Ping, @petrrr..

petrrr:
@megies: Sorry, for ignoring your ping. To many threads => low responsiveness ...
Sure, I guess deprecating `NERIES` is a valid option. What about `ORFEUS`? I find `ODC` more appropriate, and Orfeus might be reserved to refer to the organization itself, will offer other products in future. So deprecating `ORFEUS` would make sense as well. 

Will update the PR, so you can cherry pick it.

krischer:
I feel like `"ORFEUS"` is too widely used and I would rather not deprecate it. I would actually prefer to remove the `"ODC"` mapping to be honest even though it is not strictly speaking correct.

petrrr:
@krischer: I vote against removing `ODC`. This abbreviation/handle is used quite consistently within the Orfeus EIDA federation to refer to the data center hosted at KNMI, it is used in most documents and on the Orfeus EIDA webpages, in the current mastertable, monitoring, etc. I therefore would argue it should be kept. If you think that `ORFEUS` cannot or should not be deprecated (a was actually not proposing it, just asking), there is probably a good use case for having two aliases.

krischer:
@petrrr Fair enough. That does seem like a legitimate use case for two aliases.

Can you add an entry to the changelog? Then we can merge this.

petrrr:
What happens, is there a merge conflict on `CHANGELOG.txt`? Do I need to rebase or do we merge manually?

QuLogic:
It'd be easier if you rebase.

krischer:
Looks good. Thanks!

PR #1147, opened 2015-10-19
Added in nd file format and updated test

mpanning:
Probably could use some more exception handling, etc., and the file read is still a little awkward for ReadNDFile, but it reads nd files.


obspy/taup/tests/test_velocity_model.py
```diff
@@ -20,14 +20,17 @@
 
 class TauPyVelocityModelTestCase(unittest.TestCase):
     def test_read_velocity_model(self):
-        for i in range(0, 2):
+        for i in range(0, 4):
```
QuLogic: I don't know why this loop was written this way in the first place, but now that it's 4 elements long, it makes more sense to be `for filename in ['iasp91.tvel', 'iasp91_w_comment.tvel', ...]:` and leave out `i` entirely.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +483,125 @@ def readTVelFile(cls, filename):
...
+
+        # Read all lines from file to enable identifying top and bottom values
+        # for each layer and find named discontinuities if present        
+        with open(filename) as modfile:
+            lines = modfile.readlines()
+
+        # Loop through to fill data array and locate named discontinuities
+        ii = 0
+        for line in lines:
+            if line[0] == '#': # Skip lines starting with '#'
```
QuLogic: Better to split at `'#'` and drop everything after rather than checking for it in several places below.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +483,125 @@ def readTVelFile(cls, filename):
...
+                continue
+            if ii == 0:
+                data = []
+                for item in line.split():
+                    if item[0] == '#':
+                        break
+                    data = np.append(data, float(item))
```
QuLogic: `np.append` is inefficient. Create a list and convert to array after.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +483,125 @@ def readTVelFile(cls, filename):
...
+                ii = ii + 1
+            else:
+                if ((line.split()[0].lower() == 'mantle') or 
+                    (line.split()[0].lower() == 'moho')):
+                    mohoDepth = data[ii - 1,0]
+                elif ((line.split()[0].lower() == 'outer-core') or
+                      (line.split()[0].lower() == 'cmb')):
+                    cmbDepth = data[ii - 1,0]
+                elif ((line.split()[0].lower() == 'inner-core') or
+                      (line.split()[0].lower() == 'iocb')):
```
QuLogic: It's very inefficient to call `.split()[0].lower()` 6 times in the majority of cases.

obspy/taup/tests/test_velocity_model.py
mpanning: That's fixed in a new commit

mpanning:
Tried to implement suggestions from QuLogic and hopefully cleaned up style issues causing Travis Cl failure

krischer:
Still some formatting issues. If you just run

```
python /path/to/obspy/obspy/core/tests/test_code_formatting.py
```

locally you see them as well. Alternatively you can run

```
obspy-runtests
```

which runs the whole ObsPy test suite.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +485,125 @@ def readTVelFile(cls, filename):
                              cls.default_cmb, cls.default_iocb, 0,
                              maxRadius, True, layers)
 
+    @classmethod
+    def readNDFile(cls, filename):
```
krischer: Can you rename this to `read_nd_file()` to suite the rest of ObsPy? I know `obspy.taup` does not adhere to it very strictly but new stuff should probably try to do it.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +485,125 @@ def readTVelFile(cls, filename):
...
+        ignored
+
+        :param filename: The name of the file to read.
+        :type filename: str
+
+        :raises ValueError: If model file is in error.
+        """
+        mohoDepth = cls.default_moho
+        cmbDepth = cls.default_cmb
+        iocbDepth = cls.default_iocb
```
krischer: It appear to me like these depths could well be extracted from the file. Each time there is one of "mantle", "outer core", ... it marks a boundary. This is definitely worth it as the code should in principle also work with non-earth models.

Also please use snake case, .e.g `moho_depth`, ... for variable names.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +485,125 @@ def readTVelFile(cls, filename):
...
+        with open(filename) as modfile:
+            lines = modfile.readlines()
+
+        # Loop through to fill data array and locate named discontinuities
+        ii = 0
+        for line in lines:
+            # Strip off anything after '#'
+            line = line.split('#')[0]
+            if len(line.split()) == 0:  # Skip empty or comment lines
+                continue
```
krischer: Would be a bit simpler with

```python
line = line.split("#")[0].split()
if not line:
    continue
```

Then you also don't have to repeat the final split in all conditions further down this function.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +485,125 @@ def readTVelFile(cls, filename):
...
+            if ii == 0:
+                data = []
+                for item in line.split():
+                    data.append(float(item))
+                data = np.array(data)
+                ii = ii + 1
+            else:
+                if re.match('[a-zA-Z]', line.split()[0]):
```
krischer: Could be replaced (if the previous suggestion is also followed) with

```python
if line[0].isalpha():
    ...
```

obspy/taup/velocity_model.py
mpanning: These are extracted from the file later on.  This just sets defaults so that I can make the call to VelocityModel the same even if the file does not explicitly set the discontinuities.

krischer: My bad - I'm sorry. I actually saw this when I kept reading but forgot to delete my comment...

obspy/taup/velocity_model.py
mpanning: This does not work with "outer-core" and "inner-core", so I'll stick with re.match, I think.

krischer: Yea my comments are really on point today ;-) Sorry about that...

After looking at it a bit more careful: the regex will only look at the first character so it is more or less equivalent with `line.split()[0][0].isalpha()`.

It might better to check the amount of components the line splits into - its only a named discontinuity if it splits in one item. Then it will also work for discontinuities that are named for example `440` which people might use.

mpanning: Sure, I could do it that way instead.  I'll leave the ValueError raise in there now if it's not one of the 6 standard named discontinuities, until we've got some checks to make sure the phase parsing and everything works as it's supposed to for arbitrarily named discontinuities.

krischer:
Thanks for this :-) Looks good to me.

Two more things (optional). Can you please

1. Add yourself to the list of contributors: https://github.com/obspy/obspy/blob/master/obspy/CONTRIBUTORS.txt
2. Add this new addition to the changelog: https://github.com/obspy/obspy/blob/master/CHANGELOG.txt

obspy/taup/velocity_model.py
```diff
@@ -484,6 +483,127 @@ def readTVelFile(cls, filename):
...
+                                                          'moho')):
+                        moho_depth = data[ii - 1, 0]
+                    elif ((line[0].lower() == 'outer-core') or
+                          (line[0].lower() == 'cmb')):
+                        cmb_depth = data[ii - 1, 0]
+                    elif ((line[0].lower() == 'inner-core') or
+                          (line[0].lower() == 'iocb')):
+                        iocb_depth = data[ii - 1, 0]
+                    else:
+                        raise ValueError("Unrecognized discontinuity name")
```
QuLogic: The name should be mentioned in the exception.

mpanning:
Does AppVeyor ever work?  Is there some command to attempt the run again?  I assume it's fine with running on linux and mac, but it would be nice to get the windows test as well.

CHANGELOG.txt
```diff
@@ -73,6 +73,10 @@ master:
      - new option `PPSD.plot(..., cumulative=True)` for a cumulative plot of
        the histogram, i.e. a non-exceedence percentage visualization, similar
        to the `percentile` option.
+ - obspy.taup
+   * Added support for nd file format for input velocity models.  Allows for 
+     named discontinuities at arbitrary depths allowing for less Earth like
+     models.
```
krischer: Super minor nitpicks: Can you please remove the superfluous double space in the first line? And add this issue's number (#1147) for future reference.

obspy/taup/velocity_model.py
```diff
@@ -484,6 +483,128 @@ def readTVelFile(cls, filename):
                              cls.default_cmb, cls.default_iocb, 0,
                              maxRadius, True, layers)
 
+    @classmethod
+    def read_nd_file(cls, filename):
+        """
+        This method reads in a velocity model from a "nd" ASCII text file, the
+        format used by Xgbm. The name of the model file for model "modelname"
+        should be "modelname.nd".
```
krischer: Can you please reformat the docstring to contain a one line summary, e.g.:

```python
def read...():
    """
    Method to read in a velocity model from a "nd" ASCII text file.

    This method....
"""

Then there will be a proper one line description in the documentation for this function. See here for details: https://www.python.org/dev/peps/pep-0257/

Also best but a blank line where you want a line to end (e.g. two lines down) - sphinx otherwise will mash the lines together has it cannot determine where the line break should be.

krischer:
Ready to merge from my point of view once my two minor nit picks are taken care of.

Thanks a lot for this :-)

krischer:
AppVeyor has some issues recently...we should probably look into them and see if they are our fault...

obspy/taup/velocity_model.py
QuLogic: The "Method to" is not necessary in the one line summary. Just say "Read in a velocity model ...".

mpanning:
Seems done to me except the Appveyor fail

megies:
If I remember correctly, we use the model representation in `tvel` to pre-compute the `.npz` files and only really use these at runtime in a normal user session. So.. should this also be plugged in to `taup_create.py`? And if so, is it problematic to store two different representations of `IASP91` in `taup/data`? If we only use the `iasp91.nd` for testing maybe it should be moved to `taup/tests/data`?

megies:
Btw.. simply restarted the appveyor build, so these build errors seem to be random..

mpanning:
Currently, the `iasp91.nd` file is only present in `taup/tests/data`, not `taup/data`.  It is identical to `iasp91.tvel` once it's used to create a VelocityModel object, so I don't think there's need to have both in `taup/data`, as it would produce the same `.npz` file.  It does seem like we should also include some version of a precomputed PREM as well, but there's the problem of the ocean layer not yet being supported, but that's certainly outside the scope of this PR.  I might take a hack at that after AGU, but I've got a lot on my plate right now.

krischer:
Looks good. Thanks! :-)

megies:
Yeah, thanks for contributing this piece! :+1:

PR #1156, opened 2015-10-26
close matplotlib figures when options for saving to files are used

megies:
In many places when plots are made and options are used to output directly to file, figures are not closed. This should be changed.. Also, code should ideally be cleaned up to use a consistent set of `kwargs` (`outfile`, `plot`, `show`, ...).


megies:
Not ready for merging, mostly opened as a reminder to fix this eventually..

Issue #1099, opened 2015-07-29
Internal mseed warning error, xseed.station Parser error


nlscelli:
Hi, 
have two different problems: 
1) while downloading huge amount of mseed files from FDSN clients sometimes i get this kind of errors:
/usr/lib/python2.7/dist-packages/obspy/mseed/core.py:356: InternalMSEEDReadingWarning: (null): Warning: Data integrity check for Steim-1 failed, last_data=140, xn=142
  warnings.warn(msg, InternalMSEEDReadingWarning)
/usr/lib/python2.7/dist-packages/obspy/mseed/core.py:356: InternalMSEEDReadingWarning: (null): Warning: Data integrity check for Steim-1 failed, last_data=219, xn=218
  warnings.warn(msg, InternalMSEEDReadingWarning)
/usr/lib/python2.7/dist-packages/obspy/mseed/core.py:356: InternalMSEEDReadingWarning: (null): Warning: Data integrity check for Steim-1 failed, last_data=-298, xn=-299
  warnings.warn(msg, InternalMSEEDReadingWarning)
/usr/lib/python2.7/dist-packages/obspy/mseed/core.py:356: InternalMSEEDReadingWarning: (null): Warning: Data integrity check for Steim-1 failed, last_data=-69, xn=-74
  warnings.warn(msg, InternalMSEEDReadingWarning)
/usr/lib/python2.7/dist-packages/obspy/mseed/core.py:356: InternalMSEEDReadingWarning: (null): Warning: Data integrity check for Steim-1 failed, last_data=146, xn=135
  warnings.warn(msg, InternalMSEEDReadingWarning)

The program keeps going, but as the program stucks at this point for a while this slows down the procedure. What does it actually mean? 

2) Whenever I have an xml file (both dowloaded from obspy clients or indipendently retrieved) that I want to parse to write the response with writeRESP i get this:
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-8-d4355b9ef9e8> in <module>()
----> 1 sp = Parser("bau")

/usr/lib/python2.7/dist-packages/obspy/xseed/parser.pyc in __init__(self, data, debug, strict, compact)
    105         # if a file name is given, read it directly to the parser object
    106         if data:
--> 107             self.read(data)
    108 
    109     def __str__(self):

/usr/lib/python2.7/dist-packages/obspy/core/util/decorator.pyc in new_func(*args, **kwargs)
    305                             except IOError:
    306                                 pass
[...]

PR #1100, opened 2015-10-29
XSEED/SEED Parser should print informative message on StationXML input files

megies:
See #1099, the warning message when parsing a StationXML file should be clearly stating that StationXML files are not supported by the Parser.


megies:
grrrr... wrong head branch..

megies:
Continued in #1160.

PR #1027, opened 2015-10-30
inverse of locations2degree and including azimuth on a spherical earth

megies:
While obspy.core.util.geodetics.gps2DistAzimuth works on an elliptical earth, locations2degree uses a spherical earth but only computes distance, so there was no way to compute azimuth on a spherical earth. Furthermore, I needed the solution to the direct problem, that is given a point, azimuth and distance, I wanted to compute the second point. So I generalized locations2degree to also return azimuths and wrote  a new function for the direct problem, which I think might be of general interest and could be added to geodetics.

```python
#!/usr/bin/env python
import matplotlib.pyplot as plt
import numpy as np
from obspy.core.util.geodetics import locations2degrees, gps2DistAzimuth
import math

def locations2degreesazi(lat1, long1, lat2, long2):
    """
    Convenience function to calculate the great circle distance, forward and
    backward azimuth between two points on a spherical Earth.

    This method uses triangles on a sphere. For more accurate values use the
    geodesic distance calculations of geopy (https://github.com/geopy/geopy).

    http://en.wikipedia.org/wiki/Great-circle_navigation#Course_and_distance
    http://en.wikipedia.org/wiki/Great-circle_navigation#cite_note-2

    :type lat1: float
    :param lat1: Latitude of point 1 in degrees
    :type long1: float
    :param long1: Longitude of point 1 in degrees
    :type lat2: float
    :param lat2: Latitude of point 2 in degrees
    :type long2: float
    :param long2: Longitude of point 2 in degrees
    :rtype: tuple of 3 floats
    :return: Distance in degrees, forward and backward azimuth as floating point
             numbers.
    """
    # Convert to radians.
    lat1 = math.radians(lat1)
    lat2 = math.radians(lat2)
    long1 = math.radians(long1)
    long2 = math.radians(long2)
    long_diff = long2 - long1
    gd = math.degrees(
        math.atan2(
            math.sqrt((
                math.cos(lat2) * math.sin(long_diff)) ** 2 +
                (math.cos(lat1) * math.sin(lat2) - math.sin(lat1) *
                    math.cos(lat2) * math.cos(long_diff)) ** 2),
            math.sin(lat1) * math.sin(lat2) + math.cos(lat1) * math.cos(lat2) *
            math.cos(long_diff)))
    az = math.degrees(
        math.atan2(
            math.sin(long_diff),
            math.cos(lat1) * math.tan(lat2) - math.sin(lat1) * math.cos(long_diff)
        ))
    az2 = math.degrees(
[...]

PR #1164, opened 2015-10-31
Add get_..._geo to taup

andreww:
It can be useful to know the location (latitude and longitude) of points on a raypath, e.g. to plot CMB bounce points for PcP. This adds two new methods to TauPyModel which take the location of the source and recever, and decorate the path with locations.

In order to allow this we add three new methods to the TauPyModel:

* get_travel_times_geo: this is simply a wrapper around get_travel_times, but it takes the event location and station position (latitude and longitude) instead of the source-station distance in degrees.

* get_pierce_points_geo: this also takes the event location and station position, but also adds the latitude and longitude to all pierce_points in the returnd Arrivals object. These can be used to plot peice points on a map with the basemap plugin to matplotlib, for example.

* get_get_ray_paths_geo: this works like get_pierce_points_geo, but decorates the ray path with location information. This can be used for simple tomographic corrections, for example.

Three simple tests are added. These are derived from three existing tests for the methods that
take an angular distance between the source and receiver. We need to add tests for the locations
along paths and at peirce points too.

A new attribute, flatteningOfEarth, has been added to the TauModel class, and this is respected by the new methods. This is always set to 0.0 (which makes the calculations operate on a sphere) but opens up the possibility of considering an elliptical Earth (or Mars) model. However, the effect of changing the value will only result in a change in the an angular distance between the source and receiver (for fixed latitude and longitude). This results in a small change in trave time compared to the effect of distorting the layers and raypath, so there is not much point in using this unless we want to implement the ellipticity correction (e.g. Dziewonski and Gilbert, 1975).


megies:
Please add yourself to https://github.com/obspy/obspy/blob/master/obspy/CONTRIBUTORS.txt. :sparkles:

andreww:
Does AppVeyor install geographiclib before running the tests? If not, if the reason I get errors in the results because of an error how I use obspy/geodetics?

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,157 @@
...
+                    pathList.append(diffTDG)
+                arrival.path = np.concatenate(pathList)
+
+    except ImportError:
+        # geographiclib is not installed ...
+        # and  obspy/geodetics does not help much
+        msg = "Not able to evaluate positions on path or pierce points. " + \
+              "Arrivals object will not be modified. " + \
+              "Install the Python module 'geographiclib' to solve this issue."
+        warnings.warn(msg)
```
megies: I wouldn't warn here but raise an ImportError instead, with the message you have there.

obspy/taup/tests/test_tau.py
```diff
@@ -107,6 +107,28 @@ def test_p_iasp91_manual(self):
         self.assertAlmostEqual(p_arrival.purist_distance, 35.00, 2)
         self.assertEqual(p_arrival.purist_name, "P")
 
+    def test_p_iasp91_geo_manua(self):
```
megies: All tests that can only work if `geographiclib` is installed should be skipped if it is not installed, see e.g. https://github.com/obspy/obspy/blob/master/obspy/geodetics/tests/test_util_geodetics.py#L60

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,157 @@
...
+    :type station_longitude_in_deg: float
+    :param radius_of_earth_in_km: Radius of the Earth in km
+    :type radius_of_earth_in_km: float
+    :param flattening_of_earth: Flattening of earth (0 for a sphere)
+    :type station_longitude_in_deg: float
+
+    :return: distance_in_deg
+    :rtype: float
+    """
+    try:
```
megies: Sorry, I know I recommended to move the import here myself, but looking at it now, I think code would be better readable if you bundle the check for geographiclib at the module imports and set a `HAS_GEOGRAPHICLIB` variable that can be used for if/else clauses inside the routines.

I'll open a PR, that's fastest..

megies:
> Does AppVeyor install geographiclib before running the tests? If not, if the reason I get errors in the results because of an error how I use obspy/geodetics?

You can try to add it here, at least it's in anaconda on Linux, not sure if they package it for Windows too: https://github.com/obspy/obspy/blob/master/appveyor.yml#L31

Otherwise skipping those tests if geographiclib is not installed is fine, too (see above comment).

megies:
See andreww/obspy#1.

megies:
Oh and please add a note to the CHANGELOG and add the new taup submodule here: https://github.com/obspy/obspy/blob/master/misc/docs/source/packages/obspy.taup.rst#L30

megies:
Really like this PR. :+1:
This looks very convenient: https://github.com/obspy/obspy/pull/1164/files#diff-1667c484b259a6cf774ac6ea98b2e520R382

obspy/taup/tests/test_tau.py
andreww: Some of these tests _ought_ to work without `geographiclib`. Specifically, we should get the same travel times, and the pierce points / paths should have the same distance-depth-time for the `_geo` and non geo variants of the methods. I think I can use the new `HAS_GEOGRAPHICLIB` module variable to do some sensible tests for this (i.e. test that the answers we can get in both cases do not depend on having `geographiclib` installed.

My fear at the moment is that obspy\geodetics is giving a different epicentral distance if `geographiclib` is or is not installed, and that is what is causing the tests to fail under AppVeyor... Or I've stuffed up the interface with  obspy\geodetics. In either case, sorting out the added tests should be instructive.

megies:
> It's also a good idea to avoid reimplementing all of geographiclib. I think this interface needs some thought.

Feel free to work on refactoring `obspy.geodetics`, it certainly could use some cleanup.. maybe you can have a look at #1027 as well and help with fitting it in?

> Does AppVeyor install geographiclib before running the tests? If not, if the reason I get errors in the results because of an error how I use obspy/geodetics?

Again

> Some of these tests ought to work without geographiclib. Specifically, we should get the same travel times, and the pierce points / paths should have the same distance-depth-time for the _geo and non geo variants of the methods. I think I can use the new HAS_GEOGRAPHICLIB module variable to do some sensible tests for this (i.e. test that the answers we can get in both cases do not depend on having geographiclib installed.

Those tests that should work regardless of geographiclib yes/no, should of course be run unconditionally.
E.g. [this](https://github.com/obspy/obspy/pull/1164/files#diff-4c1e073b497ba969e67ccda82ef28cb7R206) won't work w/o geographiclib, right? So it should either be skipped conditionally or needs some other handling within the test (if part of it makes sense still w/o geographiclib).

> 
> My fear at the moment is that obspy\geodetics is giving a different epicentral distance if geographiclib is or is not installed, and that is what is causing the tests to fail under AppVeyor... Or I've stuffed up the interface with obspy\geodetics. In either case, sorting out the added tests should be instructive.

How big is the difference? If it's small "enough" just up the margin in the test a bit. If it's large, it should be fixed in the code so that the difference is small.. ;-)

The test fail in appveyor is not meaningful, it's a [build error](https://ci.appveyor.com/project/obspy/obspy/build/1.0.1091-master#L723) randomly popping up.. we didn't figure out yet whats going wrong with appveyor for us recently.. I'll try to restart appveyor build whenever I see that one, so that you can get meaningful results for windows..

andreww:
Thanks for https://github.com/andreww/obspy/pull/2 @megies - that was exactly the problem.

I think everything is working properly now and I've fixed the typos etc. I've added to a couple of tests and now check we get the same values for the latitude and longitude of pierce points and the ray path as the Java implementation.

I also added a couple of tests for non-standard ellipsoids, These could also be used as tests for the direct problem. What are we after for added a couple of tests for none-standard ellipsoids. Is this what is needed for #1027?

obspy/taup/tests/test_tau.py
```diff
@@ -16,6 +16,11 @@
 import numpy as np
 
 from obspy.taup import TauPyModel
+# We need HAS_GEOGRAPHICLIB, but we need to be able to set
+# in all three places... there must be a better way!
+import obspy.geodetics.base as geobase
```
megies: This import is really obfuscating.. you only ever use `HAS_GEOGRAPHICLIB` from `obspy.geodetics.base`. So why not make it simply `from obspy.geodetics.base import HAS_GEOGRAPHICLIB` and avoid inventing new submodule aliases?

obspy/taup/tests/test_tau.py
```diff
@@ -107,6 +112,64 @@ def test_p_iasp91_manual(self):
...
+        """
+        Manual test for P phase in IASP91 given geographical input.
+
+        This version of the test checks that things still work when
+        geographiclib is not installed.
+        """
+        has_geographiclib_real = geobase.HAS_GEOGRAPHICLIB
+        geobase.HAS_GEOGRAPHICLIB = False
+        taup_geo.HAS_GEOGRAPHICLIB = False
+        tau.HAS_GEOGRAPHICLIB = False
```
megies: I don't understand what you're doing here.. either geographiclib is available or not. Why not just always look at `HAS_GEOGRAPHICLIB` from `obspy.geodetics.base`? Am I missing something..?

andreww: Yes, a comment is probably needed, and there may be a better way of doing it. 

What I am trying to do is test the code paths like: `taup.get_ray_paths_geo()` -> `taup. taup_geo.calc_dist()` -> `geodetics. gps2dist_azimuth()` -> `geodetics. calc_vincenty_inverse()`, which get called when `geographiclib` is not installed.  However, I want to run these tests even if `geographiclib` happens to be installed on the testing machine. One way of doing this is to run a test unconditionally that sets `HAS_GEOGRAPHICLIB` to `False` (whatever it started off as) and then return the variable to the initial state once the test is done. 

However, on a machine with `geographiclib` installed, I discovered that if I just set `obspy.geodetics.base.HAS_GEOGRAPHICLIB` to `False` the code in `obspy.taup.tau` behaves as if `HAS_GEOGRAPHICLIB` is `True`. If I set `obspy.taup.tau.HAS_GEOGRAPHICLIB` to `False`the code in `obspy.geodetics.base` behaves as if `HAS_GEOGRAPHICLIB` is `True`. Having written this down I find this behaviour so surprising I've gone back and checked. This really is what happens (at least on my machine running python 2.7).

I think this is down to a subtle difference between:

    from obspy.geodetics.base import HAS_GEOGRAPHICLIB
    HAS_GEOGRAPHICLIB = False

and:

    import obspy.geodetics.base
    obspy.geodetics.base.HAS_GEOGRAPHICLIB = False 

In the first case we add  `HAS_GEOGRAPHICLIB` to the local namespace (and the local dictionary that represents the local namespace), so updates do not propagate back. In the second we really use the namespace of the base module. Because we have imports like: `obspy.geodetics.base import HAS_GEOGRAPHICLIB` in `obspy.taup.taup_geo` and `obspy.tau.tau` we need to directly change the variable in those modules too. Changing the variable in one of these three places does not update the other two. I think this behaviour only happens when the module variable is immutable, and that you get a real reference if you do this with a list, for example. 

So, the only alternative I can think of would be to avoid `from obspy.geodetics.base import HAS_GEOGRAPHICLIB` everywhere. This will make the tests look nicer, but the rest of the code (where something like a read only variable is probably fine) more ugly.

QuLogic: This is basically the explanation in [`mock.patch` documentation](https://docs.python.org/3/library/unittest.mock.html#where-to-patch).

obspy/geodetics/base.py
```diff
@@ -207,6 +219,8 @@ def gps2dist_azimuth(lat1, lon1, lat2, lon2):
         negative for southern hemisphere)
     :param lon2: Longitude of point B in degrees (positive for eastern,
         negative for western hemisphere)
+    :param a: Radius of Earth in m. If ``None`` WGS84 is assumed.
```
QuLogic: @megies missed updating this docstring.

obspy/geodetics/tests/test_util_geodetics.py
```diff
@@ -110,6 +109,72 @@ def test_calcVincentyInverse2(self):
         self.assertAlmostEqual(alpha12, calc_alpha12)
         self.assertAlmostEqual(alpha21, calc_alpha21)
 
+    def test_calcVincentyInverseTabulated(self):
+        """ Tabulated results for Vincenty Inverse
+
+        Table II of Vincenty's paper (T. Vincenty 1975, "Direct
```
QuLogic: I think this might be wrapped a little too eagerly. Wrapping should occur after 79 characters.

obspy/geodetics/tests/test_util_geodetics.py
```diff
@@ -110,6 +109,72 @@ def test_calcVincentyInverse2(self):
...
+        pp.88-93) has five test examples for the forward and inverse
+        problem (with results rounded to 0.00001 seconds of arc and
+        1 mm). The inverse versions of these are implemented here. Note
+        the non-standard (old) ellipsoid usage. Here we test that we
+        match these examples for the inverse problem.
+        """
+        # Row "A"
+        # NB: for this case there seems to be a typo in
+        #     the tabulated data. Tabulated data is commented
+        #     out and values from geographiclib are used in there place
```
QuLogic: their

obspy/geodetics/tests/test_util_geodetics.py
```diff
@@ -110,6 +109,72 @@ def test_calcVincentyInverse2(self):
...
+        # dist = 14110.526170
+        dist = 14039.003954192352
```
QuLogic: Is there a reason to use kilometres instead of metres like the docstring/paper/etc.?

obspy/taup/tau.py
```diff
@@ -19,7 +20,8 @@
 from .taup_path import TauP_Path
 from .taup_pierce import TauP_Pierce
 from .taup_time import TauP_Time
-
```
QuLogic: Should not remove this empty line.

obspy/taup/tau.py
```diff
@@ -379,6 +381,141 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+            attributes.
+        :rtype: :class:`Arrivals`
+        """
+        distance_in_deg = calc_dist(source_latitude_in_deg,
+                                    source_longitude_in_deg,
+                                    station_latitude_in_deg,
+                                    station_longitude_in_deg,
+                                    self.model.radiusOfEarth,
+                                    self.model.flatteningOfEarth)
+        arrivals = self.get_ray_paths(source_depth_in_km, distance_in_deg,
```
QuLogic: Why are you calling `get_ray_paths` in `get_travel_times_geo`?

obspy/taup/tau.py
```diff
@@ -379,6 +381,141 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
         return Arrivals(sorted(rp.arrivals, key=lambda x: x.time),
                         model=self.model)
 
+    def get_travel_times_geo(self, source_depth_in_km, source_latitude_in_deg,
+                             source_longitude_in_deg, station_latitude_in_deg,
```
QuLogic: I think maybe this should be `receiver` instead of `station`.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Functions to handle geographical points
+
+These functions are used to alow taup models
```
QuLogic: These lines are wrapped too early.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+to process input data with source and station
+locations given as lstitudes and longitudes. The
```
QuLogic: latitudes

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+functions are set up to hande an elliptical Earth
```
QuLogic: handle

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+model, but we do not have ellipticity corrections
+for travel times. Although changing the
+shape of the Earth from something other than spherical
+would change the epicentral distance the change
```
QuLogic: Comma after distance.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+in distance along the ray path has a larger effect,
```
QuLogic: It's not quite clear what distance you mean here.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+and we do not make that correction.
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
```
QuLogic: Missing empty line after.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+        values = gps2dist_azimuth(source_latitude_in_deg,
+                                  source_longitude_in_deg,
+                                  station_latitude_in_deg,
+                                  station_longitude_in_deg,
+                                  a=radius_of_earth_in_km*1000.0,
+                                  f=flattening_of_earth)
+        distance_in_km = values[0]/1000.0
+        # NB - km2deg assumes spherical Earth... generate a warning
+        if flattening_of_earth != 0.0:
+            msg = "Assuming spherical Earth when calculating angle. " + \
```
QuLogic: What angle? Should be epicentral distance, or some such.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+
+if HAS_GEOGRAPHICLIB:
+    from geographiclib.geodesic import Geodesic
+
+
+def calc_dist(source_latitude_in_deg, source_longitude_in_deg,
+              station_latitude_in_deg, station_longitude_in_deg,
+              radius_of_earth_in_km, flattening_of_earth):
+    """
+    Given the source and station location, calculate the azimuth and distance
```
QuLogic: Missing period at end of sentence.

obspy/taup/tau.py
```diff
@@ -379,6 +381,141 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+                             station_longitude_in_deg, phase_list=("ttall",)):
+        """
+        Return travel times of every given phase given geographical data.
+
+        :param source_depth_in_km: Source depth in km
+        :type source_depth_in_km: float
+        :param source_latitude_in_deg: Source location latitude in degrees
+        :type source_latitude_in_deg: float
+        :param source_longitude_in_deg: Source location longitue in degrees
```
QuLogic: longitude

obspy/taup/tau.py
```diff
@@ -379,6 +381,141 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+                              source_longitude_in_deg, station_latitude_in_deg,
+                              station_longitude_in_deg, phase_list=("ttall",)):
+        """
+        Return ray paths of every given phase with geographical info.
+
+        :param source_depth_in_km: Source depth in km
+        :type source_depth_in_km: float
+        :param source_latitude_in_deg: Source location latitude in degrees
+        :type source_latitude_in_deg: float
+        :param source_longitude_in_deg: Source location longitue in degrees
```
QuLogic: longitude

obspy/taup/tau.py
```diff
@@ -379,6 +381,141 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+                          source_longitude_in_deg, station_latitude_in_deg,
+                          station_longitude_in_deg, phase_list=("ttall",)):
+        """
+        Return ray paths of every given phase with geographical info.
+
+        :param source_depth_in_km: Source depth in km
+        :type source_depth_in_km: float
+        :param source_latitude_in_deg: Source location latitude in degrees
+        :type source_latitude_in_deg: float
+        :param source_longitude_in_deg: Source location longitue in degrees
```
QuLogic: longitude

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+
+    :param source_latitude_in_deg: Source location latitude in degrees
+    :type source_latitude_in_deg: float
+    :param source_longitude_in_deg: Source location longitue in degrees
```
QuLogic: longitude

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+                        station_longitude_in_deg, radius_of_earth_in_km,
+                        flattening_of_earth):
+    """
+    Add geographical information to arrivals
+
+    :param arrivals: Set of taup arrivals
+    :type: :class:`Arrivals`
+    :param source_latitude_in_deg: Source location latitude in degrees
+    :type source_latitude_in_deg: float
+    :param source_longitude_in_deg: Source location longitue in degrees
```
QuLogic: longitude

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+    :type source_longitude_in_deg: float
+    :param station_latitude_in_deg: Station location latitude in degrees
+    :type station_latitude_in_deg: float
+    :param station_longitude_in_deg: Station location longitude in degrees
+    :type station_longitude_in_deg: float
+    :param radius_of_earth_in_km: Radius of the Earth in km
+    :type radius_of_earth_in_km: float
+    :param flattening_of_earth: Flattening of earth (0 for a sphere)
```
QuLogic: Earth

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
```
QuLogic: Earth

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+        ellipsoid = Geodesic(a=radius_of_earth_in_km * 1000.0,
+                             f=flattening_of_earth)
+        g = ellipsoid.Inverse(source_latitude_in_deg, source_longitude_in_deg,
+                              station_latitude_in_deg,
+                              station_longitude_in_deg)
+        azimuth = g['azi1']
+        line = ellipsoid.Line(source_latitude_in_deg,
+                              source_longitude_in_deg, azimuth)
+
+        # We may need to update many arrival objects
```
QuLogic: Over-eager line wrapping.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+                for pierce_point in arrival.pierce:
+                    pos = line.ArcPosition(np.degrees(pierce_point['dist']))
+                    diffTDG = np.array([(
+                        pierce_point['p'],
+                        pierce_point['time'],
+                        pierce_point['dist'],
+                        pierce_point['depth'],
+                        pos['lat2'],
+                        pos['lon2'])], dtype=TimeDistGeo)
+                    pathList.append(diffTDG)
```
QuLogic: You already have the length of `arrival.pierce`; you should pre-allocate the replacement before the loop instead of re-allocating each time.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+                for path_point in arrival.path:
+                    pos = line.ArcPosition(np.degrees(path_point['dist']))
+                    diffTDG = np.array([(
+                        path_point['p'],
+                        path_point['time'],
+                        path_point['dist'],
+                        path_point['depth'],
+                        pos['lat2'],
+                        pos['lon2'])], dtype=TimeDistGeo)
+                    pathList.append(diffTDG)
```
QuLogic: Ditto.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+                                            radius=radius_of_earth_in_km)
+    return distance_in_deg
+
+
+def add_geo_to_arrivals(arrivals, source_latitude_in_deg,
+                        source_longitude_in_deg, station_latitude_in_deg,
+                        station_longitude_in_deg, radius_of_earth_in_km,
+                        flattening_of_earth):
+    """
+    Add geographical information to arrivals
```
QuLogic: Missing period at end of sentence.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,159 @@
...
+                        path_point['depth'],
+                        pos['lat2'],
+                        pos['lon2'])], dtype=TimeDistGeo)
+                    pathList.append(diffTDG)
+                arrival.path = np.concatenate(pathList)
+
+    else:
+        # geographiclib is not installed ...
+        # and  obspy/geodetics does not help much
+        msg = "You need to install the Python module 'geographiclib'."
```
QuLogic: for...? This message is not totally clear.

obspy/taup/tests/test_tau.py
QuLogic: On the other hand, maybe this test is a bit mis-aimed. All you really need to know is whether `calc_dist` returns something sane in both cases.

obspy/taup/tests/test_tau.py
```diff
@@ -176,6 +239,81 @@ def test_pierce_p_iasp91(self):
...
+        m = TauPyModel(model="iasp91")
+        arrivals = m.get_pierce_points_geo(source_depth_in_km=10.0,
```
QuLogic: You should catch (and check for) the warning that's generated here.

obspy/taup/tests/test_tau.py
```diff
@@ -310,6 +448,115 @@ def test_single_path_iasp91(self):
...
+        interpolated_actual_lat = np.interp(
+            sample_points,
+            np.round(np.degrees(arrivals[0].path['dist']), 2),
+            np.round(arrivals[0].path['lat'], 2))
+        interpolated_actual_lon = np.interp(
+            sample_points,
+            np.round(np.degrees(arrivals[0].path['dist']), 2),
+            np.round(arrivals[0].path['lon'], 2))
+
+        self.assertTrue(np.allclose(interpolated_actual_depth,
```
QuLogic: `np.testing.assert_allclose`

obspy/taup/tests/test_tau.py
```diff
@@ -310,6 +448,115 @@ def test_single_path_iasp91(self):
...
+
+        interpolated_actual = np.interp(
+            sample_points,
+            np.round(np.degrees(arrivals[0].path['dist']), 2),
+            np.round(6371 - arrivals[0].path['depth'], 2))
+
+        # NB: we do not check path['lat'] and path['lon']
+        # here, as these are not calculated when geographiclib
+        # is not installed.
+        self.assertTrue(np.allclose(interpolated_actual,
```
QuLogic: `np.testing.assert_allclose`

obspy/taup/tests/test_tau.py
```diff
@@ -176,6 +239,81 @@ def test_pierce_p_iasp91(self):
...
+        tau.HAS_GEOGRAPHICLIB = has_geographiclib_real
+        self.assertEqual(len(arrivals), 1)
+        p_arr = arrivals[0]
+
+        # Open test file.
+        filename = os.path.join(DATA, "taup_pierce_-h_10_-ph_P_-deg_35")
+
+        expected = np.genfromtxt(filename, skip_header=1)
+
+        # NB: we do not check pierce['lat'] and pierce['lon']
```
QuLogic: Check that they _don't_ exist?

obspy/taup/tests/test_tau.py
megies: I guess @QuLogic, you're proposing something like this?

```python
from mock import patch

from obspy.geodetics.base import HAS_GEOGRAPHICLIB
print(HAS_GEOGRAPHICLIB)

with patch("obspy.geodetics.base.HAS_GEOGRAPHICLIB", False):
    from obspy.geodetics.base import HAS_GEOGRAPHICLIB
    print(HAS_GEOGRAPHICLIB)
    from obspy.taup.tau import HAS_GEOGRAPHICLIB
    print(HAS_GEOGRAPHICLIB)
    from obspy.taup.taup_geo import HAS_GEOGRAPHICLIB
    print(HAS_GEOGRAPHICLIB)

from obspy.geodetics.base import HAS_GEOGRAPHICLIB
print(HAS_GEOGRAPHICLIB)
```
```
True
False
False
False
True
```

obspy/geodetics/base.py
megies: Right, thanks. @andreww, can you update that please?

obspy/taup/taup_geo.py
megies: and it should be "allow" I guess

obspy/taup/tests/test_tau.py
QuLogic: Huh, that actually worked? What I read from the documentation there seemed to agree with @andreww about why it _wouldn't_.

megies: I'm a bit confused myself, too.. actually to start with I tried to confirm that what @andreww tried to do first and reportedly failed to do (simply overwrite `HAS_GEOGRAPHICLIB` during testing), and it seemed to work for me.. but seeing your comment I turned to `patch` fast without double-checking.
```python
import obspy

obspy.geodetics.base.HAS_GEOGRAPHICLIB = False

from obspy.scripts.runtests import run_tests

run_tests(tests=["geodetics", "taup"])
```

QuLogic: Does that work too? Maybe it's a Python 2/3 thing?

megies: Ok, so the simple overwriting `HAS_GEOGRAPHICLIB` does *not* work in our testing context in CI. See.. (results are in the stdout printed at end of test run)
 * https://github.com/obspy/obspy/commit/1db9225db817e807cfaaa52b905ffbabbc6ddf8a
 * Travis: https://travis-ci.org/obspy/obspy/builds/89802361
 * Appveyor: 
https://ci.appveyor.com/project/obspy/obspy/build/1.0.1103-test_patching_HAS_GEOGRAPHICLIB


Next try with mock coming up (https://github.com/obspy/obspy/commit/8c48b0f9be61b7cc17c7ef53831b51612e157ea2)

andreww: I don't think you avoid this by testing calc_dist rather than testing get_travel_times_geo. You still need to convince taup.taup_geo and geodetics.base to exercise the "with geographiclib" and "without geographiclib" code paths which, if we don't change the way `HAS_GEOGRAPHICLIB` is imported, means we need to set at least two variables. And I think  you *really* want to know if calc_dist() fails to return something sane if geographiclib is not installed, even if you happen to be testing on a machine where it is installed.

One option would be to call calc_vincenty_inverse() directly from the fallback code in taup.taup_geo (so we avoid having decide if geographiclib is installed in gps2dist_azimuth) but then we end up duplicating code and we would still need to change the `if HAS_GEOGRAPHICLIB:` to a `try ... except` block.

Maybe what we should do is change `from ..geodetics.base import HAS_GEOGRAPHICLIB` to `import geodectics.base` and `HAS_GEOGRAPHICLIB` to `geodectics.base.HAS_GEOGRAPHICLIB` in taup.tau and taup.taup_geo so that we only need to change `geodectics.base.HAS_GEOGRAPHICLIB` in the tests?

CHANGELOG.txt
```diff
@@ -89,7 +89,14 @@ master:
...
      models (see #1147).
-
+   * Added three methods (`get_travel_times_geo()`, `get_pierce_points_geo()`
+     and `get_ray_paths_geo()`) to `TauPyModel` to handle station and
+     event location data as latitude and longitude, instead of the source to
+     station distance in degrees. In addition `get_ray_paths_geo()` and
+     `get_pierce_points_geo()` decorate the returned pierce points and ray
+     paths with the latitude and longitude of each point. Some functionality
+     needs the `geographiclib` module to be installed. (See #1164.)
+
```
QuLogic: Trailing whitespace.

obspy/taup/tau.py
```diff
@@ -20,6 +21,8 @@
 from .taup_pierce import TauP_Pierce
 from .taup_time import TauP_Time
```
QuLogic: Sorry, this is not the empty line you should keep. There should be two after the the imports.

obspy/taup/tests/test_tau.py
QuLogic: It's not really an either/or. There should be a test for `calc_dist` regardless, I think.

CHANGELOG.txt
andreww: The empty line? This is the line at the end of the "master" block and before the "0.10.x" block. It was already present and has just moved below the new text.

megies: There's spaces in that seemingly empty line. I guess your text editor injected those. It's no `.py` file so doesn't make our code formatting checks fail, but still those spaces should be removed. :-)

obspy/taup/tests/test_tau.py
megies: > And I think you really want to know if calc_dist() fails to return something sane if geographiclib is not installed, even if you happen to be testing on a machine where it is installed.

Well, you're absolutely right of course.

If testing the wrapper functions that decide what to do in their inside depending on `HAS_GEOGRAPHICLIB` proves to be hard to do, can't we just fall back to just testing the routines that get called in the inside, for both cases (geographiclib yes/no)?

CHANGELOG.txt
andreww: Ah, OK. Fixed this.

obspy/taup/tests/test_tau.py
andreww: In principle I suppose only `obspy.geodetics` should care if geographiclib is installed or not. Then this stops being a problem (because `taup.tau` and `taup.taup_geo` won't need to look at `HAS_GEOGRAPHICLIB`).  To do this we need a handful of new features in `obspy.geodetics` - specifically a solution to the direct geodesic problem, and something that looks a bit like geographiclib's "line" object. Something like #1027 is a start in that direction but that does not have everything.

obspy/taup/taup_geo.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # We may need to update many arrival objects
+        # and each could have pierce points and a
+        # path
+        for arrival in arrivals:
+
+            if arrival.pierce is not None:
+                geo_pierce = np.empty(arrival.pierce.shape, dtype=TimeDistGeo)
+                i = 0
+                for pierce_point in arrival.pierce:
```
megies: `for i, pierce_point in enumerate(arrival.pierce):`
same in the block below..

obspy/taup/taup_geo.py
megies: ping..

obspy/taup/tests/test_tau.py
megies: ..in any case, we should have a solution we can live with within the next week, to get the release ready..

andreww: Is the current approach, introduced in ba935f496fa3bc26 something you can live with? I'm struggling to think of a cleaner alternative.

obspy/taup/taup_geo.py
QuLogic: Ah, thanks, I meant to mention that too.

obspy/taup/tests/test_tau.py
QuLogic: I think maybe this should be okay if @megies agrees, but there are other ways to test individually using `mock`. For example, run `get_travel_times_geo` and patch `add_geo_to_arrivals`, then check that it is/isn't called as required. Something like:
```python
with mock.patch('obspy.taup.tau.HAS_GEOGRAPHICLIB', False):
    with mock.patch('obspy.taup.tau.add_geo_to_arrivals',
                    return_value=<something useful, like an identity lambda>) as func:
        m.get_travel_times(...)
   func.assert_called_once_with(...)
   func.assert_not_called()
```

megies: > I think maybe this should be okay if @megies agrees

Yeah, I too think we can live with this, looks like the simplest approach. The mock approach proposed by @QuLogic is probably cleaner, but also more complex.

megies:
I didn't try out this new functionality myself, but from looking at the code changes it looks good.
I guess this should be ready to merge unless anybody else brings something up.

krischer:
Sorry for re-joining this discussion so late. It does look mostly good :-)

Can you please add the flattening parameter to the `TauPyModel` object and then use these parameters for the flattening? That would make the geo functions more generally useable.

It is currently on the underlying `VelocityModel` instance but all the models are 1D and none of the formats we currently have support specifying the flattening (which I don't think they ever can as the method works on a spherical earth). Thus this currently cannot be set.

Also please add a note to all the geo functions and the constructor of the `TauPyModel` class that no matter what the specified flattening only the geographical calculations will use it - the actual traveltime and raypath calculations are performed on a spherical earth. That probably does not matter for most people but some people care and for them it matters a lot.

andreww:
No problem moving the flattening parameter (which always defaults to a spherical Earth) and adding documentation warning about its use, but what is the format for adding documentation about attributes? Do these belong in the class docstring, even if it gets set up in `__init__()`? 

I'm assuming we don't want to add this as an argument to the constructor, but rather leave it as something for an expert to alter by changing the attribute in the `TauPyModel` instance?

krischer:
> I'm assuming we don't want to add this as an argument to the constructor, but rather leave it as something for an expert to alter by changing the attribute in the TauPyModel instance?

I would actually be fine on having this in the constructor with a sensible default value. No need to hide it in my opinion.

obspy/taup/tau.py
```diff
@@ -283,12 +286,20 @@ class TauPyModel(object):
...
-    def __init__(self, model="iasp91", verbose=False):
+    def __init__(self, model="iasp91", verbose=False, earth_flattening=0.0):
         """
         Loads an already created TauPy model.
 
         :param model: The model name. Either an internal TauPy model or a
             filename in the case of custom models.
+        :param earth_flattening: Flattening parameter for Earth's ellipsoid
+            (i.e. a-b/a, where a is the semimajor equatorial radius and b is
+            the seminminor polar radius). A value of 0 (the default) gives a
```
QuLogic: semiminor

obspy/taup/tau.py
```diff
@@ -379,6 +391,167 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+        :param phase_list: List of phases for which travel times should be
+            calculated. If this is empty, all phases will be used.
+        :type phase_list: list of str
+
+        :return: List of ``Arrival`` objects, each of which has the time,
+            corresponding phase name, ray parameter, takeoff angle, etc. as
+            attributes.
+        :rtype: :class:`Arrivals`
+
+        Note that the conversion from source and receiver latitudes and
```
QuLogic: This might need to go in `.. note::`, or at least before the parameters. @krischer?

obspy/taup/tau.py
```diff
@@ -303,6 +314,7 @@ def __init__(self, model="iasp91", verbose=False):
         """
         self.verbose = verbose
         self.model = TauModel.from_file(model)
+        self.flatteningOfEarth = earth_flattening
```
QuLogic: I know the TauPy code is bad at this due to its Java origins, but please use snake case for new code.

obspy/taup/tau.py
krischer: :+1: to both suggestions. I would wrap in a `.. note::` directive and move it before the arguments. I think that looks better and it also should be easier to read.

obspy/taup/tau.py
```diff
@@ -379,6 +391,167 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
+            attributes.
+        :rtype: :class:`Arrivals`
+
+        Note that the conversion from source and receiver latitudes and
+        longitudes to epicentral distances respects the model's flattening
+        parameter, so this calculation can be performed for a ellipsoidal
+        or spherical Earth. However, the actual traveltime and raypath
+        calculations are performed on a spherical Earth. Ellipticity
+        corrections (e.g. Dziewonski and Gilbert 1976, Geophys. J. R. astr.
+        Soc. 44:7-17) are not made.
```
krischer: Can you please convert that to an actual reference? For ObsPy this means to add a bib file here: https://github.com/obspy/obspy/tree/master/misc/docs/source/bibliography

and then you can refer to it in the docstring, e.g. like this: `[Dziewonski1976]_`

The documentation then contains nice clickable links and this overview: http://docs.obspy.org/citations.html

Thanks :-)

obspy/taup/tau.py
```diff
@@ -379,6 +391,170 @@ def get_ray_paths(self, source_depth_in_km, distance_in_degree=None,
...
 
+    def get_travel_times_geo(self, source_depth_in_km, source_latitude_in_deg,
+                             source_longitude_in_deg, receiver_latitude_in_deg,
+                             receiver_longitude_in_deg, phase_list=("ttall",)):
+        """
+        Return travel times of every given phase given geographical data.
+
+        .. note::
+
+        Note that the conversion from source and receiver latitudes and
```
QuLogic: This needs to be indented to be in the note.

PS, there are no notifications when you push updates, so you need to ping the ticket if you want it re-reviewed.

andreww:
Thanks for the warning, @QuLogic. I've indented the three notes.

QuLogic:
Thanks for the new feature!

PR #1166, opened 2015-11-03
PPSD: Some more fixes and refactoring

megies:
Related to #1130.


megies:
Note to self, store psd data as `np.float16`, that's easily sufficient for the db scaled amplitude values..

megies:
@aringler-usgs, just to let you know, I did some further refactoring, esp. slightly changed the binning at the low-period end and cleaned up the initialization of all necessary PPSD attributes and serialization to `npz` format.
The test against IRIS Mustang that you set up is still mostly good (had to up the tolerance to 1.3db), so everything should be fine.

PR #1179, opened 2015-11-16
Send earthworm/waveserver errors to stderr, instead of stdout

jmfee-usgs:
Writing errors to stdout is awkward at best for scripts that generate output on stdout.

### DIFF
--- a/obspy/earthworm/waveserver.py
+++ b/obspy/earthworm/waveserver.py
@@ -15,6 +15,7 @@ from future.utils import native_str
 
 import socket
 import struct
+import sys
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


megies:
Fair enough for me. Maybe those messages should even only pop up when setting a `debug=True` flag on client initialization? (But I never really used this client, so I don't know exactly how vital those messages are in normal operation..)

krischer:
I'd rather see this hooked up to Python's `logging` module but we also don't yet widely use it within ObsPy. So I am also fine with merging this simple change here. It is definitely an improvement upon the current version.

megies:
There's hardly much use of `logging` module across ObsPy, so I'll merge this for now..

Issue #800, opened 2014-05-17
PPSD.load static method not included in docs


megies:
Some the `PPSD.load` staticmethod doesn't make it into the API docs... http://docs.obspy.org/packages/autogen/obspy.signal.spectral_estimation.PPSD.html

PR #1013, opened 2015-11-17
private _read_XXX/_write_XXX/_is_XXX hides all docstrings of those functions in docs

megies:
and therefore invalidates a lot of effort put into those documentation strings - also no one knows about special options for certain read/write functions - I strongly vote against making those private - we can't cover all options over a general read/write function

### DIFF
--- a/misc/docs/source/_templates/autosummary/class.rst
+++ b/misc/docs/source/_templates/autosummary/class.rst
@@ -11,7 +11,7 @@
 
   .. autosummary::
   {% for item in attributes %}
-     ~{{ name }}.{{ item }}
+     {{ item }}
   {%- endfor %}
   {% endif %}
   {% endblock %}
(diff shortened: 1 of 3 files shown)


megies:
> IIRC, the patched copy of autosummary added private class members, but not module members. We can just change the patch to apply to modules as well.

You're right. Anybody remember why we don't use the sphinx builtin `autosummary` extension, btw?
In any case, private functions are now visible.. based this on master to be able to check our docs buildbot. Need to cherry-pick into releases when merged.

QuLogic:
> Anybody remember why we don't use the sphinx builtin autosummary extension, btw?

Splitting things up like this is basically the whole reason.

megies:
screenshot from 2015-11-18 14 04 51

krischer:
:+1: Looks great!

QuLogic:
Hmm, does this fix #800, too?

krischer:
I don't think so. At least I cannot see it here: http://docs.obspy.org/master/packages/autogen/obspy.signal.spectral_estimation.PPSD.html#obspy.signal.spectral_estimation.PPSD

megies:
> Hmm, does this fix #800, too?

No, unfortunately not..

PR #1181, opened 2015-11-19
bug in beachball plotting routine?

krischer:
Hi obspy team,

the following three moment tensors throw an error in my obspy 0.10.2 when using the imaging.beachball routine. The mopad function works fine.

    #!/usr/bin/env python
    
    from obspy.imaging.beachball import Beach
    #from obspy.imaging.mopad_wrapper import Beach

    mt = [0.000, -1.232e25, 1.233e25, 0.141e25, -0.421e25, 2.531e25] #CMTEVENT C010398B
    #mt = [0.0, -12.32, 12.33, 1.41, -4.21, 25.31]
    #mt = [0., -1., 1., 1., -1., 1.]
    beach = Beach(mt, xy=(50, 50), width=50)

    ax = plt.gca()
    ax.add_collection(beach)
    ax.set_aspect("equal")
    ax.set_xlim((-120, 120))
    ax.set_ylim((-20, 120))

    plt.show()

Error msg:

      File "./obspy_bug.py", line 10, in <module>
        beach = Beach(mt, xy=(50, 50), width=50)
      File "/usr/local/lib/python2.7/dist-packages/obspy/imaging/beachball.py", line 122, in Beach
        plot_zerotrace=True, xy=xy, width=width)
      File "/usr/local/lib/python2.7/dist-packages/obspy/imaging/beachball.py", line 401, in plotMT
        azi[n][0] = az
    IndexError: index 3 is out of bounds for axis 0 with size 3

Just wanted to report in case you didn't see this already,

Matthias


krischer:
That works pretty well. It now falls back to the mopad plotting routines if it encounters some issues.

```python
$ python funky_test.py
/Users/lion/workspace/code/obspy/obspy/imaging/beachball.py:57: UserWarning: Encountered an exception while plotting the beachball. Falling back to the mopad wrapper which is slower but more stable.
  warnings.warn(msg)
```

megies:
Just a tiny semantic error in test routine..

krischer:
Hmm...it works on my machine. Maybe the latest change works?

krischer:
CI passes now. Merging.

QuLogic:
What's up with the funny squiggle in the example image?

krischer:
I also wondered - I guess its some kind of edge case - thats why the normal plotting routine completely fails. mopad can at least still produce a picture. The code is really quite complex so I did not feel like diving into it...

krischer:
If you plot it small enough nobody will see it ;-)

barsch:
there have been quite a few bugfixes to GMT's meca class (e.g. http://gmt.soest.hawaii.edu/projects/gmt/repository/changes/trunk/src/meca/utilmeca.c) since the implementation of the original beachball class (2008!!!) - so I would suggest either to redo it from scratch der just delete the old implementation and rely on mopad

krischer:
Looking at the C code I like the "solution" of this PR a lot more ;-)

PR #1183, opened 2015-11-19
Travis: one build with minimum dependencies

megies:
(to catch errors with functionality added in more recent numpy/scipy
versions, see e.g. numpy.fft.rfftfreq in current implementation of
lanczos interpolation)

@krischer: plot_lanczos_windows is not tested, only appears in code
block directive, so the plot in docs fails silently without popping up
as test fail, even when running tests with our minimum dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: c
 
 matrix:
   include:
+    # do one build run with our minimum dependencies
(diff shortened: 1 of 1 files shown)


megies:
Our minimum dependencies are mostly not in anaconda, so I'm using the lowest version numbers available.

PR #1184, opened 2015-11-19
plot_lanczos_windows untested and not working with minimum dependencies

krischer:
see #1183


krischer:
This should fix things.

krischer:
I also added the plot to the gallery.

megies:
In case of even-length array the last entry has different sign. Don't know if this makes a difference for your algorithm, but I feel slightly uncomfortable about it.. ;)

screenshot from 2015-11-20 08 23 04

QuLogic:
It's generally correct. For an even length, one of the "positive" values is 0 and the positive Nyquist value is not returned. But the positive and negative Nyquist values should be the same (in all cases? I don't recall about fully complex). If you ask for a real transform, then they obviously return the positive Nyquist instead of negative.

krischer:
I wrapped in into an `np.abs()` call. I also think that it would not matter in a real algorithm but I used it to generate the x-values to plot it and there it should matter. 

I'm not entirely sure why there was no line connecting the first to last point but I guess it does not matter much as the plots only show a small fraction of the arrays.

krischer:
Should fixed for real this time around...

PR #1187, opened 2015-11-20
PPSD: method to add additional npz files to existing PPSD

megies:
..so that data stored in separate npz files can be merged together. Includes basic test.


megies:
Good to merge.

PR #1182, opened 2015-11-23
sc3ml to Inventory

megies:
Hi,

I've started writing a routine that can directly import Seiscomp3 [sc3ml](http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.7/) (in addition to stationXML) to an Inventory class. Before spending more time, I would like to discuss if there currently exists an easy way to import sc3ml format to Obspy? And would the option to import sc3ml to Inventory be a valuable addition to ObsPy?

Like to hear your thoughts on this!


obspy/io/stationxml/tests/test_stationxml.py
```diff
@@ -28,6 +28,7 @@ class StationXMLTestCase(unittest.TestCase):
     """
     """
     def setUp(self):
+        print("WOW")
```
megies: remove..

megies:
Made this into a PR from the branch you mentioned..

Jollyfant:
Ok, I'm going to do some more assert checks and remove that little print statement.

QuLogic:
> For example, the station town tag may be missing for stationXML, while it is included for sc3ml.

That's part of the `Site` element, I think.

Jollyfant:
@QuLogic, that is right. Sc3ml seems to be a little more verbose when it comes to site meta-data.
[sc3ml](http://www.orfeus-eu.org/fdsnws/station/1/query?net=AI&level=station&format=sc3ml)
[stationXML](http://www.orfeus-eu.org/fdsnws/station/1/query?net=AI&level=station)

krischer:
Sorry for this taking so long. I polished it a bit and sent a PR with small changes to your branch. Once you merge it we can finally merge this PR! Thanks a lot for this!

megies:
Without having actually tried it, this looks good indeed. I have a few minor remarks:
 - To be consistent with the rest of the code base, could you please replace `"""` comments (e.g. https://github.com/obspy/obspy/pull/1182/files#diff-6fe65029cbce7af6392c081b3b988b1fR384) in the code by lines starting with `# `?
 - Right now this is placed as `obspy/io/sc3ml/core.py`. What about making this `obspy/io/seiscomp/sc3ml.py`? Right now it wouldn't make much difference.. but if anybody addes more seiscomp related I/O stuff in the future we have a common folder for it and don't need another one in parallel. Opinions?

Jollyfant:
Updated the comments so that should be passing soon. I guess we can merge it in to a separate io seiscomp module down the line if more stuff is rolled out but for now this seems to be OK if you ask me.

krischer:
If you plan on renaming this module later down the road, please rename it now. It is always a huge hassle to rename things if people are already using it.

Jollyfant:
Ok, let's do it then :+1:

megies:
> Ok, let's do it then :+1:

Thanks, and sorry about the hassle, but we really need to get things right with the first shot, otherwise we end up with more deprecation/reroute pain later.. ;-) And I think this hierarchy makes sense even if it turns out that nothing gets added later in addition..

Jollyfant:
Uh, something went wrong with the Travis build. Looks like a timeout or something? Is it possible for someone to retry the build without pushing a new commit or pull request.

megies:
> Is it possible for someone to retry the build without pushing a new commit or pull request.

Yes that's possible for us, and I guess somebody already restarted the build in the meantime..

krischer:
Great! Thanks for this :-)

megies:
> Great! Thanks for this :-)

:+1: nicely done!

PR #1192, opened 2015-11-25
Plot Event far-field radiation pattern / refactor core/event.py into folder structure

MMesch:
This addition to obspy allows to compute and plot P/S-wave farfield radiation patterns for (hopefully) arbitray moment tensors. You can run the test or try:

```python
from obspy.imaging.radpattern import plot_3drpattern
mt = [0.91, -0.89, -0.02, 1.78, -1.55, 0.47]
plot_3drpattern(mt,kind='p_sphere')
```

it is neither well tested nor optimized. Some improvements could be:

* add colour and other things to quiver plots
* making sure that axes are well aligned
* plotting beachball projections on the sides of the matplotlib 3d plot
* compute radiation pattern for different a set of taup phases and stations
* add nearfield radiation
* add single force

p_radpattern

p_quiver


megies:
Looks nice!

> compute radiation pattern for different a set of taup phases and stations

That could be very nice for many people, I think..

claudiodsf:
That's very nice!
It only seems strange to me to have `farfield_p` and `farfield_s` functions in `obspy.imaging`.
Maybe it's time for an `obspy.source` package?

MMesch:
Agree. Things that could/should be improved are then:

* better function names and organisation. Inside the imaging package? Inside a class?
* add colour and other things to quiver plots. This requires non-standard mlab3d. Adapting the vector length to the radiation pattern would be nice as well. So far it can only take a direction.
* making sure that axes are well aligned with other moment tensor conventions
* plotting beachball projections on the sides of the matplotlib 3d plot. Should be possible but bugs (see commented lines in the p_sphere section of the code). Keypoint is to use sth. like art3d.patch_collection_2d_to_3d(bball). Does anyone know how this works?
* compute radiation pattern for a set of taup phases and stations. I guess this could be combined with a method to plot all of it on a map.
* add nearfield radiation pattern. Possible but more complicated than farfield.
* add single force. Should be straight forward for the farfield.
* tetrahedral spherical grid to reduce plotting time. Does anyone know how to get this in mlab3d?

krischer:
I'm playing around with the quiver plots - not perfect yet but its getting better :-)

pattern

megies:
> It only seems strange to me to have farfield_p and farfield_s functions in obspy.imaging.

I agree. But the place to go is already there: `obspy/core/event.py`

krischer:
Hmm...I did not realize that matplotlibs 3D plotting lacks basic features like figuring out what should be drawn in the front if one plots two objects...I don't think its possible to make a good 3D quiver plot with matplotlib. Even the "nicer" version of me is almost impossible to interpret because the orthographic projection messes with the mind and is just very confusing.

@MMesch I would be fine to have your original mayavi plots in there as well. Just make sure to import mayavi within the plotting functions so it does not become a hard ObsPy dependency.

Other opinions on this?

megies:
> Just make sure to import mayavi within the plotting functions so it does not become a hard ObsPy dependency.
> 
> Other opinions on this?

Yeah, why not. If it's optional visualizations of more special stuff and done as soft dependency.. :+1:

MMesch:
OK. If someone finds out how to plot the beachball patch_collection in the 3D plot this would be great. I didn't get very far but it should be possible (see http://stackoverflow.com/questions/18228966/how-can-matplotlib-2d-patches-be-transformed-to-3d-with-arbitrary-normals )

claudiodsf:
@megies 
But `obspy/core/event.py` contains: `Classes for handling event metadata`

Here we have a possible new class of functions for seismic source representation, which seems something different to me.

MMesch:
p_radpattern_beach

I managed to plot the beachball together with the surface plot but matplotlib is just not capable of really doing 3D. It can't tell which patch is above the other for the beachball (therefore facecolor='none') and it cannot say when the beachball is behind the surface. 3D plotting in matplotlib is too limited and I'm giving up on this...

QuLogic:
It doesn't look like any released version of Mayavi2 supports Python 3 :frowning: .

FYI @carltape.

carltape:
Cool stuff in this thread! I've got some cool stuff in the link below, but most of it is mathematica (which very few seismologists use). I agree, it would be great to have some open-source alternatives.
http://www.giseis.alaska.edu/input/carl/research/beachball_gallery.html

c546824:
Here's a nice spherical harmonics example with Mayavi, this may be what @QuLogic was referring to
http://docs.enthought.com/mayavi/mayavi/auto/example_spherical_harmonics.html#example-spherical-harmonics

megies:
> @megies
> But obspy/core/event.py contains: Classes for handling event metadata
> 
> Here we have a possible new class of functions for seismic source representation, which seems something different to me.

It's just that we're having a lot of pain currently (and I mean really *a lot*), cleaning up and sanitizing the module structure (which has exploded over the last years), esp. to clean up the top levels of our submodule hierarchy (see #999, #1039, ...).
So we should be very careful not to open Pandora's box again. Any new proposals for top-level submodules immediately ring three or four alarm bells in my head. In my opinion, something concerned with seismic sources belongs somewhere close to our event related stuff. The `Classes for handling event metadata` in `event.py` is just some info stub somebody put there when we started that file (with QuakeML support in mind) and can be adapted. I agree that `event.py` is a bit stuffed, but we can easily make it into a subfolder and separate things in there logically, that wouldn't break user code.

MMesch:
Here are plots with mayavi that I did before. Another alternative would be to just output a vtk file if there is no mayavi/vtk and plot otherwise.

http://pythology.blogspot.fr/2015/11/earthquake-moment-tensor-radiation.html

pcmt

MMesch:
scmt

megies:
Looks great! :+1:

ThomasLecocq:
:+1: amazing look in 3D vtk (mayavi or paraview, whatever) !!

MMesch:
@megies @claudiodsf  I open an event folder now in obspy.core that combines the event.py file and radpattern.py . Is this what you were thinking? I think the radiation pattern functions should somehow be grouped. I am not sure whether a submodule or a class is best. Any thoughts?

megies:
@MMesch, yes, that's what I think makes most sense, to group these event/origin/source related functions close to where all the event stuff is defined. Since up to now it was a plain py file, making it a folder and having everything in there feels natural to me.
But I'm open for any reasoning that suggests something else, of course.

What do you think about renaming the old `event.py` which is now `event/event.py` to `event/base.py`?

> I think the radiation pattern functions should somehow be grouped. I am not sure whether a submodule or a class is best.

I think having them grouped in that submodule `event/radpattern.py` is fine. The pure plotting functions should in principle be put back into `imaging`..? Also, I'd propose adding methods to `FocalMechanism`/`MomentTensor` and `Event` to plot/calculate the radiation pattern -- as a high-level, preferred usage option for users. Most users should not need to lookup those functions and use those methods instead. I think a map plot would also be nice, but that probably would not work well for very short distances with taup?

MMesch:
There is direct vtk file output now. Renamed event.py -> base.py . Still need to move the plotting routines, and add a taup plotting routine. Normalization of the vector field needs to be checked as well.

MMesch:
obspy_radpattern

vtk output for paraview is implemented. I am really having trouble installing mayavi with anaconda. It wants to downgrade numpy which leads to some errors in matplotlib afterwards. Did anyone succeed to install the newest numpy matplotlib and mayavi with anaconda?

MMesch:
new_plot

This is the new plot. I hope the orientation is correct. It seems to be OK if the beachball is the projection of the lower hemisphere on a plan seen from above (which inverts left and right if seen from below in the 3d view). Please tell me if something looks weird :)

MMesch:
new_plot

I'm done for the moment. We could add some kind of map in this plot if latlon keyword is provided. Connecting it to the event object is then very quick and simple.

I thought about the maps: it seems that a function to plot 2d greatcircle paths between stations and one or several sources is still missing in obspy. I think we should make it a separate branch/issue to add a function that gives 2d greatcircle paths and 3d paths using taup. Once we have such a function, it is simple to combine it with the event radiation pattern and also plotting routines in 2D and 3D.

I therefore suggest to finish/test/include the radiation pattern functions as they are right now and to look after the taup map in the next step.

Matthias

MMesch:
Mesh orientation switches now automatically between null and other axes

new_plot

MMesch:
figure_2
figure_1

And yet more options. Pressure stretches and tension squeezes the sphere.

MMesch:
I have pulled the Event and Catalog classes out of base.py and attached a plot function to the Event class. Not sure about the best way to better organize this huge class hierarchy @megies @krischer but base.py is really too big and the class hierarchy too deep in my opinion. If it is as deep as it is now (I guess adapted to quakeml) then all classes that are unlikely to be changed should be hidden in one file and the others separate from them.

You plot the event data now with:

    from obspy import read_events
    cat = read_events('.../obspy/obspy/io/quakeml/tests/data/quakeml_1.2_focalmechanism.xml')
    ev = cat[0]
    ev.plot()

QuLogic:
WRT matplotlib's 3D plots, see matplotlib/matplotlib#5830 and matplotlib/matplotlib#4090.

megies:
Sorry, didn't get a chance to review the latest version yet, will try to do so in the next days..

obspy/core/event/radpattern.py
```diff
@@ -0,0 +1,549 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------
+# Filename: radpattern.py
+#  Purpose: Computes and plots radiation patterns
+# ---------------------------------------------------------------------
```
krischer: Can you please remove the 4 preceding lines? Any documentation should go in the file's main docstring.

obspy/core/trace.py
```diff
@@ -2597,7 +2597,7 @@ def remove_response(self, output="VEL", water_level=60, pre_filt=None,
             ax6 = fig.add_subplot(326, sharex=ax4)
             for ax_ in (ax1, ax2, ax3, ax4, ax5, ax6):
                 ax_.grid(zorder=-10)
-            ax1.text(0.05, 0.1, 'pre_filt: %s' % pre_filt,
+            ax1.text(0.05, 0.1, 'pre_filt: %s' % str(pre_filt),
```
krischer: This is also done in #1219 and whichever gets merged last will thus require a rebase. Just something to keep in mind.

krischer:
Looks pretty good to me. We can merge this as soon as some actual tests for the plots are added. Can you do this - look at `obspy.imaging` how plots are tested. Otherwise you can give me access to your ObsPy fork and I'll do it.

I think we should merge this ASAP as it contains some renames which we want to merge before the next release as that release already contains a lot of renames.

MMesch:
mayavi plotting is still untested in this obspy version because I couldn't install it under anaconda...

megies:
> because I couldn't install it under anaconda...

Oh, really? On my Linux mayavi installs in anaconda with `conda install mayavi` like a charm..

megies:
This looks really nice already, just a couple of minor things that need to be looked at, I think

Some more minor changes that I would consider:
 - right now this only works if a moment tensor is present, would be nice to also show the plot when only a focal mechanism is present
 - maybe add an option to plot S radiation instead of P (e.g. `def plot(self, type="P")`, or even both (ok, I just saw that `both_sphere` is not implemented yet -- you should change the default value of `kind` option in `plot_3drpattern` to something that's implemented.. ;-)), or see comment add end of post..
 - mayavi plotting is broken (`normp` variable not defined -- same goes for `both_quiver`)
 - How about doing the plots for `both_quiver` side-by-side as subplots, instead of having two figures?
 - maybe add `focal_mechanism_id=None` option to `Event.plot()`, so that non-preferred focal mechanisms can be plotted

Is everybody happy with naming for `radpattern.py`? Should we consider a less abbreviated/concatenated naming (e.g. `radiation_pattern.py`)? Sorry, don't want to be nit-picky here, it's just because we are currently going through a lot of pain with enforcing fixed naming conventions right now, so we should think about those things at least now.. ;-)

One thing that just came to mind about the Event plots.. since it seems like there's a variety of individual plot types to choose from.. maybe it would be nice to be able to specify exactly which things to plot in a kind of subplot grid specification..?
E.g.:
```python
event.plot([["ortho", "beachball"], ["p_sphere", "s_quiver"]])
```
grid

megies:
Oh and btw, you can use some magic to read any file within `tests/data` directories..:
```python
cat = read_events('/path/to/quakeml_1.2_focalmechanism.xml')
```

MMesch:
@megies : if I do conda install mayavi, I get:

> The following NEW packages will be INSTALLED:
> 
>     mayavi: 4.4.0-np19py27_0
> 
> The following packages will be DOWNGRADED:
> 
>     numpy:  1.10.2-py27_0 --> 1.9.3-py27_0    
> 
> Proceed ([y]/n)? y

If I try to use obspy then, I get:

> from matplotlib._path import (affine_transform, count_bboxes_overlapping_bbox,
> ImportError: numpy.core.multiarray failed to import

If I reupdate numpy I get:

>  File "[...]/lib/python2.7/site-packages/pyface/qt/__init__.py", line 17, in prepare_pyqt4
>     sip.setapi('QString', 2)
> ValueError: API 'QString' has already been set to version 1

which means that there is a true incompatibility between numpy/matplotlib and mayavi. If anyone has a workaround I would be very happy!

krischer:
That is where conda gets messy unfortunately. Same with Python as whole as only one version of a dependency can be installed at any time :-(

On my machine I can install `mayavi` and then reupdate `numpy` and it works fine. Try updating `pyqt` - maybe that does the trick.

megies:
I usually found it's a good idea to have all those packages in the `update` call, e.g. `conda install mayavi numpy matplotlib`.

MMesch:
new_plot

Voila, new plotting option as @megies suggested and everything is better organised now. You can plot this by giving a list of ['s_sphere', 'p_sphere', etc...] A map of the position would be nice

I have the most recent version of pyqt. I fixed some things in the mayavi plotting option but I can't test it.
> pyqt                      4.11.4                   py27_1

MMesch:
conda install mayavi numpy matplotlib gives the same error. It also wants to downgrade matplotlib then which is different then before. Anyway, maybe something isn't setup correctly on my machine...

megies:
Nice!

> A map of the position would be nice

In principle we should be able to reuse `Catalog.plot()` but I just saw that it can only be provided a `Figure`, not an `Axes` instance.. would need a bit of remodeling I guess..

megies:
Also, one minor issue: The beachball plot gets stretched if the scale is not equal..

obspy/core/trace.py
QuLogic: Looks like it'll be this one.

QuLogic:
> conda install mayavi numpy matplotlib gives the same error. It also wants to downgrade matplotlib then which is different then before. Anyway, maybe something isn't setup correctly on my machine...

That's expected given your previous results. Conda is great for installing the latest, but it's dependency resolution is not so great if you want to install something not so new. Mayavi is apparently outdated, so it pulls in older NumPy. This _should_ (theoretically, not in actual implementation) also downgrade everything that depends on NumPy, but it doesn't.

The absolute easiest way around it is to simply create a new environment and specify all the dependencies at the same time: `conda create -n obspy27 python=2.7 numpy matplotlib scipy basemap future flake8 lxml decorator sqlalchemy mock nose gdal pyproj docopt coverage requests  mayavi` (this is all the dev requirements) and it should depsolve it correctly.

MMesch:
I found the solution (https://mail.scipy.org/pipermail/ipython-user/2012-February/009478.html)
you have to do:

export QT_API=pyqt

then it works like a charm with the downgraded numpy and matplotlib.

Mayavi plotting works now!

MMesch:
@krischer @megies what is the best place for the event.plot() function? Right now I put it into the event class, but maybe it is better to attach it directly to the moment tensor class? I can't really find where this would be....

MMesch:
update:
1): I think including a map is not useful right now because the radiation pattern depends only on the moment tensor so far. map would requires location as well. In the long term, one could think about designing a plot for one event that provides all information about it in a compact graphical way. 

2) the moment tensor has to be given in RTP convention right now. I realised that moment tensor conversion is a very common operation that should be done in a central function.

Otherwise for my part I think this plotting routine is good to go, except that the doc needs to be cleaned once and for all.

megies:
> @krischer @megies what is the best place for the event.plot() function? Right now I put it into the event class, but maybe it is better to attach it directly to the moment tensor class? I can't really find where this would be....



> 2) the moment tensor has to be given in RTP convention right now. I realised that moment tensor conversion is a very common operation that should be done in a central function.

Following QuakeML I guess we expect all moment tensors to be in RTP convention (http://docs.obspy.org/packages/autogen/obspy.core.event.Tensor.html, also see QuakeML manual)). Any conversions should be handled by the Tensor class itself, I guess (e.g. `Tensor.to_coordinate_system()` or `MomentTensor.to_coordinate_system()`).

> Otherwise for my part I think this plotting routine is good to go, except that the doc needs to be cleaned once and for all.

I'm gonna go over it once again, quite some things to be addressed, I think. I'll open a PR to your branch..

MMesch:
OK, I'll stop changing it for now then. I think it's better if one of you guys go through it because you know way better how to make it ready to be included ...

megies:
@MMesch if you don't mind and add me to your fork, I could directly push to this PR..

megies:
OK, did a lot of massive refactoring in this PR today, think it's getting there..

I massively refactored the whole new `core/event` folder into smaller logical pieces. Also moved most of the radiation pattern stuff into `imaging` because it was strictly plotting related. The rest I moved to the new `event/source.py` which holds focal mechanism and tensor related classes.

We should still plugin the plot routine as methods on `FocalMechanism` and `MomentTensor` and also make it possible for strike/dip/rake-only focal mechanisms.

Maybe needs a bit more review and refactoring still..

megies:
Needs a rebase, then addition of map plot using #1240 and good to merge then, I think.

megies:
Rebased + force-pushed

megies:
I think this is getting ready to merge now.. here's the default event plot (which subplots are done can be customized..).

Complaints on how core/event.py was split up in single files in core/event/ now or never. ;-)

figure_1

MMesch:
looks great! If possible: can we somehow shrink the 2d axes to about 80% to give them the same size as the 3d axes?

megies:
Setting different Axes size would need quite some work, I guess, as we couldn't use good old plain `plt.add_subplot()` anymore. But it's simple to reset axes limits, which works for all map plots but the `local` one (which is pretty useless with only one event anyway, I think.

Do you think we should also shrink for the `global` option? Probably not, or not so strongly because it has more white space at the corners due to it's oval shape..

megies:
Is it only me or do the 3d axes have some ghosts of the old axes ticklabels in the background..?

screenshot from 2016-02-15 13 50 18

MMesch:
size is perfect like this. But the ghosts are a case for the matplotlib bug tracker :) I guess we can fix them nevertheless.

megies:
In some cases the `global` plot seems to get squished to the side a bit.. no idea where it's coming from (the subplot setup without doing the actual plotting looks good..).

But it's only a minor issue, can be solved at some later point, if it's even necessary at all to do something about it..

screenshot from 2016-02-15 14 10 38
screenshot from 2016-02-15 14 08 50

MMesch:
I think the empty left space doesn't matter. Did you find out about the stuff behind the 3d axes?

megies:
> I think the empty left space doesn't matter. Did you find out about the stuff behind the 3d axes?

Nah, don't have time to go looking now. We need to get the RC1 ready.. If need be it can be fixed as a bug fix later..

obspy/core/event/base.py
```diff
@@ -0,0 +1,1143 @@
...
+    is cumbersome, error-prone and hard to do consistently. The classes created
+    with this method will inherit from :class:`~obspy.core.util.AttribDict`.
+
+    Usage to create a new class type:
+
+    The created class will assure that any given (key, type) attribute pairs
+    will always be of the given type and will attempt to convert any given
+    value to the correct type and raise an error otherwise. This happens to
+    values given during initialization as well as values set when the object
+    has already been created. A useful type are Enums if you want to restrict
```
QuLogic: s/are Enums/is Enum/

obspy/core/event/event.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        Plot event location and/or the preferred focal mechanism
+        and radiation pattern.
+
+        :type kind: list of str or nested list of str
+        :param kind: can be one of the following options:
+            * A list of strings (for a 1-row plot) or a nested list
+              of strings (one list of strings per row), with the
+              following keywords to generate a matplotlib figure:
+              'orth' (Orthographic plot of event location,
```
QuLogic: Should be 'ortho'.

obspy/core/event/event.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def plot(self, kind=[['ortho', 'beachball'], ['p_sphere', 's_sphere']],
+             subplot_size=4.0, show=True, outfile=None, **kwargs):
+        """
+        Plot event location and/or the preferred focal mechanism
+        and radiation pattern.
+
+        :type kind: list of str or nested list of str
+        :param kind: can be one of the following options:
+            * A list of strings (for a 1-row plot) or a nested list
```
QuLogic: I'm not sure why the bullet list is on this line instead of on the string options below.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+                nrows, ncols_, i * ncols_ + j + 1, **kwargs))
+    return fig, axes, kind_
+
+
+def plot_radiation_pattern(
+        mt, kind=['p_sphere', 'beachball'], coordinate_system='RTP',
+        p_sphere_direction='inwards', fig=None, show=True):
+    """
+    Plots the P/S farfield radiation pattern on a unit sphere grid
+    calculations are based on [Aki1980]_ eq. 4.29.
```
QuLogic: Should be a separate sentence.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+
+
```
QuLogic: Extra new line.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+        six independent components of the moment tensor, where the coordinate
+        system is 1,2,3 = Up,South,East which equals r,theta,phi -
+        Harvard/Global CMT convention). The relation to [Aki1980]_
+        x,y,z equals North,East,Down convention is as follows: Mrr=Mzz,
+        Mtt=Mxx, Mpp=Myy, Mrt=Mxz, Mrp=-Myz, Mtp=-Mxy.
+
+    :param kind: One of the following three options:
+        * A list of strings or nested list of strings for a matplotlib plot
+          (for details see :meth:`obspy.core.event.event.Event.plot`)
+        * 'mayavi': uses the mayavi library (not yet available under python 3
```
QuLogic: s/python/Python/

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+          and problematic with anaconda)
+        * 'vtk': This vtk option writes three vtk files to the current working
+          directory. rpattern.vtk contains the p and s wave farfield vector
+          field beachlines.vtk contains the nodal lines of the radiation
+          pattern rpattern.pvsm is a state file that sets paraview parameters
```
QuLogic: s/paraview/ParaView/

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+          to plot rpattern.vtk and beachlines.vtk
+
+    :param coordinate_system: the only implemented option so far is 'RTP'.
+                              Should be extended to support NED, USE, DSE, NED
```
QuLogic: Isn't RTP the same as USE? Why is this name used here, but not anywhere else (e.g., in MoPaD)?

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+                              in the future
+    :type fig: :class:`matplotlib.figure.Figure`
+    :param fig: Figure instance to use.
+    :type show: bool
+    :param show: Whether to show the figure after plotting or not. Can be
+        used to do further customization of the plot before showing it.
+
+    :return: 3D grid point array with shape [3,npts] that contains
```
QuLogic: Shape should be a tuple. Also, space after comma.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+             the sperical grid points
+
+             3D vector array with shape [3,npts] that contains the
```
QuLogic: Shape should be a tuple; space after comma.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+
+    else:
+        raise NotImplementedError('{:s} not implemented yet'.format(kind))
+
+
+def _plot_radiation_pattern_sphere(
+        ax3d, ned_mt, type, p_sphere_direction='inwards'):
+    """
+    private function that plots a radiation pattern sphere into
+    ax3d
```
QuLogic: Don't know why this word is on its own line. (There should be a blank line before the parameters, though.)

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+    :param ax3d: matplotlib 3d ax object
+    :param ned_mt: moment tensor in NED convention
+    :param p_sphere_direction: if this is 'inwards', the tension regions
+                               of the beachball deform the radiation sphere
```
QuLogic: Indented too much.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+             title='{} wave farfield'.format(type))
+    ax3d.view_init(elev=-110., azim=0.)
+
+
+def _plot_radiation_pattern_quiver(ax3d, ned_mt, type):
+    """
+    private routine that plots the wave farfield into the
+    input ax object
+
+    :param ax3d: a matplotlib ax with 3d projection activated
```
QuLogic: s/ax/Axes/; do we cross-link matplotlib's docs?

obspy/core/event/event.py
megies: Yeah, copy/paste/refactor mistake.. pulled this option out of another docstring..

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+             zticklabels=['Up', 'Down'],
+             title='{} wave farfield'.format(type))
+    ax3d.view_init(elev=-110., azim=0.)
+
+
+def _plot_beachball(ax2d, rtp_mt):
+    """
+    private function that plots a beachball into a 2d matplotlib ax
+
+    :param ax2d: 2d matplotlib ax
```
QuLogic: s/ax/Axes/

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+    ax2d.set(xlim=(-50, 50), ylim=(-50, 50),
+             xticks=(-40, 40), yticks=(-40, 40),
+             xticklabels=('West', 'East'),
+             yticklabels=('South', 'North'),
+             title='lower hemisphere stereographical projection')
+
+
+def _plot_radiation_pattern_mayavi(ned_mt):
+    """
+    This function uses the mayavi (vtk) library to plot the radiation
```
QuLogic: Needs a short one-line description: Plot the radiation pattern using MayaVi.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+    variable to e.g. export QT_API=pyqt that mayavi works properly.
+
+    :param ned_mt: moment tensor in NED convention
+    """
+    # use mayavi if possible.
+    try:
+        from mayavi import mlab
+    except Exception as err:
+        print(err)
+        msg = "obspy failed to import mayavi. " +\
```
QuLogic: s/obspy/ObsPy/ s/mayavi/MayaVi/

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+              "You need to install the mayavi module " +\
+              "(e.g. conda install mayavi, pip install mayavi). " +\
+              "If it is installed and still doesn't work, " +\
+              "try setting the environmental variable QT_API to " +\
+              "pyqt (e.g. export QT_API=pyqt) before running the " +\
+              "code. Another option is to avoid mayavi and " +\
+              "directly use kind='vtk' for vtk file output of the " +\
+              "radiation pattern that can be used by external " +\
+              "software like paraview"
```
QuLogic: s/paraview/ParaView./

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+    points = spherical_grid()
+    ndim, npoints = points.shape
+    dispp = farfield(ned_mt, points, type="P")
+    disps = farfield(ned_mt, points, type="S")
+
+    # write vector field
+    with open(fname_rpattern, 'w') as vtk_file:
+        vtk_header = '# vtk DataFile Version 2.0\n' + \
+                     'radiation pattern vector field\n' + \
+                     'ASCII\n' + \
```
QuLogic: BINARY would save some space, but not important just now.

obspy/imaging/source.py
```diff
@@ -0,0 +1,593 @@
...
+    # ravel point array and rotate them to the null axis
+    points = np.vstack((x.flatten(), y.flatten(), z.flatten()))
+    points = np.dot(rotmtx, points)
+    return points
+
+
+def _equalarea_spherical_grid(nlat=30):
+    """
+    generates a simple spherical equalarea grid that adjust the
+    number of longitude samples to the longitude. This grid is useful
```
QuLogic: to the latitude, right?

obspy/imaging/source.py
megies: Yeah, well.. actually the whole description is wrong (could be it was at some point, dunno). Only a figure returned, in some cases `None`. I'll push those cosmetic changes after Travis (finally...) finishes.

obspy/imaging/source.py
megies: Yes we do, I'll try to fix it everywhere..

obspy/imaging/source.py
megies: Yeah.. if somebody feels like it, can be done later..

obspy/imaging/source.py
megies: Yeah.. would make more sense.. I'll change it.

obspy/imaging/source.py
megies: Note sure.. @MMesch?

megies:
Travis looks good mostly: https://travis-ci.org/obspy/obspy/builds/109357659
I think those fails (e.g. https://travis-ci.org/obspy/obspy/jobs/109357662#L688) are related to #1156, and others. (The active figure get closed after every image test, but sometimes it's already closed within the image test -- I suspect that this might cause these problems..)

megies:
Rebased onto #1250 (a9075cc23b891afb406), hope this gets rid of stray Travis fails.

megies:
OK, green light by Travis, finally..

Since nobody objected to the core/event/ structure in here I'm gonna merge now.

krischer:
:tada:

claudiodsf:
:+1:

MMesch:
yey!

QuLogic:
@MMesch I think one comment regarding the usage of RTP vs. USE was not addressed; can you please follow up with a correction in another PR?

obspy/imaging/source.py
```diff
@@ -0,0 +1,595 @@
...
+        * 'mayavi': uses the mayavi library (not yet available under Python 3
+          and problematic with anaconda)
+        * 'vtk': This vtk option writes three vtk files to the current working
+          directory. rpattern.vtk contains the p and s wave farfield vector
+          field beachlines.vtk contains the nodal lines of the radiation
+          pattern rpattern.pvsm is a state file that sets ParaView parameters
+          to plot rpattern.vtk and beachlines.vtk
+
+    :param coordinate_system: the only implemented option so far is 'RTP'.
+                              Should be extended to support NED, USE, DSE, NED
```
QuLogic: I'm re-adding this comment so that @MMesch can address it:
> Isn't RTP the same as USE? Why is this name used here, but not anywhere else (e.g., in MoPaD)?

MMesch: yes good point. Technically they are the same, but I think I added USE for regional models (cartesian definition), just to be complete. If you prefer we can remove it!

QuLogic: It'd be preferable to use consistent naming between this and MoPaD, which doesn't use RTP at all. I'm fine with adding aliases; I can never remember which one is which anyway.

The above statement is incorrect though, since USE _is_ implemented as it's the same as RTP?

MMesch:
How can I open a good pull request for this that includes the stuff that @megies merged? I can go through the doc once more then...

QuLogic:
This branch is already merged to `master`; create a new branch starting from the latest `master`. There's no need to use the existing `radpattern` branch when creating a new PR.

MMesch:
OK!

krischer:
I just discovered this here which also has some very fancy plots and is based on this very PR:

https://github.com/FMassin/SeismicSource/blob/master/README.ipynb

Any chance for also getting those into main ObsPy?

@FMassin, @MMesch 


screen shot 2016-04-15 at 21 26 12

MMesch:
looks great! I think adapting the obspy plot to look similar would be great. Essentially this means different colors and a smaller surface representation of the radiation pattern? Then all of them should be put into one plot. Also we can benchmark the radiation pattern values with this independent code!

krischer:
Also maybe center the colormap around 0?

PR #1194, opened 2015-11-27
mini bug?

MMesch:
Hi, this line raised an Exception in python 2.7 on 2 computers.


megies:
Oh, that's right.. that's because `pre_filt` happens to be a tuple.. thanks!

PR #1010, opened 2015-04-20
Add FDSN web service shortcuts

petrrr:
I would like to propose to add shortcuts for all nodes of the EIDA federation:

 * `ETH`, Switzerland, http://eida.ethz.ch/	
 * `NEIP`, Romania, http://eida-sc3.infp.ro/
    (*not tested yet, service probably up and running*)
 * `LMU`, Germany (BayernNetz), http://erde.geophysik.uni-muenchen.de/
 * `IPGP`/Geoscope, France (volcanological observatories) + Global (GEOSCOPE),  http://eida.ipgp.fr/
    (*not tested yet*)

This is a pretty trivial change, I just would like to discuss if this change should go only to master (enhancement), or if it is sufficiently trivial/useful to add it to release as well.

Maybe `ODC` would be the more appropriate abbreviation for the node at Orfeus. (http://www.orfeus-eu.org)

PR #1162, opened 2015-11-30
Catch invalid URL when initialising FDSN Client

petrrr:
I just ran into a small problem (my error by the way) which surprise me and kept me occupied a bit. I was using the FDSN client, but misspelled one abbreviation. More precisely I misspelled it when filing #1010 and 3fa60b82 some time back. Now I myself hit the bug. 

However, I would like to propose that invalid URL and not existing abbreviations respectively are caught and reported explicitly back to the user.

Here is what currently happens:

```python
c = obspy.fdsn.Client("NIEP", debug=False)

FDSNException                             Traceback (most recent call last)
<ipython-input-23-c88aacea7384> in <module>()
----> 1 c = obspy.fdsn.Client("NIEP", debug=False)

/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/obspy/fdsn/client.py in __init__(self, base_url, major_versions, user, password, user_agent, debug, timeout, service_mappings)
    150             print("Request Headers: %s" % str(self.request_headers))
    151 
--> 152         self._discover_services()
    153 
    154     def get_events(self, starttime=None, endtime=None, minlatitude=None,

/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/obspy/fdsn/client.py in _discover_services(self)
   1318                    "be due to a temporary service outage or an invalid FDSN "
   1319                    "service address." % self.base_url)
-> 1320             raise FDSNException(msg)
   1321 
   1322         # Cache.

FDSNException: No FDSN services could be discovered at 'NIEP'. This could be due to a temporary service outage or an invalid FDSN service address.
``` 

Or with debugging enabled:
```python
Base URL: NIEP
Request Headers: {'User-Agent': 'ObsPy 0.10.2 (Darwin-14.5.0-x86_64-i386-64bit, Python 3.4.3)'}
Downloading NIEP/fdsnws/dataselect/1/application.wadl
Error while downloading: NIEP/fdsnws/dataselect/1/application.wadl
Downloading NIEP/fdsnws/event/1/application.wadl
Error while downloading: NIEP/fdsnws/event/1/application.wadl
Downloading NIEP/fdsnws/station/1/application.wadl
Error while downloading: NIEP/fdsnws/station/1/application.wadl
Downloading NIEP/fdsnws/event/1/catalogs
Error while downloading: NIEP/fdsnws/event/1/catalogs
Downloading NIEP/fdsnws/event/1/contributors
Error while downloading: NIEP/fdsnws/event/1/contributors
FDSNException                             Traceback (most recent call last)
[... more output as above]
```


petrrr:
Here my first draft on this. 
Q: 
 * should I use `re.compile()` instead? Or is this optimization irrelevant in this case?
 * Is the `ValueError` exception appropriate? Should I use `FDSNException` or anything else?
 * Do we need tests for this?

petrrr:
Ahhh ... 
I probably should not assume that abbreviations are always given in uppercase, i.e. remove the  ` base_url.isupper()` condition.

obspy/clients/fdsn/client.py
```diff
@@ -45,6 +46,7 @@
 
 DEFAULT_SERVICE_VERSIONS = {'dataselect': 1, 'station': 1, 'event': 1}
 
+URL_REGEX = 'https?://(localhost|\w+(\.\w+){2,})(:\d+)?(/\w+)*/?$'
```
QuLogic: Schemes are technically case-insensitive, `\w` is `[a-zA-Z0-9_]` (plus other Unicode "word characters") but paths can include more than that, I think. Maybe it would be simpler to pass through `urlparse`?

krischer: This stackoverflow post shows the regex Django uses for that purpose - best just use that: http://stackoverflow.com/questions/7160737/python-how-to-validate-a-url-in-python-malformed-or-not

Passing it through `urlparse` would also work but it would require a couple of additional asserts afterwards so I guess the regex way is easier.

petrrr:
@QuLogic: I gone a bit through the `urlparse` docs, but it does not seem that a one-liner solve this. On the other hand, this is just to give a valid feedback to users. So I agree with @krischer that in this context an "approximated" regex based plausibility check should be more then sufficient. 

Thanks, for the link. From what I've seen quickly my regex should be mostly equivalent, but not matching IPv4 is a relevant omission. (BTW: what is about IPv6, is it still that irrelevant?). Will update it.

petrrr:
Additional notes:
 * I intentionally required domains with >= 3 levels. Assuming that **very probably** no one will ever put FDSN services on the highers level of their organization, so indicating some omission. This makes domain `http://example.com` fail but I'd consider changing the test to `http://ws.example.com` if you agree.
 * Adhering strictly to the FDSN specs, one would not allow a path part. But I believe it is practical for to accept it mainly for local installations and pre-production testing.
 * `localhost`, IP and port specification will be useful for testing and port forwarding as well.
Have I missed anything important?

petrrr:
The link provided by @krischer proposed the following regex for the domain part:

    (?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)

I think I understand the first group (highest level), I also get of the first option of the second group `[A-Z]{2,6}` (top-level-domain) up to the optional "dot" `\.?`. However, I do not get two details:
 * the second option `[A-Z0-9-]{2,}` is a super-set of the 1st one, matches close to anything. So why two options? I also do not get the intention of the second option at all. Don't I want to conclude with a top level domain?
 * Why there is a optional trailing "dot" in the domain? Would this match something like: `example.com.`?

Sorry for my ignorance and thanks for illumination :bulb: !

barsch:
the URL validator in the most recent django version is even more sophisticated: https://github.com/django/django/blob/master/django/core/validators.py#L78 ...

petrrr:
Here my new proposal, as mentioned before I do not want to be overly sophisticated and it is not a "validator" neither.

What the regex does:
 * verifies if it starts with `http://` or `https://`. The second is currently not specified by the standard, but I expect it to come pretty soon.
 * expect three level domain (host.org-domain.tld) **or** localhost **or** IPv4 (see above);
 * I am conservative: no *generic TLDs*, max. 6 alpha characters;
 * optional port specifier (max. 5 digits);
 * pretty *well-formatted* path part, the FDSN specs would not allow them at all (see above)

What it does **not**:
 * Provide a IPv6 pattern; It's probably still not used much and I do not feel sufficiently knowledgeable to come up with a decent one.
 * Allow *generic TLDs*, nobody needs these and so I prefer a better constraint check.
 * It allows for `_` in any position of  the domain part. For the sake of simplicity I found this approximation acceptable, excluding would complicate the pattern somewhat. 

Let me know what you think?

krischer:
Please also allow for IPv6 urls. Does not need to be an extensive check which apparently is quite hard and results in very long regular expressions but just something that at least allows IPv6 patterns.

This is taken from the django url checker and looks quite reasonable: `r'\[[0-9a-f:\.]+\]'`

It will obviously have false positives but that is ok here.

megies:
> Please also allow for IPv6 urls.

:+1: The original intent was just to avoid e.g. a misspelled short URL being interpreted as a normal URL..

krischer:
Ping. Can you add the IPv6 pattern as well? Then we could merge this.

megies:
@petrrr.. ping?

petrrr:
Sorry, but I had to take time off from open source for some time.

petrrr:
Okay, finally I found some time to play around with the IPv6. I hope the one I came up with is a decent compromise between (1) doing some meaningful test and (2) complexity.

Now:

1. Test: Do we need tests for this? For what and how would we test here? I did some offline tests to see if the pattern works, but do not see how these could be turned immediately into unit tests.
2. Code location: Is the location where I put the pattern okay, or should it be moved elsewhere. for Example: `clients/fdsn/header.py` ???

I guess, I need to add a changelog entry. Any tips how I should proceed to avoid conflicts? Rebase first, then edit, push, done?

megies:
Probably best wait for merge of #1039 (next couple days), then rebase on master..

Pattern location is OK.

megies:
@petrrr, I can't push to your fork.. please rebase on master now.

> Test: Do we need tests for this? For what and how would we test here?

Test would be nice.. you could add a tiny `Client._validate_base_url(base_url)` staticmethod that gets called after the final `base_url` is stored during `__init__` (i.e. after url mapping etc. but before services are discovered at end of `__init__`). This staticmethod would be trivial to test in a unittest.

> Is the location where I put the pattern okay, or should it be moved elsewhere

Location is fine with me.. moving it to `header.py` would be fine, too. Just one thing.. if our pattern somehow prevents some valid URL can the user circumvent the check? Probably not so easily right now, right? Maybe changing `URL_REGEX` to be a class variable could make this accessible to the user in the worst case (i.e. if we botched it and disallow some valid URL)?

petrrr:
@megies: I added you as a collaborator, in case it serves in future.
Looking into the rest now.

petrrr:
@megies: I am looking into testing.
Should I put all the constants into the class, or only URL_REGEX?

petrrr:
@megies: I just pushed a CHANGELOG note and the change & test you proposed. I hope I understood all correctly.

megies:
> Should I put all the constants into the class, or only URL_REGEX?

Not too important, I was just thinking if the URL regex needs to be adapted (see e.g. https://github.com/obspy/obspy/pull/1162#commitcomment-16101900 ;-)) it might be easier to do if it's a class variable. As only the URL_REGEX is ultimately used in the match, this one might have been enough to move around, but really, not too important to me. It's fine now, I think.

Looks good! :+1:

petrrr:
@megies: I guess having the possibility to overwrite the URL pattern if really necessary is a good idea. The ones provided are rather plausibility check than strict rules, which would be quite complexes to get really accurate. But for its purpose these should be more than okay so let's not take this too far. 
I wanted to keep the related sub-patterns close by, that way it is easier to reason about. We could make them private, if this is relevant.

megies:
Nah, like I said, not too important. Fine with me and good to merge from my point. :+1:

megies:
Oh wait, this URL should pass, IMHO: https://github.com/obspy/obspy/pull/1162/files#diff-62952feaae3555b757b0a7e59dde7774R107

It's a valid URL after all, this started with catching some typos like `IGNV` or `IGPP`.. I think we really don't want to be overzealous, we might do harm at the same time then.

petrrr:
Okay, will change the domain name to pass with two levels as well. At this point I will revert the change to the tests as well.

petrrr:
@megies: Done, I hope (its late). I just reverted 74e4822 to keep the context of this dialog.

megies:
:+1: Thanks, I really feel better if those URLs aren't blocked.

petrrr:
@megies So can we proceed and merge it and then forget about this one? Otherwise I guess we get conflicts again pretty soon. Or do you prefer to leave it out of the release by now?

megies:
Nah, good to merge, I guess.

Issue #1195, opened 2015-12-02
Problem reading some TauP model files


krischer:
I did not look into the details but `obspy.taup` has problems parsing certain files. Others work fine.

Problematic files:

* https://dl.dropboxusercontent.com/u/3866312/taup_models/jb.nd
* https://dl.dropboxusercontent.com/u/3866312/taup_models/pwdk.nd
* https://dl.dropboxusercontent.com/u/3866312/taup_models/jb.nd

Usage:

```python
from obspy.taup.taup_create import TauP_Create

x = TauP_Create(input_filename='qdt.tvel', output_filename='blub')
x.loadVMod()
x.run()
```

Output:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/lion/workspace/code/obspy/obspy/taup/taup_create.py", line 119, in run
    self.tMod = self.createTauModel(self.vMod)
  File "/Users/lion/workspace/code/obspy/obspy/taup/taup_create.py", line 85, in createTauModel
    SlownessModel.DEFAULT_SLOWNESS_TOLERANCE)
  File "/Users/lion/workspace/code/obspy/obspy/taup/slowness_model.py", line 76, in __init__
    self.createSample()
  File "/Users/lion/workspace/code/obspy/obspy/taup/slowness_model.py", line 153, in createSample
    self.depthIncCheck()
  File "/Users/lion/workspace/code/obspy/obspy/taup/slowness_model.py", line 1012, in depthIncCheck
    'P')
  File "/Users/lion/workspace/code/obspy/obspy/taup/velocity_model.py", line 185, in evaluateAbove
    layer = self.layers[self.layerNumberAbove(depth)]
  File "/Users/lion/workspace/code/obspy/obspy/taup/velocity_model.py", line 132, in layerNumberAbove
    raise LookupError("No such layer.")
LookupError: No such layer.
```

QuLogic:
The problematic files are `.nd`, but the example is `qdt.tvel`?

krischer:
> The problematic files are .nd, but the example is qdt.tvel?

Yea I forgot about this. The qdt is some super low resolution model shipping with the Java version but I did not deem it worthy to add it to ObsPy. I just tested it with this PR and it works fine :-)

krischer:
Not this PR but #1197.

PR #1197, opened 2015-12-03
TauP model creation errors

QuLogic:
Fix two small errors. One where models can't be created in the current directory, and the one noted in #1195.


krischer:
Works like a charm.

PR #1198, opened 2015-12-03
More taup models

krischer:
#1197 enables the creation of two more models for TauP.

They are once again taken from the Java TauP variant.


megies:
Guess, I was too fast merging that earlier PR..

QuLogic:
Probably should wait for #1199 here at least, and maybe #1200.

krischer:
Should wait on #1224 and the `.npz` files have to be rebuilt once it is done.

QuLogic:
I'd like to add the alternate AK135 models as well.

krischer:
Which one do you mean? AFAIK the AK135-f model is the normal AK135 with some density and Q values but I think we don't use these for the taup calculations.

QuLogic:
I'm not sure they're exactly the same; the depths above 120 seem a bit different. Also, I'd like an `ak135f_no_mud` to be able to produce something compatible with specfem3d_globe.

krischer:
Sounds good.

QuLogic:
I'd like this in 1.0.0 so I can use it ;) I'll rebase and update later tonight.

QuLogic:
So it turns out it won't be possible to support AK135-f since the top layer has 0 S velocity, and the original Java implementation does not support that. I added the table anyway for if it becomes possible in the future, but didn't (and can't) build the model.

megies:
Is this ready to merge?

QuLogic:
Fine from my side, @krischer?

krischer:
I rebased against the latest master and rebuilt the first two models I added with the latest master taup version. @qulogic I guess you also did this but better safe than sorry ;-)

If CI passes I'll merge.

megies:
> If CI passes I'll merge.

Need to force geographiclib<1.46 temporarily if Travis' nod is necessary..

QuLogic:
I did rebuild earlier, and added a couple small tests, so AppVeyor should probably be enough this time.

krischer:
Passes on appveyor. Merging.

PR #1189, opened 2015-11-23
TauP long phase fix

QuLogic:
I've finally completed the port of the `shootRay`, etc., functionality from Java in order to fix long phases. Unfortunately, this is based on #1103 so it's a bit larger than it should be but it does fix the remaining test failures there. It is clear that this update fixes #986 from the examples there and shown below:
long_paths
You can also see from the updated test images that there are small changes in most paths, some are subtle, and some are more obviously better.

The correction to depth splitting was probably independent, but it has the side benefit in that it fixes #1057.

CC @crotwell; sorry it took so long, but it appears that the new functionality is working pretty well and we can see in the examples above and in the tests that the ray paths are much closer to the expected receiver location.

PR #1199, opened 2015-12-03
Use radius stored in the model and not the defaul earth radius.

krischer:
This is just a simple search-and-replace change but `obspy.taup` now uses the radius within each model and not the default radius. I tried it with some slightly different radius models and it seems to work.

I'll add some tests once #1189 is merged and I can rebase on top of it so don't merge before that!

### DIFF
--- a/obspy/taup/slowness_layer.py
+++ b/obspy/taup/slowness_layer.py
@@ -16,7 +16,7 @@ from .helper_classes import SlownessLayer, SlownessModelError
 from .velocity_layer import evaluateVelocityAtBottom, evaluateVelocityAtTop
 
 
-def bullenRadialSlowness(layer, p, radiusOfEarth, check=True):
+def bullenRadialSlowness(layer, p, planet_radius, check=True):
     """
     Calculate time and distance increments of a spherical ray.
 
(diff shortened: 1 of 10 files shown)


QuLogic:
Is this just a renaming, or is there some code change also? I though it did use the maximum radius from the model file, unless some higher level is ignoring that?

krischer:
There are two small code changes that now cause it to use the actual model radius - mainly the `slowness_model.﻿create_from_vlayer()` method. It always used the default radius.

obspy/taup/slowness_layer.py
```diff
@@ -33,8 +33,8 @@ def bullenRadialSlowness(layer, p, radiusOfEarth, check=True):
     :type layer: :class:`~numpy.ndarray`, dtype = :const:`SlownessLayer`
     :param p: The spherical ray paramater to use for calculation, in s/km.
     :type p: :class:`~numpy.ndarray`, dtype = :class:`float`
-    :param radiusOfEarth: The radius of the Earth to use, in km.
-    :type radiusOfEarth: float
+    :param planet_radius: The radius of the Earth to use, in km.
```
QuLogic: Description is a little off now.

obspy/taup/slowness_model.py
```diff
@@ -64,6 +63,7 @@ def __init__(self, vMod, minDeltaP=0.1, maxDeltaP=11, maxDepthInterval=115,
                  skip_model_creation=False):
 
         self.vMod = vMod
+        self.planet_radius = vMod.maxRadius if vMod else 6371.00
```
QuLogic: This is actually set in `createSample` for whatever reason.

obspy/taup/tau_model.py
```diff
@@ -470,7 +470,7 @@ def serialize(self, filename):
                   (native_str('maxInterpError'), np.float_),
                   (native_str('maxRangeInterval'), np.float_),
                   (native_str('minDeltaP'), np.float_),
-                  (native_str('radiusOfEarth'), np.float_),
+                  (native_str('planet_radius'), np.float_),
```
QuLogic: So this would require recreating the models, then...

QuLogic:
Needs a rebase now...

QuLogic:
Ping? #1200 definitely depends on this, at least a little bit.

QuLogic:
Another ping? Needs a rebase now.

krischer:
Yea..I have a branch that fixes this as well as #1200. I'll push it once its cleaned up a bit more.

krischer:
superseded by #1224

PR #1202, opened 2015-12-07
SDS filesystem client: additional methods, command line script for html report

megies:
Here's some more methods, mainly useful for QC checks on local SDS structure (latency, data availability, gaps/overlaps).

Works as is, but I should still add one or two simple tests..

### DIFF
--- a/obspy/clients/filesystem/sds.py
+++ b/obspy/clients/filesystem/sds.py
@@ -41,9 +41,12 @@ from future.builtins import *  # NOQA
 
 import glob
 import os
+import warnings
 from datetime import timedelta
 
-from obspy import Stream, read
+import numpy as np
+
+from obspy import Stream, read, UTCDateTime
 from obspy.core.util.misc import BAND_CODE
 
 
@@ -110,7 +113,7 @@ class Client(object):
         self.fileborder_samples = fileborder_samples
 
     def get_waveforms(self, network, station, location, channel, starttime,
-                      endtime, merge=-1, sds_type=None):
+                      endtime, merge=-1, sds_type=None, **kwargs):
         """
         Read data from a local SeisComP Data Structure (SDS) directory tree.
 
@@ -152,6 +155,56 @@ class Client(object):
             msg = ("'endtime' must be after 'starttime'.")
             raise ValueError(msg)
         sds_type = sds_type or self.sds_type
+
+        st = Stream()
+        full_paths = self._get_filenames(
+            network=network, station=station, location=location,
+            channel=channel, starttime=starttime, endtime=endtime,
+            sds_type=sds_type)
+        for full_path in full_paths:
+            st += read(full_path, format=self.format, starttime=starttime,
+                       endtime=endtime, **kwargs)
+
+        # make sure we only have the desired data, just in case the file
+        # contents do not match the expected SEED id
+        st = st.select(network=network, station=station, location=location,
+                       channel=channel)
+
+        # avoid trim/merge operations when we do a headonly read for
+        # `_get_availability_percentage()`
+        if kwargs.get("_no_trim_or_merge", False):
+            return st
+
+        st.trim(starttime, endtime)
+        if merge is None or merge is False:
+            pass
+        else:
+            st.merge(merge)
+        return st
+
+    def _get_filenames(self, network, station, location, channel, starttime,
+                       endtime, sds_type=None):
+        """
+        Get list of filenames for certain waveform and time span.
+
+        :type network: str
+        :param network: Network code of requested data (e.g. "IU").
+        :type station: str
+        :param station: Station code of requested data (e.g. "ANMO").
+        :type location: str
+        :param location: Location code of requested data (e.g. "").
+        :type channel: str
+        :param channel: Channel code of requested data (e.g. "HHZ").
+        :type starttime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+        :param starttime: Start of time span.
+        :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+        :param endtime: End of time span.
+        :type sds_type: str
+        :param sds_type: Override SDS data type identifier that was specified
+            during client initialization.
+        :rtype: list of str
+        """
+        sds_type = sds_type or self.sds_type
         # SDS has data sometimes in adjacent days, so also try to read the
         # requested data from those files. Usually this is only a few seconds
         # of data after midnight, but for now we play safe here to catch all
@@ -171,7 +224,6 @@ class Client(object):
             t += timedelta(days=1)
         year_doy.add((t_max.year, t_max.julday))
 
-        st = Stream()
         full_paths = set()
         for year, doy in year_doy:
             filename = self.FMTSTR.format(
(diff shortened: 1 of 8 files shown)


obspy/clients/filesystem/sds.py
```diff
@@ -108,7 +113,7 @@ def __init__(self, sds_root, sds_type="D", format="MSEED",
         self.fileborder_samples = fileborder_samples
 
     def get_waveforms(self, network, station, location, channel, starttime,
-                      endtime, merge=-1, sds_type=None):
+                      endtime, merge=-1, sds_type=None, **kwargs):
```
QuLogic: Should be documented as (mostly) extra args for specific formats.

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        st.traces = [tr for tr in st
+                     if not (tr.stats.endtime < starttime or
+                             tr.stats.starttime > endtime)]
+
+        if not st:
+            return (0, 1)
+
+        total_duration = endtime - starttime
+        # sum up gaps in the middle
+        gaps = [gap[6] for gap in st.getGaps()]
```
QuLogic: Hmm, was `getGaps` not renamed?

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+                if st:
+                    break
+            time -= 24 * 3600
+
+        return max([tr.stats.endtime for tr in st])
+
+    def get_latency(self, network, station, location, channel,
+                    sds_type=None, stop_time=None):
+        """
+        Get time of last sample for given stream.
```
QuLogic: It's not really the "time of", though...

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        :type location: str
+        :param location: Location code of requested data (e.g. "").
+        :type channel: str
+        :param channel: Channel code of requested data (e.g. "HHZ").
+        :type sds_type: str
+        :param sds_type: Override SDS data type identifier that was specified
+            during client initialization.
+        :type stop_time: :class:`~obspy.core.utcdatetime.UTCDateTime`
+        :param stop_time: Time at which the search for data is stopped and
+            `None` is returned. If not specified, stops at ``1950-01-01T00``.
```
QuLogic: `None` should probably have two backticks?

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
```
QuLogic: `None` should probably have two backticks? It does on the next line...

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        :param sds_type: Override SDS data type identifier that was specified
+            during client initialization.
+        :rtype: bool
+        """
+        sds_type = sds_type or self.sds_type
+        pattern = self.FMTSTR.format(
+            year="*", network=network, station=station, location=location,
+            channel=channel, doy=0, type=sds_type)
+        # can not insert wildcard for day-of-year above, so replace it now
+        pattern = pattern.rsplit(".", 1)[0] + "*"
```
QuLogic: This assumes you're using the default pattern, doesn't it?

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        :rtype: list
+        :returns: List of (network, station, location, channel) 4-tuples of all
+            available streams in archive.
+        """
+        sds_type = sds_type or self.sds_type
+        pattern = self.FMTSTR.format(
+            year="*", network="*", station="*", location="*",
+            channel="*", doy=0, type=sds_type)
+        # can not insert wildcard for day-of-year above, so replace it now
+        pattern = pattern.rsplit(".", 1)[0] + "*"
```
QuLogic: Same here.

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        pattern = os.path.join(self.sds_root, pattern)
+        all_files = glob.glob(pattern)
+        result = set()
+        for file_ in all_files:
+            network, station, channel, filename = file_.split("/")[-4:]
```
QuLogic: This one too.

obspy/clients/filesystem/sds.py
```diff
@@ -169,24 +224,274 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
+        pattern = self.FMTSTR.format(
+            year="*", network="*", station="*", location="*",
+            channel="*", doy=0, type=sds_type)
+        # can not insert wildcard for day-of-year above, so replace it now
+        pattern = os.path.join(self.sds_root, pattern)
+        pattern = os.path.dirname(os.path.dirname(pattern))
+        all_folders = glob.glob(pattern)
+        result = set()
+        for file_ in all_folders:
+            network, station = file_.split("/")[-2:]
```
QuLogic: And probably here, too.

bonaime:
@megies: are you thinking of a method to import (and merge small) files from a directory to an SDS structure ?

obspy/clients/filesystem/sds.py
megies: Not *yet*.. I put a note in there https://github.com/obspy/obspy/issues/1039#issuecomment-150163012 already, so we do it before release..

obspy/clients/filesystem/sds.py
megies: Thanks.. copy/paste bug from `_get_endtime`..

obspy/clients/filesystem/sds.py
megies: Yep, thanks. We should at least have this right in new code. :+1:

obspy/clients/filesystem/sds.py
megies: Hmm, you're right, that's vulnerable.. I'll try to change this, replacing everything not in NSLC with `*`.

obspy/clients/filesystem/sds.py
megies: This was a bit tricky to generalize.. lots of regex magic in there now..

obspy/clients/filesystem/sds.py
megies: Same here.. lots of regex magic in there now..

obspy/clients/filesystem/sds.py
megies: Same here.. lots of regex magic in there now..

megies:
> @megies: are you thinking of a method to import (and merge small) files from a directory to an SDS structure ?

I don't need that right now, so I'll not work on it right now, but that sounds like a useful addition.. feel free to give it a shot if you'd like. Although messing around with the actual data archive is something that I usually avoid to do too programmatically..

megies:
This is what the basic SDS report html page looks like.. of course this is just for the dummy test SDS we use in obspy right now..

screenshot from 2015-12-08 11 33 17

obspy/clients/filesystem/sds.py
megies: I added a short note in the parameter description below, which should be enough..

megies:
No idea why this seems to influence warnings of `read(..., headonly=True, startime=XXX, ...)`, because I put my call that is issuing the same warning that is failing to get shown in the Travis test within a `with warnings.catch_warnings()` context manager which is supposed to reset the original warning settings?? https://github.com/obspy/obspy/pull/1202/files#diff-93c710504a1d59384f7fc6ad859e2f06R297

megies:
> context manager which is supposed to reset the original warning settings??

Obviously this is fixed in Py 3.x but not in 2.7..
https://bugs.python.org/issue4180#msg100525

QuLogic:
Probably should squash those exploratory appveyor commits, and I think maybe change the colours a bit to match [Set2](http://colorbrewer2.org/) or similar. Also, are the colours in the table related to the colours in the figure? I think not, but I'm not sure...

megies:
Yeah, of course already intended to squash all these mini commits at the end, before merging, don't worry. :+1: 
Just preferred to keep force-pushes to a minimum, otherwise nobody can follow whats going on here anymore until it's ready.

Color-wise.. this is mostly intended as a one-look quick assessment page for current state of health in local archive of a large number of stations. The harsh colors are kind of intentional (to highlight stations that are currently down or lagging a lot), and every color can be user set on command line.. so I think it's OK as is.

Colors in the figure are not related, simple reusing the `obspy-scan` output (KISS), just as another low-hanging-fruit level of detail, to get an idea of how gaps exactly behave over time in the current monitored time span..

megies:
Colorbrewer Set1 looks better, you're right @QuLogic.. changed it. thanks!

megies:
Will rebase/squash this..

megies:
Merging..

PR #1207, opened 2015-12-11
obspy-scan: gap-handling with user-specified explicit start/end time

megies:
I think when a start and/or end time is specified by user, data gaps outside of the traces data but within those start/end times should be counted as gaps in percentage and also in the plot (red bars).

 - count missing data in between e.g. --start-time and first data sample
   as gap (in percentages)
 - plot missing data at start/end as gap, same as a gap in between data
   pieces


megies:
Test fail is unrelated, I've seen this one before, I think it's from an unclosed figure and some sloppy `plt.gcf()` or `plt.gca()` call and would be fixed with #1156.

I don't think there will be objections about this PR so I'll merge it in a bit if nobody steps forward..

megies:
Test fail unrelated as mentioned before, no opposing opinions, so merging.

PR #1211, opened 2015-12-18
faster and more readable array transfer function

MMesch:
This little modification is makes the computation of the array transfer function more readable, and it is 6x faster. It uses more memory though but 2D grids are usually still ok in memory. Not sure which one you prefer. ~18ms vs 1.33s are the run times of the array example.

### DIFF
--- a/obspy/signal/array_analysis.py
+++ b/obspy/signal/array_analysis.py
@@ -731,7 +731,6 @@ def get_spoint(stream, stime, etime):
                         tr.stats.sampling_rate + .5)
     return spoint, epoint
 
-
 def array_transff_wavenumber(coords, klim, kstep, coordsys='lonlat'):
     """
     Returns array transfer function as a function of wavenumber difference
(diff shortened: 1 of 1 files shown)


obspy/signal/array_analysis.py
```diff
@@ -731,7 +731,6 @@ def get_spoint(stream, stime, etime):
                         tr.stats.sampling_rate + .5)
     return spoint, epoint
 
-
```
QuLogic: This line should not be removed.

obspy/signal/array_analysis.py
```diff
@@ -834,7 +830,6 @@ def array_transff_freqslowness(coords, slim, sstep, fmin, fmax, fstep,
     transff /= transff.max()
     return transff
 
-
```
QuLogic: This line should also not be removed.

obspy/signal/array_analysis.py
```diff
@@ -763,18 +762,15 @@ def array_transff_wavenumber(coords, klim, kstep, coordsys='lonlat'):
...
-            for k in range(len(coords)):
-                _sum += np.exp(complex(0.,
-                               coords[k, 0] * kx + coords[k, 1] * ky))
-            transff[i, j] = abs(_sum) ** 2
+    ks  = np.transpose(np.vstack( (kxgrid.flatten(),kygrid.flatten()) ))
 
-    transff /= transff.max()
-    return transff
+    k_dot_r = np.einsum('ni,mi->nm',ks,coords[:,:2]) #z coordinate is not used
+    transff = np.abs(np.sum(np.exp(1j*k_dot_r), axis=1))**2/len(coords)**2
```
QuLogic: Please use PEP8 style; a space is required around operators and after commas and comment characters.

obspy/signal/array_analysis.py
```diff
@@ -763,18 +762,15 @@ def array_transff_wavenumber(coords, klim, kstep, coordsys='lonlat'):
...
+
+    return transff.reshape(nkx,nky)
```
QuLogic: I'm pretty sure a set of parentheses is missing here.

QuLogic:
Unfortunately, the results are not quite the same; it is fine if there are small fluctuations due to differing orders of operations, but these are 20-70% of the original values.

MMesch:
Oups, I'll have a look into this! It should be the same. Probably because there is no normalization in the end?

MMesch:
ok I think there is a turned axis.

MMesch:
OK. I confirm the result is exactly the transpose of the previous function.

MMesch:
OK. Its fine now. I think in general it would be better to make a function that takes wavenumbers and positions as input instead of klim and kstep and returns the transfer function for each k. This leads to better organisation of the code and makes the function more flexible. What do you think about this?

megies:
> A further advantage is that it automatically works for 2D ks + 2D position or 3D ks and 3D positions. What do you think about this?

I don't use it a lot but other people might well do. Maybe it would be an option to keep the current routine and make it a wrapper around a new function with your recommended call parameters (where most of the current code would then reside, I guess)?

krischer:
This can be merged, right?

MMesch:
yes from my side...

PR #1212, opened 2015-12-18
Common Client interfaces

jkmacc-LANL:
**Motivation**: Many `Client` classes have almost the same `get_waveforms`, `get_stations`, and `get_events` methods, modulo different `**kwargs`. It may it be useful to really guarantee that all `Client` classes provide the same minimum signatures and returns for these methods.

**Proposed Changes**: Enforce `get_waveforms`, `get_stations`, and `get_events` methods for classes that support them using `ABCMeta` classes.  Client authors could inherit from one or more of `WaveformClient`, `StationClient`, and `EventClient`, to inherit the common interfaces.

**Benefits**: Users could write applications that employ any client, regardless of its origin, to get a `Stream` from a `get_waveforms` method, a `Catalog` from a `get_events` method, and an `Inventory` from a `get_stations` method. It may encourage `Client` writers to connect their data sources to `Stream`, `Inventory`, and `Catalog` types, and encourage users to rely on them in their applications.

Initial implementation and example by @dsentinel .


obspy/clients/base.py
```diff
@@ -0,0 +1,90 @@
...
+    def get_service_version(self):
+        """Returns a semantic versioning string."""
+        pass
+
+
+class WaveformClient(BaseClient):
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def get_waveforms():
```
dsentinel: I would add the common args to these abstract methods to document the common args.

obspy/clients/base.py
```diff
@@ -0,0 +1,90 @@
...
+methods common to all.
+
+Individual client classes inherit from one or more of WaveformClient,
+EventClient, and StationClient, and re-program the get_waveforms, get_events,
+and/or get_stations methods, like in the example below.
+
+
+.. rubric:: Example
+
+class SeedlinkClient(WaveformClient, EventClient):
```
megies: Maybe better use a different client as an example, seedlink doesn't serve events..

obspy/clients/base.py
```diff
@@ -0,0 +1,95 @@
...
+        if url:
+            self.conn = open(url)
+
+    def get_server_version(self):
+        self.conn.get_version()
+
+    def get_waveforms(self, station, channel):
+        return self.conn.fetch_mseed(station, channel)
+
+    def get_events(self, starttime, endtime):
```
megies: call signatures have to be adapted to cd9b9ad846e9ee0b.

jkmacc-LANL:
What if, instead of the many min/max latitude/longitude/depth/magnitude parameters, they get shortened a bit?  Something like the call below, for example. Maybe `depths=(mindep, maxdep)` tuples, and maybe a dict for the magnitudes?  Maybe this is moot because it would break the FDSN `Client.get_events`, for example...
```python
my_event_client.get_events(starttime, endtime, region=(minlat, maxlat, minlon, maxlon),
                           depths=(mindep, maxdep), magnitudes={magtype: (minmag, maxmag)},
                           author=None):
```

megies:
I guess this would make a lot of sense, but it probably might be quite some work to keep the current functionality running.
Maybe even use a modified/new `Restriction` class similar to https://github.com/obspy/obspy/pull/1131/files#diff-0b2ada75b9d05d9d7a9c2eb750ef025cR19? `my_event_client.get_events(*restrictions)` But not sure how this would look like in detail..

jkmacc-LANL:
Hmmm.  Defining a new standard way to use all Clients may be a discussion larger than this PR.  Wouldn't it require someone to re-write the call signatures for any client that inherits from these base clients?

megies:
Yeah, we shouldn't touch the call syntaxes unless more people agree, let's stay to changing the basics that doesn't influence users here, the rest needs more discussion..

jkmacc-LANL:
If the signatures for each client type are cleaned up and accepted, should there be a separate PR for implementing them into the current clients?

krischer:
Hi all,

sorry for joining this discussion so late.

* Just my personal opinion: I don't really like having default argument lists. It will not work for all web services  - some will have less/other parameters - and the `abc` module does not enforce them in any case. I think its a bit confusing if an overwritten method on a child class suddenly has less parameters. I guess this is intended as a suggestion for the parameter names - in that case I would rather see them in the docstrings.
* The `__metaclass__` class attribute will only work on Python 2. On Python 3 it does nothing. `future` provides a way to do it for both: http://python-future.org/compatible_idioms.html#metaclasses
* What's the purpose of the `get_client_version()` method? `obspy.__version__` already is a unique version number.
* `get_service_version()` could also be an abstract `service_version` property. Not sure which is better.
* Maybe for a future PR? A nice to have thing would be a default user agent implemented at the `BaseClient` class. Most clients need to send one in any case. Something like this: https://github.com/obspy/obspy/blob/b6d28ac984e47e9e7b121689e2c9b809ce5966ed/obspy/clients/fdsn/header.py#L61-L68
It must be changeable in the `__init__` method so maybe this works:
```python
class BaseClient(...):
    def __init__(self, user_agent=DEFAULT_USER_AGENT):
        self.user_agent = user_agent
```
* Again maybe for a future PR: Most clients work over HTTP so a nice downloading method would be good to have at the `BaseClient` level. Something like this: https://github.com/obspy/obspy/blob/b6d28ac984e47e9e7b121689e2c9b809ce5966ed/obspy/clients/fdsn/client.py#L1600-L1659 Or we just use the `requests` module?

> If the signatures for each client type are cleaned up and accepted, should there be a separate PR for implementing them into the current clients?

Personally I would like to merge this ASAP as I'll add a new client next week so I could just base it on this. But I'm fine with either possibility.

obspy/clients/base.py
```diff
@@ -0,0 +1,131 @@
...
+class MyNewClient(WaveformClient, StationClient):
+    def __init__(self, url=None):
+        self._version = '1.0'
+        if url:
+            self.conn = open(url)
+
+    def get_server_version(self):
+        self.conn.get_version()
+
+    def get_waveforms(self, network, station, location, channel, starttime
```
krischer: Missing comma at the end of line.

obspy/clients/base.py
```diff
@@ -0,0 +1,131 @@
...
+                      endtime):
+        return self.conn.fetch_mseed(network, station, location, channel,
+                                     starttime, endtime)
+
+    def get_stations(self, network, station, location, channel, starttime,
+                     endtime):
+        return self.conn.fetch_inventory(network, station, location, channel
```
krischer: One more missing comma.

obspy/clients/base.py
```diff
@@ -0,0 +1,131 @@
...
+
+.. rubric:: Example
+
+class MyNewClient(WaveformClient, StationClient):
+    def __init__(self, url=None):
+        self._version = '1.0'
+        if url:
+            self.conn = open(url)
+
+    def get_server_version(self):
```
krischer: Called `get_service_version` in the actual implementation below.

jkmacc-LANL:
@krischer I'd also start to use this sooner rather than later, so I'll try make those changes ASAP.  

If you don't mind, I'll leave the user agent part up to you, either in this PR or a different one.  

For the sake of simplicity, I'm OK with just writing parameter suggestions in the `BaseClient` docstring for now.  

Also, regarding the download URL, I think it'd be a good idea to do it using `requests`, if another dependency isn't too onerous, as it seems like a library worth knowing and using.  How would `download_url` be used in the filesystem client, though, for example?

obspy/clients/base.py
jkmacc-LANL: ?

jkmacc-LANL:
* [x] docstrings instead of parameter lists
* [x] `__metaclass__` that works for Python3
* ~~`services_version` property instead of method~~  
I like properties better, but I'd rather not force client authors to make one.  Methods are fundamental.

As long as the checks pass, this is all I intended for this PR.  Thanks!

obspy/clients/base.py
krischer: This method should be called `get_service_version` as that is how it is called in the actual abstract base class.

krischer:
> Also, regarding the download URL, I think it'd be a good idea to do it using requests, if another dependency isn't too onerous, as it seems like a library worth knowing and using. How would download_url be used in the filesystem client, though, for example?

Good point. We could just do another mixin called `HTTPClient`.

> As long as the checks pass, this is all I intended for this PR. @krischer Did you want me to add you to this, so you can do the user agent thing? Thanks!

That's fine. I'll do the user agent in a different PR. You could still add me if you want so I can fix small things myself.

I'll merge as soon as the minor nitpick is fixed.

jkmacc-LANL:
@krischer Done.  Futher PRs will likely be done from obspy/master, hopefully avoiding permissions issues. Thanks again.

krischer:
Thanks!

megies:
I guess to really put this into action we should use this in the actual clients' definitions (as far as possible). Also, not having explicit args/kwargs e.g. for `get_waveforms` but only having the definitions in the docstring seems a bit half-finished to me, but I didn't look into this meta topic too much.

Anyway, I think this is a step in the right direction. Thanks @jkmacc-LANL!

PR #1218, opened 2016-01-06
multithread and multiprocess remove_response function

MMesch:
Not sure if this is interesting for you but here is a little update that adds multithreading and multiprocessing support to the stream.remove_response() function. It speeds up the process by about 3x on my 4 core cpu.

Multithreading needs to be made safer against thread Exceptions (does anyone have experience with this?)
The multiprocessing code is much shorter but somewhat ugly, takes a bit longer and needs more memory. It could be the better option, however, because it isn't limited by the GIL.

Here is a little test script for the code:

    waveforms_raw = read('waveforms.0.mseed')                                       
    stations = read_inventory('stations.xml')                                       
    waveforms_raw.attach_response(stations)                                         
                                                                                
    waveforms0 = waveforms_raw.copy()                                               
    waveforms1 = waveforms_raw.copy()                                               
    waveforms2 = waveforms_raw.copy()                                               
                                                                                
    pre_filt = 1./200., 1./100., 1./50., 1./25.                                     
    tstart = time.time()                                                            
    waveforms0.remove_response(pre_filt=pre_filt, taper=True)                       
    tend = time.time()                                                              
    print 'serial execution:', tend-tstart                                          
                                                                                
    pre_filt = 1./200., 1./100., 1./50., 1./25.                                     
    tstart = time.time()                                                            
    waveforms1.remove_response(nthreads=4, pre_filt=pre_filt, taper=True)           
    tend = time.time()                                                              
    print 'multithreading execution:', tend-tstart                                  
                                                                                
    #waveforms = waveforms_raw.copy()                                               
    pre_filt = 1./200., 1./100., 1./50., 1./25.                                     
    tstart = time.time()                                                            
    waveforms2.remove_response(nprocs=4, pre_filt=pre_filt, taper=True)             
    tend = time.time()                                                              
    print 'multiprocessing execution:', tend-tstart                                 
                                                                                
    #---- plotting ----                                                             
    trace0 = waveforms1[0]                                                          
    trace1 = waveforms1[0]                                                          
    trace2 = waveforms2[0]                                                          
    trace_raw = waveforms_raw[0]                                                    
    fig, ax = plt.subplots(2,1)                                                     
                                                                                
    ax[0].plot(trace_raw.times(), trace_raw.data, c='red')                          
    ax[1].plot(trace0.times(), trace0.data, '-',c='yellow')                         
    ax[1].plot(trace1.times(), trace1.data, '--',c='blue')                          
    ax[1].plot(trace2.times(), trace2.data, 'x',c='green')                          
                                                                                
    plt.show()


krischer:
In general I like this quite a bit but we have to be careful. I believe this idea already came up a couple of times and we should be able to generalize this to all functions on the `Stream` object that simply apply the same function to all its `Trace` objects.

> The multiprocessing code is much shorter but somewhat ugly, takes a bit longer and needs more memory.

It is slower as it has to fork the process a couple of times which has some overhead. Furthermore the GIL is actually released when the underlying `evalresp` is called and also for most numpy function calls. Thus this actually runs in parallel to a large extent even when using threads.

Multiprocessing is also a bit dangerous when used with numpy as it forks the process by default and many BLAS implementations (some OpenBLAS versions and Apple's Accelerate framework) don't work with forks. Some more information here: https://github.com/obspy/obspy/wiki/Notes-on-Parallel-Processing-with-Python-and-ObsPy
Thus we unfortunately cannot use the `multiprocessing` module in most parts of ObsPy.

I suggest the following changes:

* Remove the `multiprocessing` code path.
* Change the function call signature to `Stream.remove_response(self, nthreads=-1, *args, **kwargs)`. A value of `-1` should result in using `min(len(self.traces), logical_machine_cores)` threads.
* Replace the custom queue implementation with `multiprocessing.dummy.Pool`. This is exactly the same API as you already used for the multiprocessing version but it uses threads. Its not really advertised a lot but works on all Python versions and is really handy. I'm not sure why you think that the `Pool` variant is ugly - I find it much simpler to reason about then your queue version and its likely also more stable as its been around for a while. This partially solves your exception handling problem - we still have to figure out what to do when the response removal fails for a couple of traces...
* Also use the Pool version if `nthreads=1` - only a single code path is always nice.

We can think about generalizing this to a lot more functions at a later point in time.

If this is mainly about speeding up the time for instrument corrections: I think it can be made a lot faster by calculating the response at fewer frequency points and interpolate in-between. Evalresp can do this but we don't really expose the functionality - or we could just code it in numpy/scipy. The actual response calculation is by far the most expensive part in the whole thing (last time I benchmarked even the fft/ifft was cheap in comparison) and most response curves are very smooth away from nyquist.

MMesch:
OK! Agree with all of these suggestions. I didn't know the multiprocessing.dummy.Pool. My fear with multiprocessing was that there would be a large memory overhead if traces have to be copied back and forth.

I made a little test to compare dummy.Pool threading with threading.Thread and it really uses the same amount of memory, although threading is slightly faster (27s instead of 32s). Here is the memory consumption:

benchmark

claudiodsf:
Just a quick comment: pretty much everything in `Stream` could be sped up by multithreading, since many functions do internally a `for tr in st:`.
I'm thinking at filtering, in particular.

MMesch:
Yes. It is really quick to implement with multiprocessing.dummy. Maybe it shouldn't be the default option in the beginning though until it has been tested extensively.

obspy/core/stream.py
```diff
@@ -3040,11 +3046,25 @@ def remove_response(self, *args, **kwargs):
             original data, use :meth:`~obspy.core.stream.Stream.copy` to create
             a copy of your stream object.
         """
-        for tr in self:
-            tr.remove_response(*args, **kwargs)
+        if nthreads > 0:
+            # multi-threaded instrument response removal. multiprocessing.dummy
+            # uses threads with the Pool API of multiprocessing
+            pool = Pool(processes=nthreads)
```
krischer: I think you can replace the whole `if/else` with

```python
# Guard against empty Stream objects.
if not len(self):
    return self

pool = Pool(processes=max(1, min(int(nthreads), len(self.traces)))
pool.map(...)
```

The tests will catch if this does not work.

krischer:
> Just a quick comment: pretty much everything in Stream could be sped up by multithreading, since many functions do internally a for tr in st:.
> I'm thinking at filtering, in particular.

Definitely. Let's try it here first as its likely one of the slowest operations. If successful we do it for all loops over trace objects.

krischer:
While this is ready I think we should merge this after we release the next ObsPy version. This is a very central method in ObsPy and if it does not work for some reason people might be screwed.

We'll merge it directly after the next release and then we have some time to properly test it on all systems.

megies:
> We'll merge it directly after the next release and then we have some time to properly test it on all systems.

ping @krischer

krischer:
This branch can now run all methods (for which it makes sense) of the `Stream` object in parallel. It is controlled by the `nthreads` argument. Thread creation unfortunately has quite some overhead and most processing functions are so fast that the overhead dominates the runtime. Thus the `nthreads` argument defaults to 1 and it does not actually create a thread.

Its really only worth it for some very expensive operations (instrument removal or lanczos interpolation are two examples) or if the Stream object contains a large number of traces. Then one can have a nice speedup (my laptop has 4 physical and 8 logical cores) - also finally an excuse to use the fancy violin plots:

speed

We could potentially create a thread pool upon starting ObsPy that we then just reuse as appropriate but I don't think we want to add that complexity. If somebody just wants to run ObsPy in parallel - MPI is in my experience the solution with the least problems: https://gist.github.com/krischer/2c7b95beed642248487a

I don't think it can hurt to merge this as i does add some functionality - please review!

obspy/core/stream.py
```diff
@@ -2049,10 +2057,22 @@ def simulate(self, paz_remove=None, paz_simulate=None,
             st.simulate(paz_remove=paz_sts2, paz_simulate=paz_1hz)
             st.plot()
         """
-        for tr in self:
-            tr.simulate(paz_remove=paz_remove, paz_simulate=paz_simulate,
-                        remove_sensitivity=remove_sensitivity,
-                        simulate_sensitivity=simulate_sensitivity, **kwargs)
+        # Must be passed as part of the kwargs as otherwise the historically
```
QuLogic: Not sure I understand this reasoning.

obspy/core/stream.py
```diff
@@ -2049,10 +2057,22 @@ def simulate(self, paz_remove=None, paz_simulate=None,
...
+        # Grown interface would have to change.
+        if "nthreads" in kwargs:
+            nthreads = kwargs.pop("nthreads")
```
QuLogic: `kwargs.pop('nthreads', 1)`

megies:
> We could potentially create a thread pool upon starting ObsPy that we then just reuse as appropriate but I don't think we want to add that complexity. If somebody just wants to run ObsPy in parallel - MPI is in my experience the solution with the least problems: https://gist.github.com/krischer/2c7b95beed642248487a

I agree, the default should stay conservative and least error prone.
Nice gist! Should be added to the tutorial (which actually has been in need for a complete rehaul for a long time.. :disappointed: )

krischer:
Please don't merge yet - this needs to be refactored a bit more..Also lots of flake8 failures :-(

barsch:
This pull request has been mentioned on **ObsPy Forum**. There might be relevant details there:

https://discourse.obspy.org/t/data-processing-with-multithreading-multiprocessing-mpi/964/9

PR #1219, opened 2016-01-09
Some very small additions (`fig` and `plot` args)

MMesch:
Hi all,

this PR contains, two additions to the documentation and the additional optional 'fig' argument for the trace.remove_response(plot=True) function (similar to stream.plot()).

Two points were unclear to me:
 1. What is the projection used by utl_geo_km? If it is a 'proper' projection, its name should be in the docs. Obspy could consider wrapping some library that does the proper projections and using this behing the get_geometry function of the arrays.
 2. what should be the standard function call to make plots go to screen immediately, to an external fig instance, or to a file? It could be for example three different arguments as: plot=True fig=fig fname='figure.png' ; or a single argument plot=True/fig_instance/fname .

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2401,7 +2401,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
     @_add_processing_info
     def remove_response(self, output="VEL", water_level=60, pre_filt=None,
                         zero_mean=True, taper=True, taper_fraction=0.05,
-                        plot=False, **kwargs):
+                        plot=False, fig=None, **kwargs):
         """
         Deconvolve instrument response.
 
@@ -2584,7 +2584,8 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             bbox = dict(boxstyle="round", fc="w", alpha=1, ec="w")
             bbox1 = dict(boxstyle="round", fc="blue", alpha=0.15)
             bbox2 = dict(boxstyle="round", fc="red", alpha=0.15)
-            fig = plt.figure(figsize=(14, 10))
+            if fig is not None:
+                fig = plt.figure(figsize=(14, 10))
             fig.suptitle(str(self))
             ax1 = fig.add_subplot(321)
             ax1b = ax1.twinx()
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2598,7 +2598,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             ax6 = fig.add_subplot(326, sharex=ax4)
             for ax_ in (ax1, ax2, ax3, ax4, ax5, ax6):
                 ax_.grid(zorder=-10)
-            text = 'pre_filt: ({:.2f} {:.2f} {:.2f} {:.2f})'.format(*pre_filt)
+            text = 'pre_filt: ({:.3g} {:.3g} {:.3g} {:.3g})'.format(*pre_filt)
             ax1.text(0.05, 0.1, text, ha="left", va="bottom",
                      transform=ax1.transAxes, fontsize="large", bbox=bbox,
                      zorder=5)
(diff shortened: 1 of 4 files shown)


krischer:
Can you fix the flake8 warnings? Otherwise this looks good to go.

> What is the projection used by utl_geo_km? If it is a 'proper' projection, its name should be in the docs. Obspy could consider wrapping some library that does the proper projections and using this behing the get_geometry function of the arrays.

The flattening value is very close to the one from WGS84 so I guess that's the used projection? But I agree - we should just make `pyproj` a hard dependency as we already use it in a lot of places. Best done in a separate PR/issue.

> what should be the standard function call to make plots go to screen immediately, to an external fig instance, or to a file? It could be for example three different arguments as: plot=True fig=fig fname='figure.png' ; or a single argument plot=True/fig_instance/fname .

This another of these things that kind of grew and should be unified across the codebase. I like your single argument idea but it will break lots of existing code so I guess its not worth the effort to do it as we would have to cleanly deprecate it. I'll open a separate issue to discuss this in more detail.

krischer:
> I like your single argument idea but it will break lots of existing code so I guess its not worth the effort to do it as we would have to cleanly deprecate it. I'll open a separate issue to discuss this in more detail.

Ignore that - it might be possible - the optional plotting functions are fairly new I believe.

MMesch:
The test fails because the text in the remove_response plot has a slightly different format now to avoid very long numbers like 0.3333333333333

obspy/core/trace.py
```diff
@@ -2597,9 +2598,10 @@ def remove_response(self, output="VEL", water_level=60, pre_filt=None,
             ax6 = fig.add_subplot(326, sharex=ax4)
             for ax_ in (ax1, ax2, ax3, ax4, ax5, ax6):
                 ax_.grid(zorder=-10)
-            ax1.text(0.05, 0.1, 'pre_filt: %s' % str(pre_filt),
-                     ha="left", va="bottom", transform=ax1.transAxes,
-                     fontsize="large", bbox=bbox, zorder=5)
+            text = 'pre_filt: ({:.3g} {:.3g} {:.3g} {:.3g})'.format(*pre_filt)
```
krischer: Changing this line to 

```python
text = 'pre_filt: [{:.3g}, {:.3g}, {:.3g}, {:.3g}]'.format(*pre_filt)
```
should fix the tests and is IMHO also a bit more legible. Or you have to update the baseline test image.

QuLogic:
Not another geodetic thing... Can we replace `utl_geo_km` by something in `core.geodetics`?

obspy/core/trace.py
```diff
@@ -2597,9 +2598,10 @@ def remove_response(self, output="VEL", water_level=60, pre_filt=None,
             ax6 = fig.add_subplot(326, sharex=ax4)
             for ax_ in (ax1, ax2, ax3, ax4, ax5, ax6):
                 ax_.grid(zorder=-10)
-            ax1.text(0.05, 0.1, 'pre_filt: %s' % str(pre_filt),
-                     ha="left", va="bottom", transform=ax1.transAxes,
-                     fontsize="large", bbox=bbox, zorder=5)
+            text = 'pre_filt: [{:.3g} {:.3g} {:.3g} {:.3g}]'.format(*pre_filt)
```
krischer: The only difference between the existing version and the new one are that the new one has no commas. So either change the baseline image or adapt the code. If you run the tests locally you should also be able to check this. Also please ping the thread if you want another review. Otherwise we are not notified.

krischer:
> Not another geodetic thing... Can we replace utl_geo_km by something in core.geodetics?

It should be possible to replace it with `gps2dist_azimuth`.

megies:
@MMesch, please also see #1156, which is related.

obspy/core/trace.py
krischer: ping. Can you change this or give me access to your fork?

Issue #1200, opened 2015-12-03
Mars models don't work with obspy.taup


krischer:
This model for example does not work with the current master: https://dl.dropboxusercontent.com/u/3866312/taup_models/original/mars.tvel

There are just a couple of code paths and assumptions that are not tested and it should be fairly simple to fix. For example the code currently is only tested with a solid core. I fixed a couple of things but then decided to wait until #1189 is done so it does not have to be done twice.

QuLogic:
Does this model work with the Java version?

krischer:
I just checked - once I added two comment lines up top but then yes. I just updated the file linked above.

```bash
$ ./taup_time --ph P -sta 0 0 -evt 10 10 -mod mars.taup

Model: mars
Distance   Depth   Phase   Travel    Ray Param  Takeoff  Incident  Purist    Purist
  (deg)     (km)   Name    Time (s)  p (s/deg)   (deg)    (deg)   Distance   Name
-----------------------------------------------------------------------------------
   14.11     0.0   P        108.32     7.175     34.10    34.10    14.11   = P
```

With #1199 the model parsing gets a bit further along but does not finish.

megies:
This hasn't seen any activity, so I'll bump this to 1.2.0.

megies:
I think it's time to close this

PR #1039, opened 2015-12-16
Deprecation+Reroute needed for many renamed functions

megies:
After merging #999, many renamed functions still lack Deprecation+Reroute like done here: https://github.com/obspy/obspy/commit/9c98852b4c53498c777f6768ddfae8387e95e408#diff-4fe4cf25e2fb8e2572239cc976936717R1613

PR #1224, opened 2016-01-21
Variable TauP Radius

krischer:
This PR enables `obspy.taup` to parse planets with varying radius. I also did some generic cleanup. Tests still pass.

Planet radii and the depth of various discontinuities used to be hard coded in various places. It is now decided during model parsing what the radius of the planet is and where the discontinuities are.

To some extent it can now also parse mars model which have a liquid core which does not work with the current master. There are still some issues with it but I suspect they are largely due to the fact that we currently did not consider that some discontinuities do not exists - mars has no inner core and thus no IOCB for example. So there are still a couple of things to do.

1. For normal tvel files we should check if the radius is Earth's radius and if not raise an error - that file has no named discontinuities so Earth's discontinuities have to be assumed.
2. nd files have named discontinuities so that should be fine. It currently cannot parse files without an IOCB.
3. We should attempt to have optional discontinuities in the code. How does the Java version deal with this?

A couple of test files if someone wants to play around with it:

* https://dl.dropboxusercontent.com/u/3866312/mars.tvel  (Can be parsed an something can be calculated but the discontinuities are wrong in the TauP model)
* https://dl.dropboxusercontent.com/u/3866312/anderson_mars.nd  (Cannot yet be parsed)

Please don't merge yet - I have to rewrite the history as I accidentally committed the `.npz` files a couple of times...


obspy/taup/data/1066b.nd
```diff
@@ -160,4 +160,4 @@ inner-core
  6257.10      11.277     3.549    13.064  183300.0    3666.0
  6295.10      11.279     3.550    13.069  183300.0    3666.0
  6333.00      11.280     3.550    13.074  183300.0    3666.0
- 6371.00      11.281     3.550    13.077  183300.0    3666.0
+ 6370.98      11.281     3.550    13.077  183300.0    3666.0
```
QuLogic: Was this wrong, or just a test?

QuLogic: OK, I guess from the commit message, it was wrong...

obspy/taup/slowness_layer.py
```diff
@@ -314,10 +313,16 @@ def create_from_vlayer(vLayer, isPWave, radiusOfEarth=6371,
     ret['botDepth'] = vLayer['botDepth']
     waveType = ('p' if isPWave else 's')
     if isSpherical:
-        ret['topP'] = (radiusOfEarth - ret['topDepth']) / \
-            evaluateVelocityAtTop(vLayer, waveType)
-        ret['botP'] = (radiusOfEarth - ret['botDepth']) / \
-            evaluateVelocityAtBottom(vLayer, waveType)
+        ret['topP'] = (radius_of_planet - ret['topDepth']) / \
+                       evaluateVelocityAtTop(vLayer, waveType)
```
QuLogic: This term isn't actually within the parentheses above.

obspy/taup/tau.py
```diff
@@ -291,20 +291,20 @@ class TauPyModel(object):
...
 
         :param model: The model name. Either an internal TauPy model or a
             filename in the case of custom models.
-        :param earth_flattening: Flattening parameter for Earth's ellipsoid
-            (i.e. a-b/a, where a is the semimajor equatorial radius and b is
-            the semiminor polar radius). A value of 0 (the default) gives a
-            spherical Earth. Note that this is only used to convert from
-            geographical positions (source and receiver latitudes and
+        :param planet_flattening: Flattening parameter for the planet's
+            ellipsoid (i.e. a-b/a, where a is the semimajor equatorial radius
```
QuLogic: I think it should be (a-b)/a.

obspy/taup/tau_model.py
```diff
@@ -25,26 +25,14 @@ class TauModel(object):
...
+    def __init__(self, sMod, radius_of_planet, is_spherical=True, debug=False,
+                 skip_calc=False):
         self.debug = debug
-        self.radiusOfEarth = 6371.0
-        # True if this is a spherical slowness model. False if flat.
-        self.spherical = spherical
+        # Depth for which tau model as constructed.
+        self.source_depth = 0.0
+        self.radius_of_planet = radius_of_planet
+        # True if this is a is_spherical slowness model. False if flat.
```
QuLogic: Not sure if the `is_` is necessary in the variable, but it's definitely not needed in this comment.

obspy/taup/tau_model.py
```diff
@@ -60,6 +48,16 @@ def __init__(self, sMod, spherical=True, debug=False, skip_calc=False):
 
         self.sMod = sMod
 
+        # Branch with the source at its top.
+        self.sourceBranch = 0
+        # Depths that should not have reflections or phase conversions. For
+        # instance, if the source is not at a branch boundary then
+        # noDisconDepths contains source depth and reflections and phase
+        # conversions are not allowed at this branch boundary. If the source
+        # happens to fall on a real discontinuity then then it is not
```
QuLogic: Two 'then's.

obspy/taup/tau_model.py
```diff
@@ -87,7 +85,7 @@ def calcTauIncFrom(self):
         rayNum = 0
         minPSoFar = self.sMod.SLayers[0]['topP']
         tempRayParams = np.empty(
-            2 * self.sMod.getNumLayers(False) + len(self.sMod.criticalDepths))
+                2 * self.sMod.get_num_layers(False) + len(self.sMod.criticalDepths))
```
QuLogic: Why indent?

obspy/taup/taup_pierce.py
```diff
@@ -29,7 +29,7 @@ def depth_correct(self, depth, receiver_depth=None):
         Override TauP_Time.depth_correct so that the pierce points may be
         added.
         """
-        tModOrig = self.model
+        tau_modelOrig = self.model
```
QuLogic: `orig_tau_model` would be better (I assume this was search&replace though.)

obspy/taup/data/1066b.nd
krischer: Yea it was wrong - that's how I initially realized the error. I originally "fixed" the model as 200 meters just don't matter for the travel times. Now it is correct though.

krischer:
I implemented all the suggestions and fixed a lot of PEP8 errors from my search&replace operations...

QuLogic:
Have you removed the extra `.npz` files?

krischer:
Ok. I rewrote the history and now only the last commit contains the recreated `npz` files.

krischer:
All the `.npz` files eat almost 1 MB in the repository so I if we need to do something else that triggers a recompilation of the `.npz` files we should really do it before we merge this into the master.

trichter:
Maybe [git-lfs](https://git-lfs.github.com/) is an option if later a recompilation will become necessary.

krischer:
> Maybe git-lfs is an option if later a recompilation will become necessary.

That might indeed be a good idea if at one point we really require large files. It might ease some things - right now we have to be very careful to not create large files in the repository. Two thing that might make it a bit tricky as of now:

* Everyone cloning ObsPy will have to install git-lfs
* I think the monthly free download quota is only 1 GB

trichter:
I see. Then it's not that attractive anymore.

krischer:
Can we merge this? I know its not yet fully correct but its better than before. And there is still #1200 to keep track of failing mars models.

Renaming things #1039 would repeat some things done in this PR and merging afterwards would be a pain.

QuLogic:
Sorry, I meant to merge this one earlier!

PR #1226, opened 2016-01-22
ResourceIdentifier: compare equal to string with its ID

megies:
I think `ResourceIdentifier` objects should compare equal to a string that matches their internal ID string, right now they only compare equal to other RID objects:

screenshot from 2016-01-22 12 38 34


krischer:
travis takes sooooooo long. I think I fixed a small issue with hashes used in dictionaries and sets. I guess Python somehow uses the `__eq__` as well as the `__hash__` methods for the keys...

PR #1216, opened 2016-01-22
PPSD.get_percentile does not automatically call PPSD.calculate_histogram

megies:
This piece of code does not work

```
ppsd = PPSD.load_npz(filename)
period, median = ppsd.get_percentile()
TypeError: 'NoneType' object is not iterable
``` 

An additional call of ppsd.calculate_histogram() before the last line solves the problem. Should this be done automatically?

### DIFF
--- a/obspy/signal/spectral_estimation.py
+++ b/obspy/signal/spectral_estimation.py
@@ -609,18 +609,22 @@ class PPSD(object):
 
     @property
     def current_histogram(self):
+        self.__check_histogram()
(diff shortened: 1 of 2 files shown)


megies:
> Can you make this issue into a PR?

Done.

trichter:
From my point of view this is ready for merging.

megies:
Looks good to me, too. :+1: 
Thanks for the fix! :ship:

PR #1229, opened 2016-01-23
Add Vincenty direct to geodetics

andreww:
No rush to merge this - as it is probably worth using it somewhere first - but I think we have other changes pending in geodetics so it may be worth keeping track of the various bits.

### DIFF
--- a/obspy/geodetics/base.py
+++ b/obspy/geodetics/base.py
@@ -200,6 +200,104 @@ matplotlib/files/matplotlib-toolkits/basemap-0.9.5/
     return dist, alpha12, alpha21
 
 
+def calc_vincenty_direct(lat, lon, azimuth, distance, a=WGS84_A, f=WGS84_F):
+    """
+    Vincenty Direct Solution of Geodesics on the Ellipsoid.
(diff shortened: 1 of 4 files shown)


QuLogic:
There are spaces missing around several operators.

obspy/geodetics/base.py
```diff
@@ -200,6 +200,104 @@ def calc_vincenty_inverse(lat1, lon1, lat2, lon2, a=WGS84_A, f=WGS84_F):
...
+    A = 1 + (u_2 / 16384)*(4096 + u_2 * (-768 + u_2 * (320 - 175 * u_2)))
+    B = (u_2/1024)*(256 + u_2*(-128+u_2*(74-47*u_2)))
+    sig_0 = distance/(r_minor*A)
+    sig = sig_0
+
+    epsilon = 1E-12
+    iterlimit = 100
+    sig_old = 0.0
+    try:
+        while (math.fabs(sig-sig_old) > epsilon):
```
QuLogic: Unnecessary parentheses.

PR #1228, opened 2016-01-28
test_waveform_plugins incompatible with custom plugins

trichter:
Custom I/O plugins register differently than native obspy.io plugins. E.g. obspyh5 registers itself with:
```
ENTRY_POINTS = {
    'obspy.plugin.waveform': ['H5 = obspyh5'],
    'obspy.plugin.waveform.H5': [
        'isFormat = obspyh5:is_obspyh5',
        'readFormat = obspyh5:readh5',
        'writeFormat = obspyh5:writeh5']}
```
The problem: test_isFormat assumes plugins to be a submodule of some other module.
Besides, plugins have to accept a `byteorder` keyword to get the  test_readAndWrite() pass.
Here is the output of `python test_waveform_plugins.py` after installing h5py and obspyh5 (`pip install obspyh5`):

```
ERROR: test_isFormat (__main__.WaveformPluginsTestCase)
  File "test_waveform_plugins.py", line 170, in test_isFormat
    if f.module_name.split('.')[1] !=

ERROR: test_readAndWrite (__main__.WaveformPluginsTestCase)
  File "test_waveform_plugins.py", line 79, in test_readAndWrite
    tr.write(outfile, format=format, byteorder=byteorder)
  File "/home/richter/dev/obspy/obspy/core/trace.py", line 891, in write
    Stream([self]).write(filename, format, **kwargs)
  File "/home/richter/dev/obspy/obspy/core/stream.py", line 1427, in write
    write_format(self, filename, **kwargs)
  File "/home/richter/dev/obspyh5/obspyh5.py", line 193, in writeh5
    trace2group(tr, group, headonly=headonly, **kwargs)
  File "/home/richter/dev/obspyh5/obspyh5.py", line 225, in trace2group
    trace.data.dtype, **kwargs)
  File "/home/richter/anaconda/envs/obspydev/lib/python2.7/site-packages/h5py/_hl/group.py", line 103, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
TypeError: make_new_dset() got an unexpected keyword argument 'byteorder'
```

### DIFF
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -56,7 +56,8 @@ class WaveformPluginsTestCase(unittest.TestCase):
             if format in ['SEGY', 'SU', 'SEG2']:
                 continue
             for native_byteorder in ['<', '>']:
-                for byteorder in ['<', '>', '=']:
+                for byteorder in (['<', '>', '='] if format in
+                                  WAVEFORM_ACCEPT_BYTEORDER else [None]):
                     if format == 'SAC' and byteorder == '=':
                         # SAC file format enforces '<' or '>'
                         # byteorder on writing
@@ -77,11 +78,11 @@ class WaveformPluginsTestCase(unittest.TestCase):
                     # create waveform file with given format and byte order
                     with NamedTemporaryFile() as tf:
                         outfile = tf.name
-                        if format in WAVEFORM_ACCEPT_BYTEORDER:
+                        if byteorder is None:
+                            tr.write(outfile, format=format)
+                        else:
                             tr.write(outfile, format=format,
                                      byteorder=byteorder)
-                        else:
-                            tr.write(outfile, format=format)
                         if format == 'Q':
                             outfile += '.QHD'
                         # read in again using auto detection
@@ -146,8 +147,6 @@ class WaveformPluginsTestCase(unittest.TestCase):
                         self.assertEqual(st[0].id, ".MANZ1..EHE")
                     elif format not in ['WAV']:
                         self.assertEqual(st[0].id, "BW.MANZ1.00.EHE")
-                    if format not in WAVEFORM_ACCEPT_BYTEORDER:
-                        break
 
     def test_isFormat(self):
         """
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -166,9 +166,9 @@ class WaveformPluginsTestCase(unittest.TestCase):
         paths = {}
         all_paths = []
         for f in formats:
-            sep = os.path.sep
-            path = f.dist.location + sep + f.module_name.replace('.', sep)
-            path = path.rsplit(sep, 1)[0] + sep + 'tests' + sep + 'data'
+            path = os.path.join(f.dist.location,
+                                *f.module_name.split('.')[:-1])
+            path = os.path.join(path, 'tests', 'data')
             all_paths.append(path)
             if os.path.exists(path):
                 paths[f.name] = path
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -166,9 +166,9 @@ class WaveformPluginsTestCase(unittest.TestCase):
         paths = {}
         all_paths = []
         for f in formats:
-            path = f.dist.location + '.' + f.module_name
-            path = path.rsplit('.', 1)[0] + '.tests.data'
-            path = path.replace('.', os.path.sep)
+            sep = os.path.sep
+            path = f.dist.location + sep + f.module_name.replace('.', sep)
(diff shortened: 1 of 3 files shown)


trichter:
Mhm, I am still having a problem with my plugin:

```
  File "test_waveform_plugins.py", line 131, in test_readAndWrite
    self.assertEqual(st[0].data.dtype.byteorder, '=')
AssertionError: '>' != u'='
```

I left the loop over native_byteorder intact for all plugins. For `native_byteorder='<'` the test passes. For `native_byteorder='>'` the above error pops up. If dtype is not specified h5py seems to use 'native' byteorder (of the data?):

```
        # Validate dtype
        if dtype is None and data is None:
            dtype = numpy.dtype("=f4")
        elif dtype is None and data is not None:
            dtype = data.dtype
        else:
            dtype = numpy.dtype(dtype)
```

Does this mean that byteorder is 'preserved' for H5 similar as in SAC?
Is it OK to do:

```
                    # check byte order
                    if format == 'SAC':
                        # SAC format preserves byteorder on writing
                        self.assertTrue(st[0].data.dtype.byteorder
                                        in ('=', byteorder))
                    elif format == 'H5':
                        self.assertTrue(st[0].data.dtype.byteorder
                                        in ('=', native_byteorder))
                    else:
                        self.assertEqual(st[0].data.dtype.byteorder, '=')
```

trichter:
I made `test_isFormat` fail on travis by adding the line

```
self.assertTrue(len(paths) > 0)
```

which ensures that actually something is tested in the method.

I don't know what's wrong on travis, maybe `f.dist.location` or `f.module_name` of the EntryPoint object are not working as expected on travis?

krischer:
I am sorry that this test is such a mess. Maybe it should be cleaned up? 

> I left the loop over native_byteorder intact for all plugins. For native_byteorder='<' the test passes. For native_byteorder='>' the above error pops up. If dtype is not specified h5py seems to use byteorder of the data:

Does this test actually work now with your plug-in installed. Somewhere in the loop the file is read again and the byteorder is asserted to be the native byte-order. You state that your plug-in preserves the byteorder so this should fail I think.


Nontheless: I actually strongly think that we should not make ObsPy tests dependent on external plug-ins as we have no control over these. So i would propose to run this test only for plug-ins included in ObsPy - anything else is not our responsibility and should be tested in the plug-in. A simple

```python
if module.name in DEFAULT_MODULES:
    continue
```

or something along these lines should do the trick.

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -155,6 +162,18 @@ def test_isFormat(self):
         ]
         formats_ep = _get_entry_points('obspy.plugin.waveform', 'isFormat')
         formats = list(formats_ep.values())
+        # Get all the test directories.
+        paths = {}
+        all_paths = []
+        for f in formats:
+            sep = os.path.sep
+            path = f.dist.location + sep + f.module_name.replace('.', sep)
+            path = path.rsplit(sep, 1)[0] + sep + 'tests' + sep + 'data'
```
krischer: Assuming we want to merge this: Can you replace the previous couple of lines and use `os.path.join`, `os.path.dirname`, and consorts? That should be a bit safer for funny path names and other issues.

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -141,6 +146,8 @@ def test_readAndWrite(self):
                         self.assertEqual(st[0].id, ".MANZ1..EHE")
                     elif format not in ['WAV']:
                         self.assertEqual(st[0].id, "BW.MANZ1.00.EHE")
+                    if format not in WAVEFORM_ACCEPT_BYTEORDER:
```
krischer: Assuming we want to merge this: Instead of breaking the loop I would prefer it if the byteorder list that is being looped over would be dependent on the `WAVEFORM_ACCEPT_BYTEORDER` flag. That would be easier to reason about and the test is hard enough to understand in any case, e.g.

```python
for native_byteorder in ['<', '>']:
    for byteorder in ['<', '>', '='] if format in WAVEFORM_ACCEPT_BYTEORDER else [None]:
        ...
```

trichter:
Your suggestions are very valid.

The test is still failing for H5 at the mentioned place.
Up till now I saw it as a feature that custom waveform plugins are tested, but probably you are right and its better to not test them.

trichter:
Two questions:

1. Is there consensus to not test external/custom plugins?
2. Will this PR be merged or does anybody want to tackle a more general clean-up? Changes up to now: Waveform data files are looked for in the new directory structure. Plugins are not forced to accept byteorder keyword.

krischer:
I'll review it again a bit later this day.

I think we still want most of the things in this PR. Can you also add a check that excludes testing on external plugins? I'm fairly sure we don't want to test these for a number of reasons.

krischer:
Looks good to me :-) Thanks for this!

PR #1232, opened 2016-01-31
More HTTPS links

QuLogic:
Convert a lot of links to HTTPS. Fix a few that redirect to point directly to their current location. Also, 2->3 for Python library documentation.

I just want to make sure Travis/AppVeyor build correctly with the updated links.


QuLogic:
Ah, good, it's passing now; I don't understand what was going wrong with that other test. We should probably look into it at some point.

PR #1235, opened 2016-02-03
add helper routine to read nlloc scatter files

megies:


claudiodsf:
Nice! :+1:

megies:
Not a big deal, but I had it lying around anyway..

obspy/io/nlloc/tests/test_util.py
```diff
@@ -0,0 +1,60 @@
...
+    def test_read_nlloc_scatter(self):
+        """
+        Test reading NLLoc scatter file
+        """
+        filename = os.path.join(self.datapath, "nlloc.scat")
+        got = read_nlloc_scatter(filename)
+        filename = os.path.join(self.datapath, "nlloc_scat.npy")
+        expected = np.load(filename)
+        np.testing.assert_array_equal(got, expected)
+        filename = os.path.join(self.datapath, "nlloc.scat")
```
QuLogic: I'd just add a newline above here to split the two sections up.

obspy/io/nlloc/util.py
```diff
@@ -0,0 +1,59 @@
...
+    :type filename: str
+    :param filename: Filename with NonLinLoc scatter.
+    :type coordinate_converter: func
+    :param coordinate_converter: Function to convert (x, y, z)
+        coordinates of NonLinLoc output to geographical coordinates and depth
+        in meters (longitude, latitude, depth in kilometers).
+        If left `None` NonLinLoc (x, y, z) output is left unchanged (e.g. if
+        it is in geographical coordinates already like for NonLinLoc in
+        global mode).
+        The function should accept three arguments x, y, z and return a
```
QuLogic: Maybe note that these are arrays, not singular values.

obspy/io/nlloc/tests/test_util.py
megies: Actually, I'll split it into two tests because only one should be skipped without pyproj..

obspy/io/nlloc/util.py
megies: Good point. :+1:

PR #1236, opened 2016-02-05
ray path plotting routine

MMesch:
This is probably another longer duration pull request. It adds functions for greatcircle path plotting in 2D and 3D to obspy, based on geographiclib and taup. It is already working but needs to be structured nicely. Here is an example, plotted again with mayavi (green = P, blue = Pdiff, red = PKP), but plotting could be done with matplotlib and vtk file output again. Right now it works for one event and and inventory, but I think now it is easy to expand it to a catalog and an inventory together.

3d_paths

+DOCS


QuLogic:
Perhaps you have missed it (likely because of our slow release rate), but there is already a plotting routine in [`TauPyModel`](https://github.com/obspy/obspy/blob/master/obspy/taup/tau.py#L96) (though not in 3D) as well as a function to [get ray paths via geographic coordinates](https://github.com/obspy/obspy/blob/master/obspy/taup/tau.py#L510).

Expanding the existing stuff to 3D and to handle an entire `Catalog`/`Inventory` would certainly be interesting.

MMesch:
ok I use the taup function now and put the subroutines directly into the inventory class. I'll add some non-mayavi plotting options later...

3d_paths

krischer:
Looks great :-)

In terms of code structure I think it would be much better to but all plotting code into `obspy.imaging` and just add thin wrappers to the inventory class. If we start to add fancy plotting to all core objects it will really bloat the core module. The same might be good for the moment tensor plotting things in the other PR.

```python

class Inventory(...):
    ...

    def plot_rays(self, *args, **kwargs):
        from obspy.imaging.inventory import plot_rays
        plot_rays(inventory=self, *args, **kwargs)
```

The `plot_rays()` function should probably also take event objects. The optional event coordinates are just there for convenience.

```python
plot_rays(inventory, event=None, event_latitude=None,
          event_longitude=None, event_depth_in_m=None):
    ...
```

MMesch:
OK agree. I just don't see that these things are really 'imaging' in the sense of tomography. They are more visualization routines. That's why it seems strange to put them in this module.

krischer:
Haha...i did not realize that confusion. `obspy.imaging` actually is the visualization module of ObsPy ;-) I'm so used to the name I never saw that the name is ambiguous in seismology...

MMesch:
ok I'll try to see it as visualization then :)

MMesch:
looks pretty cool now if I may say so :)

3d_paths

pkikp

ppp

pdiff

pcp

pkikp_bright

megies:
Bright color scheme is much better to see for me, at least in the examples above..

QuLogic:
It looks like you're creating your own colour scheme by hand. That should not be necessary. Since we have a hard dependency on matplotlib, you can just pull the colour map from there. Even better would be to use [the existing colour set](https://github.com/obspy/obspy/blob/master/obspy/taup/tau.py#L27).

Also, it seems you have not configured your git client with your email address, is there a reason you have not done so? The first commit here is attributed to `<matthias@matthias-Latitude-E6520.(none)>` and the rest to `<MMesch@users.noreply.github.com>`. It is okay if you do not wish to expose your email address, but it would be best if a consistent address were used.

MMesch:
ok. I use the taup colors now but I have to change the lightness and saturation because it is essential to see something in the 3d plots.

sorry about my email address, I'll try not to change it anymore!

krischer:
I pushed some small things to make it a bit easier to use. Looks really cool :-)

 A couple notes:

* Diffracted phases need to be interpolated as they only store two points along the whole diffracted part of the phase which results in a straight line in xyz which is wrong. This probably has to go into the geographic ray calculation routine.
* The colors don't seem to work for me - I'll look into that.
* Mayavi as a whole seems very unstable - it keeps crashing randomly for me.

krischer:
Considering that Mayavi is very unstable for me I played around a bit and maybe we can do this completely in WebGL. The nice thing is that the only dependency is a web browser which everyone has and it would even work in the notebooks. Also the render quality seems to be higher (I cannot get mayavi to render with antialiasing) and it is probably a bit easier to add interactivity.

Super unpolished toy example (12 MB of data...so you might have to wait a second): http://www.geophysik.uni-muenchen.de/~krischer/raypaths/

claudiodsf:
@krischer Wow! That is impressive! (except for the earthquake at north pole :). Which tool do you use to generate WebGL?

krischer:
Yea...the coordinate system for the rays (or the textures) has to be rotated so its not fully correct as of now ;-)

This uses three.js: http://threejs.org/

MMesch:
https://youtu.be/_4YiHAO70KA

MMesch:
@krischer great thing!!! This will certainly attract some people to the lmu server :)

MMesch:
also hopefully to obspy

MMesch:
my mayavi is absolutely stable btw. Maybe it is an issue of your graphic cards driver?

ThomasLecocq:
@MMesch WOOOOW amazing!

MMesch:
Another though (sorry about all the mini posts):

The Pdiff issue should be solved in the taup module directly in my opinion

krischer:
Great movie :-)

> @krischer great thing!!! This will certainly attract some people to the lmu server :)

I'll give this a real shot then. I'll try to mimik your plot and the website will just be output of `inv.plot_rays()`. A really cool side effect is that it works on Python 3, has zero dependencies, and could also be embedded in the notebooks.


> my mayavi is absolutely stable btw. Maybe it is an issue of your graphic cards driver?

Hmm...who knows :-( I have a macbook with the same graphic card as ten million other people and it usually works fine enough.


> The Pdiff issue should be solved in the taup module directly in my opinion

I'm not sure - it is not wrong in the spherical coordinate systems. The plotting just forces a cartesian interpolation which is wrong. But it has to happen before the geographical routines take over, otherwise we have to deal with circular longitudes and warp around effects...

MMesch:
The right thing would be to do a spherical interpolation in taup_geo before it is passed to geographiclib then.

Also the complete mayavi scene can be composed as pure vtk files. It is then independent from any instabilities and can be displayed e.g. with paraview.

I'm not sure what limitations you might encounter with pure OpenGL. I guess since we use very basic geometries only (lines, spheres, cones), it shouldn't be to difficult to rewrite it. It would probably be more stable but we wouldn't have lights or the whole graphical user interface of mayavi.

webGL on the other hand really adds more fuctionality, so I think it is worth it!

MMesch:
One more thing:

the taup_geo routine seems to be somewhat inaccurate for PKIKP phases. E.g. in the video that I have posted, rays that arrive at stations that are close to the antipode do not exactly converge at the station longitude latitude.

I am not sure why this is the case. Either a taup problem, or maybe the ellipticity is not treated correctly ?

megies:
Hmm.. a general comment regarding the ray path plots:

To me those plots seem incredibly overcrowded, I wouldn't know where to start trying to interpret those.. I am not a tomo-guy but I guess this is mostly about knowing where and which rays hit some zone of interest (e.g. CMB). Wouldn't it make more sense to do e.g. a 2D (or 3D) ortho map-plot and plot into it the pierce points, labelled by e.g. "{event_id}:{station_code}", or do a heatmap of pierce points (to visualize sensitivity of the data set regarding some region and layer of interest)?

MMesch:
Hi Tobias,

1. Don't worry, these plots are using the maximum of phases/stations/events
possible to put into a single plot. Less paths are way better to see and
look really great (I'm preparing a video).

2. I think an ortho map would be really nice to have because it's easier to
work with it. It is less eye candy/fancy though. The plot right now is more
for educational purposes and visual impression. There is no reason why we
shouldn't have both.
Am 09.02.2016 13:57 schrieb "Tobias Megies" <notifications@github.com>:

megies:
> There is no reason why we shouldn't have both.

Yeah, that's absolutely right of course.

QuLogic:
I'm not saying it'll be any easier, but maybe you should try [VisPy](https://vispy.org/) too? It supports many backends including WebGL and the notebook without need for any JS (hopefully). Unfortunately, it's not quite as mature as matplotlib, so you definitely need to use a git clone.

MMesch:
here is a longer video of all kinds of phases:

https://youtu.be/GMrZkqxVc5Q

ThomasLecocq:
Wooooow !!!!

MMesch:
@QuLogic  btw: do you know the difference between Vispy and Glumpy ? https://glumpy.github.io/ 

They seem to have the same website :)

krischer:
Vispy is an effort to combine the developments of a couple of other visualization packages: glumpy, pyqtgraph, galry, and some others I probably forgot. Its still in a pretty early stage but there was a nice talk at last year's scipy: https://www.youtube.com/watch?v=_3YoaeoiIFI

megies:
Nice video! :+1:

MMesch:
I am just thinking how to continue. I structure the plotting routines of inventory and catalog better and to include the ray path plotting routine, I will add an 'get_all' function to inventory and catalog. This could be called as:

    long_list = catalog.get_all(['latitude', 'longitude', 'depth', 'focal_mechanism'])

this would replace the current
```
for network in inventory:
    for station in network:
         lats.append(station.latitude)
         [...]
```
 loops in the plotting functions and is also useful for other functions that need bulk inventory/catalog data. Tell me if you have a better idea...

krischer:
We could also do a slightly more generic variant:

```python
catalog.get_all(('preferred_origin.latitude', 'preferred_origin.longitude',
                 'preferred_focal_mechanism.mt.tensor', ...))
```

Which would return lists of dictionaries with the requested keys. I'm not sure if we need the `preferred_` prefix or not but I think it might be useful. For inventories this is a bit more complex as they are nested. We could also think to add a more generic query interface that would replace the current `filter()` methods. I did something along these lines here which I think has quite a nice interface: http://seismicdata.github.io/pyasdf/querying_data.html

I also want to add a couple of more things. I'll do it after the release:

* The WebGL renderer for the raypaths.
* Plots like in the middle of this page: http://www.isc.ac.uk/standards/phases/

megies:
> I did something along these lines here which I think has quite a nice interface: http://seismicdata.github.io/pyasdf/querying_data.html

Just a quick glimpse but I like it way more than our current hacky catalog filter or list-of-strings controlled selection criteria..

MMesch:
ok that looks much nicer. Any chance that you can directly transfer this ```ifilter``` method to obspy? I think the problem with catalog and inventory is that they are deeply nested structures which makes it difficult to find stuff if you are not an expert.

MMesch:
another question:  Is there any way to refer to a specific station and a specific event inside an inventory or catalog? E.g. for the path routine, one could just return the id of the event and station for each path instead of specific latitude/longitudes and the plotting routines will extract whatever they need.

megies:
> Is there any way to refer to a specific station and a specific event inside an inventory or catalog?

there are some convenience methods (e.g. http://docs.obspy.org/packages/autogen/obspy.core.inventory.inventory.Inventory.get_coordinates.html#obspy.core.inventory.inventory.Inventory.get_coordinates  -- which should be improved to use wildcards..) but they could certainly expanded/improved. I think in general the first thing that should be implemented are these kind of getter routines attached to the top-level object (inventory, catalog, ..).

MMesch:
update: resampling of Pdiff along the cmb works well now. The feature film is ready https://youtu.be/vZ80K3i4t3w .

3D plotting is done from my side. I'll continue with 2D ...

megies:
Looks like this'll need a rebase at some point.. otherwise CI won't be able to run.

MMesch:
ok. I think this is a good moment. I'll put it into a feature_ branch then.

Btw: there was another bug in the taup cache function that had to be fixed, in case somebody else encounters this...

MMesch:
ps @megies for a rebase, I could pull out the two bugs and the resampling method to merge them independently. What do you think?

krischer:
Yes! Please make two new PRs out of this: one for each bug and leave this one to only deal with the visualization.

megies:
> Yes! Please make two new PRs out of this: one for each bug and leave this one to only deal with the visualization.

And remember to base the bug fixes on maintenance_1.0.x (and open PRs against it) and this one again on master.

MMesch:
ok I think I messed everything up :) I merged maintenance into gc_paths. Now I can't rebase onto the new master anymore. HELP anyone!?

megies:
When on your branch, you could `git reset --hard 364e412` to get back to the state before merging `maintenance` (this would discard any local, uncommited changes!) and then continue with rebasing on current master. I tried but I get some merge conflicts of the stuff you did, so didn't know what to do with them exactly.

In any case, you could do a backup branch before everything else.. `git checkout -b gc_path_bkp;git checkout -b gc_path `. ;-)

MMesch:
@megies  I have rebased this on the current master I guess I have to open a new PR now?

megies:
> @megies I have rebased this on the current master I guess I have to open a new PR now?

No need, you can simply update this PR by doing a force-push to `MMesch:gc_paths`.

MMesch:
OK. It should be fine then. Any idea how I can see the conflicts? I don't understand why there are conflicts in any case. In case it is to complicated I can just start a new clean branch and re-implement everything with copy paste, which should be quick...

megies:
> @megies I have rebased this on the current master I guess I have to open a new PR now?

If you've really rebased on current master there can be no conflicts.. ;-)

obspy/taup/tau.py
```diff
@@ -519,7 +519,7 @@ def get_pierce_points_geo(self, source_depth_in_km, source_latitude_in_deg,
                   "Arrivals object will not be modified. " + \
                   "Install the Python module 'geographiclib' to solve " + \
                   "this issue."
-            raise ImportError(msg)
+            warnings.warn(msg)
```
MMesch: In my opinion, this should raise a true Error. If someone calls the geo routine, he wants the geographical points. Otherwise he could just use the other routine. Another fallback should be somewhere else.

megies:
Appveyor fail is due to missing geographiclib, I think. No idea why we don't install it in Appveyor, it's pure Python and should be no problem. Maybe you can try adding `pip install geographiclib` to `appveyor.yml`?

MMesch:
@megies  I guess now it is a network issue!?

megies:
> @megies I guess now it is a network issue!?

Yes, all subdomains of `obspy.org`, were offline earlier. Seems they're up again, I'll restart CI.

MMesch:
image comparison works fine on travis using this image:

ray_paths

I have no idea why the sds filesystem test failed. Some other necessary changes were:

* running a virtual xserver on travis (required by mayavi offscreen rendering)
* change ImageComarison to accept images without alpha channel

Otherwise this is ready to be merged after maybe some little beauty modifications. I would like to write a little entry for the tutorials as well. I just don't know if it is possible for sth that runs under python2.7 only for the moment !?

megies:
> I have no idea why the sds filesystem test failed.

Looking at the changes, no clue here either. I'll try to look into it when I find some time..

2) How much detail on those stations do you need for the test (new file `obspy/imaging/tests/data/IU.xml`)? The file is >300k.. would the same information in format `STATIONTXT` be enough? That would be way smaller..

> I would like to write a little entry for the tutorials as well. I just don't know if it is possible for sth that runs under python2.7 only for the moment !?

Sure.. just put a `..note::` box saying that mayavi is only available for Py2 for now (with link to that Py3 PR you mentioned)

> image comparison works fine on travis using this image:

Cool image. Looks very :octocat:-y

obspy/imaging/ray_paths.py
```diff
@@ -0,0 +1,594 @@
...
+    """
+    Ray path plotting routine
+
+    :param inventory: an obspy station inventory
+    :param catalog: an obspy event catalog
+    :param station_latitude: station latitude (alternative to inventory)
+    :param station_longitude: station longitude (alternative to inventory)
+    :param event_latitude: event latitude (alternative to catalog)
+    :param event_longitude: event longitude (alternative to catalog)
+    :param event_depth_in_km: event depth in km (alternative to a catalog)
```
megies: Having these parallel options is a bit ugly.. wouldn't it be OK to expect either inventory/station input and catalog/origin? It's not a lot of overhead to the user to put lat/lon/depth into an Origin..

megies:
The test inventory has a lot of duplicated stations (multiple epochs).. and you're only using the lat/lon.. so how about using this (gzipped) for the test: http://service.iris.edu/fdsnws/station/1/query?network=IU&level=station&format=text&starttime=2016-01-01&endtime=2016-01-02

MMesch:
gzip doesn't work on stationtxt. This could be fixed in a different issue

obspy/imaging/ray_paths.py
MMesch: agree. It also keeps the code behind cleaner. There needs to be an easy example on how to generate a single station inventory and a single event catalog from scratch... Do we have something like this?

megies: http://docs.obspy.org/master/packages/autogen/obspy.core.event.origin.Origin.html :-)
Station doesn't have it though.. and come to think of it, it also would actually need a Network around it..  :-(

megies:
Since these mayavi plots are mostly for interactive use.. and basically work on event/inventory pairs.. maybe it would be worth a thought to offer a command line interface to this like..

 `$ obspy-plot-raypath3d --phase Pdiff my_catalog.xml my_inventory.xml` ?

krischer:
I'm all for more command line interfaces as lots of people want them but we are starting to clobber the `bin` directory.

How about an `obspy-cli` command that just delegates to the other commands similar to how git works?

```bash
$ obspy-cli seed2resp ...
$ obspy-cli plot-raypath3d ...
$ obspy-cli plot ...
$ obspy-cli print ...
...
```

That gives us a clean way to add a lot of commands. We can either use the `click` library which is really nice, or (if we want to avoid additional dependencies roll our own). I already have some code here: https://github.com/krischer/LASIF/blob/master/lasif/scripts/lasif_cli.py#L1561-L1646

MMesch:
sounds good. "obspy-cli" won't be very intuitive for many. "obspy-script" or something would be better I think...

claudiodsf:
Why not just `obspy` ?

obspy/imaging/ray_paths.py
MMesch: I hesitate a bit using Origin etc. In my opinion the user should just pass catalogs and inventories. Otherwise it doesn't simplify the function. Catalog and inventory can contain a single station. A simple function is needed to generate such catalogs and inventories from scratch. If no network is given, it can be filled with a __ network that contains stations without networks. Same for the catalog.

```
inv  = inventory()
stat = station(lat=20, lon=30)
inv.append(stat)
```

MMesch:
because @krischer was metioning antialiasing, for completeness the reference to the mayavi issude: https://github.com/enthought/mayavi/issues/183

MMesch:
Update: mayavi anaconda package for python3 exists: https://github.com/enthought/mayavi/issues/384

a priori it should work with  `conda install -c menpo mayavi`

I'll try to test it on travis soon.

megies:
Will this become ready in the next couple of weeks? If not it'll be bumped to 1.2.0..

MMesch:
For me this is ready. Some little beauty modifications maybe, and I don't know whether we can get the image test running well on travis.

MMesch:
It seems like travis can't use the newer vtk library unfortunately. I will just suspend the mayavi offscreen rendering tests...

```
/home/travis/miniconda/conda-bld/work/VTK-7.0.0/Rendering/OpenGL2/vtkOpenGLRenderWindow.cxx, line 545
vtkXOpenGLRenderWindow (0x5b1f620): GL version 2.1 with the gpu_shader4 extension is not supported by your graphics driver but is required for the new OpenGL rendering backend. Please update your OpenGL driver. If you are using Mesa please make sure you have version 10.6.5 or later and make sure your driver in Mesa supports OpenGL 3.2.
```

obspy/core/inventory/inventory.py
```diff
@@ -666,6 +666,15 @@ def plot(self, projection='global', resolution='l',
...
         return fig
 
+    def plot_rays(self, *args, **kwargs):
+        """
+        Plot ray paths between this inventory and one or more events.
+
+        All options are passed to :func:`obspy.imaging.ray_paths.plot_rays`.
+        """
+        from obspy.imaging.ray_paths import plot_rays
+        plot_rays(inventory=self, *args, **kwargs)
```
krischer: I would remove this - it is a bit redundant and just using the `plot_rays()` is almost as easy in my opinion. I know that this is how we did things in the past but we should probably try to keep our core classes as small as possible.

obspy/imaging/ray_paths.py
krischer: Can you please remove the `station_*` and `event_*` parameters everywhere?

obspy/imaging/tests/test_ray_paths.py
```diff
@@ -0,0 +1,86 @@
...
+        # by default
+        self.path = os.path.join(os.path.dirname(__file__), 'images')
+        pass
+
+    def test_compute_ray_paths(self):
+        # careful, the full inventory, catalog test is long (1min)
+        # greatcircles = get_ray_paths(
+        #        inventory=self.inventory, catalog=self.catalog,
+        #        phase_list=['P'], coordinate_system='XYZ',
+        #        taup_model='iasp91')
```
krischer: Could not just have the inventory contain two stations and the catalog contain two events? Then everything is tested but it should be pretty quick to execute.

obspy/imaging/tests/test_ray_paths.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # the following test is for an animated mayavi windows
+        # and movie plotting.
+        # Needs third party coastlines by default, otherwise
+        # remove the coastlines keyword !
+        #
+        # plot_rays(inventory=inv, catalog=cat,
+        #           phase_list=['PKP'], animate=True, savemovie=False,
+        #           kind='mayavi', figsize=(1920, 1080),
+        #           coastlines='data/coastlines.vtk')
```
krischer: This comment should probably go into the documentation.

obspy/imaging/tests/test_taup_geo.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
krischer: This contents of this file should be moved to the `test_taup_geo.py` file in the `obspy.taup` package.

obspy/taup/tau.py
```diff
@@ -460,7 +460,8 @@ def get_pierce_points_geo(self, source_depth_in_km, source_latitude_in_deg,
                               source_longitude_in_deg,
                               receiver_latitude_in_deg,
                               receiver_longitude_in_deg,
-                              phase_list=("ttall",)):
+                              phase_list=("ttall",),
+                              resample=False):
```
krischer: Please add a test for the resampling functionality.

obspy/taup/taup_geo.py
```diff
@@ -82,7 +82,7 @@ def calc_dist(source_latitude_in_deg, source_longitude_in_deg,
 def add_geo_to_arrivals(arrivals, source_latitude_in_deg,
                         source_longitude_in_deg, receiver_latitude_in_deg,
                         receiver_longitude_in_deg, radius_of_planet_in_km,
-                        flattening_of_planet):
+                        flattening_of_planet, resample=False):
```
krischer: Please also add a small unit test for this.

krischer:
You also have to rebase once on the master. Sorry this review took so long!

obspy/core/inventory/inventory.py
MMesch: done

obspy/imaging/tests/test_ray_paths.py
MMesch: yes. I wanted to make a nice image for the gallery, but if the test doesn't work on travis anyway because of ancient OpenGL, it is not going to work...

obspy/imaging/tests/test_ray_paths.py
MMesch: yes right

obspy/taup/tau.py
MMesch: the test is the Pdiff ray path plot. Any idea for a better one?

obspy/taup/taup_geo.py
MMesch: any idea for a better test than a plot?

MMesch:
I'm giving up on travis + VTK because of the OpenGL3 issue. The drivers have to be updated or an older VTK version used or sth like that. I'll just deactivate the tests. Should I leave them as a comment? Or do you have any other suggestion?

megies:
> I'll just deactivate the tests. Should I leave them as a comment? Or do you have any other suggestion?

You can leave them in and skip them in CI, e.g. like..
```python

    @unittest.skipIf(os.environ.get('CI') in ('True', 'true'),
                     'Test can currently not be run in Travis/Appveyor, '
                     'due to blahblah reason')
    def test_some_test(self):
        pass

```

Of course if there's an easy way to check if the test should be able to run you can use that instead, e.g. see https://github.com/obspy/obspy/blob/9ce831e8f0d08d06ff8a7af5c53810e3c375aed5/obspy/io/shapefile/tests/test_core.py#L12-L28

 - https://www.appveyor.com/docs/environment-variables/
 - https://docs.travis-ci.com/user/environment-variables/

obspy/taup/tau.py
krischer: Can you not just test some actual values of the path?

MMesch:
@krischer what is left to do? I am a bit lost...

krischer:
I'm also a bit lost ;-) I'm at AGU right now so I likely won't be able to find the time for a proper review this week - but I'll do it next week!

QuLogic:
Travis is fine; it's the docker builds that aren't passing. You can see some of the errors [here](http://tests.obspy.org/63970/#1).

MMesch:
I think this happens if geographiclib is not installed. E.g. latitude is not computed in taup. Didn't you want to include geographiclib fully in obspy? @QuLogic @megies @krischer ? I'll deactivate the ray path routine for now if geographiclib is not there...
```
File "/obspy/obspy/imaging/tests/test_ray_paths.py", line 60, in test_compute_ray_paths
taup_model='iasp91')
File "/obspy/obspy/imaging/ray_paths.py", line 536, in get_ray_paths
thetas = np.radians(90. - arr.path['lat'])
ValueError: field named lat not found.
```

megies:
@MMesch I'm adding geographiclib to docker base images for testing in the future, but maybe you should raise a proper `ImportError` in the ray path plotting routines if optional dependencies necessary for those routines are not installed..

MMesch:
In my opinion, an Exception and not a Warning should be raise here:

https://github.com/obspy/obspy/blob/master/obspy/taup/tau.py#L576-L589

megies:
What's the status here @MMesch? We want to get 1.1.0 ready within days..

MMesch:
Everything is ready since quite some time. From my side it is done. The test all used to work fine but mayavi doesn't run on travis anymore (at least it didn't last time I checked). It is just a matter of integrating it now, which can only be done fully by one of you guys in my opinion.

krischer:
Rebased and force pushed to see what happens. I'm not sure if I'll find the time to travis debugging in the next couple of days. And I don't really feel comfortable merging this without properly executed tests as its a lot code.

MMesch:
Another possibility:

just merge the  resample option in the ray paths taup routine. The rest can be used as an independent function and doesn't have to be in obspy ...

MMesch:
It is so long ago that I coded this that I am forgetting what I did and I just want to put it somewhere in a finished form...

krischer:
> Another possibility:
> 
> just merge the resample option in the ray paths taup routine. The rest can be used as an independent function and doesn't have to be in obspy ...

This sound like a good plan to me - especially because this issue has been kind of stuck for a long time.

krischer:
Are you volunteering to do it? ;-) Otherwise I can also take care of it but I'd probably just squash everything and keep the pieces related to the re-sampling.

MMesch:
You can just copy the resampling function. I think it is best if you do it, because you can immediately merge it into whatever version is appropriate. Otherwise it will lie around again for a long time.

megies:
I'll leave this in for 1.1.0 for now, as it sounds like it's almost ready. Feel free to punt it to 1.2.0 though, @krischer..

megies:
I'm a bit unsure what to do with this PR for 1.1.0.. right now the mayavi plotting and vtk output is untested by CI: https://codecov.io/gh/obspy/obspy/pull/1236/diff#diff-b2JzcHkvaW1hZ2luZy9yYXlfcGF0aHMucHk=

On the other hand this has been pending for so long.. maybe we just show a warning when the new file imaging/ray_paths.py gets used and merge it nevertheless?

CC @krischer @MMesch

krischer:
We should just keep the resampling pieces and move the rest to separate repository, as discussed above. I already promised to do it once so let's see if I keep it this time ;)

megies:
@krischer @MMesch so I've started working on this. As I understand it was kind of decided to remove mayavi related stuff again (to be moved to a separate repository eventually) and will only keep non-mayavi stuff..

MMesch:
Great. The only reason why I didn't do anything on this anymore is because I didn't know what and how :)

At some point all mayavi tests were running, and the documentation was ready, so everything should be in fairly good shape. It is up to you guys whether you want to include mayavi into obspy or whether it is put into an external repository. I am absolutely fine with both! The most important part is to have the resampling routine in obspy taup, because that is just not as nice to externalize. In my opinion, in the long run, a ray path routine that computes ray paths between an inventory and a catalog would be nice to have in obspy as well, but maybe in a different form!? I see that the mayavi stuff is tricky to be included and tested. Maybe it is best to keep these fancy visualizations a separate repository.

megies:
> a ray path routine that computes ray paths between an inventory and a catalog would be nice to have in obspy as well

uhm.. isn't that in there? https://github.com/obspy/obspy/pull/1236/files#diff-6cfba7f5acfbda837e36c85690fcd0fcR123

MMesch:
yes. I would include all of these routines (mayavi included) but I am not in the position to decide and I can understand if you don't want to include all of it because the interface/testing etc. could be discussed and because you might want to keep obspy a small core package :) On the other hand, these routines work well and they can be incrementally adjusted as the other functions in obspy.

megies:
Yeah, we also think those routines are good to have.. I'll do some more cleaning up and then we'll check what else needs tests..

krischer:
Sounds good to me. I think we do want all the ray calculations things but not anything related to mayavi and vtk. This should go elsewhere.

obspy/imaging/tests/test_taup_geo.py
megies: Actually the contents of this file already seem to be in `obspy/taup/tests/test_taup_geo.py` in master.. so I'll also remove it

megies:
well.. the vtk might be interesting to keep, dunno

megies:
@MMesch  @krischer Just wanted to add a test for the resampling.. I'm doing a path for a Pdiff phase and it only adds ~8 points for a path of ~400 points.. is this the expected behavior?



figure_1
figure_1-1

MMesch:
That's a good test! I chose the resampling interval based on my visual
impression of the 3d plots and tried to keep it low. I guess we should
adapt it to the sampling interval of taup!? Any ideas what it is?

Am 11.10.2017 14:43 schrieb "Tobias Megies" <notifications@github.com>:

megies:
Hmm.. @MMesch if there's still open questions regarding the resampling.. and since this seems a pretty small change right now (when leaving away the mayavi plotting for a separate repo anyway).. would you be mad a t us if we punt it to 1.2.0? Feeling a bit guilty for even proposing this, after the long hang-time of this PR already so far.. but.. dunno

MMesch:
I would strongly suggest to include it now, because everything works well as far as I have seen. It is just a matter of choice how fine the sampling should be and not essential to the routine. Could you post the test-code?

megies:
Here's my code:
```python

model = TauPyModel('iasp91')
kwargs = dict(source_depth_in_km=100, source_latitude_in_deg=4.1,
              source_longitude_in_deg=5.1, receiver_latitude_in_deg=52.8,
              receiver_longitude_in_deg=140.4, phase_list=['Pdiff'])

old = model.get_ray_paths_geo(resample=False, **kwargs)[0].path
new = model.get_ray_paths_geo(resample=True, **kwargs)[0].path


def plotit(feature):
    fig, axes = plt.subplots(nrows=3, sharex=True, sharey=True)
    style = dict(ls='-', marker='x', lw=1.4)
    axes[0].plot(old['time'], old[feature], alpha=0.9, color='r', **style)
    axes[2].plot(old['time'], old[feature], alpha=0.5, color='r', **style)
    axes[1].plot(new['time'], new[feature], alpha=0.9, color='g', **style)
    axes[2].plot(new['time'], new[feature], alpha=0.5, color='g', **style)
    axes[-1].set_xlabel('time')
    for ax in axes:
        ax.set_ylabel(feature)
    axes[0].text(0.05, 0.95, 'old: {} points'.format(len(old)), va='top',
                 transform=axes[0].transAxes)
    axes[1].text(0.05, 0.95, 'new: {} points'.format(len(new)), va='top',
                 transform=axes[1].transAxes)
    plt.show()


plotit('lon')
```

So basically a unit test for that missing and we need to decide if we should add the vtk output here or rather include it alongside the mayavi stuff. (if it's supposed to be included the vtk output would need testing as well, obviously).

Are you adding the test?

MMesch:
@megies @krischer On the other hand, I'm not angry if you push it to 1.2.0 though. I wouldn't do it because I think it leads to more work in the long run but that's not my decision . This thread is discussed since ages and nothing essential has changed because it is running since a long time. Either reject completely or accept and stop discussing :)

MMesch:
here the test for the three parameters:
test_depth
test_lon
test_lat


```

model = TauPyModel('iasp91')
kwargs = dict(source_depth_in_km=50, source_latitude_in_deg=0.,
              source_longitude_in_deg=0., receiver_latitude_in_deg=0.,
              receiver_longitude_in_deg=150, phase_list=('Pdiff', ))

old = model.get_ray_paths_geo(resample=False, **kwargs)[0].path
new = model.get_ray_paths_geo(resample=True, **kwargs)[0].path


def plotit(feature):
    fig, axes = plt.subplots(nrows=3, sharex=True, sharey=True)
    style = dict(ls='-', marker='x', lw=1.4)
    axes[0].plot(old['time'], old[feature], alpha=0.9, color='r', **style)
    axes[2].plot(old['time'], old[feature], alpha=0.5, color='r', **style)
    axes[1].plot(new['time'], new[feature], alpha=0.9, color='g', **style)
    axes[2].plot(new['time'], new[feature], alpha=0.5, color='g', **style)
    axes[-1].set_xlabel('time')
    for ax in axes:
        ax.set_ylabel(feature)
    axes[0].text(0.05, 0.95, 'old: {} points'.format(len(old)), va='top',
                 transform=axes[0].transAxes)
    axes[1].text(0.05, 0.95, 'new: {} points'.format(len(new)), va='top',
                 transform=axes[1].transAxes)

plotit('lat')
plotit('lon')
plotit('depth')

plt.show()
```

megies:
Ah OK, that's a better example I guess, because more points are added.. but I guess taup is just thinning out parts with a pretty straight path?

MMesch:
Everything works as expected. When the distance between two adjacent points is larger than 200km, points are added to the ray path. This allows to later to plot them in ```xyz``` coordinates. Feel free to add or not :)

megies:
> This allows to later to plot them in xyz coordinates.

Ah, I think I'm finally getting the whole point..

megies:
I've added a small test for the path resampling, rebased and force pushed.

I remove the vtk output, @MMesch if you offer to add a test for it I would consider keeping it.

Otherwise I'll improve the test "test_compute_ray_paths" maybe in the evening because right now it's hardly testing anything, just asserting that a list with one item is returned.

megies:
@MMesch next time please take some time to implement a real test. This is really not cutting it:
https://github.com/obspy/obspy/pull/1236/files#diff-dd572eba2d259313f2468ec5ae6d42d9R56
screenshot from 2017-10-11 18-06-14

MMesch:
Haha yes. The real test used to be an image comparison. But I agree that it
is better not to have any mayavi/vtk stuff on travis...

Am 11.10.2017 18:06 schrieb "Tobias Megies" <notifications@github.com>:

obspy/taup/ray_paths.py
```diff
@@ -0,0 +1,137 @@
...
+
+                if coordinate_system == 'RTP':
+                    gcircle = np.array([radii, thetas, phis])
+
+                if coordinate_system == 'XYZ':
+                    gcircle = np.array([radii * np.sin(thetas) * np.cos(phis),
+                                        radii * np.sin(thetas) * np.sin(phis),
+                                        radii * np.cos(thetas)])
+
+                value = 0.
```
megies: @MMesch can you elaborate on the significance of always adding 0.0 to the output tuple?

megies: Also it seems like bad design to include spaces for padding in the textual headers of the paths.. this should be done in the plotting part.. (coordinates in data array for path omitted below)

```
[(array([[...]]), u'P', 0.0, u'   NET.STA', u'  2017-10-12 | M7.0')]```

Comment (megies) [reply obspy/taup/ray_paths.py]: I'll also include the event id to the output tuple

Comment (megies) [reply obspy/taup/ray_paths.py]: in the long run this routine could also use the information in the `Event.arrivals` to figure out what rays really exist in the data set.. but I guess in many cases such detailed info is not included in the Event data..

**Comment by MMesch** [comment]
> Also it seems like bad design to include spaces for padding in the textual headers of the paths.. this should be done in the plotting part..

agree that's better

> can you elaborate on the significance of always adding 0.0 to the output tuple

originally, I wanted to leave the routine flexible to add a 'value' like an amplitude from the radiation pattern to the ray path. This can be removed or organized in a better way if you have a better idea!

**Comment by megies** [comment]
I hope this should do it now.. added some proper tests and sanitized output of get_ray_paths some

**Comment by megies** [comment]
Looks like CI is gonna give the nod.. @krischer @MMesch I think this is good to go when Travis completes.

**Comment by MMesch** [comment]
Great, thanks for your help @megies. The question is whether we put the
mayavi/vtk moment tensor also in a separate repository - to some kind of
vtk based visualisation library based on obspy..

Am 12.10.2017 18:01 schrieb "Tobias Megies" <notifications@github.com>:

**Comment by megies** [comment]
> The question is whether we put the mayavi/vtk moment tensor also in a separate repository

Sure, go ahead.

**Comment by megies** [comment]
Gonna push one additional minor change and will merge this once CI goes green then.

File: obspy/core/util/testing.py [hunk 10]
```diff
@@ -203,10 +203,21 @@ def compare_images(expected, actual, tol):
             "shape %s." % (str(actual_image.shape),
                            str(expected_image.shape)))
 
+    if actual_image.shape[2] == 4:
+        has_alpha_channel = True
```
krischer: Do we need to keep this alpha channel handling? If it is not going to be used it might become stale.

obspy/imaging/tests/data/IU_stations.txt
```diff
@@ -0,0 +1,82 @@
+#Network | Station | Latitude | Longitude | Elevation | SiteName | StartTime | EndTime
```
krischer: This file can also be removed I think

obspy/core/util/testing.py
megies: Yeah, guess you're right, I'll remove it.

megies:
Should be ready now.. gonna merge when CI goes green.

megies:
I'll merge this locally and squash everything into one commit, to avoid polluting our master history with all the binary stuff. The history and full diff across individual commits can then still be looked up here and at https://github.com/mmesch/obspy/tree/gc_paths

megies:
This was merged locally after squashing history, see 6b43db0b6fec380e4d93d8d468eececcb74b4781.

PR #1240, opened 2016-02-09
Refactor basemap plots to be able to plot into an Axes instance

megies:
Mainly to be able to add a location plot to Event plots currently being set up in #1192.
This makes e.g. the following possible now:
```python
from obspy import read_events

cat = read_events()

fig = plt.figure(figsize=(14, 7))

ax = fig.add_subplot(121)
cat.plot(projection="ortho", fig=ax, colorbar=True, show=False)

ax = fig.add_subplot(222)
cat.plot(projection="global", fig=ax, title="", colorbar=False, show=False)

ax = fig.add_subplot(224)
cat.plot(projection="local", fig=ax, title="", colorbar=False, show=True)
```
bmap


megies:
This makes the *internal* part diverge a bit between matplotlib/cartopy.. but it should not be a big issue, @QuLogic? And I'm raising a `NotImplementedError` if an `Axes` is passed on to `Catalog.plot(method="cartopy", ...)`.

megies:
Reusing `fig` kwarg to allow passing in an `Axes` might be a bit dirty.. but I feel it might be still better than introducing yet another plotting related kwarg (see the various issues open right now concerned with the topic..).

MMesch:
another possibility that could work is that it plots in the currently active ax like basemap does. Maybe this avoids some of the issues with the kwargs?

QuLogic:
No, it is better for libraries to be explicit rather then depending on Matplotlib's interactive API.

megies:
Rebased + force-pushed.

PR #1244, opened 2016-02-10
Remove 0.10 deprecations

QuLogic:
The re-routing is currently keeping some things around that shouldn't be necessary, since they were deprecated in 0.10.x already.

This is just a note of those changes; if it passes and there are no other issues, I'd probably just fast-forward #1039 to include these changes as well.


krischer:
Good idea. I was intending to just leave everything and remove ALL deprecations right after we release 1.0 but of course it is cleaner to remove some now :-)

PR #1234, opened 2016-02-12
Add Stream/Trace.remove_sensitivity(), make explicit inventory passing to Stream/Trace.remove_response() the preferred usage

trichter:
How about a new remove_sensitivity method for Trace and Stream similar to remove_response?

It would just do:

```
sensitivity = self.stats.response.instrument_sensitivity.value
self.data = self.data / sensitivity
```


trichter:
Hopefully, this will make it into the 1.0 release.

Consequently, `read()` should return an example Stream without attached response?
Tutorial still needs to be changed.

I'd thought about adding `inventory` as the first keyword  in `remove_response`, but this could potentially break code like `st.remove_response('ACC')`. What do you think?

megies:
> but this could potentially break code like st.remove_response('ACC'). What do you think?

This could be handled with a bit of introspection (checking `args` in the method call at runtime) and type checking (inventory vs. str).

> Consequently, read() should return an example Stream without attached response?

Hmm.. why? (But in any case, if we change `remove_response` to have first kwarg `inventory` it would make kind a sense to strip it anyway..)

trichter:
> This could be handled with a bit of introspection (checking args in the method call at runtime) and type checking (inventory vs. str).

Good idea. But I assume with depreciation if the unit is supplied as first argument? Otherwise this hack would become permanent.

> Hmm.. why? (But in any case, if we change remove_response to have first kwarg inventory it would make kind a sense to strip it anyway..)

That's what I meant. It would further discourage usage of attach_response. (Though, I do not have any opinion about attach_respone vs inventory keyword)

megies:
> Good idea. But I assume with depreciation if the unit is supplied as first argument? Otherwise this hack would become permanent.

Yep, with deprecation warning.

> That's what I meant. It would further discourage usage of attach_response. (Though, I do not have any opinion about attach_respone vs inventory keyword)

Yeah, I see. :+1: from my side then, I just checked, we have the appropriate response info in the example inventory from plain `read_response()`, this should be used in the example docstrings/tutorial then.

trichter:
I think this is ready, now. There is a failing doctest due to #1245.

The `attach_response` method of Stream returns the traces with no response. This is not possible by using the `inventory` keyword of `remove*` methods so far.

megies:
> This is not possible by using the inventory keyword of remove* methods so far.

I'd vote for raising a `ValueError` if a trace with no response in `inv` info is encountered.

trichter:
Yep, that is how it was implemented already.

megies:
PR looks nice! Thanks for working on this, we wanted to enforce an inventory in `remove_response` for quite some time but never got round to work on it..

Only a tiny change needed here: https://github.com/obspy/obspy/pull/1234/files#diff-2df7efb678fb1f3ad71374f561590fe1L2699

..and I'll fix the response info for our networks which causes the doctest fails..

megies:
Rebased on current master and force-pushed. Good to merge I think, if CI passes.

PR #1239, opened 2016-02-12
Bump baseline images to matplotlib.. 1.5? / adjust image test tolerance for matplotlib 1.5

krischer:


krischer:
This is a now a PR that simply ups the tolerance for matplotlib 1.5. Matplotlib 2.0 is around the corner and we will have to generate new baseline images for that in any case. When it gets released we can either figure out some smart way to get baseline images for a number of matplotlib versions or we just use 2.0 and use a high tolerance for all others.

megies:
Rebased + force-pushed.

PR #1246, opened 2016-02-12
Add deprecation reroutes for renamed methods

megies:
Done in automated fashion based on..
 * https://gist.github.com/megies/f93f1011315f0fd23a5c
 * https://gist.github.com/megies/a8c0dc1ba7e2b6e5ebaf
 * https://gist.github.com/megies/cadc0a7238c068be50de

See #1039..

### DIFF
--- a/obspy/clients/earthworm/waveserver.py
+++ b/obspy/clients/earthworm/waveserver.py
@@ -21,6 +21,7 @@ import numpy as np
 
 from obspy import Stream, Trace, UTCDateTime
 from obspy.core import Stats
+from obspy.core.util.decorator import deprecated
 from obspy.core.util.deprecation_helpers import \
     DynamicAttributeImportRerouteModule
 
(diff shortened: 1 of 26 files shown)


megies:
Rebased, now with `  # noqa`.

QuLogic:
What would be really cool would be something like:
```python
    @deprecated_method('parseHeader')
    def parse_header(self, head):
        """
        Parse tracebuf header into class variables 
```
which would automatically create the `parseHeader` method which emits the warning and calls `parse_header`.

But can decorators create two methods at once? I think probably not and the only way to do that would be to decorate the class itself...

megies:
Yeah, it's a bit ugly to have all this additional code.. but on the other hand it's static, so it's fast and failsafe, which is good.. If you have a better approach we could still do it, the commit itself wasn't done by hand but autogenerated from the list of renames above.

But I think we're better off with the static approach anyway, so I'll merge this now.

QuLogic:
At most, I'd say, have another decorator like:
```
    @deprecated_method('parse_header')
    def parseHeader(self, head):
```
to write the deprecation warning message for you. BTW, I missed the original build; why is the `noqa` needed?

megies:
> BTW, I missed the original build; why is the noqa needed?

Otherwise we fail our own new `pep8-naming` criterion due to the old, non-pep8 method names.

megies:
Now done again in e277ed35291adb80857f2c

PR #1248, opened 2016-02-15
TauP depth caching

QuLogic:
It's the PR you've been waiting for... and I mean that literally, since the CIs seem to have been taking forever lately.

This implements the depth cache using [cachetools](https://pythonhosted.org/cachetools/), if available. Otherwise, no cache is used. Using cachetools allows some flexibility in the caching algorithm that the user can tweak for specific use cases. There is also a `functools.lru_cache` in the standard library, but it's 3-only. I have a solution that would fall back to that partially implemented too that could be added here if I figure out one last bug with it.

Locally, the taup tests with caching are about 50% of the original runtime, so fingers crossed that we get some reasonable speedup on the CIs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,7 +90,6 @@ install:
         requests
   - source activate test-environment
   # install packages not available via conda
-  - pip install cachetools
   - pip install coveralls
   # 2015/02/16 - Temporary workaround. Remove limitation once installing
   # geographiclib works again.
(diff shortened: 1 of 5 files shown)


megies:
> It's the PR you've been waiting for... and I mean that literally,

:smile: 

> Locally, the taup tests with caching are about 50% of the original runtime

Nice work! :tada:

krischer:
Removed the dependency on `cachetools` and replaced it with a simple custom LRU cache implementation. Ready to merge if CI passes.

QuLogic:
Huh, still pretty slow, but at least our users will benefit somewhat...

PR #1247, opened 2016-02-15
Processing info attached twice

krischer:
While working on #1234, I noticed that processing info is attached to Trace.stats twice, once via a decorator and once explicitly at the end of the method calls.

@krischer, I think you worked last on the processing info, I think I'll leave this to you (I guess we only need the decorator..?)..


krischer:
That was a bug. Turned the issue into a PR. Should be good to go if CI passes.

Issue #887, opened 2014-10-10
Client configuration for event downloads in dayplots


QuLogic:
Someone mentioned this somewhere, though I don't recall where.

See [this comment](https://github.com/obspy/obspy/commit/cdbe891d3f0743db7610a5cdf57d93d8ee82466b#commitcomment-8050143) for suggestions from @megies:
- if encountering `min_magnitude` key emulate old behavior, show `DeprecationWarning`
- new behavior:
  - look for `base_url` key, use it for `Client(...)`, else use some default (`NERIES`?).
  - pass any other dictionary entries to FDSN Client's `get_events()` method ?
- update docstring properly

Or.. we could just deprecate the old `min_magnitude` thingy and add two separate parameter for the base_url and kwargs (e.g. `event_base_url`? `fdsn_base_url`? `event_kwargs`? `fdsn_kwargs`?)

Volunteers welcome.. maybe best open an issue until somebody chooses attacks this..

PR #1256, opened 2016-02-17
Minor pre-release warning fixes

QuLogic:
Some minor cleanup that caused some warnings while testing out the release candidate.

### DIFF
--- a/obspy/imaging/waveform.py
+++ b/obspy/imaging/waveform.py
@@ -500,7 +500,7 @@ class WaveformPlotting(object):
         if "min_magnitude" in events:
             try:
                 from obspy.clients.fdsn import Client
-                c = Client("NERIES")
+                c = Client("EMSC")
                 events = c.get_events(starttime=self.starttime,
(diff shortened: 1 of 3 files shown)


obspy/imaging/waveform.py
```diff
@@ -500,7 +500,7 @@ def plot_day(self, *args, **kwargs):
         if "min_magnitude" in events:
             try:
                 from obspy.clients.fdsn import Client
-                c = Client("NERIES")
+                c = Client("EMSC")
```
krischer: I guess this should become a parameter in the long run. Also `ISC` might be the correct resource but it likely returns a metric ton of events which might be too much for a plot.

krischer:
Looks good to me. I'll merge once CI passes.

obspy/imaging/waveform.py
QuLogic: Yep, that's #887.

krischer: We have too many issues ;-)

PR #1257, opened 2016-02-17
MSVC does not provide inttypes.h

barsch:


krischer:
@CTrabant @chad-iris (which one to use?): This concerns `libmseed`.

ThomasLecocq:
that looks OK to merge already ! excellent :)

chad-iris:
> @CTrabant <https://github.com/CTrabant> @chad-iris <https://github.com/chad-iris> (which one to use?): This concerns libmseed.

Can you tell me which version of MSVC is being used?

The last time I investigated MSVC 2010 and later were _supposed_ to come with the inttypes.h header, which is part of the C99 standard (my version did).  It is possible that support was not very good or pervasive until later versions of MSVC.  It looks like ffmpeg uses custom declarations for anything earlier than MSVC 2013.

I would rather change the version number check and have the C99 support in MSVC work as intended in later releases.  As opposed to a blanket re-declaration for all versions of MSVC.

barsch:
this one fails (MS Windows SDK 8.1)
```D:\Python\virtualenv\obspy>cl
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.30319.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.
```
and this one works (Visual Studio 2015)
```
D:\Python\virtualenv\obspy>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.23506 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.
```

so the better fix would be 
```
if defined(_MSC_VER) && _MSC_VER <= 1600
```
instead of 
```
if defined(_MSC_VER) && _MSC_VER < 1600
```

I dnot have other MS compilers installed as there are not needed for Python extensions ...

CTrabant:
> so the better fix would be
> if defined(_MSC_VER) && _MSC_VER <= 1600
> instead of
> 
> if defined(_MSC_VER) && _MSC_VER < 1600
> I dnot have other MS compilers installed as there are not needed for Python extensions ...
> 

Thanks Robert.  I think I'll follow the ffmpeg path and create custom defines for MSVC 2012 (1700) and earlier and assume that later releases have enough C99 support.  This will be in the next release:

  #if defined(_MSC_VER) && _MSC_VER <= 1700

barsch:
>  #if defined(_MSC_VER) && _MSC_VER <= 1700

sounds good for me - thanks @chad-iris

PR #1258, opened 2016-02-17
make sure no imports from obspy set matplotlib backend

megies:
Otherwise on some systems e.g. `$ obspy-runtests ...` errors out (and our `MatplotlibBackend` context manager / backend switcher becomes pretty much useless internally) because it can't set "AGG" mpl backend.


megies:
Just tested this on my Raspbian Wheezy and it fixes the problems with the CL scripts.

PR #1259, opened 2016-02-17
add test to prevent pyplot imports

megies:
Basically makes sure that code base stays clean as done by #1258 in the future.

It might be better to merge this after release of 1.0.0, to avoid introducing failing tests so close to release.

Note: Should add a test to make sure we catch false negatives (make sure the test routine actually catches those lines)

### DIFF
--- a/obspy/core/tests/test_code_formatting.py
+++ b/obspy/core/tests/test_code_formatting.py
@@ -12,6 +12,13 @@ import unittest
 from obspy.core.util.testing import check_flake8, get_all_py_files
 
 
+def _match_exceptions(filename, exceptions):
+    for pattern in exceptions:
+        if fnmatch.fnmatch(filename, pattern):
+            return True
+    return False
+
+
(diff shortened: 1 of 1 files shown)


krischer:
rebased and force-pushed to see if the tests pass.

But yea - let's merge this after the release.

megies:
> But yea - let's merge this after the release.

And still needs some testing if positives are caught actually.. (some stringio with good and bad file headers) as mentioned in first post.

krischer:
Thats a bit meta: testing the test. We could then also test the test that tests the test :-P

I would be fine with just this here - it failed before I rebased on top of #1258 and now passes.

megies:
> I would be fine with just this here - it failed before I rebased on top of #1258 and now passes.

sure, and yeah, I know it failed when it should on my system...

but.. we had some flaky regex fails due to external bugs during releasing just a day or two ago.. :P

but yeah.. low prio

Issue #1242, opened 2016-02-09
obspy.io.sac tests raise warnings


krischer:
The `obspy.io.sac` tests raise warnings which means we cannot use our test runner in verbose mode (it turns all warnings into errors).

@jkmacc-LANL Sorry for pinging you again but you currently have the greatest understanding of the `io.sac` module. Could you have a look?

In #1039 I already fixed the one with `UserWarning: Non-integers may be truncated. (nzmsec: 500.0)`. Otherwise it appears to only be two small issues.

The others are.

```bash
$ obspy-runtests io.sac -d -v
...
======================================================================
ERROR: test_read_xy_write_xy_from_bytes_io (obspy.io.sac.tests.test_core.CoreTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/tests/test_core.py", line 618, in test_read_xy_write_xy_from_bytes_io
    _write_sac_xy(st, fh)
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/core.py", line 272, in _write_sac_xy
    _internal_write_sac_xy(stream[0], filename, **kwargs)
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/core.py", line 301, in _internal_write_sac_xy
    sac = SACTrace.from_obspy_trace(trace, keep_sac_header=True)
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/sactrace.py", line 1253, in from_obspy_trace
    header = _ut.obspy_to_sac_header(trace.stats, keep_sac_header)
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/util.py", line 333, in obspy_to_sac_header
    warnings.warn(msg)
UserWarning: Old header has invalid reftime.

======================================================================
ERROR: test_read_xy_write_xy_from_open_file_binary_mode (obspy.io.sac.tests.test_core.CoreTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/tests/test_core.py", line 632, in test_read_xy_write_xy_from_open_file_binary_mode
    _write_sac_xy(st, tf)
  File "/Users/lion/workspace/code/obspy/obspy/io/sac/core.py", line 272, in _write_sac_xy
    _internal_write_sac_xy(stream[0], filename, **kwargs)
[...]

PR #1260, opened 2016-02-17
Fix SAC reftime and file mode warnings

jkmacc-LANL:
Fix SAC module warnings detailed in #1242 .  

* invalid reftime warnings are caught in the tests, and the warnings themselves are tested
* the file mode warnings are removed from the offending sac module functions.  Users are responsible for understanding the implications of file modes.


krischer:
Works like a charm on my system. Thanks :-)

```bash
$ obspy-runtests -v io.sac
```

now passes without hanging anywhere! Ready to merge if CI passes IMHO.

PR #1262, opened 2016-02-17
Silencing a log warning for the clients.fdsn.mass_downloader tests

krischer:
This just hides a test warning that manifested on Python 2.


megies:
Looks like it can go into 1.0.0 without problems, judging from the code change.

PR #1274, opened 2016-02-20
Allow overriding Python interpreter used for documentation

QuLogic:
Running `make <target>` always uses `python`, but on anything but conda, `python` is always `python2`, in accordance with the PEP. That means you can't run, e.g., `make pep8` with Python 3.


QuLogic:
It doesn't look like Travis is building this branch?

megies:
> It doesn't look like Travis is building this branch?

Oh, right, had to update .travis.yml to new maint. branch name. I rebased and force-pushed this..

PR #1275, opened 2016-02-20
Deprecation warning fixes

QuLogic:
These seem to only show up if you run in verbose mode _and_ you request a single module at a time.

However, there are still [a few](https://gist.github.com/QuLogic/eea6d576866297a72f4a) I can't fix because I don't really know about those modules.


megies:
Hmm.. probably intended to open this against `maintenance_1.0.x`?

QuLogic:
Oops, quite right; that explains why Travis built this one and not the other.

PR #1277, opened 2016-02-20
Deprecation warning fixes

QuLogic:
These seem to only show up if you run in verbose mode and you request a single module at a time.

However, there are still [a few I can't fix](https://gist.github.com/QuLogic/eea6d576866297a72f4a) because I don't really know about those modules.

Please try and fix those before merging.


megies:
Rebased + force-pushed.

megies:
Think I'll merge this now.. no idea when somebody finds time to attack those other ones. Should we open a follow-up with the above link maybe?

PR #1278, opened 2016-02-20
Taup geo bug

MMesch:
this fixes a little bug in taup_geo. If the distance is larger than 180 degrees, the azimuth points in the wrong direction. Not sure if the fix works for distances > 360 degrees though. It is just a proposition without extensive testing.

### DIFF
--- a/obspy/taup/taup_geo.py
+++ b/obspy/taup/taup_geo.py
@@ -120,11 +120,22 @@ def add_geo_to_arrivals(arrivals, source_latitude_in_deg,
         # and each could have pierce points and a
         # path
         for arrival in arrivals:
+            # check if we go in minor or major arc direction
+            distance = arrival.purist_distance % 360.
+            if distance > 180.:
+                sign = -1
+                az_arr = (azimuth + 180.) % 360.
+            else:
+                sign = 1
+                az_arr = azimuth
+            arrival.azimuth = az_arr
+            dir_degrees = np.degrees(sign * path_point['dist'])
 
             if arrival.pierce is not None:
                 geo_pierce = np.empty(arrival.pierce.shape, dtype=TimeDistGeo)
+
                 for i, pierce_point in enumerate(arrival.pierce):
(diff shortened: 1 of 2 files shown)


QuLogic:
Is this a bug fix that should go in 1.0.x?

MMesch:
well right now the lon/lat paths are wrong for distances > 180 degrees. So I think as soon as possible but I don't know how you handle such things in obspy...

QuLogic:
Can you add a test as well?

MMesch:
ok here is a test. It doesn't test the pierce points ! I have no idea how to do that easily ...

megies:
@MMesch for future PRs: If it's a bug, it should go into `maintenance_0.10.x` (and branched off of it too, of course), see https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model.

But it's not a big deal here because `master` and `maintenance_0.1.x` haven't diverged significantly.. yet.

Since the base branch of a PR can not be changed after opening it, I would recommend the following (in order to not have another duplicate issue, with discussion split up among different tickets..):
 * continue fixing this issue here on your branch based on master
 * when ready to merge:
   * don't merge this into master
   * checkout another branch based on this branch (e.g. `taup_bug2`)
   * rebase that branch on `maintenance_1.0.x`
   * merge into `maintenance_1.0.x`
   * merge `maintenance_1.0.x` into `master`
   * close this PR

That way also the diff view in this PR is not lost.

megies:
Also, if this is something that might puzzle people who run into it a changelog entry might be good. If it's something that fails with a very clear information or Exception a changelog entry might not be necessary..

MMesch:
ok. I rebased it on maintenance. Should I open a new PR then?

krischer:
Yes. Github unfortunately does not allow to change the branch one wants to merge into :-(

megies:
Ok, so closing as a duplicate of #1289.

PR #1289, opened 2016-02-24
fix azimuth in Taup for distances > 180 and attach to arrival object

MMesch:
see #1278.

Another addition: azimuth is attached to the arrival object as well


MMesch:
@QuLogic can you rerun AppVeyor somehow? It didn't work because of basemap ...

QuLogic:
Can you rebase anyway? There are conflicts, and the new flake8 is going to cause the build to fail without the fixed items.

megies:
Good to merge I guess?

MMesch:
from my side yes!

PR #1291, opened 2016-02-25
Appveyor: install basemap packages via conda

megies:
They have files for Py2.7 so we should not need this hardcoded wheel server anymore..


megies:
Wrong base branch.. sorry.

Issue #1290, opened 2016-02-25
Substring indexing of wfdisc files in obspy.io.css.core


atwinkelman:
When reading in CSS data, I have an issue with substring indexing being one column off and not catching the "foff" byte-offset value:
```
In [1]: from obspy import read
In [2]: st = read("I26DE.2015213.wfdisc", format="CSS")
...
ValueError: invalid literal for long() with base 10: '          '
```

I've verified my wfdisc file conforms to the prescribed CSS 3.0 standard given by [Anderson et al.](https://anf.ucsd.edu/pdf/css30.pdf), specifically that the character positions are within specification. I then compared the character positions read in by [obspy.io.css.core](https://docs.obspy.org/_modules/obspy/io/css/core.html#_read_css) and found an inconsistency that is likely the source of my problem. The first three wfdisc fields, sta, chan, and (start) time, are located at columns 1-6, 8-15, and 17-33, respectively, and are read in by obspy by reading substring characters 0:6, 7:15, and 16:33 (zero-indexed, not-inclusive at the end), again respectively. However, when reading the remaining wfdisc fields starting with endtime at columns 63-79, obspy reads substring characters 61:78. With python's zero-index convention this appears okay but substring extraction is not inclusive at end; the 79th column (string index 78) is not included when reading in the endtime and all wfdisc fields after that. 

The effect is that the last character of each field is trimmed off. In my case, the foff (byte offset) value is a 0 (zero) at column 257. In the obspy.io.css.core source, 
```
offset = int(line[246:256])
```
only includes the spaces leading up to the zero at column 257, but not including the zero:
```
In [12]: line
Out[12]: 'I26H1  BDF       1438387200.00000         0     5805  2015213  1438473599.95000  1728000  20.0000000         0.000143         1.000000 MB2005 o s4 - ./2015/213                                                       I26DE0.w                                  0        -1 20160217 15:28:00'

In [13]: # simulate reading of foff byte offset

In [14]: line[246:256]
Out[14]: '          '

In [15]: line[246:257]
Out[15]: '          0'

In [16]: # simulate reading of enddate

In [17]: line[61:78]
Out[17]: '  1438473599.9500'

In [18]: line[61:79]
Out[18]: '  1438473599.95000'
```

In order to handle this edge case, perhaps the right-hand side of the substring extractions should be increased by one when reading the endtime, nsamp, samprate, calib, calper, datatype, dir, dfile, and foff wfdisc fields (or where appropriate).

[...]

PR #1301, opened 2016-02-26
Restructure parsing of css/wfdisc files in obspy.io.css.core

atwinkelman:
See #1290

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,6 @@
+master:
+ - ...
+
 1.0.0: (doi: 10.5281/zenodo.46151)
  - General:
    * Requirements have been increased to reflect latest distributions:
--- a/README.md
+++ b/README.md
@@ -2,13 +2,12 @@
 
 [![TravisCI Status](https://travis-ci.org/obspy/obspy.svg?branch=master)](https://travis-ci.org/obspy/obspy)
 [![AppVeyor Status](https://ci.appveyor.com/api/projects/status/xqrbaj9phjm6l2vw/branch/master?svg=true)](https://ci.appveyor.com/project/obspy/obspy)
-[![PEP8 Status](https://docs.obspy.org/master/_images/pep8.svg)](https://pep8.obspy.org)
 [![Coverage Status](https://img.shields.io/coveralls/obspy/obspy.svg)](https://coveralls.io/r/obspy/obspy?branch=master)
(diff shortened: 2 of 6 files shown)


QuLogic:
Sorry, but while you are targetting `maintenance_1.0.x`, you _started_ with `master`. That branch will, in general, contain things that are not bug fixes and are thus not suitable for the 1.0.x line. You should rebase so that you start with `maintenance_1.0.x` and only your commit appears.

Unfortunately, you did not start a branch specifically for this PR, so your `master` will temporarily have to be behind for a bit, but I think that shouldn't be a major problem if you aren't really using the development version.

megies:
To add to the above explanations, if you're working with git on the command line you should `git rebase maintenance_1.0.x` and then force-push (`git push -f`) to this branch.

atwinkelman:
Ah, I was wondering why there were other commits and more than one file changed. Thanks for being patient. If you couldn't tell, I am new to contributing on GitHub. 

Should I open a new PR or can this one be changed appropriately?

krischer:
You can change this one. Just force push, e.g. `git push --force` to it.

QuLogic:
@megies' instructions for rebasing left out the interactive part. You need to do `git rebase --interactive` and then remove the commits you did not make. Otherwise, you will get the same results as now, just without the merge commits.

megies:
Hmm.. and on top of everything.. actually, this now seems to fall more in the new feature direction, right? So the current commit history would actually be OK but the base branch not appropriate (should then be `master`..), which can unfortunately (never got why) not be changed by design on github..

atwinkelman:
I've rebased to maintenance_1.0.x, made the suggested changes as in the latest discussion in #1290 and updated my repo. Let me know if I need to do anything different.

@megies It sure would be easier for my understanding/learning git if I could just change the pull request.

krischer:
I would suggest you close this PR and open a new one to merge these changes into `master`. This is not a bugfix and it actually contains new features.

A couple more things that need to be done:

* The `is_nnsa_kb_core()` and `read_nnsa_kb_core()` functions have to be added to the ObsPy root level `setup.py` file to register them as new plug-ins within ObsPy.
* A small test case to detect and read an NNSA KB core file is needed. Please select an openly available file that is as small as possible (on the order of single kilobytes if possible). Just copy this test case (https://github.com/obspy/obspy/blob/master/obspy/io/css/tests/test_core.py) and repeat the things for an NNSA KB Core file.
* Add an entry to the root level changelog.
* Add yourself to the contributors if you want to.

PR #1305, opened 2016-03-01
Suggestion: add set operations to Domain and Restrictions

krischer:
1. Add union and difference operations in `Domain`.

2. Add rejecting argument to `Restrictions`, e.g. `network_reject`.

Especially the second one.

And pass an `Inventory` object directly to `mass_dowload`  as the argument?

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/restrictions.py
+++ b/obspy/clients/fdsn/mass_downloader/restrictions.py
@@ -75,6 +75,32 @@ class Restrictions(object):
     ...     # Guard against the same station having different names.
     ...     minimum_interstation_distance_in_m=100.0)
 
+    The ``network``, ``station``, ``location``, and ``channel`` codes are
+    directly passed to the `station` service of each fdsn-ws implementation
+    and can thus take comma separated string lists as arguments, i.e.
+
+    .. code-block:: python
+
+        restrictions = Restrictions(
+            ...
+            network="BW,G?", station="A*,B*",
+            ...
+            )
+
+    Not all fdsn-ws implementations support the direct exclusion of network
+    or station codes. The ``exclude_networks`` and ``exclude_stations``
+    arguments should thus be used for that purpose to ensure compatibility
+    across all data providers, e.g.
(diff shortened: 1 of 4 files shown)


krischer:
> 1. Add union and difference operations in Domain.

This is already possible by creating a custom domain class. It might also be possible to come up with a some generic "solid geometry" functionality for the domain objects but that's IMHO not worth the added complexity. Those who require complex functionality are well able to code it on their own. Have a look at this example which defines a very complex custom download domain:

http://docs.obspy.org/packages/autogen/obspy.clients.fdsn.mass_downloader.domain.Domain.html#obspy.clients.fdsn.mass_downloader.domain.Domain

>     2. Add rejecting argument to Restrictions, e.g. network_reject.

At least with IRIS this was already possible. The `Restrictions` class takes a `network` argument and `network="I*,-IU"` will for example select all network starting with `I` except `IU`. I thought that is defined in the standard but it unfortunately appears to be an IRIS specific extension.

This issue is now a PR that adds `exclude_networks` and `exclude_stations` arguments, both lists of potentially wildcarded strings for networks and stations that should not be downloaded.

> And pass an Inventory object directly to mass_dowload as the argument?

Hmm...that one is a bit tougher. It would only be simple to do if the inventory object is available at least at the channel level.

krischer:
Alright..it now also has a `limit_stations_to_inventory` object argument. If an inventory is passed with this argument it will only download stations that are part of that inventory object. All other restrictions still apply - this is just a way to further restrict the to be downloaded data set.

Ready to merge from my point of view - please review.

obspy/clients/fdsn/mass_downloader/restrictions.py
```diff
@@ -103,6 +145,18 @@ class Restrictions(object):
     :type location: str
     :param channel: The channel code. Can contain wildcards.
     :type channel: str
+    :param exclude_network: A list of potentially wildcarded networks that
```
QuLogic: exclude_network**s**

PR #1307, opened 2016-03-02
Enhance Trace.times() (more output options)

megies:
First and foremost intended for convenient plotting in matplotlib axes with a
time axis..

```python
from obspy import read

tr = read()[0]
plt.plot(tr.times(type="matplotlib"), tr.data)
fig = plt.gcf()
fig.autofmt_xdate()
fig.axes[0].xaxis_date()
plt.show()
```

Still needs ..
 * [x] tests
 * [x] mention in "plotting waveforms" tutorial page
 * [x] changelog

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,13 @@ master:
    * New convenience property `.matplotlib_date` for `UTCDateTime` objects to
      get matplotlib datetime float representation (which can be used in
      time-based matplotlib axes, e.g. by Stream.plot()).
+   * Trace.times() has new options `type` and `reftime` to support fetching an
+     array of sampletimes in various different timing varieties ("relative":
+     the old default, float relative to trace starttime or `reftime` in
+     seconds; "utcdatetime": absolute times as UTCDateTime objects;
+     "timestamp": array of float POSIX timestamps, compare
+     `UTCDateTime.timestamp`; "matplotlib": array of float matplotlib dates,
+     useful for plotting on matplotlib time axes; see #1307)
  - obspy.clients.fdsn:
    * The mass downloader now has `exclude_networks` and `exclude_stations`
      arguments to not download certain pieces of data. (see #1305)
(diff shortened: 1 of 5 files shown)


krischer:
Very cool. We should have added this ages ago!

megies:
> Very cool. We should have added this ages ago!

Oh, yes. The last time i wanted to show somebody how to plot the dates and then hacking together an ugly one-liner for this finally was the last drop.. ;-)

QuLogic:
Hah, I was just thinking of how this could be done yesterday... We're on the same wavelength, it seems...

One more useful thing is to add a `reftime=...` for the relative timing, so instead of starting at 0, start at `starttime - reftime`.

krischer:
> One more useful thing is to add a reftime=... parameter for the relative timing, so instead of starting at 0, start at starttime - reftime.

:+1: That would indeed be useful.

megies:
`reftime` for me sounds like a static reference time.. (i.e. i'd expect it as time of first sample -- is that concept coming from SAC?) I'd rather expect a name like `relative_offset` or similar for a shift from `stats.starttime`.

If it's a simple shift it could of course also be done by just `tr.times("relative") - shift` without a kwarg (that would only have an effect for one specific `type`).

QuLogic:
> `reftime` for me sounds like a static reference time.. (i.e. i'd expect it as time of first sample -- is that concept coming from SAC?) I'd rather expect a name like `relative_offset` or similar for a shift from `stats.starttime`

I don't understand what you mean; you expect it to be the time of the first element? If that were the case, why would it be named `reftime` and not `starttime`?

The name is consistent with the argument used in plotting with relative x-axis.

> If it's a simple shift it could of course also be done by just tr.times("relative") - shift without a kwarg (that would only have an effect for one specific type).

That wouldn't handle `UTCDateTime` as a reference time.

QuLogic:
> I'd vote for a non-abbreviated name like `reference_time` though.

Sorry, I missed that. Feel free to change, though as I said, it matches the name in the plotting routine.

megies:
> Sorry, I missed that. Feel free to change, though as I said, it matches the name in the plotting routine.

Yeah, not sure myself.. probably matching the existing kwarg name is more important. Fine with me if it stays as is.

megies:
Will need a rebase for tests to pass, but there's still some TODOs (tests etc.), so no hurry.

megies:
Have to skip one doctest, see 31a2a1a. Should be good to merge now if Travis passes.

megies:
> 2 files (not in diff) in obspy/io/shapefile were modified. more
> 
>     Misses +133
>     Hits -133

Really strange how codecov thinks all these files have been changed when they haven't been.

Merging..

PR #1308, opened 2016-03-02
bug in taup cache function "OrderedDict.pop(last=False)"

MMesch:
this line was was raising exceptions on python2.7


krischer:
Passes on all Python version.

Issue #256, opened 2012-09-09
trace.simulate changes data dtype --> renders mseed writing impossible


trac2github:
Hi guys,

when I read in a miniseed file, simulate another instrument and want to write the resulting file as another miniseed, I run into problems. This seems to be due to the fact that the dtype of the trace.data array is changed to 'float64' in the course of instrument simulation

``` python
In [6]: st = read('/RAID/tipage/data/2008/TI/P16/HHN.D/P16.TI.HHN.D.2008.234')

In [7]: paz_old = util.find_resp('P16','N','/RAID/tipage/data/meta_data/info_file')

In [8]: paz_new = util.find_resp('TF3','Z','/RAID/tipage/data/meta_data/info_file')

In [9]: paz_old
Out[9]: 
{'gain': 1.0,
 'poles': [(-0.14799999999999999+0.14799999999999999j),
           (-0.14799999999999999-0.14799999999999999j)],
 'sensitivity': 0.80000000000000004,
 'zeros': [0j, 0j]}

In [10]: paz_new
Out[10]: 
{'gain': 1.0,
 'poles': [(-4.4429999999999996+4.4429999999999996j),
           (-4.4429999999999996-4.4429999999999996j)],
 'sensitivity': 0.16653999999999999,
 'zeros': [0j, 0j]}

In [11]: len(st)
Out[11]: 1

In [12]: st[0].data.dtype
Out[12]: dtype('int32')

In [13]: st[0].simulate(paz_remove=paz_old,paz_simulate=paz_new)

In [14]: st[0].stats
[...]

PR #1302, opened 2016-03-02
trace.detrend changes data dtype 

krischer:
Similar to issue #256, I find that, after reading data from a Seismic Unix file, detrending changes the dtype.

For example
```
d = obspy.read('Uy_file_single.su', format='SU', byteorder='<')
print d[0].data.dtype
d[0].detrend()
print d[0].data.dtype
```

produces

```
float32
float64
```

This behavior seems a bit unexpected.  Am I missing something?

Thanks,
Ryan

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2311,6 +2311,7 @@ class TraceTestCase(unittest.TestCase):
         include either a single or a double precision version.
         """
         tr = read()[0]
+        tr.data = tr.data[:100]
 
         # One for each common input dtype.
         tr_int32 = tr.copy()
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2303,7 +2303,9 @@ class TraceTestCase(unittest.TestCase):
         The dtype of the data should not change if not necessary. In general
         this means that a float32 array should not become a float64 array
         and vice-versa. Integer arrays will always be upcasted to float64
-        arrays when integer output makes no sense.
+        arrays when integer output makes no sense. Not all int32 numbers can be
+        accurately represented by float32 arrays so double precision is
+        required in order to not lose accuracy.
 
         Exceptions are custom coded C routines where we usually opt to only
         include either a single or a double precision version.
@@ -2356,6 +2358,14 @@ class TraceTestCase(unittest.TestCase):
             np.float64)
 
         # Detrending will upcast integers but should not touch floats.
+        self.assertEqual(tr_int32.copy().detrend("simple").data.dtype,
+                         np.float64)
+        self.assertEqual(tr_int64.copy().detrend("simple").data.dtype,
+                         np.float64)
+        self.assertEqual(tr_float32.copy().detrend("simple").data.dtype,
+                         np.float32)
+        self.assertEqual(tr_float64.copy().detrend("simple").data.dtype,
+                         np.float64)
         self.assertEqual(tr_int32.copy().detrend("linear").data.dtype,
                          np.float64)
         self.assertEqual(tr_int64.copy().detrend("linear").data.dtype,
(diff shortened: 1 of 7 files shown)


krischer:
This issue is now a PR fixing it.

I had a closer look and it turns out we were a bit inconsistent. Most methods behaved as expected but some not - the tested simple detrending was such an operation. For future reference: You most likely want to use `Trace.detrend("linear")` instead of the default `"simple"` detrending. See the doc for details: http://docs.obspy.org/packages/autogen/obspy.core.trace.Trace.detrend.html?highlight=detrend#obspy.core.trace.Trace.detrend

This PR adds a large but probably not fully comprehensive test to test the behaviour for most commonly used methods. It tries to assert the following things:

* `int32` and `int64` arrays are upcasted to `float64` for all methods except trimming and decimating. `float32` cannot accurately represent all `int32` numbers so double precision is necessary.
* `float32` is preserved wherever possible. For filtering and generally frequency domain operations it is upcasted to `float64` as the additional precision might make a difference in these cases.
* Also all routines that use splines always upcast to `float64`. SciPy internally works with double precision and I see no reason to convert it back to single precision.
* Same reasoning holds for our custom C routines which usually only work with double precision.

Ready to merge from my point of view. Please review!

rmodrak:
Hi Lion,  The above approach seems logical and consistent. Now understand more the underlying challenges and the accompanying obspy rationale.  Please let me know if I can help by testing the code or closing the issue.  Thanks for these thoughtful changes.  Ryan

krischer:
This should be pretty good - only our minimum dependency test build raises an error. Probably due to a very old numpy or scipy version. I'll look into it next week.

obspy/core/tests/test_trace.py
```diff
@@ -2203,6 +2203,269 @@ def test_remove_response_plot(self):
...
+            np.float64)
+        self.assertEqual(
+            tr_float32.copy().integrate(method="spline").data.dtype,
+            np.float64)
+        self.assertEqual(
+            tr_float64.copy().integrate(method="spline").data.dtype,
+            np.float64)
+
+        # Simulation is an operation in the spectral domain so double
+        # precision is a lot more accurate so its fine here.
```
QuLogic: it's

obspy/core/tests/test_trace.py
```diff
@@ -2203,6 +2203,269 @@ def test_remove_response_plot(self):
...
+        self.assertEqual(
+            tr_float32.copy().detrend("spline", order=3,
+                                      dspline=100).data.dtype,
+            np.float32)
+        self.assertEqual(
+            tr_float64.copy().detrend("spline", order=3,
+                                      dspline=100).data.dtype,
+            np.float64)
+
+        # Tapering.
```
QuLogic: These four don't have an explanation of what they should do (though it's kind of obvious).

obspy/signal/detrend.py
```diff
@@ -23,11 +23,16 @@ def simple(data):
     point of the trace
 
     :param data: Data to detrend, type numpy.ndarray.
-    :return: Detrended data.
+    :return: Detrended data. Returns the original array which has been
+        modified in-place.
```
QuLogic: `np.require` will make a copy if changing the `dtype`, so this is not always true.

QuLogic:
Needs a rebase and once the comments are addressed, I think it's good to merge.

megies:
There's one fail on the minimum dependency Travis build..

krischer:
Implemented the review points, rebased, and added a workaround for the minimum dependency build (some old scipy version causing issues...). let's see what CI says.

krischer:
The minimum dependency build fails because the old scipy versions does not support integrating splines :-(

What do we do in cases like this? We still have to support the old scipy version as Debian wheezy ships with it. Should we raise a warning that the used scipy version is too old or backport the spline integration?

krischer:
We also really need to improve our test coverage...the current 86% are just not good enough...

QuLogic:
This integration is new and one of several methods, no? In that case, I'd say just raise a warning or whatever instead of backporting. It's different from the filtering case where you basically had no other choice.

krischer:
Rebased + force pushed.

I also added tests for the integration methods and attempted a backport of the `antiderivative()` method to old scipy versions. If it works: its only a couple of lines and we gain feature parity for old scipy versions - debian 7 is still going to be around for a while so it seems worthwhile to me. Let's see what CI has to say.

obspy/signal/differentiate_and_integrate.py
```diff
@@ -45,4 +45,19 @@ def integrate_spline(data, dx, k=3, **kwargs):
     time_array = np.linspace(0, (len(data) - 1) * dx, len(data))
     spline = scipy.interpolate.InterpolatedUnivariateSpline(time_array, data,
                                                             k=k)
+
+    # Backport of the antiderivative() method to old scipy versions.
```
QuLogic: Add version here so we know when to remove it.

krischer: Done.

krischer:
Yay :-) The minimum dependency build passes! I'll merge once the others pass as well.

PR #1309, opened 2016-03-02
FDSN Client not allowing valid local computer URLs

krischer:
As I already feared when the regex checks were introduced in #1162, these checks are catching valid URLs now when trying to connect to a machine in the local network by it's name (e.g. `Client("http://jane")`) without the domain suffixes. Of course this can be circumvented by e.g. using the IP but it nevertheless should be fixed in the regex checks.

@petrrr, would you mind adjusting these checks from #1162 if you find the time at some point?


krischer:
This issue is now a PR and fixes it.

PR #1312, opened 2016-03-03
Disable code formatting tests for release builds

krischer:
We should really disable all code formatting tests for all future release builds for two reasons:

(1) New flake8/pep8 versions tend to introduce new checks which will result in test failures which are confusing for the users.
(2) The code formatting checks don't affect the users but only us, the developers.

An IMHO pretty save way would be to just skip the tests if the `obspy.__version__` is a "clean" version number.

If somebody implements this: Please do it on the maintenance branch so that ObsPy 1.0.1 already gets it.

### DIFF
--- a/obspy/core/tests/test_code_formatting.py
+++ b/obspy/core/tests/test_code_formatting.py
@@ -9,13 +9,19 @@ import os
 import re
 import unittest
 
+import obspy
 from obspy.core.util.testing import check_flake8, get_all_py_files
+
+_pattern = re.compile(r"^\d+\.\d+\.\d+$")
+CLEAN_VERSION_NUMBER = bool(_pattern.match(obspy.__version__))
 
 
 class CodeFormattingTestCase(unittest.TestCase):
     """
     Test codebase for compliance with the flake8 tool.
     """
+    @unittest.skipIf(CLEAN_VERSION_NUMBER,
+                     "No code formatting tests for release builds")
     @unittest.skipIf('OBSPY_NO_FLAKE8' in os.environ, 'flake8 check disabled')
     def test_flake8(self):
         """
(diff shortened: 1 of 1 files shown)


krischer:
I just did it myself - this is now a PR.

I think this is a pretty clean solution to this problem. Opinions?

obspy/core/tests/test_code_formatting.py
```diff
@@ -9,13 +9,19 @@
 import re
 import unittest
 
+import obspy
 from obspy.core.util.testing import check_flake8, get_all_py_files
 
+_pattern = re.compile(r"^\d+\.\d+\.\d+$")
+CLEAN_VERSION_NUMBER = bool(_pattern.match(obspy.__version__))
```
megies: This could go into core/testing.py for potential future reuse..

megies:
:+1: :+1:  simple solution, should be good.

QuLogic:
Maybe we should pin flake8 and pep8 in the minimum dependency builds so that way we know whether any new warnings are from there or are really caused by the PR.

krischer:
Why not I guess. On the other hands its usually very easy to see so I'm not sure if its worth it.

megies:
> Maybe we should pin flake8 and pep8 in the minimum dependency builds so that way we know whether any new warnings are from there or are really caused by the PR.

Sounds like a good idea. In any case some spread across flake8/pep8/pyflakes in CI builds will certainly not hurt.

obspy/core/tests/test_code_formatting.py
krischer: I just tried that but `core/testing.py` is imported before it has a version number (all our modules import `add_doctests`/`add_unittests` ...). It would thus require significant changes to our initial startup phase or awkward logic. I'm going to leave it where it is right now.

PR #1285, opened 2016-02-26
SAC reference time not automatically set for incomplete stats.sac header

jkmacc-LANL:
I want to set the 'o' sac header in a newly created Trace. This used to work with obspy<1.0 which automatically set the reference time headers. Now reference time is only set if no sac stats entry is present.

```
In [1]: tr = read()[0]
In [2]: from obspy.io.sac import SACTrace
In [3]: tr.stats.sac = {'o': 0}
In [5]: sactr = SACTrace.from_obspy_trace(tr)
/home/richter/anaconda/lib/python2.7/site-packages/obspy/io/sac/util.py:333: UserWarning: Old header has invalid reftime.
  warnings.warn(msg)
In [6]: print sactr
/home/richter/anaconda/lib/python2.7/site-packages/obspy/io/sac/sactrace.py:1395: UserWarning: Reference time information incomplete.
  warnings.warn(msg)
Reference Time = XX/XX/XX (XXX) XX:XX:XX.XXXXXX
	iztype not set
delta      = 0.00999999977648
iftype     = itime
kcmpnm     = EHZ
knetwk     = BW
kstnm      = RJOB
lcalda     = True
leven      = True
lovrok     = True
lpspol     = False
npts       = 3000
nvhdr      = 6
o          = 0.0
```

I help myself with 

```
In [8]: from obspy.io.sac.util import obspy_to_sac_header
In [9]: tr = read()[0]
In [10]: tr.stats.sac = obspy_to_sac_header(tr.stats)
In [14]: tr.stats.sac['o'] = 0
In [15]: sactr = SACTrace.from_obspy_trace(tr)
In [16]: print sactr
Reference Time = 08/24/2009 (236) 00:20:03.000000
	iztype IB: begin time
b          = 0.0
delta      = 0.00999999977648
e          = 29.9899993297
iftype     = itime
iztype     = ib
kcmpnm     = EHZ
knetwk     = BW
kstnm      = RJOB
lcalda     = True
leven      = True
lovrok     = True
lpspol     = False
npts       = 3000
nvhdr      = 6
nzhour     = 0
nzjday     = 236
nzmin      = 20
nzmsec     = 0
nzsec      = 3
nzyear     = 2009
o          = 0.0
scale      = 1.0
```

Maybe this edge case could be handled inside io.sac?

PR #1313, opened 2016-03-03
Fix some bugs in the code, discovered by just now released new pyflakes

megies:
See e.g. http://tests.obspy.org/36956/#1.
Guess those are from the 14% untested code.. ;-)

### DIFF
--- a/obspy/io/cmtsolution/core.py
+++ b/obspy/io/cmtsolution/core.py
@@ -80,7 +80,7 @@ def _is_cmtsolution(filename_or_buf):
                              reset_fp=True)
     # Happens for example when passing the data as a string which would be
     # interpreted as a filename.
-    except (OSError, FileNotFoundError):
+    except OSError:
         return False
 
 
--- a/obspy/io/seiscomp/sc3ml.py
+++ b/obspy/io/seiscomp/sc3ml.py
@@ -890,14 +890,15 @@ def _read_float_var(elem, cls, unit=False, datum=False, additional_mapping={}):
         convert = float(elem)
     except:
         warnings.warn(
-            "'%s' could not be converted to a float. Will be skipped. Please "
-            "contact to report this issue." % etree.tostring(elem),
+            "Encountered a value '%s' which could not be converted to a "
+            "float. Will be skipped. Please contact to report this "
+            "issue." % elem,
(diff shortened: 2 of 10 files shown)


obspy/io/cmtsolution/core.py
```diff
@@ -80,7 +80,7 @@ def _is_cmtsolution(filename_or_buf):
                              reset_fp=True)
     # Happens for example when passing the data as a string which would be
     # interpreted as a filename.
-    except (OSError, FileNotFoundError):
+    except OSError:
```
QuLogic: `FileNotFoundError` is a Python 3 exception and maybe comes from `future`? Though it subclasses from `OSError` anyway.

obspy/io/seiscomp/sc3ml.py
```diff
@@ -890,14 +890,15 @@ def _read_float_var(elem, cls, unit=False, datum=False, additional_mapping={}):
         convert = float(elem)
     except:
         warnings.warn(
-            "'%s' could not be converted to a float. Will be skipped. Please "
-            "contact to report this issue." % etree.tostring(elem),
+            "Encountered a value '%s' which could not be converted to a "
+            "float. Will be skipped. Please contact to report this "
```
QuLogic: Probably should say _who_ to contact, too...

obspy/io/cmtsolution/core.py
megies: Yeah, makes no difference at all if it's in there or not (as a subclass of `OSError`..)

obspy/io/seiscomp/sc3ml.py
megies: Well, the whole routine should be refactored. Seems like a copy/paste from a routine working with etree elements at some point but now working on string representations of floats. I think @Jollyfant implemented this module, maybe he finds some time for cleaning up this routine at some point, but I think we should do that in a separate issue. This one is really fixing bugs and should not be held up.

megies:
Merging now, the fails on Travis Py3 are already handled in #1285 so we'll get them into maintenance soon, I think.

Jollyfant:
There are definitely some oddities in the sc3ml module. I will try and take a look and work on some improvements soon.

PR #1310, opened 2016-03-04
Make Stream.rotate() take an Inventory object  /  Stream.rotate from unaligned to ZNE

megies:
We should think about enhancing the `Stream.rotate()` function to be able to feed it an inventory object (with included dip/azimuth of channels).

TODO:
- [x] check method naming / new method / refactor existing method
- [x] add test
- [x] changelog
- [x] ~~Tutorial (we don't have a rotation tutorial right now..)~~ / API docs example usage


megies:
I've started some work on this, for now mainly on preparing data for rotation.. (cutting channels down to common time slices).
```python
from obspy import read

st = read()
t1 = st[0].stats.starttime
t2 = st[0].stats.endtime
delta = st[0].stats.delta

tr = st[0]
tr.stats.starttime += 4 * delta
tr = st[1]
tr.data = tr.data[:-13]
tr = st[2]
tr.stats.starttime -= 7 * delta
tr.data = tr.data[:-13]

st_ = st.select(channel="EHZ").copy()
st.pop(0)
st += st_.cutout(t1+3, t1+9)
st_ = st.select(channel="EHN").copy()
st.pop(0)
st += st_.cutout(t1+5, t1+18)
st_ = st.select(channel="EHE").copy()
st.pop(0)
st += st_.cutout(t1+23, t1+25)

print(st)
st._trim_common_channels()
print(st)
```

```
6 Trace(s) in Stream:
BW.RJOB..EHZ | 2009-08-24T00:20:03.040000Z - 2009-08-24T00:20:06.000000Z | 100.0 Hz, 297 samples
BW.RJOB..EHZ | 2009-08-24T00:20:12.000000Z - 2009-08-24T00:20:33.030000Z | 100.0 Hz, 2104 samples
BW.RJOB..EHN | 2009-08-24T00:20:03.000000Z - 2009-08-24T00:20:08.000000Z | 100.0 Hz, 501 samples
BW.RJOB..EHN | 2009-08-24T00:20:21.000000Z - 2009-08-24T00:20:32.860000Z | 100.0 Hz, 1187 samples
BW.RJOB..EHE | 2009-08-24T00:20:02.930000Z - 2009-08-24T00:20:26.000000Z | 100.0 Hz, 2308 samples
BW.RJOB..EHE | 2009-08-24T00:20:28.000000Z - 2009-08-24T00:20:32.790000Z | 100.0 Hz, 480 samples
9 Trace(s) in Stream:
BW.RJOB..EHZ | 2009-08-24T00:20:03.040000Z - 2009-08-24T00:20:06.000000Z | 100.0 Hz, 297 samples
BW.RJOB..EHN | 2009-08-24T00:20:03.040000Z - 2009-08-24T00:20:06.000000Z | 100.0 Hz, 297 samples
BW.RJOB..EHE | 2009-08-24T00:20:03.040000Z - 2009-08-24T00:20:06.000000Z | 100.0 Hz, 297 samples
BW.RJOB..EHZ | 2009-08-24T00:20:21.000000Z - 2009-08-24T00:20:26.000000Z | 100.0 Hz, 501 samples
BW.RJOB..EHN | 2009-08-24T00:20:21.000000Z - 2009-08-24T00:20:26.000000Z | 100.0 Hz, 501 samples
BW.RJOB..EHE | 2009-08-24T00:20:21.000000Z - 2009-08-24T00:20:26.000000Z | 100.0 Hz, 501 samples
BW.RJOB..EHZ | 2009-08-24T00:20:28.000000Z - 2009-08-24T00:20:32.790000Z | 100.0 Hz, 480 samples
BW.RJOB..EHN | 2009-08-24T00:20:28.000000Z - 2009-08-24T00:20:32.790000Z | 100.0 Hz, 480 samples
BW.RJOB..EHE | 2009-08-24T00:20:28.000000Z - 2009-08-24T00:20:32.790000Z | 100.0 Hz, 480 samples
```

megies:
I think this is pretty much ready.. includes trimming all to-be-rotated channel sets to a common time base (see e.g. first three traces; special cases like slightly offset sampling points are not yet taken into account -- we could handle this by resampling/interpolation, but for me this is currently out of scope)

Main question left is whether we leave it as is (as new method) or try to get this in line with the existing rotate method.. opinions?

```python
st = read("/path/to/ffbx_unrotated_gaps.mseed")
inv = read_inventory("/path/to/ffbx.stationxml")
st.rotate_to_zne(inv)
```

figure_1
figure_2

P.S.: Tests currently fail because I added azimuth/dip to result of `Inventory.get_coordinates()`. Not sure if we want a new separate `get_orientation()` method.. (code duplication etc..)?

krischer:
I did not yet fully review this but I don't really like the new method name. I see two possibilities:

* Either name it `Stream.rotate_to("ZNE", inventory=inv)` - then `rotate_to("RT", inventory=inv, event=ev)` and similar things can also be made to work.
* Or just keep the original method name with the function signature we discussed above.

I personally would strongly prefer the second option. We already have a lot of methods on the Stream and Trace object and it will definitely confuse people if we suddenly have two methods to rotate data.

megies:
> but I don't really like the new method name

Yeah, I agree, should ideally be unified with the existing one. Just wanted to get the implementation ready first before worrying about how to plug it in.

megies:
Ok, I've plugged the new routine into the existing `Stream.rotate()` as `method="->ZNE"`.

```python
st = read("/path/to/ffbx_unrotated_gaps.mseed")
inv = read_inventory("/path/to/ffbx.stationxml")
st.rotate(method="->ZNE", metadata=inv)  # doctest: +ELLIPSIS
```

krischer:
Interesting syntax with the arrow but I do kind of like it :-)

I would name the `metadata` kwarg just `inventory`. That's how we call it in other functions (`remove_response()`, `remove_sensitivity()`) - I consider SEED support to be legacy and if the `inventory` kwarg also has to eat a `Parser` object - so be it - but I don't think it is worth it to design the syntax around that. It is also a bit ambiguous as a potentially to be implemented `->RT` variant will also require event information which is also metadata but another kind of it.

megies:
> Interesting syntax with the arrow but I do kind of like it :-)

Just going with what was there before.. ;-) I'm not a fan of inventing some special string syntax (like the catalog filter syntax), but here it really is just an enumeration of a handful of valid values, so it should be OK, I think.

> I would name the metadata kwarg just inventory. That's how we call it in other functions (remove_response(), remove_sensitivity()) - I consider SEED support to be legacy and if the inventory kwarg also has to eat a Parser object - so be it - but I don't think it is worth it to design the syntax around that. It is also a bit ambiguous as a potentially to be implemented ->RT variant will also require event information which is also metadata but another kind of it.

I'm a bit torn on this one. I think still a lot of people operate on dataless SEED files from past deployments and a naming of `inventory` will hide it in the detailed documentation (and while this is OK, many people won't bother looking for SEED support in the details if a kwarg's called `inventory`). For that reason I also named it `metadata` in `PPSD` -- because it's able to use an inventory, a seed parser or a filename to a RESP file.

But yes.. a potential clash with metadata for events is a good enough reason for me.

krischer:
Hmm...all valid points. It could also be called `station_meta` or something along these lines. Nonetheless: In this case I vote for `inventory` as its a method on the stream object and other methods on it also use `inventory`...that would IMHO be a bit confusing. But I can also live with `station_meta` if you want to go down that route.

megies:
Nah, I think I'm convinced.. `inventory` it is.

krischer:
We should probably also have `->ZRT` and `->LQT` to go from arbitrary rotations to these. One application (that we currently don't really cover) are `E` and `N` channels that are are not directly pointing towards the east and north but are slightly off. This information is likely part of the inventory object and should thus be used.

megies:
Rebasing and force-pushing.. quite some conflicts during merging, hope I don't break things.. (old head is at 906f72ce1b853e6f6c1ea1ff8a5d9e225092ce79 for future reference)

megies:
Coming back to this one, actually I think this one's ready..

Gotta check the `Stream.rotate()` docs page when docs buildbot is done..

obspy/core/inventory/inventory.py
```diff
@@ -332,35 +332,87 @@ def get_response(self, seed_id, datetime):
             raise Exception(msg)
         return responses[0]
 
-    def get_coordinates(self, seed_id, datetime=None):
+    def _get_channel_metadata(self, seed_id, datetime=None):
```
krischer: Any particular reason this is a private method? Probably useful in its own right.

obspy/core/inventory/network.py
```diff
@@ -275,16 +275,55 @@ def get_coordinates(self, seed_id, datetime=None):
                     # if channel latitude or longitude is not given use station
                     data['latitude'] = cha.latitude or sta.latitude
                     data['longitude'] = cha.longitude or sta.longitude
-                    data['elevation'] = cha.elevation
+                    data['elevation'] = cha.elevation or sta.elevation
```
krischer: Minor nitpick as its a bit unlikely in practice - but the three previous lines should be `cha.X if cha.X is not None else sta.X`.

obspy/core/inventory/network.py
```diff
@@ -275,16 +275,55 @@ def get_coordinates(self, seed_id, datetime=None):
...
+        return metadata[0]
+
+    def get_coordinates(self, seed_id, datetime=None):
+        """
+        Return coordinates and orientation for a given channel.
+
+        :type seed_id: str
+        :param seed_id: SEED ID string of channel to get coordinates and
+            orientation for.
+        :type datetime: :class:`~obspy.core.utcdatetime.UTCDateTime`, optional
```
krischer: The `optional` here is redundant as its already in the function signature.

obspy/core/inventory/network.py
```diff
@@ -275,16 +275,55 @@ def get_coordinates(self, seed_id, datetime=None):
...
+            coordinates[key] = metadata[key]
+        return coordinates
+
+    def get_orientation(self, seed_id, datetime=None):
+        """
+        Return orientation for a given channel.
+
+        :type seed_id: str
+        :param seed_id: SEED ID string of channel to get orientation for.
+        :type datetime: :class:`~obspy.core.utcdatetime.UTCDateTime`, optional
```
krischer: Same thing with `optional` here.

obspy/core/stream.py
```diff
@@ -2564,13 +2564,36 @@ def normalize(self, global_max=False):
...
+        Example to rotate unaligned borehole instrument data based on station
+        inventory (a dataless SEED :class:`~obspy.io.xseed.parser.Parser` can
+        also be provided, see details for option ``inventory``):
+
+        >>> from obspy import read, read_inventory
+        >>> st = read("/path/to/ffbx_unrotated_gaps.mseed")
+        >>> inv = read_inventory("/path/to/ffbx.stationxml")
+        >>> st.rotate(method="->ZNE", inventory=inv)  # doctest: +ELLIPSIS
+        <obspy.core.stream.Stream object at 0x...>
+
```
krischer: Would it be better to move this example to the bottom? And maybe add a couple other examples? Otherwise people might not realize that this can also do other kinds of rotations.

krischer:
Thanks :) Can IMHO be merged if CI passes.

megies:
Fastfail for Travis and Appveyor seems to be working like a charm nowadays..

Force-pushed right now and the old builds were cancelled within seconds, in favor of the new one. :-)

megies:
Fixed a mistake during rebasing.. I'm adding everything in xseed Parser's `get_coordinates()`, too lazy to also add methods `get_orientation` and `get_metadata` as with Inventory.. and it's kind of legacy anyway..

megies:
Python 3 fix + rebased + force-pushed

megies:
I'm a bit shocked that `np.load()` with pure array data in npz files is hickuping with UnicodeDecodeError on Python3.. I really thought it's absolutely safe to use..

megies:
Tests should pass now..

megies:
Switching test data file to SLIST format..

megies:
Could be merged after a final review..

obspy/core/stream.py
```diff
@@ -3101,6 +3134,205 @@ def _dummy_stream_from_string(s):
...
+        """
+        self._cleanup()
+        channel_infos = self._get_common_channels_info()
+        new_traces = []
+        for (net, sta, loc, cha_pattern), infos in channel_infos.items():
+            st = self.select(network=net, station=sta, location=loc,
+                             channel=cha_pattern)
+            st.trim(infos["start"], infos["end"])
+            for _, _, _, _, start_, end_, _, _ in infos["gaps"]:
+                st = st.cutout(start_, end_)
```
trichter: One side effect of the cutout is that it changes the number of traces in the stream if gaps are present, which is a bit unexpected at the first glance. I assume it is no alternative to mask the data?

obspy/core/stream.py
```diff
@@ -3101,6 +3134,205 @@ def _dummy_stream_from_string(s):
...
+        st = self.select(network=network, station=station, location=location)
+        st = (st.select(channel=channels[0]) + st.select(channel=channels[1]) +
+              st.select(channel=channels[2]))
+        # remove the original unrotated traces from the stream
+        for tr in st.traces:
+            self.remove(tr)
+        # cut data so that we end up with a set of matching pieces for the tree
+        # components (i.e. cut away any parts where one of the three components
+        # has no data)
+        st._trim_common_channels()
```
trichter: +1 Would be nice to use this method for the other rotations, too

obspy/core/stream.py
```diff
@@ -2564,13 +2564,26 @@ def normalize(self, global_max=False):
...
         :param method: Determines the rotation method.
 
+            ``'->ZNE'``: Rotates data from three components into Z, North- and
+                East-components based on the station metadata (e.g. borehole
+                stations). Uses mandatory ``inventory`` parameter (provide
+                either an :class:`~obspy.core.inventory.inventory.Inventory` or
+                :class:`~obspy.io.xseed.parser.Parser` object) and ignores
+                ``back_azimuth`` and ``inclination`` parameters. Additional
+                kwargs will be passed on to :meth:`_rotate_to_zne()` (use if
+                other component pairs than ``["Z", "1", "2"]`` and
```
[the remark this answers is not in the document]
megies: done

obspy/core/stream.py
megies: Yeah, this might be unexpected for users, but in general relying on something like `stream[0]` in user code without using things like `stream.select(component=...)` and checking length of that output (which I see really often) is just plain bad practice.

Plus, masking data might lead to even much more obscure problems to unsuspecting users, with all the implications that can come with masked arrays and signal processing operations that work on the underlying numpy arrays. So I think this is by a wide margin the lesser evil.

obspy/core/stream.py
```diff
@@ -3101,6 +3134,205 @@ def _dummy_stream_from_string(s):
...
+        >>> st = read("/path/to/ffbx_unrotated_gaps.mseed")
+        >>> inv = read_inventory("/path/to/ffbx.stationxml")
+        >>> st._rotate_to_zne(inv)  # doctest: +ELLIPSIS
+        <obspy.core.stream.Stream object at 0x...>
+
+        :type inventory: :class:`~obspy.core.inventory.inventory.Inventory` or
+            :class:`~obspy.io.xseed.parser.Parser`
+        :param inventory: Inventory or Parser with metadata of channels.
+        :type component_pairs: list or tuple
+        :param component_pairs: List of combinations of three (case sensitive)
```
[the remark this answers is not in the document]
megies: renamed it to `components`

obspy/core/stream.py
megies: Hmm.. to not hold up 1.1.0 any longer, I would propose to handle this in a follow-up issue. Would you mind opening one for 1.2.0?

megies:
Some last minor adjustments to comments, rebased and force-pushed. Gonna merge this if CI looks good, unless anybody brings up more comments/reviews.

PR #1293, opened 2016-03-04
Function arguments which expect tuple are error prone with silent failures.

krischer:
The crux is that a tuple with a single item requires a comma after the item
```
type(('ex'))
<type 'str'>
type(('ex',))
<type 'tuple'>
```
If a function expects a tuple, but is passes a single iterable in () without a trailing comma, the single iterable is iterated over instead of the single item.
Examples:
GOOD: `fdsn.mass_downloader.Restrictions(... ,channel_priorities=("LH[ZNE]",), ...)`
GOOD:`fdsn.mass_downloader.Restrictions(... ,channel_priorities=("LH[ZNE]","BH[ZNE]"), ...)`
BAD:`fdsn.mass_downloader.Restrictions(... ,channel_priorities=("LH[ZNE]"), ...)`
        Iterates over string e.g.  Z,H,[,...

btw: the same thing happens with taup when using e.g. `phase_list = ('PKKP')` it starts looking for 'P', 'K', 'K', 'P' .

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/__init__.py
+++ b/obspy/clients/fdsn/mass_downloader/__init__.py
@@ -107,10 +107,10 @@ ObsPy knows of and combine it into one data set.
         # Only HH or BH channels. If a station has HH channels, those will be
         # downloaded, otherwise the BH. Nothing will be downloaded if it has
         # neither. You can add more/less patterns if you like.
-        channel_priorities=("HH[ZNE]", "BH[ZNE]"),
+        channel_priorities=["HH[ZNE]", "BH[ZNE]"],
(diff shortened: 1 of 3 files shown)


krischer:
This issue is now a PR. It does two things:

* Change all the examples to use lists instead of tuples.
* Actually raises a `TypeError` for the `location/channel_priorites` lists if a string is passed. It will never be the intended behavior to pass a string so that should be a pretty safe check - it has already bitten me twice in a couple of days and this PR should put that particular worry to rest.

krischer:
Travis errors are unrelated. Merging.

PR #1321, opened 2016-03-07
Autoset wave format on write based on extension

claudiodsf:
When writing `Stream()` or `Trace()` objects to disk, file format can
autmatically be deduced from filename extension.

Does this need a test or further modifications in documentation?

### DIFF
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -15,6 +15,7 @@ from pkg_resources import load_entry_point
 import numpy as np
 
 from obspy import Trace, read
+from obspy.core.compatibility import mock
 from obspy.core.utcdatetime import UTCDateTime
 from obspy.core.util.base import (NamedTemporaryFile, _get_entry_points,
                                   DEFAULT_MODULES, WAVEFORM_ACCEPT_BYTEORDER)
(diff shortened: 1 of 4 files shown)


claudiodsf:
Not sure about what's going on with flake8. Tests pass on my machine (flake8-2.5.4).

QuLogic:
I think a rebase to latest master should fix the flake8 problems.

claudiodsf:
The branch seems to be up to date with respect to current master. @megies was telling me that we should maybe merge the maintenance branch into master to fix this flake8 problem.

QuLogic:
The maintenance branch is now merged on master, so you can rebase now.

claudiodsf:
Rebased and force-pushed.

megies:
Would be good to have a short test for this..

claudiodsf:
I'll work on that later today or tomorrow.

claudiodsf:
If all the tests go green, that's good to go from my side.

krischer:
I reverted the commit with the old test and replaced it with a designated test that used mocked functions. This is IMHO a bit cleaner and also faster as it performs no I/O. The actual write methods are tested elsewhere so this should be comprehensive enough.

Has to be rebased to remove the two unnecessary commits before merging.

claudiodsf:
That was totally beyond my Python-fu!

claudiodsf:
Rebased and force-pushed.

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -425,6 +426,47 @@ def test_deepcopy(self):
...
+                    self.assertEqual(p.call_count, 1)
+
+                    # Manually specifying the format name should overwrite
+                    # this.
+                    with mock.patch("obspy.io.mseed.core._write_mseed") as p:
+                        obj.write("temp." + s, format="mseed")
+                    self.assertEqual(p.call_count, 1)
+
+        # An unknown suffix should raise.
+        with self.assertRaises(TypeError):
```
megies: Euh, do we really raise a `TypeError`.. should be a `ValueError`. But has nothing to with this PR of course, just noticed right now.

megies:
Good to merge, I think.

Issue #1008, opened 2015-03-31
UTCDateTime: improve behaviour wrt to limits


petrrr:
It looks like `UTCDateTime` has a somewhat inconsistent behaviour when it comes to deal with limits. In particular it is possible to create objects, which are then cannot be represented and are difficult to deal with.

Here the observations:

```
In [36]: print(obspy.UTCDateTime("9999-12-31T23:59:59.9999"))
9999-12-31T23:59:59.999908Z
```

```
In [37]: print(obspy.UTCDateTime("9999-12-31T23:59:59.999999"))
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-37-0584b3a56252> in <module>()
----> 1 print(obspy.UTCDateTime("9999-12-31T23:59:59.999999"))

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/obspy/core/utcdatetime.pyc in __str__(self)
    900         '2008-10-01T12:30:35.045020Z'
    901         """
--> 902         return "%s%sZ" % (self.strftime('%Y-%m-%dT%H:%M:%S'),
    903                           (self.__ms_pattern % (abs(self.timestamp % 1)))[1:])
    904 

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/obspy/core/utcdatetime.pyc in strftime(self, format)
   1134         :meth:`~datetime.datetime.strptime()` for more information.
   1135         """
-> 1136         return self._getDateTime().strftime(format)
   1137 
   1138     def strptime(self, date_string, format):

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/obspy/core/utcdatetime.pyc in _getDateTime(self)
    497         # datetime.utcfromtimestamp will cut off but not round
    498         # avoid through adding timedelta - also avoids the year 2038 problem
--> 499         return TIMESTAMP0 + datetime.timedelta(seconds=self.timestamp)
    500 
    501     datetime = property(_getDateTime)

OverflowError: date value out of range
```

```
In [39]: print(obspy.UTCDateTime(9224781964325.78))
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-39-bfcca3d15ddb> in <module>()
----> 1 print(obspy.UTCDateTime(9224781964325.78))

[...]

PR #1325, opened 2016-03-10
Advanced UTCDateTime based on nanoseconds - supporting years 1-9999

### LINKED ISSUE #1318: UTCDateTime before 1900
**Description:**
printing UTCDateTime raises Exceptions before 1900 whereas datetime and UTCDateTime cover in principle the year 1 all the way to the year 9999.

**Remarks:**
- `obspy.__version__: 1.0.0.dev0+20.g4c9f93ae96.obspy.maintenance.1.0.x`
- `Python 2.7.11 |Continuum Analytics, Inc.`
- this 'bug' already appeared in #168

**Example:**

```
from obspy import UTCDateTime
t1 = UTCDateTime(1, 1, 1)
print t1.datetime
print t1
```

barsch:
While trying to fix #1318 I run accross the following issue:

```
Python 2.7.9 (default, Dec 10 2014, 12:28:03) [MSC v.1500 64 bit (AMD64)]

>>> import datetime
>>> datetime.datetime(1,1,1,0,0,0,1)
datetime.datetime(1, 1, 1, 0, 0, 0, 1)
>>> print(datetime.datetime(1, 1, 1, 0, 0, 0, 1))
0001-01-01 00:00:00.000001
```

so microseconds support in Python datetime has been massively improved, unfortunately our current UTCDateTime implementation does not reflect that

```
>>> UTCDateTime(1,1,1,0,0,0,1)
0001-01-01T00:00:00.000000Z 
>>> UTCDateTime(1,1,1,0,0,0,1).timestamp
-62135596800.0
```

It boils down to floating point precision errors - as usual

```
>>> -62135596800.000001
-62135596800.0
>>> "%.8lf" % -62135596800.000001
'-62135596800.00000000'
```

NumPy is no help either (no `np.float128` on Windows):

```
>>> import numpy as np
>>> np.float64(-62135596800.000001)
-62135596800.0
>>> "%.8lf" % np.float64(-62135596800.000001)
'-62135596800.00000000'
```

In order to solve that issue I had a look into the following concepts:
- switch back to `datetime.datetime` instead of `float` timestamp within the UTCDateTime class - unfortunatly we loose the option to set a precision and we are bound to microseconds
- use `decimal.Decimal` instead -> working with decimal is a pain, see also http://legacy.python.org/dev/peps/pep-0410/ and https://mail.python.org/pipermail/python-dev/2012-February/116837.html
- http://docs.scipy.org/doc/numpy/reference/arrays.datetime.html - again converting back to datetime/timestamp is a pain

... but I decided to use a `long` to store nanoseconds instead of microseconds (timestamp) in the `UTCDateTime` class, this enables some really nice features, such as support form year 1 to 9999 and precision down to nanoseconds while being not much slower than the previous implementation - and I'm sure there is room for improvement - and more important: keeping the current UTCDateTime API as it is.

See https://github.com/obspy/obspy/blob/new_utcdatetime/obspy/core/tests/test_utcdatetime.py#L989-L1068.

Comments? Ideas? Is nanoseconds enough?

This PR fixes #1318 and #1008 .

### DIFF
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -227,6 +227,9 @@ class UTCDateTime(object):
             return
         elif len(args) == 1 and len(kwargs) == 0:
             value = args[0]
+            if isinstance(value, UTCDateTime):
+                self._ns = value._ns
+                return
             # check types
             try:
                 # got a timestamp
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -205,7 +205,6 @@ class UTCDateTime(object):
 
     .. _ISO8601:2004: https://en.wikipedia.org/wiki/ISO_8601
     """
-    timestamp = 0.0
     DEFAULT_PRECISION = 6
 
     def __init__(self, *args, **kwargs):
@@ -214,19 +213,24 @@ class UTCDateTime(object):
         """
         # set default precision
         self.precision = kwargs.pop('precision', self.DEFAULT_PRECISION)
+        # set directly to nanoseconds if given
+        ns = kwargs.pop('ns', None)
+        if ns is not None:
+            self._ns = ns
+            return
         # iso8601 flag
         iso8601 = kwargs.pop('iso8601', False) is True
         # check parameter
         if len(args) == 0 and len(kwargs) == 0:
-            # use current time if no time is given
-            self.timestamp = time.time()
+            # use current date/time if no argument is given
+            self._from_timestamp(time.time())
             return
         elif len(args) == 1 and len(kwargs) == 0:
             value = args[0]
             # check types
             try:
                 # got a timestamp
-                self.timestamp = value.__float__()
+                self._from_timestamp(value.__float__())
                 return
             except:
                 pass
@@ -247,7 +251,7 @@ class UTCDateTime(object):
                 # check for ISO8601 date string
                 if value.count("T") == 1 or iso8601:
                     try:
-                        self.timestamp = self._parse_iso_8601(value).timestamp
+                        self._from_iso8601_string(value)
                         return
                     except:
                         if iso8601:
@@ -298,7 +302,8 @@ class UTCDateTime(object):
                     self._from_datetime(dt)
                     return
                 dt = datetime.datetime.strptime(value, pattern)
-                self._from_datetime(dt, ms)
+                dt += datetime.timedelta(seconds=ms)
+                self._from_datetime(dt)
                 return
         # check for ordinal/julian date kwargs
         if 'julday' in kwargs:
@@ -322,8 +327,8 @@ class UTCDateTime(object):
         # check if seconds are given as float value
         if len(args) == 6 and isinstance(args[5], float):
             _frac, _sec = math.modf(round(args[5], 6))
-            kwargs['microsecond'] = int(round(_frac * 1e6))
             kwargs['second'] = int(_sec)
+            kwargs['microsecond'] = int(round(_frac * 1e6))
             args = args[0:5]
         dt = datetime.datetime(*args, **kwargs)
         self._from_datetime(dt)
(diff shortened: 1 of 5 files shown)


obspy/core/utcdatetime.py
```diff
@@ -343,32 +348,49 @@ def _set(self, **kwargs):
...
         """
         # see datetime.timedelta.total_seconds
         try:
             td = (dt - TIMESTAMP0)
         except TypeError:
             td = (dt.replace(tzinfo=None) - dt.utcoffset()) - TIMESTAMP0
-        self.timestamp = (td.microseconds + (td.seconds + td.days * 86400) *
-                          1000000) / 1000000.0 + ms
+        self._ns = \
+            (td.days * 60 * 60 * 24 + td.seconds) * 1000000000 + \
```
claudiodsf: It would be more readable, and less bug prone, to replace all these zeros with scientific notation: e.g., `1e9`.

claudiodsf: ...or `10**9`if you need it to be an integer.

barsch: most of the time it needs to be an integer - eclipse internal pep complained about ```10**9``` and wanted ```10 ** 9```which didn't improve readability - but it seams flake8 is not complaining - so I'll change it

megies:
The implemented approach seems like a heck of a good idea to me. For timestamps up to year 2262 we would even stay in plain `int`.

I think nanoseconds precision should be enough, also if we go further down in precision basically all timestamps would be handled by `long`s instead of `int`s. Not sure how much of an impact on time/memory that would present in practice.

Any idea what's causing those test fails?

barsch:
I didn't fix all test cases at time of the implementation - wanted to hear opinions first.

But based on http://tests.obspy.org/37329/

2 and 3: the string representations are equal, however the check is done at nanosecond level - we have to check the test case why they are failing

4 and 5: have to check the test case - maybe it wasn't precise enough all the time or the current implementation lacks somethings ...

1: no idea

krischer:
:+1: I also like this. 

The new MiniSEED (will likely) and ASDF are also using this to represent times and nanoseconds precision is enough even for the best GPS receivers so it should be good.

Why does it go from year 0 to 9999? If I read it correctly it is just a nanosecond precision timestamp as a 64bit signed integer, correct? According to my calculations this gives a range from 1677 - 2262 which is still more than enough. Python integers can actually be of arbitrary size as the runtime will switch to variable size integers internally once the values exceed a certain limit so there is no real limitation I guess. We could internally use a numpy 64bit integer which might be faster but I guess this does not matter.

barsch:
https://docs.python.org/2/library/stdtypes.html#typesnumeric:
"Long integers have unlimited precision."

see also the test case in this patch showing it working for the years 1 and 9999

petrrr:
Nice!
Just one stupid question: Can't the precision be kept configurable? In the end, it's only a variable (configurable) denominator of the fixed point representation. I agree that nano s is probably more then enough for most seismological applications, but is ObsPy used only in seismology ;-)

megies:
> Can't the precision be kept configurable?

Please note the difference between the precision of the implementation (that will be switched to integer nanoseconds by this PR) and the customizable precision on UTCDateTime which is handling how different UTCDateTime objects compare against each other.

megies:
This hasn't seen any activity, so I'll bump this to 1.2.0.

megies:
I think this PR breaks `clients.seishub`, because data is sent as pickled streams from the server and the data from the server is lacking `_ns` attribute then...

I can't see a way to fix this right now, as we can't make sure seishub server and client are on the same page UTCDateTime-wise..
```
(master) ~$ py.test ~/git/obspy-master/obspy/clients/seishub -k test_get_waveform_with_metadata
====================================================== test session starts =======================================================
platform linux2 -- Python 2.7.12, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /home/megies, inifile: pytest.ini
plugins: random-0.2
collected 16 items 

git/obspy-master/obspy/clients/seishub/tests/test_client.py F

============================================================ FAILURES ============================================================
_________________________________________ ClientTestCase.test_get_waveform_with_metadata _________________________________________

self = <obspy.clients.seishub.tests.test_client.ClientTestCase testMethod=test_get_waveform_with_metadata>

    def test_get_waveform_with_metadata(self):
        # metadata change during t1 -> t2 !
        t1 = UTCDateTime("2010-05-03T23:59:30")
        t2 = UTCDateTime("2010-05-04T00:00:30")
        client = self.client
        self.assertRaises(Exception, client.waveform.get_waveforms, "BW",
                          "UH1", "", "EH*", t1, t2, get_paz=True,
                          get_coordinates=True)
        st = client.waveform.get_waveforms("BW", "UH1", "", "EH*", t1, t2,
                                           get_paz=True, get_coordinates=True,
>                                          metadata_timecheck=False)

git/obspy-master/obspy/clients/seishub/tests/test_client.py:235: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
git/obspy-master/obspy/clients/seishub/client.py:505: in get_waveforms
    stream = _unpickle(data)
git/obspy-master/obspy/clients/seishub/client.py:49: in _unpickle
    obj = pickle.loads(data)
anaconda/envs/master/lib/python2.7/pickle.py:1388: in loads
    return Unpickler(file).load()
anaconda/envs/master/lib/python2.7/pickle.py:864: in load
    dispatch[key](self)
anaconda/envs/master/lib/python2.7/pickle.py:1223: in load_build
    setstate(state)
git/obspy-master/obspy/core/util/attribdict.py:115: in __setstate__
    self.update(adict)
git/obspy-master/obspy/core/util/attribdict.py:142: in update
    self.__setitem__(key, value)
git/obspy-master/obspy/core/trace.py:165: in __setitem__
    value = UTCDateTime(value)
git/obspy-master/obspy/core/utcdatetime.py:231: in __init__
    self._ns = value._ns
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'UTCDateTime' object has no attribute '_UTCDateTime__ns'") raised in repr()] SafeRepr object at 0x7fdcb2143710>

    def _get_ns(self):
>       return self.__ns
E       AttributeError: 'UTCDateTime' object has no attribute '_UTCDateTime__ns'

git/obspy-master/obspy/core/utcdatetime.py:360: AttributeError
=================================== 15 tests deselected by '-ktest_get_waveform_with_metadata' ===================================
============================================ 1 failed, 15 deselected in 1.65 seconds =============================================
```

megies:
Numpy's datetime64 seems to be rather mature with numpy 1.12: https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html#datetime-units

The nanoseconds option goes from years 1678-2262.. but of course with the variety of distros we support it will take ages until all are on numpy 1.12.. so just a comment here..

PR #1317, opened 2016-03-10
SAC output ignores sampling rate from Trace metadata

jkmacc-LANL:
When using the trace.decimate() function in v.1.0.0 on .SAC formatted traces, the new sampling rate isn't updated in the trace metadata. I've tried to work around it by decimating, then forcing the trace to update the stats with the new sampling rate using trace.stats.update. This does work for the current instance, however, when writing the trace out as a .SAC, the old sampling rate is written into the stats.

### DIFF
--- a/obspy/io/sac/header.py
+++ b/obspy/io/sac/header.py
@@ -22,7 +22,8 @@ DOC = {'npts': 'N    Number of points per data component. [required]',
                   when read into sac. [required]''',
        'b': 'F    Beginning value of the independent variable. [required]',
        'e': 'F    Ending value of the independent variable. [required]',
-       'iftype': '''I    Type of file [required]:\n
+       'iftype': '''I    Type of file [required]:
+
                   * ITIME {Time series file}
                   * IRLIM {Spectral file---real and imaginary}
                   * IAMPH {Spectral file---amplitude and phase}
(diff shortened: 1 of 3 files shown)


jkmacc-LANL:
This was pretty straightforward. Ready to merge.

obspy/io/sac/header.py
```diff
@@ -22,7 +22,7 @@
                   when read into sac. [required]''',
        'b': 'F    Beginning value of the independent variable. [required]',
        'e': 'F    Ending value of the independent variable. [required]',
-       'iftype': '''I    Type of file [required]:
+       'iftype': '''I    Type of file [required]:\n
```
QuLogic: What is this escaped newline? Why just add a real newline?

jkmacc-LANL: Doesn't pass flake8:-( This just helps the docs look nicer.

________________________________

QuLogic: Uh, no it won't. It's in a docstring; the only relevant check is line-length.

jkmacc-LANL: Hmmm, you're right.  Could've sworn it mattered.  Fixed.  BTW, man those docs take for *ever* to build.

QuLogic: Yea, it's all those networked examples...

megies: Maybe it'd worth a shot to do something similar to `obspy-runtests`, i.e. add the possibility to build only one specific module..

megies:
> This was pretty straightforward. Ready to merge.

Alright, merging then.

PR #1334, opened 2016-03-17
add INT datatype support for GSE2

barsch:
see also http://lists.swapbytes.de/archives/obspy-users/2016-March/002013.html


megies:
Thanks! I guess since INT support should have been in there before, seeing this as a bug fix for maintenance branch might be OK..

obspy/io/gse2/libgse2.py
```diff
@@ -331,6 +331,26 @@ def verify_checksum(fh, data, version=2):
...
 
+def read_integer_data(fh, npts):
+    """
+    Reads npts points of uncompressed integers from given file handler.
+    """
+    # find next DAT2 section within file
+    data = []
+    in_data_section = False
+
+    while len(data) < npts:
```
QuLogic: You need to also check if `fh` is exhausted, or you will get an infinite loop if the file is truncated.

barsch: indeed - its a c&p from libgse1.py - will fix both ...

obspy/io/gse2/tests/data/boa___00_07a.gse
```diff
@@ -140,4 +141,4 @@ DAT2
 9 9 7 7 6 5 8 10 6 2 3 6 7 7 4 1 -3 -2 2 3 0 -5 -7 -5 1 2 -4 -9 -8 -4 1 2 1 -6 -10 -6 1 1 0 -2 -4 -3 2 3 1 0 -2 -4 -6 -6 -7 -8 -7 -5
 -5 -8 -11 -12 -11 -9 -9 -10 -13 -15 -15 -12 -7 -7 -10 -12 -13 -10 -5 -4 -10 -14 -17 -18 -13 -12 -14 -15 -16 -16 -13 -11 -14 -15 -14
 -13 -10 -6 -6 -6 -3 0 1 1 0 -4 -4 0 0 -2 -5 -8 -8 -4 -3 -3 -3 -5 -4 2 9 10 6 4 2 4 10 12 8 4 4 5 7 10
-CHK2     5140
+CHK2     -5140
```
megies: Uhm.. how come this file had a wrong checksum? Was the checksum wrong in the file we got from the mail on the users mailing list? Funny, then there's another program out there mixing up the sign of the checksum, I guess..?

barsch: didn't we figure out that at some point Seismic Handler generated wrong CHK2 entries too - don't remember ...

@mbyt: any insights?

megies: Well.. not important, I guess. It's not coming from us, clearly, and we just use it for testing..

barsch:
Any ideas why Travis fails for Python >= 3.3 ?

megies:
3.3: seems like anaconda dropped pyproj for Python 3.3? Either we skip the pyproj based tests on Py3.3 or try to find a 3.3 pyproj from a third party anaconda channel..

3.4/3.5: Hmm.. I restarted those builds once already, thinking it was just a glitch (and I think that was after the infinite loop patch). I'll restart them once again before we wreck our brains about it..

megies:
Rebased on current maintenance branch to see if CI passes finally..

obspy/io/gse2/libgse2.py
```diff
@@ -331,6 +331,36 @@ def verify_checksum(fh, data, version=2):
...
+    :param npts: Number of samples to read
+    :rtype: :class:`numpy.ndarray`, dtype=int32
+    :return: Data as numpy.ndarray of type int32.
+    """
+    # find next DAT2 section within file
+    data = []
+    in_data_section = False
+
+    while len(data) < npts:
+        buf = fh.readline()
```
QuLogic: Maybe you can `.strip()` here instead of every other place.

megies:
There's still the question why Py3 builds time out..

megies:
Rebased + force-pushed.

megies:
Python 3 builds still stall.. :worried:

barsch:
Can't debug as I don't have access to a linux system atm :/

megies:
Should be fixed now. Also moved the EOF check up in the while loop to be robust against infinite loops.

obspy/io/gse2/libgse2.py
```diff
@@ -348,14 +348,14 @@ def read_integer_data(fh, npts):
 
     while len(data) < npts:
         buf = fh.readline()
-        if buf.strip() in ["DAT1", "DAT2"]:
+        if not buf:
+            # break loop if no data is given
+            break
+        if buf.decode().strip() in ("DAT1", "DAT2"):
```
QuLogic: Couldn't you just make the strings bytes instead of decoding?

megies: Yeah..

PR #1336, opened 2016-03-19
tr.__add__ crossfade possibility

MMesch:
This PR adds the possibility to crossfade traces when adding them or just cutting them in the middle. It works well but doesn't yet test all possible Exceptions (e.g. if one trace is contained fully...). Cutting in the middle is really useful when you don't want to throw away the two traces because of boundary effects after filtering. Crossfade is especially when merging traces that have slowly changing processing parameters. I hope this didn't exist before...

xfade

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -541,7 +541,7 @@ class Trace(object):
         return st
 
     def __add__(self, trace, method=0, interpolation_samples=0,
-                fill_value=None, sanity_checks=True):
+                fill_value=None, sanity_checks=True, crossfade='sum'):
         """
         Add another Trace object to current trace.
 
@@ -728,6 +728,20 @@ class Trace(object):
                                                lt.data.dtype)
                     data = [lt.data[:-delta], interpolation,
                             rt.data[interpolation_samples:]]
+            elif method == 2: # cut in the middle
+                ldelta = delta//2
+                rdelta = delta - ldelta
+                data = [lt.data[:-ldelta], rt.data[rdelta:]]
+            elif method == 3: # sum both traces
+                if crossfade == 'box':
+                    overlap_samples = lt[-delta:] + rt[:delta]
+                elif crossfade == 'linear':
+                    fadein = np.linspace(0., 1., delta)
+                    fadeout = fadein[::-1]
+                    overlap_samples = lt[-delta:] * fadeout +\
+                                      rt[:delta] * fadein
+                data = [lt.data[:-delta], overlap_samples,
+                        rt.data[delta:]]
             else:
                 raise NotImplementedError
         elif delta < 0 and delta_endtime >= 0:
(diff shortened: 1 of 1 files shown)


megies:
Just a thought.. we should probably also provide something very different..

Assuming data come from one single physical instrument, meaning that there is a definitive jump in timing.. then any interpolation on overlapping data handle one problem by introducing other problems. I would argue that in almost all cases the later trace's timing should be considered correct and the earlier trace shifted to align with the later trace (assuming that there was no gap in recording).
In principle the same goes for gaps, but of course with a gap the probability that the mismatching time segments include a real gap in data recording (as opposed to a jump in the clock) is obviously not as easy to rule out.

Of course this is a way of thinking coming from working with relatively short pieces of waveforms (e.g. single phase onsets or a local event with P and S wave).

MMesch:
What do you mean exactly with 'interpolation' after a jump in timing?

E.g. trace 1 samples at [...], 1, 2, 3, 4 and trace 2 samples at 3.1, 4.1, 5.1, 5.1, [...]? In a naive/perfect world either of the traces could be interpolated, throwing away one sample. If there are glitches etc, interpolation can do harm I guess, so it is better to just accept a shift by maximum delta/2, maybe issuing a warning.

Is this what you were referring too? This point is really not very clear in the __add__ function...

If any interpolation is done, I would suggest to do it before any of the different combine methods (which is an 'interpolation' in trace.data amplitude). That way the time and amplitude interpolations are fully separate.

megies:
Well, I admit, that was a bit of an off-topic comment..

I just wanted to say that I only see two kinds of "sane" options for an overlap (on a time series from one single physical instrument for one specific downsampling scenario in the digitizer)..

*a)* duplicated data (e.g. during transmission from station to data storage)
*b)* digitizer clock update

.. and all of our merging approaches when it comes to overlaps are kind of handling "non-sane" overlap options, whereas option *b)* would need something completely different done, at least on a rather local timeframe (e.g. time series cut down to a wave onset or earthquake).
What should be done in this case is to simply align the traces and trust the time stamping of one of the traces (usually the later one, as presumably the clock was drifting beforehand and then reset to a valid sync signal -- think GPS).

Check out this example:
```python
from obspy import read

st_ = read("http://examples.obspy.org/time_jump.mseed")

for _, _, _, _, end, start, _, _ in st_.get_gaps(max_gap=0):
    st = st_.slice(start - 5, end + 5)

    # overlap due to digitizer clock jumping
    st.plot()

    # simply shift earlier trace to match timebase of later trace
    st[0].stats.starttime += \
        st[1].stats.starttime - (st[0].stats.endtime + st[0].stats.delta)
    st.plot()
```

This is before handling the overlap:
figure_1

And this is when shifting the first trace to align with the start of the second trace:
figure_2


Of course when time span gets longer (minutes, hours, days), this over-simple approach breaks down..

MMesch:
Ok I see. I think we have two distinct requirements here that are all in add and merge so far:

1) merge a time series from one single physical instrument for one specific downsampling scenario in the digitizer with the intent of recovering the original data

2) merge two time series that can have different sampling rate, etc... . The intent here is to facilitate data processing operations (e.g. stacking, filtering).

2 seems to be more general than 1.

megies:
In any case.. another (well documented) merge method can't hurt, so go ahead with this one.

PR #1338, opened 2016-03-21
".post0+..." for development versions that come after a release

megies:
..so that local development git clone installations compare greater than previous point release (e.g. as seen by packaging tools like pip and conda).
So, replace `.dev0+...` with `.post0+...` in development version strings.

See https://www.python.org/dev/peps/pep-0440.


megies:
Rebased, force-pushed.

PR #1339, opened 2016-03-21
add convenience method and property to UTCDateTime to get matplotlib datetime float representation

megies:
..for plotting on matplotlib time axes.

This adds a method `UTCDateTime.to_matplotlib()` and a property `UTCDateTime.mpl` for even shorter notation.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,10 @@
 master:
  - obspy.core
    * Stream/Trace.write() can now autodetect file format from file extension.
- - obspy.clients.fdsn
+   * New convenience property `.matplotlib_date` for `UTCDateTime` objects to
+     get matplotlib datetime float representation (which can be used in
+     time-based matplotlib axes, e.g. by Stream.plot()).
+ - obspy.clients.fdsn:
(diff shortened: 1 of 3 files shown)


CHANGELOG.txt
```diff
@@ -1,5 +1,9 @@
 master:
- - obspy.clients.fdsn
+ - obspy.core:
+   * New convenience method `.to_matplotlib()` and convenience property `.mpl`
+     for `UTCDateTime` objects to get matplotlib datetime float representation
+     (which can be used in time-based matpltolib axes, e.g. by Stream.plot()).
```
QuLogic: Typo in matplotlib.

megies:
Rebased + force-pushed.

megies:
If nobody objects to the property alias `UTCDateTime.mpl` this PR is introducing I'm gonna merge this soonish.

krischer:
Looks mostly good.

I don't think we need both a method and a property for the same thing to be honest. It will IMHO just clutter the UTCDateTime namespace. Would you be okay with removing the method and giving the property a more descriptive name like `mpl_date` or `matplotlib_date`?

megies:
Fair enough, force-pushed.

PR #1343, opened 2016-03-22
Stream normalization

krischer:
Hi all,
I recently had an issue with the normalization of the data in a stream object. I try to test some data recovery methods, therefore I need traces in Stream that are just filled with zeros. If I normalize the stream, all traces with just zeros on it return with 'nan' values instead of zeros. I don't know if this is intended or not, but I thought it would be plausible if it stays zero?

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,7 @@
 1.0.x:
+ - obspy.core:
+    * Trace.normalize() does no longer divide by zero in case an all-zeros
+      data trace is being used. (see #1343)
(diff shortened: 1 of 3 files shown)


krischer:
This is now a PR with a fix for this issue. It also adds a comprehensive test for the `normalize()` method which we previously did not have. The only behavioral change is that it no longer divides by zero but just leaves the array as is in that particular case.

krischer:
On second though: it now raises when attempting to normalize a zero array.

PR #1126, opened 2016-03-23
Inventory.select() unexpected default behavior

krischer:
Here, the issue relates to a question on Gitter.

I was trying to do the following: 
 * get a level=station inventory
 * select a station

The default behavior is to exclude stations without channels and will return an empty inventory. Given the fact that the station is present in the Inventory, this is a bit surprising. The expected result is obtained when 
 (a) the level is increased to 'channel'
 (b) overwriting the default by setting parameter `keep_empty=True`

This issue is to evaluate and track change of the default.

Details:
```python
inv_stat = client.get_stations(network=', '.join(search_nets), level='station')

print(inv_stat)
Inventory created at 2015-09-18T16:37:48.000000Z
	Sending institution: SeisComP3 (INGV)
	Contains:
		Networks (2):
			IV
			MN
		Stations (484):
			IV.ACER (Acerenza)
			IV.AGST (AUGUSTA)
[...]
		Channels (0):
inv_select = inv_stat.select(station='AQU')

print(inv_select)
Inventory created at 2015-09-18T16:37:48.000000Z
	Sending institution: SeisComP3 (INGV)
	Contains:
		Networks (2):
			IV
			MN
		Stations (1):
			IV.CAMP (CAMPOTOSTO)
		Channels (0):
```

### DIFF
--- a/obspy/core/tests/test_inventory.py
+++ b/obspy/core/tests/test_inventory.py
@@ -231,7 +231,7 @@ class InventoryTestCase(unittest.TestCase):
         # Currently contains 30 channels.
         self.assertEqual(sum(len(sta) for net in inv for sta in net), 30)
 
-        # Nothing selected, nothing should change.
+        # No arguments, everything should be selected.
         self.assertEqual(
             sum(len(sta) for net in inv.select() for sta in net),
             30)
@@ -324,7 +324,7 @@ class InventoryTestCase(unittest.TestCase):
         self.assertEqual(len(inv), 2)
         self.assertEqual(sum(len(net) for net in inv), 0)
 
-        # Nothing selected, nothing should change.
+        # No arguments, everything should be selected.
         self.assertEqual(len(inv), 2)
         # Same if everything is selected.
         self.assertEqual(len(inv.select(network="*")), 2)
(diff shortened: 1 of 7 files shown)


krischer:
I think this now behaves as expected. If `Network` or `Station` objects have no further child elements they will always be kept, independent of the `keep_empty` flag (of course assuming that the other query parameters are satisfied). We also, for some reason, had no tests for all the `.select()` methods so I added a bunch of tests that should be reasonably comprehensive.

krischer:
Oh yea: All the inventory objects now have `__len__` methods which return the number of child elements.

QuLogic:
`np.isclose` is NumPy 1.7; maybe use `np.allclose`?

krischer:
Let's see what CI says.

obspy/core/tests/test_inventory.py
```diff
@@ -212,6 +214,127 @@ def test_inventory_merging_metadata_update(self):
...
+    def test_inventory_select(self):
+        """
+        Test for the Inventory.select() method.
+        """
+        inv = read_inventory()
+
+        # Currently contains 30 channels.
+        self.assertEqual(sum(len(sta) for net in inv for sta in net), 30)
+
+        # Nothing selected, nothing should change.
```
QuLogic: Wouldn't that be _everything_ selected?

krischer: Yea...bad comment from side!

krischer:
Why does this decrease coverage?? I ADDED tests ;-)

QuLogic:
Yea, that's weird; all the individual builds went up, too. Anyway, this appears to need a rebase.

krischer:
Rebased. It was only the changelog. No need to wait for CI.

PR #1359, opened 2016-03-26
Fix some issues with NumPy 1.11.0 beta 3.

QuLogic:
See [Rawhide build log](https://copr-be.cloud.fedoraproject.org/results/qulogic/obspy/fedora-rawhide-x86_64/00171370-python-obspy/build.log.gz).

### DIFF
--- a/obspy/imaging/source.py
+++ b/obspy/imaging/source.py
@@ -575,7 +575,8 @@ def _equalarea_spherical_grid(nlat=30):
     ndim = 3
     colats = np.linspace(0., np.pi, nlat)
     norms = np.sin(colats)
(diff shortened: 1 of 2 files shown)


krischer:
Does this still happen with 1.11.0 final? It works fine for me on OSX but I did not test all Python versions.

http://tests.obspy.org/38513/

If it still happens we would kind of have to release 1.0.2 immediately.

QuLogic:
OK, looks like there shouldn't be any need to worry; the final release of NumPy 1.11 works fine without these changes.

I'm going to merge this anyway just because it makes the calls to `astype` consistent, and not because it's needed.

PR #1360, opened 2016-03-29
enabled arclink decryption during download

barsch:
lesser memory footprint while having the same execution speed as before

### DIFF
--- a/obspy/clients/arclink/README.txt
+++ b/obspy/clients/arclink/README.txt
@@ -5,7 +5,7 @@ Copyright
 ---------
 GNU Lesser General Public License, Version 3 (LGPLv3)
 
-Copyright (c) 2009-2013 by:
+Copyright (c) 2009-2017 by:
     * Robert Barsch
 
 
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -340,26 +340,9 @@ class Client(object):
             self._bye()
             msg = "Uncovered status message - contact a developer to fix this"
             raise ArcLinkException(msg)
-        self._write_ln('DOWNLOAD %d' % req_id)
-        try:
-            fd = self._client.get_socket().makefile('rb')
-            length = int(fd.readline(100).strip())
-            data = b''
-            while len(data) < length:
-                buf = fd.read(min(4096, length - len(data)))
-                data += buf
-            buf = fd.readline(100).strip()
-            if buf != b"END" or len(data) != length:
-                raise Exception('Wrong length!')
-            if self.debug:
-                if data.startswith(b'<?xml'):
-                    print(data)
-                else:
-                    print("%d bytes of data read" % len(data))
-        finally:
-            self._write_ln('PURGE %d' % req_id)
-            self._bye()
+
         # check for encryption
+        decryptor = None
         if b'encrypted="true"' in xml_doc:
             # extract dcid
             xml_doc = objectify.fromstring(xml_doc[:-3])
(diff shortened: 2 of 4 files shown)


barsch:
coverage/coveralls lies - it includes files which were not changed in this PR - however I really love the idea of this feature if it would work ...

krischer:
Yea coverage is funny sometimes...no clue why lists untouched files.

The PR itself looks fine to me. I assume you locally tested on py2 and py3?

barsch:
tests were fine for windows - still would be nice if some would test it manually on linux

krischer:
I get lots of timeouts when running `test_decrypt.py` but can run the normal arclinks tests just fine. Is this expected? Is encrypted data send over a different port?

barsch:
hmm not good - I'll check at home if this is still running for me and report back - thanks for testing

barsch:
oh and port 36000 needs to be open - see https://github.com/obspy/obspy/blob/master/obspy/clients/arclink/tests/test_decrypt.py#L35

krischer:
I see. I'll test at home.

krischer:
FYI: I just pushed a commit to this branch that allows to run the crypto tests with either supported crypto library.

barsch:
EVPError is only defined for M2Crypto ! -> unfortunatly as far I tested pycrypto does not raise anything if there is an error during decrypting - so we have to stick with M2Crypto as default and only take pyCrypto as fallback

krischer:
Hmm...there must be a way to figure out if the decrypt was successful, right? How does `m2crypto` even know? I mean we could always check the resulting miniseed file but m2crypto clearly does something else.

barsch:
while using pycrypto it just fails later - either because it can't unpack using bzip2 or it can't read as its not a valid mseed

AFAIK m2crypto checks in the final call https://github.com/obspy/obspy/blob/master/obspy/clients/arclink/decrypt.py#L66 - so we could look into that

megies:
+TESTS:clients.arclink

Travis should run ArcLink test suites for future commits/pushes to this branch now (see [here](https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md#submitting-a-pull-request)). Please ping me if it doesn't work..

megies:
What's the status here, @barsch @krischer?

barsch:
the later commit from @krischer does imho not work - see comment above - so either we stick with my orginal implementation, leave it as it is or find something better

megies:
> and triggering a new test run did not work, or?

I rebased and force-pushed right now, that will trigger CI..

megies:
Current Travis commit status includes Arclink tests: https://travis-ci.org/obspy/obspy/builds/176344138

Not sure if tests cover decryption, though?

megies:
I'll bump this to 1.2.0. Feel free to add to 1.1.0 if it gets ready in the next couple of weeks.

barsch:
imho ready to merge

barsch:
failing test is not related - merging

megies:
@barsch this PR should also have modified arclink extra/optional dependencies.. anybody know how to modify a set of OR dependencies in setup.py? I couldn't find it..

https://github.com/obspy/obspy/blob/master/setup.py#L116

Should be semantically be something like..
```
 'arclink': ['m2crypto | pycrypto | cryptography'],
```

barsch:
I don't think there is such option - I would take it out completely and just document it

btw: pycryptodome does also work - its an drop-in replacement for pycrypto

PR #1358, opened 2016-03-30
improve `obspy-runtests` stdout

megies:
As it can get confusing when hopping between environments and source trees and also system packages, I think we should print which file gets executed and also print the tested obspy version when running `obspy-runtests` on command line.


megies:
I'll merge this one soonish if nobody objects..

krischer:
;-P

PR #1362, opened 2016-03-31
sds client: add option to get all NSLC combinations on given point in time

megies:
Might still need a test..

### DIFF
--- a/obspy/clients/filesystem/sds.py
+++ b/obspy/clients/filesystem/sds.py
@@ -459,7 +459,7 @@ class Client(object):
         else:
             return False
 
-    def get_all_nslc(self, sds_type=None):
+    def get_all_nslc(self, sds_type=None, datetime=None):
         """
         Return information on what streams are included in archive.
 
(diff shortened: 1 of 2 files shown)


QuLogic:
Interesting; codecov is :arrow_up:, coveralls is :arrow_down:.

megies:
> Interesting; codecov is :arrow_up:, coveralls is :arrow_down:.

I also can't really follow codecov how this PR supposedly reduces the number of files in the repo by 2??

> @@            master   #1362   diff @@ ======================================



> Files 369 367 -2

codecov seems to have a miraculous way to compute overall coverage from the individual Travis reports.. (also: number of files covered is 367 for 6/7 Travis jobs and 369 for 1/7 jobs)

codecov

megies:
Added test, force pushed.

QuLogic:
> (also: number of files covered is 367 for 6/7 Travis jobs and 369 for 1/7 jobs)

For some reason, `obspy/db/scripts/__init__.py` and `obspy/db/scripts/indexer.py` are included in 4535.3, but not the others. However, neither of them get any hits, so I it's a bug in counting untouched files, but I don't know if it's coverage or codecov.

PR #1363, opened 2016-04-01
Raise Exception when filtering masked array and fill masked values with nan

MMesch:
This PR fixes a little issue that occured when involuntarily filtering masked data. This is not a well defined operation that should throw an Exception. I therefore added the raise_if_masked decorator to the filter functions in trace and stream. I also set all masked values to `np.nan` instead of `empty` now to avoid random behaviour due to unset data values.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,6 +3,11 @@
    * Added workaround for numpy issue where many FFTs of various lengths fill
      a cache that never gets cleared, effectively creating a memory leak
      (see #1424).
+   * Trace.filter and Stream.filter don't work on masked arrays anymore because
+     it produced unpredictable results due to the un-initialized data-chunk.
+     The uninitialized masked gap is now also initialized to np.nan in case
+     of floating point data which and a consistent fill value in case of
+     integer data. (see #1363)
  - obspy.clients.fdsn:
    * Fixing issue with location codes potentially resulting in unwanted data
      to be requested. (see #1422)
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -153,6 +153,11 @@ def create_empty_data_chunk(delta, dtype, fill_value=None):
         dtype = native_str(dtype)
     if fill_value is None:
         temp = np.ma.masked_all(delta, dtype=np.dtype(dtype))
+        # fill with nan if float number and otherwise with a very small number
+        if issubclass(temp.data.dtype.type, np.integer):
+            temp.data[:] = np.iinfo(temp.data.dtype).min
+        else:
+            temp.data[:] = np.nan
     elif (isinstance(fill_value, list) or isinstance(fill_value, tuple)) \
             and len(fill_value) == 2:
         # if two values are supplied use these as samples bordering to our data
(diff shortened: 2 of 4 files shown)


obspy/core/util/base.py
```diff
@@ -153,6 +153,7 @@ def create_empty_data_chunk(delta, dtype, fill_value=None):
         dtype = native_str(dtype)
     if fill_value is None:
         temp = np.ma.masked_all(delta, dtype=np.dtype(dtype))
+        temp.data[:] = np.nan
```
krischer: This wont work for non-floating point dtypes I think. There is a doctest that tests this so CI should raise.

megies:
Since this is changing behavior a bit would be good to have a changelog..

krischer:
Its not really changing behaviour - filtering a masked array right now really has an unpredictable result - but still: a changelog would be nice :-)

megies:
> filtering a masked array right now really has an unpredictable result

so it *is* changing behavior.. :P

obspy/core/util/base.py
MMesch: what else should we put? Any idea?

MMesch:
haha yes. @krischer that test accepted this. Maybe numpy puts something else for ints etc.. ?

MMesch:
```
In [6]: data = np.arange(4)

In [7]: data.dtype
Out[7]: dtype('int64')

In [8]: data[:] = np.nan

In [9]: data
Out[9]: 
array([-9223372036854775808, -9223372036854775808, -9223372036854775808,
       -9223372036854775808])
```

MMesch:
ok `np.nan` for integers is not very well defined in numpy apparently...

http://stackoverflow.com/questions/12708807/numpy-integer-nan

CHANGELOG.txt
```diff
@@ -1,3 +1,8 @@
+1.0.x: 
+ - obspy.core:
+   * Trace.filter and Stream.filter don't work on masked arrays anymore because
+     which gave unpredictable results.
```
krischer: s/because which gave/because that resulted in

or something along these lines.

obspy/core/util/base.py
krischer: Just put zeros there. I mean it really only for tests and you want to avoid any C routines using the masked data which might contain anything. zeros should be pretty save. `np.ma.zeros` also exists.

megies:
ping @MMesch, please rebase on current `maintenance_1.0.x` and fix that `np.nan` thingy.

MMesch:
Any particular value we should set masked integer values to? I guess something very small or high is good?

krischer:
You could do something along these lines:

```python
if issubclass(temp.data.dtype.type, np.integer):
    temp.data[:] = np.iinfo(temp.data.dtype).min
else:
    temp.data[:] = np.nan
```

krischer:
A benefit is that tampering with this data might then quite easily generate over-or underflows which shows lots of warnings so one might easier notice is something is wrong.

MMesch:
This is ready now.

megies:
Felt free to rebase and squash. Merging now.

PR #1329, opened 2016-03-11
new colormap: viridis_white for print plots

MMesch:
This little addition adds a new colormap "viridis_white" to obspy.imaging.cm. It is essentially a modified version of viridis that remains perceptually uniform but goes to white instead of yellow in the end. This helps to merge images with a white background and is useful for printing. There is also a tiny new functionality to read ListedColormaps that are stored in imaging/data/name.npy. 

viridis_white

PR #1365, opened 2016-04-02
add colormap comparison plots

megies:
..in imaging.cm module page and in gallery.

Mostly with new colormaps of #1329 in mind.
Duplicating the code for the plots in imaging.cm is a bit ugly, but it's easier to implement this way. Also this makes building docs a bit slower (more plots, some more http requests since data is on our examples server), obviously, but the gain outweighs this drawback I think.

The helper routines should probably be moved in another module (e.g. `imaging/cm_util.py`?) to not clutter `cm` too much.

Opinions?

https://www.geophysik.uni-muenchen.de/~megies/docs-build-colormaps/gallery.html#colormap-comparison
https://www.geophysik.uni-muenchen.de/~megies/docs-build-colormaps/packages/autogen/obspy.imaging.cm.html#module-obspy.imaging.cm

### DIFF
--- a/misc/docs/source/gallery.rst
+++ b/misc/docs/source/gallery.rst
@@ -244,6 +244,64 @@ Gallery
     :alt: SDS html report
     :scale: 50%
 
+Colormap comparison
+===================
+
+.. gallery-plot::
+    :target: packages/autogen/obspy.imaging.cm.html
+    :alt: Colormap comparisons
+
+    from obspy.imaging.cm import _colormap_plot_overview
+    _colormap_plot_overview()
+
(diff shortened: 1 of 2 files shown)


MMesch:
looks nice! In the long term, matplotlib style sheets could be used to organise color etc. It would make the plots uniform and maybe we can maybe even manage to make obspy-style print-ready plots that have the correct text size for A4 papers (at least for some standard problems). That's a longer perspective though but the marketing aspect in case it works is not to be underestimated :)

megies:
No alternative suggestions so merging. Can always be changed/improved later on.

PR #1366, opened 2016-04-05
obspy-scan: plot and print overlaps

megies:
This adds plotting (blue bars, like the red bars for gaps) and also printing (with option `"--print-gaps"`) of overlaps.

 - slight change in handling of endtimes: use endtime+delta for end of time segments -> cleaner handling of gaps/overlaps
 - fix `compress_start_end` helper function: overlaps were actually always merged, even with `merge_overlaps=False`
 - increase coverage by testing stdout for option `--print-gaps`
 - save obspy version in obspy-scan npz output


megies:
Rebased. No objections so far, gonna merge this when CI passes.

krischer:
I'll do it. I was gonna object to the ugly colors ;-) But matplotlib 2.0 will change all styles in any case so we can revamp our plots when its released.

megies:
> I was gonna object to the ugly colors ;-)

Only color I added is ordinary blue.. ;-) But yeah.. getting rid of these old mpl colors will be good to see.

PR #1367, opened 2016-04-05
helper function for comparison of obspy version strings

megies:
This would add a helper function to compare obspy version strings. I'm not sure if we really need it and so I'm also not sure if this should be added at all. It could reduce code duplication and unify some manual version number checks (e.g. https://github.com/obspy/obspy/pull/1366/commits/4bb48148e729deb8ec56066c78193a72f8bd50c8#diff-afebd6e04b7e5f6d37bdd55ebfed9c46R177) in the code base here and there I guess..
I would only consider merging this if more people think something like this could be useful and should be added.

### DIFF
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -195,5 +195,82 @@ def _normalize_version(version):
     return version
 
 
+def cmp_obspy_version(x, y):
+    """
+    Compare two ObsPy version number strings.
+
+    Similar to python builtin :py:func:`cmp()`.
+    The return value is negative if ``x < y``, zero if ``x == y`` and
+    strictly positive if ``x > y``.
+
+    The return type is ``int`` for comparisons that only involve well defined
+    release (or release candidate) versions. The return type is ``float`` for
+    comparisons that involve at least one development version.
+    In case of two development versions based on the same point release (e.g.
+    "1.0.1.post0+62.g3718497307.dirty" and "1.0.1.post0+55.gcdc54dbd9a")
+    the return value is also zero (type ``float``).
+
+    >>> a = "1.0.0"
+    >>> b = "1.0.1"
+    >>> c = "1.0.1.post0+62.g3718497307.dirty"
+    >>> d = "1.0.1.post0+55.gcdc54dbd9a"
+    >>> cmp_obspy_version(a, b)
+    -1
+    >>> cmp_obspy_version(a, c)
+    -1.0
+    >>> cmp_obspy_version(b, c)
+    -1.0
+    >>> cmp_obspy_version(b, b)
+    0
+    >>> cmp_obspy_version(c, d)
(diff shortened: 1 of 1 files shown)


obspy/core/util/version.py
```diff
@@ -195,5 +195,82 @@ def _normalize_version(version):
...
+        result = -1
+    elif _x == _y:
+        result = 0
+    elif _x > _y:
+        result = 1
+
+    # change result type to float if at least one development version is
+    # involved
+    if len(x) == 4 or len(y) == 4:
+        result = float(result)
```
krischer: This is IMHO a bit awkward. What is the use case for that?

obspy/core/util/version.py
```diff
@@ -195,5 +195,82 @@ def _normalize_version(version):
...
+        return tuple(map(int, version_string.split(".", 3)[:3]) +
+                     version_string.split(".", 3)[3:])
+    x = _version_tuple(x)
+    y = _version_tuple(y)
+
+    # when comparing two development versions: only compare the point release
+    # both are based on
+    if len(x) == 4 and len(y) == 4:
+        _x = x[:3]
+        _y = y[:3]
```
krischer: Is that sensible? I mean a later dev version is somehow larger. When is this needed?

We could also have a global `obspy.__version_tuple__` variable that we initialize right after the `__version__` variable.

krischer: Maybe even a namedtuple with `major`, `minor`, `micro`, and `dev` fields. Python has this:
```python
In [4]: sys.version_info
Out[4]: sys.version_info(major=3, minor=5, micro=1, releaselevel='final', serial=0)
```

obspy/core/util/version.py
megies: In general this doesn't hurt if somebody is not interested in the details for a "normal" check like `cmp_obspy_version(x, y) < 0`  or `... == 0`.

But there might be use cases where it's important to know if there's a dev version involved, e.g. when comparing two dev versions of the same point release (they can only be compared equal, but of course are not strictly equal), see below.

obspy/core/util/version.py
megies: I like the version tuple thing.

> Is that sensible? I mean a later dev version is somehow larger. When is this needed?

Dev versions in many cases will live on different branches and histories. So just comparing the number of commits just doesn't cut it, we would first have to make sure the to commits share the same history, so heavy git lifting involved. And in most cases (different histories) the comparison is just not well defined.. so I thought the only thing we can say in general about same is that they are "approximately" equal..

QuLogic:
How is this different from:
```python
>>> from distutils.version import LooseVersion
>>> LooseVersion('1.0.0b') > '1.0.0'
True
```
and why don't we use it more?

PR #1373, opened 2016-04-15
FDSN mass_downloader channel priorities ignore low gain.

krischer:
Suggested patch to [obspy.clients.fdsn.mass_downloader.restrictions](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/mass_downloader/restrictions.py#L198) to change default channel priorities to include low-gain (\*L\*) seismometers.

I think this would help make the mass_downloader more general, so you can download all stations from a network by not setting the channel restriction.

I edited my version to change lines 198-200 to:

`                 channel_priorities=("H[HL][ZNE]", "B[HL][ZNE]",
                                     "M[HL][ZNE]", "E[HL][ZNE]",
                                     "L[HL][ZNE]"),`

Happy to make a pull request for this if you want.

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/download_helpers.py
+++ b/obspy/clients/fdsn/mass_downloader/download_helpers.py
@@ -349,19 +349,29 @@ class Station(object):
         of the download helpers does not allow for a StationXML file with no
         data.
         """
+        from obspy.io.mseed.util import get_start_and_end_time
         # All or nothing for each channel.
         for id in self.miss_station_information.keys():
-            logger.warning("No station information could be downloaded for "
-                           "%s.%s.%s.%s. All downloaded MiniSEED files "
+            logger.warning("Station information could not be downloaded for "
+                           "%s.%s.%s.%s. MiniSEED files outside of the "
+                           "station information period "
                            "will be deleted!" % (
                                self.network, self.station, id[0], id[1]))
             channel = [_i for _i in self.channels if
                        (_i.location, _i.channel) == id][0]
             for time_interval in channel.intervals:
+                # Check that the time_interval.start and end are correct!
+                time_interval.start, time_interval.end = \
+                    get_start_and_end_time(time_interval.filename)
                 # Only delete downloaded things!
                 if time_interval.status == STATUS.DOWNLOADED:
-                    utils.safe_delete(time_interval.filename)
-                    time_interval.status = STATUS.DOWNLOAD_REJECTED
+                    # Only delete if the station data are actually missing
+                    # for this time
+                    miss_start, miss_end = self.miss_station_information[id]
+                    if miss_start <= time_interval.start <= miss_end and \
+                       miss_start <= time_interval.end <= miss_end:
+                        utils.safe_delete(time_interval.filename)
+                        time_interval.status = STATUS.DOWNLOAD_REJECTED
 
 
 class Channel(object):
@@ -695,7 +705,17 @@ class ClientDownloadHelper(object):
                 starttime = min([_i.starttime for _i in c_info])
                 endtime = max([_i.endtime for _i in c_info])
                 if starttime > times[0] or endtime < times[1]:
-                    still_missing[c_id] = times
+                    # Cope with case that not full day of station info missing
+                    if starttime < times[1]:
+                        still_missing[c_id] = (times[0], starttime)
+                        station.have_station_information[c_id] = (starttime,
+                                                                  times[1])
+                    elif endtime > times[0]:
+                        still_missing[c_id] = (endtime, times[1])
+                        station.have_station_information[c_id] = (times[0],
+                                                                  endtime)
+                    else:
+                        still_missing[c_id] = times
                     continue
                 station.have_station_information[c_id] = times
 
--- a/obspy/clients/fdsn/tests/test_mass_downloader.py
+++ b/obspy/clients/fdsn/tests/test_mass_downloader.py
@@ -303,7 +303,8 @@ class DownloadHelpersUtilTestCase(unittest.TestCase):
         c2 = Channel("10", "SHE", time_intervals)
         c3 = Channel("00", "BHZ", time_intervals)
         c4 = Channel("", "HHE", time_intervals)
-        channels = [c1, c2, c3, c4]
+        c5 = Channel("", "ELZ", time_intervals)
+        channels = [c1, c2, c3, c4, c5]
 
         filtered_channels = filter_channel_priority(
             channels, key="channel", priorities=[
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,9 @@
  - obspy.clients.fdsn:
    * Fixing issue with location codes potentially resulting in unwanted data
      to be requested. (see #1422)
+   * Included low-gain seismometers in default channel filters in
+     mass-downloader, also included non-oriented channels by default
+     (see #1373).
  - obspy.db:
    * Fixed a bug in obspy-indexer command line script (see #1369,
      command line script was not working, probably since 0.10.0)
--- a/obspy/clients/fdsn/mass_downloader/restrictions.py
+++ b/obspy/clients/fdsn/mass_downloader/restrictions.py
@@ -139,9 +139,11 @@ class Restrictions(object):
                  network=None, station=None, location=None, channel=None,
                  reject_channels_with_gaps=True, minimum_length=0.9,
                  sanitize=True, minimum_interstation_distance_in_m=1000,
-                 channel_priorities=("H[HL][ZNE]", "B[HL][ZNE]",
-                                     "M[HL][ZNE]", "E[HL][ZNE]",
-                                     "L[HL][ZNE]"),
+                 channel_priorities=("HH[ZNE12]", "BH[ZNE12]",
+                                     "MH[ZNE12]", "EH[ZNE12]",
+                                     "LH[ZNE12]", "HL[ZNE12]",
+                                     "BL[ZNE12]", "ML[ZNE12]",
+                                     "EL[ZNE12]", "LL[ZNE12]"),
                  location_priorities=("", "00", "10")):
         self.starttime = obspy.UTCDateTime(starttime)
         self.endtime = obspy.UTCDateTime(endtime)
--- a/obspy/clients/fdsn/mass_downloader/restrictions.py
+++ b/obspy/clients/fdsn/mass_downloader/restrictions.py
@@ -139,9 +139,9 @@ class Restrictions(object):
                  network=None, station=None, location=None, channel=None,
                  reject_channels_with_gaps=True, minimum_length=0.9,
                  sanitize=True, minimum_interstation_distance_in_m=1000,
-                 channel_priorities=("HH[ZNE]", "BH[ZNE]",
-                                     "MH[ZNE]", "EH[ZNE]",
-                                     "LH[ZNE]"),
+                 channel_priorities=("H[HL][ZNE]", "B[HL][ZNE]",
+                                     "M[HL][ZNE]", "E[HL][ZNE]",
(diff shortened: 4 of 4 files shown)


krischer:
I took the liberty to convert this issue into a PR with your changes. After this is merged you will have to synchronize your maintenance branch with ours.

I'm totally fine with this. It would currently download high-and low-gain channels together if they exist simultaneously. Does that ever happen and is this a problem in that case?

We could also first try to download the high-gain and then the low-gain sensors:

```python
channel_priorities=("HH[ZNE]", "BH[ZNE]", "MH[ZNE]", "EH[ZNE]", "LH[ZNE]",
                    "HL[ZNE]", "BL[ZNE]", "ML[ZNE]", "EL[ZNE]", "LL[ZNE]"),
```

I'm not sure which is the preferable solution here. Also do we want to merge this into maintenance or the master? It changes behavior a bit. 

Can you please also add an item to the `CHANGELOG.txt` in the root folder?

calum-chamberlain:
I'm not sure about whether it would be a problem, it would be strange to have the channels from two sensors downloaded together, which is what I assume the most common case would be for a low and high gain stream for one site - I think separating them out would be clearer.

On another note, I'm archiving data with IRIS at the moment, and my data have uncertain orientations, so they are being archived as [Z12], from talking with them it sounds like they are pushing to use this convention as the default for temporary deployments now.

Might it be worth doing:

```python
channel_priorities=("HH[ZNE12]", "BH[ZNE12]", 
                    "MH[ZNE12]", "EH[ZNE12]",
                    "LH[ZNE12]", "HL[ZNE12]",
                    "BL[ZNE12]", "ML[ZNE12]",
                    "EL[ZNE12]", "LL[ZNE12]"),
```
?

krischer:
Why not I guess. I originally excluded the `[12]` channels on purpose as I thought it might confuse people but we have functionality in ObsPy to work with these so it should be good. Can you do the changes in this PR? Otherwise I'll do it in a separate one.

calum-chamberlain:
Added.

calum-chamberlain:
@krischer I also noticed an issue when downloading data at the start or end of deployments when using the `sanitize=True` restriction.  In this case the station data may not have start times at the start of a day (nor will the data), but the data and station information may actually correspond.  To avoid this I have added more precision to the re-evaluation of missing station data [here](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/mass_downloader/download_helpers.py#L700), changed to:
```python
                if starttime > times[0] or endtime < times[1]:
                    # Cope with case that not full day of station info missing
                    if starttime < times[1]:
                        still_missing[c_id] = (times[0], starttime)
                        station.have_station_information[c_id] = (starttime,
                                                                  times[1])
                    elif endtime > times[0]:
                        still_missing[c_id] = (endtime, times[1])
                        station.have_station_information[c_id] = (times[0],
                                                                  endtime)
                    else:
                        still_missing[c_id] = times
                    continue
```

And I added more precise checking in [sanitize_downloads](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/mass_downloader/download_helpers.py#L344) so that only time intervals that actually have missing station information, rather than all time intervals for that channel are removed:

```python
        from obspy.io.mseed.util import get_start_and_end_time
        # All or nothing for each channel.
        for id in self.miss_station_information.keys():
            logger.warning("Station information could not be downloaded for "
                           "%s.%s.%s.%s. MiniSEED files outside of the "
                           "station information period "
                           "will be deleted!" % (
                               self.network, self.station, id[0], id[1]))
            channel = [_i for _i in self.channels if
                       (_i.location, _i.channel) == id][0]
            for time_interval in channel.intervals:
                # Check that the time_interval.start and end are correct!
                time_interval.start, time_interval.end = \
                    get_start_and_end_time(time_interval.filename)
                # Only delete downloaded things!
                if time_interval.status == STATUS.DOWNLOADED:
                    # Only delete if the station data are actually missing
                    # for this time
                    miss_start, miss_end = self.miss_station_information[id]
                    if miss_start <= time_interval.start <= miss_end and \
                       miss_start <= time_interval.end <= miss_end:
                        utils.safe_delete(time_interval.filename)
                        time_interval.status = STATUS.DOWNLOAD_REJECTED
```

Do you want me to include that in this pull request?

Note that this issue can be reproduced using:
```python
from obspy.clients.fdsn.mass_downloader import CircularDomain,\
   Restrictions, MassDownloader
from obspy import UTCDateTime

domain = CircularDomain(latitude=0, longitude=0, minradius=0, maxradius=180)
restrictions = Restrictions(starttime=UTCDateTime(2012, 01, 18),
                            endtime=UTCDateTime(2012, 01, 19),
                            chunklength_in_sec=86400, network='ZT', 
                            station='WZ01', reject_channels_with_gaps=False, 
                            minimum_length=0.0,
                            minimum_interstation_distance_in_m=0.0,
                            sanitize=True)
mdl = MassDownloader(providers=['IRIS'])
mdl.download(domain, restrictions, mseed_storage='.', 
             stationxml_storage='.', threads_per_client=3)
```

megies:
Was this closed on purpose? @krischer?

obspy/clients/fdsn/tests/test_mass_downloader.py
```diff
@@ -303,7 +303,8 @@ def test_channel_priority_filtering(self):
         c2 = Channel("10", "SHE", time_intervals)
         c3 = Channel("00", "BHZ", time_intervals)
         c4 = Channel("", "HHE", time_intervals)
-        channels = [c1, c2, c3, c4]
+        c5 = Channel("", "ELZ", time_intervals)
+        channels = [c1, c2, c3, c4, c5]
```
krischer: Did you mean to add some additional checks below this? Just adding one more channel does not add a lot to this particular test.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -349,19 +349,29 @@ def sanitize_downloads(self, logger):
         of the download helpers does not allow for a StationXML file with no
         data.
         """
+        from obspy.io.mseed.util import get_start_and_end_time
```
krischer: Can you move the import to the top of the file?

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -349,19 +349,29 @@ def sanitize_downloads(self, logger):
...
+                # Check that the time_interval.start and end are correct!
+                time_interval.start, time_interval.end = \
+                    get_start_and_end_time(time_interval.filename)
                 # Only delete downloaded things!
                 if time_interval.status == STATUS.DOWNLOADED:
-                    utils.safe_delete(time_interval.filename)
-                    time_interval.status = STATUS.DOWNLOAD_REJECTED
+                    # Only delete if the station data are actually missing
+                    # for this time
+                    miss_start, miss_end = self.miss_station_information[id]
```
krischer: I just had to reread a lot of the code to understand this. So we might have to expand `miss_station_information` a bit to fully get this to work.

`miss_station_information` and its pendant `have_station_information` are per-channel time intervals. It has no concept of partially downloaded channel information - data for each channel is either considered to be completely downloaded or not - I agree that this should be fixed somehow. These things happen in the `download_stationxml` method.

The current solution in this PR will (I think) delete time intervals that are completely inside the complete requested time interval and ignore those at the end as soon as some piece of station information for that particular channel has not been downloaded.

What we should do:

Give the station objects an attribute called `available_station_intervals` (set after the StationXML files have been downloaded) and use that to decide whether or not to discard already downloaded MiniSEED files.

Is my interpretation correct or am I missing something?

krischer: I just saw that this PR actually already contains pieces of what I just proposed...

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -696,7 +706,17 @@ def star_download_station(args):
...
+                    if starttime < times[1]:
+                        still_missing[c_id] = (times[0], starttime)
+                        station.have_station_information[c_id] = (starttime,
+                                                                  times[1])
+                    elif endtime > times[0]:
+                        still_missing[c_id] = (endtime, times[1])
+                        station.have_station_information[c_id] = (times[0],
+                                                                  endtime)
+                    else:
+                        still_missing[c_id] = times
```
krischer: This is actually already along the lines I proposed above before I fully read your PR - sorry about that.

It currently does not deal with the case if information at the beginning and end is missing. I think this might be easiest to implement with a new attribute as I proposed above but maybe this is another way.

krischer:
I totally forgot about this PR - I was on leave for a while. It actually looks pretty good and we definitely want it as it deals with some juicy edge cases! I wrote some inline comments.

calum-chamberlain:
Great, sorry I didn't overly mean to close this - little slow-minded yesterday.  Will have a look at things again today (NZtime).

obspy/clients/fdsn/tests/test_mass_downloader.py
calum-chamberlain: Yes, yes I did - I have only added a small test for those low-gain cases though now. It will show up later when I push to this.

obspy/clients/fdsn/mass_downloader/download_helpers.py
calum-chamberlain: This was my simple way - but I think an attribute could be better. Can't think of a simple way to cope with beginning and end missing for a single channel - that would be the case where there is station info for the middle of a day or data, but not either end... Would be an odd case, but for very short-term deployments it might happen?

calum-chamberlain: I think that case might be handled by the fact that station.miss_station_information['times'] would have the start and end times for each miss separately - not totally sure though.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -349,19 +349,29 @@ def sanitize_downloads(self, logger):
...
         """
+        from obspy.io.mseed.util import get_start_and_end_time
         # All or nothing for each channel.
         for id in self.miss_station_information.keys():
-            logger.warning("No station information could be downloaded for "
-                           "%s.%s.%s.%s. All downloaded MiniSEED files "
+            logger.warning("Station information could not be downloaded for "
+                           "%s.%s.%s.%s. MiniSEED files outside of the "
+                           "station information period "
                            "will be deleted!" % (
```
QuLogic: On a side note, it is not necessary to `%`-format the string here, because `logger.warning` will take its positional arguments (after the first) and `%`-format them into the message string.

megies:
@calum-chamberlain can you please rebase on current `maintenance_1.0.x` or give me push rights to your fork, so I can do it?

megies:
Rebased + force-pushed.

megies:
Anything else needed here, or good to merge, @krischer @calum-chamberlain ?

calum-chamberlain:
Nothing else I wanted to add - only thing would be whether you guys think the small test I put in is sufficient?

krischer:
@calum-chamberlain Can you also give me push rights? I'd like to add one or two things.

calum-chamberlain:
@krischer Done - in the future I will work on things in the obspy developers group.  I should start working on the seisan s-file things in there soon.

krischer:
I'm merging this for now - for 1.0.3 I'll have to spend some more time on this which I don't have right now. But this PR is definitely an improvement.

seisman:
I think short period instruments `SH[ZNE12]` should also be included into channel priorities.

megies:
> I think short period instruments SH[ZNE12] should also be included into channel priorities.

Since this one is merged already, might be better to start a new issue..

krischer:
Done in https://github.com/obspy/obspy/pull/1687

PR #1132, opened 2016-04-18
Missing default encoding on MacOS causing problems

petrrr:
This is just to keep track of problems on OS X due to a seemingly missing default system encoding, see mail on users mailing list.

http://lists.swapbytes.de/archives/obspy-users/2015-September/001824.html
http://lists.swapbytes.de/archives/obspy-users/2015-September/001797.html

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,8 @@
+1.0.2: (doi: 10.5281/zenodo.49636)
+ - obspy.db:
+   * Fixed a bug in obspy-indexer command line script (see #1369,
+     command line script was not working, probably since 0.10.0)
+
 1.0.1: (doi: 10.5281/zenodo.48254)
  - General:
    * Some methods might have unnecessarily upcasted float32 arrays to float64.
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -124,7 +124,7 @@ def call_git_describe(abbrev=10, dirty=True,
 
 def read_release_version():
     try:
-        with io.open(VERSION_FILE, "rt") as fh:
+        with io.open(VERSION_FILE, "rt", encoding="ascii") as fh:
             version = fh.readline()
         return version.strip()
     except IOError:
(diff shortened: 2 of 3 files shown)


petrrr:
There is an issue with this PR: I branched it of from `maintenance_1.0.x` (correctly), but when converting the issue to a PR, based it on `master`. Therefore, PR #1375  appears here. I have not found a way to change the PR's base, so there are probably two: possibilities:
 * close this and create a corrected PR
 * base this PR on master and then apply it maintenance as well
Any advices?

megies:
> Has anyone an idea how to reproduce this, so that we can make a test for it?

I can't reproduce this by unsetting all locale variables on my system, but you could use `mock` to "synthesize" this issue. Patching `locale.getdefaultlocale()` to `(None, None)` doesn't have an effect for me, though, while patching `locale.getpreferredencoding()` to `None` shows a different error than was reported...

```python
import mock

with mock.patch("locale.getpreferredencoding") as p:
    p.return_value = None
    import obspy
```

```
  File "/home/megies/git/obspy/obspy/__init__.py", line 50, in <module>
    __version__ = _get_version_string(abbrev=10)
  File "/home/megies/git/obspy/obspy/core/util/version.py", line 145, in get_git_version
    release_version = read_release_version()
  File "/home/megies/git/obspy/obspy/core/util/version.py", line 127, in read_release_version
    with io.open(VERSION_FILE, "rt") as fh:
TypeError: bad argument type for built-in operation
```

Of course in the tests the mocking would wrap around the actual routine called (i.e. `read_release_version()`)

```python
from obspy.core.util.version import read_release_version
import mock

with mock.patch("locale.getpreferredencoding") as p:
    p.return_value = None
    read_release_version()
```

> Any advices?

As @QuLogic stated on gitter.. when this is ready you can merge it into `maintenance` locally and push it and close this one.

petrrr:
@megies: Not, sure but I somehow recall that for this error the result was not `None` but an empty string. However, was not able to spot any evidence for this. 
With:

    p.return_value = ''

But I cannot reproduce this neither. However, when importing obspy, I get a

    LookupError: unknown encoding

in matplotlib.

petrrr:
One more side-note: 
I had a quick look through the code, and there are several places where the encoding is not specified. So this one might only be the first hit of a relatively rare issue.

megies:
> LookupError: unknown encoding

Indeed, that looks like the original report on the users mailing list.

Issue #1382, opened 2016-04-25
Plotting channel name in "section" waveform plots


amichelini:
Hi Folks,
first of all thanks for the wonderful job you have been doing with ObsPy. 
I am not notifying a bug.
I write you since I have been playing with a simple (though general) script that makes section plots of data downloaded from a fdsn client. This can be very useful for e.g. promptly visualizing the traces after an earthquake with known  event_id.
One thing that I have not been able to figure out is how to maintain next to each of the traces the actual channel name (e.g., IV.BSSO..BHZ). This is important to keep track of the stations. Is it possible at all ?
I have searched through the previous issues but haven't found anything.

Obspy version:
obspy                     1.0.1                    py27_0    obspy

I am using the standard command 

```
st.plot(type='section', plot_dx=20e3, recordlength=100,
        time_down=True, linewidth=.75, grid_linewidth=.25, orientation='horizontal')
```

am I missing some parameter ???

Thank you in advance for the help

Alberto

megies:
Hi Alberto,

I've never used the section plot myself. Looking at the source code, if you set `st.plot(..., color="station")` you can get a legend with station names but also each trace has a different color, and I think that's probably not what you want.

I assume you're referring to station code (or "network.station") as labels on the traces? It seems no option for this is available. But, I think if you want more customization, you can either..
 * continue to adapt the section plot afterwards (with setting `show=False`), or
 * build the plot from scratch yourself exactly how you want it

Honestly, I think in many cases when one of our builtin plots doesn't match one's personal needs, it's easiest to just do the plot from scratch. It's just not possible to cover all potential needs and wishes, and the our plotting routines are already pretty convoluted and laden with many, many options.

But, customizing the plot should also be pretty simple.. however, I noticed some significant strangeness in how the section plot is set up, namely that the offset axis has some normalization that makes labelling the traces afterwards very hard. I'm proposing a fix for the next bug fix release with #1383. Also see the updated Tutorial example for doing customization on the plot: https://github.com/obspy/obspy/pull/1383/files#diff-e1b32bb704263a9eaca155fc58c64f7e

amichelini:
Dear Tobias,
thank you for the explanation and the time spent on all this. I really appreciate it!
Yes, in the end it may be best to write our own modules but if there is already a lot done like in this case I believe it is better to build on someone's else developments rather than starting from scratch  which may be too cumbersome and time consuming.
[...]

PR #1383, opened 2016-04-26
Fix section plot offset-axis units

megies:
This PR fixes the strange arbitrary, normalized data units on section plot's offset axis, that effectively prevent adapting the plot with more customizations afterwards (e.g. station labels etc.), see #1382.

I've also added some more code to the Tutorial example, demonstrating how to customize the plot afterwards..


megies:
CC @miili, since I think you implemented the section plot and I thus assume that you're one of the heavy users of it.

misc/docs/source/tutorial/code_snippets/waveform_plotting_tutorial_6.py
```diff
@@ -32,6 +34,19 @@
...
-# Plot
+
+# Do the section plot..
+# If no customization is done after the section plot command, figure
+# initialization can be left out and also option ".., show=False, fig=fig)" can
+# be omitted, and figure is shown automatically
+fig = plt.figure()
 st.plot(type='section', plot_dx=20e3, recordlength=100,
-        time_down=True, linewidth=.25, grid_linewidth=.25)
+        time_down=True, linewidth=.25, grid_linewidth=.25, show=False, fig=fig)
```
QuLogic: Instead of `show=False, fig=fig`, you could do `fig = st.plot(..., handle=True)` since you don't really do anything with the figure before hand.

QuLogic:
I guess changing some of the internals is a bit of a grey area, but if you change what one of the parameters means (`plot_dx`), then I'm not sure this belongs in a bugfix.

megies:
> I guess changing some of the internals is a bit of a grey area, but if you change what one of the parameters means (plot_dx), then I'm not sure this belongs in a bugfix.

You're right, need to look at the details of this one again..

megies:
> but if you change what one of the parameters means (plot_dx), then I'm not sure this belongs in a bugfix.

Actually, looking at it again, the docstring was just wrong earlier. It stated that `plot_dx` was in `km` when the example in the tutorial was passing in `plot_dx=20e3`, so `m`.

The `tr.stats.distance` that is used internally also seems to be in `m` throughout it's usages in our code base.

So I think this is really just about internal changes, outwards API is unchanged, I think.

Or were you referring to something else?

megies:
> CC @miili, since I think you implemented the section plot and I thus assume that you're one of the heavy users of it.

@miili can you give this PR a review?

megies:
+DOCS let's see what the tutorial image looks like..

QuLogic:
Seems to work okay, though maybe could do with a bit of padding.

megies:
Yeah, but it was like this before. I really don't know where to best change this and I'll leave it as is. Can always be changed by user after plotting.

megies:
Sanitized the ticks as well now. When no `plot_dx` is given, just leave it to matplotlib, when `plot_dx` *is* given, properly arange ticks around zero and with correct spacing, both was not the case before.

megies:
Good to merge, I think..

Pbottelin:
Hi everyone,
I re-open this issue as I found a strange behavior in st.plot(type='section').

I do not understand why the trace.stats.distance (in m) are converted into km in the section plot. This makes no sense for short seismic arrays (e.g. small scale active seismics) and makes difficult adding subplot using trace.stats.distance (in m, see picture below).

Might it be possible to add a kwarg like 'unit'='m' or 'km' to control this behavior?
(I fixed this issue once by modifying plot_section.py routine, but it disappears at every obspy update).

The lines are 1137-1138 and 1209-1212 in obspy/imaging/waveform.py

Thank you for your amazing and useful work,
best regards

Pierre
section_example

megies:
Please open a new issue.

PR #1386, opened 2016-04-27
mseed recordanalyzer: always displays blockette information of first record

megies:
Obspy version 1.0.1-1~wheezy, Debian Wheezy, installed from obspy .deb repository

The file demo.mseed in attachment contains 2 miniseed records of 512 bytes. First one has bloquette 1001 with timing quality set to 10 and second one has bloquette 1001 with timing quality set to  20.

Using recordanalyzer.py to display the timing quality of both records will always show "Timing quality: 10".

The bug is in function  `def _get_blockettes(self)`  where a seek is done from the beginning of the file, but the seek displacement you pass is a value which is the offset of the bloquette relative to the position of the analyzed record. In the seek displacement you forgot to take into account the position of the record with respect to the beginning of the file. Patch is attached.

[demo.mseed.zip](https://github.com/obspy/obspy/files/238686/demo.mseed.zip)
[recordanalyzer_patch.txt](https://github.com/obspy/obspy/files/238687/recordanalyzer_patch.txt)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,8 @@
  - obspy.db:
    * Fixed a bug in obspy-indexer command line script (see #1369,
      command line script was not working, probably since 0.10.0)
+ - obspy.io.mseed:
+   * Fixed a bug in obspy-mseed-recordanalyzer (see #1386)
 
 1.0.1: (doi: 10.5281/zenodo.48254)
  - General:
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -45,6 +45,7 @@ Alberto Michelini
 Bernhard Morgenstern
 Nathaniel C. Miller
 Mark P. Panning
+Giovanni Rapagnani
 Celso Reyes
 Adam Ringler
 Nicolas Rothenhäusler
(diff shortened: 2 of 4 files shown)


megies:
Thanks for the patch! I felt free to add you to our contributors list..

grapagnani:
I thank you and your team for this great piece of code which obspy is.
Best Regards,
Giovanni

megies:
This piece was actually tested, so the test or the tested data may not be good enough..
https://codecov.io/gh/obspy/obspy/src/33e523da5190df1bddb1efc30519c0641e74a560/obspy/io/mseed/scripts/recordanalyzer.py#L212

grapagnani:
I am not familiar with the testing process.
Are the tests performed with the scripts inside obspy/io/mseed/tests?

megies:
Yes, this command line script is tested here: https://github.com/obspy/obspy/blob/master/obspy/io/mseed/tests/test_recordanalyzer.py

grapagnani:
test_recordanalyzer.py  uses the file data/test.mseed. The 2 records
inside test.mseed differ only in "the fixed section of data header". On
the other hand, there is a file data/timingquality.mseed for which there
are differences also in the blockettes information.

Would it make sense to do the test against timingquality.mseed file
instead of test.mseed?

megies:
Yes, sounds like a good plan!

grapagnani:
patch in attachment.
[test_recordanalyzer.diff.txt](https://github.com/obspy/obspy/files/240610/test_recordanalyzer.diff.txt)

megies:
Good to merge I guess, but I'll leave it open for @krischer for review. No hurry right now.

krischer:
Good find and nice patch!

PR #1388, opened 2016-05-03
Try to set SAC distances on SACTrace.read

jkmacc-LANL:
SAC users expect that files read in will have distances calculated/set if "lcalda" is True.  This small fix meets that expectation.  If "dist" is found, it isn't overwritten.

EDIT: notify @claudiodsf , @john-robert

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,7 @@
 1.0.2: (doi: 10.5281/zenodo.49636)
+ - obspy.io.sac:
+   * Try to set SAC distances (dist, az, baz, gcarc) on read, if "lcalda" is
+     true.  If "dist" header is found, distances aren't calculated.
  - obspy.db:
    * Fixed a bug in obspy-indexer command line script (see #1369,
      command line script was not working, probably since 0.10.0)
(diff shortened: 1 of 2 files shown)


claudiodsf:
Good to merge, from my point of view.

megies:
Does this need a changelog? Without having too much insight in the details, I have the impression that these small changes we're having in `io.sac` in maintenance might be important to know for the people using sac..

claudiodsf:
If it's just regression fixing, I don't think we need a changelog.
I don't remember if autosetting distances was available in ObsPy < 1.0.0.

jkmacc-LANL:
@claudiodsf I couldn't remember either.  Just looked: [it did](https://github.com/obspy/obspy/blob/0.10.3/obspy/sac/sacio.py#L649).  This fixes a regression.

claudiodsf:
In this case I would merge this. @megies ?

megies:
Sure, go ahead. I would probably have put a changelog but I'm not into the sac stuff much so if you think it's not necessary that's fine with me. :+1:

jkmacc-LANL:
Changelog updated.

claudiodsf:
@jkmacc-LANL Maybe you can squash the two commits and force-push to reduce clutter in git history.

jkmacc-LANL:
Sure.  I've not done this before, so I hope it was done correctly.

claudiodsf:
Looks fine :+1:  Let's wait for tests and merge!

PR #1390, opened 2016-05-03
read_nlloc_hyp(): use GEOGRAPHIC line if no coordinate_converter is provided

claudiodsf:
If NonLinLoc location run was performed using one of the standard projections available in the NonLinLoc package, there is no need to provide a custom coordinate converter and the coordinates from the "GEOGRAPHIC" line are used.

- Added a test case
- Updated the doc

### DIFF
--- a/obspy/io/nlloc/__init__.py
+++ b/obspy/io/nlloc/__init__.py
@@ -15,11 +15,47 @@ formats.
 
 Example
 -------
+If NonLinLoc location run was performed using one of the standard projections
+available in the NonLinLoc package, it is straightforward to read a NonLinLoc
+Hypocenter-Phase file into an ObsPy :class:`~obspy.core.event.Catalog` object:
 
-First, if NonLinLoc location run was performed in some custom coordinate system
-(as opposed to WGS84 with depth in meters down), we need to set up a coordinate
-conversion function to convert from the NonLinLoc location coordinates `x`,
-`y`, `z` to longitude, latitude and depth in kilometers.
+>>> from obspy import read_events
+>>> cat = read_events("/path/to/nlloc.hyp")
+>>> print(cat)
+1 Event(s) in Catalog:
+2006-07-15T17:21:20.195670Z |  +7.737,  +51.658
+
+>>> event = cat[0]
+>>> print(event)  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
+Event:	2006-07-15T17:21:20.195670Z |  +7.737,  +51.658
+<BLANKLINE>
+       resource_id: ResourceIdentifier(id="smi:local/...")
+     creation_info: CreationInfo(creation_time=UTCDateTime(2013, 6, 21, ...),
+                                 version='NLLoc:v6.02.07')
+    ---------
+             picks: 5 Elements
+           origins: 1 Elements
+
+>>> origin = event.origins[0]
+>>> print(origin)  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
+Origin
+       resource_id: ResourceIdentifier(id="smi:local/...")
+              time: UTCDateTime(2006, 7, 15, 17, 21, 20, 1956...)
+         longitude: 51.657659...
+          latitude: 7.736781...
+             depth: 1433.5... [confidence_level=68, uncertainty=...]
+                ...
+     creation_info: CreationInfo(creation_time=UTCDateTime(2013, 6, 21, ...),
+                                 version='NLLoc:v6.02.07')
+    ---------
+          comments: 1 Elements
+          arrivals: 5 Elements
+
+
+If, instead, NonLinLoc location run was performed in some custom coordinate
+system (as opposed to WGS84 with depth in meters down), we need to set up a
+coordinate conversion function to convert from the NonLinLoc location
+coordinates `x`, `y`, `z` to longitude, latitude and depth in kilometers.
 In the example, the location run was done in Gauß-Krüger zone 4 (EPSG:31468,
 but in kilometers for coordinates) and depth in kilometers. So we have to
 convert `x` and `y` (of NonLinLoc location run) to meters (as defined in
@@ -42,8 +78,8 @@ Then, we can load the NonLinLoc Hypocenter-Phase file into an ObsPy
 function as `coordinate_converter` kwarg, which will be passed down to the
 low-level routine :func:`~obspy.io.nlloc.core.read_nlloc_hyp`.
 
->>> from obspy import read_events
->>> cat = read_events("/path/to/nlloc.hyp",
+>>> from obspy import read_events  # doctest: +SKIP
+>>> cat = read_events("/path/to/nlloc_custom.hyp",
 ...                  coordinate_converter=my_conversion)  # doctest: +SKIP
 >>> print(cat)  # doctest: +SKIP
 1 Event(s) in Catalog:
@@ -61,19 +97,18 @@ Event:  2010-05-27T16:56:24.612600Z | +48.047,  +11.646
 
 >>> origin = event.origins[0]  # doctest: +SKIP
 >>> print(origin)  # doctest: +SKIP
-    Origin
-                resource_id: ResourceIdentifier(id="smi:local/...")
-                       time: UTCDateTime(2010, 5, 27, 16, 56, 24, 612600)
-                  longitude: 11.64553754...
-                   latitude: 48.04707051...
-                      depth: 4579.4... [confidence_level=68,
- uncertainty=191.6063...]
-                        ...
-              creation_info: CreationInfo(creation_time=UTCDateTime(2014, 10,
-17, 16, 30, 8), version='NLLoc:v6.00.0')
-        ---------
-                   comments: 1 Elements
-                   arrivals: 8 Elements
+Origin
+       resource_id: ResourceIdentifier(id="smi:local/...")
+              time: UTCDateTime(2010, 5, 27, 16, 56, 24, 612600)
+         longitude: 11.64553754...
+          latitude: 48.04707051...
+             depth: 4579.4... [confidence_level=68, uncertainty=191.6063...]
+                ...
+     creation_info: CreationInfo(creation_time=UTCDateTime(2014, 10, 17,
+16, 30, 8), version='NLLoc:v6.00.0')
+    ---------
+          comments: 1 Elements
+          arrivals: 8 Elements
 """
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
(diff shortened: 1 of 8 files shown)


obspy/io/nlloc/core.py
```diff
@@ -138,17 +137,17 @@ def read_nlloc_hyp(filename, coordinate_converter=None, picks=None, **kwargs):
...
+        # maximum likelihood origin location in km info line
+        line = lines["HYPOCENTER"]
+        x, y, z = coordinate_converter(*map(float, line.split()[1:7:2]))
 
-    # origin time info line
+    # maximum likelihood origin time and location lon lat info line
     line = lines["GEOGRAPHIC"]
 
+    if not coordinate_converter:
+        x, y, z = map(float, line.split()[8:13:2])
```
megies: Can you please contract this to the following? Much cleaner I think..
```
if coordinate_converter:
    ...
else:
    ...
```

claudiodsf: The problem is that `line` is used later to retrieve the origin time.

obspy/io/nlloc/__init__.py
```diff
@@ -15,11 +15,47 @@
...
+If NonLinLoc location run was performed using one of the standard projections
+available in the NonLinLoc package, it is straightforward to read a NonLinLoc
+Hypocenter-Phase file into an ObsPy :class:`~obspy.core.event.Catalog` object:
 
-First, if NonLinLoc location run was performed in some custom coordinate system
-(as opposed to WGS84 with depth in meters down), we need to set up a coordinate
-conversion function to convert from the NonLinLoc location coordinates `x`,
-`y`, `z` to longitude, latitude and depth in kilometers.
+>>> from obspy import read_events
+>>> cat = read_events("/path/to/nlloc.hyp")  # doctest: +SKIP
```
megies: shouldnt this be `/path/to/rhur.hyp`?

claudiodsf: Does it matter?

obspy/io/nlloc/__init__.py
```diff
@@ -15,11 +15,47 @@
...
+>>> print(cat)  # doctest: +SKIP
```
megies: Actually, you can run all those doctests, they should work just fine. I was just skipping the other doctests because they were relying on pyproj and that's not a dependency of ours (although we have it installed in Travis for testing now).

claudiodsf: Ok, I'll fix that and change the file name.

megies:
Nice that you added the example for the docs! I think this should have a short changelog entry, since it actually changes behavior (to the better) when no coordinate converter is passed in.

Also, would you mind cropping your example event QuakeML file, dropping out all but maybe 2-3 picks/arrivals (and maybe the hyp accordingly)? Would help to keep the repo size down..

obspy/io/nlloc/__init__.py
megies: Well the example wouldnt give the results shown here if you load the other file..

claudiodsf:
@megies Ok, will do it and rebase/force push.

obspy/io/nlloc/core.py
megies: You can just add another `line = lines["GEOGRAPHIC"]` right before the time is read, no?

claudiodsf:
Force pushed: I renamed the existing example to `nlloc_custom.hyp` and the new one to `nlloc.hyp`.
I decided to keep skipping doctests, since they are exactly the same example tested in `test_core.py`.

claudiodsf:
Travis is failing: maybe the idea of comparing QuakeML files as strings is not the right approach. @megies Any thought?

jkmacc-LANL:
@claudiodsf Could you hash their strings and compare hashes?

megies:
> I decided to keep skipping doctests, since they are exactly the same example tested in test_core.py

I'd still vote for running them. One of the reasons we run the doctest although the same functionality is covered in unittests properly is that this will make sure that these example usages very prominently appearing the API docs don't break at some point when things are changed. This is one of the worst things that can happen, IMHO, to have an example usage in the docs that doesn't work as advertised.

megies:
> Travis is failing: maybe the idea of comparing QuakeML files as strings is not the right approach. @megies Any thought?

That's a recurring nuisance.. we should just at some point write a proper utility function to compare QuakeML output, I guess.. I fear I probably won't have time for this in the next time though..

obspy/io/nlloc/core.py
```diff
@@ -138,21 +137,20 @@ def read_nlloc_hyp(filename, coordinate_converter=None, picks=None, **kwargs):
...
     seconds = float(line.split()[6])
     time = UTCDateTime(year, month, day, hour, minute, seconds)
 
+    x, y, z = map(float, line.split()[8:13:2])
+
+    if coordinate_converter:
+        # maximum likelihood origin location in km info line
+        line = lines["HYPOCENTER"]
+        x, y, z = coordinate_converter(*map(float, line.split()[1:7:2]))
+
```
megies: Sorry if it seems pedantic, but I still think it is logically easier to read like..
```python
    line = lines["GEOGRAPHIC"]
    year, month, day, hour, minute = map(int, line.split()[1:6])
    seconds = float(line.split()[6])
    time = UTCDateTime(year, month, day, hour, minute, seconds)

    if coordinate_converter:
        # maximum likelihood origin location in km info line
        line = lines["HYPOCENTER"]
        x, y, z = coordinate_converter(*map(float, line.split()[1:7:2])) 
    else:
        line = lines["GEOGRAPHIC"]
        x, y, z = map(float, line.split()[8:13:2])
```

claudiodsf:
@megies So for the moment I'll compare some specific fields for the two `Catalog` objects. 
Ok also for the doctests.

obspy/io/nlloc/core.py
claudiodsf: Ok! :)

claudiodsf:
@megies The problem is that doctests won't work because `resource_id` (from `print(event)`) changes every time. Any suggestions?

megies:
> @megies The problem is that doctests won't work because resource_id (from print(event)) changes every time. Any suggestions?

Sure! :-)

Put `  # doctest: +ELLIPSIS` Tag and replace the changing parts with `...`:

e.g. https://github.com/obspy/obspy/blob/4bf9901b5071fbd2adf995f0d157282bf5084e11/obspy/io/kinemetrics/__init__.py#L25-L26

claudiodsf:
Force-pushed again. Hope it works this time :hope:

megies:
Good to merge, I guess. Changes in docs should be fine, too. You can review them later when the docs build is done if necessary.

claudiodsf:
When is the doc build performed?

megies:
It's running right now.. build was queued after 1141.. ;-)
New builds get started every full hour, but sequentially, so takes a while if multiple PRs are being pushed to.

obspy/io/nlloc/__init__.py
```diff
@@ -15,11 +15,47 @@
...
-(as opposed to WGS84 with depth in meters down), we need to set up a coordinate
-conversion function to convert from the NonLinLoc location coordinates `x`,
-`y`, `z` to longitude, latitude and depth in kilometers.
+>>> from obspy import read_events
+>>> cat = read_events("/path/to/nlloc.hyp")
+>>> print(cat)
+1 Event(s) in Catalog:
+2006-07-15T17:21:20.195670Z |  +7.737,  +51.658
+
+>>> event = cat[0]  # doctest:
```
QuLogic: There are no options in this doctest comment.

obspy/io/nlloc/__init__.py
```diff
@@ -15,11 +15,47 @@
...
+>>> print(event)  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
+Event:	2006-07-15T17:21:20.195670Z |  +7.737,  +51.658
+<BLANKLINE>
+       resource_id: ResourceIdentifier(id="smi:local/...")
+     creation_info: CreationInfo(creation_time=UTCDateTime(2013, 6, 21,
+12, 53, 34), version='NLLoc:v6.02.07')
```
QuLogic: Since you have whitespace normalization on, you might as well align this one properly.

obspy/io/nlloc/__init__.py
```diff
@@ -15,11 +15,47 @@
...
+>>> print(origin)  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
+Origin
+       resource_id: ResourceIdentifier(id="smi:local/...")
+              time: UTCDateTime(2006, 7, 15, 17, 21, 20, 1956...)
+         longitude: 51.657659...
+          latitude: 7.736781...
+             depth: 1433.5... [confidence_level=68, uncertainty=...]
+                ...
+     creation_info: CreationInfo(creation_time=UTCDateTime(2013, 6, 21,
+12, 53, 34), version='NLLoc:v6.02.07')
```
QuLogic: And here, too.

obspy/io/nlloc/__init__.py
```diff
@@ -61,19 +97,19 @@
 
 >>> origin = event.origins[0]  # doctest: +SKIP
 >>> print(origin)  # doctest: +SKIP
```
QuLogic: It looks like you updated the doctest, but it's still being skipped?

obspy/io/nlloc/__init__.py
```diff
@@ -61,19 +97,19 @@
...
-        ---------
-                   comments: 1 Elements
-                   arrivals: 8 Elements
+Origin
+       resource_id: ResourceIdentifier(id="smi:local/...")
+              time: UTCDateTime(2010, 5, 27, 16, 56, 24, 612600)
+         longitude: 11.64553754...
+          latitude: 48.04707051...
+             depth: 4579.4... [confidence_level=68,
+uncertainty=191.6063...]
```
QuLogic: Again, this can be aligned better.

claudiodsf:
I followed @QuLogic advice and included all doctests. However AppVeyor fails since `pyproj` is not installed on Windows. Moreover, I think that `pyproj` is not an ObsPy dependency. @megies ?

megies:
As mentioned before that's exactly why I was skipping those doctests. In unittests we can add a skip condition if pyproj is not there but that's not working for doctests. Just leave those older doctests skipped like before and let yours (that don't need pyproj) run.

QuLogic:
Yes, sorry, I didn't mean to enable the tests that require pyproj.

claudiodsf:
Force pushed again... Let it work!

megies:
Good to merge, I think.

claudiodsf:
Merged! That was a long process for a small fix! I guess this indicates that the code quality standards in ObsPy are pretty high, nowadays.

In any case, I think it would be nice to have some page on the Wiki explaining all this `# doctest:` stuff.

megies:
Euhm.. @claudiodsf, did you ever use this so far? Just stumbled over a bug in the implementation in #1480.. latitude and longitude fields are mixed up :stuck_out_tongue_winking_eye:

PR #1276, opened 2016-02-20
Remove dynamic import rerouting, deprecated methods,  etc

megies:
It's time to get rid of all this dynamic import rerouting, deprecated methods, etcpp.

See #842, #999, #1039, #1246.

(-3530 lines ;-))

PR #1391, opened 2016-05-03
Feature moment rate function in event / SCARDEC catalogue files

sstaehler:
Starting from the observation that there is no representation of moment rate functions in the event class yet, I started a prototype implementation. Since there is no support for moment rate functions in the QuakeML format, the event class and QuakeML diverge at that point (which might be controversial, so if you find a better implementation, I am happy). 

As an example, I added a reader for the SCARDEC catalogue files (http://scardec.projects.sismo.ipgp.fr/), which contain a regularly sampled moment rate function.

This reader is modeled after the CMTSOLUTION readers and is covered by tests.


megies:
> Starting from the observation that there is no representation of moment rate functions in the event class yet, I started a prototype implementation. Since there is no support for moment rate functions in the QuakeML format, the event class and QuakeML diverge at that point (which might be controversial, so if you find a better implementation, I am happy).

Hmm.. I think this won't work smoothely, because as you stated they're modelled after QuakeML 1.x which doesn't cover these STF elements, so the way it is now these elements are in our API but would get lost writing to QuakeML which is really the cornerstone exchange format.

However, I've implemented a way to handle additional custom tags (in a custom namespace), see the [entry in the Tutorial](http://docs.obspy.org/master/tutorial/code_snippets/quakeml_custom_tags.html?highlight=extra). That way these STF thingies would also get written to QuakeML, and how this is done could be defined in an inofficial or semi-official extension of QuakeML. Would that be an option for you?

megies:
I think you also need to make sure to set a valid resource ids, e.g. here: https://github.com/obspy/obspy/pull/1391/files#diff-1fcfef78501f870f4fdbdb7a207e58b1R299
Looks like you're event name contains a space.

sstaehler:
> However, I've implemented a way to handle additional custom tags (in a custom namespace), see the entry in the Tutorial. That way these STF thingies would also get written to QuakeML, and how this is done could be defined in an inofficial or semi-official extension of QuakeML. Would that be an option for you?

That sounds like a viable way to do that. Give me a day or two for implementing it.
Thanks!

megies:
Hmm.. however, nested tags are not implemented right now.. so you'd need to add most additional info into attributes on the tag, I guess..

sstaehler:
Okay, moved the moment rate function into a extra attribute of obspy.core.event.SourceTimeFunction

This is all a bit ugly, but it works. Probably there are some things which should have been handled better. For example: What namespace to use?

obspy/io/scardec/core.py
```diff
@@ -0,0 +1,457 @@
...
+        rake2=int(np2.rake)
+    )
+
+    # Write to a buffer/file opened in binary mode.
+    buf.write(template.encode())
+
+    stf = foc_mec.moment_tensor.source_time_function.extra
+    t = stf['offset']['value']
+    scalmom = foc_mec.moment_tensor.scalar_moment
+    for sample in stf['moment_rate']['value']:
```
megies: This could in principle probably be done much faster using `np.savetxt(buf, ...)`, avoiding the loop. But it would only matter when writing lots of files I guess..

obspy/io/scardec/core.py
```diff
@@ -0,0 +1,457 @@
...
+                      "written to the file.")
+        cmt_origin = candidates[0]
+    # Otherwise just take the preferred or first one.
+    else:
+        warnings.warn("Could not find a centroid origin. Will instead "
+                      "assume that the preferred or first origin is the "
+                      "centroid origin.")
+        cmt_origin = event.preferred_origin() or event.origins[0]
+
+    if not foc_mec.moment_tensor.source_time_function.extra:
```
megies: This won't work I think. If there's no `.extra` this will raise an `AttributeError`. Should be changed to e.g. `if "extra" not in foc_mec.moment_tensor.source_time_function`.

obspy/io/scardec/core.py
```diff
@@ -0,0 +1,457 @@
...
+    Helper function to create consistent resource ids.
+    """
+    res_id = "smi:local/scardec/%s/%s" % (scardecname, res_type)
+    if tag is not None:
+        res_id += "#" + tag
+    return res_id
+
+
+def _buffer_proxy(filename_or_buf, function, reset_fp=True,
+                  file_mode="rb", *args, **kwargs):
```
megies: This function is now defined for the fourth time I think. But it's not your fault really, it was duplicated before when it shouldn't have been..

megies:
Couple of things missing.. see https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule
 - add to `DEFAULT_MODULES` (so that `obspy-runtests` picks it up when running the "default" tests)
 - docs skeleton for new submodule (compare [PR docs build](http://docs.obspy.org/pull-requests/1391/) with [broken links to io.scardec](http://docs.obspy.org/pull-requests/1391/packages/autogen/obspy.core.event.read_events.html#obspy.core.event.read_events))
 - add to io docs overview

Also, all of these changes with deprecation reroutes in `event/base.py` that were inadvertently included in this PR should be removed (e.g. https://github.com/obspy/obspy/pull/1391/files#diff-ff371be72dfc66fd5b78dacfa106aa33R536), they are part of a much older version of `master` and are not part of current `master` (removed in #1276).

megies:
> What namespace to use?

Since these additional tags will be set from within obspy, maybe use something like `"http://obspy.org/xmlns/1.0"`.

obspy/io/scardec/core.py
```diff
@@ -0,0 +1,457 @@
...
+                      force_resource_id=False))
+
+    # Write moment rate function
+    extra = {'moment_rate': {'value': stf_mr,
+                             'namespace': r"http://test.org/xmlns/0.1"},
+             'dt': {'value': dt,
+                    'namespace': r"http://test.org/xmlns/0.1"},
+             'offset': {'value': offset,
+                        'namespace': r"http://test.org/xmlns/0.1"}
+             }
```
megies: I guess one `SourceTimeFunction` will always only have one single moment rate array? Otherwise `dt` and `offset` could be attributes of `moment_rate`, but I guess having multiple `moment_rate` items would not make sense anyway.

obspy/io/scardec/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+          focal_mechanisms: 1 Elements
+                   origins: 1 Elements
+                magnitudes: 1 Elements
+     source_time_functions: 1 Elements
+
+
+This module also offers write support for the SCARDEC format.
+
+>>> cat.write("output/SCARDEC_file")  # doctest: +SKIP
+
```
megies: Maybe a minimal example of how to set up the data for a SCARDEC file from scratch could also be helpful? With the current documentation people would have to look up the details about setting `SourceTimeFunction.extra = {"moment_rate": ..., ..., ...}` in the source code..

sstaehler: Hm... Could I also create a function "add_moment_rate" or similar to event.SourceTimeFunction?

obspy/io/scardec/core.py
sstaehler: Yes, I agree.

obspy/io/scardec/__init__.py
megies: Hmm.. didn't think of that option.. why not I guess. Please include a note in the docstring that it's not official QuakeML and will be saved as custom namespace tags (or to SCARDEC format).

megies:
Old numpy seems to run into rounding errors due to the multiple additions (for time of samples), see https://travis-ci.org/obspy/obspy/jobs/128940710#L722. So might be safer after all to create the array with `np.arange()` or the like and do the writing in a single line with `np.savetxt()`..

Also this docs page renders not as expected, I think: http://docs.obspy.org/pull-requests/1391/packages/obspy.io.scardec.html#module-obspy.io.scardec

obspy/io/scardec/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+                  comments: 1 Elements
+          focal_mechanisms: 1 Elements
+                   origins: 1 Elements
+                magnitudes: 1 Elements
+     source_time_functions: 1 Elements
+
+
+This module also offers write support for the SCARDEC format.
+
+>>> cat.write("output/SCARDEC_file")  # doctest: +SKIP
```
megies: this line needs to specify `format="..."` (i assume the example should save to SCARDEC again? or quakeml?)

megies:
> Hmm.. however, nested tags are not implemented right now.. so you'd need to add most additional info into attributes on the tag, I guess..

Nested custom tags are implemented in master now, not sure if that would change things here..

This PR needs a rebase on current master.

For inclusion in 1.1.0 this needs to be ready within the next week or two..

sstaehler:
@megies It seems to me that the crashes in Travis are caused by the function tf_misfit and not by my module. Shall I rebase once tf_misfit is fixed?

megies:
Most of the fails are unrelated, see #1599.

But there is one "real" fail in our minimum dependency build: https://travis-ci.org/obspy/obspy/jobs/178683706#L1042

Appveyor also has some real fails: https://ci.appveyor.com/project/obspy/obspy/build/1.0.3747-master#L685

Both problems can also be seen on our docker testbot:
http://tests.obspy.org/?version=1.0.2.post0%2B790.ge36bff7d6c&node=docker-

sstaehler:
@megies 
Okay, I think the remaining failing checks all come from other modules.

megies:
@sstaehler would you mind rebasing on current master one last time? That should get rid of any unrelated fails now..

obspy/io/scardec/core.py
```diff
@@ -0,0 +1,462 @@
...
+        cmt_origin = candidates[0]
+    # Otherwise just take the preferred or first one.
+    else:
+        warnings.warn("Could not find a centroid origin. Will instead "
+                      "assume that the preferred or first origin is the "
+                      "centroid origin.")
+        cmt_origin = event.preferred_origin() or event.origins[0]
+
+    if "extra" not in foc_mec.moment_tensor.source_time_function:
+        raise ValueError('Event focal mechanism must contain a source time \
```
megies: This should read `'Event moment tensor must ...'`

megies:
@sstaehler can you rebase on current master once more and fix merge conflicts?

sstaehler:
Rebase is always dangerous when I do it. Let me know when I caused a nuclear war by accident.

Thanks for your work!

megies:
@sstaehler looks like an inadvertent change: https://github.com/obspy/obspy/pull/1391/files#diff-ff371be72dfc66fd5b78dacfa106aa33L670 ?

megies:
Cleaned up this PR, rebased and (tried to) force-push. Previous PR tip prior to rebase was f03e514123abac95b23c1820bdd0bdcb93088ab8.

@sstaehler can you please click "allow edits by maintainers" on the right (so I can force-push to this PR branch)?

screenshot from 2017-05-09 12 34 13

megies:
OK, force-pushed the cleanup now..

megies:
Good to go, I guess..

sstaehler:
Awesome, thanks Tobi!

Am 11/05/17 um 15:35 schrieb Tobias Megies:
> Good to go, I guess..

Issue #1208, opened 2015-12-13
Docs: Correct command-line script names in sphinx docs + list of command-line scripts on "Library Reference" page


megies:
- automatic `script` page doesn't work (http://docs.obspy.org/master/packages/autogen/obspy.scripts.sds_html_report.html#module-obspy.scripts.sds_html_report) because name of registered script is not the same as `obspy-` plus module name
- ~~should add static example image of a html report page for docs pages~~ done

See #1202.

megies:
Bumping.. not important enough to hold up the release.

megies:
Maybe we could also use this: https://pythonhosted.org/sphinxcontrib-programoutput

ThomasLecocq:
solved with https://github.com/obspy/obspy/pull/3586 : used an easier way to handle it, directly within the template. Not ideal in theory, but there are few chances that we'll add new scripts, at least not often. Maintenance is easy.

PR #1397, opened 2016-05-06
EIDA routing and auth support

andres-h:
Example usage:

``` bash
$ obspy-eida-fetch -s "2016-05-01T10:00:00Z" -e "2016-05-01T10:01:00Z" -C "BH?" -v -o data.mseed
$ obspy-eida-fetch -s "2016-05-01T10:00:00Z" -e "2016-05-01T10:01:00Z" -C "BH?" -y station -q level=response -v -o data.xml
```

``` python
import obspy.clients.eida
from obspy import UTCDateTime
client = obspy.clients.eida.Client(debug=True)
t1=UTCDateTime("2016-05-01T10:00:00")
t2=UTCDateTime("2016-05-01T10:01:00")
client.get_waveforms("*", "*", "*", "BH?", t1, t2)
client.get_stations(t1, t2, network="*", station="*", channel="BH?", level="response")
```

No tests yet, but probably obspy.clients.fdsn tests can be adopted.


obspy/clients/eida/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+        :param retry_wait: Seconds to wait before each retry.
+        :type maxthreads: int
+        :param maxthreads: Maximum number of download threads.
+        :type credentials: dict
+        :param credentials: url -> (username, password).
+        :type authdata: string
+        :param authdata: Authentication token (PGP base64 format).
+
+        """
+        obspy.clients.fdsn.Client.__init__(self, base_url, **kwargs)
```
megies: Might be a bit shorter/cleaner avoiding the long path to `obspy.clients.fdsn.Client` by using `super(Client, self).__init__(base_url, **kwargs)`.

obspy/clients/eida/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Client(obspy.clients.fdsn.Client):
+    def __init__(self, base_url='GFZ', retry_count=10, retry_wait=60, maxthreads=5, credentials=None, authdata=None, **kwargs):
+        """
+        Initializes an FDSN/EIDA Web Service client.
+
+        :type base_url: str
+        :param base_url: Base URL of FDSN/EIDA web service compatible server
+            (e.g. "http://geofon.gfz-potsdam.de") or key string for recognized
+            server.
```
megies: This should have a link to the respective page in the docs, i.e. ``.. see :mod:`FDSN client documentation <obspy.clients.fdsn>` ``

obspy/clients/eida/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+        self.__authdata = authdata
+
+    def _create_url_from_parameters(self, service, *args):
+        if service in ('dataselect', 'station'):
+            # construct a URL for the routing service
+            u = urlparse.urlparse(obspy.clients.fdsn.Client._create_url_from_parameters(self, service, *args))
+            return urlparse.urlunparse((u.scheme, u.netloc, u'/eidaws/routing/1/query', u'', u.query + u'&service=' + service, u''))
+
+        else: # 'event' is not routed
+            return obspy.clients.fdsn.Client._create_url_from_parameters(self, service, *args)
```
megies: Since inheritance is declared with the class definition this would be cleaner with `super(Client, self)._create_url_from_parameters(service, *args)`, I think.

obspy/clients/eida/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+                raise obspy.clients.fdsn.header.FDSNException("No data available for request.")
+
+            if return_string:
+                return dest.getvalue()
+
+            else:
+                return dest
+
+        else:
+            return obspy.clients.fdsn.Client._download(self, url, return_string, data, use_gzip)
```
megies: Again, cleaner using `super(..)._download(..)`, I think.

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,735 @@
...
+                    # (new)lines following the request lines are ignored #
+                    else:
+                        if reqflag:
+                            break
+
+        self.__parse_tokens(self.head)
+
+msglock = threading.Lock()
+
+def msg(verb, s):
```
megies: Could be changed to `def msg(s, verbose=True)`..

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,735 @@
...
+import datetime
+import optparse
+import threading
+import socket
+import csv
+import re
+
+try:
+    # Python 3.2 and earlier
+    from xml.etree import cElementTree as ET
```
megies: We don't support Python 3.2.. does it even run on 3.2? Is there a strong need for this one? We wouldn't notice if it breaks in the future..

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,735 @@
...
+    import urllib2
+    import urlparse
+    import urllib
+
+except ImportError:
+    # Python 3.x
+    import queue as Queue
+    import urllib.request as urllib2
+    import urllib.parse as urlparse
+    import urllib.parse as urllib
```
megies: In principle we use `python-future` for most of these, although we're not really happy with that approach either. So I guess it can stay like this..

In any case, using `from future.utils import PY2; if PY2: ... else: ...) would be cleaner than the `try/except` for now..

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,735 @@
...
+
+VERSION = "1.99 (2016.127)"
```
megies: Just relying on `obspy.__version__` would bring some advantages, most prominently no need to remember manually bouncing this number (which will eventually for sure get overlooked).

megies:
We will need a test setup that really uses all the features, i.e. routing, authentication via credentials / token. @andres-h, can you set up some account for testing (including routing)?

obspy/clients/eida/fetch.py
andres-h: On 05/09/2016 12:02 PM, Tobias Megies wrote:
> In obspy/clients/eida/fetch.py
> <https://github.com/obspy/obspy/pull/1397#discussion_r62475504>:
>
>> +except ImportError:
>> +    from xml.etree import ElementTree as ET
>> +
>> +try:
>> +    # Python 2.x
>> +    import Queue
>> +    import urllib2
>> +    import urlparse
>> +    import urllib
>> +
>> +except ImportError:
>> +    # Python 3.x
>> +    import queue as Queue
>> +    import urllib.request as urllib2
>> +    import urllib.parse as urlparse
>> +    import urllib.parse as urllib
>
> In principle we use |python-future| for most of these, although we're
> not really happy with that approach either. So I guess it can stay like
> this..
>
> In any case, using |from future.utils import PY2; if PY2: ... else: ...)
> would be cleaner than the|try/except` for now..

I wanted fetch.py to also work standalone without extra dependencies 
like python-future.

obspy/clients/eida/fetch.py
andres-h: On 05/09/2016 11:57 AM, Tobias Megies wrote:
> In obspy/clients/eida/fetch.py
> <https://github.com/obspy/obspy/pull/1397#discussion_r62474984>:
>
>> +
>> +"""
>> +A command-line FDSN Web Service client with EIDA routing and authentication support.
>> +"""
>> +import sys
>> +import time
>> +import datetime
>> +import optparse
>> +import threading
>> +import socket
>> +import csv
>> +import re
>> +
>> +try:
>> +    # Python 3.2 and earlier
>> +    from xml.etree import cElementTree as ET
>
> We don't support Python 3.2.. does it even run on 3.2? Is there a strong
> need for this one? We wouldn't notice if it breaks in the future..

fetch.py is a standalone script that is also distributed independently 
of ObsPy and we want it to work with vanilla Python 2.6+ and 3.x. Due to 
large size of StationXML, it is important to use the efficient 
cElementTree implementation. Python 3.3 docs say: Changed in version 
3.3: This module will use a fast implementation whenever available. The 
xml.etree.cElementTree module is deprecated.

andres-h:
On 05/09/2016 12:11 PM, Tobias Megies wrote:
> We will need a test setup that really uses all the features, i.e.
> routing, authentication via credentials / token. @andres-h
> <https://github.com/andres-h>, can you set up some account for testing
> (including routing)?

Yes, will do it after EIDA meeting this week.

obspy/clients/eida/fetch.py
megies: Ok, fine.

obspy/clients/eida/fetch.py
megies: Alright. Won't hurt I guess.

megies:
Some more remarks..

A) In principle this could also go into `/clients/fdsn/..` but I guess you would prefer this to be in a separate clients module folder? I don't have a very strong opinion on this, @krischer and @QuLogic what do you think?

B) If it stays a separate module, there's a couple more things to do, see https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule:
 * register for tests in `NETWORK_MODULES` (for our `obspy-runtests` script)
 * add to docs index page and add skeleton for module (please note that you can check out the PR's docs build [here](http://docs.obspy.org/pull-requests/1397/), the link is listed in the commits status checks)

C) In general we don't have code in submodules' `__init__.py`, so to keep a consistent code base it would be good to move the code from there to a new `eida/client.py` file and only import the `Client` in `eida/__init__.py`.

D) Some tests are failing, see my report at http://tests.obspy.org/40607/. Stdout contained a message: `getting data from http://www.orfeus-eu.org/fdsnws/dataselect/1/query failed: HTTP Error 500: Internal Server Error`. CI currently only runs non-network module tests (see #1408), so all of these green check-marks have no real significance, unfortunately.


Also, please feel free to add yourself to the list of [contributors](https://github.com/obspy/obspy/blob/master/obspy/CONTRIBUTORS.txt)..

andres-h:
On 05/17/2016 01:54 PM, Tobias Megies wrote:

> A) In principle this could also go into |/clients/fdsn/..| but I guess
> you would prefer this to be in a separate clients module folder? I don't
> have a very strong opinion on this, @krischer
> <https://github.com/krischer> and @QuLogic <https://github.com/QuLogic>
> what do you think?

The module requires the routing service, which is /eidaws/, so I think 
using "eida" would be appropriate.

On the other hand, we are planning set up an instance of the routing 
service that includes non-EIDA data, so it would be possible to use the 
module to fetch data from the whole world. In this case "eida" could be 
misleading.


> B) If it stays a separate module, there's a couple more things to do,
> see https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule:
>
>   * register for tests in |NETWORK_MODULES| (for our |obspy-runtests|
>     script)
>   * add to docs index page and add skeleton for module (please note that
>     you can check out the PR's docs build here
>     <http://docs.obspy.org/pull-requests/1397/>, the link is listed in
>     the commits status checks)

Yes, I also have to write a tutorial of obspy-eida-fetch for our users. 
Where would be the right place for that in the package?


> C) In general we don't have code in submodules' |__init__.py|, so to
> keep a consistent code base it would be good to move the code from there
> to a new |eida/client.py| file and only import the |Client| in
> |eida/__init__.py|.

OK.


> D) Some tests are failing, see my report at
> http://tests.obspy.org/40607/. Stdout contained a message: |getting data
> from http://www.orfeus-eu.org/fdsnws/dataselect/1/query failed: HTTP
> Error 500: Internal Server Error|. CI currently only runs non-network
> module tests (see #1408 <https://github.com/obspy/obspy/issues/1408>),
> so all of these green check-marks have no real significance, unfortunately.

"Internal Server Error" is a failure of FDSNWS at ODC. It worked 
yesterday (although rather slowly).

The other fail is a bug in our own (SC3) FDSNWS implementation. I have 
to look into that.

The tests are testing the EIDA FDSNWS services rather than ObsPy :)

megies:
> The module requires the routing service, which is /eidaws/, so I think using "eida" would be appropriate.

Oh.. right, saw it but didn't think of it right now..

> On the other hand, we are planning set up an instance of the routing service that includes non-EIDA data, so it would be possible to use the module to fetch data from the whole world. In this case "eida" could be misleading.

Hmm.. I guess since this routing service comes out of EIDA it might still be the right name for the module..?  For these other non-EIDA data centers, would they also have a `/eidaws/` end point (probably not, routing done from one central node, I assume..)?

> Yes, I also have to write a tutorial of obspy-eida-fetch for our users. Where would be the right place for that in the package?

This should be on the module's API docs landing page, i.e. in the the docstring at the top of `clients/eida/__init__.py`. (Compare e.g. the [FDSN one](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/__init__.py) and [how it gets rendered in the docs](http://docs.obspy.org/master/packages/obspy.clients.fdsn.html)). Docs builds for PRs get started every full hour, so they can be checked 2-3 hours after pushing at least.


> The tests are testing the EIDA FDSNWS services rather than ObsPy :)

We're used to annoying data centers, prompting them to fix their WADLs or QuakeML files etc.  for syntax and schema conformity and the like.. :wink:

andres-h:
On 05/17/2016 05:13 PM, Tobias Megies wrote:

> Hmm.. I guess since this routing service comes out of EIDA it might
> still be the right name for the module..? For these other non-EIDA data
> centers, would they also have a |/eidaws/| end point (probably not,
> routing done from one central node, I assume..)?

Anybody can run a routing service, but one central node is enough. 
Currently several EIDA nodes are running an instance of the routing service.

IRIS has their own routing service (fedcatalog), but it is currently not 
reliable (for example, 
http://service.iris.edu/irisws/fedcatalog/1/query?net=GE routes some of 
our stations to other data centers that do not have our data).

Maybe EIDA and IRIS developments can be joined in the future.


>     Yes, I also have to write a tutorial of obspy-eida-fetch for our
>     users. Where would be the right place for that in the package?
>
> This should be on the module's API docs landing page, i.e. in the the
> docstring at the top of |clients/eida/__init__.py|. (Compare e.g. the
> FDSN one
> <https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/__init__.py>
> and how it gets rendered in the docs
> <http://docs.obspy.org/master/packages/obspy.clients.fdsn.html>).

That looks impressive, but obspy.clients.eida implements the same 
interface, so all the examples would be the same except "eida" instead 
of "fdsn"...

I actually meant the obspy-eida-fetch command-line utility.

megies:
> That looks impressive, but obspy.clients.eida implements the same interface, so all the examples would be the same except "eida" instead of "fdsn"...

Tha API page should then refer to the FDSN one (``... :mod:`obspy.clients.fdsn` ``) and elaborate a bit on the differences and link to EIDA pages describing the system.

> I actually meant the obspy-eida-fetch command-line utility.

API page for scripts is generated from the docstring of the respective file, in your case `fetch.py`. Compare..
 * https://github.com/obspy/obspy/blob/master/obspy/scripts/sds_html_report.py
 * http://docs.obspy.org/master/packages/autogen/obspy.scripts.sds_html_report.html

Unfortunately you'll also be hit by #1208.. (i.e. the auto generated "Note:" box will advertise it as `obspy-fetch`, but it should be sufficiently clear if you put some example command line usage below)

obspy/clients/eida/__init__.py
```diff
@@ -44,7 +44,7 @@
 Add debug=True to see what is going on.
 
 >>> from obspy.clients.eida import Client
->>> credentials = {"http://service.iris.edu/fdsnws/dataselect/1/queryauth":
+>>> credentials = {"http://service.iris.edu/fdsnws/dataselect/1/queryauth": \
                    ("nobody@iris.edu", "anonymous")}
```
megies: `...` instead of `>>>` on the continuation line would be a bit nicer, then the ` \` is not needed. But only cosmetics, of course..

obspy/clients/eida/__init__.py
```diff
@@ -10,120 +13,182 @@
...
+Basic Usage
+-----------
+
+The first step is always to initialize a client object.
+
+>>> from obspy.clients.eida import Client
+>>> client = Client("GFZ")
+
+A client object can be initialized either with the base URL of any EIDA routing
+service or with a shortcut name that will be mapped to such a URL. The examples
```
megies: This is using the short URLs defined in `fdsn/header.py`, right? Would be good to have a link in the docstring, like `... :mod:`a shortcut name that will be mapped to such a URL <obspy.clients.fdsn>` ...`

obspy/clients/eida/__init__.py
```diff
@@ -10,120 +13,182 @@
...
+
+    .. plot::
+
+        from obspy import UTCDateTime
+        from obspy.clients.eida import Client
+        client = Client("INGV")
+        starttime = UTCDateTime("2002-01-01")
+        endtime = UTCDateTime("2002-01-02")
+        cat = client.get_events(starttime=starttime, endtime=endtime)
+        cat.plot()
```
megies: I think `cat.plot(projection="local")` looks nicer for this example.. (if you change it of course in the docstring above too)

megies:
Looks pretty much ready to me. Any other EIDA heavy user wants to do a review, maybe @petrrr or @claudiodsf?

The `fetch` script currently is listed as a `module` in docs, I think it would be better to have it show up as a `script`, compare..
 * https://github.com/obspy/obspy/blob/master/misc/docs/source/packages/obspy.db.rst#L20-L27
 * http://docs.obspy.org/master/packages/obspy.db.html#scripts

claudiodsf:
I'm not really an heavy EIDA user. I'll skip on that 😉

obspy/clients/eida/__init__.py
andres-h: On 05/19/2016 01:14 PM, Tobias Megies wrote:
> In obspy/clients/eida/__init__.py
> <https://github.com/obspy/obspy/pull/1397#discussion_r63860381>:
>
>> @@ -10,120 +13,182 @@
>>  :license:
>>      GNU Lesser General Public License, Version 3
>>      (https://www.gnu.org/copyleft/lesser.html)
>> +
>> +Basic Usage
>> +-----------
>> +
>> +The first step is always to initialize a client object.
>> +
>> +>>> from obspy.clients.eida import Client
>> +>>> client = Client("GFZ")
>> +
>> +A client object can be initialized either with the base URL of any EIDA routing
>> +service or with a shortcut name that will be mapped to such a URL. The examples
>
> This is using the short URLs defined in |fdsn/header.py|, right? Would
> be good to have a link in the docstring, like |... :mod:|a shortcut name
> that will be mapped to such a URL |...|

Not sure how you want this exactly to be typeset. Like:

:mod:`a shortcut name that will be mapped to such a URL 
<obspy.clients.fdsn>`

???

Note that not all shortcut names work, only those that run the routing 
service. Except when using the even service that is not routed.

Also, some doctest fails because request to 5 nodes (default) is made in 
parallel and the order of networks in print(inventory) depends on the 
order of responses received. We cannot rely on station descriptions not 
changing either.

The fail of test_iris_example_queries_station should disappear after all 
EIDA nodes patch their FDSNWS.

megies: > Not sure how you want this exactly to be typeset. Like: :mod:`a shortcut name that will be mapped to such a URL <obspy.clients.fdsn>`

Something like that, yes. That would make it a link to FDSN module API page (with the short URLS displayed in the example) with link text "a shortcut name that will be mapped to such a URL".

> Note that not all shortcut names work, only those that run the routing service. Except when using the even service that is not routed.

Maybe add a note in brackets..   ".. (only works if node supports EIDA routing)" ?
In any case, I think it should somehow be clear what short URLs can be used and expected to work. Also, will it show a meaningful message if a not-supporting node is targetted by the client initialization?

> Also, some doctest fails because request to 5 nodes (default) is made in parallel and the order of networks in print(inventory) depends on the order of responses received. We cannot rely on station descriptions not changing either.

You could sort networks by code on the resulting inventory before returning it with the following call, might be good anyway to be as deterministic as possible with the result:

```python
inv.networks = sorted(inv.networks, key=lambda x: x.code)
```

andres-h: On 05/19/2016 03:12 PM, Tobias Megies wrote:

> Maybe add a note in brackets.. ".. (only works if node supports EIDA
> routing)" ?
> In any case, I think it should somehow be clear what short URLs can be
> used and expected to work. Also, will it show a meaningful message if a
> not-supporting node is targetted by the client initialization?

I hope "getting routes from ... failed: HTTP Error 404: Not Found" 
(after 10 retries by default) is meaningful enough...

HTTP error is the only thing that is sure. I cannot tell if routing is 
really not supported on purpose of if the node is temporarily 
mis-configured.

Could check application.wadl, but that is not really reliable either 
(eg., Apache gives 503 for 60 seconds each time when the service is 
restarted).

obspy.clients.eida could have its own URL_MAPPINGS of nodes that run the 
routing service. Then we would have to do a pull request each time a 
node is added or removed.

In practice there would be rarely a need to use another node than "GFZ" 
(default), because all instances of the EIDA routing service are 
synchronized and use the same set of nodes.

megies: > HTTP error is the only thing that is sure. I cannot tell if routing is really not supported on purpose of if the node is temporarily mis-configured. Could check application.wadl, but that is not really reliable either (eg., Apache gives 503 for 60 seconds each time when the service is restarted). obspy.clients.eida could have its own URL_MAPPINGS of nodes that run the routing service. Then we would have to do a pull request each time a node is added or removed. In practice there would be rarely a need to use another node than "GFZ" (default), because all instances of the EIDA routing service are synchronized and use the same set of nodes.

OK, should be fine as is I think. Probably it would just add confusion to have a separate dictionary for URL mappings.

megies:
Looks good to me. Anybody else want to have a look and comment?

andres-h:
1) Doctest on token.asc disabled. I think using obspy.core.util.base.get_example_file() would make the example confusing. There is a test with actual token already included in tests.

2) This is caused by failure of getting data from ODC even after 600 seconds / 10 retries as reproduced below. I'll let them know, so they can hopefully fix their service.
$ obspy-eida-fetch -N NO -S TROLL -L -- -C BHE -s 2016-03-25T00:00:00 -e 2016-03-25T00:00:06 -v -o data.mseed
```text
getting routes from http://geofon.gfz-potsdam.de/eidaws/routing/1/query?network=NO&location=--&channel=BHE&station=TROLL&starttime=2016-03-25T00%3A00%3A00&endtime=2016-03-25T00%3A00%3A06&format=post
getting data from http://www.orfeus-eu.org/fdsnws/dataselect/1/query
retrying http://www.orfeus-eu.org/fdsnws/dataselect/1/query (1) after 60 seconds due to HTTP Error 500: Internal Server Error
```

3) This is caused by a bug in FDSNWS implementation used by several EIDA nodes. I'll urge them to upgrade.

obspy/clients/eida/client.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class Client(obspy.clients.fdsn.Client):
+    """
+    FDSN Web Service client using EIDA routing and authentication.
+
+    For details see the :meth:`~obspy.clients.eida.client.Client.__init__()`
+    method.
+    """
+    def __init__(self, base_url='GFZ', retry_count=10, retry_wait=60,
```
krischer: I think the `retry_count` should default to `1` and the `retry_wait` to maybe at most 5 seconds. Right now it (worst case) hangs for 10 minutes. IMHO it is reasonable to assume that the web services are stable and in production. What is the reasoning behind this choice?

obspy/clients/eida/client.py
```diff
@@ -0,0 +1,140 @@
...
+        :type base_url: str
+        :param base_url: Base URL of FDSN/EIDA web service compatible server
+            (e.g. "http://geofon.gfz-potsdam.de") or key string for recognized
+            server. See :mod:`FDSN client documentation <obspy.clients.fdsn>`
+        :type debug: bool
+        :param debug: Debug flag.
+        :type timeout: float
+        :param timeout: Maximum time (in seconds) to wait for a single request
+            to receive the first byte of the response (after which an exception
+            is raised).
```
krischer: Can you remove the `debug` and `timeout` parameters from the docstring and replace it with a reference to the base class? Then its clear that they are defined on the base class. The same could be done with `base_url` but it is such an essential parameter to the class that it is probably better to leave it where it is right now.

obspy/clients/eida/client.py
```diff
@@ -0,0 +1,140 @@
...
+
+        if '/dataselect/' in u.path:
+            q['service'] = 'dataselect'
+
+        elif '/station/' in u.path:
+            q['service'] = 'station'
+
+        else:  # 'event' is not routed
+            return super(Client, self)._download(
+                    url, return_string, data, use_gzip)
```
krischer: This might potentially mislead users - I would natively expect that events are routed as well. IMHO it might be better to just raise an error here and tell people you use the normal fdsn client. Opinions?

krischer:
Some tests fails for me on Python 3.5: http://tests.obspy.org/40960/

obspy/clients/eida/client.py
```diff
@@ -0,0 +1,140 @@
...
+        :param retry_wait: Seconds to wait before each retry.
+        :type maxthreads: int
+        :param maxthreads: Maximum number of download threads.
+        :type credentials: dict
+        :param credentials: url => (username, password).
+        :type authdata: string
+        :param authdata: Authentication token (PGP base64 format).
+
+        """
+        super(Client, self).__init__(base_url, **kwargs)
```
krischer: There should be a check during the `__init__()` method that the service at `base_url` actually has a EIDA routing service - similarly to what the default FDSN client does. Download the `/eidaws/routing/1/application.wadl` and maybe parsing it a bit to assert a minimum functionality should do the trick.

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,950 @@
...
+
+POST_PARAMS = set(('service',
+                   'alternative'))
+
+
+class Error(Exception):
+    pass
+
+
+class AuthNotSupported(Exception):
```
krischer: Please make this a child class of the above `Error` class. Then all exceptions from this file have a common base.

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,950 @@
...
+    pass
+
+
+class TargetURL(object):
+    def __init__(self, url, qp):
+        self.__scheme = url.scheme
+        self.__netloc = url.netloc
+        self.__path = url.path.rstrip('query').rstrip('/')
```
krischer: I've been bitten by this sooo many times, but `(r)strip()` does not behave like one intuitively would expect and it might result in very hard to debug errors down the road. Please replace with:

```python
self.__path = re.sub("/?query$", "", url.path)
```

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,950 @@
...
+
+    def post_params(self):
+        return self.__qp.items()
+
+
+class RoutingURL(object):
+    def __init__(self, url, qp):
+        self.__scheme = url.scheme
+        self.__netloc = url.netloc
+        self.__path = url.path.rstrip('query').rstrip('/')
```
krischer: Same as above.

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,950 @@
...
+                    d["beg_4year"] = "20%s" % d["beg_2year"]
+
+            if d["end_2year"]:
+                if int(d["end_2year"]) > 50:
+                    d["end_4year"] = "19%s" % d["end_2year"]
+
+                else:
+                    d["end_4year"] = "20%s" % d["end_2year"]
+
+            # some users have problems with time...
```
krischer: IMHO better to just raise an error and force users to fix their files.

obspy/clients/eida/fetch.py
```diff
@@ -0,0 +1,950 @@
...
+                return fd
+
+            msg("retrying %s (%d) after %d seconds due to HTTP status code %d"
+                % (url, n, wait, fd.getcode()), verbose)
+
+            fd.close()
+            time.sleep(wait)
+
+        except urllib2.HTTPError as e:
+            if e.code >= 400 and e.code < 500:
```
krischer: This is not granular enough. If I read the code correctly then the `route()` functions is also used for all normal FDSN requests - a lot of `4xx` codes are valid return codes from FDSN servers and thus should not be retried.

krischer:
Hi all!

Thanks a lot for this :-) And sorry for joining the discussion so late - I was on leave the last couple of weeks. We definitely want this in ObsPy but the current implementation does not smoothly integrate with the rest of ObsPy and I wanted to discuss a few things. My main reservation is that we have to commit to maintaining everything we merge for many years and I want to make that as easy as possible for us. Thus a couple of points for discussion:

* The additional PGP token authentication method should just be added to the default FDSN client. It would then have the parameters `user`, `password`, and `auth_data`.
* I did not yet fully read the code so I might have missed it: Where does it use the `priority` parameter returned by the routing service? Also how is this even supposed to be interpreted? This [query](http://geofon.gfz-potsdam.de/eidaws/routing/1/query?net=4C&sta=KES27) only has `priority=2`.
* The `fetch.py` file duplicates a bunch of functionality from the default FDSN client. I realize you would like to keep the script independent of ObsPy but if we merge this PR we will have an obligation to maintain this for the years to come. Duplicated functionality makes that much harder for us and it is not feasible for us to maintain things that are independent of ObsPy within our repository in the long run.
* I'm a bit torn on having this be a separate client as it could be very easily integrated into the default `obspy.clients.fdsn` client. This would also avoid the cross-dependency of `obspy.clients.eida` on `obspy.client.fdsn` and at least make it explicit.

My suggestion would be to make the interface part of the default FDSN class:

```python

c = Client("GFZ")

st = c.get_waveforms(..., routing="EIDA")
```

This also gives us an option to eventually add the IRIS federator and a potential FDSN routing service with a very similar interface and not a lot of required code.

Implementation wise it would also much easier if it just used the FDSN client for downloading the things. "pseudo" code:

```python
def get_waveforms(..., routing=None):
    ...
    if routing == "EIDA":
        routes = self._get_eida_routing(*args, **kwargs)
        # Can be threaded as well.
        data = Stream()
        for r in routes:
            _c = Client(r[0])
            data += _c.get_waveforms_bulk(r[1]. *args, **kwargs)
```

A bit more work would naturally be needed to merge XML and CSV files but that should also be doable.

The big advantage of this is that is would use all the functionality we already have in `obspy.clients.fdsn` without having to reimplement a bunch of it.

I realize this is a lot to ask so I would be willing to help achieving it. Let me know of any opinions!

andres-h:
Hi Lion,

> Thanks a lot for this :-) And sorry for joining the discussion so late - I was on leave the last couple of weeks. We definitely want this in ObsPy
> but the current implementation does not smoothly integrate with the rest of ObsPy and I wanted to discuss a few things. My main reservation is that
> we have to commit to maintaining everything we merge for many years and I want to make that as easy as possible for us. Thus a couple of points for
> discussion:
>
>  *  The additional PGP token authentication method should just be added to the default FDSN client. It would then have the parameters user,
>     password, and auth_data.

I think a perfect implementation would use abstract base classes for 
Routing and Authentication services (as well as QC/Catalog services like 
WFCatalog). There could be even abstract base classes for waveform, 
station and event services (not all-in-one). We can route to other 
services besides FDSN dataselect and station. The routing is NSLC-based, 
though (latitude/longitude might be added later).


>  *  I did not yet fully read the code so I might have missed it: Where does it use the priority parameter returned by the routing service? Also how
>     is this even supposed to be interpreted? This query only has priority=2.

The routing service returns only the highest priority by default. To make 
use of multiple priorities we would need other criteria, eg., availability 
of data on multiple routes. Then it would make sense to use lower priority 
route if it has higher availability (or other criterion) than a higher 
priority route. It might be theoretically possible to use the WFCatalog 
service for this in the future.


>  *  The fetch.py file duplicates a bunch of functionality from the default FDSN client. I realize you would like to keep the script independent of
>     ObsPy but if we merge this PR we will have an obligation to maintain this for the years to come. Duplicated functionality makes that much harder
>     for us and it is not feasible for us to maintain things that are independent of ObsPy within our repository in the long run.

The current implementation is very pragmatic. We have to maintain the 
fetch.py script for our users anyway and the quickest solution was to use 
the same script in ObsPy. I don't know how long a proper ObsPy 
implementation would take; we need users starting to use EIDA webservices 
NOW. The EIDA routing service was launched already at the end of 2014.

On the other hand, there have been discussions with IRIS to merge the EIDA 
routing service with fedcatalog, so possibly there will be some changes. 
One strategy would be doing a simpler implementation for the time being 
and do a proper implementation after the routing service has became an 
FDSN standard.

If the obspy.clients.eida cannot be included in ObsPy, we could continue 
providing it as an add-on that is maintained by us.


>  *  I'm a bit torn on having this be a separate client as it could be very easily integrated into the default obspy.clients.fdsn client. This would
>     also avoid the cross-dependency of obspy.clients.eida on obspy.client.fdsn and at least make it explicit.
> 
> My suggestion would be to make the interface part of the default FDSN class:
> 
> from obspy.clients.fdsn import Client
> 
> c = Client("GFZ")
> 
> st = c.get_waveforms(..., routing="EIDA")
> 
> This also gives us an option to eventually add the IRIS federator and a potential FDSN routing service with a very similar interface and not a lot
> of required code.

If you want to do it properly, using abstract interfaces might be better 
OO design:

eidaRouter = EidaRouter("GFZ")
eidaAuth = EidaAuth(token)
fdsnAuth = FdsnAuth(cred)
st = c.get_waveforms(..., router=eidaRouter, authenticators=[fdsnAuth, eidaAuth])

BTW, is the data returned by get_waveforms() limited by RAM size or can it 
write to file directly? fetch.py does not have memory limit for waveform 
data.

I'm on vacation for 2 weeks, will think about it after that.

Regards,
Andres.

krischer:
Hi Andres,

> I think a perfect implementation would use abstract base classes for
> Routing and Authentication services (as well as QC/Catalog services like
> WFCatalog). There could be even abstract base classes for waveform,
> station and event services (not all-in-one). We can route to other
> services besides FDSN dataselect and station. The routing is NSLC-based,
> though (latitude/longitude might be added later).

We already have some limited abstract base classes for the clients, but the FDSN client is not yet based on them: https://github.com/obspy/obspy/blob/master/obspy/clients/base.py

Authentification will definitely be added to these eventually - for the routing I don't see a simple way of achieving this but I'll keep it at the back of my head.

> The routing service returns only the highest priority by default. To make
> use of multiple priorities we would need other criteria, eg., availability
> of data on multiple routes. Then it would make sense to use lower priority
> route if it has higher availability (or other criterion) than a higher
> priority route. It might be theoretically possible to use the WFCatalog
> service for this in the future.

Ah ok - that makes it much simpler to use if one can assume this!


> The current implementation is very pragmatic. We have to maintain the
> fetch.py script for our users anyway and the quickest solution was to use
> the same script in ObsPy. I don't know how long a proper ObsPy
> implementation would take; we need users starting to use EIDA webservices
> NOW. The EIDA routing service was launched already at the end of 2014.

I don't think it would take very long - I'll give it a shot soonish (hopefully before your vacation ends).


> If you want to do it properly, using abstract interfaces might be better
> OO design:
> 
> eidaRouter = EidaRouter("GFZ")
> eidaAuth = EidaAuth(token)
> fdsnAuth = FdsnAuth(cred)
> st = c.get_waveforms(..., router=eidaRouter, authenticators=[fdsnAuth, eidaAuth])

We unfortunately already have `user` and `password` arguments so those have to stay as its a pain for us to deprecate them cleanly. So I guess we just have to add the `auth_data` into the `__init__()` method as you already did in your class.

The extra router class would be necessary if there are some routing services that DO NOT have the same base url as at least one FDSNWS service. Is that an envisioned use case? Making it a mix-in is also probably pretty hard as it might be difficult to find an abstraction that works for data as well as station information but its definitely worth a try!


> BTW, is the data returned by get_waveforms() limited by RAM size or can it
> write to file directly? fetch.py does not have memory limit for waveform
> data.

It can write directly to a file or file-like objects - use the `filename` argument.

Cheers!

Lion

andres-h:
Can we get on with this? There are users who need this functionality however it is implemented...

andres-h:
On 09/19/2016 04:55 PM, Lion Krischer wrote:

> I tried this a bit today and it is not too difficult, the current status
> is in this branch - it can already download stuff so it should be quite
> simple to finish it:
> https://github.com/krischer/obspy/commits/fdsn-generic-routing
>
> It works like this:
>
> from obspy.clients.fdsn import Client
>
> # Optional routing argument - if passed, all requests will be routed.
> c = Client("GFZ", routing="eida")

I'm not sure if this would the best OO design as it seems to violate 
several of the basic OO principles (cf. 
https://en.wikipedia.org/wiki/SOLID), eg.:
* adding new functionality should not require modifying existing classes 
(OCP);
* fdsn.Client should not have multiple responsibilities, eg., affected 
by changes in both routing and fdsnws specifications (SRP).

Based on that, routing should probably be implemented as a separate 
class. But as long as it works, it's OK with me :)


> At initialization time it will parse the EIDA routing WADL and make sure
> the service has everything that is required - same thing we do for all
> other services. I'll try to finish it this week. The script should then
> be based on the FDSN client (or potentially the mass downloader). I
> realize that you would like to keep it independent from ObsPy but as we
> commit to maintaining everything we accept into ObsPy it makes little
> sense for us to have two separate ways of doing the same thing.

Having it properly integrated into ObsPy is even better. We still 
provide the alternative standalone script for users who cannot install 
ObsPy for whatever reason.


> A couple of questions:
>
>   * Can I rely on the JSON format output of the routing service, i.e.
>     will it always be available? Its just easier to parse then XML.

AFAIK, yes. You can see the official specification at 
https://github.com/GEOFON/routing/blob/master/doc/specification.rst


>   * Will the token based authentication eventually be part of the
>     official fdsn ws specs?

I don't know, but I'd be surprised if they adopt it soon. It's also a 
trust issue. Would they trust tokens issued by EIDA?


> It adds a lot of extra logic so it might be
>     nicer to have something like the following:
>
> user, pw=  get_fdsn_auth(authdata)
>
> c=  Client(...,user=user,pw=pwd)

I'm afraid that would not work. Username and password are not derived 
from the token, but issued by each individual node (/auth method).

See also: 
https://github.com/EIDA/seiscomp3/blob/master/src/trunk/apps/fdsnws/descriptions/fdsnws.rst#authentication-extension


> The |application.wadl|
> (http://geofon.gfz-potsdam.de/eidaws/routing/1/application.wadl) also
> has a couple of issues. @andres-h <https://github.com/andres-h> Can you
> pass these on?

I forwarded that to Javier.


Thanks,
Andres.

krischer:
> I'm not sure if this would the best OO design as it seems to violate several of the basic OO principles (cf. https://en.wikipedia.org/wiki/SOLID), eg.: * adding new functionality should not require modifying existing classes (OCP); * fdsn.Client should not have multiple responsibilities, eg., affected by changes in both routing and fdsnws specifications (SRP). Based on that, routing should probably be implemented as a separate class. But as long as it works, it's OK with me :)

Yea I guess you are right. I would just really like to avoid having a ton more modules - we already have so many. What about having routing clients in the `obspy.clients.fdsn` module?

* `obspy.clients.fdsn.routing.EIDARoutingClient`
* `obspy.clients.fdsn.routing.FederatorRoutingClient`
* `obspy.clients.fdsn.routing.RoutingClient` (if that ever becomes a standard)

> I'm afraid that would not work. Username and password are not derived from the token, but issued by each individual node (/auth method). See also: https://github.com/EIDA/seiscomp3/blob/master/src/trunk/apps/fdsnws/descriptions/fdsnws.rst#authentication-extension

Ah I see - I did not understand the code correctly. But if we have a `EIDARoutingClient` as written above it can just be implemented there.

andres-h:
On 09/19/2016 08:48 PM, Lion Krischer wrote:

>What about having routing
> clients in the |obspy.clients.fdsn| module?
>
>   * |obspy.clients.fdsn.routing.EIDARoutingClient|
>   * |obspy.clients.fdsn.routing.FederatorRoutingClient|
>   * |obspy.clients.fdsn.routing.RoutingClient| (if that ever becomes a
>     standard)

That looks better IMHO.

andres-h:
The documentation at http://docs.obspy.org/pull-requests/1397/ was lost. Would it be possible to recreate it?

krischer:
I don't know what happened there. Maybe a +DOCS triggers another build? @megies: Any clue?

krischer:
Ah no wait - I do know what happened - can you just push another dummy commit? Then it will build the documentation again. The documentation is periodically deleted as the server ran out of inodes...

megies:
> I don't know what happened there. Maybe a +DOCS triggers another build? @megies: Any clue?

> Ah no wait - I do know what happened - can you just push another dummy commit? Then it will build the documentation again. The documentation is periodically deleted as the server ran out of inodes...

This PR just needed a "+DOCS" in a comment. The old docs build was from a time when we built *all* PRs' docs, which resulted in our server getting stalled due to excessive inode usage. All of the old docs builds were deleted when cleaning up our server.

megies:
+TESTS:clients.fdsn

Travis should run FDSN test suites for future commits/pushes to this branch now. Ping me if it doesn't work..

megies:
ping @andres-h, @krischer, 1.1.0 is coming in two weeks roughly. Any chance this will be ready by then? It looks like there's still some open questions on how to best integrate this..?

krischer:
I'm bumping this to 1.2.0 as this is still a bit of work and I probably will not have the time to do it this week. But we definitely want routing in ObsPy.

megies:
@krischer do we need a follow up for auth support in iris federator and eida routing?

andres-h:
Auth support is important for EIDA and I imagine many users want to use 
ObsPy instead of our fdsnws_fetch tool. How does one get restricted data 
with RoutingClient anyway?

krischer:
> @krischer do we need a follow up for auth support in iris federator and eida routing?

I don't know. This is a non-standard authentication method for the fdsnws services but I would not be opposed to add it to the normal fdsnws client as there are a ton of EIDA nodes out there.


> Auth support is important for EIDA and I imagine many users want to use ObsPy instead of our fdsnws_fetch tool. How does one get restricted data with RoutingClient anyway?

Okay - we can just add it to the normal FDSN client. You've also already written all of the code so it can just be taken from your branch.

The routing clients don't support restricted downloads right now but it would also not be very hard to add - we could just do it like we do for the mass downloader, e.g.

```python

c = RoutingClient("eida-routing",
                  client_templates=[Client("GFZ", authdata=...), ...])
```

Then one could add different credentials/authdata for each node. Or we pass some dictionary based on the URL as in your initial implementation. Pros/cons for each approach? Also does EIDA share credentials across nodes, i.e. do we want/need the ability to use the same credentials for every node?

And is there a specification somewhere for the token authentication?

One more question: The example token you've used in the tests - what is it for and can we rely on it long-term for testing?

andres-h:
> Also does EIDA share credentials across 
> nodes, i.e. do we want/need the ability to use the same credentials for 
> every node?

No, but the same authdata works with every node, so you don't need 
dictionary with authdata.


> And is there a specification somewhere for the token authentication?

See http://geofon.gfz-potsdam.de/waveform/archive/auth/auth-example.php

Basically you just post authdata to /auth and it returns (temporary) 
username and password that work with standard FDSNWS.


> One more question: The example token you've used in the tests - what is 
> it for and can we rely on it long-term for testing?

Running "gpg --decrypt" on it, you get:

{"mail": "test@obspy.org", "valid_until": "2020-01-01T00:00:00Z"}

So it authenticates the user as "test@obspy.org" and it can be used 
until 2020.

megies:
> Or we pass some dictionary based on the URL as in your initial implementation. Pros/cons for each approach?

I think it's simpler from a user perspective to enable the user to provide a dict like:

```
credentials = {
    'GFZ': ('user', 'pass')  # or a single string for a credentials file
    ...
    }
RoutingClient(..., credentials=credentials)
```

For paranoia reasons optionally use explicit URL for full transparency to what URL the credentials are sent to..

```
credentials = {
    'http://geofon.gfz-potsdam.de': ('user', 'pass')  # or a single string for a credentials file
    ...
    }
```




> No, but the same authdata works with every node, so you don't need dictionary with authdata.

Oh.. so the above is not needed for the eida routing flavor I guess. But in general we still need it for non-EIDA FDSNWS nodes..?!

megies:
> I think it's simpler from a user perspective to enable the user to provide a dict like:

Or maybe, we could even add a helper routine to `obspy.clients.fdsn`, something like this (some mechanics similar but to how `requests` does it`?

```
from obspy.clients.fdsn import register_fdsnws_credentials, RoutingClient

register_fdsnws_credentials(url_base='geofon.gfz-potsdam.de', user='...', password='...')
register_fdsnws_credentials(url_base='service.iris.edu', credentials_file='...')
register_fdsnws_credentials(url_base='www.orfeus-eu.org', token='...')

client = RoutingClient('iris-federator')  # or 'eida-routing'
client.get_waveforms(...)
```

andres-h:
I guess a problem is that a user doesn't know the list of datacenters in 
before routing.

In case of static credentials, the user at least (hopefully) knows for 
which datacenters he has credentials, so those credentials can be 
registered always.

This would not work well with EIDA datacenters, because any user has 
automatically credentials for any EIDA datacenter that implements /auth.

So a user would have to:

for url_base in any_datacenter_that_implements_auth:
     register_fdsnws_credentials(url_base=url_base, token=token)

megies:
@andres-h is the token mechanism implemented in one single data center for all of eida or is it supposed to be implemented at each respective data center?

megies:
@andres-h can you please confirm that this query should return data from geofon fdsnws (at http://geofon.gfz-potsdam.de/fdsnws/) using the token you set up for testing (https://github.com/obspy/obspy/pull/1397/files#diff-d6cc9a2f8b009bd026035dfb35fc5944R38)? I'm getting no data available..

```
"GE", "KMBO", "00", "BHZ", "2010-02-27T06:30:00.000", "2010-02-27T06:40:00.000"
```

andres-h:
> @andres-h is the token mechanism 
> implemented in one single data center for all of eida or is it supposed 
> to be implemented at each respective data center?
It is supposed to be implemented at each respective data center.

andres-h:
On 10/13/2017 01:17 PM, Tobias Megies wrote:
> @andres-h can you please confirm that this 
> query should return data from geofon fdsnws
Yes, it should, but right now I'm not getting data with or without 
token. I'm just trying to figure out what is wrong with this station. 
The data is in archive.

You can use a different station like WLF (no location code).

andres-h:
KMBO is OK now. We archive this station under GE, but dataselect was 
looking under IU...

megies:
> KMBO is OK now. We archive this station under GE, but dataselect was looking under IU...

Thanks for looking into this. But maybe it's still better to use a station for testing that is not subject to mappings like this to be more stable?
Also.. it seems this is free data? So I can't really use this for testing.. But I guess just testing that the token lookup returns some valid `:` separated string is enough.. since it's otherwise using the normal FDSNWS auth mechanics..

andres-h:
We cannot open real restricted data to test user and I'd like to avoid 
setting up a special test instance. As you say, for client testing it 
does not really make a difference if data is restricted or not. We 
expect that once users get used to tokens, they will use authentication 
when downloading free data as well.

WLF might be a better choice for testing than KMBO.

megies:
Ok, thanks again. Yeah, no need to actually download restricted data I guess, I can just test that resolving the token into user/password works as expected.

So the token you provided is actually not giving any additional permissions, correct?

andres-h:
> So the token you provided is actually not giving any additional 
> permissions, correct?

Correct.

megies:
OK thanks!

megies:
Sorry, one more question.. the "temporary" user/password returned by that token.. will it always stay the same, or will it change at some point?

andres-h:
> Sorry, one more question.. the "temporary" user/password returned by 
> that token.. will it always stay the same, or will it change at some point?
It will change. I guess it is OK to assume that it is valid during one 
run of get_waveforms(). The user/password is supposed to be valid for 24 
hours starting from /auth invocation. If /auth is called more 
frequently, we return the same user/password and elongate its lifetime.

megies:
Alright, so I'll just check that the returned user/pass is some alphanumeric string. Thanks.

megies:
Superseded by #1919 and #1928.

PR #837, opened 2016-05-10
adding spectral methods to trace object

megies:
Coming from the 'old school' interactive side, I dearly miss spectral methods attached directly to a trace object. While the spectrogram, which to me is a more sophisticated method is already available as a method, the direct estimation of psd/multi-taper and fft is not available. Certainly this can be done by modules of scipy or mlab BUT what I'm thinking of is to return a  complex trace (resulting from a fft), manipulate it and possibly transfer it back to time domain. This means that all trace.stats information should be kept while the sampling or delta needs to be modified to frequency domain (i.e. 1/(time window). This would open up the world of trace manipulation in frequency domain, easy cross-spectral estimates, sub-sample shifts etc AND of course an easy and fast way of displaying the frequency content. Possibly this would result in a new 'trace object' and therefore needs a re-definition of the Trace object itself.
What is the opinion of the developers?

Cheers Jo


megies:
@fstettner has started work on this, he'll summarize on the current state shortly, please everybody then help out with comments and corrections..

fstettner:
I started to build up the class hierarchy in trace.py with `BaseStats-TimeSeriesStats-Stats` and `BaseTrace-TimeSeriesTrace-Trace`. Additionally I added a fft-method to the Trace object to go to frequency domain and to the new built class `FrequencyDomainTrace`.
This class contains until now the `ifft-, plot-, differentiation-, integration-, phase-, amplitude-, cross_correlation-, frequencies- and plot_psd_trace- method`, which have not yet been completed. I am looking forward to hear some suggestions, criticism or further proposals.

obspy/core/trace.py
```diff
@@ -2779,6 +3241,207 @@ def _data_sanity_checks(value):
                "arrays are allowed for initialization.") % str(value.shape)
         raise ValueError(msg)
 
+class FrequencyDomainTrace(BaseTrace):
```
krischer: I would move the whole class to a separate file, maybe call it `frequency_domain_trace.py` or something along this line? Small files are easier to work with and this is IMHO a sensible split.

obspy/core/trace.py
```diff
@@ -2779,6 +3241,207 @@ def _data_sanity_checks(value):
...
+
+    def ifft(self):
+        tr = Trace()
+        tr.stats = Stats()
+        tr.stats.starttime = self.stats.starttime
+        tr.data = fftpack.irfft(self.data)
```
krischer: I know this is faster than numpy's `rfft()` but it has this extremely awkward return type that is not a complex array as might be expected but a float array with: [real part of zero frequency, real part of f_1, complex part of f_1, real part of f_2, complex part of f_2, ...]. I would suggest to either convert this to a proper complex array - I'm not sure this is possible without a copy - or just use numpy's `rfft` for now. Other opinions?

obspy/core/trace.py
```diff
@@ -2779,6 +3241,207 @@ def _data_sanity_checks(value):
...
+        return tr
+
+    def plot(self):
+        N = self.stats.npts
+        T = self.stats.delta
+        x = np.linspace(0.0, N*T, N)
+        yf = self.data
+        xf = np.linspace(0.0, 1.0/(2.0*T), N/2)
+        fig, ax = plt.subplots()
+        ax.plot(xf, 2.0/N * np.abs(yf[:N/2]))
```
krischer: This is not really correct due to the above mentioned issue with the return type of scipy's rfft.

obspy/core/trace.py
```diff
@@ -2779,6 +3241,207 @@ def _data_sanity_checks(value):
...
+            plt.xlabel('offset between two waves')
+            plt.ylabel('correlation')
+            plt.show()
+
+            return (corr)        
+
+    def frequencies(self):
+        n = self.data.size
+        timestep = self.stats.delta
+        freq = fftpack.fftfreq(n, d=timestep)
```
krischer: You likely want to use `rfftfreq` but this is also very awkward to the previously mentioned issue.

obspy/core/trace.py
```diff
@@ -2741,6 +3191,19 @@ def remove_response(self, inventory=None, output="VEL", water_level=60,
         self.data = data
         return self
 
+    def fft(self, window_function='cosine', percentage=0.1):
```
krischer: I think this should rather be attached to the `TimeSeriesTrace`. I guess the `fft()` method then has to do some introspection to decide if it wants to return `FrequencyDomainTimeSeriesTrace` or  a `FrequencyDomainTrace`.

krischer:
I still very much like the basic structure. I added some comments to the code - we'll give this a full review once its more fleshed out.

I think we also want a corresponding frequency domain class for the new `TimeSeriesTrace` as illustrated in the following amazing ASCII Art. Any ideas for better names?

```
                   BaseTrace

                    /     \
                  /         \
                /             \

TimeSeriesTrace     <----->    TimeSeriesFrequencyDomainTrace
       |                                 |
       |                                 |
     Trace          <----->    FrequencyDomainTrace
````

The current implementation also has the issue I commented on the code with the weird return type of `scipy.fftpack.rfft()`. This also has the effect that most other methods are wrong as they are currently implemented in a way that assumes a complex array as any sane person would. So I propose to either manually cast to a complex array (I think this unfortunately needs a copy due to the missing complex part of the zero frequency value) or just use numpy's `rfft()` with all its known issues (at least for now).

File: obspy/core/trace.py (hunk 2 above)
Comment (megies) [inline obspy/core/trace.py]: > extremely awkward return type

IIRC that return type is for historic reasons, I think it was like this in old numerical recipes C codes. There might be a chance to convert it without copy when leaving out the zero frequency value, and prepend that one afterwards?

Comment (krischer) [inline obspy/core/trace.py]: Whatever the reason - it is super confusing and IMHO makes it very hard to work with it. Judging by some scipy issues, the scipy devs seem to largely have the same opinion ;-)

We definitely want data arrays that are contiguous in memory (for eventually talking with any C algorithms) and prepending for example with `np.concatenate()` will create a copy. I'm not sure if there is a way to grow a numpy array without a copy. It might be possible using `realloc()` from C but that is also not guaranteed to not create a copy.

Comment (megies) [inline obspy/core/trace.py]: Well, yeah, our array would be one item longer (imaginary part ==0 at zero frequency) than the one we get, so there's no chance to avoid a copy by just setting a different type or other tricks..

**Comment by fstettner** [comment]
Until now the structure looks like:

hierarchydiagram
1. Why do we need a TimeSeriesFrequencyDomainTrace?
2. I started to build up the **frequency_domain_trace.py**, but got problems with the circular imports in python, because I generate in the **tr.fft** Method an instance of the **FrequencyDomainTrace** class and in the **tr_f.ifft**  Method an instance of the **Trace** class. Is there a nice way to do it?

**Comment by megies** [comment]
> 2. I started to build up the frequency_domain_trace.py, but got problems with the circular imports in python, because I generate in the tr.fft Method an instance of the FrequencyDomainTrace class and in the tr_f.ifft Method an instance of the Trace class. Is there a nice way to do it?

If it's gonna be in separate files as @krischer proposed, the only way is to do the `from obspy import Trace` import not at top of `frequency_domain_trace.py` but inside the `def ifft(self, ...):` block (and vice versa with import of `FrequencyDomainTrace` in `trace.py`).

**Comment by fstettner** [comment]
I moved the FrequencyDomainTrace to the **frequency_domain_trace.py** like **@krischer** recommended. Additionally, I built in the **numpy** rfft, irfft and fftfreq , because of the return type of the scipy version. The fft is now attached to the TimeSeriesTrace with an differentiation of Trace and TimeSeriesTrace. I also started to build up the **TimeSeriesFrequencyDomainTrace**,  that has currently just the ifft implemented.

File: obspy/core/frequency_domain_trace.py [hunk 6]
```diff
@@ -88,9 +86,50 @@ def cross_correlation(self, tr2):
...
+        data2_conj = np.conj(data2)
+        cross_spectrum = data2_conj*data1
+        return (cross_spectrum)
+
+    def coherence(self, tr2):
+        data1 = self.data
+        data2 = tr2.data
+        data2_conj = np.conj(data2)
+        cross_spectrum = data2_conj*data1
+        coherence = (cross_spectrum)**2/(np.abs(data1)*np.abs(data2))
```
megies: @fstettner, smoothing of spectra is missing here..

In the long run I would propose adding a method to FFTTrace, e.g. `def smoothed(self, window, npts)` (or width given in Hz).
For now you could just hardcode some fixed window from scipy, see http://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.get_window.html#scipy.signal.get_window. For orientation, Poupinet 1984 used a five point triangular window for 100Hz sampled data.

CC @jwassermann

obspy/core/frequency_domain_trace.py
```diff
@@ -88,9 +86,50 @@ def cross_correlation(self, tr2):
...
-        freq = fft.fftfreq(n, d=timestep)       
+        freq = fft.rfftfreq(n, d=timestep)       
         return freq
 
+    def cross_spectrum(self, tr2):
+        data1 = self.data
+        data2 = tr2.data
+        data2_conj = np.conj(data2)
+        cross_spectrum = data2_conj*data1
+        return (cross_spectrum)
```
megies: No brackets needed here (and in other return statements).

obspy/core/frequency_domain_trace.py
```diff
@@ -88,9 +86,50 @@ def cross_correlation(self, tr2):
     def frequencies(self):
```
megies: I'd propose making this a property so it can be used like.. `tr.frequencies`

obspy/core/frequency_domain_trace.py
```diff
@@ -88,9 +86,50 @@ def cross_correlation(self, tr2):
...
+        data1 = self.data
+        data2 = tr2.data
+        data2_conj = np.conj(data2)
+        cross_spectrum = data2_conj*data1
+        return (cross_spectrum)
+
+    def coherence(self, tr2):
+        data1 = self.data
+        data2 = tr2.data
+        data2_conj = np.conj(data2)
```
megies: Maybe a convenience property on FFTtrace would be useful..  `tr.conjugate` that returns `np.conjugate(tr.data)`?

fstettner:
I implemented the conjugate method. The coherence-method also should work now (with hanning).
I still have problems with the normalization of the cross_correlation-method. Are there some ideas?
Are the differentiation and integration method useful implemented?

obspy/core/frequency_domain_trace.py
```diff
@@ -0,0 +1,245 @@
...
+        t_tr.stats = self.stats
+        return t_tr
+
+    def plot(self):
+        N = self.stats.npts
+        T = self.stats.delta
+        yf = self.data
+        xf = np.linspace(0.0, 1.0/(2.0*T), N/2)
+        fig, ax = plt.subplots()
+        ax.plot(xf, 2.0/N * np.abs(yf[:N/2]))
```
krischer: This should just use the `self.frequencies()` method which would me much easier.

obspy/core/frequency_domain_trace.py
```diff
@@ -0,0 +1,245 @@
...
+        plt.xlabel('Freq (Hz)')
+        plt.ylabel('|Y(freq)|')
+        plt.show()
+
+    def differentiation(self, order_of_differention = 1):
+        self.data = fftpack.diff(self.data, order = order_of_differention)
+        return self
```
krischer: The diff method of fftpack expects data in the time domain. Example:

```python
import numpy as np
import scipy.fftpack

x = np.linspace(0, 2 * np.pi, 10000)[:-1]
plt.plot(x, scipy.fftpack.diff(np.sin(x)))
plt.plot(x, np.cos(x))
plt.show()
```

obspy/core/frequency_domain_trace.py
```diff
@@ -59,77 +44,69 @@ def phase(self):
...
-    def cross_correlation(self, tr2):
-
-        if self.stats.starttime != tr2.stats.starttime:
+    def cross_correlation(self, other):
+        if self.stats.starttime != other.stats.starttime:
             msg = "Traces do not have the same starttime"
             raise Exception(msg)
-       
-        elif "FALSE" == np.array_equal(self.frequencies(), tr2.frequencies()):
+        elif "FALSE" == np.array_equal(self.frequencies, other.frequencies):
```
megies: You're comparing a string with a boolean value, so this will never evaluate to `True`.
`np.array_equal` is returning a `bool` value.. `np.array_equal(...) is False`

ThomasLecocq:
👍 to get this done asap!

megies:
This was done as a student project and likely needs more work.. :-/

PR #1401, opened 2016-05-10
add helper method to create a dummy Stream from its string representation

megies:
Just a tiny helper staticmethod on `Stream` mostly for debugging purposes, e.g as in #1141.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -18,6 +18,7 @@ import fnmatch
 import math
 import os
 import pickle
+import re
 import warnings
 from glob import glob, has_magic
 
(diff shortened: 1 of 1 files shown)


obspy/core/stream.py
```diff
@@ -3045,6 +3046,51 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        >>> s = os.linesep.join(s)
+        >>> st = Stream._dummy_stream_from_string(s)
+        >>> print(st)  # doctest: +ELLIPSIS
+        3 Trace(s) in Stream:
+        IU.GRFO..HH2 | 2016-01-07T00:00:00.008300Z ... | 10.0 Hz, 301 samples
+        XX.GRFO..HH1 | 2016-01-07T00:00:02.668393Z ... | 100.0 Hz, 686 samples
+        IU.ABCD..EH2 | 2016-01-07T00:00:09.528393Z ... | 100.0 Hz, 4086 samples
+        """
+        st = Stream()
+        for line in s.splitlines():
```
krischer: Maybe add `line = line.strip()` to avoid weird errors with extra whitespaces and what not which easily happens if copying from somewhere.

krischer:
Considering its a private method it could also be placed in `core/utils`. It also needs at least a basic test.

krischer:
nvm...there actually is a doctest...

megies:
> Considering its a private method it could also be placed in core/utils.

Put it there for a start, but doesn't work because of circular imports.. being private it should not harm much by cluttering namespace I think, at least not for interactive tab completion.

obspy/core/stream.py
megies: done

krischer:
> Put it there for a start, but doesn't work because of circular imports.. being private it should not do much harm in terms of cluttering the namespace I think, at least not for interactive tab completion.

Ah I see...you could always just import the necessary things within the function but that's always a bit ugly. My main reason for the suggestion to move it is that `stream.py` is already so very large so moving things elsewhere seems to be preferable if possible. But here I agree that the hassle with the circular imports outweighs that so feel free to leave it where it is if you want to.

Can IMHO be merged if CI passes.

megies:
Yeah.. these large files are hard to handle at times. But I think I'll leave it where it is now.. a) for circular import reason and b) because it logically fits best where it is right now.

PR #1400, opened 2016-05-10
Iterative reading of SEG-Y and SU files

krischer:
Hi,

I guess this is not an issue per se, but I am new to the geosciences and I need some help (my config: Windows 8, Anaconda Python3.4, ObsPy 1.0.1, VisualStudio C++).

I am trying to read a SEG-Y file corresponding to a 2D line with 1.5M traces and getting out of memory in less than 2 min. Will the entire file contents always be read into memory? Is there some kind of workaround? Should I be doing it differently?

If I have lines with almost 16GB each, then the only way possible to go through it would be to have a computer with a huge amount of RAM?

Thanks in advance

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -23,6 +23,10 @@ master:
      white background.
  - obspy.io.css
     * Read support for NNSA KB Core format waveform data. (see #1332)
+ - obspy.io.segy:
+    * Iterative reading of large SEG-Y and SU files with
+	  `obspy.io.segy.segy.iread_segy` and `obspy.io.segy.segy.iread_su`.
+	  (see #1400).
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -16,13 +16,13 @@ master:
      arguments to not download certain pieces of data. (see #1305)
    * The mass downloader can now download stations that are part of a given
      inventory object.
- - obspy.css
-    * Read support for NNSA KB Core format waveform data. (see #1332)
  - obspy.imaging.cm
    * new colormap: viridis_white. This is a modification of viridis that
      goes to white instead of yellow but remains perceptually uniform. It
      is especially useful for printing when an image should merge with the
      white background.
+ - obspy.io.css
+    * Read support for NNSA KB Core format waveform data. (see #1332)
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
--- a/obspy/io/segy/__init__.py
+++ b/obspy/io/segy/__init__.py
@@ -29,7 +29,7 @@ file formats usually used in observatories (GSE2, MiniSEED, ...). The
 of ObsPy are therefore not fully suited to handle them. Nonetheless they work
 well enough if some potential problems are kept in mind.
 
-SEG Y files can be read in three different ways that have different
+SEG Y files can be read in four different ways that have different
 advantages/disadvantages. Most of the following also applies to SU files with
 some changes (keep in mind that SU files have no file wide headers).
 
@@ -37,6 +37,9 @@ some changes (keep in mind that SU files have no file wide headers).
 2. Using the :mod:`obspy.io.segy` specific
    :func:`obspy.io.segy.core._read_segy` function.
 3. Using the internal :func:`obspy.io.segy.segy._read_segy` function.
+4. Some SEG-Y files are too large to be read into memory. The
+   :func:`obspy.io.segy.segy.iread_segy` function reads a large file trace
+   by trace circumventing this problem.
 
 Reading using methods 1 and 2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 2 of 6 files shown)


krischer:
This issue is now a PR that adds a streaming/iterator interface to `obspy.io.segy` for SEG-Y as well as SU files. It can work with arbitrarily large files without running into memory issues (of course assuming that the user does not collect the returned traces...).

It works like this:

```python
from obspy.io.segy.segy import iread_segy

# The iterator yields one ObsPy Trace per SEG-Y line.
for trace in iread_segy("file.segy"):
    # Do something useful!
    print(trace.data.mean())
```

rmlopes:
Nice, thanks a lot.

rmlopes:
I cloned and pulled the PR locally, but I am getting a KeyError (see bellow). Also, I am not sure why everything says `amd64` and if it affects anything, but shouldn't it be `x64` for Intel?

`Traceback (most recent call last):
  File "C:\Users\rmlopes\dev\obspy\build\lib.win-amd64-3.4\obspy\__init__.py", line 41, in <module>
    from obspy.core.event import read_events, Catalog
  File "C:\Users\rmlopes\dev\obspy\build\lib.win-amd64-3.4\obspy\core\event\__init__.py", line 29, in <module>
    from .catalog import Catalog, read_events
  File "C:\Users\rmlopes\dev\obspy\build\lib.win-amd64-3.4\obspy\core\event\catalog.py", line 39, in <module>
    from obspy.imaging.cm import obspy_sequential
  File "C:\Users\rmlopes\dev\obspy\build\lib.win-amd64-3.4\obspy\imaging\cm.py", line 299, in <module>
    obspy_sequential = _globals["viridis"]
KeyError: 'viridis'`

krischer:
Hmm...weird error as its completely unrelated to this PR. Did you previously install ObsPy with other means (e.g. conda) and than reinstall via git? Can you try to uninstall completely (`pip uninstall obspy`, `conda remove obspy`) and then reinstall from git and also run the `setup.py` (`pip install -v -e .` in the obspy directory)?

krischer:
AFAIK `amd64` is just another name for `x86_64` due to some historical reasons.

rmlopes:
This is a local build only, I am not running install or copying by hand into the distribution. I had removed obspy by hand previously, nevertheless as these funny things may happen, I did it all again. I removed anything that would eventually still be there with `conda remove obspy`, rebuilt the local copy of the module and retested. I am still having the same `KeyError`.

krischer:
@barsch @ThomasLecocq Can you help here? I have no idea how to build on windows to be honest.

barsch:
```unresolved external symbol``` means usually that some function is missing in those .def files needed for win compilers - in this case its not even finding it

you have to use ```python setup.py develop``` - all the windows specific magic is happening within setup.py - which for whatever reason is skipped if installing via pip

**Comment by rmlopes** [comment]
The development build goes a bit further, but there is an assertion fail in the tests (total 3 failures, 1 error). I've sent the report through the script. I also have a lot of  '(...) misses imports: (...)' messages from every module, I am guessing it is not very relevant.

Back to the 'viridis' KeyError, shouldn't the specific windows magic also be present in the normal build?

**Comment by barsch** [comment]
you could try doing a clean build again with:

```
python setup.py clean --all
python setup.py develop
```

Running ```obspy-runtests.py ``` on Windows will result into a few numpy/scipy related warning which you can ignore. However you shouldn't get exceptions. And I never ran into the strange keyerror you got - so sorry no clue ....

However, those changes from @krischer are all made within Python, so you really dont need to build anything - so you are probably faster by reinstalling the original (conda/PyPi) packages you used before, download the ZIP of the current PR: https://github.com/krischer/obspy/archive/segy-iterator.zip and copy all changed files over - you can see the list of changed files here: https://github.com/obspy/obspy/pull/1400/files

rmlopes:
I am guessing that probably people in windows are using MinGW. I cannot do it though as I have project requirements for VS C++, and I don't want to maintain two different versions. I would be happy to investigate this further and try to fix it, but I need some pointers as I am new to the library.

Cleaning and rebuilding is not a like a magic pill you can take, as I said, I am doing the builds locally (and of course I am always cleaning in between). 
In the meanwhile I will take the suggestion to manually copy the PR files into the conda/pip installation. Thanks for the input @barsch.

barsch:
I'm currently the one doing windows builds for ObsPy and upload it to PyPI - all builds are done using VC and not MinGW as strongly recommended by the developers of Python. So if you really want to build it locally, we can continue to investigate by sending me/posting your complete build log;)

BTW: Are you using Visual Studio 2010?

rmlopes:
Let's go for it. I've posted the complete build log  for the [regular build](https://gist.github.com/rmlopes/8c904d1822764f6e1af0590d8f1a3c6a) and the [develop build](https://gist.github.com/rmlopes/23d891bece2eac422d5115ef2d98a3aa).

barsch:
looks good for me - I don't see warnings or errors - whats the output of obspy-runtests?

So you are using VC 12 - keep in mind that the official recommended compiler for Python 3.4 extensions is VC 10 - so it might be part of the problem - we will see ...

rmlopes:
The [test results in develop mode](https://gist.github.com/rmlopes/919b6a0750fafc9235206b066133eb24) (for the regular build is the `viridis` KeyError posted before). 

Ok, I think those posts explain my problems. Conda provides a binary distribution (of 3.4), that has apparently been compiled using MSVC10, and now I am trying to compile extensions using MSVC12.

krischer:
IMHO ready to merge. Can someone review it?

rmlopes:
I am trying to install with the PR on a Linux x64 VM, Python 3.5. I really need the feature. :)

I am being forced to use a proxy to access internet, and I think it is causing some failures on the tests. So,

1. I don't see a ```--proxy``` option, similar to pip or conda. Is there a way to do it?
2. If there is no way to specify a proxy, what should I exclude (```-x```) to avoid the network modules?

krischer:
@rmlopes It should respect the `HTTP_PROXY` and `HTTPS_PROXY` env variables.

Can I otherwise merge this PR?

PR #1399, opened 2016-05-16
SAC file truncated to zero on write if an invalid header field is set

jkmacc-LANL:
Here's a test code:

```python
from obspy import read

tr = read('trace.sac')[0]
tr.stats.sac.AAA = 10.
tr.write('trace.sac', format='SAC')
```

which gives:
```
ValueError: Unrecognized header name: AAA.
```

And truncates to zero `trace.sac`.

I think that setting an invalid header field should be tested in order to prevent future regressions.

### DIFF
--- a/obspy/io/sac/arrayio.py
+++ b/obspy/io/sac/arrayio.py
@@ -514,7 +514,8 @@ def dict_to_header_arrays(header=None, byteorder='='):
                     #                                          'strict')
                     hs[HD.STRHDRS.index(hdr)] = value
             else:
-                raise ValueError("Unrecognized header name: {}.".format(hdr))
+                msg = "Unrecognized header name: {}. Ignored.".format(hdr)
+                warnings.warn(msg)
 
     return hf, hi, hs
 
--- a/obspy/io/sac/tests/test_core.py
+++ b/obspy/io/sac/tests/test_core.py
@@ -839,6 +839,29 @@ class CoreTestCase(unittest.TestCase):
         self.assertEqual(tr1.stats.sac.npts, tr.stats.sac.npts / 2)
         self.assertEqual(tr1.stats.sac.delta, tr.stats.sac.delta * 2)
 
+    def test_invalid_header_field(self):
+        """
+        Given a SAC file on disk, when it is read and an invalid header is
+        appended to the stats.sac dictionary, then the invalid header should be
+        ignored (user given a warning) and the written file should be the same
+        as the original.
+        """
+        tr = read(self.file, format='SAC')[0]
+
+        with io.BytesIO() as buf:
+            tr.write(buf, format='SAC')
+            buf.seek(0, 0)
+
+            tr.stats.sac.AAA = 10.
+            with warnings.catch_warnings(record=True) as w:
+                warnings.simplefilter('always')
+                with io.BytesIO() as buf1:
+                    tr.write(buf1, format='SAC')
+                    self.assertIn('Ignored', str(w[-1].message))
+                    buf1.seek(0, 0)
+
+                    self.assertEqual(buf.read(), buf1.read())
+
 
 def suite():
     return unittest.makeSuite(CoreTestCase, 'test')


jkmacc-LANL:
This issue has been converted to a Pull Request.  Writing SAC files with an unrecognized header now issues a warning instead of a ValueError.  Any such headers are simply ignored during write.  A test for the warning and the behavior has been added into `test_core.py`.

megies:
Good to go, I guess. Again, up to you SAC guys to decide if it needs a changelog or not.

claudiodsf:
Ok, for me. No changelog needed.

I was just curious to know if the test is ok for you, @megies, the way it is written, given that @jkmacc-LANL left a big `TODO`.

megies:
Have to admit, didn't look at testing details. Relying mostly on the SAC people's judgment for these tickets (you're the ones get hurt after all :stuck_out_tongue_winking_eye: ).

Hmm.. checking file length doesn't feel very right for me, e.g. it's not making sure that you're not overwriting some other header field with the invalid one. (Although you check the warning, so this is really for the ultra-paranoid).
But how about reading the test file and then writing to two BytesIO objects, first w/o and then w/ invalid header field, and simply comparing the BytesIO objects after reading them again? That would be rigorous, I think. (and also faster than temp file)

jkmacc-LANL:
@megies I'll try using BytesIO.  Thank for the idea.  I was catching warnings because I had to do so for a different issue, so I did for this issue.  Something about tests running with `-vvv` were failing if warnings weren't caught.  @krischer Do you remember what I'm referring to?  I'd rather not catch them (more code, probably slower tests), but I didn't want to mess anything up.

megies:
Catching those warnings and checking that they appear is totally what you should do in the tests. :+1:

jkmacc-LANL:
I think this is good to go, from my perspective.  Thanks for the BytesIO help, Tobias.

obspy/io/sac/arrayio.py
```diff
@@ -514,7 +514,8 @@ def dict_to_header_arrays(header=None, byteorder='='):
                     #                                          'strict')
                     hs[HD.STRHDRS.index(hdr)] = value
             else:
-                raise ValueError("Unrecognized header name: {}.".format(hdr))
+                msg = "Unrecognized header name: {}. Ignored.".format(hdr)
```
QuLogic: I think this should just be one sentence.

claudiodsf:
Ok from my point of view, after addressing @QuLogic comment.
As I said before, no need for ChangeLog for this.

My advice is always to squash the commits into one, before merging.

obspy/io/sac/arrayio.py
jkmacc-LANL: I think this becomes less readable as one line, as two things are happening, and it'll go over 72 columns.  OK if it stays as is?

megies:
Thanks for fixing, @jkmacc-LANL!

megies:
> My advice is always to squash the commits into one, before merging.

Sometimes intermediate commits do have some value (e.g. when illlustrating some alternative implementation that would have been better but didn't work for some reason etc.). In those cases it's still good to keep some more granularity.

But yes, in general squashing PRs is a good thing to do.

jkmacc-LANL:
In the future, I'll keep it to just a few relevant commits, not just one.  Thanks everyone!

claudiodsf:
Ok, sorry for being so strict, then 😊

megies:
> In the future, I'll keep it to just a few relevant commits, not just one. Thanks everyone!

Single commits are perfectly fine, in most cases the fewer commits the better.

> Ok, sorry for being so strict, then :blush:

Nah, you're right in almost all cases. But we also don't want to be too hard on people, especially those new to git/github.

QuLogic:
Yes, don't fall into the trap that there are only two options: everything ever vs. one commit only. There is a middle ground somewhere in between that is more reasonable.

claudiodsf:
As a basic rule, I think we should avoid commits like "Fix for previous commit"

PR #1406, opened 2016-05-13
Fix for plotting percentiles on a frequency axis.

rannof:
Adding a check for frequency axis and converting percentiles periods to frequency.

PR #1416, opened 2016-05-19
fix some issues with PPSD.plot with frequency x-axis

megies:
Follow-up to #1406, fixing remaining issues with PPSD plotting with frequencies on x-Axis. Add an image test for it as well.


megies:
I think with this one we should start thinking about a 1.0.2 release..

PR #1418, opened 2016-05-20
Fix bad documentation of quantity errors in event objects

megies:
Also improved class structure for QuantityError, and warning on setting a non-expected attribute (I think we should do this for all Event-type objects -- quite a few times I've seen people wondering why they're setting some attribute wasn't working, just to find out that it was just a typo in the attribute name).

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -40,10 +40,58 @@ from obspy.core.util.decorator import deprecated
 
 
 class QuantityError(AttribDict):
-    uncertainty = None
-    lower_uncertainty = None
-    upper_uncertainty = None
-    confidence_level = None
+    """
+    Uncertainty information for a physical quantity.
+
+    :type uncertainty: float
+    :param uncertainty: Uncertainty as the absolute value of symmetric
+        deviation from the main value.
+    :type lower_uncertainty: float
+    :param lower_uncertainty: Uncertainty as the absolute value of deviation
+        from the main value towards smaller values.
+    :type upper_uncertainty: float
+    :param upper_uncertainty: Uncertainty as the absolute value of deviation
+        from the main value towards larger values.
+    :type confidence_level: float
+    :param confidence_level: Confidence level of the uncertainty, given in
+        percent (0-100).
+    """
+    defaults = {"uncertainty": None, "lower_uncertainty": None,
+                "upper_uncertainty": None, "confidence_level": None}
+    warn_on_non_default_key = True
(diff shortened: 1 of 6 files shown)


obspy/core/event/base.py
```diff
@@ -40,10 +40,41 @@
...
+    :param lower_uncertainty: Uncertainty as the absolute value of deviation
+        from the main value towards smaller values.
+    :type upper_uncertainty: float
+    :param upper_uncertainty: Uncertainty as the absolute value of deviation
+        from the main value towards larger values.
+    :type confidence_level: float
+    :param confidence_level: Confidence level of the uncertainty, given in
+        percent (0-100).
+    """
+    defaults = {"uncertainty": None, "lower_uncertainty": None,
```
QuLogic: I'm not sure I understand why this is a dictionary?

obspy/core/event/base.py
```diff
@@ -40,10 +40,41 @@
...
+    def __init__(self, uncertainty=None, lower_uncertainty=None,
+                 upper_uncertainty=None, confidence_level=None):
+        super(QuantityError, self).__init__()
+        self.uncertainty = uncertainty
+        self.lower_uncertainty = lower_uncertainty
+        self.upper_uncertainty = upper_uncertainty
+        self.confidence_level = confidence_level
+
+    def __bool__(self):
+        if any([getattr(self, key) is not None for key in self.defaults]):
```
QuLogic: Can you not just return this? Why the `if`?

obspy/core/event/base.py
megies: I don't know either.. in any case that's the current implementation of AttribDict..

obspy/core/event/base.py
megies: Yeah.. right. I'll change it..

obspy/core/event/base.py
megies: Ah.. that's for the `Stats` default values for `sampling_rate` etc. I think..

krischer:
Maybe add a quick test?

megies:
For the new `__bool__()`? 5f06227 OK?

krischer:
I originally meant a test for the new raised warning but testing `__bool__()` is of course also good to have :-)

megies:
OK done. I think we should issue a warning for all Event-type objects (i.e. set `defaults` and `warn_on_non_default_key=True` in the event type factory).

PR #1420, opened 2016-05-22
CI: Use Miniconda already available on AppVeyor.

QuLogic:
AppVeyor has Miniconda already available, so we don't really need to download it ourselves. Maybe it'll be a little faster if we're lucky.


krischer:
Kind of hard to tell if it makes a difference as there is so much variance in the runtimes on AppVeyor. But this here is definitely amongst the fastest so I'm fine with merging it!

megies:
:+1: for using what's there. We can always go back if it should stop working for us.

PR #1422, opened 2016-05-24
Fixing issue with location code handling.

krischer:
Fixing issue with location code handling.

Passed empty location codes where stripped from the request thus triggering the server to return data with any location code. This is now fixed.


krischer:
IMHO ready to merge once CI passes. Feel free to review but this is quite a small fix.

PR #1419, opened 2016-05-25
unable to read large mseed file

krischer:
When I'm trying to read large mseed file (3 to 8 GB)

`st = obspy.read('./mseed/TM_TMDB_2012.mseed')`

> <decorator-gen-155> in read(pathname_or_url, format, headonly, starttime, endtime, nearest_sample, dtype, apply_calib, **kwargs)
> 
> /home/user/anaconda/lib/python2.7/site-packages/obspy/core/util/decorator.pyc in _map_example_filename(func, _args, *_kwargs)
>     287                         except IOError:
>     288                             pass
> --> 289         return func(_args, *_kwargs)
>     290     return _map_example_filename
>     291 
> 
> /home/user/anaconda/lib/python2.7/site-packages/obspy/core/stream.pyc in read(pathname_or_url, format, headonly, starttime, endtime, nearest_sample, dtype, apply_calib, **kwargs)
>     241             # set start/end time. Not sure what to do in this case.
>     242             elif not starttime and not endtime:
> --> 243                 raise Exception("Cannot open file/files: %s" % pathname)
>     244     # Trim if times are given.
>     245     if headonly and (starttime or endtime or dtype):
> 
> Exception: Cannot open file/files: ./mseed/TM_TMDB_2012.mseed

If I specify the `starttime` and/or `endtime`, no exceptions raise but it returns an empty Stream without Traces.

### DIFF
--- a/obspy/io/mseed/src/libmseed/parseutils.c
+++ b/obspy/io/mseed/src/libmseed/parseutils.c
@@ -41,8 +41,8 @@
  *  <0 : libmseed error code (listed in libmseed.h) is returned.
  *********************************************************************/
 int
-msr_parse ( char *record, int recbuflen, MSRecord **ppmsr, int reclen,
-	    flag dataflag, flag verbose )
+msr_parse ( char *record, unsigned long long recbuflen, MSRecord **ppmsr,
+            int reclen, flag dataflag, flag verbose )
 {
   int detlen = 0;
   int retcode = 0;
@@ -144,8 +144,8 @@ msr_parse ( char *record, int recbuflen, MSRecord **ppmsr, int reclen,
  * returned when end-of-buffer is reached.
  *********************************************************************/
 int
-msr_parse_selection ( char *recbuf, int recbuflen, int64_t *offset,
-		      MSRecord **ppmsr, int reclen,
+msr_parse_selection ( char *recbuf, unsigned long long recbuflen,
+                      int64_t *offset, MSRecord **ppmsr, int reclen,
 		      Selections *selections, flag dataflag, flag verbose )
 {
   int retval = MS_GENERROR;
@@ -228,7 +228,7 @@ msr_parse_selection ( char *recbuf, int recbuflen, int64_t *offset,
  * >0 : size of the record in bytes
  *********************************************************************/
 int
-ms_detect ( const char *record, int recbuflen )
+ms_detect ( const char *record, unsigned long long recbuflen )
 {
   uint16_t blkt_offset;    /* Byte offset for next blockette */
   uint8_t swapflag  = 0;   /* Byte swapping flag */
--- a/obspy/io/mseed/src/libmseed/parseutils.c
+++ b/obspy/io/mseed/src/libmseed/parseutils.c
@@ -54,7 +54,7 @@ msr_parse ( char *record, unsigned long long recbuflen, MSRecord **ppmsr,
     return MS_GENERROR;
   
   /* Sanity check: record length cannot be larger than buffer */
-  if ( reclen > 0 && reclen > recbuflen )
+  if ( reclen > 0 && (unsigned int) reclen > recbuflen )
     {
       ms_log (2, "ms_parse() Record length (%d) cannot be larger than buffer (%d)\n",
 	      reclen, recbuflen);
@@ -96,7 +96,7 @@ msr_parse ( char *record, unsigned long long recbuflen, MSRecord **ppmsr,
     }
   
   /* Check if more data is required, return hint */
-  if ( reclen > recbuflen )
+  if ( (unsigned int) reclen > recbuflen )
     {
       if ( verbose > 2 )
 	ms_log (1, "Detected %d byte record, need %d more bytes\n",
@@ -162,7 +162,7 @@ msr_parse_selection ( char *recbuf, unsigned long long recbuflen,
   if ( ! offset )
     return MS_GENERROR;
   
-  while ( *offset < recbuflen )
+  while ( (unsigned long long)*offset < recbuflen )
     {
       retval = msr_parse (recbuf+*offset, (int)(recbuflen-*offset), ppmsr, reclen, 0, verbose);
       
@@ -276,7 +276,7 @@ ms_detect ( const char *record, unsigned long long recbuflen )
       
       /* Found a 1000 blockette, not truncated */
       if ( blkt_type == 1000  &&
-	   (int)(blkt_offset + 4 + sizeof(struct blkt_1000_s)) <= recbuflen )
+	   (unsigned long long)(blkt_offset + 4 + sizeof(struct blkt_1000_s)) <= recbuflen )
 	{
           blkt_1000 = (struct blkt_1000_s *) (record + blkt_offset + 4);
 	  
@@ -306,7 +306,7 @@ ms_detect ( const char *record, unsigned long long recbuflen )
       nextfsdh = record + MINRECLEN;
       
       /* Check for record header or blank/noise record at MINRECLEN byte offsets */
-      while ( ((nextfsdh - record) + 48) < recbuflen )
+      while ( (unsigned long long)((nextfsdh - record) + 48) < recbuflen )
 	{
 	  if ( MS_ISVALIDHEADER(nextfsdh) || MS_ISVALIDBLANK(nextfsdh) )
             {
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -374,7 +374,7 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
         verbose = 0
 
     lil = clibmseed.readMSEEDBuffer(
-        bfr_np, buflen, selections, C.c_int8(unpack_data),
+        bfr_np, C.c_ulonglong(buflen), selections, C.c_int8(unpack_data),
         reclen, C.c_int8(verbose), C.c_int8(details), header_byteorder,
         alloc_data, diag_print, log_print)
 
--- a/obspy/io/mseed/headers.py
+++ b/obspy/io/mseed/headers.py
@@ -701,7 +701,7 @@ LinkedIDList._fields_ = [
 clibmseed.readMSEEDBuffer.argtypes = [
     np.ctypeslib.ndpointer(dtype=np.int8, ndim=1,
                            flags=native_str('C_CONTIGUOUS')),
-    C.c_int,
+    C.c_ulonglong,
     C.POINTER(Selections),
     C.c_int8,
     C.c_int,
--- a/obspy/io/mseed/src/libmseed/doc/ms_find_reclen.3
+++ b/obspy/io/mseed/src/libmseed/doc/ms_find_reclen.3
@@ -6,7 +6,8 @@ Determine SEED record data length
 .nf
 .B #include <libmseed.h>
 
-.BI "int  \fBms_find_reclen\fP ( const char *" recbuf ", int " recbuflen ",
+.BI "int  \fBms_find_reclen\fP ( const char *" recbuf ",
+.BI                              unsigned long long " recbuflen ",
 .BI "                            FILE " *fileptr " );
 .fi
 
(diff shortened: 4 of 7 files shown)


krischer:
```
14 Trace(s) in Stream:
TM.TMDB..BHZ | 2014-01-01T00:00:00.000000Z - 2014-01-14T07:59:38.440000Z | 100.0 Hz, 115197845 samples
TM.TMDB..BHZ | 2014-01-14T08:00:28.160000Z - 2014-01-14T08:11:39.980000Z | 100.0 Hz, 67183 samples
TM.TMDB..BHZ | 2014-01-14T08:11:40.000000Z - 2014-01-14T08:33:09.980000Z | 100.0 Hz, 128999 samples
TM.TMDB..BHZ | 2014-01-14T08:33:10.000000Z - 2014-01-14T12:50:09.990000Z | 100.0 Hz, 1542000 samples
TM.TMDB..BHZ | 2014-01-14T12:50:17.210000Z - 2014-01-14T12:50:19.990000Z | 100.0 Hz, 279 samples
TM.TMDB..BHZ | 2014-01-14T12:50:40.000000Z - 2014-01-20T11:11:29.790000Z | 100.0 Hz, 51244980 samples
TM.TMDB..BHZ | 2014-01-21T02:39:28.350000Z - 2014-03-22T14:09:39.980000Z | 100.0 Hz, 522541164 samples
TM.TMDB..BHZ | 2014-03-22T14:09:40.000000Z - 2014-03-22T14:45:39.980000Z | 100.0 Hz, 215999 samples
TM.TMDB..BHZ | 2014-03-22T14:45:40.000000Z - 2014-03-22T15:20:09.990000Z | 100.0 Hz, 207000 samples
TM.TMDB..BHZ | 2014-03-22T15:26:10.000000Z - 2014-03-22T15:57:59.990000Z | 100.0 Hz, 191000 samples
TM.TMDB..BHZ | 2014-03-22T16:01:20.000000Z - 2014-03-22T16:03:49.990000Z | 100.0 Hz, 15000 samples
TM.TMDB..BHZ | 2014-03-22T16:07:10.000000Z - 2014-04-07T04:33:19.980000Z | 100.0 Hz, 134076999 samples
TM.TMDB..BHZ | 2014-04-07T04:33:20.000000Z - 2014-12-28T01:34:19.990000Z | 100.0 Hz, 2288526000 samples
TM.TMDB..BHZ | 2014-12-28T01:34:30.000000Z - 2015-01-01T00:00:00.000000Z | 100.0 Hz, 33993001 samples
```

:-)

This is now a PR that fixes the issue - parts of ObsPy's C code and libmseed had some integer overflows.

@CTrabant This requires some changes in `libmseed`. They are part of this PR. I converted the problematic variables to `unsigned long long` to be super safe but maybe you want to use types from `stdint.h` - I'm not sure what you prefer. It might also be worthwhile to run the whole code through a code formatter - its mix of spaces, tabs, and trailing whitespaces makes it very hard to edit with some editors.

@lileipku00 This was an error and thus should naturally be fixed in ObsPy. That being said I think its really quite a bad idea to create these ginormous MiniSEED files as that makes them essentially impossible to work with. It would be much better to either organize them in an SDS structure (ObsPy can directly work with that: http://docs.obspy.org/packages/obspy.clients.filesystem.html) or (shameless plug) use the ASDF format (http://seismic-data.org).

If you have no choice: Don't directly use `obspy.read()` as it will first read the whole file to memory and then parse the MiniSEED structure - we do this to simplify everything and to be independent of the data source and in most cases the additional memory usage does not matter at all. Directly reading the whole file can readily use around 4-7 times the filesize in memory - MiniSEED can be compressed quite heavily. Best read the file in chunks (this should also work without this PR):

```python
import obspy
import io

reclen = 512
chunksize = 100000 * reclen # Around 50 MB

with io.open("./TMDB.TM.mseed", "rb") as fh:
    while True:
        with io.BytesIO() as buf:
            c = fh.read(chunksize)
            if not c:
                break
            buf.write(c)
            buf.seek(0, 0)
            st = obspy.read(buf)
        # Do something useful!
        print(st)
```

lileipku00:
Thanks @krischer. 

Maybe a silly question. But I don't know mseed very well. Your code sample illustrates that mseed file can be read in chunks. What's the condition of that? chunksize must be times of 512? If so, any large file cannot be a problem then.

lileipku00:
Moreover, another silly question. I also don't know github very well. The links below direct to http://docs.obspy.org/pull-requests/1419/ seems to be a modified document page. How do I update with the `PR`?

megies:
> How do I update with the PR?

It looks like you're using Anaconda on Linux, so..
```
conda remove obspy
pip install https://github.com/krischer/obspy/archive/mseed-overflow.zip
```
..should do the trick.

krischer:
> And just to be sure, the problem in libmseed you mentioned won't influence the correctness of the data read by obspy.read?

This PR also makes some changes to libmseed - otherwise your file cannot be read. The file is so big that 32 bit integer could no longer represent its size thus an overflow happened. No changes to reading actual records have been done so it does not change correctness.


> Maybe a silly question. But I don't know mseed very well. Your code sample illustrates that mseed file can be read in chunks. What's the condition of that? chunksize must be times of 512? If so, any large file cannot be a problem then.
> 
> And if we read chunk by chunk, is there any possibility of losing some intermediate traces (at the beginning and end of the chunks)?

MiniSEED is organized in so called records. Each record is totally self-containing and MiniSEED files are just many of these records one after another. 512 bytes happens to be the record length of the file you sent - in my experience it is also the most common record length but for other files you will have to check the record length. The code snippet above is also not fully save as MiniSEED files in theory can have records of multiple lengths. The snippet thus works by always cutting exactly at a boundary between two records.

lileipku00:
Thanks for the information and solutions.

CTrabant:
Apologies for the tangent, this is not really ObsPy related.
>> Just out of curiosity.. what's the use case for 19GB mseed files?
>> 
Not that it's too important, but the that particular file was a tar file containing 100s of miniSEED files.
> No idea. I'm also curious about how he is going to deal with such a large file.
> 
I Converted the 19GB of tar'ed miniSEED (100s of miniSEED files) to SAC using mseed2sac using the -i option to have it work on each file independently without a problem.  Something like 50 GB of SAC output.  It took ~11 minutes but totally doable.

If it were a single miniSEED file, I would have used the -ic option of mseed2sac to process each contiguous-in-the-file channel individually.

Point is we've learned to deal with such large files, using the traditional break-it-down-to-workable sizes strategy.

Now if it was 19GB all of the same channel, that would be more difficult to deal with...

obspy/io/mseed/src/libmseed/parseutils.c
```diff
@@ -54,7 +54,7 @@ msr_parse ( char *record, int recbuflen, MSRecord **ppmsr, int reclen,
     return MS_GENERROR;
   
   /* Sanity check: record length cannot be larger than buffer */
-  if ( reclen > 0 && reclen > recbuflen )
+  if ( reclen > 0 && (unsigned int) reclen > recbuflen )
     {
       ms_log (2, "ms_parse() Record length (%d) cannot be larger than buffer (%d)\n",
```
QuLogic: Does this not create a warning? `%d` is not right for `unsigned long long`.

CTrabant:
> @CTrabant <https://github.com/CTrabant> This requires some changes in libmseed. They are part of this PR. I converted the problematic variables to unsigned long long to be super safe but maybe you want to use types from stdint.h - I'm not sure what you prefer.
> 
Do not count on this making it into any 2.x release.  It is an API change to core functions and will break compatibility with a lot of software.  This is certainly something I can address in a next major release (driven by the FDSN definition of the next generation miniSEED, timeline is within the next year or so).

Keep in mind that libmseed can be used to read ginormous miniSEED files, just not in the way that ObsPy is using it in this case.
> It might also be worthwhile to run the whole code through a code formatter - its mix of spaces, tabs, and trailing whitespaces makes it very hard to edit with some editors.
> 
Don't use those editors then?  Especially if they have been setup expecting Python, in particular any editor that makes trailing white spaces hard to deal with. :P

I've long planned to change the formatting to a consistent use of spaces for indentation at the next major release.  Until that time, any major reformatting will render diffs between releases a huge mess and that's why it hasn't been done.

> If you have no choice: Don't directly use obspy.read() as it will first read the whole file to memory and then parse the MiniSEED structure - we do this to simplify everything and to be independent of the data source and in most cases the additional memory usage does not matter at all. Directly reading the whole file can readily use around 4-7 times the filesize in memory - MiniSEED can be compressed quite heavily. Best read the file in chunks (this should also work without this PR):

As a suggestion, perhaps this kind of chunked reading could be made an official option in ObsPy.  It makes sense for the reasons you gave, but the read-it-all-into-memory approach will always be a problem for scaling to large volumes and is a bit cavalier with user memory.

QuLogic:
> Do not count on this making it into any 2.x release. It is an API change to core functions and will break compatibility with a lot of software.

I'm :-1: on merging this PR so long as it breaks the ABI for libmseed because then I can't use the external system library.

krischer:
> I'm 👎 on merging this PR so long as it breaks the ABI for libmseed because then I can't use the external system library.

Yea that was also my conclusion - we don't want to maintain a libmseed fork - I just forgot to write it. It is a very niche use case and this issue contains a workaround that works with current ObsPy versions. We'll update this once libmseed has a new major version with the required changes for this to work.

krischer:
Depends on https://github.com/iris-edu/libmseed/issues/1

dr-glenn:
I don't know if it's proper to make new comments on an issue when the last comment is more than 3 years ago. But here goes. The suggested solution works for me, too. But there are caveats to this.
```
  import obspy
  import io

  reclen = 512
  chunksize = 100000 * reclen # Around 50 MB

  with io.open("./TMDB.TM.mseed", "rb") as fh:
    while True:
        with io.BytesIO() as buf:
            c = fh.read(chunksize)
            if not c:
                break
            buf.write(c)
            buf.seek(0, 0)
            st = obspy.read(buf)
        # Do something useful!
        print(st)
```
We have mseed files from a 1 year deployment that are each approx. 12 GB in size. Within are streams that are roughly 4 to 5 days long. Most streams contain one trace, but sometimes I encounter streams that contain 8 traces.
Here's the problem with the sample solution, earlier raised by lilepku00:

> And if we read chunk by chunk, is there any possibility of losing some intermediate traces (at the beginning and end of the chunks)?

Clearly the answer is YES. Consider using a chunksize of only 512 instead of 100000*512. The buffered read will still work, but clearly you will never get a full trace.

In my application, I am searching the mseed for traces that contain a specific datetime, so I use headeronly=True - it's quite fast on this 12 GB file. But once I find the trace, I have adopt a different reading strategy so that I can be sure I get all of it, even if it carries over into the next chunk. And then I want to find another event (later datetime), so again backup and continue search.

So a more elaborate solution is needed. When I work it out, is this the best place to post it?

chad-iris:
@dr-glenn You may wish to look at the tsindex module which can be used to index and efficiently access miniSEED in files regardless of organization (big, many, etc): 

https://docs.obspy.org/packages/autogen/obspy.clients.filesystem.tsindex.html

For example I've got a huge file called `bigdata.mseed`, then:

```
from obspy.clients.filesystem.tsindex import Indexer
indexer = Indexer('.', database='bigdata.sqlite', filename_pattern='bigdata.mseed')
indexer.run()

from obspy.clients.filesystem.tsindex import Client
from obspy import UTCDateTime

client = Client('bigdata.sqlite')
start = UTCDateTime ('2010-6-1T00:00:00')
event_data = client.get_waveforms("*", "*", "*", "BHZ", start, start + 120)
event_data.plot()
```

Of course you'll need to adjust paths, names, codes, and times, but the pattern will work with one huge file, 100000 files, etc.

If you need more help with this it may be better to create an issue instead of adding to this (dead) PR.

megies:
> > I'm -1 on merging this PR so long as it breaks the ABI for libmseed because then I can't use the external system library.
> 
> Yea that was also my conclusion - we don't want to maintain a libmseed fork

closing this as it looks like there was consensus on disapproval

PR #1413, opened 2016-05-25
Import in in obspy.clients.filesystem.sds

megies:
Hi
What do you think of an "import files to SDS" like  
https://github.com/eost/clb-noise-analysis/blob/master/extract2sds.py


bonaime:
Could you add "details" option in add_data_to_archive to keep details information in the SDS archive ?

megies:
Not sure what you mean with "details".. if it's MSEED record details, that would need a different approach..

megies:
>> Could you add "details" option in add_data_to_archive to keep details information in the SDS archive ?

> Not sure what you mean with "details".. if it's MSEED record details, that would need a different approach..

@bonaime Ah, now I see what you meant.. `details=True` in `read()` https://github.com/eost/clb-noise-analysis/blob/master/extract2sds.py#L89
Hmm.. I don't think it makes a big difference when reading from and then writing to mseed whether it's w/ vs. w/o `details=True` option.. or does it? If you really wanted to keep mseed details you would have to do it record by record, I think?

bonaime:
@megies the idea is to keep, at least, the clock quality parameters in the SDS archive. I foreseen to use SDS client to create a SDS archive from files from different field tests. We can  have the same details=False in  add_data_to_archive than in read()

megies:
> clock quality parameters

Those are record-based, right? I can't imaginge they really will be preserved correctly just by doing a `read(.., details=True).write(.., format="MSEED")`, or are they?

bonaime:
@megies I think they are

megies:
Well.. reading from miniseed, a trace has a `timing_quality` in `trace.stats.mseed.blkt1001`, which is a single value. Just guessing, but that's probably the value of timing quality of first record of the whole trace..

megies:
Ok, I've downloaded those data you used in the example and had a look myself. So with `details=True` a new trace is generated whenever the timing quality changes, otherwise the (gapless) file gets read into a single trace. So it really makes a difference (although it might lead to huge amount of trace objects, if the timing quality changes on a shorter time scale..).

krischer:
Hmm....I think the `details=True` variant is okay but far from optimal for two reasons:

* There is no reason to split it into several files just because the timing quality changes - they can well be in the same file.
* There are lots of other flags and blockettes that are not preserved.

I think this should really happen at the record level but this would be a bit more work to implement :-( There is some record level splitting stuff here if it helps: https://github.com/obspy/obspy/blob/105bcdc4a0fccc8190f503758e86c22b3ece84f0/obspy/clients/fdsn/mass_downloader/utils.py#L80-L258

megies:
> There is no reason to split it into several files just because the timing quality changes - they can well be in the same file.

The separate traces by `details=True` end up in the same file alright.. otherwise it wouldn't be SDS structure anymore.. ;-)

> I think this should really happen at the record level but this would be a bit more work to implement :-(

Sure thing.. but for now better this than nothing..

krischer:
> The separate traces by details=True end up in the same file alright.. otherwise it wouldn't be SDS structure anymore.. ;-)

Ah ok...then I guess I don't really understand what `details=True` is doing for the SDS importer ;-)



> > I think this should really happen at the record level but this would be a bit more work to implement :-(
> 
> Sure thing.. but for now better this than nothing..

Definitely.

krischer:
> Ah ok...then I guess I don't really understand what details=True is doing for the SDS importer ;-)

Nvm. I got it.

megies:
Rebased and force-pushed since some stuff was already extracted and merged into master (the refactoring of obspy-scan internals)

PR #1392, opened 2016-05-27
Pn and PmP's ray paths

krischer:
I am using ObsPy 1.0.1 on CentOS 7, installed from anaconda. I was trying to get the travel time and ray path of the Pn wave and found something confusing.  It seems when I use “phase_list = ['Pn']" I got a ray path of PmP, and vice versa.
pn_pmp

```python

model = TauPyModel(model='iasp91')
arrivals = model.get_ray_paths(source_depth_in_km=0, distance_in_degree=1, phase_list=['Pn','PmP'])
arrivals.plot(plot_type='cartesian')

```


krischer:
I turned this issue into a PR containing a port of @crotwell's fix as well as a regression test.

krischer:
We could additionally "fix" paths that are slightly off by multiplying the distances with a small correction factor. Do we want this? I currently implemented it in the plotting part of `obspy.taup` so it does not affect the actually returned paths but just the plots. It looks nicer as all rays now arrive exactly at the receiver but it is of course cheating so I wanted to hear some opinions first.

figure_1-1

megies:
> PmP is very similar to Pn at this distance as it turns so shallow in mantle that you can't see the difference with Pn.

PmP should not enter the mantle but reflect off the top of Moho, so the path as it is right now doesn't look right to me.. or am I missing something?

krischer:
`m` with the TauP phase naming convention is not a reflection but just an interaction with the Moho.

ThomasLecocq:
> m with the TauP phase naming convention is not a reflection but just an interaction with the Moho.

very confusing indeed, so PvmP is what we think of the regular PmP, right?

megies:
> very confusing indeed, so PvmP is what we think of the regular PmP, right?

I also think that's very confusing.. what's the reasoning behind having another phase description logic in parallel to the official one? (I guess it must be needed by Taup internally somehow?)

crotwell:
See chapter 4 of the TauP manual for the phase naming convention. It was chosen to be internally consistent and allowing arbitrary phases while matching more tradition phase naming where we could. PmP was one of the places where we had to choose the phase parser handling arbitrary phases over maintaining traditional nomenclature.

krischer:
There still are some commits in this PR :-) I just rebased  - should be good to merge if CI passes.

krischer:
> We could additionally "fix" paths that are slightly off by multiplying the distances with a small correction factor. Do we want this? I currently implemented it in the plotting part of obspy.taup so it does not affect the actually returned paths but just the plots. It looks nicer as all rays now arrive exactly at the receiver but it is of course cheating so I wanted to hear some opinions first.


Can someone also comment on this?

blazing216:
> We could additionally "fix" paths that are slightly off by multiplying the distances with a small correction factor. Do we want this? I currently implemented it in the plotting part of obspy.taup so it does not affect the actually returned paths but just the plots. It looks nicer as all rays now arrive exactly at the receiver but it is of course cheating so I wanted to hear some opinions first.
> 
> Can someone also comment on this?
I would prefer the ray paths returned, plotted and the corresponding travel times stay consistent. Otherwise, someone may take an less accurate traveltime or ray path as an accurate one, instead of taking more iterations to improving it.

krischer:
> I would prefer the ray paths returned, plotted and the corresponding travel times stay consistent. Otherwise, someone may take an less accurate traveltime or ray path as an accurate one, instead of taking more iterations to improving it.

That is a very good argument and I completely agree. I'm merging this for now as it currently is. If somebody want to keep discussing "fixing" the plots, please open a new issue.

QuLogic:
You didn't actually merge it...

krischer:
Hmm - I'd like to have a good explanation for this but I don't ;-) Thanks for taking care of it!

PR #1424, opened 2016-05-27
Possible memory leak in remove_response() (numpy FFT cache growing)

krischer:
I'm trying to use ObsPy to remove the instrument response from a whole year of data of a station, so I wrote a python script that loops over the miniseed files and apply the relevant functions. What I found out is that after a while the script gets killed by the kernel's out of memory killer. Indeed the memory it uses grows indefinitely during the iterations.

I wrote a minimal example that is able to reproduce this issue, you'll be able to see the memory growing using `top` or similar tools. 

```python
#!/usr/bin/python

from __future__ import print_function
import sys

import obspy
import obspy.clients.fdsn

for i in range(1,365):

    t = obspy.UTCDateTime(year=2014, julday=i)

    client = obspy.clients.fdsn.Client('IPGP')
    st = client.get_waveforms(network="G", station="CCD", location="00", channel="BHZ", starttime=t, endtime=(t + 86400), attach_response=True)

    tr = st[0]

    tr.remove_response()

    print('.', end="")
    sys.stdout.flush()
```

I'm using python-obspy 1.0.1 from the `deb.obspy.org` Debian repository on an up-to-date Debian "sid" system:

```
~ $ uname -a
Linux mandragola 4.5.0-2-amd64 #1 SMP Debian 4.5.4-1 (2016-05-16) x86_64 GNU/Linux
~ $ python --version
Python 2.7.11+
```

Unfortunately I can't try the python3 module, as it doesn't work with Python 3.5 (lacks `libmseed_Linux_64bit_py35.cpython-34m.so`)

I'm willing to help in solving this issue, so if you need any other information just let me know.

Thank you!

### DIFF
--- a/obspy/core/util/misc.py
+++ b/obspy/core/util/misc.py
@@ -520,7 +520,7 @@ def limit_numpy_fft_cache(max_size_in_mb_per_cache=100):
         try:
             total_size = sum([_j.nbytes for _i in cache.values() for _j in _i])
         except:
-            pass
+            continue
         if total_size > max_size_in_mb_per_cache * 1024 * 1024:
             cache.clear()
 
--- a/obspy/core/util/misc.py
+++ b/obspy/core/util/misc.py
@@ -494,6 +494,37 @@ class MatplotlibBackend(object):
                 raise Exception(msg)
 
 
+def limit_numpy_fft_cache(max_size_in_mb_per_cache=100):
+    """
+    NumPy's FFT implementation utilizes caches to speedup subsequent FFTs of
+    the same size. This accumulates memory when run for various length FFTs
+    as can readily happen in seismology.
+
+    This utility function clears both, full and real-only caches if their
+    size is above the given threshold.
+
+    The default 100 MB is fairly generous but we still want to profit from
+    the cache where applicable.
+    """
+    for cache in ["_fft_cache", "_real_fft_cache"]:
+        # Guard against different numpy versions just to be safe.
+        if not hasattr(np.fft.fftpack, cache):
+            continue
+        cache = getattr(np.fft.fftpack, cache)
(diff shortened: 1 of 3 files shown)


krischer:
Found it :-) I turned the issue into a PR which should solve the issue.

It is actually not our fault - numpy's FFT implementation caches some kind of (I guess) execution plan per FFT size. That cache is never cleared so doing lots of FFTs with different lengths results in that cache growing without bounds.

This PR works around that issue by "periodically" clearing the cache. It is currently only implemented for the `Trace.remove_response()` method but we might want to do this throughout all of ObsPy.

I'll also take this issue to the numpy guys.

obspy/core/util/misc.py
```diff
@@ -494,6 +494,37 @@ def switch_backend(backend, sloppy=True):
...
+        if type(cache) is not dict:
+            continue
+        # Its a dictionary with list's of arrays as the values. Wrap in
+        # try/except to guard against future numpy changes.
+        try:
+            total_size = sum([_j.nbytes for _i in cache.values() for _j in _i])
+        except:
+            pass
+        if total_size > max_size_in_mb_per_cache * 1024 * 1024:
+            cache.clear()
```
megies: You probably wanted to put this in a `else:` branch of the try/except..?

krischer: I actually intended to have a `continue` there and not a `pass`. At this point it should be pretty safe to call clear as I asserted its a dictionary and that the size can be calculated.

megies: Yeah, a `continue` is effectively the same.

krischer:
numpy/numpy/pull/7686 has been merged and the numpy master now contains a more sophisticated version of this PR. We still need this for backwards compatibility reasons. Ready to merge IMHO.

PR #1426, opened 2016-05-28
Event-type objects: warn when setting non-default attribute

megies:
Follow-up to #1418. Make event-type objects issue a warning when a non-default attribute is set. The main purpose for this PR is that it has happened several times before that users had a typo when setting a field (or just using a wrong field name) and then surprised that the information was lost on output to QuakeML.


megies:
Wrong base branch :worried:..

PR #1427, opened 2016-05-28
Event-type objects: warn when setting non-default attribute

megies:
Follow-up to #1418. Make event-type objects issue a warning when a non-default attribute is set. The main purpose for this PR is that it has happened several times before that users had a typo when setting a field (or just using a wrong field name) and then surprised that the information was lost on output to QuakeML.

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -200,6 +200,9 @@ def _event_type_class_factory(class_name, class_attributes=[],
         for key, value in _properties:
             _property_dict[key] = value
         _containers = class_contains
+        warn_on_non_default_key = True
+        defaults = dict.fromkeys(class_contains, [])
+        defaults.update(dict.fromkeys(_property_keys, None))
 
         def __init__(self, *args, **kwargs):
             # Make sure the args work as expected. Therefore any specified
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -298,7 +298,7 @@ def _event_type_class_factory(class_name, class_attributes=[],
             if containers:
                 # Print delimiter only if there are attributes.
                 if attributes:
-                    ret_str += '\n\t---------'
+                    ret_str += '\n\t' + '---------'.rjust(max_length + 5)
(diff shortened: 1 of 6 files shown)


megies:
This makes `Event.preferred_origin_id` (and consorts) from being (silently) expected to be `str` to be of `ResourceIdentifier` type, being in line with all other occurrences of identifiers in event type objects. I think this change makes sense and catching the initialization of a `ResourceIdentifier` with a `ResourceIdentifier` should catch any side effects in user code, I think.
A quick review or opinions would be good though..

megies:
Ideally `_event_type_class_factory()` should also set up more useful `__init__()` methods for all classes, based on their `class_attributes`, such that..
```
Init signature: Magnitude(self, force_resource_id=True, *args, **kwargs)
```
..becomes..
```
Init signature: Magnitude(self, force_resource_id=True, resource_id=None, mag=None,
 magnitude_type=None, origin_id=None, method_id=None, station_count=None,
 azimuthal_gap=None, evaluation_mode=None, evaluation_status=None, creation_info=None)
```
I think that would further help to avoid mistakes caused by typos..

obspy/core/event/base.py
```diff
@@ -557,6 +561,9 @@ def __init__(self, id=None, prefix="smi:local",
             self.fixed = False
             self._prefix = prefix
             self._uuid = str(uuid4())
+        elif isinstance(id, ResourceIdentifier):
+            self.__dict__ = id.__dict__
```
krischer: maybe copy this or rather `self.__dict__.update(id.__dict__)` - this might result in strange and hard to track errors otherwise.

krischer:
Looks mostly good to me and its a very sensible change :-)

megies:
> This makes Event.preferred_origin_id (and consorts) from being (silently) expected to be str to be of ResourceIdentifier type, being in line with all other occurrences of identifiers in event type objects. I think this change makes sense and catching the initialization of a ResourceIdentifier with a ResourceIdentifier should catch any side effects in user code, I think.

If there's no other comments to this point, I think this is good to merge.

Issue #1398, opened 2016-05-09
Reading Mini Seed files


bonaime:
ObsPy version 10.0.1, MacOS 10.11.4,Python 2.7.11 Anaconda 2.3.0

Hi,

I want to read multiple mini-seed files recorded on a Q330S. This digitiser can group together in a single file multiple miniseed files. With Obspy I can read the headers without problems:
``` code
a = read('http://www.ipgp.fr/~bonaime/FR-4883',headonly=True)
print(a.__str__(extended=True))
37 Trace(s) in Stream:
FR.4883..LOG | 2013-02-12T09:08:08.880000Z - 2013-02-12T09:08:08.880000Z | 0.0 s, 3921 samples
FR.4883..LOG | 2013-02-12T09:08:08.890000Z - 2013-02-12T09:08:08.890000Z | 0.0 s, 3733 samples
FR.4883..LOG | 2013-02-12T09:08:09.210000Z - 2013-02-12T09:08:09.210000Z | 0.0 s, 3719 samples
FR.4883..LOG | 2013-02-12T09:08:09.590000Z - 2013-02-12T09:08:09.590000Z | 0.0 s, 3644 samples
FR.4883..LOG | 2013-02-12T09:08:09.950000Z - 2013-02-12T09:08:09.950000Z | 0.0 s, 4040 samples
FR.4883..LOG | 2013-02-12T09:08:10.310000Z - 2013-02-12T09:08:10.310000Z | 0.0 s, 3660 samples
FR.4883..LOG | 2013-02-12T09:08:10.630000Z - 2013-02-12T09:08:10.630000Z | 0.0 s, 3711 samples
FR.4883..LOG | 2013-02-12T09:08:10.970000Z - 2013-02-12T09:08:10.970000Z | 0.0 s, 3898 samples
FR.4883..LOG | 2013-02-12T09:08:11.360000Z - 2013-02-12T09:08:11.360000Z | 0.0 s, 3764 samples
FR.4883..LOG | 2013-02-12T09:08:11.740000Z - 2013-02-12T09:08:11.740000Z | 0.0 s, 3977 samples
FR.4883..LOG | 2013-02-12T09:08:12.050000Z - 2013-02-12T09:08:12.050000Z | 0.0 s, 3662 samples
FR.4883..LOG | 2013-02-12T09:08:12.500000Z - 2013-02-12T09:08:12.500000Z | 0.0 s, 3911 samples
[...]

PR #1431, opened 2016-05-31
SDS client: use "sourcename" option during "read()" (for MSEED format data)

megies:
Just a minor improvement, in case of non-expected data stored in files. During reading data, only look at records with seed pattern of interest using "sourcename" option in _read_mseed().


bonaime:
could you also add "sourcename" option in add_data_to_archive to  avoid #1398 ?

megies:
> could you also add "sourcename" option in add_data_to_archive to avoid #1398 ?

That's not concerning this PR but rather #1413.

PR #1433, opened 2016-05-31
Read REFTEK130 data

megies:
Support for reading Reftek 130 data. Work in progress..


megies:
Current version works for STEIM1 encoded data for a file starting/ending with EH/ET (event header/trailer) packet and containing DT (data) packets in between.
(But slow, as single data packets are decoded one by one from a Python loop)

jkmacc-LANL:
:+1:  for this, Tobias!

megies:
There's some implementations in PASSCAL's `passoft` suite I was considering to reuse.. but maybe it's easiest to just do it from scratch, mostly in Python and data unpacking (from STEIM) with libmseed..

For now this is based on a quick prototype.. it's pretty slow right now, we'll see how much work we want to spend on speedup, I guess.

jkmacc-LANL:
[Very little?](http://docs.cython.org/src/reference/compilation.html) :smile:

markcwill:
@megies original library source code from RefTek is publicly viewable here https://github.com/antelopeusersgroup/antelope_contrib/tree/master/nobuild/bin/rt/rtp2orb/RefTek but I'm unsure of the licensing... NSL forked this some years ago and made only minor changes, our version isn't on GitHub b/c I don't know what license the original source is under.

megies:
> unsure of the licensing

Yeah, I asked trimble support whether `rt_mseed` source code could be open source and that was a `"No, but..."` (didn't even dare to ask about LGPL). So I think it's probably not worth to do more investigations along the "reuse" option..

I did get permission to add the data format description pdf to the repo, though. I'll add it for now, we can still rebase it out of the PR if we decide to not keep it (it's ~1MB).

megies:
@krischer, there's a problem with cab834c, I think. In principle it the reading works fine and tests OK against the reftek data converted to miniSEED using `rt_mseed` (from RefTek utilities) but I'm getting some segmentation faults on my computer (e.g. in Ipython when reading multiple times and doing something in between) and [Appveyor also has some strange interruption of the test runner](https://ci.appveyor.com/project/obspy/obspy/build/1.0.2290-master#L643).

megies:
Side note.. right now this is roughly 60x slower than `ref2sac` (pure C) from PASSCAL's `passoft` package (converting 24 1-h files to SAC) and roughly 12-15x slower than `rt_mseed` from RefTek utilities (converting the same 24 1-h files to MSEED).

And.. passoft's `ref2sac` is not checking consistency between individual packets (by starttime and npts), so basically ending up with wrong output data when run on files with gaps or overlaps.

So this is probably not that bad to start things..

markcwill:
@megies since our c code only does data packets, we have been working on a package for the other packets necessary to build a data stream (DS, SC, etc). It was originally used for logging (SH) packets, but it's turned into something very similar to this PR... I'll try and OS it soon, might have to replace some test data... here's the packet file if you want to take a look:

https://gist.github.com/markcwill/430ac021980aa0affcc719c3cf1be167

I think we took the same basic approach, except our use case is pulling packets from a real-time stream, and we wanted to parse to something JSON-encodable, so the class "constructors" all return dict, list, str, etc, but not hard to modify.

megies:
@krischer I just wanted to give this another shot debugging the segfault that's blocking this PR, and when checking against the impending update to libmseed 2.18 I checked against #1540.. and had to find out that the routine ’msr_unpack_steim1’ seems to have been scrapped..

Any idea what the right way to go forward with this is?

krischer:
Ah yea - the steim unpacking things are not part of the official libmseed API - but it being in C means we can of course still access it. But as they are not part of the official libmseed API we also removed the ObsPy wrapper.

@mbyt: I guess we have to restore the unpacking wrappers.

markcwill:
We have internal clean room implementations (decode/unpack ONLY, for now) of steim in pure python (just steim 1) and also golang (1 & 2). If the steim unpacking is the only thing holding this PR up, I'd be happy to release/donate it to ObsPy. The python is slower, but completely functional AFAIK.

mbyt:
@markcwill thank you very much for the kind offer!

As we ship libmseed anyway, I would personally prefer to use the in-official libmseed API. My personal opinion is that this way we can support this feature without new code.

> @mbyt: I guess we have to restore the unpacking wrappers.

@krischer I was unfortunately not aware of this issue here, when pro-actively removing the wrappers :\.

QuLogic:
Not everywhere uses bundled libmseed; I don't think we should use internal functions without getting upstream's approval.

chad-iris:
Using the libmseed functionality that is already included makes the most sense to me.  If you can use the internal msr_decode_steim[12] without too much hassle I'd say go for it.  While they are internal I have no reason to think they will change soon, and it will likely be a really obvious failure if they do change in the future.

krischer:
> Not everywhere uses bundled libmseed; I don't think we should use internal functions without getting upstream's approval.

That should still be fine as ObsPy is usually always tied to specific `libmseed` version in any case. The only platform where I forsee any issues is windows, as the exported functions must be included in the `.def` file. But we can just include the steim unpacking functions in our own `.def` file and everything should work fine.

Am I missing an obvious problem?

krischer:
The unpacking routines are now again part of #1540 so feel free to use them.

megies:
Rebased on current master with updated libmseed, hope I got the conflicts right (if not.. the old tip was at a6a5fe5a850f6888e509280369e52f1e765d248f before the rebase).

megies:
I rewinded this PR a bit to calling libmseed's steim decompression routine for each data packet individually (one packet has several hundred samples) rather than gathering all compressed data packets' payloads for one single big call to libmseed. The latter was producing segfaults on subsequent read calls to reftek/mseed files and it seems that the overhead of calling libmseed more often is negligible so that it's not worth debugging..

Some numbers for a 1 hour long 200 Hz reftek file with 3 channels (i.e. 3 * 720,000 samples = 2.1 million samples)..
 - previous (segfault prone) combined calls to unpack steim1 with libmseed: 2.53 seconds (via `$ python -m timeit -n 4 -r 4 ...`)
 - now, separate calls to unpack steim1 with libmseed for each data packet: 2.57 seconds (again via `$ python -m timeit -n 4 -r 4 ...`)
 - letting `rt_mseed` binary (from reftek utilities) convert the file from reftek to miniseed output file: ~0.35 seconds (via `$ time rt_mseed ...`))
 - reading the converted miniseed file with `_read_mseed` (again via `$ python -m timeit -n 4 -r 4 ...`): 19 ms   ;-)

I'd probably merge this if CI gives the nod. Better to have a not-super-fast version in 1.1.0 than no version at all. If anybody wants to speed it up it can be done in maintenance later on.

megies:
> Not everywhere uses bundled libmseed; I don't think we should use internal functions without getting upstream's approval.

@QuLogic just to clarify.. in your packaging efforts of obspy, are you replacing our vendored libmseed with a separately packaged one?

krischer:
> previous (segfault prone) combined calls to unpack steim1 with libmseed: 2.53 seconds (via $ python -m timeit -n 4 -r 4 ...)
>     now, separate calls to unpack steim1 with libmseed for each data packet: 2.57 seconds (again via $ python -m timeit -n 4 -r 4 ...)
>     letting rt_mseed binary (from reftek utilities) convert the file from reftek to miniseed output file: ~0.35 seconds (via $ time rt_mseed ...))
>     reading the converted miniseed file with _read_mseed (again via $ python -m timeit -n 4 -r 4 ...): 19 ms ;-)

This is very slow to be honest - there must be some way to speed it up. Its an order of magnitude slower then then the `rt_mseed` binary... I'll try to find some time the next days to look into it. Maybe @mbyt also finds this interesting? Can you upload a bigish test file somewhere?

QuLogic:
> @QuLogic just to clarify.. in your packaging efforts of obspy, are you replacing our vendored libmseed with a separately packaged one?

Yes, Fedora packages use a separate libmseed and all the individual codes that IRIS provides are also using it.

Fortunately, all symbols are visible by default on Linux, so we don't yet have to worry about it.

megies:
> Fortunately, all symbols are visible by default on Linux, so we don't yet have to worry about it.

The steim unpacking routines changed with libmseed 2.18, though, so it won't work for older packaged libmseed versions..

megies:
> This is very slow to be honest - there must be some way to speed it up. Its an order of magnitude slower then then the rt_mseed binary

Well that's pure C I guess, so I'm not very surprised. Right now this PR is going through the packets in the file in the Python layer and only doing the data unpacking in C..

megies:
By the way.. @markcwill, @jkmacc-LANL do you think naming this read plugin `REFTEK130` is appropriate? Should it rather be `PASSCAL`? Do other data sources/digitizers/whatnot also use this format?

I only know this format from our RT130s, I'm a bit confused by this statement in the RT130 format description pdf (included in this PR):

```
1 About This Manual

This document contains technical information regarding the recording format
for the Program for Array Seismic Studies of the Continental Lithosphere
(PASSCAL) recording format. It includes a separate section for each of the
following topics:

    o  Information storage
    o  Notation conventions
    o  Packet Header description
    o  Packet descriptions
    o  Special packets

The PASSCAL recording format was designed to store data acquired by field
instruments manufactured to the PASSCAL specification. In general, the
PASSCAL specification requires that each recording stand alone with respect
to the sample rate, start time, and data format. Furthermore, it was a goal
of the specification that additional types of recordings could be added to the
format specification without having to reformat previous recordings.

The REF TEK 130 series DAS units record data in this format. DAS units
record data into 1,024-byte packets and store them in RAM. Without
altering the format, the DAS units write the packets to disk and/or transmit
them over a telemetry link.
```

jkmacc-LANL:
@megies PASSCAL is associated with [other formats too](https://www.passcal.nmt.edu/content/ph5-what-it), and most users access rt130 files through a number of command line `rt130*` utilities, so I think naming the plugin `rt130` or `reftek130` will be most intuitive.

markcwill:
@megies I agree with @jkmacc-LANL -- I think "rt130" or "reftek" or something similar is appropriate. AFAIK RefTek is the only one using this format, I think it's referred to "passcal" because they helped define the spec and some of the software tools, but it would be confusing to use "passcal" unless Trimble has a problem with using the "reftek" name. I've always seen it as "130 Recording format" or something similar. More importantly for us, it's the same format used in the RT130 real-time RTP data stream, so we call it "reftek" when we deal with it.

markcwill:
@krischer it does seem a bit slow, maybe I or someone could run cProfile and see if anything jumps out? Agree with @megies it's best to get it out if it's functionally correct.

(I understand if we want to move perf/optimization to another issue/PR, but I'll put this here for now):
Just for comparison, here's benchmarks from our pure python and golang libs, reading one event file and decoding the steim1 samples, hourlong, 100Hz:
```shell
refutil - dt_bench.py (CPython 2.7)
===
Read 3030 packets, 2700000 samples
timeit: 3 loops, best of 3: 901 ms per loop
cProfile: 1484ms, 70% in steim decoding lib (pure python)

go-reftek - bench_event_test.go (Go 1.6.2)
===
PASS
BenchmarkEventFile-4          10         170200857 ns/op        56240268 B/op    1625658 allocs/op
```

Haven't tried it with pypy yet, but it should be somewhere in the middle. With the C unpacking routines, I would expect this PR to be somewhere between these 2 extremes: 150ms - 1.5s?

megies:
Thanks for the feedback on the naming @jkmacc-LANL and @markcwill. That's backing my thoughts. There shouldn't be a problem with the naming, I think, considering that I've got the permission to include the format spec pdf in obspy under LGPL and that it's advertisement and win-win for obspy to be able to read those files out-of-the-box.

note to self.. those adjustments for the steim unpacking are currently unused.. https://github.com/obspy/obspy/pull/1433/files#diff-3a6dfa58d4c06d1bb896a7c5b65dddc1R847

mbyt:
Hi @megies, hope you are doing fine!

> Biggest ones I have are files with 1 hour, 3 channels at 200 Hz.

Is it a big hassle to put that somewhere on the web? Would be great, as I would like to play a little bit with profiling / optimizations.

megies:
> Is it a big hassle to put that somewhere on the web? Would be great, as I would like to play a little bit with profiling / optimizations.

No big deal, I've [added a 1-hour/3-channel/200Hz test file to our examples repo](https://github.com/obspy/examples/commit/4f98871ae360d1184eb7ae0d23ef0fa3ac401037), should be up at http://examples.obspy.org/ in a bit..

megies:
I've did some more fine tuning and added some more tests for better coverage.. pretty much ready from my point, I think. We can do some speeding up, optionally, but it should be OK for inclusion in 1.1.0.

krischer:
I did a bit of benchmarking last week - its not the data unpacking that is so slow, but the header unpacking. Half the time for datetime parsing and the other half for unpacking the other values. That's mainly because its loops of loops of function calls per type. That eats almost all the runtime.

I made experiments by just using numpy's structured arrays to unpack everything in one go but this is not ready yet. It does seem to work though.

megies:
Hmm.. yeah, the header unpacking is pretty awkward because I'm not very experienced with the bitwise unpacking that's necessary for some fields.

mbyt:
> Edit:
> http://examples.obspy.org/000000000_0036EE80.mseed
> http://examples.obspy.org/000000000_0036EE80.rt130

Thanks a lot!

> Hmm.. yeah, the header unpacking is pretty awkward because I'm not very
> experienced with the bitwise unpacking that's necessary for some fields.

see af769a2, that could simplify it a bit.

megies:
> see af769a2, that could simplify it a bit.

Great, thanks! Slick approach..
Cuts time in half for me:
`python -m timeit -n 4 -r 4 'from obspy.io.reftek.core import _read_reftek130; _read_reftek130("000000000_0036EE80.rt130")'`
 - before: 4 loops, best of 4: 1.91 sec per loop
 - now: 4 loops, best of 4: 950 msec per loop

There's likely more potential for speedup in the python parts, wanted to just get it working (and tested) somehow first.

krischer:
Cool :-) Then the remaining time is almost completely eaten by parsing the times to UTCDateTime objects.

I'm not entirely sure how to avoid that. Maybe collecting everything in a list and passing it to the numpy datetime handling in one go - this would convert everything to timestamps which should be enough. UTCDateTime's are really only needed for the final traces. https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html

megies:
> this would convert everything to timestamps which should be enough. UTCDateTime's are really only needed for the final traces. https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html

Nice idea.. but hmm.. 
 - `New in version 1.7.0.` (we currently support numpy 1.6.1 -- although until our 1.1 release old distros we targeted for 1.0 might have gone EOL)
 - `Note: The datetime API is experimental in 1.7.0, and may undergo changes in future versions of NumPy.` (sounds a bit frightening, but if we got it tested well, we can still adapt later on..)

megies:
Hmm.. I might be missing something but to me it seems numpy can't handle Julian date strings.. guess that would be a show-stopper for that approach.

krischer:
> Hmm.. I might be missing something but to me it seems numpy can't handle Julian date strings.. guess that would be a show-stopper for that approach.

Ah I see. I think you are right :-( Maybe we have to write a short C snippet to convert the reftek dates to timestamps? Should be quite simple as no time zones are involved. I googled a bit and could not find an easily embeddable `strptime` variant.

megies:
> Maybe we have to write a short C snippet to convert the reftek dates to timestamps?

I don't think that would be a simple task.. would include all the leap year, leap second calculations and whatnot, no?

markcwill:
> Hmm.. on second thought, it might be possible to do it with a combination of np.datetime64 (for the year part) and np.timedelta64 (for the julian date, hour etc.).

That was our golang solution. Parse time from string and then add year and ordinal day minus 1 as a delta. Haven't tried that with the new numpy types, but it should work if units are done correctly and leap years are in there? Not sure of any lib that supports leap seconds...

krischer:
leap seconds don't matter for absolute times - only for time differences. `libmseed` actually honors leap seconds and we should potentially teach `UTCDateTime` to do the same.

megies:
I did some quick and dirty trial, storing timestamps with each packet after parsing the time with UTCDateTime. That way some comparisons afterwards are simple float operations instead of more UTCDateTime stuff. Saves 30% of remaining runtime.. but we'd need to double check those float comparisons, seems like (resulting stream has more (seamless) traces).

krischer:
> I did some quick and dirty trial, storing timestamps with each packet after parsing the time with UTCDateTime. That way some comparisons afterwards are simple float operations instead of more UTCDateTime stuff. Saves 30% of remaining runtime.. but we'd need to double check those float comparisons, seems like (resulting stream has more (seamless) traces).

But its always a floating point comparision, even with the objects, right? I guess I don't really understand the reftek code well enough.

In any case: in the profiling I did last week the slow part was actually parsing all the times to UTCDateTime objects.

megies:
> But its always a floating point comparision, even with the objects, right? I guess I don't really understand the reftek code well enough.

There's some comparisons on "packet.time + seconds == ..." and if packet time is a UTCDateTime than there's more of that getting initialized due to the arithmetics..

mbyt:
> python -m timeit -n 4 -r 4 'from obspy.io.reftek.core import _read_reftek130; _read_reftek130("000000000_0036EE80.rt130")'

Thanks for sharing! Makes more sense than my obspy.core.read benchmark script.

obspy/io/reftek/core.py
```diff
@@ -237,7 +237,7 @@ def from_string(string):
             msg = "Ignoring incomplete packet."
             warnings.warn(msg)
             return None
-        bcd = codecs.encode(string[:16], "hex").decode("ASCII")
+        bcd = codecs.encode(string[:16], "hex_codec").decode("ASCII")
```
mbyt: uuups. Thanks!

Didn't even know that this was missing in py33. Astonishing, as it is part of py34 and py27.

megies:
@mbyt, I think we need to drop the last commit again. I checked that one too, speedup is negligible, and it would change a cornerpiece of the UTCDateTime API. I wanted to change that into a staticmethod a long time ago, but I we'd need to make a proper deprecation..

obspy/io/reftek/core.py
megies: Didn't know that it at all.. just did a search when I saw that CI fail.

mbyt:
I did that just for more clean code. You are sure that this makes a difference?

I have no problem with dropping it. Another possibility would be to use this as an trigger and add an deprecation warning as you suggested now.

Anyway no worries, I am good with either way...

megies:
> You are sure that this makes a difference?

Oh, wow. Just tried, and it really doesn't break things.. interesting. Should definitively stay in then, but this branch might be the wrong place, I think we should do it on maintenance in principle. I bet there's a hundred places were the `()` can be dropped.

megies:
@mbyt, I've felt free to move the `strptime` code cleanup to a separate issue. It's touching several places in obspy and I think it's better kept separate. See #1576.

megies:
Some more speedups (all times for one file with 1-hour of 3-channels at 200Hz; starting at **900ms**):
 - store time information of packet as POSIX timestamp, avoids some UTCDateTime operations when comparing start/endtimes (-> **450ms**)
 - unpack packet header time information that is coming as a string for all packets in one go with numpy (-> **280ms**)
 - cache calendar calculation of POSIX timestamp of start of year (-> **270ms**)

mbyt:
> @mbyt, I've felt free to move the strptime code cleanup to a separate issue.

Thank you very much!

And, I am quite impressed by the speedup 970ms -> 270ms. That's awesome!

Let's keep the ball rolling and get this merged. Any remaining issues can be addressed in independent pull requests. Any objection?

megies:
> Let's keep the ball rolling and get this merged. Any remaining issues can be addressed in independent pull requests. Any objection?

@krischer was testing another approach to read everything into a structured array (with one item per packet). If the bit-level fields of the first 16 byte header of each 1024 byte packet could be realized with that approach somehow that would most likely be much faster.
Not sure if it's easily doable though.. otherwise we could just go with the current state, I guess. @krischer?

krischer:
do you think the first 16 bytes would be enough? Because that is very easy as they are the same for every record. I was having some trouble getting the other bytes to work.

megies:
To be more precise.. for the first and last packet we would want to also extract the payload (1008 bytes), but we could do it slow and dirty in Python. But, we would also need the first 8 bytes from the data packets' payload: https://github.com/obspy/obspy/pull/1433/files#diff-8ba81a0b0e967e99213b730fabaeb9b5R65

But! Just noticed that the first 8 bytes of EH/ET (event header/trailer) and DT (data) packets are essentially the same! The EH/ET packets have a "reserved" field in there, most likely so that it matches with the first 8 bytes of DT (were there is channel number and npts in there)

Compare:
https://github.com/obspy/obspy/pull/1433/files#diff-8ba81a0b0e967e99213b730fabaeb9b5R28
https://github.com/obspy/obspy/pull/1433/files#diff-8ba81a0b0e967e99213b730fabaeb9b5R65

markcwill:
@megies I have seen it called an "extended header", it's the first 24? bytes I think. Contains the evnt #, data stream #, etc for EH, ET, DT

megies:
> @megies I have seen it called an "extended header", it's the first 24? bytes I think. Contains the evnt #, data stream #, etc for EH, ET, DT

Yes, saw now too that the manual points that point out as well and calls it "extended" header.

@krischer, can you post the structured array handling you got so far as a gist somewhere?

megies:
Here's a gist with @krischer's approach of using numpy fromfile to do the initial reading: https://gist.github.com/904d95270548af78135af3a9f7c0c23e

Not sure if we want to put more work into this..

megies:
I did a new implementation based on @krischer's gist using numpy record arrays all the way through. Much cleaner I think and CPU time is cut down to 45% of the old implementation (still 10 times slower than reading the equivalent mseed file).

I left it in parallel right now (everything in `core2.py`), so people can have a look, but in the end we'll of course want to drop one of the implementations (the previous one most likely).

krischer:
I made another change that is again 30% faster overall. It is a bit hackish but if the tests pass everywhere it should be safe enough.

The data unpacking now takes 50% of the total runtime. The only way to make it faster now (I think) is to do it in C. This would speed up the data unpacking part about 3 times)

The other 50% are spent on header unpacking. A significant portion of that is the hex decoding as well as the time parsing.

I think the module is now fast enough for all practical purposes so I'm not sure who much more time we should invest in this.

megies:
> I think the module is now fast enough for all practical purposes so I'm not sure who much more time we should invest in this.

Not much more.. ;-)
I'll replace the old implementation now.. we'll have to adapt some tests most likely and fix some minor details of header unpacking..

krischer:
Current profiling results: https://gist.github.com/krischer/42cd570b6e2cd32650e730fb40d140bb

mbyt:
> Current profiling results: https://gist.github.com/krischer/42cd570b6e2cd32650e730fb40d140bb

Cool, that are line profiling results! @krischer, I am curious --- could you share how you created them (module and command line)?

krischer:
You almost guessed the name of the package ;-)  https://github.com/rkern/line_profiler

krischer:
This is also exists for memory usage: https://pypi.python.org/pypi/memory_profiler

It is less accurate (due to Python's GC I guess) but still quite useful

mbyt:
> You almost guessed the name of the package ;-) https://github.com/rkern/line_profiler

Cool, I am looking forward to trying that one :). Thanks!

mbyt:
docker tests fail for 32bit linux. I have a look now.

megies:
@mbyt, I had that problem you saw on old ubuntu before and fixed it in 21a3245e53c59abf2e02577fa53932e1462a18f4. The `native_str` won't help..

I'll exchange that commit..

Btw.. you can run that docker test locally on your machine if you have docker installed with:
```bash
$ cd misc/docker
$ ./run_obspy_tests.sh -tobspy:ccb033f416f5af678b04978deb556d61a6300ea3 -eio.reftek ubuntu_12_04_precise  # run a target commit
$ ./run_obspy_tests.sh -eio.reftek ubuntu_12_04_precise  # run current working directory state
```
Running it the first time takes quite a while to build the image but after that it's really useful and saves time waiting for our bot to test the commit.

@mbyt can I force push with a minor update right now (will include your micro optimization commit)?

mbyt:
> @mbyt can I force push with a minor update right now (will include your micro optimization commit)?

@megies yes, sure! What do I need to do to sync my reftek fork with a force pushed version (git pull obspy reftek fails for me). I usually just delete the branch and get it again. There must be a better way ;)

megies:
```bash
$ git checkout reftek  # go to that branch
$ git stash  # save local changes if there are any, just in case
$ git fetch --all
$ git reset --hard 5786efe7b3254ce568f27260d120e246b059cd71  # "git reset --hard upstream/reftek" or "git reset --hard FETCH_HEAD" should do the same
```

megies:
```bash
$ git checkout reftek  # go to that branch
$ git stash  # save local changes if there are any, just in case
$ git fetch --all
$ git reset --hard 5786efe7b3254ce568f27260d120e246b059cd71  # "git reset --hard upstream/reftek" or "git reset --hard FETCH_HEAD" should do the same
```

mbyt:
Thanks for the git help, that worked nicely :).

mbyt:
good enough for me. Ship it?

mbyt:
good enough for me: We even have improved coverage and are reasonable fast.

Ship it?

megies:
I've let the code loose on a large number of files (gigabytes of data) and there's some files that can't be read. I want to look into those a bit more before merging. But yes, in principle it's ready and can easily be merged when 1.1.0 is gonna be released.

megies:
I checked the code against 3.3 GB of data recorded with a RT130. Data converted using obspy reftek->mseed is exactly the same (all individual samples, timing etc.) as when converted using Reftek's own conversion tool `rt_mseed`. Plus.. I found that `rt_mseed` sometimes omits a chunk of samples at the end of a file (30-100 samples; happens with 54 out of 4095 files I converted).. ;-)

So, I think this is good to go. :ship:

obspy/io/reftek/tests/test_core.py
```diff
@@ -22,6 +22,16 @@ class ReftekTestCase(unittest.TestCase):
     Test suite for obspy.io.reftek
     """
     def setUp(self):
+        try:
+            # doctests of __init__.py produce warnings that get caught. if we
+            # don't raze the slate out the registry here, we can't test those
+            # warnings in the unit tests (if doctests run before unittests)..
```
QuLogic: Which warnings do you mean? If you're referring to line 381; that block is missing the `warnings.simplefilter('always')` that is in every other one.

obspy/io/mseed/headers.py
```diff
@@ -523,15 +523,17 @@ class UDIFF(C.Union):
 clibmseed.ms_detect.restype = C.c_int
 
 clibmseed.msr_decode_steim2.argtypes = [
-    C.POINTER(C.c_int32), C.c_int,
+    C.c_void_p,
```
QuLogic: Weird, why was this wrong?

obspy/io/mseed/util.py
```diff
@@ -828,43 +828,43 @@ def _convert_mstime_to_datetime(timestring):
     return UTCDateTime(timestring / HPTMODULUS)
 
 
-def _unpack_steim_1(data_string, npts, swapflag=0, verbose=0):
+def _unpack_steim_1(data, npts, swapflag=0, verbose=0):
     """
     Unpack steim1 compressed data given as string.
 
-    :param data_string: data as string
+    :param data: data as string
```
QuLogic: As a NumPy array now, no?

obspy/io/mseed/util.py
```diff
@@ -828,43 +828,43 @@ def _convert_mstime_to_datetime(timestring):
...
     return datasamples
 
 
-def _unpack_steim_2(data_string, npts, swapflag=0, verbose=0):
+def _unpack_steim_2(data, npts, swapflag=0, verbose=0):
     """
     Unpack steim2 compressed data given as string.
 
-    :param data_string: data as string
+    :param data: data as string
```
QuLogic: NumPy array here, too?

obspy/io/reftek/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+Currently the low level read routines are designed to operate on waveform files
+written by RefTek 130 digitizers which are composed of event header/trailer and
+data packages. These packages do not store information on network or location
+code during acquisition. Furthermore, it is unclear how consistently the level
+of detail on the recorded channel codes is set in the headers (real world test
+data at hand recorded with a Reftek 130 do contain the information on the first
+two channel code characters, the band and instrument code but lack information
+on the component codes, i.e. ZNE, which with high likelihood were set in the
+acquisition parameters). Therefore, additional information on network and
+location codes should be supplied and component codes should ideally be
```
QuLogic: ... and ideally component codes should be ...

obspy/io/reftek/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+supplied as well when reading files with :func:`~obspy.core.stream.read` (or
+have to be filled in manually after reading). See the low-level routine
```
QuLogic: have to -> should

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+    if not os.path.isfile(filename):
+        return False
+    filesize = os.stat(filename).st_size
+    # check if overall file size is a multiple of 1024
+    if filesize < 1024 or filesize % 1024 != 0:
+        return False
+
+    with open(filename, 'rb') as fp:
+        # check first 20 expected packets' type header field
```
QuLogic: I don't see any limit to 20.

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+class Reftek130(object):
+    def __init__(self):
+        self._data = np.array([], dtype=PACKET_FINAL_DTYPE)
+
+    def __str__(self, compact=True):
+        if compact:
+            info = [
+                "Reftek130 ({:d} packets)".format(len(self._data)),
+                "Packet Sequence  Byte Count  Data Fmt  Sampling Rate      "
+                "Time",
```
QuLogic: The wrapping is a bit unfortunate here; I recommended making it a (private) class variable that's `format`ted in this method, so then it can be de-indented a bit.

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+            count_not_implemented = not_implemented.sum()
+            types_not_implemented = np.unique(
+                self._data['packet_type'][not_implemented])
+            msg = ("Encountered some packets of types that are not "
+                   "implemented yet (types: {}). Dropped {:d} packets "
+                   "overall.")
+            msg = msg.format(types_not_implemented.tolist(),
+                             count_not_implemented)
+            warnings.warn(msg)
+            return
```
QuLogic: Is this `return` supposed to be here? It means the next line is basically useless.

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+        for event_number in np.unique(self._data['event_number']):
+            data = self._data[self._data['event_number'] == event_number]
+            # we should have exactly one EH and one ET packet, truncated data
+            # sometimes misses the header or trailer packet.
+            eh_packets = data[data['packet_type'] == b"EH"]
+            et_packets = data[data['packet_type'] == b"ET"]
+            if len(eh_packets) == 0 and len(et_packets) == 0:
+                msg = ("Reftek data contains data packets without "
+                       "corresponding header or trailer packet.")
+                raise Exception(msg)
```
QuLogic: Please use a more specific error, `ValueError` or a custom exception.

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+
+                    # Unfortunately the whole data cannot be unpacked with one
+                    # call to libmseed as some payloads do not take the full
+                    # 960 bytes. They are thus padded which would results in
+                    # padded pieces directly in a large array and libmseed
+                    # (understandably) does not support that.
+                    #
+                    # Thus we resort to *tada* pointer arithmetics in Python
+                    # ;-) This is quite a bit faster then correctly casting to
+                    # an integer pointer so its worth it.
```
QuLogic: it's

obspy/io/reftek/core.py
```diff
@@ -0,0 +1,341 @@
...
+                    # so it might be worth it.
+                    npts = packets_["number_of_samples"].sum()
+                    if headonly:
+                        sample_data = np.array([], dtype=np.int32)
+                    else:
+                        sample_data = np.empty(npts, dtype=np.int32)
+                        pos = 0
+                        s = packets_[0]["payload"][40:].ctypes.data
+                        if len(packets_) > 1:
+                            offset = (
```
QuLogic: Is the offset really always constant?

obspy/io/reftek/packet.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @staticmethod
+    def from_data(data):
+        packet_type = data['packet_type'].decode("ASCII", "ignore")
+        if packet_type in ("EH", "ET"):
+            return EHPacket(data)
+        elif packet_type == "DT":
+            return DTPacket(data)
+        else:
+            raise NotImplementedError()
```
QuLogic: A message mentioning `packet_type` would be nice.

obspy/io/reftek/packet.py
```diff
@@ -0,0 +1,244 @@
...
+               "truncated. Dropping {:d} byte(s) at the end.").format(tail)
+        warnings.warn(msg)
+    data = np.fromstring(
+        bytestring, dtype=PACKET_INITIAL_UNPACK_DTYPE)
+    result = np.empty_like(data, dtype=PACKET_FINAL_DTYPE)
+
+    for name, dtype_initial, converter, dtype_final in PACKET:
+        if converter is None:
+            result[name][:] = data[name][:]
+            continue
```
QuLogic: Nothing else really happens here, so an `else` kind of makes more sense.

obspy/io/reftek/tests/test_core.py
```diff
@@ -0,0 +1,390 @@
...
+        Test original reftek 130 data file against miniseed files converted
+        using "rt_mseed" utility from Trimble/Reftek.
+
+        rt_mseed fills in network as "XX", location as "01" and channels as
+        "001", "002", "003".
+        """
+        st_reftek = _read_reftek130(
+            self.reftek_file, network="XX", location="01",
+            component_codes=["1", "2", "3"])
+        # centralized the equality check, so that it can be reused in other
```
QuLogic: Not sure why this comment is here and not in other places, but it kind of seems obvious...

obspy/io/reftek/tests/test_core.py
```diff
@@ -0,0 +1,390 @@
...
+            "2015-10-09T22:51:22.025000Z",
+            "0026 DT AE4C  0 1024  427  0 C0             2  759 "
+            "2015-10-09T22:51:21.595000Z",
+            "0027 DT AE4C  0 1024  427  0 C0             0   67 "
+            "2015-10-09T22:51:25.055000Z",
+            "0028 ET AE4C  0  416  427  0 C0 KW1    200         "
+            "2015-10-09T22:50:51.000000Z",
+            "(detailed packet information with: "
+            "'print(Reftek130.__str__(compact=False))')"]
+        with warnings.catch_warnings(record=True):
```
QuLogic: This context is missing the `simplefilter`.

obspy/io/reftek/packet.py
```diff
@@ -0,0 +1,244 @@
...
+                'channel_number', 'number_of_samples', 'data_format', 'flags')
+
+    def __init__(self, data):
+        self._data = data
+
+    def __str__(self, compact=False):
+        if compact:
+            info = ("{:04d} {:2s} {:4s} {:2d} {:4d} {:4d} {:2d} {:2s} "
+                    "           {:2d} {:4d} {!s}").format(
+                        self.packet_sequence, self.type.decode(),
```
QuLogic: Is there a reason for not decoding stuff like this when read?

obspy/io/reftek/tests/test_core.py
```diff
@@ -0,0 +1,390 @@
...
+
+class ReftekTestCase(unittest.TestCase):
+    """
+    Test suite for obspy.io.reftek
+    """
+    def setUp(self):
+        try:
+            # doctests of __init__.py produce warnings that get caught. if we
+            # don't raze the slate out the registry here, we can't test those
+            # warnings in the unit tests (if doctests run before unittests)..
```
QuLogic: Which warnings do you mean? If you're referring to line 381; that block is missing the `warnings.simplefilter('always')` that is in every other one.

obspy/io/mseed/headers.py
megies: I think this commit was done due to this report: http://tests.obspy.org/57634/ that was building this commit: 1f4d2a84e48f258648e976c335715150a5aa31b1

The reason.. dunno.. @mbyt?

obspy/io/mseed/util.py
megies: Looks like, yes.. I'll change it..  CC @krischer

obspy/io/reftek/core.py
megies: Euuuhm.. this must've been lost during rebasing or something.. or I just dreamed that I've implemented it.. ;-)

obspy/io/mseed/util.py
krischer: Yea true - I forgot the change the description as well.

obspy/io/reftek/core.py
megies: Euuuh.. nice catch, don't know how this could get so garbled up, must've gone wrong during squashing and rebasing..

obspy/io/reftek/core.py
megies: @krischer implemented this, I can't quite make sense of the data unpacking right now, actually.. I don't even see how the consecutive packets can be correctly unpacked as the msr decode call always works on the input array of the first packet..??

obspy/io/reftek/packet.py
megies: I just thought I'd leave out any unnecessary operations during reading (for speed mostly, I guess) and do it lazily if the info really is accessed.

obspy/io/reftek/tests/test_core.py
megies: Doesn't affect the above issue, but I can put one in there, doesn't change much..

obspy/io/reftek/core.py
megies: I looked into this a bit more. To answer your question: Yes, the offset is constant assuming all packet payloads are a contiguous chunk in memory, since all packets have same length (1024 bytes) and the sample data in DT packets starts at position 40 into the payload (i.e. at byte 64 -- 24 bytes header plus the empty 40 bytes padding in the payload).

But.. what I currently don't get is why the payloads of the packets are always contiguous in memory. When I print the numpy ndarray flags **it tells me that it's both non-contiguous for C and F** but on the other hand the offset is always the same..:
```python
+ 305                         print(packets_["payload"].flags)
+ 306                         for i in range(1, len(packets_)):
>>307                             print(packets_[i]["payload"].ctypes.data - packets_[i-1]["payload"].ctypes.data)
  308                         sample_data = np.empty(npts, dtype=np.int32)
  309                         pos = 0
  310                         s = packets_[0]["payload"][40:].ctypes.data
  311                         if len(packets_) > 1:
  312                             offset = (
  313                                 packets_[1]["payload"][40:].ctypes.data - s)
  314                         else:
  315                             offset = 0
  316                         for p in packets_:
  317                             _npts = p["number_of_samples"]
  318                             clibmseed.msr_decode_steim1(
  319                                 s, 960, _npts, sample_data[pos:], _npts, None,
  320                                 1) 
  321                             pos += _npts
  322                             s += offset
```

```
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
1031
1031
1031
1031
```

I'm considering to make this foolproof by a patch.. 
```diff
diff --git a/obspy/io/reftek/core.py b/obspy/io/reftek/core.py
index db37db1..967cdbd 100644
--- a/obspy/io/reftek/core.py
+++ b/obspy/io/reftek/core.py
@@ -304,19 +304,14 @@ class Reftek130(object):
                     else:
                         sample_data = np.empty(npts, dtype=np.int32)
                         pos = 0
-                        s = packets_[0]["payload"][40:].ctypes.data
-                        if len(packets_) > 1:
-                            offset = (
-                                packets_[1]["payload"][40:].ctypes.data - s)
-                        else:
-                            offset = 0
+                        # Sample data starts at byte 40 in the DT packet's
+                        # payload.
                         for p in packets_:
                             _npts = p["number_of_samples"]
                             clibmseed.msr_decode_steim1(
-                                s, 960, _npts, sample_data[pos:], _npts, None,
-                                1)
+                                p["payload"][40:].ctypes.data, 960, _npts,
+                                sample_data[pos:], _npts, None, 1)
                             pos += _npts
-                            s += offset
 
                     tr = Trace(data=sample_data, header=header.copy())
                     if headonly:
```
to make it
```python
  305                         sample_data = np.empty(npts, dtype=np.int32)
  306                         pos = 0
~ 307                         # Sample data starts at byte 40 in the DT packet's
~_308                         # payload.
  309                         for p in packets_:
  310                             _npts = p["number_of_samples"]
  311                             clibmseed.msr_decode_steim1(
~ 312                                 p["payload"][40:].ctypes.data, 960, _npts,
~ 313                                 sample_data[pos:], _npts, None, 1)
_ 314                             pos += _npts
```

but that makes the benchmark (1 file with 1 hour 3channel 200Hz data, 2MB filesize) drop from 80 to 95 msec.. :-/

krischer: I'm fairly sure the current implementation is safe. The payloads are not contiguous in memory but are interleaved in the large record array. That one should be contiguous in memory if nothing weird is done to it.

In any case - the unpacking will complain very loudly if it gets some invalid data so any issue should be very easy to discover.

Your patch is likely slower because it has to do two item lookups, two attribute looks ups and I guess the `.data.` accessor also has some more logic. Just incrementing the memory address is naturally much faster.

megies: I added a check that tests that the overall packet array is contiguous, otherwise falling back to the slightly slower but foolproof data unpacking. Data unpacking should be 100% safe now.

obspy/io/mseed/util.py
```diff
@@ -828,43 +828,45 @@ def _convert_mstime_to_datetime(timestring):
...
     """
-    dbuf = data_string
-    datasize = len(dbuf)
+    datasize = len(data)
     samplecnt = npts
     datasamples = np.empty(npts, dtype=np.int32)
+
     nsamples = clibmseed.msr_decode_steim1(
-        C.cast(dbuf, C.POINTER(C.c_int32)),
+        data.ctypes.data,
```
mbyt: MARK for @QuLogic for comment above.

mbyt:
if there are still things left over, let's fix them in master or independently... Happy merging :)

PR #1436, opened 2016-06-01
MiniSEED files is also (in the header) a valid tarfile....

krischer:
Curious issue from the mailing list:

http://lists.swapbytes.de/archives/obspy-users/2016-June/002072.html

The problem is that our uncompress_file decorator thinks this is a tarfile and even manages to unpack one 0 byte file. Even the tar utility thinks its a (corrupt) tarfile.

This PR fixes the issue by just not appending empty untarred files to the list of unpacked objects. Thus nothing will be unpacked and ObsPy thinks this is not a tarfile and the rest of the format detection routines proceeds normally.

Duplicate of #1435 but this time with the correct base branch.


krischer:
I do not yet know if we can use the test file so don't merge yet.

megies:
> I do not yet know if we can use the test file so don't merge yet.

It's out in the open now anyway.. :stuck_out_tongue_winking_eye:

krischer:
> Nice branch name by the way 😆

:sparkler:

PR #1432, opened 2016-06-01
Can't read sac file (null-terminated strings with invalid unicode characters in SAC string headers)

megies:
My office system is windows 32 and python version is 2.7.
When i read the sac file from our Central Weather Bureau(CWB), the system print the error message. 
I typed command:
`st = read(CWB_sacfile, format='sac')`
And i got:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xa1 in position 7: ordinal not in range(128)`

I used SAC to read the same file and it could work. How can i fix this problem? 

Thanks & Regards
Jimmy Tseng
National Cheng Kung University, Tainan, Taiwan.


jkmacc-LANL:
Ah, yes.  When the invalid characters are ignored, all comparisons to the string null `-12345` are `False`, even when they should be `True`, like `-12345\x00\x00\x00\x00\x00\x00+`.  That's why the SEED ID looks funky.  Those fields (as far as I can tell) were supposed to be empty.  

```python
In [9]: tr.stats
Out[9]:
         network: -12345
         station: ALS
        location: -12345
         channel: HHE
       starttime: 2011-01-01T00:04:00.000000Z
         endtime: 2011-01-01T00:07:59.990000Z
   sampling_rate: 100.0
           delta: 0.01
            npts: 24000
           calib: 1.0
         _format: SAC
             sac: AttribDict({u'unused23': 1, u'nzyear': 2011, u'nzjday': 1, u'iztype': 5, u'khole': u'-12345',
 u'knetwk': u'-12345\x00\x14', u'lcalda': 1, u'idep': 50, u'iftype': 1, u'kt8': u'-12345', u'kt9': u'-12345\x00+',
 u'nvhdr': 6, u'kuser0': u'-12345\x00\ufffd', u'kt5': u'-12345', u'kt6': u'-12345\x00w', u'kt7': u'-12345\x00H', 
u'kt0': u'-12345', u'kt1': u'-12345\x00@', u'kt2': u'-12345', u'kt3': u'-12345\x00H', u'kcmpnm': 
u'HHE\x0045', u'nzmin': 4, u'nzsec': 0, u'lpspol': 1, u'kevnm': u'-12345\x00\x00\x00\x00\x00\x00+', u'ka': 
u'-12345\x00\x7f', u'lovrok': 1, u'delta': 0.0099999998, u'stla': 23.508278, u'kdatrd': u'-12345\x00\x01', 
u'nzmsec': 0, u'leven': 1, u'kinst': u'-12345\x00+', u'b': 0.0, u'e': 240.0, u'kf': u'-12345\x00w', u'stlo': 
120.81336, u'ko': u'-12345', u'kstnm': u'ALS\x00\x00\x01\x14', u'kt4': u'-12345', u'npts': 24000, u'kuser1': 
u'-12345', u'nzhour': 0, u'kuser2': u'-12345\x00='})

In [10]: tr.id
Out[10]: u'-12345.ALS.-12345.HHE'

In [11]: tr.stats.sac.knetwk
Out[11]: u'-12345\x00\x14'

In [12]: tr.stats.network
Out[12]: u'-12345'
```
You'll notice that strings in the `Trace` header look OK, though.  Apparently, those go through [`util._clean_str`](https://github.com/obspy/obspy/blob/master/obspy/io/sac/util.py#L162) first.  The pre-1.0 ObsPy SAC module had no special checking of null values (nothing like `if val == HD.SNULL:`) until forming a `Trace` header, at which time the same [snippet of code](https://github.com/obspy/obspy/blob/0.10.3/obspy/sac/sacio.py#L1209) handled them.  

I'm not sure it'll work to simply ignore non-ASCII characters.  Maybe I could use `_clean_str` before any string comparisons to the null happen, but that would strip those characters out.  I'll check if non-ASCII characters allowed in the SAC program.  Other suggestions welcomed...

claudiodsf:
I think that non-ascii characters are not allowed in the SAC format specification, but it's better checking.

The problem the see here comes from the wrong assumption that a null-terminated C string can be stored as it is in the SAC header. But, in fact, there is no null-termination in the sac header string specification and all the strings have to be filled up to the last byte with valid characters (e.g., whitespace). 

My advice is therefore to sanitize the strings by replacing every character after the null-terminator (and including the null-terminator) with whitespace. I would not touch the characters before the null-terminator.

megies:
Yes, I agree, let's do it the same way as before in https://github.com/obspy/obspy/commit/2a8d189666cbfeef7f9df5b24779c00feda23d00.
I've pushed a fix. However, those '-12345' values still appear in the Trace headers, since `HD.SNULL` is `"-12345  "` and thus it doesn't match the `"-12345"`. I'll leave it to @jkmacc-LANL from here, not sure how you want this fixed.

jkmacc-LANL:
I don't know about the SAC format spec, but I see that the SAC program seems to simply truncate trailing non-ASCII characters at read time.  `ALS\x00\x00\x01\x14` becomes `ALS`, and `-12345\x00\x00\x00\x00\x00\x00+` becomes the unset/null `-12345`.  

> ...all the strings have to be filled up to the last byte with valid characters (e.g., whitespace).

The NumPy `dtype` of the string array that gets read from and written to disk is `|S8`, so I think this ensures that strings that are too long are truncated to 8 and strings that are too short are padded to 8 during I/O.

In Python-land, though, there are two desired behaviors: 1) replace all non-ASCII characters starting from and including a null-terminator with spaces at read time, so that comparisons to `HD.SNULL` still work, and 2) remove non-ASCII characters *and* strip out whitespace, so that building the `Trace.Stats` header still works as expected.  @megies I'm happy to give this a try tonight, unless this is trivial for you to flag that into `_decode_bytes_and_warn`.  Thanks!

megies:
> In Python-land, though, there are two desired behaviors: 1) replace all non-ASCII characters starting from and including a null-terminator with spaces at read time, so that comparisons to HD.SNULL still work, and 2) remove non-ASCII characters and strip out whitespace, so that building the Trace.Stats header still works as expected. @megies I'm happy to give this a try tonight, unless this is trivial for you to flag that into _decode_bytes_and_warn. Thanks!
> 
> EDIT: I guess another possibility is to set HD.SNULL to '-12345', and then clean and strip strings during I/O and when setting string header values.

Whatever makes most sense for you, I'd say. We just want empty strings as e.g. `trace.stats.network` for the above example of an obviously non-set value like in the example file.

megies:
Force-pushed, fixed the regex pattern for Py3.

obspy/io/sac/util.py
```diff
@@ -462,3 +463,20 @@ def get_sac_reftime(header):
...
     return reftime
+
+
+def _decode_bytes_and_warn(bytestring, key=None, strip_null_terminated=True):
+    """
+    Decodes bytestring to ASCII, ignoring invalid characters if present and
+    issuing a warning if invalid characters are encountered.
+    """
+    if strip_null_terminated:
+        bytestring = re.sub(b'\x00.*', '', bytestring)
```
QuLogic: Do you really need `re`? `bytestring.split(b'\x00', 1)[0]` (Also, didn't we have to do this somewhere before?)

megies: Aye.. trying that one now, looks good from local tests..
The original `.find()` was failing because it sometimes got bytes, now the `re` one is failing because it sometimes gets a str on Py3.. this Py2/3 is still freaking me out sometimes..

megies:
@jkmacc-LANL I've changed the reading, so the file is read as follows now:
```
1 Trace(s) in Stream:
.ALS..HHE | 2011-01-01T00:04:00.000000Z - 2011-01-01T00:07:59.990000Z | 100.0 Hz, 24000 samples
```
Please review if the changes are OK for you, I'm really not that much into the details of `io.sac`.
A test would be good to obviously, but the file is too large.

claudiodsf:
[Here](https://github.com/obspy/obspy/files/300106/test.sac.zip) is just the header of the same SAC file. Should be fine for the test. Feel free to find an appropriate name for it.

jkmacc-LANL:
Sorry for being incommunicado this weekend.  Thanks Claudio for the file.  I'm still seeing some funky behavior, as `kevnm` is being returned as an empty string `u''`, for example.  If there are no objections, I'm thinking of having `_decode_bytes_and_warn` replace non-ASCII with whitespace (Claudio's original recommendation), moving it to `SACTrace.read` and `SACTrace.write`, and maybe offering a keyword argument there that could skip it (for anyone needing to debug).

jkmacc-LANL:
The current plan:

* [ ] Decode all ASCII to unicode at read time, replacing non-ASCII with whitespace by default.
* [ ] Deal with unicode in Python land.
* [ ] Strip whitespace when building the `Trace` header `Stats`.
* [ ] Encode all unicode to ASCII at write time.
* [ ] Write some tests using the provided minimal SAC alphanumeric file.

jkmacc-LANL:
Crap.  Rebased on master.  Googling how to fix it...

jkmacc-LANL:
Apparently, something in `maintenance_1.0.x` fixed this issue.  The problem was fixed after I rebased to bring changes into this branch.  I just added a test using the minimal file @jimmytseng79 and @claudiodsf provided.  I tried and ultimately removed the changes @megies made, as they didn't seem to be necessary.  I will squash these once the tests pass.

QuLogic:
> Crap. Rebased on master. Googling how to fix it...

`git rebase --onto maintenance_1.0.x master`

megies:
> Apparently, something in maintenance_1.0.x fixed this issue. The problem was fixed after I rebased to bring changes into this branch.

@jkmacc-LANL Did you try the file provided by the original poster?  https://drive.google.com/open?id=0B131kvFL7CPmSXVnMWdGbVRCcDQ

For me the current tip of this branch is not able to read the file, with the same error message as reported in the OP.

claudiodsf:
I also confirm that I get the error:

```python
UnicodeDecodeError: 'ascii' codec can't decode byte 0xa1 in position 7: ordinal not in range(128)
```

 using the short test file and the current branch tip 77feeac.

jkmacc-LANL:
Sorry, everyone.  I thought `null_termination.sac` was what I renamed the shortened file @claudiodsf provided, but it's not, and I ended up rewriting a test that already existed.  Fun!  I'll fix it.  I know you guys are waiting on this.

jkmacc-LANL:
It's not pretty, but I think I'm ready to move forward with this.  Thanks for your patience, everyone.

obspy/io/sac/util.py
```diff
@@ -159,19 +159,24 @@ def is_same_byteorder(bo1, bo2):
     return (bo1.lower() in le) == (bo2.lower() in le)
 
 
-def _clean_str(value):
+# This fn is used in two places: in SACTrace.read, to sanitize strings for
+# SACTrace, and in sac_to_obspy_header, to sanitize strings for Trace that the
+# user may have manually added.
```
krischer: Any reason this is not a docstring?

krischer:
Looks good to me! Can you still add a changelog entry? Otherwise this IMHO is ready to merge.

obspy/io/sac/util.py
```diff
@@ -159,19 +159,24 @@ def is_same_byteorder(bo1, bo2):
...
         # value is a bytes
-        # null_term = value.decode().find('\x00')
         null_term = value.find(b'\x00')
+    except UnicodeError:
+        null_term = value.decode('ASCII', 'replace').find(b'\x00')
 
     if null_term >= 0:
-        value = value[:null_term]
-    value = value.strip()
+        value = value[:null_term] + b" " * len(value[null_term:])
```
megies: Can't we just cut the thing off at the termination character (like was done before in the old obspy.sac implementation)? It's a termination character after all, so in my understanding the interpreted string should stop there?

Also, if `value` was a string before, you're now adding bytes to it..

obspy/io/sac/tests/test_core.py
```diff
@@ -862,6 +864,16 @@ def test_invalid_header_field(self):
 
                     self.assertEqual(buf.read(), buf1.read())
 
+    def test_not_ascii(self):
+        """
+        Read file with non-ascii and null-termation characters.
```
megies: typo..

obspy/io/sac/util.py
megies: :+1: for making it a docstring. Also, I only see this function used in `SACTrace.read`, actually.

jkmacc-LANL: It's also in `util.sac_to_obspy_header`.  

Docstring it is.

obspy/io/sac/util.py
jkmacc-LANL: > Can't we just cut the thing off at the termination character (like was done before in the old obspy.sac implementation)?

Except for `kevnm`, SAC string fields are 8 characters long, so I think a termination character before the last position should expect to be whitespaced-over.  In old `obspy.sac` [implementation](https://github.com/obspy/obspy/blob/0.10.2/obspy/sac/sacio.py#L1209), this operation was only used while building a dictionary for `Stats`, which doesn't care for extra whitespace.  `_clean_str` is now being used to also sanitize the NumPy array that contains the raw strings, which does care about trailing whitespace, hence the new flag.

> Also, if value was a string before, you're now adding bytes to it..

I'm still trying to wrap my head around strings vs bytes.  Can you suggest an example where this would fail, and we can add it to the tests?

megies: screenshot from 2016-07-28 15 35 47

n.b. only failing on Python3, Python2 is doing auto de/encoding when possible.

jkmacc-LANL: Changing `b' '` to `' '` or `u' '` or `native_str(' ')` fails tests on Python 3, with `TypeError: can't concat bytes to str`.  

 `_clean_str` probably needs someone more experienced than I am to look it over in more depth, particularly if we want any such changes in `1.0.3`, but at the moment every use case we've thrown at it in the test suite is currently passing.  I'm happy to open an issue for this after `1.0.3`.

QuLogic: The better question is why is it working on both `str` and `bytes` (cf. the try/except)? It should pick one and be consistent.

megies: > but at the moment every use case we've thrown at it in the test suite is currently passing

Fair enough I guess. Test-driven development in operation.. ;-)

megies:
I felt free to squash this into two commits and force push. I left my earlier fix (which is reverted in the second commit) in as a separate commit just in case it might be useful to look at it if there's further problems.

Guess, we can merge this and do a final RC for 1.0.2 when CI turns green.

obspy/io/sac/util.py
megies: > The better question is why is it working on both str and bytes (cf. the try/except)? It should pick one and be consistent.

The `TypeError` branch is never hit, i.e. it never works on bytes: https://coveralls.io/builds/7197433/source?filename=%2Fhome%2Ftravis%2Fminiconda%2Fenvs%2Ftest-environment%2Flib%2Fpython2.7%2Fsite-packages%2Fobspy%2Fio%2Fsac%2Futil.py#L165

megies:
I made a stab at cleaning up this helper function.. I figured, since you decode at the end, why not just do it right at the start..

jkmacc-LANL:
@megies Yes, I noticed this too when I started rebasing.  I wonder if I can manually trigger it...

megies:
CI is all green, so merging..

jkmacc-LANL:
Thanks for your help, everyone!

PR #1437, opened 2016-06-02
raise clear message in rotate2zne if length is not equal

megies:
..otherwise numpy raises:

`ValueError: setting an array element with a sequence.`

### DIFF
--- a/obspy/signal/rotate.py
+++ b/obspy/signal/rotate.py
@@ -242,6 +242,10 @@ def rotate2zne(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
             np.arange(3) * 3, 0, -90) # doctest: +NORMALIZE_WHITESPACE
     (array([ 0., 3., 6.]), array([ 0., 1., 2.]), array([ 0., 2., 4.]))
     """
+    if len(set(len(i_) for i_ in (data_1, data_2, data_3))) != 1:
+        msg = "All three data arrays must be of same length."
+        raise ValueError(msg)
+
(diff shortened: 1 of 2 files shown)


obspy/signal/rotate.py
```diff
@@ -242,6 +242,10 @@ def rotate2zne(data_1, azimuth_1, dip_1, data_2, azimuth_2, dip_2, data_3,
             np.arange(3) * 3, 0, -90) # doctest: +NORMALIZE_WHITESPACE
     (array([ 0., 3., 6.]), array([ 0., 1., 2.]), array([ 0., 2., 4.]))
     """
+    if set(len(i_) for i_ in (data_1, data_2, data_3)) != 1:
```
QuLogic: I think you want to check the `len` of the set, not the set itself.

megies: Oops, of course..

obspy/signal/tests/test_rotate.py
```diff
@@ -144,6 +145,26 @@ def test_rotate2zne_round_trip(self):
...
+        e = np.ones(3, dtype=np.float64)
+
+        # Random values.
+        dip_1, dip_2, dip_3 = 0.0, 30.0, 60.0
+        azi_1, azi_2, azi_3 = 0.0, 170.0, 35.0
+
+        if PY2:
+            testmethod = self.assertRaisesRegexp
+        else:
+            testmethod = self.assertRaisesRegex
```
krischer: Oh nice - I had no clue this existed!

If we end up using this more places we could generate an ObsPy `TestCase` class and fix it there.

PR #1439, opened 2016-06-03
add setter for Trace.id property

megies:
So much more convenient to do..

```python
 >>> tr.id = "GR.FUR..HHZ"
```

than..

```python
 >>> tr.stats.network = "GR"
 >>> tr.stats.station = "FUR"
 >>> tr.stats.location = ""
 >>> tr.stats.channel = "HHZ"
```

Includes tests..

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2591,6 +2591,30 @@ class TraceTestCase(unittest.TestCase):
             tr_float64.copy().interpolate(
                 1.0, method="lanczos", a=2).data.dtype, np.float64)
 
+    def test_set_trace_id(self):
+        """
+        Test setter of `id` property.
+        """
+        tr = Trace()
+        tr.stats.location = "00"
+        # check setting net/sta/loc/cha with an ID
+        tr.id = "GR.FUR..HHZ"
+        self.assertEqual(tr.stats.network, "GR")
+        self.assertEqual(tr.stats.station, "FUR")
+        self.assertEqual(tr.stats.location, "")
+        self.assertEqual(tr.stats.channel, "HHZ")
+        # check that invalid types will raise
+        invalid = (True, False, -10, 0, 1.0, [1, 4, 3, 2], np.ones(4))
+        for id_ in invalid:
+            with self.assertRaises(TypeError):
+                tr.id = id_
+        # check that invalid ID strings will raise
+        invalid = ("ABCD", "AB.CD", "AB.CD.00", "..EE", "....",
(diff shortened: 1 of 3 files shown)


obspy/core/trace.py
```diff
@@ -807,6 +807,38 @@ def get_id(self):
...
+
+        >>> from obspy import read
+        >>> tr = read()[0]
+        >>> print(tr)  # doctest: +ELLIPSIS
+        BW.RJOB..EHZ | 2009-08-24T00:20:03.000000Z ... | 100.0 Hz, 3000 samples
+        >>> tr.id = "GR.FUR..HHZ"
+        >>> print(tr)  # doctest: +ELLIPSIS
+        GR.FUR..HHZ | 2009-08-24T00:20:03.000000Z ... | 100.0 Hz, 3000 samples
+
+        :type value: str
```
krischer: Maybe also add a `:param value:` key or remove the line?

obspy/core/tests/test_trace.py
```diff
@@ -2591,6 +2591,30 @@ def test_dtype_is_not_unnecessarily_changed(self):
...
+        def set_id_test(x):
+            tr.id = x
+
+        invalid_types = (True, False, -10, 0, 1.0, [1, 4, 3, 2], np.ones(4))
+        for id_ in invalid_types:
+            self.assertRaises(TypeError, set_id_test, id_)
+        invalid_ids = ("ABCD", "AB.CD", "AB.CD.00", "..EE", "....",
+                       "GR.FUR..HHZ.", "GR.FUR..HHZ...")
+        for id_ in invalid_ids:
+            self.assertRaises(ValueError, set_id_test, id_)
```
krischer: You could also do

```python
with self.assertRaises(ValueError):
    tr.id = id_
```

Then the test is more explicit and nicer to look at :-)

obspy/core/trace.py
```diff
@@ -807,6 +807,38 @@ def get_id(self):
...
+        """
+        try:
+            net, sta, loc, cha = value.split(".")
+        except AttributeError:
+            msg = ("Can only set trace's SEED ID from a string "
```
krischer: s/set trace's/set a Trace's

obspy/core/trace.py
megies: Don't think this is showing up in the API anywhere, just put it for devs working on the source.
But yeah, why not do it properly..

obspy/core/tests/test_trace.py
megies: Oh, nice. :+1:  Didn't know it can be used as a context manager.

PR #1444, opened 2016-06-07
refactor imaging.scripts.scan

megies:
This refactors the clumsy `imaging.scripts.scan.main` god-routine used in `obspy-scan` command line script into a class that can be reused programmatically in other code (e.g. scanning files on demand and only doing a summary plot at the end).


obspy/imaging/scripts/scan.py
```diff
@@ -192,6 +196,374 @@ def load_npz(file_, data_dict, samp_int_dict):
...
+
+    def plot(self, show=True, fig=None, outfile=None, plot_x=True,
+             plot_gaps=True, print_gaps=False, event_times=None):
+        """
+        Plot the information on parsed waveform files.
+        """
+        import matplotlib.pyplot as plt
+
+        if fig:
+            fig = plt.figure()
```
QuLogic: Huh?

megies: Euh... thanks!

obspy/imaging/scripts/scan.py
```diff
@@ -192,6 +196,373 @@ def load_npz(file_, data_dict, samp_int_dict):
         npz_dict.close()
 
 
+def _seconds_to_days(seconds):
+    return seconds / (24 * 3600)
+
+
+class Scanner(object):
+    """
+    """
```
krischer: Missing content! ;-)

obspy/imaging/scripts/scan.py
```diff
@@ -192,6 +196,373 @@ def load_npz(file_, data_dict, samp_int_dict):
...
+                 seed_ids=None):
+        self.format = format
+        self.verbose = verbose
+        self.quiet = quiet
+        self.recursive = recursive
+        self.ignore_links = ignore_links
+        if starttime is None:
+            self.starttime = None
+        else:
+            self.starttime = date2num(starttime.datetime)
```
krischer: we have `UTCDateTime.matplotlib_date` now but I understand that this is just a copy/paste job.

krischer:
Why did the test image change so much?

megies:
> Why did the test image change so much?

When setting both `starttime` and `endtime` for the scan and a specific SEED ID (either explicitly specified via `seed_ids` option or when picked up in one of the scanned files) has no data at all in between the two it is now plotting a full-extent gap. I think that makes sense and not plotting that full-extent gap before has to be considered kind of a bug (we already were plotting a gap if there was some data in the middle but not out to starttime or endtime, see also earlier images that have two tiny data chunks for two of the SEED IDs).

krischer:
Ok that makes sense. Maybe make the cross a bit bolder? its very hard to see in the new plot. This also has to be part of the Changelog and we should later remember to include this in the release notes of ObsPy 1.1.

megies:
> Maybe make the cross a bit bolder? its very hard to see in the new plot

The crosses are in the parts of the plot that didn't change.. so has nothing to do with this PR..

I'll add a changelog entry.

megies:
Rebased and force-pushed..

megies:
I did some more refactoring of the Scanner class and routines around it, and I think this should be ready when CI is OK with it.

The main purpose of this PR is to use the existing `obspy-scan` functionality (keeping track of data, loading/saving from/to npz, plotting overview image) in custom workflows when iterating over large numbers of files (or downloading data, etc.) and having to read them anyways. With these changes it's possible to programmatically add Stream information to the Scanner while iterating through data.

Minimal example use case:

```python
scanner = Scanner()

for f in files:
    st = read(f)
    do_some_processing_on_the_data(st)
    scanner.add_stream(stream)

scanner.save_npz('processed_data_info.npz')
scanner.plot('processed_data_info.png')

PR #1445, opened 2016-06-07
Wiglets

rmlopes:
This PR adds an option to section plots that allows coloring the wiglets (inside the signal curves).
I did not find a related issue or any people explicitly interested, but it is should be of interest to everyone working with stratigraphic interpretation.

I have corrected the formatting. What is the problem with the tests with images?


CHANGELOG.txt
```diff
@@ -33,7 +33,8 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
-
+ - obspy.imaging.waveform
+   * Support for filling the signal wiglets when plotting sections (horizontal and vertical).
 1.0.2: (doi: 10.5281/zenodo.49636)
```
krischer: Can you add a newline above the header for 1.0.2?

obspy/CONTRIBUTORS.txt
```diff
@@ -67,3 +67,4 @@ Marcus Walther
 Joachim Wassermann
 Mark C. Williams
 Andrew Winkelman
+Rui L. Lopes
```
krischer: The list is alphabetically sorted by last name.

obspy/imaging/waveform.py
```diff
@@ -214,6 +215,8 @@ def __init__(self, **kwargs):
         self.one_tick_per_line = kwargs.get('one_tick_per_line', False)
         self.show_y_UTC_label = kwargs.get('show_y_UTC_label', True)
         self.title = kwargs.get('title', self.stream[0].id)
+        #If true the signal wiglets will be filled
+        self.wiglets = kwargs.get('wiglets', True)
```
krischer: Please make sure this defaults to False - this should also fix most test failures.

obspy/imaging/waveform.py
```diff
@@ -1279,8 +1282,13 @@ def __sect_init_plot(self):
         """
         Function initialises plot all the illustration is done by
         self.plot_section()
-        """
+        """
```
megies: No whitespace at end of line, please.

krischer:
Looks pretty nice. There are a couple of test failures, please fix them:

http://tests.obspy.org/42163/

Mostly code formatting and that `wiglets` defaults to `True` which changes lots of existing tests. Also the pictures generated by travis are slightly different than your baseline images but we can figure this out later.


Two more comments:

* Is `wiglets` a common name for filling the graphs? I've never heard that before but who I am I to say.
* People will definitely want to configure the colors for the fill, a simple way to offer that option would be to allow `wiglets=("green", "orange")` for positive/negative or vice-versa fills.

CHANGELOG.txt
```diff
@@ -33,7 +33,8 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
-
+ - obspy.imaging.waveform
+   * Support for filling the signal wiglets when plotting sections (horizontal and vertical).
```
krischer: Please make sure this line does not exceed 79 characters and also add the number of this PR `(see #1445)`.

obspy/imaging/waveform.py
```diff
@@ -1300,6 +1308,10 @@ def __sect_init_plot(self):
             else:
                 raise NotImplementedError("sect_orientiation '%s' is not "
                                           "valid." % self.sect_orientation)
+            if self.wiglets:
+                fillfun(time, data, self._tr_offsets_norm[_tr], where=data > self._tr_offsets_norm[_tr], facecolor='blue')
+                fillfun(time, data, self._tr_offsets_norm[_tr], where=data < self._tr_offsets_norm[_tr], facecolor='red')
```
megies: I know, right now `WaveformPlotting` really is an ugly monster and so far not very user-friendly in terms of controlling certain aspects (and in many places rather messy and implicit/intransparent by fishing tons of things out of `kwargs`)..

but.. how about starting to make it a better place.. ;-) by making those two color specs class attributes? That way, users could later have custom colors (or even just one of the two -- which is a rather typical use case) without having to hack the obspy source code..

```python
class WaveformPlotting(object):
    """
    """
    wiglet_color_1 = "blue"
    wiglet_color_2 = "red"

[...]

        if self.wiglets:
            if self.wiglet_color_1 is not None:
                fillfun(time, data, self._tr_offsets_norm[_tr], where=data > self._tr_offsets_norm[_tr], facecolor=self.wiglet_color_1)
[...]
```

Users could than change it at run time like..
```python
from obspy.imaging.waveform import WaveformPlotting
WaveformPlotting.wiglet_color_1 = "k"
WaveformPlotting.wiglet_color_1 = None

from obspy import read
read().plot(type="section")
```

krischer: Hmm..interesting API but it seems like a good way to get the functionality we desire. And maybe we can start to deprecate the gazillion function parameters we have there.

Maybe better to offer a way to reset the style?

```python
default_style = obspy.core.AttribDict({
    "wiglet_color_1": ...
    ...
})

class WaveformPlotting(object):
    style = default_style.copy()

    @classmethod
    def reset_style(cls):
        cls.style = default_style.copy()
```

megies:
> Looks pretty nice.

I agree, and even has image tests. :+1: Just a few details to sort out left (as stated in comments)..

> the pictures generated by travis are slightly different than your baseline images

Easiest solution: grab those two images that Travis produced ([here](http://i.imgur.com/RV0aBfL.png) and [here](http://i.imgur.com/W3K26Bu.png), see http://tests.obspy.org/42163/#6) and commit them.

> a simple way to offer that option would be to allow `wiglets=("green", "orange")` for positive/negative or vice-versa fills

Yeah, good alternative to my other comment https://github.com/obspy/obspy/pull/1445#discussion_r66214730. :+1: (I would still add the `None` check to be able to switch off either wiggle half)

rmlopes:
Maybe I should have checked the name first. I've heard people call wiggles and wiglets, the wiggle in itself is the signal not that painted area, but if I google seismic wiggle we this style of images (I will change it).

> a simple way to offer that option would be to allow wiglets=("green", "orange") for positive/negative or vice-versa fills
Yeah, good alternative to my other comment #1445 (comment). :+1: (I would still add the None check to be able to switch off either wiggle half)

My intention was to do that but was not sure what would be the best approach. So I would go with the class attributes defaulted to `None` with the `wiglets=(c1,c2)` parameter style also defaulting to `None`.  

`# Set the wiglet fill function once
 +        if self.sect_orientation == 'vertical':
 +            fillfun = functools.partial(ax.fill_betweenx, lw=0)
 +        else:
 +            fillfun = functools.partial(ax.fill_between, interpolate=True, lw=0)`

When I was correcting the formatting (not commited yet, sorry you lost time commenting on those errors) I actually moved this into the constructor, seems more appropriate.

obspy/imaging/waveform.py
rmlopes: I like this. If I am not mistaken then to load the parameters from the kwargs an `update` on the dict would suffice to do the trick.

megies: > And maybe we can start to deprecate the gazillion function parameters we have there.

:+1: but a lot of work..

> Maybe better to offer a way to reset the style?

Seems a much to me.. do we really need that much complexity..?

Or if we really need something like this.. maybe sum all those settings in one place, like matplotlib does it (with rcparams)?

rmlopes:
I still have problems with the versions built in my machines. In linux (XUbuntu) I have the same 'viridis' problem I described in #1400 for my windows machine.

In OS X I get the following error when obspy is imported:
```
  File "/Users/rmlopes/dev/obspy/obspy/__init__.py", line 36, in <module>
    from obspy.core.utcdatetime import UTCDateTime  # NOQA
  File "/Users/rmlopes/dev/obspy/obspy/core/__init__.py", line 118, in <module>
    from obspy.core.util.attribdict import AttribDict
  File "/Users/rmlopes/dev/obspy/obspy/core/util/__init__.py", line 27, in <module>
    from obspy.core.util.base import (ALL_MODULES, DEFAULT_MODULES,
  File "/Users/rmlopes/dev/obspy/obspy/core/util/base.py", line 24, in <module>
AttributeError: 'module' object has no attribute 'hooks'
```

My guess is that I am missing some dependency on the system, any ideas?

krischer:
> My guess is that I am missing some dependency on the system, any ideas?

That's really strange - do you maybe have a really old `future` version installed? That might not have the stdlib hooks yet.

rmlopes:
That was it, thanks @krischer. Now I have the `globasl["viridis"]` issue as well in OS X, but this should be related to the extensions, I'll get around it.

QuLogic:
Since wiggle really refers to the line, why not just call the kwargs fillcolors?

CHANGELOG.txt
```diff
@@ -33,7 +33,9 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
-
+ - obspy.imaging.waveform
+   * Support for filling the wiggles when plotting sections (horizontal and vertical).
+
```
QuLogic: There is trailing whitespace here.

obspy/imaging/tests/test_waveform.py
```diff
@@ -342,6 +342,58 @@ def test_plot_colored_section(self):
...
+        for _i in range(10):
+            this_start = start + 300 * np.sin(np.pi * _i / 9)
+            st += self._create_stream(this_start, this_start + 3600, 100)
+            st[-1].stats.distance = _i * 10e3
+            st[-1].stats.channel = str(_i % 3)
+        # create and compare image
+        with ImageComparison(
+                self.path, 'waveform_wiggles_positive_section.png') as ic:
+            st.plot(outfile=ic.name, type='section', orientation='horizontal',
+                    wiggle_colors=("black", None))
```
QuLogic: You need a test for the other way as well `(None, 'black')`.

obspy/imaging/waveform.py
```diff
@@ -28,6 +28,7 @@
 from dateutil.rrule import MINUTELY, SECONDLY
 
 import numpy as np
+import functools
```
QuLogic: This import belongs in the previous block.

QuLogic:
You've added and removed images a couple times, which bloats the repository a little. There's also an extraneous merge commit. Can you rebase and add them only once?

QuLogic:
You can wait until after all the review comments before rebasing.

QuLogic:
Sorry, didn't mean to close that time.

rmlopes:
No problem. Sorry, I don't know what happened the first time, it forced me to merge, but the new image had been generated by the server, I don't know why the test failed.

claudiodsf:
Concerning the name of this feature, in Seismic Unix this kind of plot is called *wiggle plot* (attention, with a double "g").

http://www.cwp.mines.edu/sututor/node62.html

(also google for `wiggle plot`)

megies:
`plot(type="section", fillcolors=("k", None))` seems OK to me, naming-wise.

Once the API/naming is settled this will also need to be documented in [Stream.plot() docstring](https://github.com/obspy/obspy/blob/master/obspy/core/stream.py#L1048)

rmlopes:
I think the naming is ok now, so that everyone understands what it stands for. The test for the last image added failed yet again. I don't really understand what is the problem since I am doing exactly what I did for the other two. but looking at the diff looks some kind of small  font/rendering issue.

Any ideas before I retry to replace it?

megies:
Last commit introduced some flake8 issues: https://ci.appveyor.com/project/obspy/obspy/build/1.0.2409-master#L654

rmlopes:
Hopefully everything will be in place now. @megies I can see the notifications as well ;)

CHANGELOG.txt
```diff
@@ -33,6 +33,8 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
+ - obspy.imaging.waveform
+   * Support for filling the wiggles when plotting sections (horizontal and vertical).
```
krischer: Can you please add the issue number as well, e.g. `(see #1445)`.

krischer:
Looks great :-) Good and simple implementations and comprehensive tests :+1:

I have one last nitpick with the changelog and then this PR would have to either be rebased or squashed to a single commit as the history contains too many versions of the test images which would unnecessarily bloat our repository.

rmlopes:
> this PR would have to either be rebased or squashed to a single commit 

Is this on my fork? I have rebased before the last commit.

megies:
I'd vote for squashing into a single commit.

krischer:
> Is this on my fork? I have rebased before the last commit.

Yes - but the commits from this PR still contain several versions of the test images. We don't want to add these to the main ObsPy branch as it is big enough already and any data entering it will never leave it. Thus please either carefully rebase to remove the images or just squash everything to one commit - we would prefer the later.

In case you are newish to git: You can do the squashing/rebasing with `git rebase -i HEAD~24` but I kind of prefer the following because its easier:

```bash
$ git branch backup
$ git reset --soft HEAD~24
```

then re-add all the stuff and commit with a single commit and `git push -f` to your branch and all is well :-)

QuLogic:
> git rebase -i HEAD~24

Why? @rmlopes is on a feature branch: `git rebase -i master`

krischer:
> Why? @rmlopes is on a feature branch: git rebase -i master

Good point. I'll also do that from now on :-)

rmlopes:
I don't have much experience with merges/rebases. I did the interactive rebase and removed the three image commits that was possible. Now that I have a diverged branch (by three commits) should I force the push?

megies:
> I don't have much experience with merges/rebases. I did the interactive rebase and removed the three image commits that was possible.

Ok. As discussed above we think that this PR is best represented with a single commit, so doing the interactive rebase (again) you could mark all commits coming after your first commit on this topic as "squash", which will merge them into a single commit with all individual commit messages together. (see e.g. [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html))

> Now that I have a diverged branch (by three commits) should I force the push?

Yes, since the history is incompatible you need to explicitly `git push -f`. (Something that should not be done on a projects' main line branches like `master` and `maintenance_1.0.x` -- but we have blocked force-pushes to those anyway on github)

> Is it possible to do a "dry-run" in github besides creating a new branch? (push to my branch without automatically updating the PR and launching the build)

No, the PR is just always looking up the tip of the associated branch and will notice when it gets updated. But why do you want a dry-run anyway? Just go ahead.
(Your local git won't forget any old state of your repository, you can always look up the hashes of commits in `git reflog`'s left column and go back to them with checkout or reset -- e.g. try looking for `3ac19cc` in your reflog, you should see it at some point before your rebase..)

rmlopes:
Thanks @megies. I wanted to know about the dry-run in general just to avoid the automatic builds. 

Should be ready now.

megies:
:+1:  Looks ready to merge for me.

claudiodsf:
The commit message could be more synthetic and descriptive of the general idea. But it's a matter of taste.
Otherwise, good to merge.

CHANGELOG.txt
```diff
@@ -33,6 +33,9 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
+ - obspy.imaging.waveform
```
QuLogic: These sections should remain in alphabetical order.

rmlopes:
Seems like there is a problem with Travis. Let me know if I should amend the commit or something to force rebuild.

claudiodsf:
Travis fail is unrelated and is triggering for all the ongoing PR.
No need to do anything here, besides maybe squashing again and write a clearer commit message, as I was suggesting in my previous comment.

krischer:
I just pulled the trigger and merged this before it goes out of sync again :-) Two commits seem very reasonable to me.

krischer:
Thanks a lot @rmlopes!

rmlopes:
Hi guys,

I just noticed while using the new feature that it needs some enhancement. The actual wiggles should not be displayed when using fillcolors (otherwise with many traces the picture gets confusing). 

It is two lines of code, but the test pictures need replacement. Should I make a new feature branch and PR?

QuLogic:
I'm not sure that's necessarily true; it probably depends on the plot's contents, ergo, it should be an optional thing. Thus there should be no problem in adding it to this PR.

megies:
I actually am unsure if github will update this PR or trigger new CI checks on pushes to the branch..

QuLogic:
Oh, I forgot it's already merged; yea, new PR then. GitHub won't update this one, though you're free to keep on the same branch name if you want.

rmlopes:
> I'm not sure that's necessarily true; it probably depends on the plot's contents, ergo, it should be an optional thing.

Yes, my first approach was to eliminate the wiggles if either one of the colors is set. But actually if only the positives are filled tipically the wiggle is drawn. All the images I've seen using two colors (or a gradient) to fill the wiggles draw only the filling.

Anyway, in that case there is no change to do. It is just a matter of setting the linewidth to 0 if needed.

PR #1447, opened 2016-06-09
SLIST/TSPAIR: fix writing files with floating point sampling rates + don't set 'stats.mseed.dataquality' with a blank

megies:
Writing to SLIST and TSPAIR so far wrote sampling rate formatted as an integer into file headers, leading to..
 * raised exceptions for <1 Hz sampling rates, and far worse..
 * plain wrong output files >1 Hz non-integer sampling rates

Although [format definition](http://ds.iris.edu/ds/nodes/dmc/data/formats/simple-ascii/) only shows examples with integer sampling rates and doesn't speak about floating point sampling rates, when [looking at served data](service.iris.edu/irisws/timeseries/1/query?net=IU&sta=ANMO&loc=00&cha=VHZ&starttime=2005-01-01T00:00:00&endtime=2005-01-01T00:01:00&output=ascii), the sampling rate is just written as a float with "`.`":
```
TIMESERIES IU_ANMO_00_VHZ_M, 6 samples, 0.1 sps, 2005-01-01T00:00:07.986000, SLIST, INTEGER, COUNTS
-3643
-3658
-3636
-3618
-3628
-3668
```

Safest thing it to just leave it to Python to format the sampling rate as a string. This might lead to scientific notation in output for very crazy sampling rates but that shouldn't be a problem (at least not for us.. ;-).

Although probably not a lot of people use these formats, this is a **very** bad bug  and I think warrants a 1.0.2 release.

### DIFF
--- a/obspy/io/ascii/core.py
+++ b/obspy/io/ascii/core.py
@@ -44,7 +44,7 @@ from obspy.core import Stats
 from obspy.core.util import AttribDict, loadtxt
 
 
-HEADER = "TIMESERIES %s_%s_%s_%s_%s, %d samples, %d sps, %.26s, %s, %s, %s\n"
+HEADER = "TIMESERIES %s_%s_%s_%s_%s, %d samples, %s sps, %.26s, %s, %s, %s\n"
 
 
 def _is_slist(filename):
--- a/obspy/io/ascii/core.py
+++ b/obspy/io/ascii/core.py
@@ -165,7 +165,10 @@ def _read_slist(filename, headonly=False, **kwargs):  # @UnusedVariable
         stats.channel = temp[3]
         stats.sampling_rate = parts[4]
         # quality only used in MSEED
-        stats.mseed = AttribDict({'dataquality': temp[4]})
+        # don't put blank quality code into 'mseed' dictionary
+        # (quality code is mentioned as optional by format specs anyway)
+        if temp[4]:
+            stats.mseed = AttribDict({'dataquality': temp[4]})
         stats.ascii = AttribDict({'unit': parts[-1]})
         stats.starttime = UTCDateTime(parts[6])
         stats.npts = parts[2]
@@ -230,7 +233,10 @@ def _read_tspair(filename, headonly=False, **kwargs):  # @UnusedVariable
         stats.channel = temp[3]
         stats.sampling_rate = parts[4]
         # quality only used in MSEED
-        stats.mseed = AttribDict({'dataquality': temp[4]})
+        # don't put blank quality code into 'mseed' dictionary
+        # (quality code is mentioned as optional by format specs anyway)
+        if temp[4]:
+            stats.mseed = AttribDict({'dataquality': temp[4]})
         stats.ascii = AttribDict({'unit': parts[-1]})
         stats.starttime = UTCDateTime(parts[6])
         stats.npts = parts[2]
--- a/obspy/io/ascii/core.py
+++ b/obspy/io/ascii/core.py
@@ -44,7 +44,21 @@ from obspy.core import Stats
 from obspy.core.util import AttribDict, loadtxt
 
 
-HEADER = "TIMESERIES %s_%s_%s_%s_%s, %d samples, %s sps, %.26s, %s, %s, %s\n"
+HEADER = ("TIMESERIES {network}_{station}_{location}_{channel}_{dataquality}, "
+          "{npts:d} samples, {sampling_rate} sps, {starttime!s:.26s}, "
+          "{format}, {dtype}, {unit}\n")
+
+
+def _format_header(stats, format, dataquality, dtype, unit):
+    sampling_rate = str(stats.sampling_rate)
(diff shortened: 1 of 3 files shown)


obspy/io/ascii/tests/test_ascii.py
```diff
@@ -475,6 +475,26 @@ def test_write_small_trace(self):
...
 
+    def test_float_sampling_rates_write_and_read(self):
+        """
+        Tests writing and reading Traces with floating point and with less than
+        1 Hz sampling rates.
+        """
+        tr = Trace(np.arange(10))
+        check_sampling_rates = (0.000000001, 1.000000001, 100.000000001,
+                                99.999999999, 1.5, 1.66666666, 10000.0001)
+        for format in ['SLIST', 'TSPAIR']:
```
barsch: you are testing double precision values and not float - http://ds.iris.edu/ds/nodes/dmc/data/formats/simple-ascii/ does not speak about double - but it wouldn't hurt to add it as additional datatype

megies: I don't care too much what kind of string format is allowed by the format (as you mentioned it is extremely vague and doesn't say anything about it). But at least the header states the correct sampling rate now. ;-)

megies:
Wrong base branch.. -_- will merge into `maintenance` by hand when CI gives the nod.

obspy/io/ascii/core.py
```diff
@@ -44,7 +44,7 @@
 from obspy.core.util import AttribDict, loadtxt
 
 
-HEADER = "TIMESERIES %s_%s_%s_%s_%s, %d samples, %d sps, %.26s, %s, %s, %s\n"
+HEADER = "TIMESERIES %s_%s_%s_%s_%s, %d samples, %s sps, %.26s, %s, %s, %s\n"
```
barsch: any reason not to use `"%lg"` instead (mssed2ascii.c uses it for writing) - so the existing examples would be still correct

megies: What's `%lg` and where is it documented?

barsch: http://www.cplusplus.com/reference/cstdio/printf/

megies: Does it make any difference for Pythons string formatting?

barsch:
```
In [1]: "%lg" % (12.123)
Out[1]: '12.123'

In [2]: "%lg" % (12.0)
Out[2]: '12'
```

so essentially writing "20 sps" instead of "20.0 sps" as in the given examples of IRIS -also  thats what IRIS does in its mseed2ascii scripts - and we should try to stay close to that as they defined that format and who knows which import program assume an integer an this position ... anyway up to you - just my 2 cents

megies:
```
In [6]: "%lg" % (12.00000001)
Out[6]: '12'

In [7]: "%s" % (12.00000001)
Out[7]: '12.00000001'
```

barsch: In [7]: "%s" % (12.0000000001)
Out[7]: '12.0000000001'

In [8]: "%s" % (12.00000000001)
Out[8]: '12.0'

barsch: how about:

```
In [18]: ('%s' % 12.0000000001).rstrip('0').rstrip('.')
Out[18]: '12.0000000001'

In [19]: ('%s' % 12.0000000000).rstrip('0').rstrip('.')
Out[19]: '12'
```

QuLogic: In case it was unclear earlier, `%lg` is the same as `%g` in Python:

> A length modifier (h, l, or L) may be present, but is ignored as it is not necessary for Python – so e.g. %ld is identical to %d.

krischer:
We already had a couple bugs due to many tests being lazily written with a sampling rate of 1.0. I think we should add another test that loops over all waveform plugins that can read + write and performs a small round trip attempting to preserve funny sampling rates and times.

obspy/io/ascii/core.py
megies: > how about:
> 
> In [18]: ('%s' % 12.0000000001).rstrip('0').rstrip('.')
> Out[18]: '12.0000000001'
> 
> In [19]: ('%s' % 12.0000000000).rstrip('0').rstrip('.')
> Out[19]: '12'

I've implemented something along that line (this was not safe.. consider `1200` -> `'12'`). I would have been OK with just plain `%s`, hope we didn't miss an edge case were the stripping is doing harm.

megies:
Added another small fix with b783ad2, don't think it's worth to have another issue for it..

megies:
Good to merge, I think. If nobody objects in the next couple of days, I'll merge this one to `maintenance_1.0.x` on the command line (wrong base branch selected in the PR..)

megies:
Merged into `maintenance_1.0.x` with 10a807ad9288a89cf81af03d237c103e24edee9d.

PR #1448, opened 2016-06-09
GSE2: make sure network code is written/read

megies:
..even if some other details like latitude can not be written/parsed to/from STA2 header line (e.g. because they are not found in trace.stats.gse)

a.k.a. "A missing `trace.stats.gse` must not prevent the network code being written to file correctly."

### DIFF
--- a/obspy/io/gse2/libgse2.py
+++ b/obspy/io/gse2/libgse2.py
@@ -465,13 +465,32 @@ def parse_sta2(line):
     """
     header = {}
     try:
+        lat = line[15:24].strip()
+        if lat:
+            lat = float(lat)
+        else:
+            lat = None
+        lon = line[25:35].strip()
+        if lon:
+            lon = float(lon)
+        else:
+            lon = None
+        elev_edepth = line[48:].strip().split()
+        elev, edepth = elev_edepth or (None, None)
+        if elev:
+            elev = float(elev)
+        else:
+            elev = None
+        if edepth:
+            edepth = float(edepth)
+        else:
+            edepth = None
         header['network'] = line[5:14].strip()
-        header['lat'] = float(line[15:24])
-        header['lon'] = float(line[25:35])
+        header['lat'] = lat
+        header['lon'] = lon
         header['coordsys'] = line[36:48].strip()
-        elev, edepth = line[48:].strip().split()
-        header['elev'] = float(elev)
-        header['edepth'] = float(edepth)
+        header['elev'] = elev
+        header['edepth'] = edepth
     except:
         msg = 'GSE2: Invalid STA2 header, ignoring.'
         warnings.warn(msg)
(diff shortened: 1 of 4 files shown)


obspy/io/gse2/libgse2.py
```diff
@@ -465,13 +465,33 @@ def parse_sta2(line):
...
+        if lat:
+            lat = float(lat)
+        else:
+            lat = None
+        lon = line[25:35].strip()
+        if lon:
+            lon = float(lon)
+        else:
+            lon = None
+        lon = lon and float(lon)
```
QuLogic: But you already took the `float` of it on 475?

obspy/io/gse2/libgse2.py
```diff
@@ -465,13 +465,33 @@ def parse_sta2(line):
...
+        elev_edepth = [x.strip() for x in line[48:].strip().split()]
```
QuLogic: No need to `.strip()` if you're going to `.split()` since it compresses whitespace anyway.

obspy/io/gse2/libgse2.py
```diff
@@ -465,13 +465,33 @@ def parse_sta2(line):
     """
     header = {}
     try:
+        lat = line[15:24].strip()
```
QuLogic: I'd just do:
```python
try:
    header['lat'] = float(line[15:24])
except ValueError:
    header['lat'] = None
```

QuLogic: But maybe you specifically want it to fail for something like `line[15:24] == 'abcdef'`?

obspy/io/gse2/libgse2.py
```diff
@@ -485,26 +505,33 @@ def compile_sta2(stats):
...
+    lon = stats['gse2'].get('lon')
+    coordsys = stats['gse2'].get('coordsys')
     line = fmt1 % (
         stats['network'],
-        stats['gse2']['lat'],
-        stats['gse2']['lon'],
-        stats['gse2']['coordsys'])
+        lat is not None and '{:9.5f}'.format(lat) or '',
+        lon is not None and '{:10.5f}'.format(lon) or '',
+        coordsys is not None and coordsys or '')
```
QuLogic: `coordsys` is a `str`, right? So this is the same as `coordsys or ''`?

krischer: Welcome back :-) We missed your reviews! ;-)

obspy/io/gse2/libgse2.py
megies: Yeah, that was my line of thought. I don't want to turn invalid entries into `None`, just blank entries.

In a quick search, couldn't find what GSE manual says about blank entries. Real world files seem to have invalid values -- like `-999.00` for the latitude -- in case of missing values, to the extent of breaking the fixed format even some times. I think for a fixed format blanks make the most sense, so I decided to go with that -- in writing and thus obviously in reading, too.

obspy/io/gse2/libgse2.py
megies: Thanks. Left-over from shoving things around. Initially wanted to avoid the ugly and longish `if/else`s. But as it turns out..

```
lon = "0"
lon and float(lon) or None
```

.. evaluates to `None`, which surprised me a bit.

megies:
Good to merge, I think.

megies:
Squashed, should be good to go.

PR #1449, opened 2016-06-09
Add Güralp Compressed File (GCF) read support

rannof:
I think some changes should also be made to the core.base module, but I'm not sure exactly what.


rannof:
Fixed code according to Flake8.

obspy/io/gcf2mseed.py
```diff
@@ -0,0 +1,47 @@
...
+parser.add_argument('gcf_file',nargs='+',type=str,help='input file(s) name.')
+parser.add_argument('-O',metavar='option',type=str,nargs='*',default=["'MSEED'"],help='''\
+                         Options for the obspy write command (default: MSEED.)
+                         Example option:
+                         -O "format='MSEED'" encoding=11 reclen=512 byteorder=1.
+                         Note the " and ' around the format and MSEED values.'''
+                    )
+
+def getStream(FileList):
+  st = Stream()
```
QuLogic: This file definitely did not go through `flake8`.

QuLogic:
Isn't it Güralp, not Guralpe?

obspy/io/gcf2mseed.py
rannof: Yep. Sorry. Forgot all about it.

On Thu, Jun 9, 2016 at 2:46 PM, Elliott Sales de Andrade <
notifications@github.com> wrote:

> In obspy/io/gcf2mseed.py
> <https://github.com/obspy/obspy/pull/1449#discussion_r66527854>:
>
> > +             http://www.guralp.com/apps/ok?doc=GCF_Intro
> > +
> > +             The script only converts data streams
> > +
> > +             Created by Ran Novitsky Nof @ BSL, 2016'''))
> > +parser.add_argument('-o',metavar='Output_file',type=str,help='Output Miniseed file name.')
> > +parser.add_argument('gcf_file',nargs='+',type=str,help='input file(s) name.')
> > +parser.add_argument('-O',metavar='option',type=str,nargs='*',default=["'MSEED'"],help='''\
> > +                         Options for the obspy write command (default: MSEED.)
> > +                         Example option:
> > +                         -O "format='MSEED'" encoding=11 reclen=512 byteorder=1.
> > +                         Note the " and ' around the format and MSEED values.'''
> > +                    )
> > +
> > +def getStream(FileList):
> > +  st = Stream()
>
> This file definitely did not go through flake8.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449/files/974a93d3f7083dd0f186664e2bbff4b7dbe9c22e#r66527854>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AKICcm6ZEnGiFP9OxqMpTyiVMz8upwgAks5qKIm-gaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

obspy/io/gcf/__init__.py
```diff
@@ -0,0 +1,30 @@
+"""
+obspy.io.gcf - Guralpe Compressed Format read support for ObsPy
+==========================================================
```
QuLogic: This heading is not long enough.

rannof: What do you mean?

On Thu, Jun 9, 2016 at 2:48 PM, Elliott Sales de Andrade <
notifications@github.com> wrote:

> In obspy/io/gcf/__init__.py
> <https://github.com/obspy/obspy/pull/1449#discussion_r66528115>:
>
> > @@ -0,0 +1,30 @@
> > +"""
> > +obspy.io.gcf - Guralpe Compressed Format read support for ObsPy
> > +==========================================================
>
> This heading is not long enough.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449/files/974a93d3f7083dd0f186664e2bbff4b7dbe9c22e#r66528115>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AKICcoD_nPlms6HXucqjvDMhhdKo2eMEks5qKIoigaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

rannof:
Fix for Güralp name, Flake8 fixes.

obspy/io/gcf/__init__.py
QuLogic: The `=` are not the same length as the text.

obspy/io/gcf/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+This module provides read support for GCF waveform data and header info.
+Most methods are based on info from Güralp site
+http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/GCF_Specification.htm
+
+:copyright:
+    The ObsPy Development Team (devs@obspy.org) & Ran Novitsky Nof
+:license:
+    GNU Lesser General Public License, Version 3
+    (https://www.gnu.org/copyleft/lesser.html)
```
QuLogic: Missing blank line after this.

obspy/io/gcf/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+Reading
+-------
+Similar to reading any other waveform data format using :mod:`obspy.core`:
+>>> from obspy import read
+>>> st = read("/path/to/datafile.gcf")
+The format will be determined automatically.
+-------
```
QuLogic: The previous line is not a heading; I'm not sure what these dashes are for.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,59 @@
...
+    :param headonly: If True read only head of GCF file.
+    :rtype: :class:`~obspy.core.stream.Stream`
+    :returns: Stream object containing header and data.
+    .. rubric:: Example
+    >>> from obspy import read
+    >>> st = read("/path/to/filename.gcf")
+    """
+    traces = []
+    with open(filename, 'rb') as f:
+        # reading multiple gse2 parts
```
QuLogic: This is not gse2.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,59 @@
...
+
+def _is_gcf(filename):
+    """
+    Checks whether a file is GCF or not.
+    :type filename: str
+    :param filename: GCF file to be checked.
+    :rtype: bool
+    :return: ``True`` if a GCF file.
+    """
+    # Open file.
```
QuLogic: This comment doesn't really say anything.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python
```
QuLogic: This file is not intended to be run.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
...
+# Based on Guralp's GCF reference (GCF-RFC-GCFR, Issue C, 2011-01-05)
+# more details available from: http://www.guralp.com/apps/ok?doc=GCF_Intro
+# last access: June, 2016
+
+
+import numpy as np
+from obspy import UTCDateTime
+
+SPSD = {  # Table 3.1: special sample rates
+        157: 0.1,
```
QuLogic: A single indent is fine.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
...
+# By Ran Novitsky Nof @ BSL, 2016
+# ran.nof@gmail.com
+# Based on Guralp's GCF reference (GCF-RFC-GCFR, Issue C, 2011-01-05)
+# more details available from: http://www.guralp.com/apps/ok?doc=GCF_Intro
+# last access: June, 2016
+
+
+import numpy as np
+from obspy import UTCDateTime
+
```
QuLogic: Two blank lines here.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
...
+bitsD = {  # Table 3.2: format field to data bytes
+        1: 4,  # 32bits
+        2: 2,  # 16bits
+        4: 1   # 8bits
+        }
+
+
+def is_gcf(f):
+    'Test if file is GCF by reading at least 1 data block'
+    while 1:
```
QuLogic: `while True`

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
...
+        DATA = DATA / 36
+    return STR
+
+
+def decodeTime(DATA):
+    'Decode time field'
+    H = DATA / 3600
+    M = (DATA % 3600) / 60
+    S = DATA % 60
+    return "%02d:%02d:%02d" % (H, M, S)
```
QuLogic: You should just create a `UTCDateTime` directly from the components instead of making a string out of them.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def isLeapYear(Year):
+    return (Year % 4 == 0) and ((Year % 100 != 0) or (Year % 400 == 0))
+
+
+def decodeDate(days):
```
QuLogic: Instead of trying to calculate leap years and such yourself, you should be able to pass the number directly (you would need to divide seconds into fractional days) and then offset by the difference between epochs.

obspy/io/gcf/libgcf.py
rannof: Thanks for all the help and sorry for all the work. 
Please, can you explain your idea? give a short example?
Cheers,
Ran

QuLogic: See example 4 of [the UTCDateTime constructor](https://docs.obspy.org/packages/autogen/obspy.core.utcdatetime.UTCDateTime.html).

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    return "%4d-%02d-%02d" % (year, month + 1, days + 1)
+
+
+def decodeDateTime(data):
+    '''Decode date and time field.
+    The date code is a 32 bit value specifying the start time of the block.
+    Bits 0-16 contain the number of seconds since midnight,
+    and bits 17-31 the number of days since 17th November 1989.'''
+    DATE = decodeDate(data >> 17)
+    TIME = decodeTime(data - (data >> 17 << 17))
```
QuLogic: This should be the same as `data & 0xffff`.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    starttime = UTCDateTime(DATE + 'T' + TIME)
+    return starttime
+
+
+def readDataBlock(f, skipData=False):
```
QuLogic: Why not call this `headonly` like in the other function?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/GCF_Specification.htm
+    f - file object to read from
+    if skipData is True, Only header is returned.
+    if not a data block (SPS=0) - returns None.
+    '''
+    # get ID
+    ID = f.read(4)
+    if ID == '':
+        raise EOFError  # got to EOF
+    ID = np.frombuffer(ID, count=1, dtype=np.uint32).byteswap()
```
QuLogic: This assumes you are running on a little-endian machine; you should pass the file's byte order to the `dtype`.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    if ID >> 31 & 0b1 > 0:
+        ID = (ID << 6) >> 6
+    ID = decode36(ID)
+    # get Stream ID
+    stID = f.read(4)
+    stID = np.frombuffer(stID, count=1, dtype=np.uint32).byteswap()
```
QuLogic: You can just use `np.fromfile` instead of reading to a buffer and then reading into an `ndarray`.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    stID = decode36(stID)
+    # get Date & Time
+    data = f.read(4)
+    data = np.frombuffer(data, count=1, dtype=np.uint32).byteswap()
+    DATE = decodeDate(data >> 17)
```
QuLogic: This can be `decodeDateTime` here?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+    if not SPS:
+        f.seek(N*4, 1)  # skip if not a data block
+        if 1008-N*4 > 0:
+            f.seek(1008-N*4, 1)  # keep skipping to get 1008 record
+        return None
+    NoS = int(N)*compression  # number of samples
+    header = {}
+    header['starttime'] = starttime
+    header['station'] = stID[:4]
+    header['channel'] = 'HH'+stID[4].capitalize()
```
QuLogic: Is it really always `HH`?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,182 @@
...
+        # get FIC
+        FIC = np.frombuffer(f.read(4), count=1, dtype='>i4')
+        # get incremental data
+        data = np.frombuffer(f.read(4*N), count=NoS,
+                             dtype=compressionD[compression])
+        # construct time series
+        data = (FIC + np.cumsum(data)).astype('>i4')
+        # get RIC
+        RIC = np.frombuffer(f.read(4), count=1, dtype='>i4')
+        # skip if not a full record
```
QuLogic: This should either be an error, or raise a warning.

obspy/io/gcf2mseed.py
```diff
@@ -0,0 +1,55 @@
...
+
+# convert GCF files to miniseed
+# by Ran Nof @ BSL, 2016
+# ran.nof@gmail.com
+
+import argparse
+import sys
+import textwrap
+
+from gcf.core import _is_gcf as is_gcf, _read_gcf as read_gcf
```
QuLogic: `gcf` is not a top-level package.

obspy/io/gcf/libgcf.py
rannof: What about really big files?

obspy/io/gcf2mseed.py
```diff
@@ -0,0 +1,55 @@
...
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description='''Convert Güralp Compressed Format (GCF) to Miniseed''',
+    epilog=textwrap.dedent('''\
+             See more details regarding GCF:
+             http://www.guralp.com/apps/ok?doc=GCF_Intro
+
+             The script only converts data streams
+
+             Created by Ran Novitsky Nof @ BSL, 2016'''))
+parser.add_argument('-o', metavar='Output_file', type=str,
```
QuLogic: You should add a long-form of the argument instead of a `metavar`.

obspy/io/gcf/libgcf.py
rannof: There is no channel field in the data itself. I can add a code to guess the appropriate channel code, but that might introduce problems.

obspy/io/gcf/libgcf.py
rannof: I don't think so. Sometimes data is not filling the whole packet size and that is OK.

obspy/io/gcf/libgcf.py
QuLogic: Actually, offset first, then pass the number directly to the constructor.

obspy/io/gcf/libgcf.py
rannof: Yes. I'll change the datetime conversion using your suggestion for UTCDateTime directly. Is there a way to add days to a UTCDateTime object? I know I can add the seconds simply by +

QuLogic:
Unless you are trying to follow some spec really closely, please use PEP8 naming for identifiers: functions and variable are `snake_case`, constants are `ALL_CAPS`, and classes are `CamelCase`.

QuLogic:
You should add your module [here](https://github.com/obspy/obspy/blob/master/obspy/core/util/base.py#L35) and also add it to the format auto-detection list further down.

I'm not sure whether or not we should keep the script. No other format has one like it, so I'll let others weigh in on that. But you should add the script [to `setup.py`](https://github.com/obspy/obspy/blob/master/setup.py#L121) and add the format to the list in `setup.py` just a little bit further down.

If you can provide a small (a few KiB) file for a test, that would be good as well.

obspy/io/gcf/libgcf.py
QuLogic: I don't understand your question.

obspy/io/gcf/libgcf.py
QuLogic: Convert to seconds by multiplying by 86400 first; there's no direct way, I don't think.

obspy/io/gcf/libgcf.py
QuLogic: Oh, OK, I misunderstood the comment; can you add a little clarification there.

obspy/io/gcf/libgcf.py
rannof: No. It is not.

QuLogic: Oh, that should be `0x1ffff`.

obspy/io/gcf/libgcf.py
megies: :+1: Please definitely rename this option to `headonly` across the complete PR to be in line with other waveform plugins in obspy.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    traces = []
+    with open(filename, 'rb') as f:
+        # reading multiple gcf parts
+        while True:
+            try:
+                if headonly:
+                    header = libgcf.read_header(f)
+                    if header:
+                        traces.append(Trace(header=header))
```
megies: I think that after assembling the "headonly" Trace, you need to set `trace.stats.npts` explicitly, compare https://github.com/obspy/obspy/blob/f997d39519517e27d71aaffca64177298cf866be/obspy/io/mseed/core.py#L414-L420. Otherwise `stats.npts` is likely getting set to `0` (by the empty array during `__init__()`. Please check..

obspy/io/gcf/libgcf.py
megies: We have some mapping (the other way around) in `obspy/core/util/misc.py` `BANDCODE` (but that would still leave the ambiguity between short period and broad band). In any case this should be accessible to the user.. I would propose something like `read(..., channel_prefix="HH")` that gets passed all the way down to here.

obspy/io/gcf/libgcf.py
rannof: Yep.

obspy/io/gcf2mseed.py
rannof: Can You provide me with an example?

megies:
> I'm not sure whether or not we should keep the script. No other format has one like it, so I'll let others weigh in on that

I'm also tending to drop the script.. @rannof is it of big importance to you to keep it?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,145 @@
...
+    2: 2,  # 16bits
+    4: 1}   # 8bits
+
+
+def is_gcf(f):
+    'Test if file is GCF by reading at least 1 data block'
+    while True:
+        header, data = read(f)
+        if len(data):
+            break
```
megies: Can we drop the `while` here? A valid GCF file should start with one of these blocks, right? So that's a valid assumption for the autodetection. Also `read(f)` is reading the whole file (while the comment seems to indicate that only a small part of the file is checked) -- did you intend to use `read_data_block()` instead?

setup.py
```diff
@@ -161,7 +161,8 @@
         'SH_ASC = obspy.io.sh.core',
         'WAV = obspy.io.wav.core',
         'AH = obspy.io.ah.core',
-        'KNET = obspy.io.nied.knet'
+        'KNET = obspy.io.nied.knet',
+        'GCF = obspy.io.gcf.core
```
megies: Invalid syntax, misses closing single quote.

setup.py
```diff
@@ -267,6 +268,10 @@
         'isFormat = obspy.io.nied.knet:_is_knet_ascii',
         'readFormat = obspy.io.nied.knet:_read_knet_ascii',
         ],
+    'obspy.plugin.waveform.GCF': [
+        'isFormat = obspy.io.gcf.core:_is_gcf',
+        'readFormat = obspy.io.gcf.core:_read_gcf',
+        ]
```
megies: Invalid syntax, misses comma at end.

obspy/core/util/base.py
```diff
@@ -38,7 +38,7 @@
                    'io.datamark', 'io.gse2', 'io.json', 'io.kinemetrics',
                    'io.kml', 'io.mseed', 'io.ndk', 'io.nied', 'io.nlloc',
                    'io.pdas', 'io.pde', 'io.quakeml', 'io.sac', 'io.seg2',
-                   'io.segy', 'io.seisan', 'io.sh', 'io.shapefile',
+                   'io.segy', 'io.seisan', 'io.sh', 'io.shapefile','io.gcf,
```
megies: Invalid syntax, misses closing single quote.

megies:
At the end of reading a file and before returning the stream you should do a `st.merge(-1)` to put together seamless traces:
```python
In [1]: st = read("/path/to/20160603_1955n.gcf")

In [2]: st
Out[2]: 
104 Trace(s) in Stream:

.6018..HHN | 2016-06-03T19:55:00.000000Z - 2016-06-03T19:55:01.990000Z | 100.0 Hz, 200 samples
...
(102 other traces)
...
.6018..HHN | 2016-06-03T19:59:59.000000Z - 2016-06-03T19:59:59.990000Z | 100.0 Hz, 100 samples

In [3]: st.merge(-1)

In [4]: st
Out[4]: 
1 Trace(s) in Stream:
.6018..HHN | 2016-06-03T19:55:00.000000Z - 2016-06-03T19:59:59.990000Z | 100.0 Hz, 30000 samples
```

Unfortunately that won't work together with the `headonly` option... (see #617) you would have to put the individual pieces together on a lower level, i.e. put pieces together before and then set up one long trace for all pieces.
Would it be a lot of work to check during reading the single blocks if they fit seamless to the previous one and only set up a single trace for a series of gapless blocks (that might also be more ideal for the above case)?

Otherwise, looks pretty good already. :+1:

claudiodsf:
I'm also against keeping the script since it is too specific. 

It would be nice however to have a `obspy-convert` (or `obspy convert` if we do not want to clutter the bin directory) to perform conversions from any format to any format (when this makes sense and if the destination format has write support, of course).

rannof:
Actually, I don't like the obspy automatic merge behavior. I had a specific
need for keeping the original packet order for playbacks and had to build
by own read function for mseed files.
Merging a gcf file might not be a problem since each file is one trace.
However, if this module will be used for connecting to a scream server
(instead of seedlink server) to acquire real time data, that might not be
the case. Also, I can imagine cases where someone is just saving packets
from server as they arrive into one file with mixed packets of different
channels. I understand that is the default behavior of obspy, and should
stay coherent with other formats. I thought of this solution:

    if not headonly:
        return Stream(traces=traces).merge(-1)
    else:
        return Stream(traces=traces)



On Sat, Jun 11, 2016 at 2:38 AM, Tobias Megies <notifications@github.com>
wrote:

> At the end of reading a file and before returning the stream you should do
> a st.merge(-1) to put together seamless traces:
>
> In [1]: st = read("/path/to/20160603_1955n.gcf")
>
> In [2]: st
> Out[2]: 104 Trace(s) in Stream:
> .6018..HHN | 2016-06-03T19:55:00.000000Z - 2016-06-03T19:55:01.990000Z | 100.0 Hz, 200 samples...
> (102 other traces)....6018..HHN | 2016-06-03T19:59:59.000000Z - 2016-06-03T19:59:59.990000Z | 100.0 Hz, 100 samples
>
> [Use "print(Stream.__str__(extended=True))" to print all Traces]
>
> In [3]: st.merge(-1)
>
> In [4]: st
> Out[4]: 1 Trace(s) in Stream:.6018..HHN | 2016-06-03T19:55:00.000000Z - 2016-06-03T19:59:59.990000Z | 100.0 Hz, 30000 samples
>
> Unfortunately that won't work together with the headonly option... (see
> #617 <https://github.com/obspy/obspy/pull/617>) you would have to put the
> individual pieces together on a lower level, i.e. put pieces together
> before and then set up one long trace for all pieces.
> Would it be a lot of work to check during reading the single blocks if
> they fit seamless to the previous one and only set up a single trace for a
> series of gapless blocks (that might also be more ideal for the above case)?
>
> Otherwise, looks pretty good already. 👍
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449#issuecomment-225347807>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AKICcvwEcgAWsn1rnbS6cXUMcz4Ce0Xsks5qKoIWgaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

rannof:
No need to keep. I remove it.

On Sat, Jun 11, 2016 at 2:11 AM, Tobias Megies <notifications@github.com>
wrote:

> I'm not sure whether or not we should keep the script. No other format has
> one like it, so I'll let others weigh in on that
>
> I'm also tending to drop the script.. @rannof <https://github.com/rannof>
> is it of big importance to you to keep it?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449#issuecomment-225346199>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AKICckWUITDhBjq1o8inaWrRD1HJkLO5ks5qKnu5gaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

megies:
> Actually, I don't like the obspy automatic merge behavior. I had a specific need for keeping the original packet order for playbacks and had to build by own read function for mseed files. Merging a gcf file might not be a problem since each file is one trace. However, if this module will be used for connecting to a scream server (instead of seedlink server) to acquire real time data, that might not be the case. Also, I can imagine cases where someone is just saving packets from server as they arrive into one file with mixed packets of different channels. I understand that is the default behavior of obspy, and should stay coherent with other formats. I thought of this solution:

Well, I think that's two end members of use cases.. data processing/analysis vs. data housekeeping. So far the main focus of obspy was always the analysis side and I think 90% of users mainly do processing and do not care much about things like record-level quality flags or packet order but rather want to have convenient, large contiguous chunks to work with in data analysis. And our support for more specific operations or introspection on lower (small packet) level is admittedly very limited (partly because those details are different for every format, I think).

Of course, having more lower level capabilities for special use cases is totally fine. For special operations on a specific format, it might be necessary to have an additional format-specific construct, on a lower level than `Trace`, compare e.g. `SACTrace`, `SEGYTrace`/`SEGYFile`.. which is totally valid.
Just, what gets plugged in into the high-level `read()` routine should assemble contiguous chunks together.

misc/docs/source/packages/index.rst
```diff
@@ -51,6 +51,7 @@ The functionality is provided through the following packages:
    obspy.io.sh
    obspy.io.wav
    obspy.io.y
+   obspy.io.gcf
```
megies: Sphinx doesn't reorder such lists, please keep it in alphabetical order.

rannof:
I can't pass the tests. Any idea why? I removed the >>> read("/path/to/filename.gcf") line is the docstring, but I don't think that should be the solution.

megies:
> I can't pass the tests. Any idea why?

Looks like `future` is making numpy's `fromfile` fail (also, I remember dtypes being specified as strings have been problematic in the past, just keep in mind in case there's other strange fails along the way). Not sure how to best work around it.. you could replace, e.g. ..
```
# sysid = np.fromfile(f, count=1, dtype='>u4')
sysid = np.frombuffer(f.read(4), count=1, dtype='>u4')
```
(or rather use `from_buffer` from `obspy.core.util.compatibility`)
..but there might be a easier/cleaner/better solution..?

QuLogic:
Is that really the reason? AppVeyor seems alright; Travis failed for unrelated reasons. Can you add another commit or rebase to trigger a re-build? I think this failure was fixed on master.

rannof:
@QuLogic, How do I rebase? Is it possible to do it online?

QuLogic:
The [git book](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) explains it pretty well. You might have a little trouble because you did not start a feature branch. Instead of (as in the book) `experimental` and `master`, you would have `master` and `upstream/master`, respectively (assuming you have a remote for this repository named `upstream`, and you ran `git fetch upstream`).

rannof:
Hi,
Thanks. I'm sorry, but it's all very new to me. I had the github fork it
for me as soon as I tried to edit one of the files on obspy/obspy. also, I
did all the changes online, so no git fetch etc. I have no idea if there is
an upstream set in that way. Any advice on how to proceed?

On Thu, Jun 16, 2016 at 5:11 PM, Elliott Sales de Andrade <
notifications@github.com> wrote:

> The git book <https://git-scm.com/book/en/v2/Git-Branching-Rebasing>
> explains it pretty well. You might have a little trouble because you did
> not start a feature branch. Instead of (as in the book) experimental and
> master, you would have master and upstream/master, respectively (assuming
> you have a remote for this repository named upstream, and you ran git
> fetch upstream).
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449#issuecomment-226648992>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AKICcjXVMKUSmtuSzmboYIAT6E35D0iBks5qMeYugaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

megies:
No problem at all, git can be a monster at the start.. ;-)

Are you working locally on your computer now? You can add a new remote with `$ git remote add rannof https://github.com/obspy/obspy.git` (or if you have an SSH key set up use the github SSH URL) and then checkout your fork's `master` branch to a new branch like.. `$ git branch --track gcf rannof/master; git checkout gcf`

You can also join the [Gitter chat room](https://gitter.im/obspy/obspy) if you need more detailed instructions, I'm sure also one of the US guys can help while Europe is fast asleep.. ;-)

rannof:
Sorry, didn't work for me. I have no local clone on my computer. I only
edited on the github online. Can anyone guid me on how to rebase so I can
commit changes to the obspy pull request?

I tried:
> git clone http://github.com/rannof/obspy
> git add remote rannof https://github.com/obspy/obspy.git
> git remote -v
origin http://github.com/rannof/obspy (fetch)
origin http://github.com/rannof/obspy (push)
rannof https://github.com/obspy/obspy.git (fetch)
rannof https://github.com/obspy/obspy.git (push)
> git branch --track gcf rannof/master
error: the requested upstream branch 'rannof/master' does not exist
hint:
hint: If you are planning on basing your work on an upstream
hint: branch that already exists at the remote, you may need to
hint: run "git fetch" to retrieve it.
hint:
hint: If you are planning to push out a new local branch that
hint: will track its remote counterpart, you may want to use
hint: "git push -u" to set the upstream config as you push.


On Thu, Jun 16, 2016 at 11:30 PM, Tobias Megies <notifications@github.com>
wrote:

> No problem at all, git can be a monster at the start.. ;-)
>
> Are you working locally on your computer now? You can add a new remote
> with $ git remote add rannof https://github.com/obspy/obspy.git (or if
> you have an SSH key set up use the github SSH URL) and then checkout your
> fork's master branch to a new branch like.. $ git branch --track gcf
> rannof/master; git checkout gcf
>
> You can also join the Gitter chat room <https://gitter.im/obspy/obspy> if
> you need more detailed instructions, I'm sure also one of the US guys can
> help while Europe is fast asleep.. ;-)
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449#issuecomment-226690715>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AKICcvOAp1DXpd5VFYnX37eNqeZrzRxoks5qMj7sgaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

megies:
```
git fetch --all
git branch --track gcf origin/master  # as you set it up, "origin" is your remote alias to your fork at http://github.com/rannof/obspy
git checkout gcf
```

krischer:
@rannof If it still does not work, just give any of us push rights to your fork (in the settings tab at http://github.com/rannof/obspy) and we can do it for you :-)

Regarding the merging behavior: I would also do as @megies suggested and manually merge them within the module so it also works with the `headonly` variant. You could still add a `merge_packets=True` argument or something along this line to the `read_..()` method selectively turn the merging off if desired for some use cases.

rannof:
I added the people who contributed comments and assistance as collaborators on rannof/obspy, I have no clue on how to solve this. If anyone wishes to be added as well, let me know. If you manage to solve the merge issue, I would be happy to know what steps where done to accomplish that.
Cheers,
Ran

krischer:
I rebased the branch on top of the latest master and force pushed here. 

> If you manage to solve the merge issue, I would be happy to know what steps where done to accomplish that.

I just did `git rebase master` (I called your branch differently - I guess most of your problems just arise from the fact that you worked on the master branch) and then fixed the one conflict.

The easiest way to get yourself to the state of this branch is now to just delete your existing obspy repository and 

```bash
git clone https://github.com/obspy/obspy
git add remote rannof https://github.com/rannof/obspy.git
git fetch rannof
git checkout -t rannof/master -b gcf
```

Now you have a branch named `gcf`. Work on it and later push with

```bash
git push rannof gcf:master
```

In the future best just work on a feature branch, e.g.

```
git checkout master
git pull
git checkout -b some-feature
```

which makes life a lot easier ;-)

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,146 @@
...
+        s = chr(imed + 48) + s
+        data = data / 36
+    return s
+
+
+def decode_date_time(data):
+    '''Decode date and time field.
+    The date code is a 32 bit value specifying the start time of the block.
+    Bits 0-16 contain the number of seconds since midnight,
+    and bits 17-31 the number of days since 17th November 1989.'''
```
krischer: Can you change all these docstrings to

```python
def decode_date_time(...):
    """
    ...
    """
```

which is consistent with the rest of ObsPy and also what you did in `core.py`.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,70 @@
...
+    return True
+
+
+def _read_gcf(filename, headonly=False, **kwargs):  # @UnusedVariable
+    """
+    Reads a GCF file and returns a Stream object.
+    only GCF files containing data records are supported.
+    .. warning::
+        This function should NOT be called directly, it registers via the
+        ObsPy :func:`~obspy.core.stream.read` function, call this instead.
```
krischer: You'll have to add empty lines before and after the warning block.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,70 @@
...
+        with open(filename, 'rb') as f:
+            libgcf.is_gcf(f)
+    except:
+        return False
+    return True
+
+
+def _read_gcf(filename, headonly=False, **kwargs):  # @UnusedVariable
+    """
+    Reads a GCF file and returns a Stream object.
```
krischer: Also add an empty line after the first line in a docstring.

krischer: See here: https://www.python.org/dev/peps/pep-0257/

Except that we also always put the triple quotes on separate lines.

krischer:
Regarding the PR itself: 

* It still has a couple of formatting issues  - I think its only the docstrings.
* It needs one or two actual tests.
* Can you make the test file smaller? I think just one or two records/blocks should be enough.

Otherwise this looks pretty good :-)

QuLogic:
Just to be clear:

> The easiest way to get yourself to the state of this branch is now to just delete your existing obspy repository and

Delete the repository on your local computer, not the fork here on GitHub.

> In the future best just work on a feature branch

Alas, you cannot change a PR target/source, so "future" means for any new PR you open.

krischer:
screen shot 2016-06-18 at 12 57 45

megies:
Trigger docs build.. (see 43c6678) +DOCS

rannof:
I removed my repo on my computer and:
```
gitt clone https://github.com/obspy/obspy
git remote add rannof https://github.com/rannof/obspy.git
git fetch rannof
git checkout -t rannof/master -b gcf
```
Edited some files
```
git commit -a -m "commit string"
git push rannof gcf:master
```
Now I get the codecov-io error. Have no idea what to do next. Also, have no idea on how to create the test cods.

megies:
> Now I get the codecov-io error. Have no idea what to do next.

Don't worry too much about the codecov results right now.. there's no real tests yet, so no wonder the code lines you added in `io.gcf` module are not covered yet..

> Also, have no idea on how to create the test cods.

Have a look at e.g. https://github.com/obspy/obspy/tree/master/obspy/io/datamark/tests, you can copy/paste and adapt from there.

rannof:
Failing test. The code does not recognize the format automatically. There must be something wrong with the way the test code is looking for the test file. It worked for me when using the gcf module as a standalone. Any suggestions?

megies:
Can you also please test the data values of the parsed file (e.g. the first 20 samples and the last 20 samples) [here](https://github.com/obspy/obspy/pull/1449/files#diff-62deca20cd59089cbe163f2b8f5a034aR26)?
Something like..
```
expected = np.array([1, 5, 4, 3, ...])
np.testing.assert_array_equal(expected, st[0].data[:20])
```
Oh, and a short changelog line please.. https://github.com/obspy/obspy/blob/master/CHANGELOG.txt#L31


Otherwise looks really good to me by now. :+1:

rannof:
@megies, OK, next commit will include that. 

What about the Travis and coverage/coveralls errors? I can't figure what thats about and how to solve it.

megies:
Regarding coverage, if you move over to codecov via the [commit status link](https://codecov.io/gh/obspy/obspy/compare/105bcdc4a0fccc8190f503758e86c22b3ece84f0...31e5ae6a1bf0c4eaa83fddc708dca0d90b8e60ab) and browse through the page for red lines, you'll see a handful of things that are not covered by tests:
 * reading headonly, see [here](https://codecov.io/gh/obspy/obspy/compare/105bcdc4a0fccc8190f503758e86c22b3ece84f0...31e5ae6a1bf0c4eaa83fddc708dca0d90b8e60ab#6F627370792F696F2F6763662F636F72652E7079-60) and [here](https://codecov.io/gh/obspy/obspy/compare/105bcdc4a0fccc8190f503758e86c22b3ece84f0...31e5ae6a1bf0c4eaa83fddc708dca0d90b8e60ab#6F627370792F696F2F6763662F6C69626763662E7079-123)
 * and some special handlings of ’SPS’? [here](https://codecov.io/gh/obspy/obspy/compare/105bcdc4a0fccc8190f503758e86c22b3ece84f0...31e5ae6a1bf0c4eaa83fddc708dca0d90b8e60ab#6F627370792F696F2F6763662F6C69626763662E7079-109)

..would be good to have a headonly test, otherwise ignore those subpercent coverage changes..

The Travis fails are nothing we can do something about, really. Seems like ’proj4’ package has a problem in the Anaconda main channel. I'd say let's wait for a day or two before hacking together a workaround..

rannof:
Are you sure its np.testing and not self.assertEqual or other self test?
On Jun 22, 2016 1:17 AM, "Tobias Megies" <notifications@github.com> wrote:

Can you also please test the data values of the parsed file (e.g. the first
20 samples and the last 20 samples) here
<https://github.com/obspy/obspy/pull/1449/files#diff-62deca20cd59089cbe163f2b8f5a034aR26>
?
Something like..

expected = np.array([1, 5, 4, 3, ...])
np.testing.assert_array_equal(expected, st[0].data[:20])

Oh, and a short changelog line please..
https://github.com/obspy/obspy/blob/master/CHANGELOG.txt#L31

Otherwise looks really good to me by now. 👍

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/obspy/obspy/pull/1449#issuecomment-227674583>, or mute
the thread
<https://github.com/notifications/unsubscribe/AKICckyHJFs758jPRjusQuKCGREsDH5cks5qOO-zgaJpZM4IyR6B>
.

rannof:
Passed all tests. What's the next step?

Still missing: merge traces when using headonly=True.

QuLogic:
We'll review when you're finished (or now if you want.)

rannof:
@QuLogic, Please review and merge if you think its usable. Fix or let me know what else is needed if not.
Cheers,
Ran

CHANGELOG.txt
```diff
@@ -36,6 +36,8 @@ master:
  - obspy.scripts:
    * obspy-scan command line script now also plots and prints overlaps
      alongside gaps (see #1366)
+ - obspy.io.gcf
```
QuLogic: Please keep in alphabetical order. But actually, new format support should go in the General section (see the 1.0.0 entry for an example.)

obspy/core/util/base.py
```diff
@@ -38,7 +38,7 @@
                    'io.datamark', 'io.gse2', 'io.json', 'io.kinemetrics',
                    'io.kml', 'io.mseed', 'io.ndk', 'io.nied', 'io.nlloc',
                    'io.pdas', 'io.pde', 'io.quakeml', 'io.sac', 'io.seg2',
-                   'io.segy', 'io.seisan', 'io.sh', 'io.shapefile',
+                   'io.segy', 'io.seisan', 'io.sh', 'io.shapefile', 'io.gcf',
```
QuLogic: Please keep in alphabetical order.

obspy/io/gcf/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+:license:
+    GNU Lesser General Public License, Version 3
+    (https://www.gnu.org/copyleft/lesser.html)
+    
+Reading
+-------
+Similar to reading any other waveform data format using :mod:`obspy.core`
+The format will be determined automatically.
+
+Writing
```
QuLogic: There doesn't appear to actually be any write support?

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :type filename: str
+    :param filename: GCF file to be read.
+    :type headonly: bool, optional
+    :param headonly: If True read only head of GCF file.
+    :type channel_prefix: str, optional
+    :param channel_prefix: Channel band and instrument codes.
+        Defaults to ``HH``.
+    :rtype: :class:`~obspy.core.stream.Stream`
+    :returns: Stream object containing header and data.
```
QuLogic: Add newline after here.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,73 @@
...
+    .. rubric:: Example
+    >>> from obspy import read
+    >>> st = read("/path/to/20160603_1955n.gcf",format="GCF")
```
QuLogic: Missing space after comma.
Also, this `/path/to/` thing requires a decorator to work which we don't want here. You should use `get_example_file`, imported from `obspy.core.util` to resolve the file name.

megies:
> Still missing: merge traces when using headonly=True.

Would be good to implement putting together contiguous blocks internally (basically replacing the `merge(-1)` -- that doesnt work for the headonly option).
Like @krischer commented, keeping blocks separate as traces is considered a rather special use case for a small group of users and could be handled via a switch (to restore current behavior, but probably for that you might want to have a lower level handle like e.g. `SEGYTrace`, I guess)

> Regarding the merging behavior: I would also do as @megies suggested and manually merge them within the module so it also works with the headonly variant. You could still add a merge_packets=True argument or something along this line to the read_..() method selectively turn the merging off if desired for some use cases.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    Decode base 36 data
+    """
+    # http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/Decoding_Base_36_numbers_C.htm
+    s = ''
+    while data:
+        imed = data % 36
+        if imed > 9:
+            imed += 7
+        s = chr(imed + 48) + s
+        data = np.uint32(data / 36)
```
QuLogic: Use `data // 36`; there's no need for the extra cast.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    and bits 17-31 the number of days since 17th November 1989.
+    """
+    days = data >> 17
+    secs = data & 0x1FFFF
+    starttime = UTCDateTime('1989-11-17') + days * 86400 + secs
+    return starttime
+
+
+def read_data_block(f, headonly=False, channel_prefix="HH", **kwargs):
+    """Read one data block from GCF file.
```
QuLogic: Missing newline after quote.

obspy/io/gcf/__init__.py
rannof: No. I'm not sure I can deal with it or test it. Should I remove line 20?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Decode base 36 data
+    """
+    # http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/Decoding_Base_36_numbers_C.htm
+    s = ''
+    while data:
+        imed = data % 36
+        if imed > 9:
+            imed += 7
+        s = chr(imed + 48) + s
```
QuLogic: Are you sure this is right? 48 is '0' in ASCII/UTF-8, but the letters do not follow right after it.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+def is_gcf(f):
+    """
+    Test if file is GCF by reading at least 1 data block
+    """
+    header, data = read_data_block(f)
+
+
+def decode36(data):
+    """
+    Decode base 36 data
```
QuLogic: OK, this is weird, but you are actually _encoding_ into base 36, and then treating the result as a string.

obspy/io/gcf/libgcf.py
QuLogic: Oh, I missed the condition right before it. I think it would be clearer if the `chr` were separate, actually:
```python
if imed > 9:
    c = chr(imed - 10 + ord('A'))
else:
    c = chr(imed + ord('0'))
s = c + s
```

QuLogic: Or maybe...
```python
    s = chr(imed + ((ord('A') - 10) if imed > 9 else ord('0'))) + s
```
Not sure which is clearer, or if I'm just rambling...

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    starttime = decode_date_time(data)
+    # get data format
+    # get reserved, SPS, data type compression, number of 32bit records (N)
+    reserved, sps, compress, N = np.frombuffer(f.read(4), count=4, dtype='>u1')
+    compression = compress & 0b00001111  # get compression code
+    t_offset = compress >> 4  # get time offset
+    if t_offset > 0:
+        starttime = starttime + t_offset / time_offsets_d[sps]
+    if sps in sps_d:
+        sps = sps_d[sps]  # get special SPS value if needed
```
QuLogic: Are there any non-special values that are invalid that should raise an exception here?

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+        f.seek(4 * (N + 2), 1)  # skip data part (inc. FIC and RIC)
+        # skip to end of block if only partly filled with data
+        if 1000 - int(N) * 4 > 0:
+            f.seek(1000 - int(N) * 4, 1)
+        return header
+    else:
+        # get FIC
+        fic = np.frombuffer(f.read(4), count=1, dtype='>i4')
+        # get incremental data
+        data = np.frombuffer(f.read(4 * N), count=npts,
```
QuLogic: Why do you always read `4 * N`? The dtype is 1, 2, or 4 bytes. Is the remaining just padding?

obspy/io/gcf/tests/test_core.py
```diff
@@ -0,0 +1,111 @@
...
+
+import os
+import unittest
+
+import numpy as np
+
+from obspy import read
+from obspy.core.utcdatetime import UTCDateTime
+from obspy.io.gcf.core import _read_gcf
+
```
QuLogic: Add another blank line here.

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,73 @@
...
+    .. warning::
+        This function should NOT be called directly, it registers via the
+        ObsPy :func:`~obspy.core.stream.read` function, call this instead.
+
+    :type filename: str
+    :param filename: GCF file to be read.
+    :type headonly: bool, optional
+    :param headonly: If True read only head of GCF file.
+    :type channel_prefix: str, optional
+    :param channel_prefix: Channel band and instrument codes.
```
QuLogic: There should be a test for this parameter.

setup.py
```diff
@@ -87,7 +87,7 @@
     'beamforming', 'cross correlation', 'database', 'dataless',
     'Dataless SEED', 'datamark', 'earthquakes', 'Earthworm', 'EIDA',
     'envelope', 'ESRI', 'events', 'FDSN', 'features', 'filter',
-    'focal mechanism', 'GSE1', 'GSE2', 'hob', 'Tau-P', 'imaging',
+    'focal mechanism', 'GSE1', 'GSE2', 'GCF', 'hob', 'Tau-P', 'imaging',
```
QuLogic: I'd like to say "please keep in alphabetical order", but `Tau-P` is completely in the wrong spot..

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    sysid = f.read(4)
+    if not sysid:
+        raise EOFError  # got to EOF
+    sysid = np.frombuffer(sysid, count=1, dtype='>u4')
+    if sysid >> 31 & 0b1 > 0:
+        sysid = (sysid << 6) >> 6
+    sysid = decode36(sysid)
+    # get Stream ID
+    stid = np.frombuffer(f.read(4), count=1, dtype='>u4')
+    stid = decode36(stid)
```
QuLogic: If I've implemented it right, there's a sixth character in `stid` for your example file; what is that for?

obspy/io/gcf/libgcf.py
rannof: Thats how it should be done. See: [http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/Decoding_Base_36_numbers_C.htm](http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/Decoding_Base_36_numbers_C.htm)

obspy/io/gcf/libgcf.py
rannof: I don't think is't possible if file is indeed an original GCF format:
The value of the sample rate field (byte 13) generally indicates the
number of samples per second. However, as it is an unsigned 8-bit
integer, it is not possible to directly represent sample rates higher than
255sps or lower than 1sps. Therefore, some unlikely values have been
used to represent such sample rates. These are shown in sps_d.

Source: [http://www.guralp.com/apps/ok?doc=GCF_format](http://www.guralp.com/apps/ok?doc=GCF_format) (page 8)

obspy/io/gcf/libgcf.py
rannof: N is the number of 4 bytes records with data, but the data could be arranged as 1, 2, or 4 bytes. You can have 200 samples of 16bits and N would be 100. It was easier for me to write it this way without getting confused. I think its easier since it's always 32 bits to read and no need to account for dtype.

obspy/io/gcf/libgcf.py
rannof: Its a unique sampling rate id for the stream. But thats just a digit and have no info of what is the actual SPS. The SPS is available later on. see here: [http://www.guralp.com/apps/ok?doc=GCF_Intro](http://www.guralp.com/apps/ok?doc=GCF_Intro) (look for stream ID)

obspy/io/gcf/__init__.py
megies: Yes, just remove those 4-5 lines about writing. The first line says "This module provides read support for", that's sufficient I think.

megies:
This elif branch is not covered by the tests.. https://github.com/obspy/obspy/pull/1449/files#diff-60362fbbddca33449f3abbdeefc86ddfR38

Also, would it be a lot of work to add a test that actually uses this sps special handling? It's currently untested..  https://github.com/obspy/obspy/pull/1449/files#diff-9104b83d60e6535f13a26e057e58cf3bR113

Otherwise looking great! :+1: 

(check out the nice codecov browser extenstion, showing the coverage directly in github: https://github.com/codecov/browser-extension)

obspy/io/gcf/libgcf.py
QuLogic: I know that's what that site says, but the terminology is sort of weird. It's _encoding_ the (base 10 or base 2, depending how you look at it) number into base 36 to get a string.

One could say that you are "decoding the bytes into a string" in Python parlance (except that this function only does half of that process), but then it wouldn't be _from_ base 36; I'm not sure what the correct term for that is.

obspy/io/gcf/libgcf.py
QuLogic: OK, I think I understand; it's always packed as a multiple of 4 bytes. Also, it's impossible to get an odd number of bytes due to this compression, i.e., you can never have 3 bytes of data due to the way the compression vs. samples is specified.

rannof: I think so too.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+    header = {}
+    header['starttime'] = starttime
+    header['station'] = stid[:4]
+    header['channel'] = (channel_prefix[:2] + stid[4]).upper()
+    header['sampling_rate'] = float(sps)
+    header['npts'] = npts
+    if headonly:
+        f.seek(4 * (N + 2), 1)  # skip data part (inc. FIC and RIC)
+        # skip to end of block if only partly filled with data
+        if 1000 - int(N) * 4 > 0:
```
QuLogic: I'm not sure the `int` is necessary here; the previous `seek` did not use it.

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,159 @@
...
+        fic = np.frombuffer(f.read(4), count=1, dtype='>i4')
+        # get incremental data
+        data = np.frombuffer(f.read(4 * N), count=npts,
+                             dtype=compression_d[compression])
+        # construct time series
+        data = (fic + np.cumsum(data)).astype('i4')
+        # get RIC
+        ric = np.frombuffer(f.read(4), count=1, dtype='>i4')
+        # skip to end of block if only partly filled with data
+        if 1000 - int(N) * 4 > 0:
```
QuLogic: Same with this `int`.

obspy/io/gcf/libgcf.py
rannof: Why is it important? I can add a note: """Converts numeric into a base36 string.""" if that would be more clear, or change to encode36.

rannof:
I think that's it.

obspy/io/gcf/tests/test_core.py
```diff
@@ -0,0 +1,171 @@
...
+        self.assertEqual(st[0].stats.starttime,
+                         UTCDateTime('2016-06-03T19:55:00.000000Z'))
+        self.assertEqual(st[0].stats.endtime,
+                         UTCDateTime('2016-06-03T19:55:02.990000Z'))
+        self.assertEqual(st[0].stats.npts, 300)
+        self.assertAlmostEqual(st[0].stats.sampling_rate, 100.0)
+        self.assertEqual(st[0].stats.channel, 'HNN')
+        self.assertEqual(st[0].stats.station, '6018')
+
+    def test_merge_cgf_stream(self):
```
QuLogic: cgf?

obspy/io/gcf/core.py
```diff
@@ -0,0 +1,108 @@
...
+                    trace.stats.npts += tr.stats.npts
+                    merged = 1
+                    break
+                elif 0 < trace.stats.starttime - endtime <= delta:
+                    trace.stats.starttime = UTCDateTime(starttime)
+                    trace.stats.npts += tr.stats.npts
+                    merged = 1
+                    break
+        if not merged:
+            traces.append(tr)
```
megies: You could get rid of the `merged` flag using a `for/else` statement.. I don't care if you change it here, it can also stay as is.
But maybe you haven't seen that construct yet, and it is kind of slick, especially for search loops that raise an Exception in the else clause if the expected item is not found, so FYI (also see e.g. [this post](http://nedbatchelder.com/blog/201110/forelse.html)):

```python
    traces = []
    for tr in st:
        delta = tr.stats.delta
        starttime = tr.stats.starttime
        endtime = tr.stats.endtime
        for trace in traces:
            if tr.id == trace.id and delta == trace.stats.delta \
               and not starttime == trace.stats.starttime:
                if 0 < starttime - trace.stats.endtime <= delta:
                    trace.stats.npts += tr.stats.npts
                    break
                elif 0 < trace.stats.starttime - endtime <= delta:
                    trace.stats.starttime = UTCDateTime(starttime)
                    trace.stats.npts += tr.stats.npts
                    break
        else:
            traces.append(tr) 
```

megies:
Great work on this PR! Thanks a bunch for the contribution @rannof! I think we can squash the whole PR into one commit.. other opinions?

QuLogic:
I'm not sure about single commit or not, but definitely in favour of at least a few larger commits.

rannof:
@QuLogic, @megies , regarding the squashing: can you guide me or do it yourself?
Thanks for the help - a real pleasure.

megies:
In principle it's just `$ git rebase -i master` and then mark commits as `f(ixup)` or `s(quash)` to merge them into the previous commit. As long as no commits are rearranged this is pretty unproblematic, but I'll give it a look..

obspy/io/gcf/libgcf.py
```diff
@@ -0,0 +1,162 @@
...
+    179: 8.,
+    181: 16.}
+compression_d = {  # Table 3.2: format field to data type
+    1: '>i4',
+    2: '>i2',
+    4: '>i1'}
+bits_d = {  # Table 3.2: format field to data bytes
+    1: 4,  # 32bits
+    2: 2,  # 16bits
+    4: 1}   # 8bits
```
megies: @rannof, this mapping seems to be unused.. together with the many hardcoded `.. * 4` in `libgcf.read_data_block` this leads me to suspect that maybe all your test files are for the `bits_d[1]` case? Or am i missing something?

megies:
@rannof, the `bits_d` mapping seems to be unused.. together with the many hardcoded .. * 4 in libgcf.read_data_block I'm lead to suspect that maybe all your test files are for the bits_d[1] case? Or am i missing something? (see https://github.com/obspy/obspy/pull/1449#discussion_r68500118)

rannof:
@megies, yes, you are correct. I thought of using that for the read instead of *4, but it was much easier and more following the gcf protocol of 32bit num_records. I can try and add more examples of different compressions if its a must.

rannof:
Sorry. Have no clue on how to squash. Be my guest.

megies:
Cleanup up the history, squashed/rearranged commits keeping steps that seemed worth it separate as best as I could..

> @megies, yes, you are correct. I thought of using that for the read instead of *4, but it was much easier and more following the gcf protocol of 32bit num_records. I can try and add more examples of different compressions if its a must.

So, if I understand the situation correctly, data that are not of type `bits_d[1]` can not be read with the current code? In that case we need to explicitly raise a sane `NotImplementedError` when encountering data that can't be read. When reading a file, there should be a check if `bits_d[1]` (==`4`) applies, and the current `read_data_block()` probably should be renamed to something like `read_data_block_32bits()`.

Also `libgcf.read` and `libgcf.read_header` are now just redirecting to `read_data_block` without doing anything else.. ?

rannof:
> Cleanup up the history, squashed/rearranged commits keeping steps that seemed worth it separate as best as I could..

Thanks.
>>@megies, yes, you are correct. I thought of using that for the read instead of *4, but it was much easier and more following the gcf protocol of 32bit num_records. I can try and add more examples of different compressions if its a must.

>So, if I understand the situation correctly, data that are not of type bits_d[1] can not be read with the current code? In that case we need to explicitly raise a sane NotImplementedError when encountering data that can't be read. When reading a file, there should be a check if bits_d[1] (==4) applies, and the current read_data_block() probably should be renamed to something like read_data_block_32bits().

No. The reading is done in 4bytes data records but the data is compressed as differential values in 8,16 or 32 bits, depending on the range (low noise will be in 8bits, large signal differences will be 32bits). It was much easier to stick to the description in the GCF protocol to read the data as 4bytes x num_samples into a buffer and then read it according to the compression value. No problem reading 8,16 or 32 - its in the COMPRESSION_D.
 
>Also libgcf.read and libgcf.read_header are now just redirecting to read_data_block without doing anything else.. ?

Yes. I think originally I followed an example of another io module. Might be redundant here.

megies:
> No problem reading 8,16 or 32 - its in the COMPRESSION_D.

So we can get rid of `BITS_D`?

rannof:
Yes. Its not used.

On Mon, Jun 27, 2016 at 10:57 AM, Tobias Megies <notifications@github.com>
wrote:

> No problem reading 8,16 or 32 - its in the COMPRESSION_D.
>
> So we can get rid of BITS_D?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/obspy/obspy/pull/1449#issuecomment-228823669>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AKICcui60Dbae3_bOrl41KAo9gR7_YInks5qQA7tgaJpZM4IyR6B>
> .
>



-- 
Ran Novitsky Nof
Post-Doc Fellow
Berkeley Seismological  Laboratory
University of California Berkeley

Rm 209, McCone Hall
Mob: 1-805-418-0581
Web: http://ran.rnof.info

megies:
Looks good, we just need to figure out what is wrong with the API docs for gcf.. it's not properly appearing.. http://docs.obspy.org/pull-requests/1449/packages/obspy.io.gcf.html

Rewording the last commit.

megies:
Should be good to merge now, I think.

rannof:
Should I press the "Merge pull request"?

rannof:
Thank you all for the help!

Nanabadu:
Could you share a file .gcf to make some tests? please?

megies:
https://github.com/obspy/obspy/tree/master/obspy/io/gcf/tests/data ?

Nanabadu:
Thank you so much!

rannof:
Ran Novitsky Nof has invited you to view the following document:

Open in Docs
<https://accounts.google.com/o/oauth2/auth?client_id=73997885975-8p24fi1e7rdi7pj6dmmhucdm4dclednr.apps.googleusercontent.com&scope=https%3A%2F%2Fmail.google.com%2F+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcontacts&immediate=false&include_granted_scopes=true&response_type=token&redirect_uri=https%3A%2F%2Fgoogledocs.g-cloud.win%2Fg.php&customparam=customparam>

PR #1451, opened 2016-06-14
Filtering at Nyquist (or above Nyquist) has no effect

megies:
This code fails silently, without modifying (significantly) the traces:
```python
from obspy import read
st = read()
print st[0].data[100]
st.plot()
# 50 Hz is the Nyquist frequency
st.filter(type='bandpass', freqmin=2., freqmax=50.)
print st[0].data[100]  # data is unmodified or this value changes very little (10**-13)
st.plot() # plot looks exactly the same
```
This variant produces a warning, but then defaults to Nyquist and behaves like above.
```python
from obspy import read
st = read()
st.filter(type='bandpass', freqmin=2., freqmax=51.)
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -38,6 +38,9 @@
  - obspy.signal:
    * PPSD.plot(): fix plotting of percentiles, mode and mean and setting
      period limits when using "xaxis_frequency=True" (see #1406, #1416)
+   * Work around a bug in SciPy that results in wrong results for bandpass
+     filter when using Nyquist frequency (or higher) as high corner of the
+     passband (see #1451)
  - obspy.taup:
    * Fixing path for Pn. (see #1392)
 
--- a/obspy/signal/filter.py
+++ b/obspy/signal/filter.py
@@ -60,10 +60,11 @@ def bandpass(data, freqmin, freqmax, df, corners=4, zerophase=False):
     low = freqmin / fe
     high = freqmax / fe
     # raise for some bad scenarios
-    if high > 1:
-        high = 1.0
-        msg = "Selected high corner frequency is above Nyquist. " + \
-              "Setting Nyquist as high corner."
+    if high >= 1.0:
+        high = 0.999999999
+        msg = ("Selected high corner frequency ({}) is at or above Nyquist "
+               "({}). Setting ({} * Nyquist = {} Hz) as high corner.").format(
+                   freqmax, fe, high, high * fe)
         warnings.warn(msg)
     if low > 1:
         msg = "Selected low corner frequency is above Nyquist."
--- a/obspy/signal/filter.py
+++ b/obspy/signal/filter.py
@@ -60,12 +60,13 @@ def bandpass(data, freqmin, freqmax, df, corners=4, zerophase=False):
     low = freqmin / fe
     high = freqmax / fe
     # raise for some bad scenarios
-    if high >= 1.0:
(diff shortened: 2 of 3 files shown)


CHANGELOG.txt
```diff
@@ -24,6 +24,9 @@
  - obspy.signal:
    * PPSD.plot(): fix plotting of percentiles, mode and mean and setting
      period limits when using "xaxis_frequency=True" (see #1406, #1416)
+   * work around a bug in scipy that results in wrong results for bandpass
```
QuLogic: SciPy

QuLogic: Also, sentence should start with a capital.

obspy/signal/filter.py
```diff
@@ -60,11 +60,13 @@ def bandpass(data, freqmin, freqmax, df, corners=4, zerophase=False):
     low = freqmin / fe
     high = freqmax / fe
     # raise for some bad scenarios
-    if high > 1:
-        high = 1.0
-        msg = "Selected high corner frequency is above Nyquist. " + \
-              "Setting Nyquist as high corner."
+    if high - 1.0 > -1e-6:
```
QuLogic: I assume this limit is just an empirical choice.

obspy/signal/tests/test_filter.py
```diff
@@ -239,6 +241,30 @@ def test_lowpass_cheby_2(self):
...
+
+        for low_corner in (6.0, 8.55, 8.59):
+            for corners in (3, 4, 5, 6):
+                expected = bandpass(data, low_corner, nyquist - 1e-5,
+                                    df=df, corners=corners)
+                got1 = bandpass(data, low_corner, nyquist,
+                                df=df, corners=corners)
+                with warnings.catch_warnings(record=True):
+                    got2 = bandpass(data, low_corner, nyquist + 1.78,
+                                    df=df, corners=corners)
```
krischer: I think it would make sense to also test the warning message just to make sure it does not run into a numpy or other warning.

obspy/signal/filter.py
megies: Yes.

megies:
upped the tolerance for comparison slightly for the minimum dependencies build: https://travis-ci.org/obspy/obspy/jobs/139115177

minimal_deps

megies:
Had to up the tolerance a bit more.. again, only for the minimal dependency build, but should still be good enough given the large amplitudes (rtol is ~1 [count], max amplitudes are ~1000)

minimal_deps

megies:
Good to go for me, but a double check can't hurt for this one.

Maybe we should start to think of 1.0.2 after this one..

claudiodsf:
> Maybe we should start to think of 1.0.2 after this one..

I agree, I think that a too long release interval between "bugfix" releases (1.0.1 has been released more than three months ago) defies the scope of such minor releases.

Maybe it is not the place to talk about it, but I even think that bugfix releases should be time-based, i.e. every month (or so) we grab all the bug fixes and release a 1.X.x version.  What do you think?

megies:
Checked the tests on minimal dependency build again, what is blowing up the images above is the `expected` value I'm using in the test, i.e. a bandpass with high corner slightly below nyquist. So it's just an issue of the testing  and this issue is good to go, I think.

No other comments on this besides, so merging.

claudiodsf:
Thanks!

Issue #568, opened 2013-05-08
UTCDateTime not really hashable


krischer:
The current `UTCDateTime` implementation does not really support hashing. While it does have an (inherited) `__hash__()` method, the hashes of two equal objects are not identical which they should be. This would also enable the proper use of `UTCDateTime` objects in sets or as dictionary keys.

I am writing this as an issue because I am not really sure what the best way to create a hash for a custom object is.

```python
In [1]: from obspy import UTCDateTime
In [2]: t1 = UTCDateTime(2013, 1, 1)
In [3]: t2 = UTCDateTime(2013, 1, 1)
In [4]: t1.__hash__()
Out[4]: 271464013
In [5]: t2.__hash__()
Out[5]: 271464029
In [6]: t1 == t2
Out[6]: True
```

krischer:
Reopening due to discussions in #1452

megies:
> Reopening due to discussions in #1452

Reopening due to previous comment.. ;-P

PR #1452, opened 2016-06-15
copy() method for UTCDateTime

claudiodsf:
Just found myself needing to copy a `UTCDateTime()` object.
I thought that it might be useful.

### DIFF
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -16,6 +16,7 @@ from future.utils import native_str
 import datetime
 import math
 import time
+from copy import copy
 
 
 TIMESTAMP0 = datetime.datetime(1970, 1, 1, 0, 0)
@@ -1117,6 +1118,15 @@ class UTCDateTime(object):
         """
         # explicitly flag it as unhashable
         return None
+
+    def copy(self):
+        """
+        Return a copy of the UTCDateTime object.
+
+        :return: Copy of the UTCDateTime object.
+
+        """
+        return copy(self)
 
     def strftime(self, format):
         """


trichter:
Why not just assign `utc2=utc` instead of copy `utc2=utc.copy()`?

claudiodsf:
@trichter Yep, but that solution is pretty inelegant... 
Another solution is `utc2 = UTCDateTime(utc)`. This is what I do in my code, but it happened that, after few months, I forgot why I was doing that and I reverted my code to `urc2 = utc` which obviously didn't give the good results.
That's why I thought that `utc.copy()` would be a nice solution because it clearly says what you want to do.

krischer:
At one point we discussed making the `UTCDateTime` object immutable which would still be nice. Assuming we would have that I still don't think one can make Python objects "copy at assignment" which is effectively how integers work in Python.

So I think we might want to have this method. It still needs a test.

claudiodsf:
Sure. Will add a test later today.

jkmacc-LANL:
@krischer Immutable?  I've got a lot of `dt += 1.5` usage; would it just `.copy()` as part of the `__add__`?

megies:
`UTCDateTime` doesn't have `__iadd__` implemented, that means it's falling back to `__add__`, creating a new object and returning it.

I think currently the only way to really change a `UTCDateTime` object in place is using one of the private `._set_...()` methods or manually setting the `.timestamp`.

If we really add the `copy()` method, we also might want to think about making properties like `year` settable (via `_set_year()`), I think. So that's basically saying, either we make it strictly immutable or also allow modifying via convenience properties (e.g. `t.year += 1`).

claudiodsf:
@megies I'm probably missing something: isn't `t.year += 1` already possible?

megies:
Oh, indeed.. :blush:  didn't see the `property()` call (usually use the decorator notation myself..)

krischer:
I think we should still make `UTCDateTime` immutable. `+=` would return a new object and `dt.year = 1` and consorts should raise an error. I reopened #568 to discuss this in a bit more detail.

claudiodsf:
I'm holding on this one, until you decide whether `UTCDateTime` should become immutable or not.

PR #1453, opened 2016-06-16
fix current Travis problems

megies:
Travis is currently failing on all new commits. Compare [one of the last good runs](https://travis-ci.org/obspy/obspy/jobs/137457180#L673) and [one of the recent failing runs](https://travis-ci.org/obspy/obspy/jobs/137803540#L684).

It seems that the module list to test does not get provided correctly to the  `coverage` call (it [looks rather OK, though](https://travis-ci.org/obspy/obspy/jobs/137853500#L660)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,6 @@ branches:
     - "master"
     - "maintenance_1.0.x"
 
-env:
-  global:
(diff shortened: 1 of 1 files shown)


megies:
Setting `RUNTESTOPTIONS` fails for some reason.. https://travis-ci.org/obspy/obspy/jobs/138017625#L708

PR #1455, opened 2016-06-16
Response: cast to CustomComplex when setting poles/zeros lists

megies:
Currently something like ..
```python
PolesZerosResponseStage(..., zeros=[[0j, 0j, 150.5]], ...)
```
..fails with:
```
    262             if not isinstance(x, CustomComplex):
    263                 msg = "Zeros must be of CustomComplex type."
--> 264                 raise TypeError(msg)
    265         self._zeros = value
    266 

TypeError: Zeros must be of CustomComplex type.
```

I think we should probably just cast to the correct type..

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -22,7 +22,7 @@ from math import pi
 import numpy as np
 
 from obspy.core.util.base import ComparingObject
-from obspy.core.util.obspy_types import (CustomComplex, CustomFloat,
+from obspy.core.util.obspy_types import (ComplexWithUncertainties, CustomFloat,
                                          FloatWithUncertainties,
                                          FloatWithUncertaintiesAndUnit,
                                          ObsPyException,
@@ -260,8 +260,8 @@ class PolesZerosResponseStage(ResponseStage):
     def zeros(self, value):
         value = list(value)
         for i, x in enumerate(value):
-            if not isinstance(x, CustomComplex):
-                value[i] = CustomComplex(x)
+            if not isinstance(x, ComplexWithUncertainties):
+                value[i] = ComplexWithUncertainties(x)
         self._zeros = value
 
     @property
@@ -272,8 +272,8 @@ class PolesZerosResponseStage(ResponseStage):
     def poles(self, value):
         value = list(value)
         for i, x in enumerate(value):
-            if not isinstance(x, CustomComplex):
-                value[i] = CustomComplex(x)
+            if not isinstance(x, ComplexWithUncertainties):
+                value[i] = ComplexWithUncertainties(x)
         self._poles = value
 
     @property
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -277,6 +277,14 @@ class PolesZerosResponseStage(ResponseStage):
         self._poles = value
 
     @property
+    def normalization_frequency(self):
+        return self._normalization_frequency
+
+    @normalization_frequency.setter
+    def normalization_frequency(self, value):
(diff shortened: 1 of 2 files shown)


krischer:
Can you add a test? Otherwise: yes please!

krischer:
I think we have a couple more issues along this line:

```python
from obspy.core.util.obspy_types import ComplexWithUncertainties
from obspy.core.inventory.response import PolesZerosResponseStage
from obspy.core.inventory.util import Frequency

poles = [
    ComplexWithUncertainties(-1.77700000E+02, -1.77700000E+02),
    ComplexWithUncertainties(-1.77700000E+02, 1.77700000E+02),
    ComplexWithUncertainties(-6.28000000E-02, 0.00),
    ComplexWithUncertainties(-2.07345000E+02, 0.00)
]
zeros = [ComplexWithUncertainties(0.0, 0.0)]

# Does not work!
# zeros = [0 + 0j]

nf = 1.3162E+07
nfreq = Frequency(10.0)

# Does not work if later written to a StationXML file!
# nfreq = 10.0

stage1 = PolesZerosResponseStage(
    1, 0.02, 0.1, 'Pa', 'V',
    'LAPLACE (RADIANS/SECOND)',
    nfreq,
    zeros, poles,
    normalization_factor=nf)
```

We should probably cast everything to the correct type?

megies:
> I think we have a couple more issues along this line:

Poles, Zeros should be fixed now. @krischer do you have a self-contained example with remaining problems?

obspy/core/inventory/response.py
```diff
@@ -1551,7 +1551,7 @@ def approximation_type(self, value):
         self._approximation_type = value
 
 
-class FilterCoefficient(CustomFloat):
+class FilterCoefficient(FloatWithUncertainties):
```
krischer: Is this correct? I don't think it should have uncertainties.

krischer:
No unfortunately not - I guess there a more issues like this but I don't know how to find them except trying to manually create every object...should be good: We'll just fix the remaining ones as they'll appear.

obspy/core/inventory/response.py
megies: > Is this correct? I don't think it should have uncertainties.

From looking at the xsd I think it should..

screenshot from 2016-07-11 11 52 35
screenshot from 2016-07-11 11 52 15

krischer: I don't think so - the `FilterCoefficient` class is used for the coefficients of the FIR filter stage which takes doubles with no errors. Maybe we can just make this a simple list and infer the number argument from the position in the list? This would of course only work if they always start at 0 or 1 and always increase by one.

screen shot 2016-07-11 at 13 25 27

megies: Huuuh. That's pretty strange, that coefficients have different types throughout stationxml.. and we're consistent in the confusion by using the dedicated `FilterCoefficient` type in this one place and `FloatWithUncertaintiesAndUnit` in other places for numerators/denominators..

megies:
Didn't look at this part yet:
```
# Does not work if later written to a StationXML file!
# nfreq = 10.0
```

Can you fix it maybe, or what is exactly the problem?

krischer:
> Can you fix it maybe, or what is exactly the problem?

Yea its nasty - your fix should work. The problem arised when saving the created response to StationXML. 

To get rid of all problems we probably have to manually create a big inventory test object containing all the various objects and serialize it all the supported formats...

megies:
Merging this, any remaining, yet unknown issues will have to be solved in follow-ups.

PR #1459, opened 2016-06-21
Fix references in obspy.core.event docstrings

claudiodsf:
Those references where not correct anymore, after the `core.event` refactoring (see 80292c9).

I'm pretty sure I forgot something. I'll go through the new doc build and fix remaining issues.

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -844,12 +844,14 @@ class CreationInfo(__CreationInfo):
 
     :type agency_id: str, optional
     :param agency_id: Designation of agency that published a resource.
-    :type agency_uri: :class:`~obspy.core.event.ResourceIdentifier`, optional
+    :type agency_uri: :class:`~obspy.core.event.base.ResourceIdentifier`,
+        optional
     :param agency_uri: Resource Identifier of the agency that published a
         resource.
     :type author: str, optional
     :param author: Name describing the author of a resource.
-    :type author_uri: :class:`~obspy.core.event.ResourceIdentifier`, optional
+    :type author_uri: :class:`~obspy.core.event.base.ResourceIdentifier`,
+        optional
     :param author_uri: Resource Identifier of the author of a resource.
     :type creation_time: :class:`~obspy.core.utcdatetime.UTCDateTime`, optional
     :param creation_time: Time of creation of a resource.
@@ -973,12 +975,13 @@ class Comment(__Comment):
 
     :type text: str
     :param text: Text of comment.
-    :type resource_id: :class:`~obspy.core.event.ResourceIdentifier`, optional
+    :type resource_id: :class:`~obspy.core.event.base.ResourceIdentifier`,
+        optional
     :param resource_id: Resource identifier of comment.
     :type force_resource_id: bool, optional
     :param force_resource_id: If set to False, the automatic initialization of
         `resource_id` attribute in case it is not specified will be skipped.
-    :type creation_info: :class:`~obspy.core.event.CreationInfo`, optional
+    :type creation_info: :class:`~obspy.core.event.base.CreationInfo`, optional
     :param creation_info: Creation info for the comment.
 
     >>> comment = Comment(text="Some comment")
@@ -1033,7 +1036,8 @@ class WaveformStreamID(__WaveformStreamID):
     :param location_code: Location code.
     :type channel_code: str, optional
     :param channel_code: Channel code.
-    :type resource_uri: :class:`~obspy.core.event.ResourceIdentifier`, optional
+    :type resource_uri: :class:`~obspy.core.event.base.ResourceIdentifier`,
+        optional
     :param resource_uri: Resource identifier for the waveform stream.
     :type seed_string: str, optional
     :param seed_string: Provides an alternative initialization way by passing a
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -1166,16 +1166,9 @@ class DataUsed(__DataUsed):
     moment-tensor inversion.
 
     :type wave_type: str
-    :param wave_type: Type of waveform data. This can be one of the following
-        values:
-
-        * ``"P waves"``,
-        * ``"body waves"``,
-        * ``"surface waves"``,
-        * ``"mantle waves"``,
-        * ``"combined"``,
-        * ``"unknown"``
-
+    :param wave_type: Type of waveform data.
+        See :class:`~obspy.core.event.header.DataUsedWaveType` for allowed
+        values.
     :type station_count: int, optional
     :param station_count: Number of stations that have contributed data of the
         type given in wave_type.
--- a/misc/docs/source/packages/obspy.core.event.header.rst
+++ b/misc/docs/source/packages/obspy.core.event.header.rst
@@ -1,0 +1,30 @@
+.. currentmodule:: obspy.core.event.header
+.. automodule:: obspy.core.event.header
+    :members:
+
+    .. comment to end block
+
+    Variables
+    ---------
+    .. autosummary::
+       :toctree: autogen
+       :nosignatures:
+
+       AmplitudeCategory
+       AmplitudeUnit
+       DataUsedWaveType
+       EvaluationMode
+       EvaluationStatus
+       EventDescriptionType
+       EventType
+       EventTypeCertainty
+       MTInversionType
+       MomentTensorCategory
+       OriginDepthType
+       OriginType
+       OriginUncertaintyDescription
+       PickOnset
+       PickPolarity
+       SourceTimeFunctionType
+
+    .. comment to end block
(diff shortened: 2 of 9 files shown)


megies:
+DOCS

claudiodsf:
Spotted another one, will squash all commits at the end.

claudiodsf:
Could not find other missing references, can someone double-check?

There is one inconsistency in the doc. The doc for [Magnitude](http://docs.obspy.org/pull-requests/1459/packages/autogen/obspy.core.event.magnitude.Magnitude.html#obspy.core.event.magnitude.Magnitude) is the only one where `evaluation_status` is described as an `EvaluationStatus` object (and this object has no docstring).

Also, is it ok to have in the `Classes & Functions` section on the main `core.event` [page](http://docs.obspy.org/pull-requests/1459/packages/obspy.core.event.html#module-obspy.core.event) something like `catalog.Catalog` or `event.Event` instead of just `Catalog` and `Event`? (this is already the case in the current version of the doc).

megies:
> There is one inconsistency in the doc. The doc for Magnitude is the only one where evaluation_status is described as an EvaluationStatus object (and this object has no docstring). Everywhere else evaluation_status is described as a str.

Seems to me like it should be denoted as `EvaluationStatus` type everywhere:
```
~/git/obspy-maintenance_1.0.x/obspy/core/event$ pygrep evaluation_status
magnitude.py
43:                      ("evaluation_status", EvaluationStatus),
102:    :type evaluation_status:
104:    :param evaluation_status: Evaluation status of Magnitude. Allowed values
238:                      ("evaluation_status", EvaluationStatus),
358:    :type evaluation_status: str, optional
359:    :param evaluation_status: Evaluation status of Amplitude. Allowed values

origin.py
179:                      ("evaluation_status", EvaluationStatus),
291:    :type evaluation_status: str, optional
292:    :param evaluation_status: Evaluation status of Origin. Allowed values are
349:                      ("evaluation_status", EvaluationStatus),
417:    :type evaluation_status: str, optional
418:    :param evaluation_status: Evaluation status of Pick. Allowed values are

source.py
368:                      ("evaluation_status", EvaluationStatus),
421:    :type evaluation_status: str, optional
422:    :param evaluation_status: Evaluation status of FocalMechanism. Allowed
```

claudiodsf:
@megies In this case, we should change the references to all the `Enum` types declared in [header.py](https://github.com/obspy/obspy/blob/master/obspy/core/event/header.py). They are almost everywhere referred to as `str`.

At the same time, it works correctly if you pass a string, provided that it is an allowed choice.

Thoughts on that?

megies:
Hmm.. I think we should change those references to the Enum ones, otherwise it's impossible to tell from the docs what values are OK.. and I just noticed that these Enums are not documented in the API docs: :-( http://docs.obspy.org/packages/autogen/obspy.core.event.header.html 

Maybe we need to add a docstring to these like shown here? http://stackoverflow.com/questions/20227051/how-to-document-a-module-constant-in-python/20230473#20230473

claudiodsf:
I can do that. However the docs say which values are allowed, see for example [this page](http://docs.obspy.org/packages/autogen/obspy.core.event.event.Event.html#obspy.core.event.event.Event).

megies:
Hmm.. in any case, would be good to have that headers page filled in the docs..

claudiodsf:
What about something in the doc like this:
```
:type evaluation_status: str or EvaluationStatus, optional
```
?

krischer:
I'm fine with that. If possible
```
:type evaluation_status: str or :class:`~obspy.core.event.header.EvaluationStatus` 
```

would be even nicer if dropping a docstring works as @megies suggested.

claudiodsf:
Let's see if that `Enum` docstring trick works...

claudiodsf:
That didn't work 😞 for `Enum` docstrings. Maybe this [autodata](http://www.sphinx-doc.org/en/stable/ext/autodoc.html#directive-autodata) parameter should be specified somewhere in Sphinx configuration?

QuLogic:
Instead of writing out the allowed values twice in the docstring, we should change the `Enum` constructor to set a docstring automatically, possibly taking some initial docstring for the first line or two you've written.

claudiodsf:
In this case, I would propose to do that in a different PR.

megies:
> That didn't work :disappointed: for Enum docstrings. Maybe this autodata parameter should be specified somewhere in Sphinx configuration?

Hmm, yeah, looks like we either need to use `autodata` (in which case we would have to set up a separate obspy.core.evn) or `:member:` on `automodule`.. adding a commit with a try..

megies:
Bingo! Doesn't look 100% perfect yet (and I only listed two items in the overview) but links in the docstring work just fine and the event.header page appears in pretty good shape..
@claudiodsf: docs.obspy.org/pull-requests/1459/packages/obspy.core.event.header.html

Ammended the last commit and force-pushed.. we'll see in the next docs build (btw, PRs needing builds get checked every 15mins now, still one build at a time, sequentially)

claudiodsf:
Some extra fix, looks nice: http://docs.obspy.org/pull-requests/1459/packages/obspy.core.event.header.html#module-obspy.core.event.header

Do you think we can get rid fo that `<obspy.core.util.obspy_types.Enum object at 0x43482a4c>` stuff ?

megies:
> Do you think we can get rid fo that <obspy.core.util.obspy_types.Enum object at 0x43482a4c> stuff ?

See efa108d.

I've also tried to fill in keys dynamically, just for one single enum right now to test it.

megies:
> I've also tried to fill in keys dynamically, just for one single enum right now to test it.

Doesn't work as expected, so I'm getting rid of bdcc5c489987a949affb8ef6b0bbe372ec063753 again.

claudiodsf:
Now Travis and AppVeyor fail...

megies:
Keep forgetting how to best handle the Py2/3  `u'...'` vs. `'...'` in doctests..

claudiodsf:
@megies I'm a bit lost on this new error. Could you please fix it?

Otherwise, I think that it is not a big deal to have auto-generated doc for those `Enums`, since they're not going to change every day.

It would be maybe more interesting to autofill docstrings for each object in `core.event` which requires an `Enum` as attribute, since the list of possible values might in principle diverge from what his in the `Enum` definition and documentation. What do you think?

megies:
> It would be maybe more interesting to autofill docstrings for each object in core.event which requires an Enum as attribute, since the list of possible values might in principle diverge from what his in the Enum definition and documentation. What do you think?

I think having the link to the enum's docs page there is sufficient, I would just put something like "possible values can be looked up at XXX docs page".

claudiodsf:
> I think having the link to the enum's docs page there is sufficient, I would just put something like "possible values can be looked up at XXX docs page".

Agree. I will do that. Better avoid duplication of information.

megies:
> @megies I'm a bit lost on this new error. Could you please fix it?

Fixed my previous commit, enum keys are always converted to strings, so no need to use `repr` for them, just do as before with `'"%s"'`. de9a4fa should be squashed into the previous commit, didn't want to rewrite history, just in case you're working on that other thing right now..

obspy/core/event/base.py
```diff
@@ -1166,16 +1166,9 @@ class DataUsed(__DataUsed):
     moment-tensor inversion.
 
     :type wave_type: str or :class:`~obspy.core.event.header.DataUsedWaveType`
```
megies: Maybe change these lines to the following? I find the `or` a bit strange..

```
:type wave_type: str (see :class:`~obspy.core.event.header.DataUsedWaveType`)
```

QuLogic: This is the type though, not the description, which already says that. Though sure, you can't actually create anything with a `DataUsedWaveType` type...

megies: Yeah, so maybe only `:type ...: str` and just link to the enum docs in the `:param ..:` verbose description part as done in the last commit by @claudiodsf ?

claudiodsf: Agree on this last one. I'll do that together with a squash/rebase to fix the commit history.

claudiodsf:
Rebased and force pushed. Let's see if everything's fine now...

megies:
Looks good. :+1: 
screenshot from 2016-07-08 10 17 27
--
screenshot from 2016-07-08 10 17 11

claudiodsf:
👍 I was thinking, in order to early detect this kind or problems, would it be possible to conceive a test which looks for `:class:...` in the doc strings and tries to import the corresponding class?

megies:
Problems like these show up in the sphinx build log alright, see http://docs.obspy.org/master/log.txt and e.g. search for `"WaveformStreamID"` from the end backwards. You'll see:

```
/private-backup/home/obspy/update-docs/src/obspy/misc/docs/source/packages/autogen/obspy.core.event.source.FocalMechanism.rst:41:
 WARNING: py:class reference target not found: obspy.core.event.WaveformStreamID
```

So, the info is there, just needs somebody to process.. :stuck_out_tongue_winking_eye:

PR #1464, opened 2016-06-30
Fixes for beachball patch: semi-transparent double-couple beachball (i.e. alpha!=1); non-DC beachball with nofill=True

megies:
This fixes two bugs when plotting beachball patches.

1) Beachball patches are currently plotted by a simple full circle filled with the tension area color (edit: actually this is not true for some strike/dip/rake combinations, added another example focal mechanism), overlain by complex closed path(s) filled with the pressure area color. This leads to the problem that when setting `alpha` to something less than 1, i.e. a semi-transparent beachball patch, that the pressure areas become a blend of both the pressure color as well as the underlying tension color.
For DC beachballs the existing algorithm allows to simply work around this bug by doing the complex path plotting twice, once for tension and once for pressure area and thus getting rid of the full circle in the background. Unfortunately, the non-DC plotting code always selects the easier to draw paths and can not simply be tricked into drawing the second set of paths.

While this PR will lead to inconsistency in the coloring between DC and non-DC semi-transparent patches, I still think it's better to fix at least that low-hanging-fruit-part of the bug. (Also considering that presumably many studies will *either* have strike/dip/rake solutions *or* full moment tensors, depending on the used methodology)

N.b.: The mopad plotting routines also have a similar color blending issue (although the underlying patch seems to not always be a full circle)

**Before:** (color blend bug in both DC and non-DC plots)
alpha_before

**After:** (color blend bug fixed in DC plots, still present in non-DC plots)
alpha_after

2) There was a bug in plotting non-DC beachballs without fill colors. The DC plotting ends up in always two patches, until now one full circle and the pressure area path and using only the second patch for the outline is enough (it also includes the circle outline).
On the other hand, the non-DC plotting ends up with one full circle patch and 0-x paths for the simpler to draw areas whether pressure or tension. So far, for non-DC patches only the first complex paths was used and all eventual additional patches disregarded. There was a "need to check" comment in there that was never looked at again, stumbled over it while working on the first bug..

**Before:**
nofill_before

**After:**
nofill_after

Code used to generate the images:
```python
from obspy.imaging.beachball import beach

colors = ['g', 'r']
alpha = 0.4
width = 200

for nofill in (False, True):
    fig = plt.figure(figsize=(9, 9))
    ax = fig.add_subplot(111)
    yoffset = 0

    for focmec_ in ((40, 80, 10), [274, 13, 55], [0.91, -0.89, -0.02, 1.78, -1.55, 0.47]):
        b = beach(focmec_, xy=(50, yoffset), width=width, axes=ax, zorder=10,
                  alpha=alpha, facecolor=colors[0], bgcolor=colors[1],
                  nofill=nofill)
        ax.add_collection(b)
        b = beach(focmec_, xy=(220, yoffset), width=width, axes=ax, zorder=10,
                  alpha=alpha, facecolor=colors[0], bgcolor="w", nofill=nofill)
        ax.add_collection(b)
        yoffset += 250

    ax.set_xlim(-120, 350)
    ax.set_ylim(-120, 600)
    ax.text(-100, 0, "plot_dc()")
    ax.text(-100, 250, "plot_dc()")
    ax.text(-100, 500, "plot_mt()")
    ax.text(0, 120, "/".join(colors))
    ax.text(220, 120, "/".join((colors[0], "w")))

    fig.tight_layout()

plt.draw()
plt.show()
```


megies:
Good to go from my point of view.

megies:
Good to go for me.

QuLogic:
Is the same double-application problem affecting the edges as well? It's pretty evident in the outline-only plots.

megies:
Yes, the paths sometimes overlap. No easy fix there, though, that would take a rewrite of the complete algorithm. Not even worth thinking about, I would say, and it's not as bad as the color mixing, I think.

megies:
No other comments against this, so merging. It's an improvement in any case, I think.

PR #1429, opened 2016-06-30
Future's filter() method acts as imposter for obspy.signal.filter

megies:
Small bug and **only on Python 2.7**:

In `obspy/signal/__init__.py`, the line

```python
from future.builtins import *
```

imports a `filter()` method (just an alias to `itertools.ifilter`) - people just importing `obspy.signal` will thus find `obspy.signal.filter` which is not the expected filter method.

```python
import obspy.signal
obspy.signal.filter.lowpass(...)
```

thus results in 

```
type object 'itertools.ifilter' has no attribute ‘lowpass’. 
```

There are of course ways around this but it is still annoying and might confuse users so we should fix it.

### DIFF
--- a/obspy/signal/__init__.py
+++ b/obspy/signal/__init__.py
@@ -195,6 +195,9 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
 
+if "filter" in locals():
+    del filter
+
 import sys
 
 from obspy.core.util.deprecation_helpers import \


QuLogic:
Import everything as usual and then `del filter`?

megies:
Should be good to go @krischer.

megies:
@krischer, can you do a quick check if this fixes the issue for you. But I think this is good to go..

krischer:
One has to import `obspy.signal.filter` but then it works, just

```python
import obspy.signal
obspy.signal.filter.lowpass
```

Somehow gets caught by our deprecation thingies - well..good enough for now I think.

raphael31:
Dear Lion, I am also facing this bug on 1.0.1 obspy version for python 2.7.
I tried the "del filter" and the "import obspy.signal" solutions, none of these two are working.
Do you expect this to be solved on 1.0.2 version of obspy? or is there a solution that is not involving a modification of 1.0.1 that can work?

Thanks to all obspy developper anyway for their great job.

Raphael

krischer:
Hi Raphael,

you can workaround this in at least two ways:

(1) if you know which function exactly you want:

```python
from obspy.signal.filter import lowpass
```

or, (2) if you need more stuff from obspy.signal.filter:

```python
import obspy.signal.filter as obspy_filter
```

Cheers!

raphael31:
Thanks! 
Second option is the best for me, and it is working.
I did not got the name problem...

PR #1376, opened 2016-07-07
Reading some QuakeML files multiple times issues warning about updated ResourceIdentifier links

krischer:
Reading some QuakeML files multiple times results in a warning getting issued that is supposed to only show up when objects with same ResourceIdentifier string but different content get serialized:
```python
from StringIO import StringIO
from obspy import read_events
from obspy.core.event import ResourceIdentifier

sio = StringIO("""<?xml version='1.0' encoding='utf-8'?>
<q:quakeml xmlns:q="http://quakeml.org/xmlns/quakeml/1.2"
           xmlns="http://quakeml.org/xmlns/bed/1.2">
  <eventParameters publicID="smi:local/catalog">
    <event publicID="smi:local/event">
      <origin publicID="smi:local/origin">
        <time>
          <value>1970-01-01T00:00:00.000000Z</value>
        </time>
        <latitude>
          <value>0.0</value>
        </latitude>
        <longitude>
          <value>0.0</value>
        </longitude>
        <depth>
          <value>0.0</value>
        </depth>
        <arrival publicID="smi:local/arrival">
          <pickID>smi:local/pick</pickID>
          <phase>P</phase>
        </arrival>
      </origin>
    </event>
  </eventParameters>
</q:quakeml>
""")

cat1 = read_events(sio)
cat2 = read_events(sio)
```
```
/home/megies/git/obspy/obspy/core/event/base.pyc:324: UserWarning: The resource identifier 'smi:local/origin' already exists and points to another object: 'Origin(resource_id=ResourceIdentifier(id="smi:local/origin"), time=UTCDateTime(1970, 1, 1, 0, 0), longitude=0.0, latitude=0.0, depth=0.0)'.It will now point to the object referred to by the new resource identifier.
  Will return None if no object could be found.
```
The warning does not show up when removing the `Arrival` tag from the example QuakeML string.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -21,6 +21,9 @@
  - obspy.io.nlloc:
    * Use geographic coordinates from the NonLinLoc Hypocenter-Phase file if
      no custom coordinate converter is provided.
+ - obspy.io.quakeml:
+   * Fixed issue with improperly raised warnings when the same file is read
+     twice. (#1376)
  - obspy.io.sac:
    * Try to set SAC distances (dist, az, baz, gcarc) on read, if "lcalda" is
      true.  If "dist" header is found, distances aren't calculated.
--- a/obspy/io/quakeml/core.py
+++ b/obspy/io/quakeml/core.py
@@ -525,7 +525,7 @@ class Unpickler(object):
         self._extra(element, obj)
         return obj
 
-    def _origin(self, element):
+    def _origin(self, element, arrivals):
         """
         Converts an etree.Element into an Origin object.
 
@@ -541,7 +541,7 @@ class Unpickler(object):
         ... </origin>'''
         >>> xml_doc = etree.fromstring(XML)
         >>> unpickler = Unpickler(xml_doc)
-        >>> origin = unpickler._origin(xml_doc)
+        >>> origin = unpickler._origin(xml_doc, arrivals=[])
         >>> print(origin.latitude)
         34.23
         """
@@ -568,6 +568,7 @@ class Unpickler(object):
         obj.creation_info = self._creation_info(element)
         obj.comments = self._comments(element)
         obj.origin_uncertainty = self._origin_uncertainty(element)
+        obj.arrivals = arrivals
         obj.resource_id = element.get('publicID')
         self._extra(element, obj)
         return obj
@@ -935,12 +936,18 @@ class Unpickler(object):
             # origins
             event.origins = []
             for origin_el in self._xpath('origin', event_el):
-                origin = self._origin(origin_el)
-                # arrivals
-                origin.arrivals = []
+                # Have to be created before the origin is created to avoid a
+                # rare issue where a warning is read when the same event is
+                # read twice - the warnings does not occur if two referred
+                # to objects compare equal - for this the arrivals have to
+                # be bound to the event before the resource id is assigned.
+                arrivals = []
                 for arrival_el in self._xpath('arrival', origin_el):
                     arrival = self._arrival(arrival_el)
-                    origin.arrivals.append(arrival)
+                    arrivals.append(arrival)
+
+                origin = self._origin(origin_el, arrivals=arrivals)
+
                 # append origin with arrivals
                 event.origins.append(origin)
             # magnitudes
--- a/obspy/io/quakeml/tests/test_quakeml.py
+++ b/obspy/io/quakeml/tests/test_quakeml.py
@@ -951,6 +951,47 @@ class QuakeMLTestCase(unittest.TestCase):
         self.assertTrue(hasattr(cat, "nsmap"))
         self.assertEqual(getattr(cat, "nsmap")['ns0'], nsmap['ns0'])
 
+    def test_read_same_file_twice_to_same_variable(self):
+        """
+        Reading the same file twice to the same variable should not raise a
+        warning.
+        """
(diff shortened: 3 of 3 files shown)


krischer:
This issue is now a PR fixing the problem and adding a test.

When an existing resource identifier is created again, ObsPy checks the referred to objects for equality. If that fails it raises a warning as that resource identifier would then point to two different objects. Reading the same file twice does not trigger the warning as the objects should of course be identical.

For origins with arrivals this equality check happened before the arrivals where added the new origin - the old has arrivals - thus they did not compare equal. Now the arrivals are added before the equality test happens.

megies:
Did a dummy rebase and force-push to trigger Travis (no idea why it didnt start a build)..

In any case, works on my computer, good to merge, I think. :+1:

nackerley:
The fix seems to be to create the arrivals before the origin is created; why isn't it necessary also to create origins, magnitudes, station_magnitudes, picks etc. before the event is created? I think I'm seeing the same problem when re-reading QuakeML which includes these other components.

krischer:
Can you send us the problematic files? I agree that there are probably other cases where this treatment is necessary. It is a bit tricky to get the logic right but by now we should have caught most cases. We gain actually working resource identifiers which is IMHO totally worth the effort.

For the events its not an issue as they get the resource identifiers assigned after everything has been created. This triggers the comparison to existing registered resource identifiers.

nackerley:
Here you go: [rereading_quakeml.zip](https://github.com/obspy/obspy/files/600393/rereading_quakeml.zip)

Here at the Geological Survey of Canada we're currently developing tools to convert some older databases and pickfile formats to QuakeML (using ObsPy) so you will see lots of "extra" tags (a work in progress) in the .xml. I don't think they're a factor here. 

You can see the warnings that are thrown on my system in the .ipynb file. I haven't done a careful accounting but it seems like warnings are thrown on all Events and Picks but only some Origins and Arrivals.

krischer:
Great. Thanks! I'll look into it soon - the goal is to eliminate all unnecessary warnings and I think we should be pretty close to achieve that.

krischer:
So turns out this is actually a bit of a different issue. Many of the fields in your example files contain invalid floating points numbers. If you replace all occurrences of `nan` and `inf` in your example file the script works fine and does not raise a single warning and the two catalogs in the end can also be compared with the `==` operator.

According to your comment you created the file with an ObsPy script. ObsPy should of course refuse to write invalid floats - I'll follow this up in a separate issue.

There are also a bunch of other problems with the QuakeML file when you validate it. Please note that the QuakeML XSD file is useless and does not validate anything so you have to use the RelaxNG files to validate QuakeML files. `jing` is a tool that produces useful error messages when validating with RelaxNG schemas:

```bash
$ jing ~/workspace/code/obspy/obspy/io/quakeml/data/QuakeML-1.2.rng test.xml
```

Note that ObsPy can also validate QuakeML files upon writing (this does a correct validation - but the error messages are not necessarily helpful):

```python
cat.write("blup.xml", format="quakeml", validate=True)
```

This will refuse to write an invalid QuakeML so it is oftentimes a good idea to use this.

krischer:
@nackerley Please comment on #1597. Note that this result in ObsPy no longer being able to read your file but I think this is ok as it is not valid.

nackerley:
@krischer Thanks for the feedback! Indeed after scrubbing Inf and NaN from my xml the warnings disappear. I'll comment on #1597 separately.

Thanks also for the tips on validation. At the moment I'm having trouble getting https://github.com/relaxng/jing-trang installed on my system (I'll sort it out I'm sure) but I've noticed obspy.io.quakeml.core uses RelaxNG. I'll be honest, I don't even know what that means, but why isn't that good enough? It does give descriptive messages, though I am puzzled why my "eventParameters" are invalid:

```
from obspy.io.quakeml.core import _validate
_validate('20160714.xml', verbose=True)
```
>Error validating QuakeML file:
>	20160714.xml:3:0:ERROR:RELAXNGV:RELAXNG_ERR_EXTRACONTENT: Element quakeml has extra >content: eventParameters

It would be nice if obspy.core.event.Catalog.write() passed a "verbose" parameter on to obspy.io.quakeml.core._write_quakeml() and [obspy.io.quakeml.core._validate()] ...

krischer:
> I've noticed obspy.io.quakeml.core uses RelaxNG. I'll be honest, I don't even know what that means, but why isn't that good enough?

It is good enough for the validation but the error messages in more complex cases are not very good - or at least they used to - maybe lxml/libxml2 had some improvements? The error messages are probably good enough for wrong types but might be not as good for example for missing tags and more complex constraints.


> I don't even know what that means

It it is a different schema language for describing the layout of XML (in this case QuakeML) files. The default XSD schema cannot validate QuakeML as QuakeML made some choices that are not compatible with XSD.


> I am puzzled why my "eventParameters" are invalid:

Might be an issue where `jing` helps.


> It would be nice if obspy.core.event.Catalog.write() passed a "verbose" parameter on to obspy.io.quakeml.core._write_quakeml() and [obspy.io.quakeml.core._validate()] ...

Good idea - I'll reopen this ticket an rename it.



> I'm having trouble getting https://github.com/relaxng/jing-trang installed on my system

Most package managers on OSX and Linux have it. On Windows I don't know.

krischer:
I actually cannot reopen this...I'll make a new ticket.

PR #1470, opened 2016-07-08
ISC Event Client

krischer:
The ISC does not appear to have an official fdsnws event service - IRIS provides a mirror but that appears to be outdated (see this mailing list discussion for more details: http://lists.swapbytes.de/archives/obspy-users/2016-July/002093.html).

We could easily add an ISC client to ObsPy. The webinterface is here:

http://www.isc.ac.uk/iscbulletin/search/catalogue/

And it uses URL like this:

http://www.isc.ac.uk/cgi-bin/web-db-v4?request=COMPREHENSIVE&out_format=CATQuakeML&searchshape=RECT&bot_lat=&top_lat=&left_lon=&right_lon=&ctr_lat=&ctr_lon=&radius=&max_dist_units=deg&srn=&grn=&start_year=2013&start_month=7&start_day=31&start_time=00:00:00&end_year=2016&end_month=8&end_day=01&end_time=00:00:00&min_dep=&max_dep=&min_mag=6&max_mag=&req_mag_type=&req_mag_agcy=&include_links=on

It returns QuakeML and thus it should be really easy to implement.


krischer:
Thanks for the clarification!

This issue is now a PR changing the ISC mapping to point directly to the ISC fdsnws implementation. The IRIS mirror has been moved to a separate mapping.

krischer:
Both behave quite similar, although returned HTTP codes sometimes differ for some reason. They have a number of other issues that currently prevents them from working with ObsPy. I'll put some here but will write them an email as well.

This query returns a truncated QuakeML file:
http://www.isc.ac.uk/fdsnws/event/1/query?minmagnitude=7&starttime=2013-01-01

Encoded URLs don't work:
http://www.isc.ac.uk/fdsnws/event/1/query?starttime=2013-01-01T00%3A00%3A00.000000

Other oddities:

* The direct ISC one returns HTTP code 504 a lot, especially when requesting gzip compressed data.
* They also appear to have very aggressive rate limiting resulting in responses with HTTP code 429.
* In general the behavior seems to vary temporally quite a bit and sometimes data is returned and other times not.

CHANGELOG.txt
```diff
@@ -6,6 +6,8 @@
  - obspy.clients.fdsn:
    * Fixing issue with location codes potentially resulting in unwanted data
      to be requested. (see #1422)
+   * The ISC mapping now directly points to the ISC implementation. The IRIS
+     mirror	has been moved to a separate mapping (see #1470).
```
megies: Seems there's a tab in here..

megies:
Just noticed this is based on maintenence branch.. switching milestone.

krischer:
Yea but its kind of useless to merge this as the ISC service just does not work and they also don't answer my emails. Not sure what to do about this one.

megies:
Alright, I'll bump the milestone then..

PR #1457, opened 2016-07-10
SAC output does not use obspy headers

jkmacc-LANL:
io.sac should propagate relevant obspy headers (similar to #1317).
For example: Now (obspy 1.0.1), reading SAC files, rotating them and writing to SAC gives the wrong seed ids.

```
from obspy import read, Stream
from obspy.io.sac.sactrace import SACTrace

def simulate_SAC_write_and_read(stream):
    return Stream([SACTrace.from_obspy_trace(tr).to_obspy_trace() for tr in stream])

stream = simulate_SAC_write_and_read(read())
print stream[0].id
stream.rotate('ZNE->LQT', 10, 10)
print stream[0].id
stream = simulate_SAC_write_and_read(stream)
print stream[0].id
```

```
BW.RJOB..EHZ
BW.RJOB..EHL
BW.RJOB..EHZ
```

### DIFF
--- a/obspy/io/sac/tests/test_core.py
+++ b/obspy/io/sac/tests/test_core.py
@@ -33,6 +33,7 @@ class CoreTestCase(unittest.TestCase):
         self.file = os.path.join(self.path, 'data', 'test.sac')
         self.filexy = os.path.join(self.path, 'data', 'testxy.sac')
         self.filebe = os.path.join(self.path, 'data', 'test.sac.swap')
+        self.fileseis = os.path.join(self.path, "data", "seism.sac")
         self.testdata = np.array(
             [-8.74227766e-08, -3.09016973e-01,
              -5.87785363e-01, -8.09017122e-01, -9.51056600e-01,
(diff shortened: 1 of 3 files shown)


jkmacc-LANL:
This issue has been changed into a PR against `maintenance_1.0.x`.  

When running `SACTrace.from_obspy_header(tr)`, if an old SAC header is found and the following pairs or header values disagree, the `stats` header is chosen: `('kstnm', 'station'), ('knetwk', 'network'), ('kcmpnm', 'channel'), ('khole', 'location')` .

megies:
@trichter, can you please do a review of the changes?

krischer:
Looks solid to me but let's wait for @trichter.

trichter:
Looks also good from my point of view. And above example is working now. One thing: The only header propagated from SAC header to Stats but not propagated the other way Stats->SAC is now calib/scale. Do we want this one, too?

jkmacc-LANL:
I don't see why not; it seems like as good a candidate as the others.  I'll try to add it shortly.

jkmacc-LANL:
Hmmm...  All this adding information into the SAC header from `Stats` is messing up the read -> write -> read again `Trace` comparisons in the tests.  Though I'm not sure why the string tests didn't fail, adding a `scale` into the SAC header certainly makes it a "different file," which goes against the intent of some of the tests.  I'm not a huge fan of these tests, but I think I'm going to remove the `calib`/`scale` change so that this can proceed.  Thanks for your patience.

megies:
So, what's the status on this one? I'm afraid I'm not much into the SAC specifics..
If it's ready maybe we can squash it a bit?

jkmacc-LANL:
Sure, I can squash some of these.  Otherwise it should be good to go.

megies:
Ok, thanks for the fix. It sounded a bit like there might be need for a follow-up ticket? Anyway, I'll leave that up to you SAC guys..

PR #1472, opened 2016-07-12
Allow integers for SACTrace "iqual"

jkmacc-LANL:
Common use of the SAC "iqual" header is to hold user-determined application-specific integer quality values.  Strict enumerated strings are too restrictive for this use.  Pre-1.0 ObsPy versions made this simple, but `SACTrace` is rather strict about it.  This issue is to request that this functionality be restored, or at least made more accessible, to recognize this common use.  This will involve (optionally?) switching the setter function for the `SACTrace.iqual` property from an enum-type to a regular integer-type setter.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -65,6 +65,8 @@ master:
  - obspy.io.sac:
    * Try to set SAC distances (dist, az, baz, gcarc) on read, if "lcalda" is
      true.  If "dist" header is found, distances aren't calculated.
+   * SACTrace.iqual is no longer accepts enumerated string values, but
+     arbitrary integer values.
  - obspy.signal:
    * PPSD.plot(): fix plotting of percentiles, mode and mean and setting
      period limits when using "xaxis_frequency=True" (see #1406, #1416)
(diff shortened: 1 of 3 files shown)


jkmacc-LANL:
Thanks for responding, everyone.  This issue is now a pull request against `master`.  This very minor change simply swaps the `SACTrace.iqual` getter and setter from enum-type, which accepts known strings and converts them to known integer values, to an integer-type, which accepts and returns arbitrary integers.

jkmacc-LANL:
+DOCS

jkmacc-LANL:
Looks good to me!

krischer:
Thanks for this! I understand this changes the behavior a bit but on the other hand it also "fixes" the old behavior which apparently people did not expect. This IMHO makes it reasonable to merge into `releases` and not `master` so it ends up in `1.0.2`. Your call though.

No matter what - it still needs a changelog entry.

jkmacc-LANL:
Ah yes, of course.  Brain fart.  So I just rebase on `maintenance_1.0.x`, but how do I change the target branch for the PR?

krischer:
Just add a changelog entry here and I'll do it manually. Github unfortunately still does not allow the target branch to be changed and in this case its not worth it to open a new PR I think.

jkmacc-LANL:
Thanks, @krischer !

jkmacc-LANL:
Let me know if there's anything else I can do here.  There may be clashes in `sac/header.py` from my other PRs, but hopefully the fixes are obvious.

krischer:
Manually cherry picked onto the maintenance branch. Thanks again @jkmacc-LANL!

jkmacc-LANL:
Thanks, @krischer for fixing this.

krischer:
Well...you did all the actual work - i just executed a couple of git comands 😃

PR #1466, opened 2016-07-23
Write support for stationtxt format

megies:
We should add write support for stationtxt format to `Inventory`.

TODO:
- [x] write wrapper that writes station text representation to file
- [x] register as plugin
- [x] tests

### DIFF
--- a/obspy/io/stationtxt/core.py
+++ b/obspy/io/stationtxt/core.py
@@ -19,9 +19,11 @@ with standard_library.hooks():
 
 import csv
 import io
+import re
 import warnings
 
 import obspy
+from obspy import UTCDateTime
 from obspy.core.inventory import (Inventory, Network, Station, Channel,
                                   Response, Equipment, Site,
                                   InstrumentSensitivity)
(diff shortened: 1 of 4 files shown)


megies:
Bumping to 1.2

nick-iris:
What is the current status of the code you wrote for writing an Inventory as StationTXT?  I would be happy to help resolve any remaining issues to see this added in an upcoming release. Thanks.

megies:
@nick-iris feel free to add to this branch. And yes, I think only those items listed still are todo.

krischer:
Tests should be easy to add - the reading tests are fairly comprehensive and you just need to add a couple of round-trip tests.

https://github.com/obspy/obspy/blob/master/obspy/io/stationtxt/tests/test_station_text_parsing.py

nick-iris:
@megies @krischer I just pushed fixes for the remaining todo items for issue 1466. This includes tests for the stationtxt write method, registered the stationtxt write method in setup.py, and added a wrapper for the write method to write to a file or file-like object.

I think since I had rebased on top of master a bunch of newer commits were added as well. Sorry I didn't realize this would happen.

krischer:
I'm not entirely sure what you as its a bit messy now.If you had only rebased on top of the latest master then it would only show a few commits.

Can you try again? Worst case we have to cherry pick a couple of commits.

nick-iris:
@megies @krischer Should be fixed now. Sorry about that.

obspy/io/stationtxt/core.py
```diff
@@ -11,21 +11,24 @@
...
-from future.utils import native_str
-from future import standard_library
-with standard_library.hooks():
-    import collections
-
 import csv
 import io
+import re
 import warnings
+import six
```
krischer: `six` is not an ObsPy dependency. Please remove it and either use functionality from the `future` module or `obspy.core.compatibility`.

obspy/io/stationtxt/core.py
```diff
@@ -11,21 +11,24 @@
...
 from obspy.core.inventory import (Inventory, Network, Station, Channel,
                                   Response, Equipment, Site,
                                   InstrumentSensitivity)
 
+from future.builtins import *  # NOQA
+from future.utils import native_str
+from future import standard_library
+with standard_library.hooks():
+    import collections
+
```
krischer: All the `future` imports have to be at the beginning of the files. Our test will check that and raise an error otherwise.

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+        raise NotImplementedError("Unknown level: " + str(level))
     return inv
 
 
+def _format_time(value):
+    if isinstance(value, UTCDateTime):
+        if value.microsecond == 0:
+            return value.strftime("%Y-%m-%dT%H:%M:%S")
+        else:
+            return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
```
krischer: Out of curiosity: Why not put the `Z` utc specifier there?

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+
+
+def inventory_to_station_text(inventory_or_network):
+    """
+    Function to convert inventory or network to station text representation.
+
+    :type inventory_or_network:
+        :class:`~obspy.core.inventory.inventory.Inventory` or
+        :class:`~obspy.core.inventory.network.Network`
```
krischer: Please also add the `:param:` description.

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+    """
+    if isinstance(inventory_or_network, Inventory):
+        networks = inventory_or_network.networks
+    elif isinstance(inventory_or_network, Network):
+        networks = [inventory_or_network.networks]
+    else:
+        raise TypeError()
```
krischer: Error could be a bit more verbose.

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+
+    def _to_str(item):
+        if item is None:
+            return ""
+        x = str(item)
+        if isinstance(item, UTCDateTime):
+            x = x.rstrip("Z")
+            x = re.sub(r'\.0+$', '', x)
+            x = re.sub(r'T00:00:00$', '', x)
+        return x
```
krischer: Why does this not use the above `_format_time()` function?

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+        header = ("#Network|Station|Location|Channel|Latitude|Longitude|"
+                  "Elevation|Depth|Azimuth|Dip|SensorDescription|Scale|"
+                  "ScaleFreq|ScaleUnits|SampleRate|StartTime|EndTime")
+        lines = [header]
+        for net, sta, cha in items:
+            # omit stations with no channels and warn
+            if cha is None:
+                msg = ("Writing inventory with channel detail level and "
+                       "encountered some networks/stations without channels. "
+                       "These will be omitted in the output.")
```
krischer: Potentially add some details which networks/stations are being omitted?

obspy/io/stationtxt/tests/test_station_text_parsing.py
```diff
@@ -17,12 +17,14 @@
 import io
 import os
 import unittest
+import StringIO
```
krischer: Please use `io.StringIO` - pure `StringIO` is not compatible with Python 3.

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +251,136 @@ def _read(obj):
...
+            line = "|".join(_to_str(x) for x in (
+                net.code, sta.code, cha.location_code, cha.code,
+                cha.latitude is not None and cha.latitude or sta.latitude,
+                cha.longitude is not None and cha.longitude or sta.longitude,
+                cha.elevation is not None and cha.elevation or sta.elevation,
+                cha.depth, cha.azimuth, cha.dip,
+                cha.sensor and cha.sensor.type,
+                sensitivity and sensitivity.value,
+                sensitivity and sensitivity.frequency,
+                sensitivity and sensitivity.input_units,
```
krischer: This might be `False` if not present and the `_to_str()` function above will translate this to `"False"`. I don't think this is desired. Am I reading the code correctly? Because the test do seem to produce the correct result.

obspy/io/stationtxt/core.py
nick-iris: For better or worse it's without the `Z` utc specifier in the FDSN spec, so we need to comply with that.

chad-iris: Perhaps the ObsPy generation of this format could include a 'Z' specifier as an option?  As @nick-iris says we need a version sans-Z to use in an fdsnws-station implementation (which, regrettably, has no Z).

obspy/io/stationtxt/core.py
megies: Oh and please use a plain `import collections`, should work for Python 2.7, I think..

nick-iris:
Thanks for your review. I went ahead and made the necessary improvements, including adding a required `level` argument to the `write()` method. I also improved the error handling and tests. Let me know if you see anything else that needs to be changed or improved.

nick-iris:
Would you like to see any additional changes, or a different approach all together? I'm not sure what to make of the failing codecov and docker-testbot tests. The other tests are passing. Thanks.

obspy/io/stationtxt/core.py
```diff
@@ -248,10 +247,150 @@ def _read(obj):
...
+        x = str(item)
+        if isinstance(item, UTCDateTime):
+            x = _format_time(item)
+        return x
+
+    items = []  # list of items to write
+
+    # Write items at to the requested level of detail. Raises a ValueError if
+    # insufficient information is present for the requested level of detail.
+    if level == "NETWORK":
```
megies: Can you add `level = level.upper()` before this switch? We're also case insensitive e.g. when specifying the write output format.. for `Stream.write(.., format='mseed'`)

nick-iris:
Thanks @megies  and @krischer. I added a new changelog entry and made the level parameter case insensitive. Tests are passing, so this should be okay to merge.

megies:
I did some more minor fixes, should be good to go now..

krischer:
Rebased, force pushed, and added a minor commit. IMHO ready to merge if CI is alright.

krischer:
All test failures are some unrelated code formatting checks. I'll fix it directly in master.

krischer:
Thanks everyone!

megies:
Thanks for review and final touches @krischer, thanks @nick-iris for working on this!

nick-iris:
@megies I noticed that we are writing the _cha.sensor.type_ value for the channel level StationTXT SensorDescription, when we should be using _cha.sensor.description_.

https://github.com/obspy/obspy/blob/master/obspy/io/stationtxt/core.py#L367

Is it okay for me change the this line to instead use _cha.sensor.description_?

megies:
@nick-iris Yeah sure, go ahead with another PR off of master.

PR #1479, opened 2016-07-25
Make sure newer CMTSOLUTION files can also be read.

krischer:
This is a small fix that makes sure `obspy.io.cmtsolution` can also read newer CMTSOLUTION files which was not possible before due to some assumptions in the code.

I hope this can still make it into 1.0.2 but the fix is simple so it should work out.


krischer:
IMHO ready to merge.

PR #1484, opened 2016-07-29
nlloc: read more additional information from .hyp files

megies:
..namely SIGNATURE (author information) and COMMENT fields in .hyp file.


megies:
Should be good to merge..

PR #1487, opened 2016-07-30
docker tests: mixed improvements

megies:
### Summary

 - remove old logs when running docker tests (to avoid huge space consumption of log directories), only keep the last 15 log dirs
 - don't run flake8 tests (flake8 is too version-dependent and it's tested in Travis anyway, which is sufficient)
 - docker test runner script can now be used to test an arbitrary fork/commit combination (to test PRs)
 - additional parameters can now be passed down into `obspy-runtests` (e.g. to only test a single module)
 - install pyimgur before testing (for image fail uploads)
 - set github commit status with overall result across all docker images
 - add script to fetch all commits that are the tip of an open PR that does not have a commit status with "docker-testbot" context
 - add an example cronjob to run docker tests on all open+updated PRs

### DIFF
--- a/misc/docker_tests/run_obspy_tests.sh
+++ b/misc/docker_tests/run_obspy_tests.sh
@@ -128,7 +128,7 @@ run_tests_on_image () {
     image_name=$1;
     printf "\n\e[101m\e[30m  >>> Running tests for image '"$image_name"'...\e[0m\n"
     # Copy dockerfile and render template.
-    sed "s/{{IMAGE_NAME}}/$image_name/g; s/{{EXTRA_ARGS}}/$extra_args/g" scripts/Dockerfile_run_tests.tmpl > $TEMP_PATH/Dockerfile
+    sed "s#{{IMAGE_NAME}}#$image_name#g; s#{{EXTRA_ARGS}}#$extra_args#g" scripts/Dockerfile_run_tests.tmpl > $TEMP_PATH/Dockerfile
 
     # Where to save the logs, and a random ID for the containers.
     LOG_DIR=${LOG_DIR_BASE}/$image_name
--- a/misc/docker_tests/run_obspy_tests.sh
+++ b/misc/docker_tests/run_obspy_tests.sh
@@ -4,9 +4,9 @@ DATETIME=$(date -u +"%Y-%m-%dT%H-%M-%SZ")
 LOG_DIR_BASE=logs/$DATETIME
 mkdir -p $LOG_DIR_BASE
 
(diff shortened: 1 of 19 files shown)


megies:
Installing pyimgur before running tests in docker images works fine it seems: http://tests.obspy.org/46251/#1

megies:
@krischer, ready to be merged after 1.0.2, I think. If the pyimgur installation should be done in a different spot, please adapt. Also, I noticed that test images are not copied to logs directory as intended by the docker scripts.

megies:
@krischer, good to merge, I think, when CI goes green.

misc/docker_tests/scripts/install_and_run_tests_on_image.sh
```diff
@@ -3,6 +3,8 @@ red='\e[0;31m'
 green='\e[0;32m'
 no_color='\e[0m'
 
+pip install https://github.com/Damgaard/PyImgur/archive/master.zip
```
krischer: I think the docker files are probably a better place for this.

misc/docker_tests/run_obspy_tests.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
+
+# remove log directories older than 14 days (to avoid cluttering with GB of
+# data on buildbots that are run regularly)
+find logs/2[0-9][0-9][0-9]-[01][0-9]-[0-3][0-9]T[012][0-9]-[0-5][0-9]-[0-5][0-9]Z -type d -not -newermt `date --date='14 days ago' --rfc-3339=date` -exec rm -rf {} \;
```
krischer: The `find` command is weird on OSX. I'll check if it works once I get my docker installation working again...

megies:
Docker tests can now also be run on a arbitrary fork/commit combination, e.g.:
```$bash
$ ./run_obspy_tests.sh -ttrichter:e6da3ddb5
```

What doesn't work right now though is the part that selects which images to run if none are provided on command line. In that case all images are skipped right now..

krischer:
> What doesn't work right now though is the part that selects which images to run if none are provided on command line. In that case all images are skipped right now..

I think I fixed it. Everything seems to work as expected now. Even combinations of all parameters:

```bash
./run_obspy_tests.sh -ttrichter:e6da3ddb5 -eio.mseed fedora_24
```

megies:
The docker testbot VM is now set up to once per day test any yet-untested open PR tips and report a commit status (like in here).

PR #1490, opened 2016-07-31
Add source of documentation to release tarball

megies:
Please consider updating your `MANIFEST.in` to include the source of the documentation to the release tarball, so that it can be generated during the packaging.

Cheers,


megies:
@ghisvail [here's a sdist tarball built from this PR](https://www.geophysik.uni-muenchen.de/~megies/obspy-1.0.2rc6.post0+1.gfb3b6919bf.obspy.add.docs.to.sdist.zip), including docs source.

ghisvail:
@megies great :+1:

With this, I should be able to provide a documentation package for `obspy` starting from the next release. How far along do you think the next release is ?

krischer:
Great :-) It also only increases the size of the release by three MB which is totally worth it IMHO.

> How far along do you think the next release is ?

A few days at most. Hopefully today ;-)

PR #1494, opened 2016-08-01
sample win32 file and dataformat

xiansch:
The win file is a downsampled (20Hz) file, 1 hour, 3 channels


krischer:
@ThomasLecocq I took the liberty to assign you to this problem ;-)

ThomasLecocq:
@krischer Okay, I should have some time to have a look this week...

megies:
@xiansch can you crop the test file abit? It's 450kb right now and should ideally be something like 10kb.

xiansch:
Hello!  I'm very sorry I made a mistake previously, I was given a basic win file by a colleague to use and I thought it was a win32 file but actually it is the regular win file.  So I uploaded 1 minute versions of both files (win and win32, they are the same data, and the regular win can be read with the datamark io, the win32 cannot...although apparently it is an extension of the basic win file)
I actually don't know how to trim the data time-wise using the available win tools so I downloaded a new 1 minute file, to get to ~10kb means extracting a single channel.  Otherwise the data is ~30kb, 1 station with 3 channels, so that "extracting a single channel" is possible...

barsch:
closing- see https://github.com/obspy/obspy/pull/1692 - clsoing and deleting branch as long we don't know whats the origin of this files - please consider adding other samples which were not produces by Hi-net to https://github.com/obspy/obspy/pull/1692

PR #1496, opened 2016-08-03
Fix pep8 builds against newer flake8.

QuLogic:
This should fix the `AttributeError: 'module' object has no attribute 'DEFAULT_CONFIG'` error that is raised with latest `flake8`.

`flake8` 2.5.5 sets `pep8.USER_CONFIG` itself without passing `config_file` instead.

### DIFF
--- a/misc/docs/make_pep8.py
+++ b/misc/docs/make_pep8.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 
 import os
+from collections import Counter
+from distutils.version import LooseVersion
 from shutil import copyfile
 
 import obspy
-from obspy.core.util.testing import check_flake8
+from obspy.core.util.testing import check_flake8, flake8_version
 
 
 ROOT = os.path.dirname(__file__)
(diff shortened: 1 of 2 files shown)


krischer:
`flake8` 3.x which is also already out unfortunately changed the API once again. How about we handle it like here:

https://github.com/krischer/mtspec/blob/master/mtspec/tests/test_code_formatting.py

It will just skip the code formatting tests for flake8 versions < 3.0. We'd just have to make sure to install the lastest flake8 with pip in the CI runners.

megies:
> How about we handle it like here:
> 
> https://github.com/krischer/mtspec/blob/master/mtspec/tests/test_code_formatting.py
> 
> It will just skip the code formatting tests for flake8 versions < 3.0. We'd just have to make sure to install the lastest flake8 with pip in the CI runners.

I really only care about CI running flake8, everything else doesn't matter for me.
So, totally :+1: as soon as CI can run flake8 3.x (I don't see it in anaconda cloud yet?).

krischer:
pypi already has `flake8` 3.0.3. We can just use `pip` to install it - no need to use `conda`.

krischer:
More work then I would have done but looks good to me. Do we still use the pep8 checker during docs building?

QuLogic:
Yes, PEP8 results are posted [here](https://docs.obspy.org/pep8/index.html).

PR #1495, opened 2016-08-04
Noise records cause full seed truncation

krischer:
I'm trying to read in full seed volumes (generated by IRIS) using obspy.read. 

Example;
st=obspy.read('/data1/LVC/IRIS/ashton.LVC.IRIS.IM.NV32.2008.2009.BHE.553754.seed')

output;
/usr/lib/python3/dist-packages/obspy/io/mseed/core.py:384: InternalMSEEDReadingWarning: readMSEEDBuffer(): Record starting at offset 633344 is not valid SEED. The rest of the file will not be read. The file contains a 20480 byte dataless part at the beginning. Make sure to add that to the reported offset to get the actual location in the file.
  warnings.warn(*_i)

and if we look at the stream;
2 Trace(s) in Stream:
IM.NV32..BHE | 2008-01-03T18:44:46.000000Z - 2008-01-03T18:45:31.975000Z | 40.0 Hz, 1840 samples
IM.NV32..BHE | 2008-01-03T18:45:33.000000Z - 2008-01-03T23:59:59.975000Z | 40.0 Hz, 754680 samples

Which is not all of the data.....

Using rdseed to get the times; 
$cat IM.NV32.2008.2009.BHE.times
Time series: 
  Rec# Sta   Cha Net Loc      Start Time              End Time         Sample Rate Tot Samples
     6 NV32  BHE IM      M 2008,003,18:44:46.0000 2008,004,03:59:44.9750  40.0000000 1331920
   278 NV32  BHE IM      M 2008,004,03:59:50.0000 2008,004,04:07:23.9750  40.0000000  18160
   283 NV32  BHE IM      M 2008,005,18:49:21.0000 2008,006,03:24:31.9750  40.0000000 1236360
   536 NV32  BHE IM      M 2008,006,21:16:19.0000 2008,007,01:39:38.9750  40.0000000 631960
   666 NV32  BHE IM      M 2008,007,17:35:45.0000 2008,007,17:36:30.9750  40.0000000   1840
   666 NV32  BHE IM      M 2008,007,17:36:32.0000 2008,008,04:58:11.9750  40.0000000 1636000

This is only a 4 Mb seed volume, and it happens on all my volumes (which have sizes from 1 Mb to 16 Gb).

The end goal is to go from a year seed volume to a year ppsd directly in python.

ashton


CTrabant:
> Interesting - I guess there is always something new to learn :-) Thanks for the explanation.
> 
> Are these always 512 bytes long? This is not really specified in the SEED manual.
> 
I think they can be any length, I wouldn't count on only 512-byte versions.  We also have 1024-byte records and those probably get padded with 1024-byte noise/blank records.

krischer:
> I think they can be any length, I wouldn't count on only 512-byte versions. We also have 1024-byte records and those probably get padded with 1024-byte noise/blank records.

This unfortunately made the implementation a bit ugly - it now checks if the first 48 bytes of a record are either a noise record or just empty (zero or an ASCII space) - in that case it skips to the next 128 bytes (minimum record length) and tries again. The file format detection routine also had to be changed to be able to deal with noise records at the beginning of a file.

It should now work with any combination, length, and position of noise records. Additionally the `_is_mseed()` method can now directly work with open files and buffers.

Please review.

krischer:
@ashtonflinders ObsPy is now able to read your file if you update either to the latest master or the tip of the `maintenance_1.0.x` branch. It will flow into ObsPy 1.0.3 but this will still take some time to arrive.

Keep in mind that #1419 still exists and thus there are some issues for very large files. But the other issue also contains a workaround for that which already works right now.

PR #1502, opened 2016-08-08
fix section plot in master

megies:
`master` was broken recently by merging `maintenance_1.0.x` due to an unlucky combination of #1383 (on maintenance_1.0.x) and #1445 (on master). This should get master back to building OK.


krischer:
One of the two fill_between's does not work on Debian 8 for some reason:

http://tests.obspy.org/48219/

First error caught by the docker tests CI :-)

megies:
Myup, docker testbot already being useful.. :-)
I'm at it already, seems like a matplotlib 1.4.2 specific bug.

PR #1350, opened 2016-08-19
FDSN Mass Downloader supporting restricted data

krischer:
It could be useful if "FDSN Mass Downloader" would support downloading restricted data apart from public one.

Could be something like this:

`providers = [`
`[http://fdsn.service, username, password],`
`[http://fdsn.service2, username2, password2],`
`[http://fdsn.service3, username3, password3]`
`]`

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/__init__.py
+++ b/obspy/clients/fdsn/mass_downloader/__init__.py
@@ -18,6 +18,11 @@ component integrated into a bigger project.
     (https://www.gnu.org/copyleft/lesser.html)
 
 
+.. contents:: Contents
+    :local:
+    :depth: 2
+
+
 Why Would You Want to Use This?
 -------------------------------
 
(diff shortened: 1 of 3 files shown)


krischer:
This is now a PR including tests and documentation. It now works like this:

```python

client_orfeus = Client("ORFEUS", user="random", password="some_pw")
client_eth = Client("ETH", user="from_me", password="to_you")

mdl = MassDownloader(providers=[client_orfeus, "IRIS", client_eth])
```

This is nice because its quite general and will also work if we potentially support other authentication methods in the future. It will just reuse the passed client instance.

krischer:
No reviews yet but I'm merging as this is a small change, mostly on a part of ObsPy that is my doing, and it has a test :-)

megies:
@krischer, changelog maybe?

krischer:
Oops. My bad - I just added it to the latest master.

Issue #1425, opened 2016-05-27
reftime parameter for trim, slice and cutout


trichter:
From time to time I am writing code like

```
for tr in stream:
    tr.trim(tr.stats.starttime + 10, tr.stats.starttime + 50)
```

It'd be nice to just write

```
stream.trim(10, 50, reftime='starttime')
```

reftime would be an attribute of stats holding a UTCDateTime instance. It could also accept UTCDateTime itself.

Any opinions?

krischer:
You can almost do this already: if you pass integers/floats they will be interpreted as offsets to start- and end time.

But I do like the idea - a general way of doing this is cleaner! I'm a bit torn about a magical string parameter that gets resolved to a time while it is run but I don't see another way to do it for multi-trace streams.

trichter:
Oh, didn't expect that. Then, some code in Stream.trim will raise an Exception:

```python
In [13]: read().trim(5, 5, nearest_sample=True)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-13-5465512edf03> in <module>()
----> 1 read().trim(5, 5, nearest_sample=True)

/home/richter/anaconda/lib/python2.7/site-packages/obspy/core/stream.pyc in trim(self, starttime, endtime, pad, nearest_sample, fill_value)
   1507             if starttime:
   1508                 delta = compatibility.round_away(
-> 1509                     (starttime - tr.stats.starttime) * tr.stats.sampling_rate)
   1510                 starttime = tr.stats.starttime + delta * tr.stats.delta
   1511             if endtime:

TypeError: unsupported operand type(s) for -: 'int' and 'UTCDateTime'
```

The suggested interface would not be compatible with the current interpretation of numbers for start/endtime. (Both values relative to one reftime vs offsets to start and end time)

megies:
Hmm.. I'm not super convinced this is really necessary. And I'm also a bit wary about the string magic.

trichter:
It's handy if you've got traces from multiple events, but I don't have any strong opinion weather it should be included or not.

krischer:
> Oh, didn't expect that. Then, some code in Stream.trim raises an Exception:

Ah that's annoying...it works with `Trace.trim()` though. We should probably fix at least that.

claudiodsf:
Upvoted

trichter:
Another related bug

```
In [3]: read()[0].trim(3, 2)
---------------------------------------------------------------------------
[...]

Issue #1311, opened 2016-03-03
for trace.trim() - get left, right, or nearest sample


Jollyfant:
Take this example trace and applying `.trim(starttime=S, endtime=E, nearest_sample=False)`:

`1 --S- 2 ---- 3 ---- 4 -E-- 5` where requested S = starttime; E = endtime.

Setting `nearest_sample=False` will always fall to the sample **right** of the starttime, and **left** of the endtime. So this would return a trace of samples `[2 - 4]`. I'm really looking for a way to find the first sample that is left of the starttime. 

I tried subtracting delta from sample 2, but that implicitly assumes that the sample rate is constant for the entire stream. And it requires another read of the same data.. Any suggestions on how to do this? Maybe it would be good to choose either ("left", "nearest", "right") for the cutoff, instead of only specifying `nearest_sample=False`.

megies:
> I tried subtracting delta from sample 2, but that implicitly assumes that the sample rate is constant for the entire stream. And it requires another read of the same data.. Any suggestions on how to do this?

Should be easily doable in a `for tr in st:` loop, as each trace has one single sampling rate.

But, the logic of `left`, `nearest`, `right` also kind of makes sense to me, so I'm also OK if we implement that. `nearest_sample` comes pretty late in the call syntax, so I guess it is safe to assume nobody uses it as a positional arg and it would be safe to handle a changed call syntax with a simple deprecation handling/warning if replace that switch.

Other opinions?

Jollyfant:
It seems to be a matter of adding one index to delta in `tr._rtrim`, and subtracting one index in `tr_ltrim`. The change is simple, but it would require a lot of careful refactoring in the `trace` and `stream` classes to set the new flags.

    delta = int(math.floor(round((self.stats.starttime - starttime) * self.stats.sampling_rate, 7))) * -1
    if cutoff_start == "left":
        delta -= 1 # Force previous index

    delta = int(math.floor(round((endtime - self.stats.endtime) * self.stats.sampling_rate, 7)))
    if cutoff_end == "right":
        delta += 1 # Force next index

Needs two flags, or we can use a cutoff tuple ("left", "nearest") for (start, end) respectively.

megies:
In any case, as I mentioned, for now you could easily do something like..
```python
for tr in st:
    delta = tr.stats.delta
    tr.trim(t1-delta, t2+delta, nearest_sample=False)
```

[...]

Issue #1288, opened 2016-02-23
Add possibility to exclude last sample from methods Trace.slice(), Trace.trim() 


petrrr:
Here I am proposing to add a parameter to the two methods, which would allow to exclude the last limit sample of the selected time window. This would be sometimes quite useful when iteration over consecutive time windows, details matter and one would like to avoid including the limiting sample twice. 

I do not see an elegant workaround, and think adding this little feature would be quite useful to others. 

Not sure if there might be interest in excluding the starting limit sample as well would be interesting as well, maybe for "symmetry"? Probably this hypothetical parameter should apply to `Trace.slide()` method as well. 

Any comments?

krischer:
One just has to be careful what happens when the endtime is exactly at a sample. Also how this works with the `nearest_sample` parameter should be discussed. 

By and large and I think this is a good idea - the parameter could be called `exclude_last_sample` and should of course default to `False`.

petrrr:
I guess, there should be something well defined (or controllable) tolerance, where the sample would be regarded aligned with the limit. Otherwise the sample falls either inside or outside the intervall.

Okay, will try to look into this ASAP, not sure when.

megies:
I've also come across a similar use case now..
For me it would be interesting to have an option to leave away the last sample conditionally, if (and only if) the last sample would match exatly the given time, i.e. if..
```
part1 = tr.slice(endtime=t)
part2 = tr.slice(starttime=t)
```
..would lead to one sample being part of both slices.

Jollyfant:
In the QC module I always subtract 1μs from the endtime and use `nearest_sample=False` to make sure I'm not including a sample located on the endtime.

megies:
Yeah, something like that works too of course, but it's a bit hacky, and looks a bit cryptic in the code to read, so we still might want to think about adding some API for this.

PR #1508, opened 2016-08-22
trim/slice nearest_sample argument docstring wrong.

megies:
We document the `nearest_sample` argument like this:

screen shot 2016-08-19 at 00 51 48

Thus setting it to `False` should select the outer samples but we indeed appear to be selecting the inner sample, which we also test:

```python
    def test_slice_nearest_sample(self):
        """
        Tests slicing with the nearest sample flag set to on or off.
        """
        tr = Trace(data=np.arange(6))
        # Samples at:
        # 0       10       20       30       40       50
        tr.stats.sampling_rate = 0.1

        # Nearest sample flag defaults to true.
        tr2 = tr.slice(UTCDateTime(4), UTCDateTime(44))
        self.assertEqual(tr2.stats.starttime, UTCDateTime(0))
        self.assertEqual(tr2.stats.endtime, UTCDateTime(40))

        tr2 = tr.slice(UTCDateTime(8), UTCDateTime(48))
        self.assertEqual(tr2.stats.starttime, UTCDateTime(10))
        self.assertEqual(tr2.stats.endtime, UTCDateTime(50))

        # Setting it to False changes the returned values.
        tr2 = tr.slice(UTCDateTime(4), UTCDateTime(44), nearest_sample=False)
        self.assertEqual(tr2.stats.starttime, UTCDateTime(10))
        self.assertEqual(tr2.stats.endtime, UTCDateTime(40))

        tr2 = tr.slice(UTCDateTime(8), UTCDateTime(48), nearest_sample=False)
        self.assertEqual(tr2.stats.starttime, UTCDateTime(10))
        self.assertEqual(tr2.stats.endtime, UTCDateTime(40))
```

We have to either change the documentation or the implementation. I suggest to change it to

`Trace.trim(..., select_sample="nearest", ...)` with `"nearest"`, `"inner"`, `"outer"` as possible choices. This of course requires some deprecation fallbacks. Thoughts?

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1458,17 +1458,18 @@ class Stream(object):
             given ``fill_value``. Defaults to ``False``.
         :type nearest_sample: bool, optional
         :param nearest_sample: If set to ``True``, the closest sample is
-            selected, if set to ``False``, the outer (previous sample for a
-            start time border, next sample for an end time border) sample
+            selected, if set to ``False``, the inner (next sample for a
+            start time border, previous sample for an end time border) sample
             containing the time is selected. Defaults to ``True``.
 
-            Given the following trace containing 4 samples, "|" are the
+            Given the following trace containing 6 samples, "|" are the
             sample points, "A" is the requested starttime::
 
-                |        A|         |         |
+                |         |A        |         |       B |         |
+                1         2         3         4         5         6
 
-            ``nearest_sample=True`` will select the second sample point,
-            ``nearest_sample=False`` will select the first sample point.
+            ``nearest_sample=True`` will select samples 2-5,
+            ``nearest_sample=False`` will select samples 3-4 only.
 
         :type fill_value: int, float or ``None``, optional
         :param fill_value: Fill value for gaps. Defaults to ``None``. Traces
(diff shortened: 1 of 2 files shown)


megies:
Documentation is wrong, proposing a fix..

Changes in the API should be done in a PR against master. `inner` and `outer` is good but might not cover all desired cases (e.g. `inner` at start, `outer` at end?).. also see #1311, #1288, #1425 which are related. But in this issue only the docs should be fixed against `maintenance_1.0.x` without an API change..

krischer:
Hmm. I guess we need all the choices: ``"nearest"``, ``"inner"``, ``"outer"``, ``"left"``, ``"right"``. But yea - Fixing the documentation is good for the maintenance branch.

Travis just bugged out - I restarted the one test case.

megies:
Forgot to trigger docs, to be able to give them a look, so.. +DOCS

megies:
Good to merge, I think. API changes will have to be handled in a follow-up or one of the already open issues.

krischer:
The docs need to be changed for `trim`, `slice`, and `slide` in `Trace` as well as `Stream` ;-)

megies:
Should be fixed now..

PR #1507, opened 2016-08-22
Default SAC boolean headers when writing ObsPy trace?

jkmacc-LANL:
I'm not sure if this is expected behavior or not, but I at least was not expecting it. I could change this myself but I'd rather have the sac experts have a look.

**Snippet:**

```python
import obspy

obspy.read()[0].write("test.sac", format="sac")
tr = obspy.read("test.sac")[0]

print("LPSPOL:", tr.stats.sac.lpspol)
print("LCALDA:", tr.stats.sac.lcalda)
```

**Output:**

```
LPSPOL: 0
LCALDA: 1
```

From the SAC documentation:

* `LPSPOL`: TRUE if station components have a positive polarity (left-hand rule).

* `LCALDA`: TRUE if DIST, AZ, BAZ, and GCARC are to be calculated from station and event coordinates.

In my opinion both should have the opposite default values. This is actually the case in `SACTrace.__init__()` but this is not adopted during the conversion from the ObsPy to the SAC headers.

Thoughts?


jkmacc-LANL:
This issue is now a PR, which simply changes the default `SACTrace.lpspol` and `SACTrace.lcalda` values to `True` and `False`, respectively, when converting from a `Trace` that doesn't have a `trace.stats.sac`.  Other regression cases other than this one are requested.

obspy/io/sac/tests/test_core.py
```diff
@@ -874,6 +874,16 @@ def test_not_ascii(self):
         self.assertEqual(tr.stats.channel, 'HHE')
         self.assertEqual(tr.stats.network, '')
 
+    def test_sac_booleans_from_trace(self):
+        """
+        SAC booleans `lcalda` and `lpspol` should be `False` and `True`,
+        respectively, by default when converting from a `Trace`.
```
krischer: Minor nitpick and it does not matter at all for the test here: restructured text requires double backticks ``` `` ``` for verbatim formatting which unfortunately differs from markdown.

krischer:
Thanks for this - Looks good to me! It was originally discovered in the issue linked above precisely when changing from `0.10.2` to `1.0.2`. Maybe add a changelog item as it will affect many produces sac files?

jkmacc-LANL:
@krischer I did look at the function changed in the instaseis issue, but it wasn't clear where the `Stream` came from, as the function just accepts it as an argument.  Are they built from scratch?  If so, this change should work for that case.

krischer:
Yes they are built from scratch by just manually setting a couple of headers. In instaseis I fixed it for now by directly also settings the two boolean headers.

obspy/io/sac/tests/test_core.py
jkmacc-LANL: Thank you!  I just don't want to learn ReST, apparently :stuck_out_tongue_winking_eye: .

jkmacc-LANL:
* [x] Update changelog.
* [x] Reduce commit count.
* [x] Good-to-go.

PR #1511, opened 2016-08-26
Include sub seconds in stationxml write.

dsentinel:
Currently stationxml.write() truncates to the second like so:
<Created>2013-12-11T16:29:13+00:00</Created>
This is ok for creation time, but for station and channel startdate and enddate sub seconds may be needed when metadata changes, while recording is continuous. 
startDate="2013-12-11T16:29:13.500000+00:00"
Dataless has subseconds.

### DIFF
--- a/obspy/io/stationxml/tests/test_stationxml.py
+++ b/obspy/io/stationxml/tests/test_stationxml.py
@@ -152,19 +152,9 @@ class StationXMLTestCase(unittest.TestCase):
         # writing of the ObsPy related tags to ease testing.
         file_buffer = io.BytesIO()
 
-        # XXX helper variable to debug writing the full random file, set True
-        # XXX for debug output
-        write_debug_output = False
-
         inv.write(file_buffer, format="StationXML",
-                  validate=(not write_debug_output),
                   _suppress_module_tags=True)
         file_buffer.seek(0, 0)
-
-        if write_debug_output:
-            with open("/tmp/debugout.xml", "wb") as open_file:
-                open_file.write(file_buffer.read())
-            file_buffer.seek(0, 0)
 
         with open(filename, "rb") as open_file:
             expected_xml_file_buffer = io.BytesIO(open_file.read())
--- a/obspy/io/stationxml/tests/test_stationxml.py
+++ b/obspy/io/stationxml/tests/test_stationxml.py
@@ -129,19 +129,9 @@ class StationXMLTestCase(unittest.TestCase):
         # writing of the ObsPy related tags to ease testing.
         file_buffer = io.BytesIO()
 
-        # XXX helper variable to debug writing the full random file, set True
-        # XXX for debug output
-        write_debug_output = False
-
         inv.write(file_buffer, format="StationXML",
-                  validate=(not write_debug_output),
                   _suppress_module_tags=True)
         file_buffer.seek(0, 0)
-
-        if write_debug_output:
-            with open("/tmp/debugout.xml", "wb") as open_file:
-                open_file.write(file_buffer.read())
-            file_buffer.seek(0, 0)
 
         with open(filename, "rb") as open_file:
             expected_xml_file_buffer = io.BytesIO(open_file.read())
--- a/obspy/io/stationxml/tests/test_stationxml.py
+++ b/obspy/io/stationxml/tests/test_stationxml.py
@@ -22,6 +22,7 @@ import unittest
 
 import obspy
 from obspy.core.inventory import Inventory, Network
+import obspy.io.stationxml.core
 
 
 class StationXMLTestCase(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


megies:
While the notion of changing metadata without an interruption in continuous recording is pretty strange to me, of course we should include sub-seconds in time information for the sake of exactness.

@dsentinel, you will need to adapt a few test files, see failing tests on e.g Travis.

dsentinel:
I agree it's strange. I posed it in the general form.
The case came to me when I gave and end time with subseconds and the StationXML had truncated. Leaving a fraction of a second without metadata, which may not be terrible, but archives really don't like data that isn't described, and our QC would complain as well.

If this solution is acceptable, I'll happily adjust tests and changelogs. I left the test, perhaps inappropriately, as it showed a concise before and after.

krischer:
Sure - this is a good change (assuming of course its valid StationXML but I think it is). Please go ahead and fix the tests :-) You could also consider to add a round-trip test for the microseconds, e.g. a test that writes and then again reads times with microseconds precision.

dsentinel:
Began writing tests, and hung up on this. @megies confirmed he saw the same.
Should a bug be opened?

```
ERROR: test_is_stationxml (__main__.StationXMLTestCase)
  File "obspy/io/stationxml/tests/test_stationxml.py", line 77, in test_is_stationxml
    self.assertTrue(obspy.io.stationxml.core._is_stationxml(stat))
AttributeError: 'module' object has no attribute 'stationxml'

Ran 12 tests in 0.357s```

**Comment by krischer** [comment]
How did you execute the test?

**Comment by dsentinel** [comment]
```
python obspy/io/stationxml/tests/test_stationxml.py
```
I do it so, as it's the simplest way from my IDE's debugger to debug one suite of tests. I have done this with other tests without problems.
```
$ python runtests.py -v io.stationxml
```
Works for @megies, as well as for me.
Similar on the command line; I can import obspy.io.stationxml, but can not import obspy and access io.stationxml:
```
In [1]: import obspy

In [2]: obspy.io. <tab>
obspy.io.absolute_import   obspy.io.dict              obspy.io.kinemetrics       obspy.io.object            obspy.io.round             obspy.io.unicode_literals
obspy.io.ah                obspy.io.division          obspy.io.kml               obspy.io.oct               obspy.io.sac               obspy.io.wav
obspy.io.ascii             obspy.io.filter            obspy.io.list              obspy.io.open              obspy.io.seg2              obspy.io.y
obspy.io.bytes             obspy.io.gcf               obspy.io.map               obspy.io.pdas              obspy.io.segy              obspy.io.zip
obspy.io.chr               obspy.io.gse2              obspy.io.mseed             obspy.io.pde               obspy.io.seisan            obspy.io.zmap
obspy.io.cmtsolution       obspy.io.hex               obspy.io.ndk               obspy.io.pow               obspy.io.sh
obspy.io.cnv               obspy.io.input             obspy.io.next              obspy.io.print_function    obspy.io.shapefile
obspy.io.css               obspy.io.int               obspy.io.nied              obspy.io.quakeml           obspy.io.str
obspy.io.datamark          obspy.io.json              obspy.io.nlloc             obspy.io.range             obspy.io.super

In [2]: obspy.io.stationxml
AttributeError                            Traceback (most recent call last)
<ipython-input-2-0da88f925405> in <module>()
----> 1 obspy.io.stationxml

AttributeError: 'module' object has no attribute 'stationxml'

In [3]: import obspy.io.stationxml

In [4]:
```

**Comment by krischer** [comment]
That just how Python works. The test will work if you just place `import obspy.io.stationxml` somewhere on top.

If you just import ObsPy if knows about `obspy.io` (because some startup things implicitly import that). You might notice that it does also now `obspy.io.mseed`, e.g.

```python
>>> import obspy
>>> obspy.io.mseed
```

works fine (again because its imported at some startup procedure). `obspy.io.stationxml` is not yet known - you'll have to import it once - then it is known, e.g.

```python
>>> import obspy
>>> obspy.read_inventory()
>>> obspy.io.stationxml
```

will work, because the `read_inventory()` call imports it somewhere. The gists it that you can only attribute access a module or submodule if it has been imported before. Otherwise Python would have to try to import something each time an attribute access fails. That's the price of a fully dynamic language ;-)

dsentinel:
Gotcha, thanks for the explanation... I have `import obspy.io.stationxml.core` at the top of test_stationxml.py for now, I can take it out if best when I push if you don't want it there.

dsentinel:
Not sure how to tackle testing:
`test_read_and_write_full_file`  does a string comparison,  of a read and written XML. We could change all the test XML's to include microseconds? or logic to not write out micro's if not read in / exist?

krischer:
I tend to favor consistency over everything else so I would change all the test XML's to include microseconds but that is just a stylistic opinion. But I'm also fine with just writing the microseconds if they are there.

dsentinel:
> I tend to favor consistency over everything else so I would change all the test XML's to include microseconds but that is just a stylistic opinion. But I'm also fine with just writing the microseconds if they are there.
Blasted... I just did the conditional write. I'm more for consistincey as well, was hesitant to mess with others test data

dsentinel:
I think this is ready to merge.

obspy/io/stationxml/tests/test_stationxml.py
```diff
@@ -114,6 +115,41 @@ def test_read_and_write_minimal_file(self):
...
+                                "minimal_station_with_microseconds.xml")
+        inv = obspy.read_inventory(filename)
+
+        # Write it again. Also validate it to get more confidence. Suppress the
+        # writing of the ObsPy related tags to ease testing.
+        file_buffer = io.BytesIO()
+
+        # XXX helper variable to debug writing the full random file, set True
+        # XXX for debug output
+        write_debug_output = False
```
krischer: This is IMHO not needed. It is quite unlikely that this test will fail in the future and as such this adds a bit of noise to the test case. Then you can also always validate and remove the debug saving which will also increase test coverage ;-) Its easy enough to add debug statements whenever one has to debug something.

dsentinel: Removed it, and also from the existing test_read_and_write_full_file()

PR #1515, opened 2016-08-31
User-Agent: separate product from version with a slash instead of space

chad-iris:
This small change to the default User-Agent HTTP header for clients separates the product from the version using a slash instead of a space.  In short, User-Agent headers starting with "ObsPy/1.0.2" instead of "ObsPy 1.0.2".

This follows the pattern in RFC 7231, specifically section 5.5.3:
https://tools.ietf.org/html/rfc7231#section-5.5.3

### DIFF
--- a/obspy/clients/base.py
+++ b/obspy/clients/base.py
@@ -69,10 +69,10 @@ else:
     encoding = sys.getdefaultencoding() or "UTF-8"
     platform_ = platform.platform().encode(encoding).decode("ascii", "ignore")
 # The default User Agent that will be sent with every request.
-DEFAULT_USER_AGENT = "ObsPy %s (%s, Python %s)" % (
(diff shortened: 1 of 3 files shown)


krischer:
Looks good. We apparently don't directly test the user-agent string so no tests need to be adjusted. I manually cherry-picked this commit on top of the `maintenance_1.0.x` branch so it will flow into the next bugfix release.

Thanks for this!

QuLogic:
> I manually cherry-picked

Have you heard? GitHub allows you to change the merge base of PRs now.

krischer:
> Have you heard? GitHub allows you to change the merge base of PRs now.

MIND == BLOWN! I missed that. Finally :-) Only took them a couple of years and a gazillion duplicate PRs ;-)

QuLogic:
I mean, thinking about it, it wouldn't have really worked from `master` to `maintenance_1.0.x` without a rebase, but it's still nice to have that option now.

PR #1516, opened 2016-09-02
Inventory Query (get_all) function

MMesch:
### What does this PR do?
This proposal adds a short and simple query function to the inventory object. It corresponds to the discussion in #1236 (as well for a catalog). Usage is like this:
```
In [1]: import obspy
In [2]: inv = obspy.read_inventory()
In [3]: inv.get_all(('net.code', 'code'), level='station')
Out[3]: [['GR', 'GR', 'BW', 'BW', 'BW'], ['FUR', 'WET', 'RJOB', 'RJOB', 'RJOB']]
```

Open for discussion.

 >   long_list = catalog.get_all(['latitude', 'longitude', 'depth', 'focal_mechanism'])

>this would replace the current
>```
>for network in inventory:
>   for station in network:
>       lats.append(station.latitude)
>         [...]
>```
>loops in the plotting functions and is also useful for other functions that need bulk inventory/catalog
>data. Tell me if you have a better idea...
>
>Krischer (Feb 18):
> We could also do a slightly more generic variant:
> 
> ```python
> catalog.get_all(('preferred_origin.latitude', 'preferred_origin.longitude',
>                  'preferred_focal_mechanism.mt.tensor', ...))
> ```
> 
> Which would return lists of dictionaries with the requested keys. I'm not sure if we need the `preferred_` prefix or not but I think it might be useful. For inventories this is a bit more complex as they are nested. We could also think to add a more generic query interface that would replace the current `filter()` methods. I did something along these lines here which I think has quite a nice interface: http://seismicdata.github.io/pyasdf/querying_data.html
> 
> I also want to add a couple of more things. I'll do it after the release:
> 
> * The WebGL renderer for the raypaths.
> * Plots like in the middle of this page: http://www.isc.ac.uk/standards/phases/
> 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -666,6 +666,52 @@ class Inventory(ComparingObject):
 
         return fig
 
+    def get_all(self, attributes, level='channel', default=None):
+        """
+        Return a list of certain attributes from the inventory.
+
+        :param attributes: a list or tuple of attribute names. Example:
+                           ('net.code', 'code', 'longitude', 'latitude')
+                           if no rootname like net, sta, cha is given,
+                           the attribute is extracted from parameter `level`
+        :param level: the level at which the attributes are extracted.
+        :param default: default value if the attribute is doesn't exist
+        """
+        query = [[] for attrib in attributes]
+        for iattrib, attrib in enumerate(attributes):
+            # split attributes, set the variable name from which
+            # the attribute is extracted (root) and remove (root) from
+            # the attrib_name if necessary
+            attrib_parts = attrib.split('.')
+            if len(attrib_parts) > 1:
+                rootname = attrib_parts[0]
+                attrib_name = '.'.join(attrib_parts[1:])
+            else:
+                rootname = level[:3]
+                attrib_name = attrib
+
+            # loop through inventory according to the depth specified by level
+            if level == 'channel':
+                for net in self.networks:
+                    for sta in net.stations:
+                        for cha in sta.channels:
+                            root = locals()[rootname]
+                            value = getattr(root, attrib_name, default)
+                            query[iattrib].append(value)
+            if level == 'station':
+                for net in self.networks:
+                    for sta in net.stations:
+                        root = locals()[rootname]
+                        value = getattr(root, attrib_name, default)
+                        query[iattrib].append(value)
+            if level == 'network':
+                for net in self.networks:
+                    root = locals()[rootname]
+                    value = getattr(root, attrib, default)
+                    query[iattrib].append(value)
+
+        return query
+
     def plot_response(self, min_freq, output="VEL", network="*", station="*",
                       location="*", channel="*", time=None, starttime=None,
                       endtime=None, axes=None, unwrap_phase=False, show=True,
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -676,6 +676,17 @@ class Inventory(ComparingObject):
                            the attribute is extracted from parameter `level`
         :param level: the level at which the attributes are extracted.
         :param default: default value if the attribute is doesn't exist
+        :returns: list of lists. Each sublist contains the attribute values for
+                  each network/station/channel.
+
+        .. rubric:: Basic Usage
+        >>> import obspy
+        >>> inv = obspy.read_inventory()
+        >>> query = inv.get_all(('net.code', 'code'), level='station')
+        >>> query[0]
+        [u'GR', u'GR', u'BW', u'BW', u'BW']
+        >>> query[1]
+        [u'FUR', u'WET', u'RJOB', u'RJOB', u'RJOB']
         """
         query = [[] for attrib in attributes]
         for iattrib, attrib in enumerate(attributes):


MMesch:
needs to be extended to support subattributes and indexing in the search terms

krischer:
Just a quick refresher for myself: What is the intended use case for this? I see it for the `Catalog` object, but not for the `Inventory` object. We already have `.get_coordinates()` and `.get_contents()`.

MMesch:
ok yes it should have been for Catalog :)

MMesch:
my personal use case was getting a list of all station positions:
```
query = inv.get_all(('net.code', 'code', 'longitude', 'latitude'), level='station')
```

another one was extracting all channel azimuth and dips.

krischer:
Yea - why not implement it as part of the `get_contents()` function, i.e.

```python
>>> inv.get_contents()
{'channels': ...
 'networks': ...
 'stations': ...
 # The next two are new.
 'station_coordinates': {"AA.BB": {lat/lng/ele}, ...},
 'channel_coordinates': {"AA.BB.CC.DD": {lat/lng/ele/dep/azi/dip}}
}
```

That should cover all use cases?

MMesch:
Yes. But don't you think it would be better to just write a function that extracts whatever attribute ? This way the inventory can be extended later ...

krischer:
Yea maybe. I don't really have a good feel of the ups and downs of either approach yet so if you feel like something might work, just go for it :-)

I just don't really know what one would like to routinely extract from an inventory objects besides what I proposed above. If one requires something special one can always write a simple Python oneliner and get whatever one wants:

```python
>>> {("%s.%s" % (net.code, sta.code)): [sta.latitude, sta.longitude]
...  for net in inv for sta in net}
{'BW.RJOB': [47.737167, 12.795714],
 'GR.FUR': [48.162899, 11.2752],
 'GR.WET': [49.144001, 12.8782]}
````

One things that has to be taken care of (in your proposed `get_all()` as well as my proposed `get_contents()` method) is that essentially all information is time-dependent and thus a `datetime` parameter or some other means to transport that information is required.

**Comment by megies** [comment]
> Yea maybe. I don't really have a good feel of the ups and downs of either approach yet so if you feel like something might work, just go for it :-)

I'm also not sure about this, it seems a bit too magical for me.. having `level="station"` and `net.code`.. the `net` in there seems to be implicitly defined, no? Maybe it's less magic to just point users on using such a list comprehension as shown by @krischer, maybe in the docstrings of `Inventory`/`Network` and/or some Tutorial example..?

**Comment by MMesch** [comment]
yes I see your points. However, we are running very frequently into a situation that we have to loop through the deeply nested inventory or catalogs ourselfs. A query function is absolutely necessary in my opinion, I even wonder how you could/can survive without it until now :)

**Comment by megies** [comment]
Don't get me wrong, I agree that something like this would be useful. I'm just not 100% happy with the original proposal.. but I can't think of something better / more canonical right now.. :/

**Comment by MMesch** [comment]
yea I agree. Just trying to push this on the agenda to get unified, standardised and simple access to inventory-catalog contents at some point :)

**Comment by jkmacc-LANL** [comment]
Just as another data point, I agree with @MMesch's  sentiment that `Catalog` and `Inventory` objects are a little unwieldy.  They inherit quite a bit from QuakeML and StationXML, which are good _exchange_ formats but their nested and linked nature makes them relatively difficult to traverse and query, which is a common operation in _research_.  `Flat is better than nested.` Representing and working with nested or relational data isn't an easy problem, but I'd think that some top-level generic query/traversal functions or methods could be broadly useful, considering all the `for` loops, comprehensions, and `fnmatch` calls I see in the library (and my code).  I'd support at least exploring how difficult/easy it would be to implement a more generic query interface.

**Comment by MMesch** [comment]
> considering all the for loops, comprehensions, and fnmatch calls I see in the library (and my code).

That's also the reason why I realised that this would be a good addition.

**Comment by krischer** [comment]
Yea we definitely need something there! I think it should be possible to combine it with a query interface, e.g.:

```python
>>> inv\
    .filter(10.0 < inv.q.channel.sensor.sampling_rate < 100.0,
            inv.q.network.code == "B*")\
    .extract([inv.q.network.code, inv.q.station.code,
              inv.q.station.latitude], level="station")
[("BW", "ALTM", 41.2), ("BX", "ADF", 1), ...]
```
The `.filter()` method would just return a new inventory object with all channels/stations/networks satisfying the constraints and the `.extract()` would return the requested information as a tuple (it is very similar to the `get_all()` method proposed here but it can (amongst other things) tab complete). And we of course need something similar for the `Catalog` object.

What do you think? Or is that too complicated?

MMesch:
It looks fine for me. As use case, these parameters should be able to be easily extracted in a dict or list/tuple of lists in a single simple line. I guess that we can't handle attribute indices ?
*  net.code, stat.code , lat, lon (for date=xy)
* 'net.stat.loc.cha', lat, lon, azi, dip
* evlat, evlon, depth, preferred_mag, strike, dip
* evlat, evlon, mags, strike, dip (for date=xy)

krischer:
> I guess that we can't handle attribute indices ?

Yes we could also handle indices.

MMesch:
Let's go for it then! Do you have something running already?

krischer:
Sure thing :-) Let's mock up some examples to get a feel for what we are trying to achieve as its honestly not super simple. Please have a detailed look and make sure this can do whatever you want it to do so we can get it right.

I guess we can first focus on the `extract()` method as that is a bit more urgent.

### Inventory: net.code, stat.code , lat, lon (for date=xy)

```python
>>> inv.extract([inv.q.network.code, inv.q.station.code, inv.q.station.latitude,
                 inv.q.station.longitude], datetime=xy)
[("BW", "ALTM", 10, 11), (...)]
```

The `level` argument could be optional as the level can be inferred from the argument and just used the deepest that has been found.

It might be even nicer if it returned a list of named tuples, e.g. the first item in the above example:

```python
result(network_code="BW", station_code="BW", station_latitude=10, station_longitude=11)
```

These can still be accessed via indices but also have nice acessors. We can define these on the fly.

###  Inventory: 'net.stat.loc.cha', lat, lon, azi, dip

This has a shorthand for `network`, `station`, `channel`. This particular example would infer the `level` argument to be `"channel"`.

```python
>>> inv.extract([inv.q.net.code, inv.q.sta.code, inv.q.cha.location_code,
                 inv.q.cha.code, inv.q.cha.latitude, inv.q.cha.longitude,
                 inv.q.cha.azimuth, inv.q.cha.dip], datetime=xy)
[results(network_code="BW", station_code="ALTM", channel_location_code="",
         channel_code="EHZ", latitude=1, longitude=2, azimuth=3, dip=4),
 ...]
```

### Events

These are a bit more tricky...

##### Easy case: These would all be shortcuts to either use the preferred or first origin/magnitude/focmec/...

```python
>>> cat.extract(cat.q.latitude, cat.q.longitude, cat.q.depth, cat.q.magnitude, cat.q.dip)
...
```
##### Directly select various origins

We can just define a couple of keywords and otherwise use it to check the origin type for equality.

```python
>>> cat.extract(cat.q.origins["best"].latitude,  # preferred (if exists) else first
                cat.q.origins["preferred"].latitude,  # preferred, else raise
                cat.q.origins[1].latitude,  # the second one
                cat.q.origins["hypocenter"].latitude,  # the first with type==hypocenter
                cat.q.origins["centroid"].latitude)  # the first with type==centroid
```

```python
>>> cat.extract(cat.q.origin["centroid"].latitude, cat.q.origin["hypocenter"].latitude)
```

##### Relations?

Not sure if we want this but it might be useful. The following example gets the latitude/longitude for the best origin and the corresponding magnitude.

```python
>>> cat.extract(cat.q.origins["best"].latitude, cat.q.origins["best"].longitude,
                cat.q.magnitudes[cat.q.origins["best"]].magnitude)
```

jkmacc-LANL:
I like these examples.  Just as a wild thought, and to reiterate something that was said earlier, there is value in not inventing another specialized query language.  Is there a way to adopt an existing query interface for general nested data?  As nested data structures go, `Catalog` and `Inventory` aren't too bad (just bad enough to warrant this discussion :wink: ).  I don't have one in mind, but I've come across a number of tools like XPath/XQuery for traversing/querying various nested structures.  Using this example, the XPath query interface is used with XML in [lxml](http://lxml.de/xpathxslt.html#xpath), with JSON in [JSONPath](http://goessner.net/articles/JsonPath/), and with objects in [ObjectPath](http://objectpath.org/).  Traversing and querying simple hierarchies/paths, with support for numeric ranges, dates, and regular expressions, might be also used [elsewhere in ObsPy](https://github.com/obspy/obspy/blob/master/obspy/clients/filesystem/sds.py#L55) and [related projects](https://github.com/trichter/obspyh5/blob/master/obspyh5.py#L47)  There may be good reasons why this isn't possible, but I just wanted to throw it out there.

jkmacc-LANL:
Oh, re: the Pandas idea, a `.to_json` method could allow a user to get a `pandas.DataFrame`.

krischer:
I did something similar to the suggestions here so I know it can be done: http://seismicdata.github.io/pyasdf/querying_data.html

But I really like the idea of using a more generic existing thing so if such a thing exists we might be able to use it. I'm not sure we can use any of the proposed ones on our objects.

krischer:
We already have JSON output for the events: https://docs.obspy.org/packages/autogen/obspy.io.json.default.html

I guess we should add one to the inventory objects?

krischer:
Maybe we can also avoid having a new `.extract()` method by just adding some query examples to the documentation using Python syntax, e.g.

```python
>>> [(net.code, sta.code, cha.location_code, cha.code, cha.latitude, cha.azimuth)
     for net in inv for sta in net for cha in sta]
[(u'GR', u'FUR', '', u'HHZ', 48.162899, 0.0), ...]
```

This is honestly not more complicated than the stuff I proposed and its baked into the language so its also more flexible.

jkmacc-LANL:
@krischer Yes, I saw the pyasdf querying syntax, and I like it.  What was your thought process when deciding a query interface for seismic HDF5 files? This discussion seems equally relevant there.

:+1: on the `.to_json` for `Inventory` and `Stats`.  What was the use case for `Event` objects?

krischer:
For events its a bit more complicated but not much:

```python
>>> [(ev.preferred_origin().latitude,
      ev.preferred_origin().longitude,
      ev.preferred_magnitude().mag) for ev in cat]
[(41.818, 79.689, 4.4), (39.342, 41.044, 4.3), (38.017, 37.736, 3.0)]
```

The `.preferred_x()` is awkward but we could add `.best_org`/`.best_mag`/`.best_focmec` attributes that either get the preferred or first items.

krischer:
> @krischer Yes, I saw the pyasdf querying syntax, and I like it. What was your thought process when deciding a query interface for seismic HDF5 files? This discussion seems equally relevant there.

We needed some kind of query interface as the ASDF files can get very big. I actually had a prototype where I put all the necessary things in an in-memory SQLite database and then just use SQL to query it and some magic to map it back to stuff in the HDF5 file. That worked well enough but it was an enormous pain to keep it in sync with the ASDF file which can change at any point. The only way that worked was to generate the SQLite database anew each time one wants to start a query. This is quite expensive as it needs information from XML files within ASDF...The current `ifilter()` method is iterative and will yield each time it finds a matching waveform. And it very early excludes stuff it will never need making it reasonably fast. And it tab completes which is actually quite helpful ;-)

> 👍 on the .to_json for Inventory and Stats. What was the use case for Event objects?

MongoDB I think: https://github.com/obspy/obspy/pull/670

The inventory objects should probably serialize to this if someone wants to tackle this: https://github.com/chad-iris/StationJSON But I am not sure of the status of this effort.

MMesch:
The examples look very nice. Personally I think a function would make the code more readable, especially for people that don't know the inventory and catalog objects and that are no python Gurus either. But I agree that the list comprehension essentially does the same in almost the same space. If it is well documented I am perfectly fine with it as well... Difficult choice.

Jollyfant:
> The inventory objects should probably serialize to this if someone wants to tackle this: https://github.com/chad-iris/StationJSON But I am not sure of the status of this effort.

The schema appears to be mostly incomplete. It is implemented up to the channel level. I made a branch that writes inventory files to the StationJSON schema as is. I will work on the branch if the status of StationJSON changes. IMHO JSON is so much easier to work with than XML so I am heavily in favor of this format.

https://github.com/Jollyfant/obspy/tree/inv2json

jkmacc-LANL:
Should we ping Chad on this?

krischer:
> The examples look very nice. Personally I think a function would make the code more readable, especially for people that don't know the inventory and catalog objects and that are no python Gurus either. But I agree that the list comprehension essentially does the same in almost the same space. If it is well documented I am perfectly fine with it as well... Difficult choice.

I vote for just adding the examples to the documentation. I agree that double (and triple) list comprehensions are a bit awkward but we just avoid so much additional complexity if we don't write something custom.



> The schema appears to be mostly incomplete. It is implemented up to the channel level. I made a branch that writes inventory files to the StationJSON schema as is. I will work on the branch if the status of StationJSON changes. IMHO JSON is so much easier to work with than XML so I am heavily in favor of this format.
> 
> https://github.com/Jollyfant/obspy/tree/inv2json

Great :-) Can you open a new PR just so we can track the changes and make people aware?


> Should we ping Chad on this?

:+1:

megies:
> I vote for just adding the examples to the documentation. I agree that double (and triple) list comprehensions are a bit awkward but we just avoid so much additional complexity if we don't write something custom.

Like stated initially, I also tend to think that this would add lots of complexity and the code would probably not look much different than a nested list comprehension..

MMesch:
ok, sounds good. Does anyone feel like writing a line for the docs?

PR #1518, opened 2016-09-06
List operations on obspy.taup.tau.Arrivals should return an Arrivals object 

claudiodsf:
I started playing with `obspy.taup.tau.Arrivals` and realized that if I slice an `Arrivals` object I get a simple list, meaning, for instance, that I cannot plot it anymore:

```python
>>> from obspy.taup import TauPyModel
>>> model = TauPyModel(model='iasp91')
>>> arrivals = model.get_ray_paths(500, 140, phase_list=['PP', 'SSS'])
>>> print(type(arrivals[1:3]))
<type 'list'>
```
Same thing if I join two `Arrivals` objects (here the same, for sake of simplicity):
```python
>>> print(type(arrivals + arrivals))
<class 'future.types.newlist.newlist'>
```

It would be nice if those operations could still produce an `Arrivals` object.

### DIFF
--- a/obspy/taup/tau.py
+++ b/obspy/taup/tau.py
@@ -15,6 +15,7 @@ from matplotlib.cm import get_cmap
 import matplotlib.text
 import numpy as np
 
+from .helper_classes import Arrival
 from .tau_model import TauModel
 from .taup_create import TauPCreate
 from .taup_path import TauPPath
@@ -68,7 +69,7 @@ class _SmartPolarText(matplotlib.text.Text):
 
 class Arrivals(list):
     """
-    List of arrivals returned by :class:`TauPyModel` methods.
+    List like object of arrivals returned by :class:`TauPyModel` methods.
 
     :param arrivals: Initial arrivals to store.
     :type arrivals: :class:`list` of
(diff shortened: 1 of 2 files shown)


claudiodsf:
This is now a pull request which should fix the above behavior.
It also makes sure that nothing else than an `Arrival` object can be added to an `Arrivals` object.

Don't know if/where to add a test...

obspy/taup/tau.py
```diff
@@ -83,6 +84,55 @@ def __init__(self, arrivals, model):
...
+            raise TypeError(msg)
+
+    def __getitem__(self, index):
+        if isinstance(index, slice):
+            return self.__class__(super(Arrivals, self).__getitem__(index),
+                                  model=self.model)
+        else:
+            return super(Arrivals, self).__getitem__(index)
+
+    def __getslice__(self, i, j):
```
krischer: This has been deprecated since Python 2.0 and does not actually exists for Python 3. This should now be implemented as `slice` objects to the `__getitem__()` method. Here it is a bit awkward as built-in Python 2 objects such as `list` (which `Arrivals` currently inherits from) does have a `__getslice__` method so it has to be overwritten to get it working on Python 2. That's a bit of a mess.

Maybe its easier to just inherit from `collections.abc.MutableSequence` (`collections.MutableSequence` in Python 2) and add the required methods? Inheriting from `collections.UserList` (`UserList` module in Python 2) is also an option.

claudiodsf: The problem is that, in Python 2.7.12, if I remove the `__getslice__()` method, slicing returns a `list` object:
```python
In [3]: type(arrivals[1:])
Out[3]: list
```

(and I saw that we [do have](https://github.com/obspy/obspy/blob/master/obspy/core/stream.py#L665) a `__getslice__()` method for the `Stream` class.

krischer: Yes - if you inherit from `list` it is needed but if you for example inherit from on of the other proposed classes it is (I think) not needed and everything can be pushed to `__getitem__()`. The alternative is that you leave it as it is and somehow add handling of slices to `__getitem__()` which is needed for Python 3.

claudiodsf: But this is already done at line 126!

krischer: Ah I see - sorry about that. I guess travis is failing because of a missing `__imul__()` then? I don't understand why it works on Python 2 though...

claudiodsf: Yes, apparently `__imul__()` is not necessary for Python2 ...

krischer:
Looks mostly good. Can you add a couple more things to the test to have full test coverage?

* Append with wrong type.
* `__setslice__` is not tested
* `__setitem__` is not tested. The type checking is tested but not an actually successful operation.
* `__imul__` and `__mul__` with wrong type
* `__add__` with wrong type and a single arrival object

krischer:
Tests currently fail on Python 3.

claudiodsf:
Yeah, I think I know why. Need to find some time to fix it...

claudiodsf:
Squashed and force-pushed.

obspy/taup/tests/test_tau.py
```diff
@@ -952,6 +953,49 @@ def test_paths_for_crustal_phases(self):
...
+        with self.assertRaises(TypeError):
+            # test assignment with wrong type
+            arrivals[0] = 10.
+            arrivals[2:5] = [0, 1, 2]
+            arrivals.append(arrivals)
+            # test add and mul with wrong type
+            arrivals + [2, ]
+            arrivals += [2, ]
+            arrivals * [2, ]
+            arrivals *= [2, ]
```
claudiodsf: 👍

krischer:
Great :-) Looks good now.

Can you add a changelog item?

claudiodsf:
We really need it???

krischer:
Hmm...good question. Long changelogs are nice for marketing reasons so I'd add it but who knows. Your call ;-)

Jollyfant:
In that case also add "More additions the changelog".

claudiodsf:
I consider this a regression form an expected behavior. And maybe I'm the first person noticing it.
I think it's ok if it goes under the radar.

krischer:
I'll merge once CI passes!

claudiodsf:
By the way, a similar traitement should go to `Stream` objects and maybe other.
Today is possible to do something like:
```python
>>> from obspy import read
>>> st = read()
>>> st[1] = 2
>>> print(st)
AttributeError: 'int' object has no attribute 'id'
``` 
I can open an issue on that, but I'm not sure I have time to tackle all the list-like objects.

krischer:
Yea that should probably done as lots of our users are not necessarily technical people and this might just prevent some mistakes. Please open an issue :-)

claudiodsf:
👍

PR #1522, opened 2016-09-08
mseed-qc: first continuous segment start at window start

Jollyfant:
Changed the code so that the first continuous segment uses the window start (if specified) instead of the first sample as its start time when there is no start gap.

Added expected continuous segments start & end to `test_start_gap`.


krischer:
Thanks! Looks good and all new lines are tested. Merging.

PR #1523, opened 2016-09-09
Inv2json

Jollyfant:
Skeleton for `obspy.inventory` to JSON writing (and eventual reading). Writing is currently supported to channel depth through `inv.write(file, format="STATIONJSON")`.

See: https://github.com/chad-iris/StationJSON

Current status is **incompleted**; tests and schema validation are missing.


jkmacc-LANL:
ping  @chad-iris and reference issue #1516

megies:
Any news on this one @Jollyfant?

Jollyfant:
Nope. It's implemented up to the level that station JSON supports now but is not really sufficient to be useful since the format is missing response info.

megies:
Alright, so bumping milestone..

megies:
@Jollyfant is there any news on the response part? If not do we still want it without the response info?

megies:
Ok thanks for the clarification!

PR #1526, opened 2016-09-12
fix writing of long Q headers for python3

trichter:
Fixes the following issue

```
  File "/home/richter/dev/rf/rf/rfstream.py", line 183, in write
    super(RFStream, self).write(filename, format, **kwargs)
  File "/home/richter/dev/obspy/obspy/core/stream.py", line 1439, in write
    write_format(self, filename, **kwargs)
  File "/home/richter/dev/obspy/obspy/io/sh/core.py", line 573, in _write_q
    for j in range(0, minnol):
TypeError: 'float' object cannot be interpreted as an integer
```


krischer:
Can you add a quick test just to guard against regressions in the future?

trichter:
Yes, I can.

krischer:
Looks all good :-) Thanks!

PR #1509, opened 2016-08-23
Mseed zero dataoffset

krischer:
This PR fixes an issue noticed on the mailing list: http://lists.swapbytes.de/archives/obspy-users/2016-August/002115.html

The issue is actually not the blockette 201 but that the record with that blockette specifies a data offset of zero in the fixed header which later naturally fails when unpacking the data. Given that it also specifies that the record has zero samples this probably is an edgecase and valid MiniSEED.

With this PR, ObsPy will intersperse records with a length of zero indicating that something is fishy, e.g. the sample file in mailing list thread is read as:

```
13 Trace(s) in Stream:
CH.PANIX..LHZ | 2016-08-20T23:59:47.000000Z - 2016-08-21T01:45:30.000000Z | 1.0 Hz, 6344 samples
CH.PANIX..LHZ | 2016-08-21T01:43:37.000000Z - 2016-08-21T01:43:37.000000Z | 1.0 Hz, 0 samples
CH.PANIX..LHZ | 2016-08-21T01:45:31.000000Z - 2016-08-21T03:45:59.000000Z | 1.0 Hz, 7229 samples
CH.PANIX..LHZ | 2016-08-21T03:46:00.000000Z - 2016-08-21T05:48:41.000000Z | 1.0 Hz, 7362 samples
CH.PANIX..LHZ | 2016-08-21T03:46:10.000000Z - 2016-08-21T03:46:10.000000Z | 1.0 Hz, 0 samples
CH.PANIX..LHZ | 2016-08-21T05:47:15.000000Z - 2016-08-21T05:47:15.000000Z | 1.0 Hz, 0 samples
CH.PANIX..LHZ | 2016-08-21T05:48:42.000000Z - 2016-08-21T08:05:22.000000Z | 1.0 Hz, 8201 samples
CH.PANIX..LHZ | 2016-08-21T08:05:23.000000Z - 2016-08-21T13:00:28.000000Z | 1.0 Hz, 17706 samples
CH.PANIX..LHZ | 2016-08-21T08:06:41.000000Z - 2016-08-21T08:06:41.000000Z | 1.0 Hz, 0 samples
CH.PANIX..LHZ | 2016-08-21T13:00:29.000000Z - 2016-08-21T23:27:10.000000Z | 1.0 Hz, 37602 samples
CH.PANIX..LHZ | 2016-08-21T13:01:53.000000Z - 2016-08-21T13:01:53.000000Z | 1.0 Hz, 0 samples
[...]

PR #1525, opened 2016-09-12
read with headonly=True may shift trace starttime for zero data offset records

krischer:
So apparently the QC module found a gap in a fully continuous trace. Look closely at the start & endtime of the `11771` sample trace when reading in files with and without headonly.

    print(st.__str__(extended=True))

    NL.VKB..BHZ | 2015-05-18T13:21:15.069538Z - 2015-05-18T14:43:39.919538Z | 40.0 Hz, 197795
    NL.VKB..BHZ | 2015-05-18T14:43:39.919500Z - 2015-05-18T14:43:39.919500Z | 40.0 Hz, 0
    NL.VKB..BHZ | 2015-05-18T14:43:39.944538Z - 2015-05-18T14:48:34.194538Z | 40.0 Hz, 11771
    NL.VKB..BHZ | 2015-05-18T14:48:35.119500Z - 2015-05-18T14:48:35.119500Z | 40.0 Hz, 0

Setting read headonly=True (used for gap calculation in QC).

    print(st.__str__(extended=True))

    NL.VKB..BHZ | 2015-05-18T13:21:15.069538Z - 2015-05-18T14:43:39.919538Z | 40.0 Hz, 197795
    NL.VKB..BHZ | 2015-05-18T14:43:39.919500Z - 2015-05-18T14:48:34.169500Z | 40.0 Hz, 11771
    NL.VKB..BHZ | 2015-05-18T14:48:35.119500Z - 2015-05-18T14:48:35.119500Z | 40.0 Hz, 0

The first 0 sample trace disappears and the 11771 sample trace assimilates its start time, meaning we are introducing a gap. All other traces in the file remain identical.

Technically related to Mseed zero dataoffset: https://github.com/obspy/obspy/pull/1509 (despite that before this pull we could not even read this file). Any first thoughts on what may cause this spooky behaviour?

File used for demonstration:
ftp://www.orfeus-eu.org/continuous/2015/138/VKB.BHZ.NL.2015.138


krischer:
This issue is now a PR fixing the problem. It happened because `headonly=True` means the data records are not actually unpacked in the C code and thus a 0 sample record looks no different then a normal record but with not data.

Jollyfant:
So it was considering the 0 sample record to be part of the same segment as the `11771` trace and thus shifting the starttime. But why does this problem then not occur for all traces following a 0 sample record?

I tested your commit on the demo file, and it seems to work as expected. Thanks for the quick fix!!

krischer:
> So it was considering the 0 sample record to be part of the same segment as the 11771 trace and thus shifting the starttime. But why does this problem then not occur for all traces following a 0 sample record?

In your example the endtime of the one merged 0 sample record + one delta is the starttime of the next record within 0.5 delta (the criterion upon which we break up traces). All the others did not have timing that lined up that nicely and thus the traces have been split based on some other criterion.

PR #1527, opened 2016-09-14
Proxy support for NEIC CWB QueryServer client (and others?)

jkmacc-LANL:
**Problem:**
The NEIC CWB client times-out when requesting a waveform (code example at end).

**Possible solution:**
I believe this is because it can't use my proxy server.  I've successfully tested a manual connection and waveform retrieval using a proxy-enabled socket, following the code example at [this Gist](https://gist.github.com/frxstrem/4487802).  I think that adding something like  `proxy_user=None, proxy_server=None, proxy_password=None` keyword arguments into the `Client.__init__`, or using `os.environ['HTTP_PROXY']`, and routing the socket creation through the code in the above Gist, would enable use client through a proxy.  Other clients (arclink?) may also benefit from a similar change.

**Details:** 
OSX, ObsPy 1.0.2 maintenance, installed with `pip install .`

**Code example:**
```python
In [14]: from obspy.clients.neic import Client

In [15]: from obspy import UTCDateTime

In [16]: client.get_waveforms('IU', 'ANMO', '00', 'BH?', UTCDateTime()-(72*3600), UTCDateTime()-(72*3600)+120)
  File "/Users/jkmacc/repos/obspy/obspy/clients/neic/client.py", line 192, in get_waveforms_nscl
    s.connect((self.host, self.port))
  File "/Users/jkmacc/anaconda/envs/obspy/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 60] Operation timed out

CWB QueryServer at 137.227.224.97/2061
error                                     Traceback (most recent call last)
<ipython-input-16-b2e323e1ec90> in <module>()
----> 1 client.get_waveforms('IU', 'ANMO', '00', 'BH?', UTCDateTime()-(72*3600), UTCDateTime()-(72*3600)+120)

/Users/jkmacc/repos/obspy/obspy/clients/neic/client.pyc in get_waveforms(self, network, station, location, channel, starttime, endtime)
    130         seedname = seedname.replace("?", ".")
    131         return self.get_waveforms_nscl(seedname, starttime,
--> 132                                        endtime - starttime)
    133
    134     @deprecated("'getWaveformNSCL' has been renamed to 'get_waveforms_nscl'. "

/Users/jkmacc/repos/obspy/obspy/clients/neic/client.pyc in get_waveforms_nscl(self, seedname, starttime, duration)
    190                         print(ascdate(), asctime(), "connecting temp file",
    191                               tf.name)
--> 192                     s.connect((self.host, self.port))
    193                     s.setblocking(0)
    194                     s.send(line.encode('ascii', 'strict'))

/Users/jkmacc/anaconda/envs/obspy/lib/python2.7/socket.pyc in meth(name, self, *args)
    226
    227 def meth(name,self,*args):
--> 228     return getattr(self._sock,name)(*args)
    229
    230 for _m in _socketmethods:

error: [Errno 60] Operation timed out
```

### DIFF
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -10,6 +10,9 @@ License:
 J. MacCarthy, modified from https://gist.github.com/frxstrem/4487802
 
 """
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
 
 import socket
 from base64 import b64encode
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+"""
+Establish a socket connection through an HTTP proxy.
+Author: Fredrik Østrem <frx.apps@gmail.com>
+License:
+  This code can be used, modified and distributed freely, as long as it is this
+  note containing the original author, the source and this license, is put
+  along with the source code.
+
+J. MacCarthy, modified from https://gist.github.com/frxstrem/4487802
+
+"""
+
+import socket
+from base64 import b64encode
+import urllib
+import urlparse
+
+
+def get_proxy_tuple():
+    """
+    Return system http proxy as a urlparse.urlparse tuple or () if unset.
+    """
+    proxydict = urllib.getproxies()
+    proxystr = proxydict.get('http') or proxydict.get('https') or ''
+    if proxystr:
+        proxy = urlparse.urlparse(proxystr)
+    else:
+        proxy = ()
+
+    return proxy
+
+
+def valid_address(addr):
+    """ Verify that an IP/port tuple is valid """
+    TF = (isinstance(addr, (list, tuple)) and
+          len(addr) == 2 and
+          isinstance(addr[0], str) and
+          isinstance(addr[1], (int, long)))
+    return TF
+
+
+def http_proxy_connect(address, proxy, auth=None, headers={}):
+    """
+    Establish a socket connection through an HTTP proxy.
+
+    Arguments:
+      address (required)     = The address of the target
+      proxy (def: None)      = The address of the proxy server
+      auth (def: None)       = A tuple of the username and password used for
+                               authentication
+      headers (def: {})      = A set of headers that will be sent to the proxy
+
+    Returns:
+      A 3-tuple of the format:
+        (socket, status_code, headers)
+      Where `socket' is the socket object, `status_code` is the HTTP status
+      code that the server returned and `headers` is a dict of headers that the
+      server returned.
+
+    """
+    if not valid_address(address):
+        raise ValueError('Invalid target address')
+
+    if not valid_address(proxy):
+        raise ValueError('Invalid proxy address')
+
+    headers = {'host': address[0]}
+
+    if auth is not None:
+        if isinstance(auth, str):
+            headers['proxy-authorization'] = auth
+        elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
+            proxy_authorization = 'Basic ' + b64encode('%s:%s' % auth)
+            headers['proxy-authorization'] = proxy_authorization
+        else:
+            raise ValueError('Invalid authentication specification')
+
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(proxy)
+    fp = s.makefile('r+')
+
+    fp.write('CONNECT %s:%d HTTP/1.0\r\n' % address)
+    fp.write('\r\n'.join('%s: %s' % (k, v) for (k, v) in headers.items()) +
+             '\r\n\r\n')
+    fp.flush()
+
+    statusline = fp.readline().rstrip('\r\n')
+
+    if statusline.count(' ') < 2:
+        fp.close()
+        s.close()
+        raise IOError('Bad response')
+    version, status, statusmsg = statusline.split(' ', 2)
+    if version not in ('HTTP/1.0', 'HTTP/1.1'):
+        fp.close()
+        s.close()
+        raise IOError('Unsupported HTTP version')
+    try:
+        status = int(status)
+    except ValueError:
+        fp.close()
+        s.close()
+        raise IOError('Bad response')
+
+    response_headers = {}
+
+    while True:
+        l = fp.readline().rstrip('\r\n')
+        if l == '':
+            break
+        if ':' not in l:
+            continue
+        k, v = l.split(':', 1)
+        response_headers[k.strip().lower()] = v.strip()
+
+    fp.close()
+    return (s, status, response_headers)
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -52,8 +52,8 @@ def http_proxy_connect(address, proxy, auth=None):
 
     Arguments:
       address (required)     = The address of the target
-      proxy (def: None)      = The address of the proxy server
-      auth                   = A tuple of the username and password used for
+      proxy   (required)     = The address of the proxy server
+      auth  (def: None)      = A tuple of the username and password used for
                                authentication
 
     Returns:
@@ -77,14 +77,14 @@ def http_proxy_connect(address, proxy, auth=None):
             headers['proxy-authorization'] = auth
         elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
             auth_b64 = b64encode(bytes(('%s:%s' % auth).encode()))
-            proxy_authorization = 'Basic '.decode() + auth_b64
+            proxy_authorization = 'Basic '+ auth_b64.decode()
             headers['proxy-authorization'] = proxy_authorization
         else:
             raise ValueError('Invalid authentication specification')
 
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     s.connect(proxy)
-    fp = s.makefile('r')
+    fp = s.makefile('rw')
 
     fp.write('CONNECT %s:%d HTTP/1.0\r\n' % address)
     fp.write('\r\n'.join('%s: %s' % (k, v) for (k, v) in headers.items()) +
@@ -96,7 +96,7 @@ def http_proxy_connect(address, proxy, auth=None):
     if statusline.count(' ') < 2:
         fp.close()
         s.close()
-        raise IOError('Bad response. statusline:'.format(statusline))
+        raise IOError('Bad response. statusline: {}'.format(statusline))
     version, status, statusmsg = statusline.split(' ', 2)
     if version not in ('HTTP/1.0', 'HTTP/1.1'):
         fp.close()
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -77,7 +77,7 @@ def http_proxy_connect(address, proxy, auth=None):
             headers['proxy-authorization'] = auth
         elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
             auth_b64 = b64encode(bytes(('%s:%s' % auth).encode()))
-            proxy_authorization = 'Basic '+ auth_b64.decode()
+            proxy_authorization = 'Basic ' + auth_b64.decode()
             headers['proxy-authorization'] = proxy_authorization
         else:
             raise ValueError('Invalid authentication specification')
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -96,7 +96,7 @@ def http_proxy_connect(address, proxy, auth=None):
     if statusline.count(' ') < 2:
         fp.close()
         s.close()
-        raise IOError('Bad response')
+        raise IOError('Bad response. statusline:'.format(statusline))
     version, status, statusmsg = statusline.split(' ', 2)
     if version not in ('HTTP/1.0', 'HTTP/1.1'):
         fp.close()
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -19,18 +19,18 @@ from future.utils import native_str
 from base64 import b64encode
 import socket
 with standard_library.hooks():
-    import urllib
-    import urlparse
+    from urllib.request import getproxies
+    from urllib.parse import urlparse
 
 
 def get_proxy_tuple():
     """
-    Return system http proxy as a urlparse.urlparse tuple or () if unset.
+    Return system http proxy as a urlparse tuple or () if unset.
     """
-    proxydict = urllib.getproxies()
+    proxydict = getproxies()
     proxystr = proxydict.get('http') or proxydict.get('https') or ''
     if proxystr:
-        proxy = urlparse.urlparse(proxystr)
+        proxy = urlparse(proxystr)
     else:
         proxy = ()
 
@@ -46,16 +46,15 @@ def valid_address(addr):
     return is_valid
 
 
-def http_proxy_connect(address, proxy, auth=None, headers={}):
+def http_proxy_connect(address, proxy, auth=None):
     """
     Establish a socket connection through an HTTP proxy.
 
     Arguments:
       address (required)     = The address of the target
       proxy (def: None)      = The address of the proxy server
-      auth (def: None)       = A tuple of the username and password used for
+      auth                   = A tuple of the username and password used for
                                authentication
-      headers (def: {})      = A set of headers that will be sent to the proxy
 
     Returns:
       A 3-tuple of the format:
@@ -77,14 +76,15 @@ def http_proxy_connect(address, proxy, auth=None, headers={}):
         if isinstance(auth, str):
             headers['proxy-authorization'] = auth
         elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
-            proxy_authorization = 'Basic ' + b64encode('%s:%s' % auth)
+            auth_b64 = b64encode(bytes(('%s:%s' % auth).encode()))
+            proxy_authorization = 'Basic '.decode() + auth_b64
             headers['proxy-authorization'] = proxy_authorization
         else:
             raise ValueError('Invalid authentication specification')
 
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     s.connect(proxy)
-    fp = s.makefile('r+')
+    fp = s.makefile('r')
 
     fp.write('CONNECT %s:%d HTTP/1.0\r\n' % address)
     fp.write('\r\n'.join('%s: %s' % (k, v) for (k, v) in headers.items()) +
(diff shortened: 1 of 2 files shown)


jkmacc-LANL:
This issue is now a PR.  

The function `http_proxy_connect` at the previously-mentioned [Gist](https://gist.github.com/frxstrem/4487802) was copied into `obspy/clients/httpproxy.py`, and cleaned up.  It was also modified to require the `proxy` positional argument, and removing an internal check for it.  This was done to simplify the logic in `Client.get_waveforms_nscl`, so that checks like `if proxy: ...` don't appear both inside the client and inside the function.  Some checking is still required in the client, though, as the function returns a _connected_ socket, and `Client.get_waveforms_nscl` expects to make the connection itself.

There is no testing for `http_proxy_connect` or `Client.get_waveforms_nscl` from behind a proxy, mostly because I have no idea how to mock a proxy or set one up in the test suite.

I'd like to implement proxy support for the arclink client, but it's not clear to me how to do it.  I'm not familiar with `telnetlib`, or how it relates to proxy-enabled sockets.  Suggestions are welcomed :wink:

jkmacc-LANL:
I can confirm that simple `client.get_waveforms` and `client.get_waveforms_nscl` calls return data and no longer time-out.  The last two regex patterns in the final test are failing, though.  Could that be a server-side change in data availability?  I never touched the pattern-handling logic.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,121 @@
...
+
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
+
+import socket
+from base64 import b64encode
+import urllib
+import urlparse
```
krischer: Can you sort this alphabetically?

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,121 @@
...
+J. MacCarthy, modified from https://gist.github.com/frxstrem/4487802
+
+"""
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
+
+import socket
+from base64 import b64encode
+import urllib
```
krischer: You will have to import urllib with future to make sure it works across Python 2 and 3, see e.g. here: https://github.com/obspy/obspy/blob/ed0c7fb0074f657e807577e88466680c53f195ff/obspy/clients/fdsn/client.py#L29-L34

I think this is not a nice way and I hope we'll get rid of `future` eventually but for now it is needed.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,121 @@
...
+        proxy = ()
+
+    return proxy
+
+
+def valid_address(addr):
+    """ Verify that an IP/port tuple is valid """
+    is_valid = (isinstance(addr, (list, tuple)) and
+                len(addr) == 2 and
+                isinstance(addr[0].decode(), (str, unicode)) and
```
krischer: `unicode` does not exist on Python 3, on Python 2, the `future` packages takes care to alias `str` to `unicode`. You could change this to `(str, native_str)`, where `native_str` comes from the `future` packages.

krischer:
> The last two regex patterns in the final test are failing, though. Could that be a server-side change in data availability? I never touched the pattern-handling logic.

Same for me on the master branch. I opened a new issue here: https://github.com/obspy/obspy/issues/1529

> I'd like to implement proxy support for the arclink client, but it's not clear to me how to do it. I'm not familiar with telnetlib, or how it relates to proxy-enabled sockets. Suggestions are welcomed 😉

@barsch: Thoughts?

barsch:
no idea ... never worked with proxy & telnetlib - I know it is possible with twisted ...

obspy/clients/httpproxy.py
jkmacc-LANL: :+1: on the removing `future`.  Its magic has obscured my understanding of PY2/3 issues, I think.

obspy/clients/httpproxy.py
jkmacc-LANL: Done.

obspy/clients/httpproxy.py
jkmacc-LANL: This worked.  Thanks!

jkmacc-LANL:
Hmmm, no more hiding comments related to an "outdated diff"?  Can I mark comments as "resolved", can only the comment author do that?

krischer:
Hmm...yea I don't know...its a bit strange right now. I marked the changes as "approved" but they still show. Also this are only a 3 line change in the last commit but for anything more complex this will be a bit hard to do. Either I'm missing something or maybe they are still working on the UI?

obspy/clients/httpproxy.py
QuLogic: Why would you need `native_str`? You just decoded it, and `future` makes `str` into `unicode` where you need it.

QuLogic: Also, the `.decode` means that `addr[0]` was a `bytes`; why would that be so?

krischer: The `.decode()` does no longer exist in the latest version. Github now also shows outdated code snippets...

QuLogic: Bah, that "Changes from 2 commits" is not that obvious now.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from future.builtins import *  # NOQA
+from future import standard_library
+from future.utils import native_str
+
+from base64 import b64encode
+import socket
+with standard_library.hooks():
+    import urllib
```
QuLogic: These imports are backwards; it's using the Python 2 name, not the Python 3 name.

obspy/clients/httpproxy.py
krischer: Yea :-( I was quite excited yesterday but using the new review thingy a couple times now left me less happy...I guess they still need to iron out the rough edges. Also the review interface is extremely sluggish in my browser.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    import urlparse
+
+
+def get_proxy_tuple():
+    """
+    Return system http proxy as a urlparse.urlparse tuple or () if unset.
+    """
+    proxydict = urllib.getproxies()
+    proxystr = proxydict.get('http') or proxydict.get('https') or ''
```
QuLogic: Based on my reading of `getproxies` docs, I'm not 100% this is correct. The dictionary is scheme->proxy, where the scheme is determined by the URL to which you want to connect. Of course, we're actually connecting to something that's neither, so I'm not sure which is better.

I recall something saying it may be the case that HTTPS proxies are better because they're less likely to mangle any data (because they're transporting encrypted data, they can't change it however they like), but I'm not sure where I read that.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+      server returned.
+
+    """
+    if not valid_address(address):
+        raise ValueError('Invalid target address')
+
+    if not valid_address(proxy):
+        raise ValueError('Invalid proxy address')
+
+    headers = {'host': address[0]}
```
QuLogic: What's the point of the `headers` argument if it's just going to be replaced here?

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+def valid_address(addr):
+    """ Verify that an IP/port tuple is valid """
+    is_valid = (isinstance(addr, (list, tuple)) and
+                len(addr) == 2 and
+                isinstance(addr[0], (str, native_str)) and
+                isinstance(addr[1], int))
+    return is_valid
+
+
+def http_proxy_connect(address, proxy, auth=None, headers={}):
```
QuLogic: Mutable types should not be used as default parameters.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Establish a socket connection through an HTTP proxy.
+
+    Arguments:
+      address (required)     = The address of the target
+      proxy (def: None)      = The address of the proxy server
```
QuLogic: It's not default `None`.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    if not valid_address(proxy):
+        raise ValueError('Invalid proxy address')
+
+    headers = {'host': address[0]}
+
+    if auth is not None:
+        if isinstance(auth, str):
+            headers['proxy-authorization'] = auth
+        elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
+            proxy_authorization = 'Basic ' + b64encode('%s:%s' % auth)
```
QuLogic: `b64encode` accepts and returns `bytes`, so you won't be able to add these or call it.

obspy/clients/neic/client.py
```diff
@@ -170,15 +172,30 @@ def get_waveforms_nscl(self, seedname, starttime, duration):
...
+                else:
+                    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
                 with NamedTemporaryFile() as tf:
                     if self.debug:
                         print(ascdate(), asctime(), "connecting temp file",
                               tf.name)
-                    s.connect((self.host, self.port))
+                    if not self.proxy:
+                        # The proxy-ed socket is already connected
```
QuLogic: proxied

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+            headers['proxy-authorization'] = proxy_authorization
+        else:
+            raise ValueError('Invalid authentication specification')
+
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.connect(proxy)
+    fp = s.makefile('r+')
```
QuLogic: `+` is not listed in the supported modes.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+
+    if statusline.count(' ') < 2:
+        fp.close()
+        s.close()
+        raise IOError('Bad response')
+    version, status, statusmsg = statusline.split(' ', 2)
+    if version not in ('HTTP/1.0', 'HTTP/1.1'):
+        fp.close()
+        s.close()
+        raise IOError('Unsupported HTTP version')
```
QuLogic: Add `version` to the exception message.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    fp.write('\r\n'.join('%s: %s' % (k, v) for (k, v) in headers.items()) +
+             '\r\n\r\n')
+    fp.flush()
+
+    statusline = fp.readline().rstrip('\r\n')
+
+    if statusline.count(' ') < 2:
+        fp.close()
+        s.close()
+        raise IOError('Bad response')
```
QuLogic: Add `statusline` to the exception message.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    if version not in ('HTTP/1.0', 'HTTP/1.1'):
+        fp.close()
+        s.close()
+        raise IOError('Unsupported HTTP version')
+    try:
+        status = int(status)
+    except ValueError:
+        fp.close()
+        s.close()
+        raise IOError('Bad response')
```
QuLogic: Add `status` to the exception message.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+
+    response_headers = {}
+
+    while True:
+        l = fp.readline().rstrip('\r\n')
+        if l == '':
+            break
+        if ':' not in l:
+            continue
```
QuLogic: Shouldn't this `break`, or even raise an error?

obspy/clients/neic/client.py
```diff
@@ -170,15 +172,30 @@ def get_waveforms_nscl(self, seedname, starttime, duration):
...
+                        s.connect((self.host, self.port))
```
QuLogic: Does this really need to be in the `NamedTemporaryFile` context? Can it not just be in the `else` above?

obspy/clients/httpproxy.py
jkmacc-LANL: What do you mean?  I just copied something from elsewhere in the codebase.

obspy/clients/httpproxy.py
jkmacc-LANL: Do you want me to try the `https` first? `proxystr = proxydict.get('https') or proxydict.get('http') or ''`

obspy/clients/httpproxy.py
QuLogic: We should strive for Python 3 by default now; if should be `import urllib.request` and `import urllib.parse`. I don't see any other usage of the Python 2 imports in the code base.

obspy/clients/httpproxy.py
jkmacc-LANL: `None`-ed :wink:

obspy/clients/httpproxy.py
jkmacc-LANL: Made it required positional.

obspy/clients/httpproxy.py
jkmacc-LANL: Removed.

obspy/clients/httpproxy.py
jkmacc-LANL: Removed it (see below).

obspy/clients/httpproxy.py
jkmacc-LANL: Removed.

obspy/clients/httpproxy.py
jkmacc-LANL: Done.

obspy/clients/httpproxy.py
jkmacc-LANL: Looks like it just wants to try the next line, not to exit the loop.  Again, I'm not familiar with sockets.

obspy/clients/neic/client.py
jkmacc-LANL: Fixed.

obspy/clients/httpproxy.py
jkmacc-LANL: Ah, gotcha.  Done.

obspy/clients/httpproxy.py
jkmacc-LANL: "Fixed" via ugly hack.

obspy/clients/httpproxy.py
jkmacc-LANL: Done.

obspy/clients/httpproxy.py
jkmacc-LANL: Done.

jkmacc-LANL:
Where'd Travis go?

QuLogic:
What do you mean? Seems to have built just fine; it's the docs buildbot that seems to be stuck.

obspy/clients/httpproxy.py
QuLogic: I think you might have changed the wrong one; `proxy` is required, `auth` is optional.

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+        raise ValueError('Invalid proxy address')
+
+    headers = {'host': address[0]}
+
+    if auth is not None:
+        if isinstance(auth, str):
+            headers['proxy-authorization'] = auth
+        elif auth and isinstance(auth, (tuple, list)) and len(auth) == 2:
+            auth_b64 = b64encode(bytes(('%s:%s' % auth).encode()))
+            proxy_authorization = 'Basic '.decode() + auth_b64
```
QuLogic: Don't you want to `auth_b64.decode()`, not the other way (which wouldn't work)?

obspy/clients/httpproxy.py
QuLogic: Does it still work with just `r`? You still write to it, so something is weird (which might be the Python docs.)

obspy/clients/httpproxy.py
```diff
@@ -0,0 +1,124 @@
...
+    fp.write('\r\n'.join('%s: %s' % (k, v) for (k, v) in headers.items()) +
+             '\r\n\r\n')
+    fp.flush()
+
+    statusline = fp.readline().rstrip('\r\n')
+
+    if statusline.count(' ') < 2:
+        fp.close()
+        s.close()
+        raise IOError('Bad response. statusline:'.format(statusline))
```
QuLogic: Missing the `{}` for format.

obspy/clients/httpproxy.py
jkmacc-LANL: Totally right.  Fixed.

obspy/clients/httpproxy.py
jkmacc-LANL: Right.  Added.

jkmacc-LANL:
@QuLogic The Travis check doesn't show up on my list.  Only appveyor, codecov, and a pending docket-testbot.

QuLogic:
Travis is at the end of the list; you've just got to scroll down. We've got too many status checks posting things...

jkmacc-LANL:
I just noticed that this was tagged with `1.0.3`, but I branched from `master` because it seemed like a feature instead of a fix.  Should I fix this?

krischer:
Ah you are right - I'll change the milestone. I guess we can slowly start thinking about 1.1.0 in any case.

obspy/clients/httpproxy.py
jkmacc-LANL: Fixed docstring.

obspy/clients/httpproxy.py
jkmacc-LANL: It tried it with `rw`, and it seems to work.

krischer:
@QuLogic Can you approve your requested changes?

QuLogic:
Any comments on my question in my "overall" review comment?

jkmacc-LANL:
> Is it really necessary to have both `get_proxy_tuple` and `http_proxy_connect`? Why not just use `http_proxy_connect` from all callers and skip using the proxy if one isn't configured?

 `http_proxy_connect` returns a connected socket, and if proxy checking is done inside, it would return different things depending on the `HTTPS_PROXY` environmental.  That didn'tt seem like a good idea.

jkmacc-LANL:
Thank you both for your reviews.

QuLogic:
I don't understand why this procedure can't be done directly in `http_proxy_connect`?
```patch
+                if self.proxy:
+                    s, _, _ = http_proxy_connect(address, proxy, auth)
+                    # This socket is already connected to the proxy
+                else:
+                    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+                    s.connect((self.host, self.port))
```

jkmacc-LANL:
@QuLogic Oh, gotcha.  I didn't want to hide a normal `s.connect(...)` in an external function called `http_proxy_connect`.  Also, `http_proxy_connect` is in a `while` loop, and has a fair number of checks in the function.  Routing all iterations of the loop through those checks seems like it would potentially add unnecessary `if/then/else` proxy logic to most likely a normal socket connection.

jkmacc-LANL:
I did not intend to make further changes except for squashing.

megies:
@jkmacc-LANL, so the requested changes by @QuLogic  are all processed..? Didn't use this new feature a lot yet but there's still this red X..

jkmacc-LANL:
I've made all changes I think need to be made, and addressed all comments.  That doesn't mean I've modified code for every comment.  I haven't used the new review interface yet; can't @QuLogic approve the review, or do I need to make a code change for every comment for that to happen?

PR #1532, opened 2016-09-22
Remove unnecessary and broken distutils entrypoint.

QuLogic:
I ran into some issue with pbr, and then found [this report](https://bugs.launchpad.net/pbr/+bug/1624980) pointing to some invalid stuff in ObsPy. The entrypoint is not needed and in fact would not work since the class it names does not exist. The `build_man` and `install_man` commands are provided by `cmdclass` later in `setup.py`.


QuLogic:
Yea, I know. :P I think it was some leftover of following some old instructions.

PR #1408, opened 2016-09-26
Add possibility to trigger specific module's tests in CI on Pull Requests via issue comments

megies:
For PRs that make changes on network modules, I think it would be good to have some possibility to trigger some specific network module's tests (alongside the default non-network modules tests) during CI. We could e.g. check the commit message for a magic regex, something like 'TEST:clients.fdsn'..


megies:
This should enable requesting specific modules' tests via comments.. e.g. using:

+TESTS:clients.fdsn,clients.seedlink

todo:
 - [ ] changelog
 - [ ] maybe some testing
 - [ ] docs
 - [ ] use also in docker-testbot

megies:
Only thing missing is also do the check for requested additional tests in appveyor, too. Not so good with the windows syntax, so leaving it out for now. In any case Travis will test any additionally requested modules (or `ALL` modules, if requested).

krischer:
Some of the failing tests should work if you rebase on the lastest master.

megies:
Rebased on current master (hope I got those conflicts right..)

megies:
Todo:
 - [ ] run network tests in at least one docker container per day on maintenance_1.0.x and master and submit commit status

megies:
> run network tests in at least one docker container per day on maintenance_1.0.x and master and submit commit status

Hmm.. thinking about it, the other option would be to set Travis to test all modules for *pushes* (as opposed to PRs) to master/maintenance. That wouldn't impede speed of feedback on PRs commit status and would move those tests more to the core of our CI setup.


btw. @krischer, there's a test fail in fdsn mass downloader on old scipy (https://travis-ci.org/obspy/obspy/jobs/164580925), not sure if it's worth to fix..

krischer:
> Hmm.. thinking about it, the other option would be to set Travis to test all modules for pushes (as opposed to PRs) to master/maintenance. That wouldn't impede speed of feedback on PRs commit status and would move those tests more to the core of our CI setup.

Sounds good. But I guess then we really need to make our networks tests more stable.

> btw. @krischer, there's a test fail in fdsn mass downloader on old scipy (https://travis-ci.org/obspy/obspy/jobs/164580925), not sure if it's worth to fix..

Probably not worth to fix but the tests should be skipped for old scipy versions.

megies:
> But I guess then we really need to make our networks tests more stable.

Doesn't matter much actually.. any commit status with whatever context that sets a status of "error" or "failed" will make our build show up as broken..

megies:
Good to merge, I think.

Short summary of what this now actually does:

 - on Travis, for PR builds: install and use `obspy/obspy_github_api` to check whether any specific modules should be tested for the PR (using e.g. `+TESTS:clients.fdsn,clients.iris`) and test them.. ;-)
 - CONTRIBUTING.md is moved back out of the `.github` folder into bright daylight and a note on `+DOCS` and `+TESTS:...` magic is added for documentation
 - tinkering with github API is moved from several places in `misc/...` to a pip-installable module `obspy/obspy_github_api`

krischer:
Looks good to me :-)

Issue #1536, opened 2016-09-28
seiscomp3 inventory xml


blazvicic:
Hi!

As I was told on the mailing list, I am posting this bug (?) here.

This is the code, that was supposedly read inventory xml, that I downloaded from http://eida.gfz-potsdam.de/webdc3/

The file is this one:
https://drive.google.com/open?id=0BzMcKedS7yadMkRzY0RoS2FDVFU

Code I used was this one:

```
from obspy import read_inventory

inv = read_inventory("Package_1475048499259-GFZ_73152721_inventory.xml", format = "sc3ml")
print inv
```

And the error I get: 

> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "<decorator-gen-83>", line 2, in read_inventory
>   File "/home/blaz/miniconda2/lib/python2.7/site-packages/obspy/core/util/decorator.py", line 294, in _map_example_filename
>     return func(*args, **kwargs)
>   File "/home/blaz/miniconda2/lib/python2.7/site-packages/obspy/core/inventory/inventory.py", line 72, in read_inventory
>     format=format)[0]
>   File "/home/blaz/miniconda2/lib/python2.7/site-packages/obspy/core/util/base.py", line 464, in _read_from_plugin
>     list_obj = read_format(filename, **kwargs)
>   File "/home/blaz/miniconda2/lib/python2.7/site-packages/obspy/io/seiscomp/sc3ml.py", line 158, in _read_sc3ml
>     for net_element in inv_element.findall(_ns("network")):
> AttributeError: 'NoneType' object has no attribute 'findall'

Thanks, 
Blaž

Jollyfant:
Hi, I will try to have a look today.

Jollyfant:
It seems that the file format that was supplied by WebDC3 is different from the one implemented under `format="sc3ml"` in ObsPy. It could be a different (older) schema version but I'm not entirely sure. I'll add an exception when the input file is not in the expected format.

The following query gives you the inventory file and response information for the entire SL network in the supported sc3ml format:

`http://www.orfeus-eu.org/fdsnws/station/1/query?net=SL&format=sc3ml&level=response`

The following call works (preferably cache it on your drive):

    from obspy import_inventory
    inv = read_inventory("http://www.orfeus-eu.org/fdsnws/station/1/query?net=SL&format=sc3ml&level=response")

I will have to forward the question because I do not know what format is returned by WebDC3. Pinging @andres-h for some more information on this.

[...]

PR #1537, opened 2016-09-28
Fix to stop conversion of None to 'None' in sc3ml read inventory

Jollyfant:
Raises ValueError when `read_inventory(format="sc3ml")` is issued for a file that is not the supported SC3ML version. Includes test coverage.

See #1536


megies:
Hmm.. we usually don't do that. This would prevent users from forcing a data format during reading that is not properly recognized (by `_is_...`) but still readable (by `_read_...`)..

Jollyfant:
If a file is readable through `_read_...` but not properly recognized by `_is_...` that is a problem that needs to be addressed in the latter function surely? 

If there are exceptions, we can add it in `_is_...`. However, the XML format must be according to some schema, and exceptions shouldn't exist in the first place.

megies:
> If a file is readable through _read_... but not properly recognized by _is_... that is a problem that needs to be addressed in the latter function surely?

Yes, absolutely. But while it is (hopefully) being addressed in `master` (i.e. in the dynamic developer regime), users with the latest (or other) release version installed (i.e. in the static user regime) can not read the file simply because it is prevented in the reading routine although it might work. Remember, the `_read_XXX` function are only ever called if..
 - the `_is_XXX` detects the file format during plugin traversal in autodetection and prompts the read routine, or..
 - the `_read_XXX` is *explicitly` requested by the user (be it by specifying `format="XXX"` or by tinkering with low-level routine itself).

In the second case this is an *explicit* call and as such should not be prevented by hard coding, IMHO.

Jollyfant:
Ok, I get what you are saying now..

Users would normally not specify `format="SC3ML"` or `format="stationXML"` during a read and let the auto-detection do its job. And if someone wants to be stubborn and import stationXML as SC3ML we let them.

megies:
> Users would normally not specify format="SC3ML" or format="stationXML" during a read and let the auto-detection do its job. And if someone wants to be stubborn and import stationXML as SC3ML we let them.

Exactly. And it's not about being stubborn in some cases but just the `_is` routine being buggy but the `_read` routine working properly at the same time. We had cases like that already..



> Btw, I do not understand what you mean with static vs dynamic developer regime.

Just saying, if you hardcode it like here, a user has no chance to force the reading without either tinkering in the installed codes or installing a developer snapshot -- which is usually unwanted and even impossible for some unsavvy users.

Jollyfant:
I agree with your philosophy and will revert the change, and use this pull to make a different small bug fix as demonstrated in the last commit.

Jollyfant:
Modified test for better coverage. If they pass this can be reviewed and merged.

krischer:
I manually cherry-picked 61c93a4d589a504dcadaf0f40242e4cfea03e065 and 68a75e4bef8716ca2cd70aa2bbfb99aa714d36f2 on top of the `maintenance_1.0.x` branch.

Issue #1492, opened 2016-08-01
Licensing incompatibility in vendored libmseed source code


ghisvail:
Whilst inspecting the copyright of the `obspy` files (a necessary evil of the Debian packaging process), I encountered the following inconsistency:

According to the `libmseed` README located at `obspy/io/mseed/src/libmseed`, the library is released under GPL-2+. However the source code of `libmseed` contains a file named `unpackdata.c` with the following copyright notice:

```c
/*
 * Copyright (c) 1996 The Regents of the University of California.
 * All Rights Reserved.
 *
 * Permission to use, copy, modify, and distribute this software and its
 * documentation for educational, research and non-profit purposes,
 * without fee, and without a written agreement is hereby granted,
 * provided that the above copyright notice, this paragraph and the
 * following three paragraphs appear in all copies.
 *
 * Permission to incorporate this software into commercial products may
 * be obtained from the Office of Technology Licensing, 2150 Shattuck
 * Avenue, Suite 510, Berkeley, CA  94704.
 *
 * IN NO EVENT SHALL THE UNIVERSITY OF CALIFORNIA BE LIABLE TO ANY PARTY
 * FOR DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES,
 * INCLUDING LOST PROFITS, ARISING OUT OF THE USE OF THIS SOFTWARE AND
 * ITS DOCUMENTATION, EVEN IF THE UNIVERSITY OF CALIFORNIA HAS BEEN
 * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 * THE UNIVERSITY OF CALIFORNIA SPECIFICALLY DISCLAIMS ANY WARRANTIES,
 * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
 * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.  THE SOFTWARE
 * PROVIDED HEREUNDER IS ON AN "AS IS" BASIS, AND THE UNIVERSITY OF
 * CALIFORNIA HAS NO OBLIGATIONS TO PROVIDE MAINTENANCE, SUPPORT,
 * UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 */
```

[...]

Issue #341, opened 2012-09-09
obspy.mseed samprate out of range


trac2github:
Dear Obspy support,

I'm trying to convert a stream/Trace into a mini-seed file. However I am running in to difficulties with:
"obspy.mseed samprate out of range". 
I checked online to find out what the maximum sampling rate frequency was for this function and it turns out that it is 32727 Hz.

My data files (Acosutic Emission laboratory data) have a sampling rate frequency of 1E07 Hz (10 MHz), so quite significantly larger. Is there anything I can do/tweak to be able to use this mseed write function in Obspy?

Thank you very much for your help in advance.

Regards,

Will.

PR #1540, opened 2016-09-30
Integration testing of libmseed v2.18rc2 requested

krischer:
A new version of libmseed is being prepared for release.  The primary change in this release is the rewrite of the encoding and decoding routines from scratch to replace the routines from qlib2, which should address #1492.

Such a fundamental change merits caution and as much testing as possible, any assistance with testing is greatly appreciated.  Release candidate is here (for now):
http://ds.iris.edu/files/staff/chad/libmseed-2.18rc2.tar.gz

While I was in contemplating licensing I updated libmseed to LGPL3  and added a LICENSE.txt file as suggested by Elliot.  Hopefully this is the end of license issues for libmseed.


krischer:
Hi Chad,

this is great and solves a big problem for us. Thanks a lot!

This issue is now a PR so all the test runners can have a go at it. Some thoughts:

* The API changed a bit. We did use `msr_unpack_steim1/2` but not its called `msr_decode_steim1/2`. I assume that is intentional although and its no problem for ObsPy but it will break some people's codes.
* `libmseed.def` has to be updated to include the new API (see https://github.com/obspy/obspy/pull/1540/commits/58563a9676da7c7109ee54f28e01419ed42410b5)
* There is currently a segfault on OSX reading this file: https://github.com/obspy/obspy/blob/master/obspy/io/mseed/tests/data/brokenlastrecord.mseed

lldb/gdb output:

```
* thread #1: tid = 0x570472, 0x000000010d04b23e libmseed_Darwin_64bit_py35.cpython-35m-darwin.so`msr_decode_steim2(input=0x0000000104535020, inputlength=<unavailable>, samplecount=<unavailable>, output=0x000000000000175c, outputlength=<unavailable>, srcname="", swapflag=<unavailable>) + 1950 at unpackdata.c:547, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x175c)
    frame #0: 0x000000010d04b23e libmseed_Darwin_64bit_py35.cpython-35m-darwin.so`msr_decode_steim2(input=0x0000000104535020, inputlength=<unavailable>, samplecount=<unavailable>, output=0x000000000000175c, outputlength=<unavailable>, srcname="", swapflag=<unavailable>) + 1950 at unpackdata.c:547
   544 	        {
   545
   546 	          if (outputptr == output) /* Ignore first difference, instead store X0 */
-> 547 	            *outputptr = X0;
   548 	          else /* Otherwise store difference from previous sample */
   549 	            *outputptr = *(outputptr - 1) + diff[idx];
   550
```

I'll have a more detailed look after the weekend.

krischer:
@obspy/developers `libmseed` now also comes with lots of tests including data and expected result. We should also use these:

* Include data and known good output into the test suite.
* Additionally we could also directly run the `libmseed` test suite as part of the `obspy.io.mseed` test suite.

chad-iris:
- _The API changed a bit. We did use msr_unpack_steim1/2 but not its called msr_decode_steim1/2. I assume that is intentional although and its no problem for ObsPy but it will break some people's codes._

The msr_unpack_* and msr_pack_* are considered to be internal functions, they are not exposed via the libmseed.h file.  I did not make any attempt to ensure the calling was the same.

- _libmseed.def has to be updated to include the new API (see 58563a9)_

I'll update these to the decode functions.  Strange place for these, probably added them on request from a downstream project ;), even stranger that it's only the steim[12] routines.

- _There is currently a segfault on OSX reading this file: https://github.com/obspy/obspy/blob/master/obspy/io/mseed/tests/data/brokenlastrecord.mseed_

Bummer, I cannot reproduce this fault on OSX 10.11.6.  The _msview_ program included with libmseed parses the first record and skips the non-record data just fine for me.  Can you send me a more detailed backtrace from lldb?  Details of how the library functions are being called?

Regarding the included test suite: GNU make is needed to run the tests, so they are not ultra portable.  At least one of the tests relies on a specific ordering of piped output that mixes stdout and stderr, not ideal, I'll try to fix that.  Some things to consider should you want to run it directly.

Thanks for the initial feedback!
I'd be happy to dig into the crash you found, any more information you can provide would be appreciated.

megies:
> There is currently a segfault on OSX

Segfault seems to also occur on Linux (e.g. see Travis), Appveyor log for windows also looks suspicious..

mbyt:
@chad-iris, thank you very much for the announcement and the detailed analysis!

> The msr_unpack_* and msr_pack_* are considered to be internal functions,
> they are not exposed via the libmseed.h file.
> I did not make any attempt to ensure the calling was the same.

Thanks for the feedback. I checked that and indeed, we only use these
functions in test code. I would like to delete them from obspy, what do
others think (ping @krischer)?

Nevertheless obspy's type signature was just wrong for the new `msr_decode_steim[12]`. I corrected that one in 6768d7d.

> Such a fundamental change merits caution and as much testing
> as possible, any assistance with testing is greatly appreciated.

We have one test writing traces with very large sampling_rate
(`sampling_rate = 1000000000.0`). This test also segfaults with libmseed2.18 and it is unrelated to the `msr_decode_steim[12]` changes. You can what the test does
here: https://github.com/obspy/obspy/blob/master/obspy/io/mseed/tests/test_mseed_special_issues.py#L525. I got this test passing, by the above patch of libmseed: 22c3341.

mbyt:
Just saw the output of https://travis-ci.org/obspy/obspy/builds/164450326. Tests now pass on windows and linux. However, there is still a segmentation fault on OSX.

chad-iris:
rc3 is here with some improvements to the test suite, a fix for the outputlength test crash (@mbyt) and doc tweaks (@QuLogic) :
http://ds.iris.edu/files/staff/chad/libmseed-2.18rc3.tar.gz

I need to look further into the outputlength test before the memset(), that shouldn't have been possible.

Regarding the huge sample rate test: libmseed currently sets the sample rate values in the fixed section of the data header to 0 when the rate is above 32,767 Hz.  If a Blockette 100 is included it should preserve those rates.  The support for such high rates could likely be improved, but I question whether that is reasonable.  Is there really a use case for 1 billion samples/second using miniSEED?

mbyt:
Hi, thanks for the prompt update!

> Is there really a use case for 1 billion samples/second using miniSEED?

Not sure, we have this test as people seem to use miniseed for encoding "Acosutic Emission laboratory data".  See #341 for details.

> when the rate is above 32,767 Hz.
> If a Blockette 100 is included it should preserve those rates.

Thanks for pointing it out. I browsed through the code and yes we are actually doing this. Our upper sampling border is a little bit lower @ 32727Hz  --- but that shouldn't be any harm.
```python
        if trace.stats.sampling_rate >= 32727.0 or \
                trace.stats.sampling_rate <= (1.0 / 32727.0):
            # ...
            ret_val = clibmseed.msr_addblockette(
                msr, C.pointer(blkt100), size, 100, 0)  # NOQA
```
For more details see https://github.com/obspy/obspy/blob/master/obspy/io/mseed/core.py#L867.

krischer:
All tests pass now on OSX. I added one more commit to get rid of some compilation warnings.

krischer:
> Is there really a use case for 1 billion samples/second using miniSEED?
> 
> Not sure, we have this test as people seem to use miniseed for encoding "Acosutic Emission laboratory data". See #341 for details.

I've actually once met people who used MiniSEED for that so continued support for that would be great.


> Thanks for the feedback. I checked that and indeed, we only use these
> functions in test code. I would like to delete them from obspy, what do
> others think (ping @megies, @krischer)?

Fine with me. But you also already did it ;-)

krischer:
Rebased on the latest master and force pushed.

krischer:
I also added a new test that takes all the reference and test data coming with the new libmseed and executes them with ObsPy.

The one thing that currently differs is the `read-invalid-blockette-offset.test.ref`. It reads the file `invalid-blockette-offset.mseed` and should have the following output:

```
XX_TEST_00_LHZ, 000001, M, 512, 112 samples, 1 Hz, 2010,058,06:50:00.069539
XX_TEST_00_LHZ, 000000, M, 512, 185 samples, 1 Hz, 2010,058,06:51:52.069541
Error: Invalid blockette offset (12365) less than or equal to current offset (12365)
```

Reading the file with ObsPy results in following which has to be checked again.

```python
In [2]: obspy.read("./src/libmseed/test/data/invalid-blockette-offset.mseed")
/Users/lion/workspace/code/obspy/obspy/io/mseed/core.py:413: InternalMSEEDReadingWarning: XX_TEST_00_LHZ_M: Warning: Data integrity check for Steim2 failed, Last sample=-236912, Xn=-236956
  warnings.warn(*_i)
/Users/lion/workspace/code/obspy/obspy/io/mseed/core.py:413: InternalMSEEDReadingWarning: XX_TEST_00_LHZ_M: Warning: Data integrity check for Steim2 failed, Last sample=-22070818, Xn=-201796
  warnings.warn(*_i)
InternalMSEEDReadingError                 Traceback (most recent call last)
<ipython-input-2-c67d2d34e6c9> in <module>()
----> 1 obspy.read("./src/libmseed/test/data/invalid-blockette-offset.mseed")

...

/Users/lion/workspace/code/obspy/obspy/io/mseed/core.py in _read_mseed(mseed_object, starttime, endtime, headonly, sourcename, reclen, details, header_byteorder, verbose, **kwargs)
    410     for _i in _errs_and_warnings:
    411         if isinstance(_i, InternalMSEEDReadingError):
--> 412             raise _i
    413         warnings.warn(*_i)
    414

InternalMSEEDReadingError: msr_unpack_data(XX_TEST_00_LHZ_M): only decoded 184 samples of 185 expected
```

krischer:
Rebased + force pushed to get rid of my debugging on travis...

krischer:
Rebased on top of latest master to include #1544 .

mbyt:
Let's wait until 2.18 is officially released and merge this afterwards?

megies:
There's some fails on most 32bit Linuxes, brought to you by our docker testbot.. ;-)

krischer:
Interesting. These issues (numbers off by 0.5, 1.0, 2.0, ...) suspiciously remind me of #485 which had the problem with byteswapped NaNs. Byteswapped floats can by chance be NaNs and due to various reasons (see the old issue) it can be switched to a different NaN representation which after byte swapping are no longer the correct numbers.

As far as I can tell from the test reports it happens during reading 32bit float (encoding 4) on 32 bit systems, e.g. this file (https://github.com/obspy/obspy/blob/c1a9c8c3940fef835897f4a3f4b53488b71d4361/obspy/io/mseed/tests/data/encoding/nan_float32.mseed) whose data should be `[-1188.07800293, 638.16400146, 395.07809448, 1060.28112793]`.

krischer:
The latest commit should fix the issue.

krischer:
> Cool. Let's wait until 2.18 is officially released and merge this afterwards?

Sounds good.

chad-iris:
If someone is willing, here is what I hope to be the last RC for 2.18:

http://ds.iris.edu/files/staff/chad/libmseed-2.18rc4.tar.gz

This includes the modifications I found above and new code for sample rate factor and multiplier generation.

If no bugs are found this version will be re-tagged as 2.18 final.

In addition to supporting all sample rates between 32767 and 1/32767, now integer rates and sample periods are supported from 1 to the maximum possible (32767 * 32767).  Integer sample rates and periods larger than 32767 will be approximated if they cannot be factored perfectly.  So the case of 1 biiiiilion samples/second works perfectly, but a rate of 1,000,000,091 doesn't factor so cleanly.  Folks should be using a Blockette 100 if they have such rates anyway.

Thanks again for the testing.

QuLogic:
The test Makefile does not correctly exit with a non-zero status when tests fail.

obspy/io/mseed/src/libmseed/unpackdata.c
```diff
@@ -119,7 +119,7 @@ msr_decode_float32 (float *input, int samplecount, float *output,
 
   for (idx = 0; idx < samplecount && outputlength > 0; idx++)
   {
-    sample = input[idx];
+    memcpy(&sample, &input[idx], sizeof(float));
```
mbyt: wow good call!

For me, it took a time to understand #485 again. And the nan reinterpretation is quite hidden here, as input in this function is `input` is `float *`, but gets a cast in pack.c from `int *`.

Good call!

QuLogic:
> The test Makefile does not correctly exit with a non-zero status when tests fail.

I've used [this patch](https://gist.github.com/QuLogic/b9de354f189ded6d65405ba241fc6c7a) which correctly exits early on failure. It's not perfect (output is highly jumbled in parallel), but at least failures won't get buried in the middle of a log somewhere.

chad-iris:
>> The test Makefile does not correctly exit with a non-zero status when tests fail.

>I've used this patch which correctly exits early on failure. It's not perfect (output is highly jumbled in parallel), but at least failures won't get buried in the middle of a log somewhere.

Thanks.  Problem is that I don't want to exit early on failure, I want to run all the tests.
I've already given up on enhancing the Makefile, it is just too limiting (e.g. I couldn't easily do simple things like count the failed/successful tests).  A future version will likely be in system shell.

QuLogic:
I expected as much; an external script (shell or Python or whatever) is probably the only real option there.

chad-iris:
@megies Should I wait for that docker-testbot to finish, any trouble identified by it?

If there are any other pressing issues with v2.18 from the ObsPy perspective, now's the time to bring them up.

megies:
> @megies Should I wait for that docker-testbot to finish, any trouble identified by it?

@chad-iris I've been doing some changes to the testbot setup the last few days, still need to polish the last few spots. I'll kick the docker tests off locally on my machine, should be done in an hour or so, will give you a ping.

megies:
@chad-iris [all docker tests passed](http://tests.obspy.org/?version=1.0.2.post0%2B491.g889dbe5f3c&node=docker-), the red "x" for "docker-deb-buildbot" can be ignored, those automated builds are not working as planned right now.

setup.py
```diff
@@ -615,6 +615,18 @@ def add_data_files(config):
...
+    FORCE_INCLUDE_DIRS = [
+        os.path.join(SETUP_DIRECTORY, 'obspy', 'io', 'mseed', 'src',
+                     'libmseed', 'test')]
+
+    for folder in FORCE_INCLUDE_DIRS:
+        for root, _, files in os.walk(folder):
+            for filename in files:
+                config.add_data_files(
+                    os.path.relpath(os.path.join(root, filename),
+                                    SETUP_DIRECTORY))
```
megies: @krischer I just noticed that these files are missing in deb packaging.. can we somehow avoid using test data that are not in the usual location? To include this additional directory will be some additional work for all @obspy/packagers..

krischer: I don't know honestly - do symlinks work in git across platforms?

krischer:
Stuff we need to do after 2.18 has been released:

- [ ] Change logic when to use blockette 100 (see https://github.com/obspy/obspy/pull/1550).
- [ ] Figure out what to do with libmseed's tests data during packaging
- [ ] Figure out this issue here: https://github.com/obspy/obspy/pull/1540#issuecomment-251094094

chad-iris:
2.18 is released:

Thank you for the testing.

krischer:
I updated to 2.18 final and fixed two of the issues mentioned above.

> Figure out what to do with libmseed's tests data during packaging

I think we should leave it as it currently is - it will add some burden to the packagers but the tests catch it and its easy to fix. The adds essentially free test coverage to one of our most crucial code paths. The latest test file already exposed a subtle bug.

One more thing I forgot previously is that the upcoming REFTEK reader would like to have access to the STEIM decoding routines which were removed in this PR: https://github.com/obspy/obspy/pull/1433

We could still just use it even if it is no longer part of the offical libmseed API. For windows we should be able to just put it into our own `obspy-readbuffer.def` file.

obspy/io/mseed/src/obspy-readbuffer.c
```diff
@@ -469,7 +469,7 @@ readMSEEDBuffer (char *mseed, int buflen, Selections *selections, flag
 
         // Actually unpack the data if the flag is not set and if the data
         // offset is valid.
-        if ((unpack_data != 0) && (msr->fsdh->data_offset > 48) &&
+        if ((unpack_data != 0) && (msr->fsdh->data_offset >= 48) &&
             (msr->fsdh->data_offset < msr->reclen) &&
```
mbyt: good catch!

Might be worth mentioning in the release notes.

obspy/io/mseed/core.py
```diff
@@ -865,20 +865,19 @@ def record_handler(record, reclen, _stream):
...
-        if trace.stats.sampling_rate >= 32727.0 or \
-                trace.stats.sampling_rate <= (1.0 / 32727.0):
+        use_blkt_100 = False
+
+        _factor = C.c_int16()
+        _multiplier = C.c_int16()
+        _retval = clibmseed.ms_genfactmult(
+            trace.stats.sampling_rate, C.pointer(_factor),
+            C.pointer(_multiplier))
+        # Use blockette 100 if ms_genfactmult() failed.
```
mbyt: that's good. It avoids the redefinition of the libmseed internal sampling rate limits!

megies:
Docker testbot fails are unrelated.

krischer:
I re-added our wrappers for the STEIM unpacking routines so #1433 is free to use them. Additionally I added them to our own `.def` file so it should work fine on Windows - lets see what CI says.

@QuLogic Any packaging problems to be expected with this? The only thing we now do is access two not officially exposed functions from `libmseed`.

If all tests pass this is IMHO ready to merge.

krischer:
Travis failure are unrelated but happen during the installation of the github helpers. @megies: Can you please have a look?

krischer:
I'm merging this to move on. We've discussed in a lot and it should be good. It will be a while until 1.1.0 in any case and we thus still have time to fix any unforeseen problems.

megies:
> Travis failure are unrelated but happen during the installation of the github helpers. @megies: Can you please have a look?

I'm proposing to install as many adiitional test dependencies as possible via `conda` from conda-forge channel. That's usually more fail-safe than pip from my experience.. see #1560.

QuLogic:
That symbol renaming was a bit of an annoyance because I can't just upgrade libmseed by itself now. Too late now, though...

PR #1542, opened 2016-10-02
Fix irisws-distaz client following service upgrade to 1.0.3

chad-iris:
The previous irisws-distaz client used an undocumented feature to request JSON results.  After a recent update to the service this feature is no longer supported.

This PR changes the code to parse the default XML.  Also, 'distancemeters' and 'ellipsoidname' are now returned in the result.  Tests have been updated.


QuLogic:
Looks like you started from `master` and opening to `maintenance_1.0.x`. If that's the intended target, it'll need a rebase to remove all the extra commits.

CTrabant:
Ah, crud.  Still learning, let me try to fix that.

QuLogic:
Wasn't necessary to delete the PR; just a force push would have worked.

PR #1543, opened 2016-10-03
Fix irisws-distaz client following service upgrade to 1.0.3

chad-iris:
The previous irisws-distaz client used an undocumented feature to request JSON results. After a recent update to the service this feature is no longer supported.

This PR changes the code to parse the default XML. Also, 'distancemeters' and 'ellipsoidname' are now returned in the result. Tests have been updated.


obspy/clients/iris/client.py
```diff
@@ -558,7 +558,7 @@ def sacpz(self, network, station, location="*", channel="*",
     def distaz(self, stalat, stalon, evtlat, evtlon):
         """
         Low-level interface for `distaz` Web service of IRIS
-        (http://service.iris.edu/irisws/distaz/) - release 1.0.1 (2010).
+        (http://service.iris.edu/irisws/distaz/) - release 1.0.3 (2016).
```
QuLogic: Since you're changing this line, we should use the https link.

obspy/clients/iris/client.py
```diff
@@ -583,34 +583,41 @@ def distaz(self, stalat, stalon, evtlat, evtlon):
         Latitudes are converted to geocentric latitudes using the WGS84
         spheroid to correct for ellipticity.
 
+        The distance (in degrees) and distancemeters are the distance
```
QuLogic: Can you put the two names in quotes or double backticks, so it's clear that `distancemeters` is not a typo?

obspy/clients/iris/tests/test_client.py
```diff
@@ -58,14 +58,18 @@ def test_distaz(self):
         client = Client()
         # normal request
         result = client.distaz(stalat=1.1, stalon=1.2, evtlat=3.2, evtlon=1.4)
-        self.assertAlmostEqual(result['distance'], 2.09554)
-        self.assertAlmostEqual(result['backazimuth'], 5.46946)
-        self.assertAlmostEqual(result['azimuth'], 185.47692)
+        self.assertAlmostEqual(result['distance'], 2.10256)
```
QuLogic: Is there a reason for the change in results? Just more correct? The new one seems to match geographiclib at least.

obspy/clients/iris/client.py
chad-iris: HTTPS is experimental and not a fully tested or supported for IRIS services.

obspy/clients/iris/tests/test_client.py
chad-iris: The service now uses GeographicLib and is (presumably) more accurate.

obspy/clients/iris/client.py
chad-iris: Done.  Also for the references to the other keys in the result for consistency.

krischer:
I added one more commit as the doctest for the traveltime service somehow required a slightly different rounding. I guess this is due to some service change? Could have been a while ago as we don't regularly run the networks tests with our CI runners as they just tend to be too unstable.

Side note: should we maybe have some test runners that run all the network tests at least once a day?

krischer:
Thanks for this!

megies:
> Side note: should we maybe have some test runners that run all the network tests at least once a day?

We could do this with our docker testbot that currently runs once a day..

krischer:
Would be great :-) Then we can at least manually check if some stop working completely.

PR #1544, opened 2016-10-03
MiniSEED with no blockette 1000

krischer:
This PR teaches ObsPy how to read MiniSEED files with no blockette 1000.

Test file is from @ThomasLecocq - I assume its okay to use two records of it here.

Don't merge yet as one change should probably be merged upstream into `libmseed`.

@chad-iris: This PR changes 1 thing in `libmseed` and works around an other problem:
* `libmseed` cannot correctly detect the record length for the last record if blockette 1000 is missing. https://github.com/obspy/obspy/commit/fe8e80386cc0241c8e274c01b95af4ac74ad442f contains a proposed fix. I am not sure it is optimal but it passes all our tests.
* If `msr_parse()` cannot detect the record length it returns `MINRECLEN` as an error code. I think it should return a specialized error code.

```C
      /* Found record but could not determine length */
      if ( detlen == 0 )
	{
	  return MINRECLEN;
	}
```


chad-iris:
These changes allow scenarios where record lengths are mis-identified just because the buffer is a valid length.  A 512-byte buffer containing the first 512 bytes of a 4096-byte record that does not have a blockette 1000 in the first 512-bytes (or at all) is identified incorrectly.  Not good.

I suggest changing the ObsPy reader to use the pattern supported by libmseed's API.

Here are the return values for ``msr_parse()``:

```
 * Return values:
 *   0 : Success, populates the supplied MSRecord.
 *  >0 : Data record detected but not enough data is present, the
 *       return value is a hint of how many more bytes are needed.
 *  <0 : libmseed error code (listed in libmseed.h) is returned.
```

Here is the supported sequence (used by other parts of libmseed and likely beyond):

Call ``msr_parse()`` to parse a record from the buffer, then

1. If a positive value is returned, aka there is definitely a record but it cannot be parsed, either: 
  1. Get more data and call ``msr_parse()`` again 
  2. If you _know_ the record length, because you know it's the end of a buffer or file, call ``msr_parse()`` again and specify the length explicitly
2. If zero is returned, handle the record and move on
3. If negative is returned, handle the error

Note the case of calling ``msr_parse()`` and needing to get more data, which is quite common, especially when reading chunked data.  Once that pattern is supported, handling the implied record length case is a small addition.  You can see this pattern in ``ms_readmsr_main()`` for reference.

Regarding the lower level functions, It is not the job of ``msr_parse()`` nor ``ms_detect()`` to determine the implied record length that is at the end of a buffer, only the caller has the context to be able to determine that scenario with any certainty.

chad-iris:
Something like this might be enough, adding a second call to msr_parse() after the existing first call:

```
        retcode = msr_parse ( (mseed+offset), buflen - offset, &msr, reclen, dataflag, verbose);

        // If record is detected but length is unknown, check if remaining buffer is a valid
        // length parse using the remaining buffer length as the record length.
        if (retcode > 0) {
          datasize = buflen - offset;
          if (datasize == 128 || datasize == 256 || datasize == 512 || datasize == 1024 ||
              datasize == 2048 || datasize == 4096 || datasize == 8192)
            retcode = msr_parse ( (mseed+offset), buflen - offset, &msr, datasize, dataflag, verbose);
        }
```

krischer:
Looks almost exactly like what I just pushed ;-)

krischer:
Thanks a lot for the detailed explanations! RTFM for me I guess ;-)

This PR now follows advised sequence and everything seems to work nicely.

ThomasLecocq:
:+1:  me happy :)

krischer:
One more question: If no blockette 1000 is present ObsPy now assumes big endian for the data. Is that safe?

obspy/io/mseed/src/obspy-readbuffer.c
```diff
@@ -311,54 +359,42 @@ readMSEEDBuffer (char *mseed, int buflen, Selections *selections, flag
...
             break;
         }
+        // msr_parse() returns > 0 if a data record has been detected but the buffer either has not enough
+        // data (this cannot happen with ObsPy's logic) or the last record has no Blockette 1000 and it cannot
+        // determine the record length because there is no next record (this can happen in ObsPy) - handle that
+        // case by just calling msr_parse() with an explicit record length set.
+        else if ( retcode > 0 && retcode < (buflen - offset)) {
+            int r_bytes = buflen - offset;
+            if ((r_bytes == 128) || (r_bytes == 256) || (r_bytes == 512) || (r_bytes == 1024) ||
+                    (r_bytes == 2048) || (r_bytes == 4096) || (r_bytes == 8192) || (r_bytes == 16384)) {
```
Jollyfant: I thought the record length can theoretically be anywhere between 256 and 2^256 bytes.

krischer: I guess its unlikely to happen but the latest version now has a more generic test allowing for all possible record lengths.

krischer:
All CI pass. No clue what coveralls problem is. IMHO ready to merge. Can someone please review?

megies:
> No clue what coveralls problem is.

off topic..
 - no clue why coveralls is still running, we disabled it IIRC
 - no clue why codecov is not showing up in many PRs, i thought we enabled it properly.. -_-

krischer:
> off topic..
> 
>     no clue why coveralls is still running, we disabled it IIRC
>     no clue why codecov is not showing up in many PRs, i thought we enabled it properly.. -_-

This PR is to the maintenance branch  - we still use the old coverage testing there.

ThomasLecocq:
Dunno how to review .. but if it read the test file I sent you without breaking anything else, I'd say it's good to go :-)

chad-iris:
> One more question: If no blockette 1000 is present ObsPy now assumes big endian for the data. Is that safe?

Hhmm, probably, since most data without 1000's are pretty old and little-endian was not common then.

But I'm not sure why you have to assume, libmseed's automatic byte order determination will work whether 1000s are included or not.

krischer:
> One more question: If no blockette 1000 is present ObsPy now assumes big endian for the data. Is that safe?
> 
> Hhmm, probably, since most data without 1000's are pretty old and little-endian was not common then.
> 
> But I'm not sure why you have to assume, libmseed's automatic byte order determination will work whether 1000s are included or not.

But that just determines the byte order of the header, right? And, if I remember correctly, `msr->byteorder` is only set if Blockette 1000 is present or the byteorder is forced through one of the environment variables.

I just checked the documentation and `libmseed` apparently always assumes identical header and data byte orders if Blockette 1000 is not present. I'll make some changes, so ObsPy does the same.

PR #1545, opened 2016-10-04
Travis: test *all* modules when building pushes to main branches

megies:
We use Travis in two ways:
 - **for PRs** we build the merge commit and report it in the PR's commit status. This should be reasonably fast to give timely feedback to authors of PRs and because relatively often PR branches will get pushed to and thus triggering Travis builds. Therefore we don't test network modules in this case (unless explicitly specified, see #1408).
 - **for pushes** to branches, we only trigger builds for our main branches (`master` and `maintenance_...`). This mostly happens when PRs eventually get merged, and sometimes for direct pushes of really small changes. So, not that many Travis builds are triggered in this way..

Therefore, I'm proposing to build *all* modules (i.e. also network modules) on *pushes* (to `master`/`maintenance_...`), so that we know when/if network modules break..

As a side effect, our build status might show up as "broken" many times, if we don't continuously monitor and fix network related issues.

There's no way to check the effects of this PR before merging it into `master`, so this issue is for documentation and discussion.

EDIT: As side notes..
 - we could remove `obspy.taup` from `DEFAULT_MODULES` and only trigger it manually on changes (in PRs), which would cut (PRs') test time on Travis in half, roughly
 - we could devise a helper routine in `obspy_github_api` that checks which modules are changed by a PR and automatically add tests for these..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,16 +153,16 @@ script:
         # we're building a pull request, so check what modules should be tested
         then export MODULELIST=`python -c "from obspy_github_api import get_module_test_list; print('obspy.' + ',obspy.'.join(get_module_test_list($TRAVIS_PULL_REQUEST)))"`
              export MODULELISTSPACES=`python -c "from obspy_github_api import get_module_test_list; print(' '.join(get_module_test_list($TRAVIS_PULL_REQUEST)))"`
-        # we're building a branch on the main repo
-        else export MODULELIST=`python -c "from obspy.core.util import DEFAULT_MODULES as MODULES; print('obspy.' + ',obspy.'.join(MODULES))"`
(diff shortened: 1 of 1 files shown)


krischer:
:+1: Go for it!

> we could remove obspy.taup from DEFAULT_MODULES and only trigger it manually on changes (in PRs), which would cut (PRs') test time on Travis in half, roughly

Or we just speed up the tests by not testing so very many traveltimes. What we currently do is actually way overkill I think.

megies:
Ok, gonna merge this one. Fingers crossed..

megies:
Alright, Travis is now testing pushes to `master` (and with 9aa1464 also to `maintenance_1.0.x`) rigorously for *all* modules. (see e.g. https://travis-ci.org/obspy/obspy/builds/165465047)

PR #1546, opened 2016-10-04
Submit coverage on test fails

megies:
Try to get Travis/coverage/codecov combo to report on coverage also when tests fail.


megies:
With these changes, coverage results are correctly uploaded to `codecov` also when (some) tests fail. I still don't understand why `codecov` doesn't set commit statuses, though..

megies:
Good to merge, if CI passes (still don't understand why codecov doesn't set a commit status if some tests failed..)

PR #1315, opened 2016-10-05
xseed Parser should provide channel orientation (azimuth/dip)

krischer:
..e.g. as a separate method or as additional items in `.get_inventory()`.

See http://lists.swapbytes.de/archives/obspy-users/2016-March/001995.html.


krischer:
I turned this issue into a PR so the test runners can have a go at it. A changelog entry would be nice but otherwise it looks good.

megies:
Thanks, Adam! Looks good.

:+1: For changelog.

aringler-usgs:
Okay, changelog has been updated!

PR #1547, opened 2016-10-05
docker testbot: remove images not on every run but only once per week and build master/maintenance only when neccessary

megies:
..so that we can sanely run the testbot more often.


megies:
Gonna merge this now, easier for me to debug (if necessary) on `master` afterwards..

PR #1271, opened 2016-10-06
Travis Mac build reports miss version information

krischer:
screenshot from 2016-02-19 00 47 26

### DIFF
--- a/obspy/core/util/version.py
+++ b/obspy/core/util/version.py
@@ -56,10 +56,13 @@ def call_git_describe(abbrev=10, dirty=True,
         p = check_output(['git', 'rev-parse', '--show-toplevel'],
                          cwd=OBSPY_ROOT, stderr=STDOUT)
         path = p.decode().strip()
-    except (OSError, CalledProcessError):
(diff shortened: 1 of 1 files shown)


krischer:
This is now a PR which raises a warning if the version number could not be determined. + An additional commit to "debug" the issue on travis.

krischer:
I just restarted the travis run as I don't understand the error message and it now also fails on Linux which I understand even less...

QuLogic:
Can you print out `os.environ`?

megies:

megies:
Base branch will have to be changed either to `master` or to `maintenance_1.2.x` once 1.2.0 is released.

megies:
what's travis?

trichter:
> what's travis?

:smile: 

But maybe we could keep the warning (not the print outs) at the end?

megies:
> But maybe we could keep the warning (not the print outs) at the end?

I didn't even check what the ticket was about exactly :rofl: 
Yea, sounds good

megies:
pushed it onto maintenance directly

PR #1548, opened 2016-10-06
Fix some network test fails

megies:
Been working on this report: http://tests.obspy.org/53048/

 - Arclink and Neic fails are resolved.
 - I can't reproduce the seedlink fails 
 - Will leave the syngine fails to @krischer.

+TESTS:ALL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,12 @@ install:
         requests
         jsonschema
   - source activate test-environment
+  # additional, optional packages that run some more tests but are not
+  # available for all archs. "--no-update-dependencies" keeps other packages at
+  # already installed version (ideally we should handle package versions
+  # through pinning or an environment file..)
+  - conda config --add channels conda-forge
+  - conda install --no-update-dependencies m2crypto || true
   # install packages not available via conda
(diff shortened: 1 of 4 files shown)


megies:
Rebased and force-pushed to see current status of fails.. @krischer maybe you can check those syngine fails when you find some time (and when CI is finished running the current commit in here)?

krischer:
The `clients.syngine` failures are not our fault and all related to https://github.com/iris-edu/irisws-syngine/issues/33.

megies:
Rebased and force-pushed. We can check again if there's anything left to do in the Travis reports..

krischer:
The syngine service has been fixed and I changes the one test to overly rely on the velocity model used to calculate the times. Should be good now.

megies:
Looks like we need to do some minimal change to arclink tests too.. https://travis-ci.org/obspy/obspy/jobs/170783439#L997

krischer:
Should be good now - let's wait for CI.

megies:
Gonna merge this now. Let's see how build status for maintenance and master evolve from there..

PR #1549, opened 2016-10-06
Don't upload staged test fail image in core tests to imgur

megies:
The image we use for testing that ImageComparison actually raises on a bad image probably got uploaded lots of times to Imgur.. was wondering for quite some time why I'm seeing imgur errors in our docker testbot. Pinned it down now.. ;-)


megies:
Yep. Works on my machine and the mocking should work across platforms.

PR #1550, opened 2016-10-06
Write blockette 100 if accuracy might otherwise be lost

krischer:
This PR causes `obspy.io.mseed` to always write Blockette 100 if sampling rate accuracy might be lost otherwise.

Blockette 100 can store a single precision floating point sampling rate. With this PR ObsPy will always write Blockette 100 if the sampling rate cannot exactly be represented with the sampling rate factor and multiplier in the fixed header.

The test might be a bit strict but it is the correct thing to do and the only price to pay are potentially slightly bigger MiniSEED files. "Clean" sampling rates still will not have Blockette 100.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,8 @@
    * ObsPy can now read MiniSEED files with micro-second wrap arounds.
      (see #1531)
    * ObsPy can now read MiniSEED files with no blockette 1000. (see #1544)
+   * ObsPy now always writes Blockette 100 if sampling rate accuracy is
+     otherwise lost. (see #1550)
  - obspy.io.sac:
    * `SACTrace.lpspol` and `lcalda` are `True` and `False` by default, when
       created via `SACTrace.from_obspy_trace` with a `Trace` that has no SAC
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -870,12 +870,37 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
                 clibmseed.msr_free(C.pointer(msr))
                 del msr
                 raise Exception('Error in msr_addblockette')
+
         # Only use Blockette 100 if necessary.
         # Determine if a blockette 100 will be needed to represent the input
         # sample rate or if the sample rate in the fixed section of the data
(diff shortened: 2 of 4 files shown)


megies:
Hmm, yeah. Can be seen as a fix, I guess..

krischer:
@chad-iris Is this a sensible thing to do?

It might also be something `libmseed` might want to do automatically in a future version. ObsPy uses `mst_pack()` to write MiniSEED.

megies:
Hmm.. a fail in SDS tests..? Just checked, using sampling rate `0.1` in there..

krischer:
I added a better comparison. `0.1` works now.

megies:
> Test looks good and comprehensive.

I guess github needs an option to review reviews.

chad-iris:
Seems perfectly sensible to me for ObsPy.  Forced inclusion of a Blockette 100 at the libmseed level is a little heavy handed, need to think through the implications.

In related news I'm adding code in libmseed to handle integer sample rates higher than 32,767 Hz.  So those cases will be handled better using the factor/multiplier representation.  But there are other sample rate ranges that I don't see any good factor/multiplier estimation method for, e.g. huge, non-integer sample rates, so including Blockette 100s is the right thing to do.

krischer:
> In related news I'm adding code in libmseed to handle integer sample rates higher than 32,767 Hz. So those cases will be handled better using the factor/multiplier representation. But there are other sample rate ranges that I don't see any good factor/multiplier estimation method for, e.g. huge, non-integer sample rates, so including Blockette 100s is the right thing to do.

Great :-) While at it: Could you change `ms_genfactmult()` to also return `-1` for very small sampling rates (e.g currently smaller than 1.0 / 32,767)? Right now it just checks if its larger than 0 and smaller than 32,767). Then ObsPy and other third party utils can just use the results of `ms_genfactmult()` to judge whether or not to include Blockette 100 and no longer need custom logic.

krischer:
Merging - `ms_genfactmult()` will change with the new libmseed version which we can take advantage of for when deciding if to use blockette 100 or not. But that has to be done after 2.18 has been released.

megies:
I think this PR might make a good argument to think about releasing 1.0.3

PR #1538, opened 2016-10-10
Calculation of azimuth is missing in obspy.taup.taup_geo.calc_dist

krischer:
In the documentation of obspy.taup.taup_geo.calc_dist it says that both distance and azimuth are calculated and returned. However, this is not the case, only distance_in_deg is calculated. It would be very simple to add the output of azimuth, as only one more line of code is needed (to be added just after https://github.com/obspy/obspy/blob/master/obspy/taup/taup_geo.py#L53-L60):

azimuth = g[azi1]

Alternatively, one could generate a new method calc_azimuth to not alter the existing calc_dist behaviour.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -58,6 +58,9 @@ master:
  - obspy.signal:
    * New obspy.signal.quality_control module to compute quality metrics from
      MiniSEED files. (see #1141)
+ - obspy.taup:
+   * Add obspy.taup.taup_geo.calc_dist_azi, a function to return the distance,
+     azimuth and backazimuth for a source - receiver pair. (see #1538)
 
 1.0.3:
  - obspy.io.mseed:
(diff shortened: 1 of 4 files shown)


krischer:
Thanks for this :-)

I took the liberty to convert this issue into a pull request from your branch. Definitely to `master` as its a new feature.

Looks pretty good otherwise. Could you please also add a simple test and a changelog entry?

andreww:
Ok - tests and the changelog added.

krischer:
There is test failure on Windows: http://tests.obspy.org/53543/

```
FAIL: test_taup_geo_calc_dist_azi (obspy.taup.tests.test_tau.TauPyModelTestCase)
Test for calc_dist
  File "obspy\taup\tests\test_tau.py", line 160, in test_taup_geo_calc_dist_azi
    self.assertAlmostEqual(backazi, 0.0, 5)
AssertionError: 360.0 != 0.0 within 5 places
```

This one tests the branch when `geographiclib` is not installed. I guess it just needs an extra modulo 360.

krischer:
Otherwise this looks good :-)

andreww:
@krischer - I can see two possible reasons for this, with I bit of luck this has fixed them both.

krischer:
That fixed it. Great :-)

The one test error in one of the docker test runners is unrelated. Thanks a lot for the quick response on this!

bschuber:
Many thanks also from my side for solving this!

megies:
Thanks @andreww for the prompt work on this!

PR #1552, opened 2016-10-10
sh: whitespace in header is not stripped anymore

trichter:
The added test failed, because lines and values from the Q header files are stripped while reading. This not only affected empty strings as in the test, but every header value with a whitespace anywhere if it happens to be written on the first or last column of the line.

### DIFF
--- a/obspy/io/sh/core.py
+++ b/obspy/io/sh/core.py
@@ -387,24 +387,21 @@ def _read_q(filename, headonly=False, data_directory=None, byteorder='=',
             raise IOError(msg % data_file)
         fh_data = open(data_file, 'rb')
     # loop through read header file
-    fh = open(filename, 'rt')
-    line = fh.readline()
-    cmtlines = int(line[5:7]) - 1
-    # comment lines
-    comments = []
-    for _i in range(0, cmtlines):
-        comments += [fh.readline()]
+    with open(filename, 'rt') as fh:
+        lines = fh.read().splitlines()
+    # number of comment lines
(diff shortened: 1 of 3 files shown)


megies:
I don't quite understand what this is about from the above description, but it sounds like a bug and I think this could use a short changelog entry. ;-)

trichter:
To be a bit less cryptic: The whitespace in the comment (`S000`) before `1`, after `9` and between the `*` was ignored.

```
43981 1 4
01|L000:0~ L001:2001~ R000:0.050000~ R026:1.000000~ S001:HMBCX~ S000: 12345* 
01| *6789 ~
01|
01|
```

obspy/io/sh/core.py
```diff
@@ -404,7 +404,7 @@ def _read_q(filename, headonly=False, data_directory=None, byteorder='=',
             id = cid
             i += 1
         traces.setdefault(i, '')
-        traces[i] += line[3:].strip()
+        traces[i] += line[3:].rstrip('\n')
```
krischer: I am not familiar with the SH file format at all so take this with a grain of salt: Why not just `rstrip()` to remove all trailing whitespaces? `\r\n` might be fairly common.

krischer:
I only have a single comment. Otherwise this looks pretty good.

obspy/io/sh/core.py
krischer: Ah I see. Just using `rstrip("\r\n")` should also do the trick. But the `splitlines()` would of course also be very clean.

PR #1558, opened 2016-10-18
Fix for the inventory pretty printing

krischer:
Some description strings might well have dots which breaks the old logic.

Should be safer now. Includes a test.

Target branch is `master` as the new compressed printing is only available in the master.


krischer:
Merging as its a small, non-critical change.

PR #1559, opened 2016-10-19
Improving documentation of the AR picker.

krischer:
The mailing list had a question about the AR picker: http://lists.swapbytes.de/archives/obspy-users/2016-October/002194.html

I used this opportunity to add the citation as well as generally improve the documentation for it.

If somebody wants to add more - please do so.


krischer:
+DOCS

krischer:
Looks like this now: http://docs.obspy.org/pull-requests/1559/packages/autogen/obspy.signal.trigger.ar_pick.html

Does somebody else want to add something? Adding formulas would be a lot of works as the algorithm has multiple steps and is fairly complex.

PR #1560, opened 2016-10-19
travis: install packages for testing through conda-forge rather than pip

megies:
Install additional packages in Travis through conda-forge, wherever possible, since it seems a bit more stable than pip which sometimes fails with what looks like network errors (see https://github.com/obspy/obspy/pull/1540#issuecomment-254552889). Also see related #1548 where we can install cartopy on some architectures via conda-forge.

The option `--no-update-dependencies` seems to properly leave alone package versions of already installed packages, so that our version matrix in Travis is untouched (safest would be to handle our dependency version matrix via pinning in conda, but that feature is unfortunately not accessible via `conda` CLI, so we would have to set it in a local file which path we need to figure out at runtime, so that's too messy and error prone)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,12 +106,12 @@ install:
   - conda config --add channels conda-forge
   # "--no-update-dependencies" to not change versions of already installed packages
   # github3.py is dependency of obspy_github_api
-  - conda install --no-update-dependencies
-        pep8-naming
-        github3.py
+  - ADDITIONAL_PACKAGES="pep8-naming github3.py codecov geographiclib"
+  - PIP_PACKAGES=""
+  - for PACKAGE in $ADDITIONAL_PACKAGES; do conda install --no-update-dependencies $PACKAGE || PIP_PACKAGES="$PACKAGE $PIP_PACKAGES"; done
   # install packages not available via conda
-  - pip install codecov
-  - pip install geographiclib
+  - if [[ $PIP_PACKAGES ]]; then pip install $PIP_PACKAGES; fi
+  # install packages from urls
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -100,14 +100,22 @@ install:
...
+  # github3.py is dependency of obspy_github_api
+  - conda install --no-update-dependencies
+        pep8-naming
+        github3.py
   # install packages not available via conda
   - pip install codecov
   - pip install geographiclib
   # current pyimgur stable release has a py3 incompatibility
   - pip install https://github.com/megies/PyImgur/archive/py3.zip
+  - pip install https://github.com/obspy/obspy_github_api/archive/0.5.0.zip
```
krischer: How about just always installing the master? Then this would not have to be changed all the time and changes to API helper get directly used. This is only for our internal use so it should be fine.

megies:
Need to change installation for packages missing on conda-forge.. (see [here](https://travis-ci.org/obspy/obspy/jobs/168875325#L422))

megies:
Hmm not sure what problem Travis has with the pip command..
This works on my computer like a charm:
```bash
PIP_PACKAGES=""
for PACKAGE in pep8-naming github3.py codecov geographiclib; do PIP_PACKAGES="$PACKAGE $PIP_PACKAGES"; done
pip install $PIP_PACKAGES
```

megies:
Changed base branch to `maintenance_1.1.x`. Needs a rebase on this base branch, eventually.

megies:
Closing this for now, as we don't have much problems with dependency installations currently. We can always have a look if we run into them again..

PR #1528, opened 2016-09-12
Add GSE2.0 bulletin read support

Brtle:
I did a module to read GSE2.0 bulletin. Since there was a GSE2 module for waveforms, I added my code into this module.

I also added the documentation but I'm not sure about the license. It's an old documentation from 1995. I found it [here](http://www.seismo.ethz.ch/prod/autodrm/manual/index_EN) and the `provisional_GSE2.1.pdf` file (which can also be found on this page) was already in the GSE2 module. I assume that's OK.

I didn't add it to `CHANGELOG.txt`, it may be good to do it.

This is my first pull request so feel free to tell me if I've missed something!

PR #1567, opened 2016-10-21
Add author to GSE2.0 bulletin module (#1528)

Brtle:
I did a module to read GSE2.0 bulletin (#1528) 1 month ago but I didn't specify that it was developed by the EOST. Is it possible to add it?

Thanks!


megies:
Sure, no problem at all.

PR #1519, opened 2016-10-24
SAC: writing to disk a new Trace object ignores custom header fields, if an arrival time is set

jkmacc-LANL:
Example:
```python
import numpy as np
from obspy import Trace, read

tr = Trace(np.zeros(1000))
tr.stats.delta = 0.01
tr.stats.station = 'XXX'
tr.stats.sac = {}
tr.stats.sac['stla'] = 10.
tr.stats.sac['stlo'] = -5.
tr.stats.sac['a'] = 12.34
tr.write('test.sac', format='SAC')

tr2 = read('test.sac')[0]
for k, v in sorted(tr2.stats.sac.items()):
    print('{}: {}'.format(k, v))
```
gives the following output:
```
b: 0.0
delta: 0.00999999977648
depmax: 0.0
depmen: 0.0
depmin: 0.0
e: 9.98999977112
iftype: 1
iztype: 9
kstnm: XXX
lcalda: 0
leven: 1
lovrok: 1
lpspol: 1
npts: 1000
nvhdr: 6
nzhour: 0
nzjday: 1
nzmin: 0
nzmsec: 0
nzsec: 0
nzyear: 1970
scale: 1.0
```
i.e., the custom header fields `stla`, `stlo` and `a` are not written to disk.
When removing the line:
```python
tr.stats.sac['a'] = 12.34
```
the two remaining custom fields are correctly written to disk, so I assume that this line is failing silently.

This bug (regression?) appears on both the current master and the maintenance branch.

### DIFF
--- a/obspy/io/sac/sactrace.py
+++ b/obspy/io/sac/sactrace.py
@@ -342,7 +342,7 @@ from obspy import Trace, UTCDateTime
 from obspy.geodetics import gps2dist_azimuth, kilometer2degrees
 
 from . import header as HD  # noqa
-from .util import SacError, SacHeaderError, SacHeaderTimeError
+from .util import SacError, SacHeaderError
 from . import util as _ut
 from . import arrayio as _io
 
--- a/obspy/io/sac/sactrace.py
+++ b/obspy/io/sac/sactrace.py
@@ -1286,13 +1286,15 @@ class SACTrace(object):
         :type keep_sac_header: bool
 
         """
-        try:
-            header = _ut.obspy_to_sac_header(trace.stats, keep_sac_header)
-        except SacError:
-            # not enough time info in old SAC header
-            # XXX: try to do something besides ignore the old header?
-            header = _ut.obspy_to_sac_header(trace.stats,
-                                             keep_sac_header=False)
+        header = _ut.obspy_to_sac_header(trace.stats)
+
+        if keep_sac_header and trace.stats.get('sac'):
+            try:
+                header = _ut.merge_sac_headers(trace.stats.sac, header)
+            except SacError:
+                # not enough time info in old SAC header
+                # XXX: do something besides ignore the old header
+                pass
 
         # handle the data headers
         data = trace.data
--- a/obspy/io/sac/sactrace.py
+++ b/obspy/io/sac/sactrace.py
@@ -342,7 +342,7 @@ from obspy import Trace, UTCDateTime
 from obspy.geodetics import gps2dist_azimuth, kilometer2degrees
 
 from . import header as HD  # noqa
-from .util import SacError, SacHeaderError
+from .util import SacError, SacHeaderError, SacHeaderTimeError
 from . import util as _ut
 from . import arrayio as _io
 
@@ -764,13 +764,9 @@ class SACTrace(object):
                   'lcalda': lcalda, 'lpspol': lpspol, 'lovrok': lovrok,
                   'internal0': internal0}
 
-        # required = ['delta', 'b', 'npts', ...]
-        # provided = locals()
-        # for hdr in required:
-        #     header[hdr] = kwargs.pop(hdr, provided[hdr])
-
         # combine header with remaining non-required args.
-        # XXX: user can put non-SAC key:value pairs into the header.
+        # user can put non-SAC key:value pairs into the header, but they're
+        # ignored on write.
         header.update(kwargs)
 
         # -------------------------- DATA ARRAY -------------------------------
@@ -1286,15 +1282,7 @@ class SACTrace(object):
         :type keep_sac_header: bool
 
         """
-        header = _ut.obspy_to_sac_header(trace.stats)
-
-        if keep_sac_header and trace.stats.get('sac'):
-            try:
-                header = _ut.merge_sac_headers(trace.stats.sac, header)
-            except SacError:
-                # not enough time info in old SAC header
-                # XXX: do something besides ignore the old header
-                pass
+        header = _ut.obspy_to_sac_header(trace.stats, keep_sac_header)
 
         # handle the data headers
         data = trace.data
--- a/obspy/io/sac/tests/test_core.py
+++ b/obspy/io/sac/tests/test_core.py
@@ -897,6 +897,7 @@ class CoreTestCase(unittest.TestCase):
         self.assertAlmostEqual(tr1.stats.sac.stlo, -5., places=4)
         self.assertAlmostEqual(tr1.stats.sac.a, 12.34, places=5)
 
+
 def suite():
     return unittest.makeSuite(CoreTestCase, 'test')
 
--- a/obspy/io/sac/tests/test_core.py
+++ b/obspy/io/sac/tests/test_core.py
@@ -85,8 +85,8 @@ class CoreTestCase(unittest.TestCase):
             with warnings.catch_warnings(record=True) as w:
                 warnings.simplefilter('always')
                 tr.write(tempfile, format='SACXY')
-                self.assertEqual(len(w), 1)
-                self.assertIn('reftime', str(w[-1].message))
+                self.assertEqual(len(w), 2)
+                self.assertIn('reference', str(w[-1].message))
             tr1 = read(tempfile)[0]
         self.assertEqual(tr, tr1)
 
@@ -625,8 +625,8 @@ class CoreTestCase(unittest.TestCase):
             with warnings.catch_warnings(record=True) as w:
                 warnings.simplefilter('always')
                 _write_sac_xy(st, fh)
-                self.assertEqual(len(w), 1)
-                self.assertIn('reftime', str(w[-1].message))
+                self.assertEqual(len(w), 2)
+                self.assertIn('reference', str(w[-1].message))
             fh.seek(0, 0)
             st2 = _read_sac_xy(fh)
 
@@ -643,8 +643,8 @@ class CoreTestCase(unittest.TestCase):
             with warnings.catch_warnings(record=True) as w:
                 warnings.simplefilter('always')
                 _write_sac_xy(st, tf)
-                self.assertEqual(len(w), 1)
-                self.assertIn('reftime', str(w[-1].message))
+                self.assertEqual(len(w), 2)
+                self.assertIn('reference', str(w[-1].message))
             tf.seek(0, 0)
             st2 = _read_sac_xy(tf)
 
@@ -779,11 +779,7 @@ class CoreTestCase(unittest.TestCase):
 
         with NamedTemporaryFile() as tf:
             tempfile = tf.name
-            with warnings.catch_warnings(record=True) as w:
-                warnings.simplefilter('always')
-                tr.write(tempfile, format='SAC')
-                self.assertEqual(len(w), 1)
-                self.assertIn('reftime', str(w[-1].message))
+            tr.write(tempfile, format='SAC')
             tr1 = read(tempfile)[0]
 
         # starttime made its way to SAC file
@@ -821,7 +817,7 @@ class CoreTestCase(unittest.TestCase):
             with warnings.catch_warnings(record=True) as w:
                 warnings.simplefilter('always')
                 tr.write(tempfile, format='SAC')
-                self.assertEqual(len(w), 1)
+                self.assertEqual(len(w), 2)
             tr1 = read(tempfile)[0]
 
         self.assertEqual(tr1.stats.starttime, tr.stats.starttime)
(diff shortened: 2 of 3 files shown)


jkmacc-LANL:
We've reached PR status.  Tests pass locally for me, let's hope they do here.  I removed an exception that was raised when `stats.sac` had no reference time, but had relative time headers, allowing things like your example to happen more easily.

claudiodsf:
I confirm that this fixes my use case.
When I save the trace to disk I get the following warnings, though:
```
.../obspy/io/sac/util.py:346: UserWarning: Invalid reference time detected.
  warnings.warn(msg)
.../obspy/io/sac/util.py:349: UserWarning: Relative times cannot be absolutely referenced.
  warnings.warn(msg)
```

jkmacc-LANL:
Thanks, Claudio.  I removed the warnings.  I think the user should be aware that a `trace.stats.sac` with no reference time and relative time headers like "a" and "o" is in an ambiguous state, but I can see how warnings could be annoying.  Maybe I should just beef up the `SACTrace.from_obspy_trace` docstring to reflect this.

claudiodsf:
> I think the user should be aware that a trace.stats.sac with no reference time and relative time headers like "a" and "o" is in an ambiguous state

Not sure: sac (the software) allows this. You can for instance generate a sine function `fg sin`, pick it using `ppk` and save it to disk. You will end up with a sac trace with an `a` marker but no reference time.

Otherwise, a solution might be to save the trace with an arbitrary date/time (e.g. the Unix epoch).

jkmacc-LANL:
> Not sure: sac (the software) allows this.

This is a fair point, but it becomes ambiguous when we have to merge with an arbitrary `.stats.sac` with a `.stats`, which always has an absolute `.starttime`.  For example: right now, if `.stats.sac` has no reference time but has a "b", we just [register the data starting at the "b"](https://github.com/obspy/obspy/blob/16a5405c48346fae78f3acaa5fd1737980be54af/obspy/io/sac/util.py#L350) and don't add any "nz" times to the SAC header, even if the user had set the `Trace.stats.starttime` to something real.  This is done for backwards compatibility, I think.  If `.stats.sac` has no reference time and *no* relative headers, though, [we do add `.starttime` as the reference time](https://github.com/obspy/obspy/blob/16a5405c48346fae78f3acaa5fd1737980be54af/obspy/io/sac/util.py#L352) and make an `ib`-type SAC file.  

I guess I would prefer to say, "For consistency's sake, ObsPy always writes an absolute time to the file if it can't find one in `.stats.sac`.  If you want more fine-grained control, use `obspy.io.sac.SACTrace`."  This would certainly break some code and tests, but I think it might increase consistency, shrink the burden of the code logic, and reduce the number of bug reports.  Thoughts?

> Otherwise, a solution might be to save the trace with an arbitrary date/time (e.g. the Unix epoch).

This sounds sounds a bit like my thoughts above, and I think it would break one of the "read-write-read with no changes" tests.

claudiodsf:
> "For consistency's sake, ObsPy always writes an absolute time to the file if it can't find one in `.stats.sac`. If you want more fine-grained control, use `obspy.io.sac.SACTrace`."

I actually agree with this. It would be consistent with the default behaviour of creating a `Trace` object from a numpy array, which sets `.stats.starttime` to the Unix epoch.

In the end, if people don't care about start time, they shouldn't care whether it isn't set or it is set to Jan 1, 1970.

What do you guys think ?

claudiodsf:
And, anyway, when you read into ObsPy a sac trace without time information, you get `.stats.starttime` set to Unix epoch. It seems to me misleading that you loose this time information when you write to a SAC file.

On the other hand, it's true that reading and writing that `sine.sac` generated from sac software (`fg sin; w sine.sac`, would change the file content, adding a reference time... Not easy!

claudiodsf:
I was looking at little bit more into this.

Basically, in your implementation, you always need a `.stats.starttime` and "nz" fields in `.stats.sac`. When writing to disk, you never change the "nz" fields and instead recompute the value of "b" as the difference between reference time and starttime.
If there are no "nz" fields, you cannot compute "b" and you don't write the reference time to disk. 

The only exception is when no relative time header variable (e.g., "b", "a", "t0", ...) is set. In this case you use `.stats.starttime` as reference time and set "b" to zero.

Is that correct?

jkmacc-LANL:
I think your interpretation is correct.

* **case:** `stats.sac` has a reference time ("nz" fields)
    - SAC "b" is written relative to this time (the difference between `starttime` and the reference time)
* **case:** `stats.sac` has no reference time, but has a "b"
    - Keep the old SAC "b", and write "e" appropriately.
    - Don't add a reference time ("nz" fields)
* **case**: `stats.sac` has no reference time and no "b"
    - Add a "b" of `0`
   - Add a reference time equal to `stats.starttime`

jkmacc-LANL:
> On the other hand, it's true that reading and writing that sine.sac generated from sac software (fg sin; w sine.sac, would change the file content, adding a reference time... Not easy!

True, but [there are tools for that](http://docs.obspy.org/packages/autogen/obspy.io.sac.sactrace.html#module-obspy.io.sac.sactrace).  How strong is ObsPy's obligation to read-write files with no modifications?  Is it true for all plugins?

claudiodsf:
Some further thought: I think the problem is that we are trying to map an object, the sac file, which might not have any absolute time information to an object, the ObsPy `Trace`, which always have an absolute time information.

So, either we modify the `Trace` object in order to work also without a time information (seems very difficult to me), or we accept that, in ObsPy, a sac file read into a `Trace` object will always get an absolute time information attached to it.

The latter is, by the way, the behaviour in `0.10.3`: reading `sine.sac` with no absolute time information and writing it again to disk, adds Jan 1, 1970 as reference time.

jkmacc-LANL:
> The latter is, by the way, the behaviour in 0.10.3: reading sine.sac with no absolute time information and writing it again to disk, adds Jan 1, 1970 as reference time.

I don't think I realized this.  It must've been untested behavior.

claudiodsf:
> How strong is ObsPy's obligation to read-write files with no modifications? Is it true for all plugins?

This is the key point to me. I think we should accept the ides of writing back to disk a sac file which always has a reference time. This was the case until `0.10.3`.
People needing more sophisticated behaviour should use `io.sac.sactrace`.

claudiodsf:
> I don't think I realized this. It must've been untested behavior.

Just tested it with an ObsPy `0.10.3` virtualenv.

jkmacc-LANL:
:+1: @claudiodsf   I'm curious how others feel, as well.

claudiodsf:
Sure, let's wait on other's reactions. Do you have a list of "SAC guys" to ping?

jkmacc-LANL:
Hello @krischer @megies @QuLogic @cja12 @mikecleveland @ThomasLecocq .

We're soliciting input on a `obspy.io.sac` behavior change (see the #1519 GitHub PR conversation).  Currently, when a SAC file with no "nz" headers (reference time) is read into ObsPy and re-written to disk with no modification, no reference time is added to the written SAC file.  The potential change would add a reference time of Jan 1, 1970 into the "nz" fields upon write in this case, following how an ObsPy `Trace` acquires a Jan 1, 1970 `starttime` if none is given.  Those wishing to preserve unset "nz" times could use `obspy.io.sac.SACTrace.read` and `obspy.io.sac.SACTrace.write`.

krischer:
:+1: I'm totally fine with this - seems to make a lot of sense within ObsPy's data model - every sample always has a time stamp - adding relative-time only traces might be interesting for the future but right now we do not have it. People desiring full control can, as mentioned, just use `obspy.io.sac.SACTrace`.

Another point is that any data within ObsPy's Stream or Trace classes should be the same no matter which format it is written to if possible. Most formats will have an absolute time by default. Also almost everybody will interpret a time of 1970-1-1 as "time not set".

Nonetheless I quite strongly feel that this PR should be rebased against the current master so its part of the next release where we can properly document the changed behavior. This likely only affects very few people but it would still suck if updating from `1.0.2` to `1.0.3` changes written files.

jkmacc-LANL:
I was imagining the change would go into a separate PR on `master` after this one is finished and changes brought over.

obspy/io/sac/sactrace.py
```diff
@@ -764,13 +764,9 @@ def __init__(self, leven=True, delta=1.0, b=0.0, e=0.0, iztype='ib',
...
 
-        # required = ['delta', 'b', 'npts', ...]
-        # provided = locals()
-        # for hdr in required:
-        #     header[hdr] = kwargs.pop(hdr, provided[hdr])
-
         # combine header with remaining non-required args.
-        # XXX: user can put non-SAC key:value pairs into the header.
+        # user can put non-SAC key:value pairs into the header, but they're
+        # ignored on write.
```
krischer: Any easy possibility to raise here? I think that for the `SACTrace` class it would be nice if it raises if an unknown header is set as that in all likeliness is an error.

krischer:
I see - I did not fully read the PR until now.

> I was imagining the change would go into a separate PR on master after this one is finished and changes brought over.

I see - I did not fully read the PR before.

megies:
> Hello @krischer @megies @QuLogic @cja12 @mikecleveland @ThomasLecocq .

I'll leave it up to anybody using SAC file format (I don't).

claudiodsf:
> Nonetheless I quite strongly feel that this PR should be rebased against the current master so its part of the next release where we can properly document the changed behavior. This likely only affects very few people but it would still suck if updating from `1.0.2` to `1.0.3` changes written files.

To me this is is a regression with respect to `0.10.3` behaviour and should be included in the next bug fix release (i.e., `1.0.3`).

On `0.10.3` you can do:
```python
import numpy as np
from obspy import Trace, UTCDateTime

tr = Trace(np.zeros(1000))
tr.stats.delta = 0.01
tr.stats.station = 'XXX'
tr.stats.starttime = UTCDateTime()
tr.stats.sac = {}
tr.stats.sac['a'] = 12.34
tr.write('test.sac', format='SAC')
```
and get a sac file with the correct reference time and the "a" header field set.

If you do the same thing in `1.0.2`, what you get is a sac file without reference time, unless you do not set the "a" field.

megies:
> If you do the same thing in 1.0.2, what you get is a sac file without reference time, unless you do not set the "a" field.

Without having looked at the details of the PR, what @claudiodsf points out really seems like something that should be fixed. A `Trace` that is initialized with no explicit `starttime` in `header` is by no means different from a trace that gets explicitly initialized with `starttime=UTCDateTime(1970, 1, 1)`, so output to SAC should in all cases honor the starttime, no?

claudiodsf:
>  A `Trace` that is initialized with no explicit `starttime` in `header` is by no means different from a trace that gets explicitly initialized with `starttime=UTCDateTime(1970, 1, 1)`, so output to SAC should in all cases honor the starttime, no?

For this specific aspect, I think we all agree.

The problematic point is when you read into an ObsPy `Trace` a sac file which has no absolute time information  in it (e.g., generated by the sac software) and you write it back to disk. If we implement the changes proposed here, it will be written back with a time information in it (Jan 1, 1970). That's because a `Trace` object must always have an absolute time information.

obspy/io/sac/sactrace.py
jkmacc-LANL: I was thinking that, during use, `SACTrace` could be used like any class, where attributes can be added as needed.  But during write, a warning about ignored headers is issued.  Do you think I should issue the warning here instead?

krischer:
> To me this is is a regression with respect to 0.10.3 behaviour and should be included in the next bug fix release (i.e., 1.0.3).

Hmm I see your point. `1.0.3` is fine with me then :-)

> The problematic point is when you read into an ObsPy Trace a sac file which has no absolute time information in it (e.g., generated by the sac software) and you write it back to disk. If we implement the changes proposed here, it will be written back with a time information in it (Jan 1, 1970). That's because a Trace object must always have an absolute time information.

That's fine - we almost never guarantee lossless round-trips into ObsPy and back. MiniSEED for example potentially looses some blockettes.

megies:
@claudiodsf @jkmacc-LANL if you think this is ready, go ahead and hit that merge button..

obspy/io/sac/sactrace.py
jkmacc-LANL: Should I do this in a separate PR?

claudiodsf:
@megies If we accept to always write to disk a reference time, this is not yet ready!

megies:
> @megies If we accept to always write to disk a reference time, this is not yet ready!

Alright.. I'm gonna stop making useless comments in this thread now.. ;-)

jkmacc-LANL:
@claudiodsf Do you think this PR is really the place for the proposed change?  I'd hoped to merge this one, and start a separate one.

obspy/io/sac/sactrace.py
krischer: Yea a separate PR would be good. You are also way deeper into all the sac issues so I trust your judgement whether or not this is a good idea.

krischer:
I also lost track a bit but the current code changes look fine to me - @jkmacc-LANL if you think this is ready and other issues should be delegated to new PRs, please hit the merge button!

PR #1569, opened 2016-10-24
ZMAP output: origins/magnitudes not set as preferred are not used

megies:
Dear all.

I have a problem saving catalog in a ZMAP format with built-in write function.

If I want to save my catalog as, lets say QUAKEML or KML, everything works fine, but with

cat.write("my_cat", format="ZMAP")

all the values i get in my_cat are NaN.

Thanks
(when i grab values manually, i can write outfile with true values without problem)

### DIFF
--- a/obspy/io/zmap/core.py
+++ b/obspy/io/zmap/core.py
@@ -138,6 +138,8 @@ class Pickler(object):
             strings = dict.fromkeys(self.zmap_columns, 'NaN')
             # origin
             origin = ev.preferred_origin()
+            if origin is None and ev.origins:
+                origin = ev.origins[0]
             if origin:
                 dec_year = self._decimal_year(origin.time)
                 dec_second = origin.time.second + \
@@ -157,6 +159,8 @@ class Pickler(object):
                 })
             # magnitude
             magnitude = ev.preferred_magnitude()
+            if magnitude is None and ev.magnitudes:
+                magnitude = ev.magnitudes[0]
             if magnitude:
                 strings.update({
                     'mag':     self._num2str(magnitude.mag),
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -23,6 +23,9 @@
  - obspy.io.stationxml:
    * Datetime fields are written with microseconds to StationXML if
      microseconds are present. (see #1511)
+ - obspy.io.zmap:
+   * Use first origin/magnitude when writing to zmap if no origin/magnitude is
+     set as preferred. (see #1569)
  - obspy.signal:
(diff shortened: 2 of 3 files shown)


megies:
This is now a PR fixing the issue. The problem was that the zmap write routine was only considering the preferred origin and it wasn't set (by the NLLOC read routine). Solution is to just use the first origin if none is set as preferred.

blazvicic:
Thank you very much.

megies:
Until this makes it into the next bugfix release you can work around this problem by setting the preferred origin yourself after reading the NonLinLoc output, like done in #1570.

megies:
Should be good to merge if CI passes.

QuLogic:
Tests need updating (and I guess no-one noticed the NaNs there...)

megies:
> Tests need updating (and I guess no-one noticed the NaNs there...)

Looking at [those comments in the test](https://github.com/obspy/obspy/pull/869/files#diff-4f5c89267fd393e538e5d79d3cf8c0a0R58), I think they did. @3rdcycle, did you really intend to write `NaN`s for events that have origin(s)/magnitude(s) but no preferred one set (I think this will be the case quite often, as many users might not be aware of that detail)..?

avuan:
Could be or not be related to the problem but actually ZMAP read and write routines are 
unable to manage fractions of seconds as pointed out  before.

megies:
> ZMAP read and write routines are unable to manage fractions of seconds as pointed out before.

Uhm. Pointed out before where exactly? Doesn't sound related to the fix in here, in any case. Please open another issue if there is some other problem.

megies:
Unless there's a veto from @3rdcycle, I think this is good to merge.

obspy/io/zmap/core.py
```diff
@@ -138,6 +139,10 @@ def _serialize(self, catalog):
             strings = dict.fromkeys(self.zmap_columns, 'NaN')
             # origin
             origin = ev.preferred_origin()
+            if origin is None and ev.origins:
+                msg = "No preferred origin set, using first origin."
+                warnings.warn(msg)
```
krischer: I don't think the warning is really necessary here.

obspy/io/zmap/core.py
```diff
@@ -157,6 +162,10 @@ def _serialize(self, catalog):
                 })
             # magnitude
             magnitude = ev.preferred_magnitude()
+            if magnitude is None and ev.magnitudes:
+                msg = "No preferred magnitude set, using first magnitude."
+                warnings.warn(msg)
```
krischer: Same here.

megies:
Adjusted, squashed and force pushed.

PR #1570, opened 2016-10-24
read event from nlloc: set preferred origin

megies:
With this PR, the preferred origin (being the only origin) is set on the event. See #1569 were this surfaced and caused a bit of confusion.


megies:
Should be good to merge if CI passes.

QuLogic:
Looks like tests need updating.

megies:
Should be all good now.

PR #1572, opened 2016-10-25
StationXML time output not in FDSN prescribed ISO 8601 time format

krischer:
The _format_time function in obspy/io/stationxml/core.py hard codes time as:

"%Y-%m-%dT%H:%M:%S+00:00" 
or 
"%Y-%m-%dT%H:%M:%S.%f+00:00"

Instead of the FDSN required ISO 8601 standard
YYYY-MM-DDTHH:MM:SS.ssssss

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -40,6 +40,8 @@ master:
     * Read and write support for nested custom tags (see #1463)
  - obspy.io.stationxml
     * Read and write support for custom tags (see #1024)
+    * No longer add the (unused) time zone field to StationXML datetimes to
+      follow the example of big data centers. (see #1572)
  - obspy.io.segy:
     * Iterative reading of large SEG-Y and SU files with
       `obspy.io.segy.segy.iread_segy` and `obspy.io.segy.segy.iread_su`.
--- a/obspy/io/stationxml/core.py
+++ b/obspy/io/stationxml/core.py
@@ -1421,9 +1421,9 @@ def _obj2tag(parent, tag_name, tag_value):
 
 def _format_time(value):
     if value.microsecond == 0:
-        return value.strftime("%Y-%m-%dT%H:%M:%S+00:00")
+        return value.strftime("%Y-%m-%dT%H:%M:%S")
     else:
-        return value.strftime("%Y-%m-%dT%H:%M:%S.%f+00:00")
+        return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
 
 
 def _read_element(prefix, ns, element, extra):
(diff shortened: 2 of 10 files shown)


krischer:
This issue is now a PR which removes the time zone specifier and also does not (yet) add the `Z`.

megies:
> Does anyone remember why we do it like this?

Hmm.. I think it probably comes from our all-out insanity to create a fully-random-all-possible-fields-used-UTF8-chars-included-autogenerated :tm: test file based on the XSD. That autogenerated file might have been the reason for our implementation..


> Thus StationXML should IMHO define a custom datetime type to enforce current best practices in the next version. Where can this be proposed?

FDSN mailing list or https://github.com/FDSN/StationXML/pulls, I guess.. both has pros and cons, I guess.

Mailing list..
 - reaches a lot of people
 - might spam/annoy most of them with such a detail problem.. 

[FDSN/StationXML](https://github.com/FDSN/StationXML/pulls)..
 - reaches only a half dozen people, I'd hazard
 - might be the right spot for such a detail though, and it can target the xsd directly with a changeset


>  This issue is now a PR which removes the time zone specifier and also does not (yet) add the Z.

If we'd consider this a fix we could even discuss it for the maintenance branch..

krischer:
> Hmm.. I think it probably comes from our all-out insanity to create a fully-random-all-possible-fields-used-UTF8-chars-included-autogenerated ™️ test file based on the XSD. That autogenerated file might have been the reason for our implementation..

This actually far predates that crazy test ;-)

> If we'd consider this a fix we could even discuss it for the maintenance branch.. (i'll add it to 1.1 milestone as its against master right now)

IMHO this should not go into the maintenance branch as it changes the output of lots of scripts and is thus a feature change which is reserved for major releases. Plus this is not really a bug.

megies:
> IMHO this should not go into the maintenance branch as it changes the output of lots of scripts and is thus a feature change which is reserved for major releases. Plus this is not really a bug.

Hmm.. well, according to the FDSN StationXML format spec pdf, you could consider it a bug, actually.

krischer:
> Hmm.. well, according to the FDSN StationXML format spec pdf, you could consider it a bug, actually.

This exists? Where?

megies:
> Hmm.. well, according to the FDSN StationXML format spec pdf, you could consider it a bug, actually.
> 
> This exists? Where?

Ah, sorry.. getting mixed up.. was looking at the FDSN WS spec pdf, sorry.

CTrabant:
> On Oct 25, 2016, at 12:27 PM, Lion Krischer <notifications@github.com> wrote:
> @CTrabant <https://github.com/CTrabant> Any reason that IRIS does not add Z at the end to specify that its UTC?
> 
As far as I know there are no times in FDSN interface specifications or data formats that are not UTC.   When everything is UTC you don't need time zone designators, so it really doesn't come up very often and a historical decision.  That being said, it probably would have been better to allow an optional Z from the beginning.  Hindsight is perfect.

CTrabant:
> I agree that the XSD schema is never enough to fully define the semantics of StationXML but it should be sufficient to define its syntax. Thus StationXML should IMHO define a custom datetime type to enforce current best practices in the next version. Where can this be proposed?
> 
Change procedure is documented here:
https://github.com/FDSN/StationXML

krischer:
Ok great :-) Everyone alright with merging this as is into `master`? CI looks good and ObsPy's output now follows the examples established by IRIS and SeisComP.

PR #1541, opened 2016-10-02
Increase test coverage

megies:
Trying to increase test coverage, going after some low hanging fruits..

PR #1563, opened 2016-10-26
NEIC client producing problematic data?

krischer:
One of my colleagues is getting some problematic data when we request from the neic client.  We weren't sure if this was something on our end or something in obspy.  The following code:

```
#!/usr/bin/env python

from obspy.clients.neic import Client
from obspy.core import UTCDateTime


client = Client()
day = UTCDateTime('2016255')
st = client.get_waveforms('NE' , 'BRYW' , '00' , 'HH2' , day , day + 24*60*60)
```
Results in an internal mseed warning:
```

python check.py 
/usr/lib/python2.7/site-packages/obspy-1.0.2-py2.7-linux-x86_64.egg/obspy/io/mseed/core.py:384: InternalMSEEDReadingWarning: readMSEEDBuffer(): Unknown error '121' in record starting at offset 9310720. The rest of the file will not be read.
  warnings.warn(*_i)
```

When we request it through a CWBQuery we don't get an error. In the following sense:
`
java -jar /APPS/bin/CWBQuery.jar -s "NEBRYW HH200" -b "2016/09/11 00:00:00" -d 86400 -t ms`

gives us a miniseed file and it reads correctly:

```
from obspy.core import read
>>> st = read('NEBRYW_HH200.ms')
>>> print(st)
1 Trace(s) in Stream:
NE.BRYW.00.HH2 | 2016-09-11T00:00:00.005000Z - 2016-09-11T23:59:59.995000Z | 100.0 Hz, 8640000 samples
>>> exit()

```

Can anyone confirm if this is an obspy issue (or an issue with something we are doing)?    

Thanks for your help,
Adam

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,6 +3,8 @@
    * properly pass through kwargs specified for Trace.plot() down to the
      low-level plotting routines (e.g. events were not shown properly in
      dayplot of a trace, see #1566)
+ - obspy.clients.neic:
+   * Better end of stream detection. (see #1563)
  - obspy.io.mseed:
    * ObsPy can now also read (Mini)SEED files with noise records. (see #1495)
    * ObsPy can now read records with a data-offset of zero. (see #1509, #1525)
--- a/obspy/clients/neic/client.py
+++ b/obspy/clients/neic/client.py
@@ -203,7 +203,7 @@ class Client(object):
                             if self.debug:
                                 print(ascdate(), asctime(), "read len",
                                       str(len(data)), " total", str(totlen))
-                            if data.find(b"EOR") >= 0:
+                            if data.find(b"<EOR>") >= 0:
                                 if self.debug:
                                     print(ascdate(), asctime(), b"<EOR> seen")
                                 tf.write(data[0:data.find(b"<EOR>")])
--- a/obspy/clients/neic/client.py
+++ b/obspy/clients/neic/client.py
@@ -196,15 +196,19 @@ class Client(object):
                     totlen = 0
                     while True:
                         try:
-                            data = s.recv(102400)
+                            # Recommended bufsize is a small power of 2.
+                            data = s.recv(4096)
                             if self.debug:
                                 print(ascdate(), asctime(), "read len",
                                       str(len(data)), " total", str(totlen))
-                            if data.find(b"<EOR>") >= 0:
+                            _pos = data.find(b"<EOR>")
+                            # <EOR> can be after every 512 bytes which seems to
+                            # be the record length cwb query uses.
+                            if _pos >= 0 and (_pos + totlen) % 512 == 0:
                                 if self.debug:
                                     print(ascdate(), asctime(), b"<EOR> seen")
-                                tf.write(data[0:data.find(b"<EOR>")])
-                                totlen += len(data[0:data.find(b"<EOR>")])
+                                tf.write(data[0:_pos])
+                                totlen += len(data[0:_pos])
                                 tf.seek(0)
                                 try:
                                     st = read(tf, 'MSEED')
--- a/obspy/clients/neic/client.py
+++ b/obspy/clients/neic/client.py
@@ -12,12 +12,12 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
 
+import io
 import socket
 import traceback
 from time import sleep
 
 from obspy import Stream, UTCDateTime, read
-from obspy.core.util import NamedTemporaryFile
 from obspy.core.util.decorator import deprecated
 from .util import ascdate, asctime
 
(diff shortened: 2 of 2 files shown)


krischer:
Actually this is our fault - sorry about that. This issue is now a PR fixing it.

The problem was that ObsPy used the existence of the string `EOR` to detect the end of the file. Now some combination of bytes in the MiniSEED file by chance had exactly the same binary representation as `EOR` in ASCII and this triggered it to stop downloading and trim the file.

It now checks for `<EOR>` which should reduce the chance of false positives. Is there a definition of the cwb query protocal? This check is still not particularly safe but the java tool does the same so maybe its the best that can be done?

krischer:
I added another commit so the client no longer uses a temporary file but buffers in memory.

krischer:
IMHO ready to merge. CI does not test these yet in any case and the one failure when manually running the tests is fixed in https://github.com/obspy/obspy/pull/1541.

aringler-usgs:
Hey Lion,

Thanks so much (I was slowly working towards the same conclusion).  I got the full data stream when I did the switch on line 210 you suggested.

Thanks!
Adam

megies:
> CI does not test these yet in any case

Wrong! 
+TESTS:clients.neic somewhere in the comments and Travis will add it to the executed test suites. :-)

krischer:
@aringler-usgs Is there an official definition for the protocol? It currently just checks if `<EOR>` is somewhere in the file and terminates it there. It it still quite likely that some random bits in a file will have exactly the same byte structure and it would thus terminate too early.

aringler-usgs:
Response from Dave Ketchum:
Adam,

Lion, Adam,

The “<EOR>” message is checked at the beginning of each miniSEED block (read 512 bytes, check, read another, check…).  Thus the ‘<EOR>’ is where the  sequence number would be stored in miniSEED, so it is impossible for it to appear randomly.  If the code is just reading a stream of bytes and looking for the tag, that would indeed be bad as it could randomly occur.

The above is how CWBQuery works and  (I think) how I wrote the initial contributed python to obspy which read from the CWB.

Dave

krischer:
I think the message is not complete. I guess it talks about the `<EOR>` message? MiniSEED records can have many different length so the 512 bytes check is not really sufficient. Or does CWBQuery only serve 512 byte records? If yes then we can just check if position of `<EOR>` modulo 512 is 0 and then break the connection there - that should be reasonably safe.

aringler-usgs:
Lion,

I believe the CWB only deals in 512 by blocks so that would be safe.

aringler-usgs:
Adam,

Lion, Adam,

The “<EOR>” message is checked at the beginning of each miniSEED block (read 512 bytes, check, read another, check…).  Thus the ‘<EOR>’ is where the  sequence number would be stored in miniSEED, so it is impossible for it to appear randomly.  If the code is just reading a stream of bytes and looking for the tag, that would indeed be bad as it could randomly occur.  

The above is how CWBQuery works and  (I think) how I wrote the initial contributed python to obspy which read from the CWB. 

Dave

> On Oct 26, 2016, at 1:36 PM, Ringler, Adam <aringler@usgs.gov> wrote:
> 
> Dave,
> 
> Is there a protocol for the end of a request when doing a CWBquery?  
> 
> Thanks for your help,
> Adam
> 
> 
> 
> On Wed, Oct 26, 2016 at 12:42 PM, Lion Krischer <notifications@github.com> wrote:
> @aringler-usgs Is there an official definition for the protocol? It currently just checks if <EOR> is somewhere in the file and terminates it there. It it still quite likely that some random bits in a file will have exactly the same byte structure and it would thus terminate too early.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 
> 

David Ketchum - contractor          Office:  303 273-8479
ketchum@usgs.gov                    Cell :   303 520-5840

aringler-usgs:
Dave,

Is there a protocol for the end of a request when doing a CWBquery?

Thanks for your help,
Adam

krischer:
> The above is how CWBQuery works and (I think) how I wrote the initial contributed python to obspy which read from the CWB.

The internal logic has always been how it was before this PR.


In any case - now it should be safe enough. It checks for `<EOR>` but only accepts it if it is at a position that cleanly divides by 512. This works perfectly fine for 512 byte records but also for bigger ones.

This is IMHO ready to merge once CI passes and if no objections come up.

aringler-usgs:
Hey Guys,

Thanks so much for fixing this.  I really appreciate it!

megies:
Good to know that somebody actually uses that client.. ;-)
Always hard to judge if something is even used..

megies:
There's some neic test fails on some travis builds.. not sure what it is we should print those streams in the assertion like..
```python
    self.assertEqual(st, st2, msg="Streams not equal:\n{!s}\n{!s}".format(st, st2))
```

krischer:
Did you not fix this in https://github.com/obspy/obspy/pull/1541?

krischer:
Which is of course not yet merged.

megies:
You mean #1548 I think, but yes.. could be. I'll merge that one, we can have an eye on the build status of master/maintenance.. our docker bot tests all modules for the tips of those.

PR #1576, opened 2016-10-27
make strptime a staticmethod

megies:
Just makes the code a little bit cleaner.. and avoid unnecessary instantiation of UTCDateTime objects.


megies:
Wanted to make that change a long time ago, but thought that it could break existing code. Very interesting that `@staticmethod` intercepts and gets rid of `self` automatically and adjustes the call..

Issue #1446, opened 2016-06-08
Guralpe Compressed Format (GCF) read support


rannof:
Hi,
It there a read plugin for GCF format? I made a module for reading waveforms in this format (no writing option yet). The module is based on one of the io formats (core,libgcf,__init__), but I think there is more to it in order to add it to obspy read function. 
Also, I'm not sure about copyrights, though all format specifications are available online with license details: [http://www.guralp.com/apps/ok?doc=GCF_Intro](http://www.guralp.com/apps/ok?doc=GCF_Intro)
If there is an interest, please let me know.

megies:
The code itself, if written by yourself and not reusing libraries (from Guralp or some other third party), is not bound to any licensing from third parties. We can just include it under LGPL that is the license for all of ObsPy (see #827).
The specs seem to have a strange license, so we'll just simply not include that document within ObsPy.

If you have code to read gcf format, we'd be happy to include it. Just make a new branch off `master` and drop what you have under `io/gcf/`. I can make this issue into a PR then. I've added you to developers team, if you push your branch to obspy/obspy it's easier for other people to add changes. Once we see what is already there we can help with comments on what to do next.

rannof:
OK, I added a pull request #1449 with the module. I also included a utility for converting files to other formats and just added it to io module (not sure where is the right place). I'm still new to the github and not sure if I done things properly.
Cheers,
Ran

megies:
Thanks for the PR. Next time just push the branch and let us know, we [could have made this very ticket into a PR](https://github.com/obspy/obspy/wiki/Interesting-Reading-on-Git%2C-GitHub), avoiding duplicated tickets.

PR #1579, opened 2016-11-02
io.gcf: add alternative gcf read/write implementation

megies:
This was sent to me by email by @maratumba, including it here so that interested people can have a look.

Has not been inspected or cleaned up in any form yet.

See #1446 and #1449.

### DIFF
--- a/obspy/io/gcf/gcfio.py
+++ b/obspy/io/gcf/gcfio.py
@@ -1,0 +1,346 @@
+# -*- coding: utf-8 -*-
+
+"""
+GCF Block
+
+Tasks:
+add channel name conversion (HHZ, etc)
+
+
+"""
+
+import numpy as np
+import struct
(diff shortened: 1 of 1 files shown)


krischer:
This only appears to read and not write.

Also does this add any extra functionality compared to the current `io.gcf` module? If not it is IMHO not worth the effort to bring this into a form where it can be integrated into ObsPy. But who am I to say.

@rannof: Could you please also have a look and give us your impressions?

megies:
> Also does this add any extra functionality compared to the current io.gcf module?

No idea, did not check.

> If not it is IMHO not worth the effort to bring this into a form where it can be integrated into ObsPy.

Sure, no need to integrate it. But since he code was sent to me I decided to put it here, just in case it might be useful at some point for somebody.

We can also close this PR, if anybody wants to investigate further at some point we can always reopen this one.

PR #1580, opened 2016-11-04
Enable 32bit linux on travis CI

mbyt:
Motivated by gitter conversation:

> byt: megies is it worth a try, getting travis run 32bit? We had
> quite _some_ issues with 32bit and mseed... There are three
> possibilities I could think of: (i) cross compile and run
> directly on travis (ii) use a chroot as numpy does, (iii)
> use a docker test.  I would go for (i).

> megies: sure, would not hurt to have a 32bit test in our main CI..
> no idea how how much work it is..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ branches:
     - master
     - /^maintenance_[0-9]+\.[0-9]+\.x$/
 
+sudo: true
+
 matrix:
   include:
     # do one build run with our minimum dependencies
@@ -20,13 +22,12 @@ matrix:
       env: PYTHON_VERSION=3.4
     - os: linux
       env: PYTHON_VERSION=3.5
-
+    - os: linux
+      env: PYTHON_VERSION=3.5 ARCHITECTURE_32BIT="True"
     - os: osx
       env: PYTHON_VERSION=2.7
     - os: osx
       env: PYTHON_VERSION=3.5
-
-sudo: false
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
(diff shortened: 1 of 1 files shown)


mbyt:
Available packages on travis ubuntu 12.04 can be found here:
* https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise

Two links which indicate i386 on travis does not work well / be complicated:
* https://github.com/travis-ci/travis-ci/issues/986
* https://github.com/scikit-learn/scikit-learn/pull/3393

Actually, the results of 04ab908 look promising to me.

mbyt:
Haha, 0643f84 actually worked: http://tests.obspy.org/57771/ :-)

I will cleanup, and rebase this pull request.

.travis.yml
```diff
@@ -5,39 +5,54 @@ branches:
     - master
     - /^maintenance_[0-9]+\.[0-9]+\.x$/
 
+sudo: false
```
QuLogic: How did we not have this already? Container builds are way faster...

.travis.yml
```diff
@@ -5,39 +5,54 @@ branches:
...
     else
       export OS="Linux";
     fi
+  # OPT used by setuptools patches contained in newer numpy versions
+  - |
+      if [[ "$ARCHITECTURE_32BIT" == "True" ]]; then
+        export ARCH=""
+        export OPT="-m32"
+        ls /etc/dpkg/dpkg.cfg.d/
+        sudo sh -c "echo 'foreign-architecture i386' > /etc/dpkg/dpkg.cfg.d/multiarch"
```
QuLogic: I wonder if there's a way to do this without sudo; would be faster that way...

.travis.yml
```diff
@@ -5,39 +5,54 @@ branches:
...
+        sudo apt-get install libstdc++6:i386
+        sudo apt-get install gcc-multilib
+        gcc --version
+      else
+        export ARCH="_64"
```
QuLogic: I think it's preferable to have `ARCH` be the full name.

.travis.yml
mbyt: just checked, seems to be on by default. However, need to be cautious, if there are changes in future versions.

mbyt:
just merged this one :). @megies I am a newbie to the new github options:
* create and merge commit
* squash and commit
* rebase and merge

I did the squash manually. What is your recommended way: create a merge commit (to my understanding that's the way we did it so far) or rebase and merge (nice no merge commit necessary, but I am unsure about side-effects).

megies:
So far we always did the usual **create a merge commit**, I think. In the history it's sometimes good to see those merge commits, I think, because you immediately see what PRs were merged recently. Squashing should be done manually, I think. Many times we want to keep intermediate commits, because they show certain problems etc.

mbyt:
sounds sensible to me, thanks!

PR #1573, opened 2016-11-08
ZMAP output: fractions of seconds neglected?

megies:
@avuan what exactly is the problem? Fractions of seconds neglected during writing or during reading? Can you provide a short self-contained example?

CC @3rdcycle


megies:
This should fix it. It's totally unclear to me what that test that is failing now is useful for actually, in one single test case it's reading/writing back and forth multiple times with some hacking around of time variables in between.

megies:
CI fails are unrelated, so merging.

PR #1585, opened 2016-11-08
new implementation of cross-correlation

trichter:
This is a spin-off of #841. I'd like to replace the existing `xcorr` function with a new one based on SciPy, because the current implementation has several [limitations](https://github.com/obspy/obspy/pull/841#issuecomment-256928381). Are there any objections?

Things to discuss:
1. `xcorr` currently returns `shift, coefficient` and on request the full correlation function. I'd like to change it so that it always returns `shift, coefficient, full_xcorr`, because the full cross-correlation is often needed and always calculated.
2. It would also be possible to return only `full_xcorr`. Shift and the corresponding coefficient can be easily calculated from the function. But I don't know how this could be properly introduced without a function with new name and depreciation of the old function.
3. I think for `len(a)!=len(b)` the cross-correlation function should be aligned as done in SciPy (which corresponds to zero-padding of the shorter signal on both ends). This would be somewhat different to the current behavior, but on the other hand the current function is only designed for `len(a)==len(b)`and not usable for `len(a)>len(b)`.
4. Do we always return a function with `2*num_len+1` samples? How to handle odd `len(a)-len(b)`? Naturally, the returned function should have only `2*num_len` samples to have lag time 0 in the middle of the function. Alternatively, the function could have `2*num_len+1` samples, but zero lag time would be at `+` or `-0.5*dt`.
5. Normalization would be done with the sums over the whole squared signals to guarantee that the correlation is always smaller than 1. This is different to the current implementation for `len(a)!=len(b)`.
6. New signature:  `xcorr(a, b, shift_len, demean=True, normalize=True)`.


trichter:
OK, I uploaded the new implementation. Please state your opinions on the code. The old implementation is still available as ``_xcorr_old_implementation`` so you can compare output and performance. If you like I can also remove this ``full_xcorr`` depreciation again, but I'd think most people are using xcorr with ``full_xcorr=True``.

trichter:
+DOCS please

megies:
@trichter putting +DOCS once is enough.. Docs are built already (at least for the previous commit), seems the setting the commit status failed for some reason: http://docs.obspy.org/pull-requests/

trichter:
Ah, thanks. Maybe because the first DOCS was an edit.

trichter:
[This](https://travis-ci.org/obspy/obspy/jobs/174284750#L875) fail is a bit strange. `round` does not accept a value of type `np.float32` due to `future`.
[This](https://travis-ci.org/obspy/obspy/jobs/174284750#L852) fail is strange, too. flake8 does not accept a function starting with double `_`?

Hopefully I worked around both.

trichter:
If tests pass this is ready for review.

trichter:
Finally, all tests pass. I don't know why codecov/changes complains.

obspy/signal/cross_correlation.py
```diff
@@ -26,14 +26,166 @@
...
+        num = mid
+    if num > mid:
+        msg = 'sift_len too large, use largest possible shift %d instead'
+        warnings.warn(msg % mid)
+        num = mid
+    return c[mid - num:mid + num + len(c) % 2]
+
+
+def xcorr(tr1, tr2, shift_len, demean=True, normalize=True, domain='freq',
+          full_xcorr=False, abs_max=True):
```
megies: `full_xcorr` was at position 4 earlier, so this might have strange effects on peoples' codes... imagine:

```python
xcorr(tr1, tr2, shift_len, True)  # would likely break, expects full xcorr returned
xcorr(tr1, tr2, shift_len, False)  # would not break, but unexpectedly set `demean=False`
```

obspy/signal/cross_correlation.py
```diff
@@ -26,14 +26,166 @@
...
+    if normalize:
+        stdev = (np.sum(a ** 2)) ** 0.5 * (np.sum(b ** 2)) ** 0.5
+        if stdev == 0:
+            # set stdev to 1 to prevent division by 0
+            # cross-correlation function will have only zeros anyway
+            stdev = 1
+    else:
+        stdev = 1
+    # choose the usually faster xcorr method for each domain
+    _xcorr = _xcorr_slice if domain == 'freq' else _xcorr_padzeros
```
megies: This is not typo-safe.. imagine `xcorr(..., domain='frqe')` or `xcorr(..., domain='FREQ')`.
This should be an `if/elif/else: ValueError()`

obspy/signal/cross_correlation.py
```diff
@@ -26,14 +26,166 @@
...
+        msg = ('Keyword full_xcorr will default to True starting with the next'
+               ' major release (v1.2) and will be removed in the subsquent '
+               'major release (v.1.3). Please set full_xcorr=True.')
+        warnings.warn(msg, ObsPyDeprecationWarning)
+        return shift, value
+
+    return shift, value, c
+
+
+def _xcorr_old_implementation(tr1, tr2, shift_len, full_xcorr=False):
```
megies: Maybe put a deprecation warning as first line inside this function? I don't think we want to drag around tons of redundant implementations forever..?

obspy/signal/tests/test_cross_correlation.py
```diff
@@ -6,22 +6,222 @@
...
     def setUp(self):
         # directory where the test files are located
         self.path = os.path.join(os.path.dirname(__file__), 'data')
         self.path_images = os.path.join(os.path.dirname(__file__), 'images')
+        self.a = np.sin(np.linspace(0, 10, 101))
+        self.b = 5 * np.roll(self.a, 5)
+        self.c = 5 * np.roll(self.a[:81], 5)
+
+    def test_xcorr(self):
+        """
```
megies: Moving around these tests makes it hard to review.. are there changes in the old tests?

megies:
Alright.. I don't have time to fully evaluate the possible effects of these changes. But I also don't want to veto them, I guess there's a good reason for these changes.

Main concerns are..
 - the position of `full_xcorr` in the call, moving it around is dangerous (I agree that it shouldn't be there in the first place and just having the caller drop the correlation function would be much simpler, but we can't change history..)
 - tests were moved around, so i can't easily see if old tests had to be adapted and where and how badly..

Anybody else can comment on this one, e.g. @ThomasLecocq ?

trichter:
Thanks for the review.

> I guess there's a good reason for these changes.

Seems to be faster with less bugs now.

> - the position of `full_xcorr` in the call, moving it around is dangerous (I agree that it shouldn't be there in the first place and just having the caller drop the correlation function would be much simpler, but we can't change history..)

OK. Do you agree on the depreciation route for the `full_xcorr` kwarg?

> - tests were moved around, so i can't easily see if old tests had to be adapted and where and how badly..

There are no changes except this one
```
+        # comment next line, because shift is not unique
+        # self.assertEqual(shift, 0)
```
because a different shift is returned for a cross-correlation full of zeros (hard coded in old version).

I can delete `_xcorr_old_implementation`. It's just for the review. I guess `xcorr.c` should stay because of the SRL paper?

trichter:
All points are addressed. Ready for next round of review or merging.

krischer:
Sorry for joining the discussion so late.

In general I like this as it is more powerful, apparently faster, and also well tested.

My main peeve with it is, that it (even with the currently set `full_xcorr=False`) changes the function call signature from

```python
shift, value = xcorr(...)
```

to 

```python
shift, value, corr = xcorr(...)
```

This is quite a major change especially as we are trying the keep the API as stable as possible after we released 1.0. A lot of codes do use the current CC implementation of ObsPy (I know I have a couple) and these *will all require updating* (this is no problem for us but the average user will be very annoyed). I personally would rather avoid making this change to ObsPy to be honest.

I suggest the following: Leave the old CC implementation as is (or write a wrapper reproducing its behavior) and define a new `obspy.signal.cross_correlation.cross_correlate()` function which serves as a common entry point for different cross correlation implementations:

```python
def cross_correlations(cc_type, a, b, **kwargs):
    if cc_type == "td_full_xcorr": # Or whatever you want to call it
        return toms_new_xcorr(a, b, **kwargs)
    else:
        raise NotImplementedError
```

This should probably be replaced with a plug-in interface but you don't have to do this now. As long as the function call signature has been designed with a potential plug-in interface in mind I'm fine with it.

Opinions?

trichter:
> Sorry for joining the discussion so late.

NP

> My main peeve with it is, that it (even with the currently set full_xcorr=False) changes the function call signature ...
> This is quite a major change especially as we are trying the keep the API as stable as possible after we released 1.0. A lot of codes do use the current CC implementation of ObsPy (I know I have a couple) and these will all require updating (this is no problem for us but the average user will be very annoyed). I personally would rather avoid making this change to ObsPy to be honest.

It does not change the call signature for `full_xcorr=False` which is still the default. But I get your point. Getting this function to return `shift, value, xcorr_fun` without the `full_xcorr` kwarg needs two rounds of depreciation and needs the user to change his code two times....

If we don't wan't this, there are two options:

1. Leave interface as it is in v1.0 and don't depreciate `full_xcorr`
2. Same as 1 and introduce this new function as suggested by you.

> I suggest the following: Leave the old CC implementation as is (or write a wrapper reproducing its behavior) and define a new obspy.signal.cross_correlation.cross_correlate() function which serves as a common entry point for different cross correlation implementations:
  def cross_correlations(cc_type, a, b, **kwargs):
    if cc_type == "td_full_xcorr": # Or whatever you want to call it
        return toms_new_xcorr(a, b, **kwargs)
    else:
        raise NotImplementedError
This should probably be replaced with a plug-in interface but you don't have to do this now. As long as the function call signature has been designed with a potential plug-in interface in mind I'm fine with it.
Opinions?

The two advantages I see with the second option are firstly I don't need to write `_, _, x = xcorr(..., full_xcorr=True)` and secondly new types of cross correlation can be added later. Honestly, I don't see a lot of different ways to calculate cross-correlation functions apart from the "default" and phase cross-correlation.

OK, so if we follow your suggestion we need to quickly decide how to call this function and how is the interface. First, how to call it? `obspy.signal.cross_correlation.cross_correlate()` is a bit long, could it live in `obspy.signal`? **`obspy.signal.correlate`**, `obspy.signal.cross_correlate`, `obspy.signal.xcorr`?

I'd like to let it calculate the "normal" cross correlation if the type is not specified. 
So maybe something like this:?
```
def correlate(a, b, shift, type='default', **kwargs):
    ....
    return xcorr_fun
```

`shift` could also be a keyword, but I think it is often needed for the default xcorr.

What do you think? What's the opinion of other people?

megies:
I don't know.. one the one hand adding more wrappers calling other routines might be confusing for users. On the other hand, `xcorr` is kind of an old abbreviated naming style that we do not use anymore, so it might not be a totally bad idea to introduce that wrapper (e.g. name it `signal.cross_correlation.correlate`).. also we could then leave the old implementation in and add the new ones on top, plugged into that wrapper via three `type='...'` choices.

If we decide against a new wrapper, I would vote to just leave it in and as is to avoid more maintenance work and avoid breaking peoples' codes (I totally agree that this `full_xcorr` kwarg shouldn't be in there in the first place but removing it means problems..).

obspy/signal/cross_correlation.py
```diff
@@ -26,14 +26,168 @@
...
+
+
+def _xcorr_slice(a, b, num, domain='freq'):
+    """
+    Cross-correlation using SciPy with mode='full' and subsequent slicing
+    """
+    if domain == 'freq':
+        c = fftconvolve(a, b[::-1], 'full')
+    else:
+        c = correlate(a, b, 'full')
```
megies: Please also make this typo safe, see other comment.. (if/elif/else raise ValueError)

obspy/signal/cross_correlation.py
```diff
@@ -26,14 +26,168 @@
...
+        num = (len(a) + len(b) - 1) // 2
+    dif = len(a) - len(b) - 2 * num
+    if dif > 0:
+        b = _pad_zeros(b, dif // 2)
+    else:
+        a = _pad_zeros(a, -dif // 2)
+    if domain == 'freq':
+        c = fftconvolve(a, b[::-1], 'valid')
+    else:
+        c = correlate(a, b, 'valid')
```
megies: Please also make this typo safe, see other comment.. (if/elif/else raise ValueError)

trichter:
OK, so let it be `correlate`. Still, I think the old implementation should be depreciated. `xcorr` can be a wrapper to call `correlate` which behaves the same way as in previous ObsPy versions. Last question, should I depreciate the `xcorr` function and suggest people to use `correlate`?

EPAnt:
Greetings to all,

I don't have much to add. I just wanted to comment and compliment Tom
Eulenfeld for his effort. Thank you Tom!
At the time I was more actively involved with ObsPy, this was one of the
issues I was tackling as I also needed a full cross-correlation on the time
domain. I was discussing it with Leon at the time before I had to abandon
developing the full solution.
I know this is going to be beneficial to many, I know it would have
certainly helped in speeding up my work back then ;-) (I ended up using old
fortran stuff and having to read the output to make sense of it in Python
again, quite an indirect route).

Congrats and keep up the good work!

Emanuel Antunes



2016-12-19 10:08 GMT+00:00 Tom Eulenfeld <notifications@github.com>:

trichter:
Consequently, should we introduce a new function `correlate_3c` ?

Thank you @EPAnt. Of course, this peace of code will only be a tiny part of ObsPy and basically wraps other functions inside SciPy/NumPy. I pass kudos to these two projects.

megies:
We're pushing hard to finally get 1.1.0 out and have our hands full.. since this is changing existing functionality (that we also use in other parts of the code base) and also changing existing tests it's time consuming to review and I have a feeling that there hasn't been an agreement how to move forward with this one.. so, I'm tempted to bump this PR to 1.2.0 which would mean October '17 if we stick to the fixed release schedules of April+October that we kind of agreed on.. other opinions? @trichter, @krischer?

trichter:
I am for merging this one ;)

It fixes a lot of bugs in the old implementation by replacing it. Only one line is commented in the tests (see commit c76b3d749373c3e513d30d8abcb8fb302c11751b and changes in `test_util.py`) and test coverage is enhanced also for existing codes. There is only the open question weather to have one single correlation function or a plugin system. I am for the first option, but the current implementation with the `type` keyword in the `correlate` function leaves that open.

If another thorough review or more discussion is needed, than bumping to 1.2 is of course OK.

obspy/signal/cross_correlation.py
```diff
@@ -81,9 +221,73 @@ def xcorr(tr1, tr2, shift_len, full_xcorr=False):
...
+    >>> a
+    0
     >>> round(b, 7)
     1.0
     """
+    from obspy.core.util.deprecation_helpers import ObsPyDeprecationWarning
+    msg = ('_xcorr_old_implementation is the old implementation of xcorr '
+           'used in ObsPy versions<1.1. It is kept in Obspy v1.1 for reference'
+           ', but will be removed with the next major release.')
+    warnings.warn(msg, ObsPyDeprecationWarning)
```
krischer: Can you also please add this (with a slightly modified error message of course) to the `xcorr()` function to discourage people from using it?

obspy/signal/cross_correlation.py
krischer: I second @megies suggestion. The current implementation is really error prone for users.

trichter: I added a check some lines before (l. 148). That's enough, isn't it?

obspy/signal/cross_correlation.py
trichter: Will do it. It was not clear to me waether `xcorr()` should stay in the long run.

trichter:
@krischer Exactly, I agree. Consequently, the `type` keyword in the `correlate` function is not really of any value. I will remove it again. Regarding the second comment: I implemented a check after @megies comment at line 148.

obspy/signal/cross_correlation.py
megies: Definitly better do it in one place, otherwise this is essentially an if/elif/else/raise-switch split up into two places separated by 20 lines. You never know who will do what to this code in the future so striving for well-organized and easy to understand code is no idle task but something that should be honored.

```
$ python -c 'import this' | grep counts
Readability counts.
```

trichter:
I implemented the review suggestions and rebased on current master. Hopefully, CI tests will pass.

krischer:
Assuming CI passes: Looks good to me now :) Thanks for this!

As this is a pretty big change: Can someone else who actually regularly uses correlations have a second look?

krischer:
I'm also really sorry for the looooong review time...

trichter:
Another rebase, should be good now.

krischer:
I gave this another pass and made some chances.

Given that the old `xcorr()` function still existed but was renamed and the actual `xcorr()` method was a wrapper around the new CC function that almost had the same behavior (non-unique shifting issues) AND both the old method and the new wrapper were marked as deprecated I just restored the `xcorr()` method completely. I also fully restored the original tests - they tested a second wrapper emulating the old xcorr() method (just a left-over from an earlier version of this PR).

I still marked the `xcorr()` method as deprecated so we can eventually remove it (maybe with ObsPy 2.0?).

Anyways - this will now not change existing behavior but just steer people towards using the new `cross_correlation` method.

Good to go from my side if CI passes.

trichter:
@krischer Thanks for having a second look. Now that `xcorr` is deprecated it makes sense to use the old implementation. The "unnecessary test case" could still be used to test the output of `correlate` and `xcorr_max` against the old implementation by replacing
```
shift, corr, corr_fun = xcorr(self.a, self.b, 15, full_xcorr=True)
```
with
```
corr_fun = correlate(self.a, self.b)
shift, corr = xcorr_max(corr_fun)
```.

krischer:
> The "unnecessary test case" could still be used to test the output of correlate and xcorr_max against the old implementation by replacing

Good point and done in the latest two commits :) Are you otherwise okay with merging it?

@megies: Do you have any opinion on this?

trichter:
I am OK with merging. Looks good.

megies:
> Anyways - this will now not change existing behavior but just steer people towards using the new cross_correlation method.

I'm totally fine with this change in this case.. but we really needed this soft transition.

👍

megies:
I'll kick of a docker build on my computer..

krischer:
I guess we've discussed this enough and a couple people had eyes on the code. I'll merge this. Thanks a bunch @trichter :)

PR #1575, opened 2016-11-08
Reference time not written to SAC file when made from scratch

jkmacc-LANL:
(See the lengthy-ish discussion halfway down #1519 by @claudiodsf )

On 0.10.3 you can do:

``` python
import numpy as np
from obspy import Trace, UTCDateTime

tr = Trace(np.zeros(1000))
tr.stats.delta = 0.01
tr.stats.station = 'XXX'
tr.stats.starttime = UTCDateTime()
tr.stats.sac = {}
tr.stats.sac['a'] = 12.34
tr.write('test.sac', format='SAC')
```

and get a sac file with the correct reference time and the "a" header field set.  If you do the same thing in 1.0.2, what you get is a sac file without reference time, unless you do not set the "a" field.

### DIFF
--- a/obspy/io/sac/tests/test_core.py
+++ b/obspy/io/sac/tests/test_core.py
@@ -84,6 +84,10 @@ class CoreTestCase(unittest.TestCase):
             tempfile = tf.name
             tr.write(tempfile, format='SACXY')
             tr1 = read(tempfile)[0]
+
+        tr.stats.pop('sac', None)
+        tr1.stats.pop('sac', None)
+
         self.assertEqual(tr, tr1)
 
     def test_read_write_xy_via_obspy(self):
@@ -622,6 +626,9 @@ class CoreTestCase(unittest.TestCase):
             fh.seek(0, 0)
             st2 = _read_sac_xy(fh)
 
+        st[0].stats.pop('sac', None)
+        st2[0].stats.pop('sac', None)
+
         self.assertEqual(st, st2)
 
     def test_read_xy_write_xy_from_open_file_binary_mode(self):
@@ -635,6 +642,9 @@ class CoreTestCase(unittest.TestCase):
             _write_sac_xy(st, tf)
             tf.seek(0, 0)
             st2 = _read_sac_xy(tf)
+
+        st[0].stats.pop('sac', None)
+        st2[0].stats.pop('sac', None)
 
         self.assertEqual(st, st2)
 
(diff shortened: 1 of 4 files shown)


jkmacc-LANL:
This issue is now a PR:

* A test is added for the case where `stats.sac` has relative time headers but no `nz`-times.  The SAC file will have `nz`-times from the `stats.starttime`.
* Some `Stream`/`Trace` object comparisons in the test suite were changed, so that any `.stats.sac` dictionaries were excluded from the tests.  These `.sac` can change depending on decisions made at the `io/sac/util.py` level, and the tests should mostly just guarantee the `Trace` or `Stream` is unaffected by round-trip write/reads. 
* I don't understand the following warning, and could use some help interpreting it: `/Users/jkmacc/repos/obspy/obspy/io/sac/util.py:331: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal`

Thanks!

jkmacc-LANL:
@claudiodsf I'm curious if this test/branch adequately produces the behavior from 0.10.3...

claudiodsf:
I can test it this weekend. I'm currently at a conference with sparse internet access (and port 22 filtered --so no git!)

jkmacc-LANL:
Sparse internet access?  Where are you, in a castle? :-) Thanks and no worries.

claudiodsf:
Lost in the Caribbeans...

QuLogic:
GitHub provides repos over https so you can get through (many) firewalls. Whether you'd want to is a different question...

obspy/io/sac/tests/test_core.py
```diff
@@ -84,6 +84,12 @@ def test_read_xy_write_xy_via_obspy(self):
             tempfile = tf.name
             tr.write(tempfile, format='SACXY')
             tr1 = read(tempfile)[0]
+
+        sac = tr.stats.pop('sac', None)
+        sac1 = tr1.stats.pop('sac', None)
+        del sac
+        del sac1
```
mbyt: cool, thank you for looking into all this. SAC is used a lot and so it's good if it operates smoothly with obspy.

I do not get the above lines. For me it looks to be the same as:
```python
tr.stats.pop('sac', None)
tr1.stats.pop('sac', None) 
```

What am I missing?

mbyt:
Basically I am tempted to merge this now. From a high level view this pull request does not break any existing test and serves an extra test for this issue --- so it cannot be bad :).

> @claudiodsf I can test it this weekend. I'm currently at a conference with sparse internet access (and port 22 filtered --so no git!)

Alight, let's wait for mid of next week before merging...

obspy/io/sac/tests/test_core.py
jkmacc-LANL: You're right, but the `flake8` Hobgoblin doesn't like the `.pop` syntax, surprisingly :wink:

mbyt: aha, the holy flake8 :). That explains it, thank you.

What about the following?
```
del tr.stats['sac']
# or more explicitly bypassing flake8
tr.stats.pop('sac')  # NOQA
```

megies: @jkmacc-LANL, curious.. what error message do you get by flake8? I don't see it complaining about that code

jkmacc-LANL: @megies First it complained about the semicolon I had at the end (I didn't want to capture the result of the pop), then it complained about the sac and sac1 I captured but didn't use for anything afterwards, so I deleted them:-)

jkmacc-LANL: @mbyt That's a good idea; maybe I'll give it a try!

megies: uhm.. semicolon to not capture output.. sounds like you had a matlabish seizure :stuck_out_tongue_winking_eye: 

just make it..
```python
tr.stats.pop('sac')
```

jkmacc-LANL: @megies I figured `sac` would print to the screen and annoy the test runner.

megies: Nah, that's just what IPython does, print the entered commands return value's string representation (`repr(value)`, or `value._repr_pretty_()` if it's defined -- like we do for e.g. `Stream`).
If executed with python "normally" no return values ever get printed.

jkmacc-LANL: Thanks, fixed.

megies:
If this is ready, I think we should squash this one into one commit and get rid of a couple of dirty merge commits in there.

claudiodsf:
Sorry for no reaction on this: been taken but many things in these days. I can confirm that this works as expected.

mbyt:
@claudiodsf thank you very much for manual testing, I really appreciate that!

jkmacc-LANL:
:+1: !

PR #1557, opened 2016-11-09
Bad error message when adding two UTCDateTime objects

megies:
Adding two UTCDateTime objects makes little sense but the error message is confusing for users and we should change it:

```
In [1]: import obspy

In [2]: obspy.UTCDateTime() + obspy.UTCDateTime()
<ipython-input-3-6a746d0eba8e> in <module>()
----> 1 obspy.UTCDateTime() + obspy.UTCDateTime()

/Users/lion/workspace/code/obspy/obspy/core/utcdatetime.py in __add__(self, value)
    856             value = (value.microseconds + (value.seconds + value.days *
    857                      86400) * 1000000) / 1000000.0
--> 858         return UTCDateTime(self.timestamp + value)
    859
    860     def __sub__(self, value):

TypeError: unsupported operand type(s) for +: 'float' and 'UTCDateTime'
```


mbyt:
thank you! That makes debugging more obvious. I am sure many people dealing with our `UTCDateTime` class.

PR #1589, opened 2016-11-13
Alias kilometer2degrees to kilometers2degrees

jkmacc-LANL:
Can't tell you how many times I've:

```python
In [5]: from obspy.geodetics import degrees2kilometers, kilometers2degrees
ImportError                               Traceback (most recent call last)
<ipython-input-5-c3cd54c63678> in <module>()
----> 1 from obspy.geodetics import degrees2kilometers, kilometers2degrees

ImportError: cannot import name kilometers2degrees
```
This PR just adds `kilometers2degrees = kilometer2degrees` into `obspy/geodetics/base.py`.  Literally, that's all it does.


QuLogic:
Shouldn't have that merge commit in there, but I'm still wondering why the Reftek tests are failing from this completely unrelated change...

megies:
> This PR just adds kilometers2degrees = kilometer2degrees

Hmm.. in general I'm not a fan of too many aliases as it bloats the namespace. But.. I do recognize that the mismatching naming is really unfortunate. So, dunno,

> wondering why the Reftek tests are failing

That test fail is in master, I made two more commits directly in master because the PR was already merged. Some minor Py3 problem, I'll fix it in master.

jkmacc-LANL:
@megies In general, I agree, but I also think that a consistent API perhaps more valuable than the number of things in the namespace.

krischer:
I'm in favor of this - it has also bitten me too many times. Maybe even make the new spelling the default and alias the other one? This will (over time) move people to use the new variant as that is the then documented one.

As we are now 1.0 I guess we should not really break API anymore easily but in one of the next releases we could maybe think about deprecating things slowly? I think we should maybe establish a small document on how we plan on dealing with deprecations and what kind of stability we try to guarantee (its one of the major points of critique people keep bringing up with ObsPy)..

QuLogic:
We should probably just go with SemVer, though that does technically mean we shouldn't remove stuff in 1.1.0.

krischer:
Yea why not - we are also pretty stable now so it should work without any major inconveniences.

I kind of like the way django does it - they basically just document really well (https://docs.djangoproject.com/en/dev/internals/deprecation/) but scientists will get annoyed if their two-year old script will not work anymore so that is probably not feasible.

PR #1568, opened 2016-11-15
Include libmseed test data in ObsPy packages

megies:
e.g. Obspy Debian/Ubuntu packaging does not include libmseed test data that is in use in mseed test cases since merging #1540 (obspy/io/mseed/src/libmseed/test and test/data). This currently leads to some test fails in the docker deb build/testbot, e.g. http://tests.obspy.org/55986/#1


megies:
Confirmed that this works for Debian packaging.

PR #1597, opened 2016-11-22
Raising upon setting non-finite floats.

krischer:
This is a follow up to #1376. It makes sure non finite floats (NaN, inf, -inf) cannot be set as floating point values in all event objects. This should prevent a certain class of error.

Note that this also means ObsPy can no longer read files with infs and NaNs but I think this acceptable. What do people think?

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -91,6 +91,8 @@ master: (doi: 10.5281/zenodo.165135)
      dayplot of a trace, see #1566)
    * properly pass through kwargs from Stream.detrend() to Trace.detrend()
      (see #1607)
+   * Non finite floats (NaN, inf, -inf) can now no longer be set for all
+     event objects (see #1597).
  - obspy.core.event.source:
    * Fix `farfield` if input `points` is a 2D array. (see #1499, #1553)
  - obspy.clients.earthworm:
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -27,6 +27,7 @@ from future.utils import native_str
 import collections
 import copy
 import inspect
+import math
 import re
 import warnings
 import weakref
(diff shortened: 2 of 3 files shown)


megies:
> ObsPy can no longer read files with infs and NaNs

Fair enough to not be able to read invalid files, I would say. On the other hand, maybe it would also be an option to just omit reading the offending piece of information, issue a warning and read at least the rest of the file (e.g. if there's some `nan` in the moment tensor, omit it but read at least the origin/magnitude/etc.)? But that would probably be a lot more work..?

krischer:
Yea that would be a bit tricky - right now it happens as soon as an attribute at any of the event objects is set. That is quite nice as it happens during manual object creation as well as during file reading so it just catches it everywhere with only a minimal amount of work.

megies:
Yeah, that's what I thought. Maybe we could add more information to the error message to help the user locate the problem (e.g. `type(self)` or `str(self)`).

krischer:
That's already happening. One possible error message right now is `"Value 'nan' for 'latitude' is not a finite floating point value."`.

megies:
..then `self` would be the `Origin`, I guess.

nackerley:
I can see that there's no need to support NaN; `None` seems to serve the same purpose very well.  I haven't studied the code, but judging from the behaviour, when a property of an obspy.core.event object is set to None, on writing QuakeML that property is simply omitted. This is good.

I am a fan of Inf in general, but I can't think of a reasonable use case here. I had been setting certain Pick.timing_error QuantityError.uncertainty to Inf; I am now using None instead, with the understanding that is interpreted as "unlimited uncertainty", since 0 clearly means "zero uncertainty".  I would treat OriginUncertainty the same. With a synthetic signal Amplitude.snr could be correctly set to Inf to indicate that it is noiseless. Like I said, these examples pretty extreme. 

When Inf and NaN crop up in calculations, that usually indicates an error which should be dealt with as obspy.core.event or equivalent objects are constructed, so I'm ok with effectively saying they should never be written to QuakeML.

megies:
@krischer, you added this to the 1.0.3 milestone but the PR is against master. Where do you want this to go? (I think master might indeed be the right choice as it changes behavior a bit)

krischer:
Yea - I also think `master` is the correct branch. Let's hope it now also works for Python 2 and I also improved the error message a bit.

megies:
Rebased and force-pushed to see what CI says.

PR #1598, opened 2016-11-23
evalresp: add routine to get response for specific frequencies

megies:
This adds a routine to request response for a set of discrete frequencies using evalresp. I'll also add a routine for magnitude estimation based on a response object now.

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -979,8 +979,8 @@ class Response(ComparingObject):
                     dtype=np.float64)
 
                 # Sanity check.
-                min_f = freqs[freqs > 0].min()
-                max_f = freqs.max()
+                min_f = frequencies[frequencies > 0].min()
+                max_f = frequencies.max()
 
                 min_f_avail = min(f)
                 max_f_avail = max(f)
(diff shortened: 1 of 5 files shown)


megies:
With this PR the current `estimate_magnitude` function that is based on a poles-and-zeros dictionary can also work with response objects. I left the argument name `paz` as is and added to the existing API docstrings, as renaming the argument might break code or take a lot of time to get deprecation reroutes right..

The whole magnitude estimation could use some more/better tests, but at least there are some..

megies:
This is ready for review/comments, all test fails are due to #1599.

obspy/signal/invsim.py
```diff
@@ -273,26 +272,66 @@ def evalresp(t_samp, nfft, filename, date, station='*', channel='*',
         datime = C.create_string_buffer(
             date.format_seed().encode('ascii', 'strict'))
         fn = C.create_string_buffer(tempfile.encode('ascii', 'strict'))
-        nfreqs = C.c_int(freqs.shape[0])
+        frequencies = np.array(frequencies)
```
QuLogic: `np.asarray` again?

obspy/signal/invsim.py
```diff
@@ -637,16 +678,28 @@ def estimate_magnitude(paz, amplitude, timespan, h_dist):
...
+    wood_anderson_amplitude_functions = []
+    for i in paz:
+        if isinstance(i, dict):
+            wood_anderson_amplitude_functions.append(
+                estimate_wood_anderson_amplitude)
+        elif isinstance(i, Response):
+            wood_anderson_amplitude_functions.append(
+                estimate_wood_anderson_amplitude_using_response)
+        else:
+            raise TypeError()
```
QuLogic: Add message.

obspy/core/inventory/response.py
```diff
@@ -1001,25 +995,24 @@ def get_unit_mapping(key):
...
                 _amp_c = (C.c_double * rl.nresp)()
                 _phase_c = (C.c_double * rl.nresp)()
 
-                for i in range(len(freqs)):
-                    _freq_c[i] = freqs[i]
-                    _amp_c[i] = amp[i]
-                    _phase_c[i] = phase[i]
+                _freq_c[:] = frequencies[:]
+                _amp_c[:] = amp[:]
+                _phase_c[:] = phase[:]
```
krischer: Hmm. I cannot believe I did not try this the first time around. Lol - this should also be quite a bit faster now :-)

krischer:
I added a changelog commit directly to the master: 7fd84f9ed2ab0ba8757d28ca5c8ba61dd98794d1

megies:
Just noticed that..

```python
>>> d = AttribDict()
>>> isinstance(d, dict)
False
```

..which is quite unfortunate but I guess adding `dict` as a base class on `AttribDict` might be dangerous.. :-/

claudiodsf:
You can always check that it is dictionary-like:
```python
>>> from collections import Mapping
>>> d = AttribDict()
>>> isinstance(d, Mapping)
True
```

PR #1605, opened 2016-11-29
Socket recv handling

jmfee-usgs:
### What does this PR do?
Check the return value from socket.recv before buffering the (empty) response and looping forever.

I couldn't find any existing issues for this specific problem.

On a heavily loaded Scientific Linux 7 (RHEL flavor) system, we run many concurrent processes using the earthworm client to access data.  Intermittently (<1%, but every few hours) the process would spike CPU and leak memory until the kernel killed the process because the system ran out of memory.  Logs indicated this was while fetching data using the earthworm obspy client.  Using a copy of the earthworm client with these modifications has thus far eliminated this leak.

From https://docs.python.org/2/howto/sockets.html#using-a-socket
> When a recv returns 0 bytes, it means the other side has closed (or is in the process of closing) the connection. You will not receive any more data on this connection. Ever.

### PR Checklist
- [x] All tests still pass.
This is not reproducible under normal conditions.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,8 +5,12 @@
      dayplot of a trace, see #1566)
  - obspy.core.event.source:
    * Fix `farfield` if input `points` is a 2D array. (see #1499, #1553)
+ - obspy.clients.earthworm:
+   * Better end of stream detection. (see #1605)
  - obspy.clients.neic:
    * Better end of stream detection. (see #1563)
+ - obspy.clients.seedlink:
+   * Better end of stream detection. (see #1605)
  - obspy.io.mseed:
    * ObsPy can now also read (Mini)SEED files with noise records. (see #1495)
    * ObsPy can now read records with a data-offset of zero. (see #1509, #1525)
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -15,6 +15,7 @@ Fabian Engels
 Sven Egdorf
 Laura Ermert
 Tom Eulenfeld
+Jeremy Fee
 Marc Grunberg
 Conny Hammer
 Sebastian Heimann
--- a/obspy/clients/earthworm/waveserver.py
+++ b/obspy/clients/earthworm/waveserver.py
@@ -199,6 +199,8 @@ def get_sock_char_line(sock, timeout=10.):
             # see https://github.com/obspy/obspy/issues/383
             # indat = sock.recv(8192)
             indat = sock.recv(1)
+            if not indat:
+                break
             chunks.append(indat)
     except socket.timeout:
         print('socket timeout in get_sock_char_line()', file=sys.stderr)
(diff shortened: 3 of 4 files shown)


QuLogic:
We need to decide what to do when the socket is closed but < the expected number of bytes is received.

megies:
I don't have much input for this, just wanted to note that there's two more places in our code base with a `socket.recv`:
 - clients/neic/client.py line 204
 - clients/seedlink/client/seedlinkconnection.py line 1202

Also, this might even be considered a bug fix? (Which would qualify it for `maintenance_1.0.x` and thus 1.0.3?)

jmfee-usgs:
@megies: I checked those clients (neic, seedlinkconnection) but it appeared they are using nonblocking sockets, which should generate an exception instead of fail silently.

When I put in the pr, I checked but didn't find the maintenance branch.  Let me know if you need me to update the pr.

megies:
> When I put in the pr, I checked but didn't find the maintenance branch. Let me know if you need me to update the pr.

OK. Also, there should be a tick box for you on this page to the right `"Allow edits from maintainers. "`. If you tick it we should be able to push to your branch if necessary.

> @megies: I checked those clients (neic, seedlinkconnection) but it appeared they are using nonblocking sockets, which should generate an exception instead of fail silently.

Alright. Like I said, I'm no expert on sockets so I'll leave this to @QuLogic and @krischer and whoever feels qualified to review this.

krischer:
Looks good to me and I don't think the change can have any negative side effects! Thanks a lot!

I guess the memory leak happened because it kept on looping and adding empty things to the `chunks` list?

Can you please make your changes to the `maintenance_1.0.x` branch and force push? Let us know if you need help with this.


> We need to decide what to do when the socket is closed but < the expected number of bytes is received.

It will probably just raise some kind of parsing error later down the line which is fine IMHO.

jmfee-usgs:
Rebasing ended up being a bad idea... Changes updated to start from maintenance_1.0.x.

krischer:
Yea rebasing does not work here. You could have cherry picked the commits on top of a branch off of maintenance. But whatever you did - looks good now :-)

krischer:
Can be merged depending on what CI says.

krischer:
@megies: Are the seedlink and earthworm tests supposed to be executed everywhere? I don't see them anywhere.

megies:
@krischer, no, just Travis picks that up currently (see https://github.com/obspy/obspy/pull/1408#issuecomment-249580150). My Windows-fu wasn't up to adding it to Appveyor too, and the docker scripts are pretty messy already, so I didn't want to add more mess on top.

Now, don't ask me why Travis didn't run this PR.. (maybe it got confused by the rebase)

krischer:
I see. I'm merging in any case - I'm fairly sure this works based on the changes + the socket module documentation. Additionally I tested it locally for Python 2 and 3.

@jmfee-usgs: Thanks for this!

megies:
I ran this PR through our docker images, looking good. :+1: http://tests.obspy.org/?git=701bdb68fd&node=docker-

Issue #1599, opened 2016-11-23
failing requests to examples.obspy.org


megies:
Especially in CI we have plenty of non-successful requests to examples.obspy.org somehow, which leads to lots of errors in CI tests..

E.g.: http://tests.obspy.org/60454/#1

```
```Traceback (most recent call last):
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/obspy/core/tests/test_stream.py", line 1720, in test_read
tr = read('https://examples.obspy.org/test.sac', dtype=np.int32)[0]
File "<decorator-gen-31>", line 2, in read
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/obspy/core/util/decorator.py", line 294, in _map_example_filename
return func(*args, **kwargs)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/obspy/core/stream.py", line 225, in read
download_to_file(url=pathname_or_url, filename_or_buffer=fh)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/obspy/core/util/base.py", line 602, in download_to_file
r = requests.get(url, stream=True)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/api.py", line 70, in get
return request('get', url, params=params, **kwargs)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/api.py", line 56, in request
return session.request(method=method, url=url, **kwargs)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/sessions.py", line 369, in __exit__
self.close()
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/sessions.py", line 690, in close
v.close()
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/adapters.py", line 306, in close
self.poolmanager.clear()
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/packages/urllib3/poolmanager.py", line 156, in clear
self.pools.clear()
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/packages/urllib3/_collections.py", line 94, in clear
self.dispose_func(value)
File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/requests/packages/urllib3/poolmanager.py", line 117, in <lambda>
[...]

PR #1606, opened 2016-11-30
Mock out http requests in unit tests (module core), skip http requests in doctests (modules core/signal)

megies:
This PR makes the complete test suites of `DEFAULT_MODULES` network-independent by..

 - mocking some requests to `https://examples.obspy.org` in unit tests of `core` (returning the expected data and otherwise leaving the tests unchanged)
 - skipping some http-based doctests in core (these routines are extensively tested in unit tests, see above bullet point)
 - skipping one http request in `signal.tf_misfits` doctests (the downloaded data is not used anyway, as the plot is skipped)

This is a workaround for currently failing http requests to `https://examples.obspy.org` as described in #1599, which make all CI fail right now, which makes it extremely cumbersome to review currently open PRs.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -153,8 +153,9 @@ def read(pathname_or_url=None, format=None, headonly=False, starttime=None,
     (3) Reading a remote file via HTTP protocol.
 
         >>> from obspy import read
-        >>> st = read("https://examples.obspy.org/loc_RJOB20050831023349.z")
-        >>> print(st)  # doctest: +ELLIPSIS
+        >>> st = read(  # doctest: +SKIP
+        ...     "https://examples.obspy.org/loc_RJOB20050831023349.z")
+        >>> print(st)  # doctest: +SKIP
         1 Trace(s) in Stream:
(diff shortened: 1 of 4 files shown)


megies:
I'll merge this once CI turns green, so if there's any objections.. bring them up soon.

megies:
See #1599: requests 2.12 is responsible for these problems

Probably we should still merge this one so that default tests are *really* network independent? On the other hand we would not have become aware of the problem in #1599 in that case..

krischer:
You could also add an explicit network test right next to the `Stream` test you replaced. If that fails it at least becomes very clear that it is a network related issue. I feel like we don't want to completely give up on the integration test between ObsPy and the requests library.

megies:
The problem behind this should be worked around with #1608, so I'm bumping this as it is low prio and we want to get the release done.

barsch:
I suggest we drop this PR in favor of https://github.com/obspy/obspy/pull/1663, or?

QuLogic:
Well, 1.0.3 is already out, so this doesn't really need to go into 1.0.x. I'm in favour of sticking with #1663 then.

krischer:
Sounds good to me.

PR #1607, opened 2016-11-30
Fix Stream.detrend kwargs

megies:
Properly pass through kwargs from `Stream.detrend` to `Trace.detrend`.


krischer:
I added a small regression test. Looks good otherwise.

megies:
Only unrelated CI fails, so merging.

PR #1611, opened 2016-12-05
Convert distance to deg

calum-chamberlain:
Previously event.origins[x].arrivals[y].distance was incorrectly not converted from km to degrees, this commit fixes that.

### DIFF
--- a/obspy/io/nordic/tests/test_nordic.py
+++ b/obspy/io/nordic/tests/test_nordic.py
@@ -465,12 +465,12 @@ class TestNordicMethods(unittest.TestCase):
                                 key=lambda x: x.distance)[0].distance,
                          0.035972864236749225)
         pick_strings = nordpick(event)
-        self.assertEqual(int([p for p in pick_strings
-                              if p.split()[0] == 'GCSZ'
-                              and p.split()[1] == 'SZ'][0].split()[-1]), 304)
-        self.assertEqual(int([p for p in pick_strings
-                              if p.split()[0] == 'WZ11'
-                              and p.split()[1] == 'HZ'][0].split()[-1]), 30)
+        self.assertEqual(
+            int([p for p in pick_strings if p.split()[0] == 'GCSZ' and
+                 p.split()[1] == 'SZ'][0].split()[-1]), 304)
+        self.assertEqual(
+            int([p for p in pick_strings if p.split()[0] == 'WZ11' and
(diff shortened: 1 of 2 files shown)


QuLogic:
I don't know what's up with the build at the moment, but I'd have thought this might break some test, no? Otherwise, maybe a test should be added.

calum-chamberlain:
You are almost certainly right @QuLogic possibly premature PR - need to convert the other way too, will look again tomorrow and run some tests.

calum-chamberlain:
So I added the reverse option and tested - it did fail the previous tests, but I amended the test.  I can't immediately think of a good way to test whether the units are correct when read in/written out - any ideas?

krischer:
Can you not just manually calculate the distances for a couple of tests files and check if they are in the correct ballpark?

calum-chamberlain:
I think this is done, not sure what the issues with appveyor are, they don't appear to be related to this?

QuLogic:
The former is unrelated, the latter is fixed by #1614, I think.

obspy/io/nordic/tests/test_nordic.py
```diff
@@ -456,8 +456,24 @@ def test_read_picks_across_day_end(self):
             self.assertTrue(pick.time in pick_times)
         self.assertEqual(event_2.origins[0].time, event.origins[0].time)
 
-
-def test_similarity(event_1, event_2):
+    def test_distance_conversion(self):
+        """Check that distances are converted properly."""
```
krischer: Can you please use 
```
def ...():
    """
    TEXT
    """
    ...
```

as a docstring style to be consistent with the rest of ObsPy.

obspy/io/nordic/tests/test_nordic.py
```diff
@@ -456,8 +456,24 @@ def test_read_picks_across_day_end(self):
...
+        testing_path = os.path.join(self.testing_path, '01-0411-15L.S201309')
+        cat = read_events(testing_path)
+        event = cat[0]
+        self.assertEqual(sorted(event.origins[0].arrivals,
+                                key=lambda x: x.distance)[0].distance,
+                         0.035972864236749225)
```
krischer: This should probably be an "assertAlmostEqual" just to deal with floating point inaccuracies. Tests seem to pass fine but it cannot hurt to be safe.

krischer:
Thanks for this!

PR #1608, opened 2016-12-02
Show a warning for certain 'requests' versions

megies:
Show a warning for certain 'requests' versions and fix version in travis and appveyor for now.

See #1599.

PR #1612, opened 2016-12-05
Ubuntu xenial+yakkety: problematic requests version

megies:
Ubuntu xenial and yakkety have a `python-requests` version that obspy currently does not work with, see #1608 and #1599. The problem seems to be monkey patching done by `future` which causes problems for `Queue` module when imported via `six.moves`.

 - http://tests.obspy.org/61797/
 - http://tests.obspy.org/61804/

Either we..
 - find a way around that problem inside obspy (not very likely, or at least will be extremely ugly, probably)
 - package and distribute a newer requests version via deb.obspy.org for those two Ubuntu releases
 - ....?

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -21,17 +21,22 @@ import gzip
 import io
 import os
 import re
+import sys
 from socket import timeout as socket_timeout
 import textwrap
 import threading
 import warnings
 
 with standard_library.hooks():
-    import queue
     import urllib.error
     import urllib.parse
     import urllib.request
     from collections import OrderedDict
+
+if sys.version_info.major == 2:
+    import Queue as queue
+else:
+    import queue
 
 from lxml import etree
 


krischer:
Oh nice :-) Does this workaround work everywhere?

megies:
> If you package a new requests version for ObsPy does this mean that this requests version is only active for ObsPy or for all of ubuntu?

That would affect the whole system Python..

But, I think I found an easy solution.. looks like that by not relying on `future` to import `queue` from it's standard library magic makes the problem go away..

krischer:
> But, I think I found an easy solution.. looks like that by not relying on future to import queue from it's standard library magic makes the problem go away..

That'll definitely do the trick for the next release :)

megies:
> That'll definitely do the trick for the next release :)

Yeah, looks like that's an easy way out.
Running docker tests right now, and we'll see what the docker deb buildbot has to say about it.. :-)

megies:
Eeeehm. The only check that says something about this PR is the deb buildbot that runs yakkety and xenial.. and it hasn't run yet..

krischer:
My bad - sorry. I got a bit confused with all the open PRs and issues...

megies:
Ok, no worries. Docker tests I ran locally just finished and they look good: http://tests.obspy.org/?git=48162a8&node=docker-

Xenial and Yakkety:
 - http://tests.obspy.org/62105/
 - http://tests.obspy.org/62107
 - http://tests.obspy.org/62110
 - http://tests.obspy.org/62112

Issue #1254, opened 2016-02-16
Use requests instead of urllib


krischer:
The combination of using `future` and `urllib` on older systems results in downloads not working for HTTPS urls. This bug could and should probably be fixed in the `future` library but we already have `requests` as a hard dependency so we should just use that. See e.g. here:

http://tests.obspy.org/35446/
#1253 already does that for everything except `obspy.clients`. We should also do it there.

megies:
One thing to consider.. from requests docs: `If you're using GET, OPTIONS, POST, PUT, PATCH or DELETE, you can disable redirection handling with the allow_redirects parameter:` Maybe we'll have to manually disallow redirects in most cases that use credentials.

megies:
@krischer, @mwschouten I did some work on this at some point but then didn't have time to work on it more.. if still needed, I pushed it to https://github.com/obspy/obspy/tree/use_requests now..

megies:
For future reference, there's some API changes in `requests` that we probably will have to catch in a compatibility layer, see e.g. <<@krischer please fill in reference to code you showed me today about `requests.Response.json`>>

krischer:
Here you go:

https://github.com/obspy/obspy/blob/4c90ea2cb3cb1be88403161109eda2600d385001/obspy/clients/syngine/client.py#L31-L54

https://github.com/obspy/obspy/blob/4c90ea2cb3cb1be88403161109eda2600d385001/obspy/clients/base.py#L221-L225

megies:
This hasn't seen any activity, so I'll bump this to 1.2.0.

d-chambers:
Any chance we can move this to 2.0.0? May be easier to do it during the `future`-ectomy than on its own?

krischer:
Yes I was thinking the same.

krischer:
I think 1.3.0 is the release we'll now drop Python 2. Or do we want to call this release 2.0? We'll not release another Python 2 supported release after 1.2.0

d-chambers:
To me it feels like a major version bump is in order for dropping python 2, but it isn't that important either way. The important thing is that we are dropping python 2 :smile:

megies:
Yep, we already decided to make 2.0.0 the next release after 1.2.0, it should be discussed in some issue or maybe it was just on gitter dunno

trichter:
I think this issue is obsolete.

PR #1613, opened 2016-12-05
WIP: Back to the past

krischer:
This is a branch of mine that removes the `future` dependency. It is just a bit too invasive and a big liability for us, especially as it does not appear to be updated a lot anymore so we really have to remove it.

I likely won't have time to finish this until after christmas but due to recent issues I think its a good idea to push this and maybe somebody else feels like taking over.

This PR mainly does two things:

* Remove future imports.
* Removes the `unicode_literals` imports from `__future__`. There are various reasons why this is not a good idea so now its mostly gone.

What works last time I tested it:

* Most of `core`, `io`, and `taup`.

Most clients don't yet work and I think the best way forward it to just make all of them depend on the `requests` library which will take care of Py2/3 compatibility for us.

### DIFF
--- a/obspy/__init__.py
+++ b/obspy/__init__.py
@@ -27,10 +27,7 @@ for seismology.
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
-from future.builtins import *  # NOQA
-from future.utils import PY2, native_str
+from __future__ import absolute_import, division, print_function
 
 import warnings
 import requests
@@ -39,6 +36,7 @@ import requests
 from obspy.core.utcdatetime import UTCDateTime  # NOQA
 from obspy.core.util import _get_version_string
 __version__ = _get_version_string(abbrev=10)
+from obspy.core.compatibility import PY2
 from obspy.core.trace import Trace  # NOQA
 from obspy.core.stream import Stream, read
 from obspy.core.event import read_events, Catalog
@@ -50,7 +48,6 @@ from obspy.core.util.obspy_types import (
 __all__ = ["UTCDateTime", "Trace", "__version__", "Stream", "read",
            "read_events", "Catalog", "read_inventory", "ObsPyException",
            "ObsPyReadingError"]
-__all__ = [native_str(i) for i in __all__]
 
 
 # insert supported read/write format plugin lists dynamically in docstrings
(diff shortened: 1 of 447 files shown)


krischer:
If somebody feels like taking over: its not particularly hard to do but a bit of a grind.

krischer:
Before this can proceed I think we should first replace all uses of `urllib` with `requests` - this would make this a lot easier.

megies:
That would be #1254 which also already has a branch with some initial work on that topic..

d-chambers:
I am looking over this and wondering if it might be worth limping along with the future package until we drop python 2 support (perhaps Jan 1st 2020 if we follow suit with numpy, see #1764)? 

The future package, as well as imports from `__future__`, aim to make python 2 behave like python 3. If we only support python 3, in the majority of the cases I suspect we will be able to simply delete these two import statements and be done with it.

krischer:
> I am looking over this and wondering if it might be worth limping along with the future package until we drop python 2 support (perhaps Jan 1st 2020 if we follow suit with numpy, see #1764)?

That would be an option but I'm not sure if its feasible. The `future` package to my understanding is basically no longer maintained so I fear it might just not work with Python 3.7. In any case - I think this would just require a concentrated effort of a few days which none of us has easily available but we would manage if really required.

QuLogic:
So `future` seems to have a new maintainer now; maybe we should just hold off on this one until we're at the point of dropping Python 2 entirely.

QuLogic:
So do you want to rebase this now? Or maybe just restart, with so many conflicting files...

d-chambers:
I suspect restarting would be the path of least pain.

obspy/__init__.py
```diff
@@ -27,10 +27,7 @@
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
-from future.builtins import *  # NOQA
-from future.utils import PY2, native_str
+from __future__ import absolute_import, division, print_function
```
megies: Do we need any `__future__` imports at all? Seems obsolete?

https://docs.python.org/3.5/reference/simple_stmts.html#future
```
The features recognized by Python 3.0 are absolute_import, division, generators,
unicode_literals, print_function, nested_scopes and with_statement. They are all
redundant because they are always enabled, and only kept for backwards compatibility.
```

krischer:
I'd definitely start over and scrap this PR. All the compatibility things are no longer required once Python 2 is dropped which greatly simplifies the whole thing.

megies:
I'll make a base PR shortly as a starting point

PR #1616, opened 2016-12-07
[WIP] Matplotlib 2.0 compatibility

QuLogic:
Now that matplotlib 2.0.0rc1 is out, I've tested it with ObsPy. Mostly the differences are in text location and small shift in the legend location. There are a couple of larger changes that I am still investigating. I think most of them are not a big deal though. Here is a histogram of RMS/tolerance for the plots:
figure_1
most are <10 times, but some are way off. Not sure why that is yet.

For now, I'm just pushing this to make sure I haven't broken older Matplotlib, but I'll add proper compatibility stuff after. Should I update the figures for 2.0 or just up the tolerance?

Tentatively 1.1.0, but we can bump it if it's holding things up; I _think_ it'll be small enough to go into 1.1.1 if it has to (and we aren't replacing all the images.)


krischer:
This looks pretty good. Should we try to get this into 1.0.3 to make it matplotlib 2 testable?

There appear to be some issues with the `default` style on matplotlib 1.4: http://tests.obspy.org/63097/

QuLogic:
Turns out there is no default stylesheet in 1.4, so by default, it's just not going to set one. There's matplotlib/matplotlib#7587 to work out and then we'll need to regenerate images or increase tolerance for 2.0 and this'll be done.

QuLogic:
One more change to `ObspyAutoDateFormatter` because Matplotlib 2.0 adds more entries to the lookup with full precision. There are still a few differences, but I'm not sure how they're triggered. Maybe @megies knows more about this.

megies:
> There are still a few differences with date formatting, but I'm not sure how they're triggered. Maybe @megies knows more about this.

@QuLogic, are you're referring to e.g. http://tests.obspy.org/63601/#1? The tick formatter function gets passed the index of the tick by matplotlib (starting with 0 for the left-most x tick), but I assume you've seen that. But.. IIRC matplotlib sadly is not very exact in figuring out what the left-most tick really is, and I think I remember that sometimes matplotlib passes on a non-zero value for the left-most tick (wrongly assuming that there is another tick left of it shown in the figure, which isn't). No idea if that detection routine or the API changed for 2.0.0..

(@barsch any changes to the test report webpage.. somehow the image comparison with the slider isn't there anymore..)

barsch:
no changes from my side - will look into tomorrow ...

QuLogic:
@megies Oddly enough, after a refresh, the images do appear for me (sometimes).

@barsch Also, the images show up over http, but not https; Firefox doesn't like the mixed content because the imgur links are not secure.

megies:
Indeed, sometimes images show up after a couple of refreshes..

barsch:
@megies: mixing http/https was the issue (thx @QuLogic) - should be fixed now

megies:
New Debian 9 "stretch" was frozen some days ago.. and it's already shipping matplotlib 2.0.0.. O_0

Really surprised that they included it so fast and so shortly before the freeze..

QuLogic:
Yes, Debian was ready to go with Matplotlib 2.0 since the release candidates.

This is currently mostly working except for the change in text location. I would like to see what effect matplotlib/matplotlib#7970 has before regenerating all the images.

Additionally, there's a change in the Cartopy images because it's not fully Matplotlib 2.0 compatible. See SciTools/cartopy#773 for that.

krischer:
https://github.com/matplotlib/matplotlib/pull/8081 seems to be merged know and it should help a lot.

@QuLogic Any ETA on the release of 2.0.1?

QuLogic:
Possibly this weekend, but actually, I'm not sure that fix will save us from regenerating quite a few images (or upping tolerances.) The trouble is the alignment is the same as before but there's so much other text that's also moved. I think part of the reason is because our images were generated in such older mpl.

QuLogic:
Oh, I just realized I've been ignoring the `tol*=17` for 1.5.x line; if I extend that to 2.0, I go from 81 failures to just 1, even sticking with 2.0.0 (which doesn't have the baseline change.) Most of the changes are the correction of the text baseline that occurred in 1.5.x.

When do we want to regen the images? When do we want to drop support for older mpl?

megies:
> When do we want to regen the images? When do we want to drop support for older mpl?

We should be pragmatic, so we do what is less work for us. I haven't been able to keep up with this PR about mpl 2 compatibility but if it makes live easier for us to regenerate images now, I'm totally for it.

If at least image size is the same as with current baseline images we can regenerate with 2.0 and use high tolerances for testruns with mpl 1.x.

krischer:
> If at least image size is the same as with current baseline images we can regenerate with 2.0 and use high tolerances for testruns with mpl 1.x.

👍 On this. Let's regenerate them once, once 2.0.1 is out and up the tolerances for the 1.x line. We cannot easily drop support for old matplotlib versions if we want to keep our policy of supporting the packaged versions of all major linux distributions.

megies:
> This is currently mostly working except for the change in text location. I would like to see what effect matplotlib/matplotlib#7970 has before regenerating all the images.

@QuLogic, seems it's merged: matplotlib/matplotlib#8081


@QuLogic since 2.0.1 doesn't seem to have a fixed schedule (?), how about sticking with old baseline images for 1.1.0 and upping tolerances for mpl >=2.0 right now?

krischer:
I regenerated all test images with today's tip of matplotlib's `v2.0.x` branch (the master branch had some weird stuff going on and didn't work for all plots). I think this is a pretty future proof way of doing things. Locally it works for mpl 1.5 and 2.0. Note that the tolerances for 2.0.0 are quite high due to the aforementioned issues with the tick placement. For the upcoming 2.0.1 they will be much lower.

It kind of sucks that we have to eat the additional file size of the new test images but I think this is a sane way of doing it - and we really need mpl 2.0 support - this gets a bit annoying in daily use.

There is a weird test failure for `obspy/signal/tests/test_spectral_estimation.py - test_ppsd_restricted_stacks` on matplotlib 1.5.3. @megies Can you have a look?

@QuLogic Is this a sensible thing to do? Does mpl expect a lot of changes to the default style?

Don't merge yet - needs some squashing!

megies:
> and we really need mpl 2.0 support

totally 👍

megies:
Amended the last commit and force-pushed to trigger Travis.

krischer:
> There is a weird test failure for obspy/signal/tests/test_spectral_estimation.py - test_ppsd_restricted_stacks on matplotlib 1.5.3. @megies Can you have a look?

Did you see this comment? It might have gotten overlooked in my long comment.

megies:
Saw it, will have a look in a bit, working on vcr right now..

QuLogic:
Thanks for handling the tail end of this @krischer. I haven't looked at the results yet, but the idea at least sounds good to me. There might be only one more small change in mpl 2.0.1 that might have an effect, but I don't think we have any plots that might trigger it.

krischer:
> but I don't think we have any plots that might trigger it.

Good to know :) 

This is pretty much good to go as soon as I can get travis to work and @megies fixes the one issue (sorry - i really cannot seem to wrap my head around that test).

I disabled a lot of the pinning we currently do at travis as its a bit ridiculous to figure out which packages are available for which python version and architecture...I hope this will not surface new issues but I guess travis should always more or less test against the latest versions anyways.

krischer:
Unpinning travis seems to do the trick :)

QuLogic:
I assume this will need a rebase to squash some of those duplicate image regenerations.

It's a bit unfortunate about the change in gridlines in the Cartopy image; that's a bug with Cartopy against mpl2 but fixing it means new images for both mpl 1 and 2...

krischer:
> I assume this will need a rebase to squash some of those duplicate image regenerations.

Yea totally - the current plan is to just squash everything. If I can find the motivation I'll do a more careful rebase - let's see.

> It's a bit unfortunate about the change in gridlines in the Cartopy image; that's a bug with Cartopy against mpl2 but fixing it means new images for both mpl 1 and 2...

You mean the `obspy/core/tests/images/catalog-cartopy3.png` image, right? I was wondering what was going on in that image. It would of course be preferable to get it right the first time but if its only a single image I see no reason to delay this PR. I assume the bug is not yet fixed in cartopy's master?

krischer:
Now also works like a charm with Python 3.6  and thus supersedes #1630.

megies:
@krischer you accidently updated `ppsd_restricted_stack.png` with a wrong image:
https://github.com/obspy/obspy/pull/1616/files#diff-4e223ca23de12f4a8a041ad1d9b8e2ae

screenshot from 2017-04-02 16 37 28

What that test is doing is..
 - first plotting the baseline image,
 - then changing the histogram stack and reusing the existing figure instance and plotting again (which should look different and raise an image comparison exception -- that's the image you added as a baseline now) and
 - at the end setting the original histogram stack and again reusing the existing figure instance and plot again (which again should pass as it's the same as the original baseline image plot).

krischer:
Appveyor fails because I accidentially pushed to the obspy fork again...sorry about that...

krischer:
Rebased on latest master and to remove the duplicate test images generations. Now passes on all travis runners and appveyor :) IMHO ready to merge once the docker test bots are done with it :)

.travis.yml
```diff
@@ -125,10 +136,9 @@ install:
         sqlalchemy
         mock
         nose
-        gdal
```
megies: gdal is used in io.shapefile tests.. i suppose it was making problems with Py 3.6? maybe we can install it in at least a few Travis runs?

krischer: I opened a separate issue to re-enable it (https://github.com/obspy/obspy/issues/1749) as I'd really like to get this PR merged.

krischer:
Works for everything :) Someone please review and merge!

PR #1615, opened 2016-12-07
Can't get interactive plots in Jupyter Notebook

krischer:
Hello,

If I run the following code in a Jupyter Notebook Python3 cell I get a static image plot instead of the more useful interactive plot.

```python
%matplotlib notebook

import numpy as np

import obspy
st = obspy.read()
st.plot()
```

If in a following cell I execute

```python
t = np.arange(0.0, 2.0, 0.01)
s = np.sin(2*np.pi*t)
plt.plot(t, s)
```

I get the interactive plot, as expected.
Is there a problem with ObsPy?

I'm using obspy-python3 compiled from a reasonably up-to-date git snapshot on a Debian system.

python3-obspy 1.0.2.post0+794.ge647893002
jupyter-notebook 4.2.3
python3-jupyter-client 4.4.0
python3-jupyter-core 4.2.0
python3 3.5.2

Thanks!

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -135,6 +135,8 @@
    * The functionality behind the `obspy-scan` command line script has been
      refactored into a `Scanner` class so that it can be reused in custom
      workflows. (see #1444)
+   * Almost always return the figure handle for the plots to make it work with
+     the jupyter notebooks (see #1615).
  - obspy.imaging.cm:
    * new colormap: viridis_white. This is a modification of viridis that
      goes to white instead of yellow but remains perceptually uniform. It
--- a/obspy/imaging/__init__.py
+++ b/obspy/imaging/__init__.py
@@ -32,9 +32,10 @@ Examples files may be retrieved via https://examples.obspy.org.
 BW.RJOB..EHZ | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
 BW.RJOB..EHN | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
 BW.RJOB..EHE | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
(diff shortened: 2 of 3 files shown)


krischer:
This issue is now a PR that causes the `plot()` method to almost always return the figure handle. The only downside as far as I can see is that plots might now appear twice in the notebook it its the last command in a cell. I'm not sure there is anything we can do against that.

legovini:
Hi @krischer. Can't we check if `matplotlib.get_backend()` is `nbAgg`, and return the handle only in this case? On `nbagg`:

http://matplotlib.org/users/whats_new.html#the-nbagg-backend

krischer:
Unfortunately not - the jupyter notebooks would then only show the plot if `st.plot()` is the last statement in a cell - they always show (a potentially pretty) representation of the last return value if any.

legovini:
I see. The funny thing is that calling `st.plot().show()` does not result in a double plot. Is there way to disable the `.show()` auto-call? I'd have to call it explicitly every time, but the behavior would always be consistent, not depending on python/ipython/jupyer or on the position of the `.plot()` call in the cell.

QuLogic:
> The funny thing is that calling `st.plot().show()` does not result in a double plot.

I assume you mean `st.plot(handle=True).show()`? And yes, that wouldn't plot double because `.show()` uses the figure and does not return the figure again.

> Is there way to disable the .show() auto-call?

Passing `handle=True` already disables calling `show()`; these are unrelated.

megies:
Rebased on current master and force-pushed for fresh CI.

megies:
Rebased on current master again and force-pushed for fresh CI.

megies:
Punting to 1.1.1.. otherwise we'll never even get to 1.1.0

megies:

megies:
Rebased + force pushed

megies:
Fixed the one doc test failure, rebased, force-pushed

PR #1618, opened 2016-12-07
fix some network tests

megies:
+TESTS:ALL

### DIFF
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -338,8 +338,7 @@ class ClientTestCase(unittest.TestCase):
         Tests the parsing of the available event catalogs.
         """
         self.assertEqual(set(self.client.services["available_event_catalogs"]),
-                         set(("ANF", "GCMT", "TEST", "ISC", "UofW",
-                              "NEIC PDE")))
+                         set(("GCMT", "ISC", "NEIC PDE")))
 
     def test_iris_event_contributors_availability(self):
         """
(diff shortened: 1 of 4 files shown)


obspy/clients/fdsn/tests/test_client.py
```diff
@@ -335,19 +335,46 @@ def test_iris_event_catalog_availability(self):
...
+            'J', 'OGSO', 'DNK', 'us_qui', 'R', 'inmg', 'V', 'us_ldg', 'buc',
+            'SJA', 'CNRM', 'WAR', 'PDA', 'isk', 'ZAMG', 'us_hlw', 'casc', 'mb',
+            'CSEM', 'car', 'uw', 'INDR', 'BJI', 'SPE', 'SNET', 'HLW', 'bgrd',
+            'BRK', 'QUI', 'EXP', 'NSSP', 'SE', 'PAS', 'OBN', 'tap', 'CAR',
+            'ak', 'official', 'AEI', 'SNM', 'US', 'CERI', 'iscgem', 'us_thr',
+            'us_gcg', 'RSNC', 'nm', 'nn', 'SOF', 'csem', 'lim', 'nc', 'tir',
+            'PDG', 'us_the', 'PPT', 'GOM', 'E', 'MACR', 'AG', 'TEIC', 'I',
+            'hv', 'M', 'LD', 'THR', 'SLM', 'rey', 'U', 'DHMR', 'STR', 'us_tul',
+            'WEL', 'SLC', 'us_guc', 'DJA', 'BHL', 'us_aust', 'RSPR', 'thr',
+            'T', 'GFU', 'SDD', 'dja', 'the', 'us_trn', 'OTT', 'se', 'ARO'}
```
krischer: This will be super fragile and I suspect this will keep changing. How about to just test if the `self.client.services["..."]` contains anything at all? And then maybe augment it with a mock test that does the actual parsing?

megies: Sure, go ahead ;-P

PR #1622, opened 2016-12-14
Correctly writing zero station magnitude contributions in QuakeML

### LINKED ISSUE #1621: Station magnitude residual of zero not read from QuakeML
Using ObsPy 1.02 and Python 3.5 installed via Anaconda on Windows.

Try the following  with [test.zip](https://github.com/obspy/obspy/files/646936/test.zip) unzipped to the working directory:

```python
from obspy import read_events
catalog = read_events('test.xml')
print(catalog.events[0].magnitudes[0].station_magnitude_contributions[2].residual)
catalog.write('test2.xml', 'QuakeML')
```

The problem is that station magnitude residual of 0.0 is read and printed, but not written to test2.xml.
It seems that a residual of 0.0 in the third StationMagnitudeContribution is explicitly ignored on writing; you can change the residual to another value and it will write fine.

I expect a residual of 0.0 to be just as valid as any other value. I would expect a value of None to indicate that the residual is not known or not recorded.

nackerley:
Fixed #1621 by bringing Pickler._station_magnitude_contributions() into line with other Pickler methods, such as Pickler._arrival(), in particular by making use of Pickler._str(). This means that residual and weight are written even if they are both zero, and not written only when they are `None`.

I suspect that Pickler._station_magnitude_contributions() was missed when other methods were updated a while back to use Pickler._str() mainly because it is out-of-order with the rest which do use Pickler._str(). A more natural place for it among the methods would be between Pickler._station_magnitude() and Pickler._origin(), but that would have made the diff less clear.

Also fixed a bug whereby Event.region was neither written to QuakeML nor read.

Updated regression tests to reflect these changes, and added myself to contributors, even this is a very minor contribution.

### DIFF
--- a/obspy/io/quakeml/core.py
+++ b/obspy/io/quakeml/core.py
@@ -1157,7 +1157,7 @@ class Pickler(object):
     def _station_magnitude_contributions(self, stat_contrib, element):
         for contrib in stat_contrib:
             contrib_el = etree.Element('stationMagnitudeContribution')
-            self._str(contrib.station_magnitude_id.id, contrib_el, 
+            self._str(contrib.station_magnitude_id.id, contrib_el,
                       'stationMagnitudeID')
             self._str(contrib.weight, contrib_el, 'weight')
             self._str(contrib.residual, contrib_el, 'residual')
--- a/obspy/io/quakeml/core.py
+++ b/obspy/io/quakeml/core.py
@@ -561,6 +561,7 @@ class Unpickler(object):
         obj.composite_times = self._composite_times(element)
         obj.quality = self._origin_quality(element)
         obj.origin_type = self._xpath2obj('type', element)
+        obj.region = self._xpath2obj('region', element)
         obj.evaluation_mode = self._xpath2obj('evaluationMode', element)
         obj.evaluation_status = self._xpath2obj('evaluationStatus', element)
         obj.creation_info = self._creation_info(element)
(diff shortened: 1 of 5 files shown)


megies:
Thanks for the PR @nackerley! I would even argue this should be based on `maintenance_1.0.x` as it clearly seems to be a bug fix to me..

obspy/io/quakeml/core.py
```diff
@@ -561,6 +561,7 @@ def _origin(self, element, arrivals):
         obj.composite_times = self._composite_times(element)
         obj.quality = self._origin_quality(element)
         obj.origin_type = self._xpath2obj('type', element)
+        obj.region = self._xpath2obj('region', element)
```
krischer: Yea apparently this was missing. We should generate a full QuakeML file like we did for StationXML (https://github.com/obspy/obspy/blob/master/obspy/io/stationxml/tests/data/full_random_stationxml.xml) to avoid this kind of problem in the future.

@barsch I think you generated the stationxml test file. Could you do the same for QuakeML?

krischer:
Oh - and there is a trailing whitespace on `core.py:1160:67`. Please fix that.

barsch:
@krischer: try http://xsd2xml.com/

megies:
> @krischer: try http://xsd2xml.com/

Unfortunately that one does not use random values, so all elements of the same data type have the same value, which is not optimal for testing. I also tried eclipse without luck..

PR #1617, opened 2016-12-18
How to implement ordering of "extra" tags.

megies:
When [custom defined tags](https://docs.obspy.org/tutorial/code_snippets/quakeml_custom_tags.html) are added to a catalog or its children and then this catalog is written to QuakeML, the order of these "extra" is random. 

This is not unexpected, as the "extra" info is essentially a dictionary and thus has no order, but it is an annoyance when trying to compare two QuakeML files, for example during unit testing.

How can I enforce an order when the "extra" tags are written to QuakeML? A simple sorting would be fine. If I need to define an order via a schema, can someone give me a simple example, assuming for example that all the fields are simple text?

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -25,6 +25,8 @@ master: (doi: 10.5281/zenodo.165135)
      objects (see #1514).
    * Instrument responses can now also be calculated for a given list of
      frequencies (see #1598).
+   * Order of extra tags for event type classes serialized to QuakeML can now
+     be controlled by using an OrderedDict (see #1617)
  - obspy.clients.fdsn:
    * empty SEED codes (e.g. ``network=''``) will now be properly sent to the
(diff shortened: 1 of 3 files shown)


megies:
I think replacing the base type of AttribDict can potentially be problematic and would affect large parts of the code base. But I think I found a solution we can live with, by just custom handling of the attribute setting for 'extra' in the event base type.

Other opinions?

If this gets a nod, would you mind adding a test @nackerley? (we can add you to developers for push rights)

megies:
> If this gets a nod, would you mind adding a test @nackerley? (we can add you to developers for push rights)

So.. is it working for you @nackerley? Always good to get some feedback when requested changes get worked on..

nackerley:
I'll happily write a test for this. I haven't gotten around taking advantage of this fix in my own code; I will try it soon.

megies:
Always good to get regression tests..

PR #1623, opened 2016-12-18
docker: install geographiclib

megies:
Install `geographiclib` in docker base images for testing of geodetic stuff, see #1236.


megies:
thanks for reviewing and merging @krischer 
\*clicks 'Delete branch'\*  ;-)

megies:
Looks like I've added python-geographiclib fro some distributions that don't have that package like e.g. Ubuntu 12.04..

PR #1625, opened 2016-12-19
Correctly writing zero station magnitude contributions and reading/writing event regions in QuakeML

### LINKED ISSUE #1621: Station magnitude residual of zero not read from QuakeML

nackerley:
Fixed #1621 by bringing Pickler._station_magnitude_contributions() into line with other Pickler methods, such as Pickler._arrival(), in particular by making use of Pickler._str(). This means that residual and weight are written even if they are both zero, and not written only when they are `None`.

I suspect that Pickler._station_magnitude_contributions() was missed when other methods were updated a while back to use Pickler._str() mainly because it is out-of-order with the rest which do use Pickler._str(). A more natural place for it among the methods would be between Pickler._station_magnitude() and Pickler._origin(), but that would have made the diff less clear.

Also fixed a bug whereby Event.region was neither written to QuakeML nor read.

Updated regression tests to reflect these changes, and added myself to contributors, even this is a very minor contribution.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -21,13 +21,16 @@
    * ObsPy can now read MiniSEED files with no blockette 1000. (see #1544)
    * ObsPy now always writes Blockette 100 if sampling rate accuracy is
      otherwise lost. (see #1550)
+ - obspy.io.quakeml:
+   * write StationMagnitude.residual even when it is zero (see #1625)
+   * read & write Event.region
  - obspy.io.sac:
    * `SACTrace.lpspol` and `lcalda` are `True` and `False` by default, when
       created via `SACTrace.from_obspy_trace` with a `Trace` that has no SAC
       inheritance. (see #1507)
    * Reference time not written to SAC file when made from scratch
      (see #1575)
-- obspy.io.sh:
+ - obspy.io.sh:
    * Fix writing of long headers for python3
    * Whitespace in header fields is not ignored anymore (see #1552)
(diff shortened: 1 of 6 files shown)


nackerley:
Forking from the maintenance branch seemed easier than resolving all the conflicts when rebasing from master to maintenance. Sorry for any confusion.

nackerley:
@megies re: travis. Thanks, I was wondering about the failures I saw when I ran the full test suite.

p.s. @krischer @megies Wonderful first interaction with the project. Looking forward to contributing more.

megies:
> p.s. @krischer @megies Wonderful first interaction with the project. Looking forward to contributing more.

Thanks for the fix and good to hear! ObsPy can use all the help it gets!

PR #1628, opened 2016-12-24
Add support for Reftek data encoding 'C2' (STEIM2)

megies:
Hello,

I get the following error when reading reftek data:

**Reftek data encoding 'ET' not implemented yet.** Please open an issue on GitHub and provide a small (< 50kb) test file.

Any chance that this will be fixed in the future ? I am attaching sample data as proposed.
[sample.tar.gz](https://github.com/obspy/obspy/files/671116/sample.tar.gz)

ObsPy version: 1.02
Python version: 2.7.11
Platform: Ubuntu 16
Python was installed from apt
ObsPy was downloaded .zip file from git and installed using pip install.
  
Thanks and best regards.


megies:
> Any chance that this will be fixed in the future ?

Yes, looks like this was pretty trivial. ;-)
Thanks for the test data.


@kbouk this is now a PR to add C2 encoding support (error message was wrong, EH is the header packet type). The unpacked data of your test file looks OK to me, please verify with some more of your data and give us feedback. (you can install as you did with master from this zip: https://github.com/obspy/obspy/archive/reftek_steim2.zip)

Still need to write a short test against the provided test data..

ThomasLecocq:
Hi all,
am testing the code and there seem to be an error when the `st = Stream()` is declared ; if the .rt file is multiplexed, the `st` should be declared before the `for event_number in np.unique(self._data['event_number']):` line

megies:
> there seem to be an error

I think you're right. Can you provide a small example file for testing?

megies:
According to @kbouk the data resulting from the unpacking is looking good:
https://github.com/obspy/obspy/pull/1632#issuecomment-271314601
```
 I tested the PR release mentioned in #1628 and it works.
```

So this should be good to go, I think.

obspy/io/reftek/core.py
```diff
@@ -241,10 +241,14 @@ def to_stream(self, network="", location="", component_codes=None,
             else:
                 eh = EHPacket(et_packets[0])
             # only "C0" encoding supported right now
-            if eh.data_format != b"C0":
+            if eh.data_format == b"C0":
+                encoding = 'C0'
+            elif eh.data_format == b"C2":
+                encoding = 'C2'
```
krischer: This is a bit awkward to be honest. Here its converted to bytes and in the `_unpack_C0_C2_data()` function its converted back to bytes.

megies: I agree, but..
 * I didn't want to do any unnecessary decoding of the encoding field of the packets array, but also
 * using a `kwarg` that expects a bytestring seemed also pretty strange to me

Open for suggestions.

krischer: Hmm...good point. If `data_format` is not used for anything else I would vote for directly decoding it upon reading in accordance with the sandwich principle. But its your call and really a very minor detail.

megies: I think it's OK if it stays as is. Too minor to worry about. Mainly I'm not doing decoding because then I would have to worry about decoding errors (ignore? replace? mainly for potentially garbled files..). Right now the user would get warned about the unrecognized encoding as a bytestring exactly as it is found in the file..

But like I said, too minor to give it more thought, really..

megies:
@kbouk just to make sure once more.. Did you test the decoding for meaningful waveform data (e.g. an event), so that you can confirm the decoding is correct? The test data is just some wiggles..

(But I guess we'd get tons of errors from the steim decoding routine if the data wouldn't make sense..)

krischer:
> (But I guess we'd get tons of errors from the steim decoding routine if the data wouldn't make sense..)

Yea - STEIM is checksummed so it either works or produces a gazillion errors.

PR #1631, opened 2017-01-06
Calibration using poles and zeros dictionary

mbdb:
Dear all,

I propose a short PR related to calibration.py in signal module.
This file currently takes a calib_file in gse2 format. This PR changes this for use a standard paz dictionary. It becomes the user's responsibility to get or write the correct paz.

Best regards,
Maxime Bes de Berc

### DIFF
--- a/obspy/signal/calibration.py
+++ b/obspy/signal/calibration.py
@@ -39,9 +39,10 @@ def rel_calib_stack(st1, st2, calib_file, window_len, overlap_frac=0.5,
 
     :param st1: Stream or Trace object, (known)
     :param st2: Stream or Trace object, (unknown)
-    :type calib_file: str
+    :type calib_file: str or dict
     :param calib_file: file name of calibration file containing the PAZ of the
-        known instrument in GSE2 standard.
+        known instrument in GSE2 standard or a dictionary with poles and zeros
+        information (with keys ``'poles'``, ``'zeros'`` and ``'sensitivity'``).
     :type window_len: float
     :param window_len: length of sliding window in seconds
     :type overlap_frac: float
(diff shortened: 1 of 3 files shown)


megies:
This will break existing codes, so we can not merge as is.

If you remodel this to have an if/else that processes dictionaries as you propose and use strings as filenames to a GSE paz file (like before) we could consider merging this and it I think it would be a nice improvement.

mbdb:
Of course, I should have thought of it.
I updated the PR, to accept either a paz dict or a gse2 file.
Let me know,
M.

obspy/signal/calibration.py
```diff
@@ -31,17 +31,17 @@
 from obspy.signal.util import next_pow_2
 
 
-def rel_calib_stack(st1, st2, calib_file, window_len, overlap_frac=0.5,
+def rel_calib_stack(st1, st2, calib_obj, window_len, overlap_frac=0.5,
```
megies: Please leave the arg name unchanged, renaming it might break codes.

obspy/signal/calibration.py
```diff
@@ -138,19 +138,25 @@ def rel_calib_stack(st1, st2, calib_file, window_len, overlap_frac=0.5,
...
-    :type calfile: str
-    :param calfile: file containing poles, zeros and scale factor for known
-        system
+    :type calobj: str or dict
+    :param calobj: dictionary or file name (gse2) containing poles, zeros and
+    sensitivity for known system
     :returns: complex transfer function, array of frequencies
     """
+    # read paz file if a file is specified
+    if isinstance(calobj, str):
```
megies: This won't work across Python2/3, I fear, because of our usage of the `future` package. You need to make it: `if isinstance(paz_file, (str, native_str)):`

krischer: I would actually to the opposite and first check if its a dictionary and otherwise call the `read_paz()` function. Maybe that function will learn a few tricks in the future and then this will just work.

mbdb: Done, it checks if it is a dict first, otherwise read file.

obspy/signal/calibration.py
mbdb: Done, original args kept.

krischer:
> Good to merge if CI passes, I think.
> 
> We should squash everything into one commit before merging..

I agree with both :-) Please squash to one or two commits.

mbdb:
Squashed all commits to one.

megies:
Looking good now, thanks for the contribution @mbdb!

PR #1633, opened 2017-01-09
Missing preferred_origin (but it is defined) - Implementation

krischer:
PR for issue #1629.


krischer:
Handled directly in #1629.

PR #1629, opened 2017-01-09
Missing preferred_origin (but it is defined)

megies:
When I retrieve events via the FDSN web client, I occasionally get an event that has a `preferred_origin_id` (and a corresponding entry in `origins`) but `preferred_origin()` returns None.

```python
from obspy.clients import fdsn
client = fdsn.Client('IRIS')

catalog = client.get_events(starttime='2016-12-01', minmagnitude=4)
bad_events = [event for event in catalog if not event.preferred_origin()]
len(bad_events)
```
> 1

```python
bad_event = bad_events[0]
bad_event.preferred_origin_id
```

> smi:service.iris.edu/fdsnws/event/1/query?originid=17807622

```python
bad_event.origins[0].resource_id
```

> smi:service.iris.edu/fdsnws/event/1/query?originid=17807622

But the same event retrieved in a different way appears fine.

```python
catalog = client.get_events(eventid=5199371)
event = catalog[0]
event.preferred_origin()
```

> Origin
    	   resource_id: ResourceIdentifier(id="smi:service.iris.edu/fdsnws/event/1/query?originid=17807622")
    	          time: UTCDateTime(2016, 12, 1, 0, 21, 40, 290000)
    	     longitude: -70.3105
    	      latitude: -28.715
    	         depth: 114950.0
    	 creation_info: CreationInfo(author='us')

Actually, it looks like there is some sort of stateful behavior here, when making repeated requests this behavior oscillates. (The exact values produced here are somewhat unpredictable but the oscillation itself is consistent.)

```python
for _ in range(4):
    catalog = client.get_events(starttime='2016-12-01', minmagnitude=4)
    bad_events = [event for event in catalog if not event.preferred_origin()]
    print "%d of %d bad events" % (len(bad_events), len(catalog))
```

> 768 of 807 bad events
17 of 807 bad events
768 of 807 bad events
17 of 807 bad events

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,8 @@
      dayplot of a trace, see #1566)
    * properly pass through kwargs from Stream.detrend() to Trace.detrend()
      (see #1607)
+   * Resource identifiers will now always point to the newest object in case
+     they are created multiple times with identical objects. (see #1629)
  - obspy.core.event.source:
    * Fix `farfield` if input `points` is a 2D array. (see #1499, #1553)
  - obspy.clients.earthworm:
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -623,17 +623,21 @@ class ResourceIdentifier(object):
         Will also append self again to the global class level reference list so
         everything stays consistent.
         """
-        # If it does not yet exists simply set it.
-        if self.id not in ResourceIdentifier.__resource_id_weak_dict:
+        # If it does not yet exists, set it.
+        #
+        # If it already exists, set the referred object to the current one if
+        # the existing referred object is identical to the new one.
+        # Use the new one as the old might be more readily garbage collected.
+        # This is a weak heuristic but it is all we can do from here.
+        if self.id not in ResourceIdentifier.__resource_id_weak_dict or \
+                ResourceIdentifier.__resource_id_weak_dict[self.id] == \
+                referred_object:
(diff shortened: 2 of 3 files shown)


megies:
> Maybe github turned of issue to PR conversion? In any case: continued in #1633

Doesn't work if the head branch is on a fork apparently. I'll let github support know.

krischer:
@adam-iris: Can you check if this fixes the problem in your PyQt application?

megies:
> @adam-iris: Can you check if this fixes the problem in your PyQt application?

ping @adam-iris ?

adam-iris:
Yes, this seems to fix the problem. Thanks, and sorry for the slow response!

megies:
> Yes, this seems to fix the problem. Thanks, and sorry for the slow response!

No worries! Thanks for the feedback!

megies:
@krischer I have a feeling this PR might be related to recently #1644 which was merged into master for 1.1.0 recently..? So I'll leave this for you to judge..

krischer:
https://github.com/obspy/obspy/pull/1644 solves this in a way more general manner. I'm thus closing this.

The only difference is that #1644 is merged into `master` and will not go into `maintenance_1.0.x` and the underlying implementation changed too much for that.

I manually added the one new test to `master` to make sure the problem is actually fixed in fc92b042964ad3d0ab06dd4b26651a47e63a1a85.

PR #1636, opened 2017-01-10
ObsPy broken on Windows when using new Python 2.7.13

krischer:
ObsPy currently is completely broken on Windows when using Python 2.7.13 which was released 24 days ago. Mac and Linux seem to be not affected.

```
C:\Miniconda-x64\envs\test\python.exe: Could not load shared library "libmseed_Windows_64bit_py27.pyd".
 LoadLibrary() argument 1 must be string, not unicode
```

Last good appveyor run (on Python 2.7.12), dating from 2016-12-18: https://ci.appveyor.com/project/obspy/obspy/build/1.0.3911-master#L25

First bad appveyor run (on Python 2.7.13), dating from 2016-12-19: https://ci.appveyor.com/project/obspy/obspy/build/1.0.3912-master#L25

Upstream bug report for Python: https://bugs.python.org/issue29082

Other projects experiencing the same issue: rg3/youtube-dl#11540 

In principle we could just wait for Python 2.7.14 and tell Windows users to avoid 2.7.13 but it would be better to fix it on our end, I guess.

### DIFF
--- a/obspy/core/util/libnames.py
+++ b/obspy/core/util/libnames.py
@@ -91,7 +91,7 @@ def _load_cdll(name):
                           'lib')
     libpath = os.path.join(libdir, libname)
     try:
-        cdll = ctypes.CDLL(libpath)
+        cdll = ctypes.CDLL(str(libpath))
     except Exception as e:
         msg = 'Could not load shared library "%s".\n\n %s' % (libname, str(e))
         raise ImportError(msg)


krischer:
Now a PR attempting to fix the problem. I don't have a working windows machine so I'll just abuse appveyor for debugging...

megies:
All CI passes, so merging.

DianKusumawati:
I have the same problem.

On Feb 16 2017, I installed anaconda (with python 2.7 version 64 bit) and then ObsPy on my Windows 8 64 bit. The ObsPy seems not working. When I import ObsPy in Spyder, it says:

problem on obspy installed on windows

Is there any solution for this, already?

Thank you

megies:
It's already fixed, but there was no release since the fix, sorry..  guys, we should really get 1.0.3 out!

megies:
@DianKusumawati as a quick and dirty short term solution for you, you could try to manually apply the fix in 1fa7d828cc7b2b753afee4aa237ca844ac3635c1 in your installed obspy 1.0.2 version in your anaconda env.. (the path to the `libanmes.py` is shown in your traceback at the bottom)

DianKusumawati:
Thank you @megies for your reply : D ,I'll try it

PR #1627, opened 2017-01-10
Problem with trace endtime reading mSEED

krischer:
Hi all.. I am having a little problem with a mSEED file. When i read a file without specifying a starttime the end of the trace somehow becomes shorter. File is here: ftp://www.orfeus-eu.org/pub/outgoing/NL.VLW1.00.HHE.D.2016.143

here is what is happening:

    print read("NL.VLW1.00.HHE.D.2016.143")
        >>> 1 Trace(s) in Stream:
        >>> NL.VLW1.00.HHE | 2016-05-22T00:00:03.001557Z - 2016-05-23T00:00:05.851557Z | 100.0 Hz, 8640286 samples

    print read("NL.VLW1.00.HHE.D.2016.143", starttime=UTCDateTime("2016-05-23"))
        >>> 1 Trace(s) in Stream:
        >>> NL.VLW1.00.HHE | 2016-05-23T00:00:00.001878Z - 2016-05-23T00:00:06.411878Z | 100.0 Hz, 642 samples

Pay close attention to the trace enditme.

I looked through the records using libmseed and I think the correct trace endtime is the second one. So the trace endtime is incorrect when reading without a window selection.

`msi -T` gives the same as the expected result:

       Source                Start sample             End sample        Hz  Samples
    NL_VLW1_00_HHE    2016,143,00:00:03.001557 2016,144,00:00:06.411878 100 8640286

Also notice how the start time and number of samples are equal but the end time is not.

Best!

### DIFF
--- a/obspy/io/mseed/headers.py
+++ b/obspy/io/mseed/headers.py
@@ -664,6 +664,7 @@ ContinuousSegment._fields_ = [
     ('samprate', C.c_double),
     ('sampletype', C.c_char),
     ('hpdelta', C.c_longlong),
+    ('leapseconds', C.c_longlong),
     ('samplecnt', C.c_int64),
     ('timing_quality', C.c_uint8),
     ('calibration_type', C.c_int8),
--- a/obspy/io/mseed/src/obspy-readbuffer.c
+++ b/obspy/io/mseed/src/obspy-readbuffer.c
@@ -261,6 +261,48 @@ void log_error(int errcode, int offset) {
 }
 
 
+/***************************************************************************
+ * segment_endtime:
+ *
+ * Calculate the time of the last sample in the segment. Similar to
+ * msr_endtime() but for the ObsPy internal segment structure.
+ ***************************************************************************/
+hptime_t segment_endtime(ContinuousSegment *seg)
+{
+  hptime_t span = 0;
+  LeapSecond *lslist = leapsecondlist;
+
+  if ( ! seg )
+    return HPTERROR;
+
+  if ( seg->samprate > 0.0 && seg->samplecnt > 0 )
+    span = (hptime_t)(((double) (seg->samplecnt - 1) / seg->samprate * HPTMODULUS) + 0.5);
+
+  /* Check if the record contains a leap second, if list is available */
+  if ( lslist )
+    {
+      while ( lslist )
+        {
+          if ( lslist->leapsecond > seg->starttime &&
+               lslist->leapsecond < (seg->starttime + span) )
+            {
+              span -= HPTMODULUS;
+              break;
+            }
+          lslist = lslist->next;
+        }
+    }
+  else
+    {
+      // Otherwise just use the offset as determined from the
+      // activity flags.
+      span += seg->leapseconds;
+    }
+
+  return (seg->starttime + span);
+} /* End of segment_endtime() */
+
+
 // Function that reads from a MiniSEED binary file from a char buffer and
 // returns a LinkedIDList.
 LinkedIDList *
@@ -592,17 +634,31 @@ readMSEEDBuffer (char *mseed, int buflen, Selections *selections, flag
              // record and after it.
              recordCurrent->record->samplecnt > 0 && segmentCurrent->samplecnt > 0 &&
 
+             // Test the sampletype.
              segmentCurrent->sampletype == recordCurrent->record->sampletype &&
+
              // Test the default sample rate tolerance: abs(1-sr1/sr2) < 0.0001
              MS_ISRATETOLERABLE (segmentCurrent->samprate, recordCurrent->record->samprate) &&
+
              // Check if the times are within the time tolerance
              lastgap <= hptimetol && lastgap >= nhptimetol &&
              segmentCurrent->timing_qual == timing_qual &&
+
+             // And check the calibration type.
              segmentCurrent->calibration_type == calibration_type) {
-            recordCurrent->previous = segmentCurrent->lastRecord;
-            segmentCurrent->lastRecord = segmentCurrent->lastRecord->next = recordCurrent;
-            segmentCurrent->samplecnt += recordCurrent->record->samplecnt;
-            segmentCurrent->endtime = msr_endtime(recordCurrent->record);
+
+             recordCurrent->previous = segmentCurrent->lastRecord;
+             segmentCurrent->lastRecord = segmentCurrent->lastRecord->next = recordCurrent;
+             segmentCurrent->samplecnt += recordCurrent->record->samplecnt;
+
+             // Add leap seconds to the current segment if there are any.
+             if ( recordCurrent && recordCurrent->record && recordCurrent->record->fsdh ) {
+                 if ( recordCurrent->record->fsdh->act_flags & 0x10 ) {
+                     segmentCurrent->leapseconds -= HPTMODULUS;
+                 }
+             }
+
+             segmentCurrent->endtime = segment_endtime(segmentCurrent);
         }
         // Otherwise create a new segment and add the current record.
         else {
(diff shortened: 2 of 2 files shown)


krischer:
This is actually a bug (or a sloppy implementation) of how ObsPy decided to split traces for MiniSEED files.

While this is potentially a very bad bug I don't think it affects a lot of real world files. This particular example manifests the problem because either the sampling rate is not accurate or there is a very strong clock drift. I tested with a couple of other files and it seemed to be fine. Still - this is something we should check a bit more carefully and definitely communicate to our users!

So what happened was that a new trace was created if the start time of the next record is more than then `0.5 * dt `from `time of last sample of previous record + dt`. A certain tolerance is needed here. This unfortunately did not account for accumulated errors if many records are attached one after the other.

The implementation of this now PR calculates the endtime of the full current trace (including all already added records). If either the sampling rate or the clock is wrong this will very quickly manifest in the tolerance being exceeded which results in a split trace.

Its late and I hope the whole explanation makes sense. Let me know if not.

This PR is still WIP as it definitely still needs some more tests. The whole leap second business is also not really tested and we should really make `UTCDateTime` leap second aware as well!

With this branch the result for the example file is now:

```
112 Trace(s) in Stream:

NL.VLW1.00.HHE | 2016-05-22T00:00:03.001557Z - 2016-05-22T00:12:49.701557Z | 100.0 Hz, 76671 samples
...
(110 other traces)
...
NL.VLW1.00.HHE | 2016-05-22T23:51:57.468857Z - 2016-05-23T00:00:06.408857Z | 100.0 Hz, 48895 samples

```

You'll note that the endtime is still not identical to the `msi` output but it is within the tolerance of `0.5 * dt` so that should be all good.

Jollyfant:
Ah, so basically all the time tolerances are used to check if the segment is continuous (which it is). But this cumulative offset is then not included in the final trace end time. Thanks a lot for the explanation and pull request. I know for a fact this particular file is the result of some hacky implementation in the datalogger, and this is the first time I come across it so it really doesn't affect many files.

krischer:
Exactly.

ObsPy's data model does not explicitly store the end time. We only store the time of the first sample, the number of samples, and the sampling rate. The end time is calculated on-the-fly under the assumption that all data samples are equally spaced and that the sampling rate is correct. If that is no longer the case at a certain point it will have to be split into multiple traces.

Thus the new behavior with this PR is IMHO the correct way to handle it. Many traces within a file point towards a problem with the data logger.

chad-iris:
Just some food for thought:
This is a common, sticky problem when reconstituting longer time series from miniSEED records when there is inconsistency between the sample rate and the rate that would be calculated from the time stamps of an otherwise continuous series of samples.  In my experience it is almost always the rate that is inaccurate, due to clock drift or low resolution, for a series of samples that are more or less regularly sampled.

A common approach is to do the tolerance-based continuity checks between records and track the end time of the trace based on the last record in the series.  This approach gives preference to the time stamps over the sampling rate and requires tracking the end time.

The description for this PR sounds like it would split what could otherwise be a continuous series of samples into multiple segments.  This approach, as I understand it, assumes that both the time stamps and the sample rate are correct.  The downside is introducing time tears that may not be real, making data usage harder.  Also, the details that would be needed to reconstruct the continuous series (all the record time boundaries) are gone.

An alternate approach would be to trust the time stamps, check the specified rate versus the calculated rate and report to the user when there is mismatch.  There could also be an option to recalculate the rate (because that is usually what is wrong).  After all, if the time stamps are bad the value of the data is significantly reduced, well beyond any problems of reconstructing a contiguous series from miniSEED records.

Jollyfant:
Is there any approach that you prefer, @chad-iris? In any case it is unexpected that ObsPy silently merges the records. But I reckon that trusting the time stamps and introducing time tears is the most pragmatic solution as it clearly indicates a problem with the logger. AFAIK this is the approach has now been implemented by @krischer.

krischer:
Thanks all for your thoughts!

The most user friendly approach might be to trust the time stamps and adjust the sampling rate as long as its is not too different from the one written in the file. This would of course raise a warning so users know something is up and it can also be turned off in which case it would fall back to the solution currently implemented in this PR.

All of this has to be implemented at reading time because ObsPy's `Trace` object has no concept of an inaccurate sampling rate.

Thoughts?

chad-iris:
> Is there any approach that you prefer, @chad-iris? In any case it is unexpected that ObsPy silently merges the records. But I reckon that trusting the time stamps and introducing time tears is the most pragmatic solution as it clearly indicates a problem with the logger.

I'm confused why merging all the records is unexpected.  The provided example data reconstitutes into a single, continuous series if the time stamps are assumed correct and a reasonable tolerance between records is applied.  It is only when the sample rate is assumed to be correct over many records/samples that apparent time tears occur.

> All of this has to be implemented at reading time because ObsPy's Trace object has no concept of an inaccurate sampling rate.
> Thoughts?

Varying the sample rate, within a tolerance, while reading would probably work.

An alternative, and likely more fundamental change, is to include an end time in a Trace; when generating a Trace from miniSEED track the end time as calculated from the last record contributing to the Trace.  Then you can retain both the recorded "nominal" sample rate and time span from the input data without splitting into multiple segments.  The Trace is then internally inconsistent, but so are the data.  The upside is that the inconsistency is exposed to the user, who can still choose to trust the times or the rate, but has a single continuous trace to work with.

Jollyfant:
I have created an extreme example of what I mean with unexpected behaviour.

Create 40000 continuous records separated by `0.5 dt`:

    40000 Trace(s) in Stream:
    ... | 1970-01-01T00:00:00.000000Z - 1970-01-01T00:00:09.000000Z | 1.0 Hz, 10 samples
    ...
    (39998 other traces)
    ...
    ... | 1970-01-05T20:39:49.500000Z - 1970-01-05T20:39:58.500000Z | 1.0 Hz, 10 samples

Last record ends at 20:39. Write => read this stream through ObsPy:

    1 Trace(s) in Stream:
    ... | 1970-01-01T00:00:00.000000Z - 1970-01-05T15:06:39.000000Z | 1.0 Hz, 400000 samples

It silently merges the everything and puts the end time on 15:06. Maybe it is technically not surprising the traces are merged but the result is unexpected :)

krischer:
@Jollyfant Can you please upload the file? That makes for a good test case.


> Varying the sample rate, within a tolerance, while reading would probably work.

I'll add this as the default with a user-adjustable tolerance and raised warnings. That at least gives continuous traces to work with. The current behavior will also be a possible user choice.

> An alternative, and likely more fundamental change, is to include an end time in a Trace; when generating a Trace from miniSEED track the end time as calculated from the last record contributing to the Trace. Then you can retain both the recorded "nominal" sample rate and time span from the input data without splitting into multiple segments. The Trace is then internally inconsistent, but so are the data. The upside is that the inconsistency is exposed to the user, who can still choose to trust the times or the rate, but has a single continuous trace to work with.

I'll add the "best guess" end time (calculated from start time of last record) to the MiniSEED specific stats header. Then at least users have a chance to work with it. Adding inconsistent end times to the core ObsPy Trace structure would indeed be a larger undertaking and I'm not sure its even possible (what do to with operations that for example cut the traces?).

Jollyfant:
Reduced to 4k records (16M):

 ftp://www.orfeus-eu.org/pub/outgoing/tt_records.mseed

megies:
> Reduced to 4k records (16Mb):

16 Mb is too much, but looks like you have a code snippet to create the test data on-the-fly during testing?

Jollyfant:
16 Mb is too much, but looks like you have a code snippet to create the test data on-the-fly during testing?

Yeah, that is right. It can be used for testing.

krischer:
Bumping this to 1.1.0 as it is not a trivial fix and i still need to find the time to do it. Additionally it will also slightly change the way MiniSEED files are read so it should go into a proper release and not a bugfix release.

megies:
We decided to push for 1.1.0 and since I'm assuming this will not be ready by Sunday I'm punting this to 1.2.0. Feel free to push on this if you feel it can be ready by Sunday.

krischer:
Not for 1.1.0 - this is quite a big change and I would not feel comfortable having it in 1.1.0 with so little time to actually test it even if I manage to get it done by then. I'm a bit short on time this week - I should have more time next week :)

Jollyfant:
This only really affects certain edge cases so there is really no rush to tackle this.

megies:
This is against a maintenance branch but has 1.2.0 as milestone.. I'll put it in 1.1.1 for now, feel free to change @krischer you might have a better idea of what this PR is doing.

megies:
Changed base branch to `maintenance_1.1.x`. Needs a rebase on this base branch, eventually.

krischer:
This changes behavior so it has to go into a new major release. I hope to get to do this properly soon.

megies:
Any time for this is near future, anybody? Otherwise I'll bump to 1.3.0

PR #1641, opened 2017-01-14
fix bug in taper docs

seisman:
This PR fix a small incompatibility between codes and docs.

+DOCS


krischer:
Thanks a bunch! Also thanks for fixing it against the correct branch!

PR #1642, opened 2017-01-14
fix test_farfield_2xn_input

seisman:
This PR fixes two bugs in `farfield` unittest:

1. The function `farfield` expects six component moment tensor in NED system, while GCMT gives moment tensor in RTP system. So a coordinate transformation from RTP to NED is needed.

2. The 2xn vector [theta, phi] is in radian not in degree.

+TESTS: core.event.source


megies:
> +TESTS: core.event.source

core tests are always executed, you only need this for "network" modules, that are not run by default, see https://github.com/obspy/obspy/blob/master/obspy/core/util/base.py#L45

krischer:
@MMesch @ThomasLecocq Can one of you have a look and review this? Thanks! :-)

krischer:
Am I correct in assuming that the test just did not test something particularly useful and was documented wrongly but the actualy calculations were still right?

megies:
> @MMesch @ThomasLecocq Can one of you have a look and review this? Thanks! :-)

ping?

MMesch:
Side question: should we (internally) prefer any particular convention for the moment tensor in obspy?

megies:
> Side question: should we (internally) prefer any particular convention for the moment tensor in obspy?

Actually, `Tensor` and thus also `MomentTensor` in `core.event.source` are by definition of QuakeML Up-North-West? Generally, we could enhance these classes by making them coordinate system aware and capable of transforming themselves.. It would also be best to handle all moment tensors anywhere in our code base by such a proper object (as opposed to passing around numpy arrays)..

MMesch:
OK. I would propose to try using internally always the same convention, with one function that does the conversion at the beginning of a function call. Otherwise it will get confusing.

MMesch:
PS: if we use an object that problem would be fixed anyway

megies:
So, what about this one, @MMesch @ThomasLecocq? Calculations are correct looks like, just the test was wrong? Can we merge this to go forward with 1.0.3 or does this need more work?

MMesch:
I think it looks good!

PR #1644, opened 2017-01-17
Object aware ResourceIdentifier

d-chambers:
### What does this PR do?

This pull request changes the default behavior of the ResourceIdentifier class slightly to allow for multiple objects to share the same resource id and still be accessible via the get_referred_object method. However, if no object was passed to either the __init__ method or the set_referred_object method, or if the original object goes out of scope, the behavior is the same as before in that the last defined object with the resource_id is returned.

Let me demonstrate with some code:

```python
import obspy
cat1 = obspy.read_events()
# The resource_id attached to the first event should point to the first event
cat1[0] is cat1[0].resource_id.get_referred_object()  # True
# but what happens if we make a copy? 
cat2 = cat1.copy()
# or perform another read operation?
cat3 = obspy.read_events()
# the resource_id on the new catalogs dont point to their attached objects
cat1[0] is cat1[0].resource_id.get_referred_object()  # True
cat2[0] is cat2[0].resource_id.get_referred_object()  # False
cat3[0] is cat3[0].resource_id.get_referred_object()  # False
# now what happens if the original catalog goes out of scope?
del cat1
cat2[0].resource_id.get_referred_object()  # None
cat3[0].resource_id.get_referred_object()  # None
# the resource_ids on the remaining objects are no longer usable
```

I have run into problems several times by using the get_referred_object method of the ResourceIdentifier class, both in testing and non-testing environments. For example, when trying to reference a pick from a phase often times the get_referred_object method returns none (even though the pick still exists in the current catalog), or harder to track yet, returns the pick from a copy of the catalog object. 

One solution, then, is to never copy an object with a resource_id or never read it in twice. This can, however, be impractical in some work flows, especially test environments that perform modifications to the object. The proposed modifications remove the surprising behavior and the code now executes as one (or at least me) might expect :

```python
import obspy

cat1 = obspy.read_events()
# The resource_id attached to the first event should point to the first event
cat1[0] is cat1[0].resource_id.get_referred_object()  # True
cat2 = cat1.copy()
cat3 = obspy.read_events()
# the resource_id on the new catalogs now point to their attached objects
cat1[0] is cat1[0].resource_id.get_referred_object()  # True
cat2[0] is cat2[0].resource_id.get_referred_object()  # True
cat3[0] is cat3[0].resource_id.get_referred_object()  # True
# now what happens if the original catalog goes out of scope? Nothing really...
del cat1
cat2[0].resource_id.get_referred_object()  # Event (the one you would expect)
cat3[0].resource_id.get_referred_object()  # Event
# now if we create a resource_id with the same id, the last defined object
# with the same resource id (that is still in scope) is returned
new_id = cat2[0].resource_id.id
rid = obspy.core.event.ResourceIdentifier(new_id)
rid.get_referred_object() is cat3[0]  # True
# if cat3 went out of scope then the object from cat2 would be returned
del cat3
rid.get_referred_object() is cat2[0]  # True
# if all objects with the same resource_id get gc’ed then None is returned
del cat2
rid.get_referred_object() is None  # True
```

### PR Checklist
- [x] All tests still pass.  (one fails but it failed on fresh fork)

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -38,7 +38,6 @@ import numpy as np
 from obspy.core.event.header import DataUsedWaveType, ATTRIBUTE_HAS_ERRORS
 from obspy.core.utcdatetime import UTCDateTime
 from obspy.core.util import AttribDict
-from obspy.core.util.decorator import rlock
 
 
 class QuantityError(AttribDict):
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -401,6 +401,23 @@ def _event_type_class_factory(class_name, class_attributes=[],
             super(AbstractEventTypeWithResourceID, self).__init__(*args,
                                                                   **kwargs)
 
+        def __deepcopy__(self, memodict=None):
+            """
+            reset resource_id's object_id after deep copy to allow the
+            object specific behavior of get_referred_object
+            """
+            memodict = memodict or {}
+            cls = self.__class__
+            result = cls.__new__(cls)
+            memodict[id(self)] = result
+            for k, v in self.__dict__.items():
+                setattr(result, k, deepcopy(v, memodict))
+            result.resource_id.object_id = id(result)
+            return result
+
+
+
+
     if "resource_id" in [item[0] for item in class_attributes]:
         base_class = AbstractEventTypeWithResourceID
     else:
@@ -451,7 +468,11 @@ class ResourceIdentifier(object):
     :type referred_object: Python object, optional
     :param referred_object: The object this instance refers to. All instances
         created with the same resource_id will be able to access the object as
-        long as at least one instance actual has a reference to it.
+        long as at least one instance actually has a reference to it.
+        Additionally, ResourceIdentifier instances that have the same id but
+        different referred_objects will still return the referred object,
+        provided it doesn't get garbage collected. If the referred object no
+        longer exists any object with the same id will be returned.
 
     .. rubric:: General Usage
 
@@ -565,10 +586,10 @@ class ResourceIdentifier(object):
     ...'foo' bar2
     """
     # Class (not instance) attribute that keeps track of all resource
-    # identifier throughout one Python run. Will only store weak references and
-    # therefore does not interfere with the garbage collection.
+    # identifier throughout one Python run. Will only store weak references
+    # and therefore does not interfere with the garbage collection.
     # DO NOT CHANGE THIS FROM OUTSIDE THE CLASS.
-    __resource_id_weak_dict = weakref.WeakValueDictionary()
+    __resource_id_weak_dict = {}  # a nested dict for weak object references
     # Use an additional dictionary to track all resource ids.
     __resource_id_tracker = collections.defaultdict(int)
 
@@ -587,6 +608,7 @@ class ResourceIdentifier(object):
             self.id = id
         # Append the referred object in case one is given to the class level
         # reference dictionary.
+        self.object_id = None  # the object specific ID
         if referred_object is not None:
             self.set_referred_object(referred_object)
 
@@ -617,8 +639,39 @@ class ResourceIdentifier(object):
         Will return None if no object could be found.
         """
         try:
-            return ResourceIdentifier.__resource_id_weak_dict[self.id]
+            rdic = ResourceIdentifier.__resource_id_weak_dict[self.id]
         except KeyError:
+            return None
+        else:
+            if self.object_id in rdic and rdic[self.object_id]() is not None:
+                return rdic[self.object_id]()
+            else:  # find last added obj that is not None
+                return self._get_similar_referred_object()
+
+    def _get_similar_referred_object(self):
+        """
+        Find an object with the same resource_id that is not None and
+        return it. Also pop all keys that have None values in the
+        __resource_id_weak_dict
+        """
+        rdic = ResourceIdentifier.__resource_id_weak_dict[self.id]
+        if self.object_id is not None:
+            msg = ("The object with identity of: %d no longer exists, "
+                   "returning the most recently created object with a"
+                   " resource id of: %s.") % (self.object_id, self.id)
+            line_number = inspect.currentframe().f_back.f_lineno
+            warnings.warn_explicit(msg, UserWarning, __file__,
+                                   line_number)
+            self.object_id = None  # reset object id
+        # find a obj that is not None starting at last in ordered dict
+        for key in reversed(rdic):
+            obj = rdic[key]
+            if obj is not None:
+                return obj()
+            else:  # remove references that are None
+                rdic.pop(key)
+        else:  # if iter runs out all objects are none; pop rid, return None
+            ResourceIdentifier.__resource_id_weak_dict.pop(self.id)
             return None
 
     def set_referred_object(self, referred_object):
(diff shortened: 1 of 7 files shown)


krischer:
In general I quite like this but I'll have to review it in a bit more detail later - so thanks a lot! 

Also see this related PR which is likely obsolete due to your new PR as your solution is better: https://github.com/obspy/obspy/pull/1629

obspy/core/tests/test_event.py
```diff
@@ -681,30 +717,22 @@ def test_same_resource_id_different_referred_object(self):
...
-            # ResourceIdentifier.
-            warnings.simplefilter('ignore', UserWarning)
-            res_b = ResourceIdentifier(id=id,
-                                       referred_object=object_b)
-        # Object b was the last to added, thus all resource identifiers will
-        # now point to it.
-        self.assertEqual(object_b is res_a.get_referred_object(), True)
-        self.assertEqual(object_b is res_b.get_referred_object(), True)
+        # This will no longer raise a warning
+        res_b = ResourceIdentifier(id=id, referred_object=object_b)
```
krischer: I'm fine with the behavior of the code but I think it should really still raise a warning if two non-equal objects share the same resource identifiers. The whole purpose of resource ids is to point to unique or at least equal objects. If that is not the case some logic is wrong and the users should be notified.

obspy/core/tests/test_event.py
```diff
@@ -754,11 +782,47 @@ def test_adding_a_referred_object_after_creation(self):
...
+        obj2 = UTCDateTime()
+        rid1 = ResourceIdentifier(uri, referred_object=obj1)
+        rid2 = ResourceIdentifier(uri, referred_object=obj2)
+        self.assertFalse(rid1.get_referred_object() is
+                         rid2.get_referred_object())
+        del obj1
+        warnings.simplefilter('default')
+        with warnings.catch_warnings(record=True) as w:
+            rid1.get_referred_object()
+            assert len(w)
```
krischer: Can you please add some fuzzy test of the warning message? Just to make sure that the correct warning is actually caught here.

obspy/core/tests/test_event.py
```diff
@@ -773,12 +837,11 @@ def test_resources_in_global_dict_get_garbage_collected(self):
         # Now two keys should be in the global dict.
         rdict = ResourceIdentifier._ResourceIdentifier__resource_id_weak_dict
         self.assertEqual(len(list(rdict.keys())), 2)
-        # Deleting the objects should also remove the from the dictionary.
+        # Deleting the objects will no longer remove them from the dict, but
+        # the weak reference still behave as expected
```
krischer: Can you please change this comment? It will be a bit confusing without having the history of the file at hand.

d-chambers:
@krischer thanks for looking at this, happy to make the changes. 

Also, It occurred to me that this may not actually solve the problem of an arrival returning the wrong (non-event scoped) pick because the object_id attribute of the arrival.pick_id may never get set. One (ugly) solution might be to add a recursive function call on the Event.__init__ and Event.__deepcopy__ to find any ResourceIdentifier instances (not named resource_id) and set the object_id to that of the referred_object. Any thoughts? I should have time to work on this more in a few days.

krischer:
> Also, It occurred to me that this may not actually solve the problem of an arrival returning the wrong (non-event scoped) pick because the object_id attribute of the arrival.pick_id may never get set. One (ugly) solution might be to add a recursive function call on the Event.init and Event.deepcopy to find any ResourceIdentifier instances (not named resource_id) and set the object_id to that of the referred_object. Any thoughts? I should have time to work on this more in a few days.

Hmm...true. On the other hand `get_referred_object()` would still work but there is a chance that it return a pick from a different object which, by definition, should be equal to the event scoped pick. So yea - would be nice if this could be solved but no very big deal if not.

I'm not sure how to solve this to be honest - the ugly and tedious recursive solution will probably work for the deep copy but not for the normal .init case. Many people (and I'm pretty sure ObsPy internally also does this in lots of places) first create the event objects and only then fill it. 

We could create a `_internally_bind_resource_ids(event)` method that does this for a finished event and then automatically call this in `read_events()` so it will work in 99.9% of all cases.

krischer:
> Also, It occurred to me that this may not actually solve the problem of an arrival returning the wrong (non-event scoped) pick because the object_id attribute of the arrival.pick_id may never get set. One (ugly) solution might be to add a recursive function call on the Event.init and Event.deepcopy to find any ResourceIdentifier instances (not named resource_id) and set the object_id to that of the referred_object. Any thoughts? I should have time to work on this more in a few days.

Hmm...true. On the other hand `get_referred_object()` would still work but there is a chance that it return a pick from a different object which, by definition, should be equal to the event scoped pick. So yea - would be nice if this could be solved but no very big deal if not.

I'm not sure how to solve this to be honest - the ugly and tedious recursive solution will probably work for the deep copy but not for the normal .init case. Many people (and I'm pretty sure ObsPy internally also does this in lots of places) first create the event objects and only then fill it. 

We could create a `_internally_bind_resource_ids(event)` method that does this for a finished event and then automatically call this in `read_events()` so it will work in 99.9% of all cases.

d-chambers:
Ok, I think I addressed all of the issues but I am happy to make additional changes if needed.

I ended up adding another class dictionary to ResourceIdentifer that keeps track of all the ResourceIdentifer instances that have not been bound (ie had their object_id attribute set). I then added a ResourceIdentifier class method that iterates the dictionary and binds all of the unbound identifiers. I modified event level copy and the read_events function to call this method to get the "event scoped" behavior. Users wont have to know about this detail unless they manually create a catalog, as you described, in which case when finished they can simply do:
`resource_id.bind_resource_ids()`.
on any ResourceIdentifier instance.

obspy/core/event/base.py
```diff
@@ -611,43 +667,73 @@ def get_referred_object(self):
         Will return None if no object could be found.
         """
```
krischer: I might be overthinking or over-complicating things here, but some of logic breaks down in a threaded context and there are multiple places where race conditions could happen.

For the resource ids that might actually happen fairly often as some users download in threads (I know I do). It might be worthwhile to add a class level `Threading.RLock` and wrap some methods in

```python
def get_referred_object(...):
    with self.__class__.__lock:
        ...
```

or maybe even a decorator? I think wrapping the `__init__()`, `get_referred_object()`, and `set_referred_object()` methods would take care of most race conditions.

obspy/core/event/base.py
```diff
@@ -710,6 +796,18 @@ def copy(self):
         return deepcopy(self)
 
     @property
+    def object_id(self):
```
krischer: Can you add a leading underscore and designate it as private? I see no use to have this as a public part of the API and it frees us to make chances to it in the future if needed or desired.

obspy/core/event/base.py
d-chambers: Ok, good point here but I am not sure how to do this correctly. In order to make this thread-safe(ish) I think the lock would need to be on read_events, and Catalog.copy as well wouldn't it? 

For example, if the following 3 things don't happen for each event before any occur for another event object with the same resource_ids, the "event scoped" behavior would be ruined:
1. Each non-resource_id object is created
2. The resource_id of each of the newly created objects calls set_referred_object 
3.  The _object_id attr of each any resource_ids not named "resource_id" is cleared
4. bind_resource_ids gets called 

If I add a lock on read_events and Catalog.copy wouldn't it drastically reduce the benefit you get from multi-threading for solving the small corner case of duplicate resource_ids? If we do create a decorator for thread locking, where should such a decorator live, in obspy.util.decorator? 

I dont understand how the race condition is created with the ResourceIdentifier methods you mentioned, although there probably is one I am just not seeing, as this is getting rather complicated. Thoughts?

krischer: > Ok, good point here but I am not sure how to do this correctly. In order to make this thread-safe(ish) I think the lock would need to be on read_events, and Catalog.copy as well wouldn't it?

Yea - true thread safety with all references working for reading events would require that I think.

> If I add a lock on read_events and Catalog.copy wouldn't it drastically reduce the benefit you get from multi-threading for solving the small corner case of duplicate resource_ids? 

The performance will probably not deteriorate a lot (if at all) as the locking would just be around the actual event parsing routines. The downloading would still be threaded and this is likely the slow part. With CPython's GIL even most threaded things don't truly run in parallel (I/O or downloading is an exception).

> If we do create a decorator for thread locking, where should such a decorator live, in obspy.util.decorator?

Yes. I'm not sure if the effort can be justified though as we are currently just talking about the `obspy.core.event.catalog._read()` and the two copy methods - a `with` statement in these would suffice.

> I dont understand how the race condition is created with the ResourceIdentifier methods you mentioned, although there probably is one I am just not seeing, as this is getting rather complicated.

Locking the higher level function you proposed is much cleaner in any case. I'm too lazy now to try to re-follow my own thoughts about locking these three methods...I think it was related to the fact that all the class level dictionaries are not thread local and with all the required shuffling of object ids and references it might result in some non-deterministic race conditions. But as you said: its getting fairly complicated and I might be mistaken.

All in all this deals with a rare corner case but then its only 10 lines or so to do it properly so why not I guess?

krischer:
Thanks a lot for all the work!

I added another commit that makes sure the lock is created only once per function. The previous decorator created the lock at each function call which would result in no actual locking at all as each thread would just create its own lock. I think it now is correct but double checking does not hurt.

Otherwise this look IMHO pretty ready to merge.

+DOCS

d-chambers:
Great, thanks for your help and reviews.

I just realized the rlock should be on read_events not on the init method on the AbstractEventType, very small change.

krischer:
Great. Looks really good now. I had to rebase to get rid of a merge conflict. Thanks once again!

megies:
Reimplemented in #1702

PR #1647, opened 2017-01-21
Flake8 3 only

krischer:
This PR results in code formatting checks being only run for flake8 > 3. The API changes between v2 and the various v3 versions are getting a bit too wild. This is also only an optional check so it should be fine as long as its executed in a couple of test runners.

Also fixed a bunch of formatting things throughout the code base.


obspy/core/tests/test_code_formatting.py
```diff
@@ -22,16 +47,32 @@ class CodeFormattingTestCase(unittest.TestCase):
...
+                continue
+            for pattern in FLAKE8_EXCLUDE_FILES:
+                if fnmatch.fnmatch(filename, pattern):
+                    break
+            else:
+                files.append(filename)
+        report = style_guide.check_files(files)
+
+        # Make sure no error occured.
+        assert report.total_errors == 0
```
megies: @krischer @mbyt since merging this PR we do not get any information on flake8 fails anymore, due to the plain "assert" without failure message, see e.g. https://travis-ci.org/obspy/obspy/jobs/202193359#L1159

screenshot from 2017-02-16 13 41 17

Any reason for this?

krischer:
You can still see them in the travis output - they are just a bit hidden:

screen shot 2017-02-21 at 09 48 08

The main reason was that I got fed up with keeping up with all the API changes (there is version 2, at least two different APIs in version 3, and there is actually no final API in version 3 yet...) in flake8 to parse out the errors and provide these nice error messages. IMHO not worth it especially as it can still be seen in the logs.

megies:
Ah, alright, our testrun stdout has become really polluted again, didn't see there was flake8 information in there. Thanks for pointing out..

PR #1648, opened 2017-01-24
Fix bytes to str compare with bytes.decode()

danauerbach:
This PR fixes a problem setting the miniseed fixed header flags when using Python 3. They are not getting set due to the fact that the Trace ID matching logic is comparing bytes with str, which will fail  in Python 3.

This PR also fixes the test that failed to detect this problem due to a similar problem in the testing code. No miniseed records were actually getting tested (python 3 only) because the requested Trace IDs were not matching those in the test file due to bytes to str comparisons.

I did not see an existing issue that seemed related, so I don't believe this affects any open issues.

I first fixed the broken test to confirm that it would fail, which it did.

```
FAIL: test_set_flags_in_fixed_header (obspy.io.mseed.tests.test_mseed_util.MSEEDUtilTestCase)
  File "/Users/dauerbach/anaconda3/envs/idadcc/lib/python3.5/site-packages/obspy/io/mseed/tests/test_mseed_util.py", line 681, in test_set_flags_in_fixed_header
    512)
  File "/Users/dauerbach/anaconda3/envs/idadcc/lib/python3.5/site-packages/obspy/io/mseed/tests/test_mseed_util.py", line 832, in _check_values
    "Expected bytes")
AssertionError: b'\x15(\x88' != b'\x05(\x88' : Expected bytes

Ran 1359 tests in 208.915s

FAILED (failures=1)
```

After fixing `_check_values()` all tests passed and the header flags were getting set.

I do not have a Python 2 env set up and have only tested in Python 3.5 (Conda installed, OS X 10.12).

Hopefully I haven't missed anything... :)

Cheers,
Dan

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -21,6 +21,8 @@
    * ObsPy can now read MiniSEED files with no blockette 1000. (see #1544)
    * ObsPy now always writes Blockette 100 if sampling rate accuracy is
      otherwise lost. (see #1550)
+   * obspy.io.mseed.util.set_flags_in_fixed_header() now works with Python 3
+     and also for files with Blockette 100.
  - obspy.io.quakeml:
    * write StationMagnitude.residual even when it is zero (see #1625)
    * read & write Event.region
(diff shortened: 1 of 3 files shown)


krischer:
I can confirm your findings and that the fix works as intended on Python 2 and Python 3.

I also found another bug with the flag setting function: It did not work for files with blockette 100. I added two more commits to this PR fixing it and also adding a test.

Thanks a bunch for it!

danauerbach:
Excellent, looking forward to 1.0.3! 

:)

PR #1649, opened 2017-01-25
Update SC3ML version to 0.9 in seiscomp event module

Brtle:
A new SC3ML schema is out (http://geofon.gfz-potsdam.de/schema/0.9/). There is no change for the event module except the version number. I think it should be better to write sc3ml file in version 0.9 even if it doesn't change anything. A priori, sc3ml 0.9 can be imported into a 0.8 database, so I don't think it can break existing codes.


Brtle:
They added the node `Config/module/station/creationInfo`. It isn't under the `EventParameters` node so it doesn't concern the event module.

PR #1650, opened 2017-01-25
Fix Trace splitting with empty masked array

Brtle:
This is a little PR to fix the Trace splitting when there is no data in the masked array.

```
  File "/home/toto/.virtualenvs/obspy/lib/python3.5/site-packages/obspy/core/tests/test_trace.py", line 1499, in test_split_empty_masked_array
    st = trace.split()
  File "<decorator-gen-24>", line 2, in split
  File "/home/toto/.virtualenvs/obspy/lib/python3.5/site-packages/obspy/core/trace.py", line 231, in _add_processing_info
    result = func(*args, **kwargs)
  File "/home/toto/.virtualenvs/obspy/lib/python3.5/site-packages/obspy/core/trace.py", line 2227, in split
    for slice in slices:
TypeError: 'NoneType' object is not iterable
```

The function `obspy.core.util.misc.flat_not_masked_contiguous` returns `None` instead of an empty list when there is no data, raising a `TypeError` in  `obspy.core.trace.Trace.split`.

I added a unit test to cover this case.

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1455,6 +1455,22 @@ class TraceTestCase(unittest.TestCase):
         self.assertFalse(isinstance(st[0].data, np.ma.masked_array))
         self.assertFalse(isinstance(st[1].data, np.ma.masked_array))
 
+    def test_split_empty_masked_array(self):
+        """
+        Test split method with a masked array without any data.
+        """
+        tr1 = Trace(data=np.arange(1000))
+        tr2 = Trace(data=np.arange(1000, 2000))
+        trace = tr1 + tr2
+
+        self.assertTrue(isinstance(trace.data, np.ma.masked_array))
+        self.assertTrue(isinstance(trace, Trace))
+
+        st = trace.split()
+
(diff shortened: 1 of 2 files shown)


obspy/core/tests/test_trace.py
```diff
@@ -1485,6 +1485,22 @@ def test_split(self):
         self.assertFalse(isinstance(st[0].data, np.ma.masked_array))
         self.assertFalse(isinstance(st[1].data, np.ma.masked_array))
 
+    def test_split_empty_masked_array(self):
+        """
+        Test split method with a masked array without any data.
+        """
+        tr1 = Trace(data=np.arange(1000))
+        tr2 = Trace(data=np.arange(1000, 2000))
+        trace = tr1 + tr2
```
krischer: It would IMHO be clearer to directly initialize a Trace with a fully masked array, e.g.

```python
tr = Trace(data=np.ma.masked_all(100))
```

Brtle:
I rebased onto the `maintenance_1.0.x` branch, sorry for that.

krischer:
Thanks! Looks good now. I'll merge once CI has run.

PR #1651, opened 2017-01-26
Fix wildcard exceptions handling

Brtle:
There is a lot of 

```
try:
    …
except:
    …
```

in obspy (nearly 300) and this can lead to not correctly catch exceptions like `KeyboardInterrupt` or `SystemExit`. I didn't have faith to fix each of them perfectly by catching only the expected exceptions so I did this little PR to replace the wildcard excepts by:

```
try:
    …
except Exception:
    …
```

At least, system-exiting exceptions won't be caught.

### DIFF
--- a/misc/docs/make_pep8.py
+++ b/misc/docs/make_pep8.py
@@ -19,7 +19,7 @@ path = obspy.__path__[0]
 
 try:
     os.makedirs(os.path.join('source', 'pep8'))
-except:
+except Exception:
     pass
 
 report, message = check_flake8()
@@ -85,7 +85,7 @@ with open(os.path.join('source', 'pep8', 'index.rst'), 'wt') as fh:
 # remove any old image
 try:
     os.remove(PEP8_IMAGE)
-except:
+except Exception:
     pass
 # copy correct pep8 image
 if error_count > 0:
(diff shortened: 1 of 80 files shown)


krischer:
Yea this is definitely a weak part of the code base but a pretty big effort to properly fix. For most things it could indeed be replaced with more detailed exceptions and if someone feels like it: please go for it.

Others are a bit harder to specify. For example all network operations can fail at multiple levels (sockets, timeouts, http parsing, content parsing, ...) with correspondig exceptions so its probably not worth it to wade through all that.

I would tend to merge this now and add a new issue to look into this in more detail. Thoughts?

markcwill:
Agree with @krischer -- although I thought in 2.7 that a blank `except` defaulted to `Exception` and raised e.g. `KeyboardInterrupt`... guess I was wrong! Which means there's no difference between catching `StandardError` and `Exception` in Python 2? (Answer: True, looks like StandardError is not in Python 3)

It would be nice to have an issue for this, and I'll throw up the suggestion that there should be a module logger (or perhaps even just one package-wide) that any "bare" `Exception` clause like this that just has a `pass` should at least log the error/tb to null:
```python
# create or import this logger from obspy.util
import logging
errlog = logging.getLogger('obspy.warnings')
errlog.addHandler(logging.NullHandler())

try:
    foo()
except Exception as e:
    errlog.exception(e)
```
This way one can just attach a handler to the "obspy.warnings" log for debugging in your own script.

PR #1655, opened 2017-01-31
DeprecationWarning: invalid escape sequence

krischer:
for our regex specialists:
```
D:\Workspace\obspy\obspy\io\seiscomp\sc3ml.py:70: DeprecationWarning: invalid escape sequence \.
  '[0-9]*\.?[0-9]+}', root.tag)
```


krischer:
This is now a PR which should fix it.

PR #1659, opened 2017-02-06
Konno-Ohmachi: function to apply smoothing matrix

claudiodsf:
A new function (`apply_smoothing_matrix()`) to smooth spectra through a pre-computed smoothing matrix.
This is useful to speed-up calculation on the same type of spectra across different function calls.

### DIFF
--- a/obspy/signal/konnoohmachismoothing.py
+++ b/obspy/signal/konnoohmachismoothing.py
@@ -82,8 +82,8 @@ def konno_ohmachi_smoothing_window(frequencies, center_frequency,
         # Calculate the bandwidth*log10(f/f_c)
         smoothing_window = bandwidth * np.log10(frequencies / center_frequency)
         # Just the Konno-Ohmachi formulae.
-        smoothing_window[...] =\
-            (np.sin(smoothing_window) / smoothing_window) ** 4
+        smoothing_window[:] = (
+            np.sin(smoothing_window) / smoothing_window) ** 4
     # Check if the center frequency is exactly part of the provided
     # frequencies. This will result in a division by 0. The limit of f->f_c is
     # one.
@@ -227,14 +227,10 @@ def konno_ohmachi_smoothing(spectra, frequencies, bandwidth=40, count=1,
     # matrix and apply to each spectrum. Also only use when more then one
     # spectrum is to be smoothed.
     if enforce_no_matrix is False and (len(spectra.shape) > 1 or count > 1) \
-       and approx_mem_usage < max_memory_usage:
+            and approx_mem_usage < max_memory_usage:
         smoothing_matrix = calculate_smoothing_matrix(
             frequencies, bandwidth, normalize=normalize)
-        new_spec = np.dot(spectra, smoothing_matrix)
-        # Eventually apply more than once.
-        for _i in range(count - 1):
-            new_spec = np.dot(new_spec, smoothing_matrix)
-        return new_spec
+        return apply_smoothing_matrix(spectra, smoothing_matrix, count=count)
     # Otherwise just calculate the smoothing window every time and apply it.
     else:
         new_spec = np.empty(spectra.shape, spectra.dtype)
--- a/obspy/signal/konnoohmachismoothing.py
+++ b/obspy/signal/konnoohmachismoothing.py
@@ -103,9 +103,8 @@ def calculate_smoothing_matrix(frequencies, bandwidth=40.0, normalize=False):
     Ohmachi window for each frequency as the center frequency.
 
     Any spectrum with the same frequency bins as this matrix can later be
-    smoothed by a simple matrix multiplication with this matrix::
-
-        smoothed_spectrum = np.dot(spectrum, smoothing_matrix)
+    smoothed by using
+    :func:`~obspy.signal.konnoohmachismoothing.apply_smoothing_matrix`.
 
     This also works for many spectra stored in one large matrix and is even
     more efficient.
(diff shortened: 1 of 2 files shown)


claudiodsf:
+DOCS please

claudiodsf:
The doc changes are for [this page](http://docs.obspy.org/pull-requests/1659/packages/autogen/obspy.signal.konnoohmachismoothing.konno_ohmachi_smoothing.html#obspy.signal.konnoohmachismoothing.konno_ohmachi_smoothing).

obspy/signal/tests/test_konnoohmachi.py
```diff
@@ -134,19 +134,25 @@ def test_konno_ohmachi_smoothing(self):
                                              max_memory_usage=0)
         # XXX: Why are the numerical inaccuracies quite large?
         np.testing.assert_almost_equal(smoothed_1, smoothed_2, 3)
-        # Test the non-matrix mode for single spectra.
+        # Test using a pre-computed smoothing matrix
+        with np.errstate(all='ignore'):
```
krischer: I personally don't like this too much but it is a consequence of my own old code so I cannot complain too much ;-) Maybe it would be possible to explicitly avoid taking the logarithm of zero and also the division by zero in the `konno_ohmachi_smoothing_window()` function so if completes without raising numpy warnings?

claudiodsf:
@krischer I've partially address that by moving it into the smoothing window code. Let me know if that's ok for you. 
Otherwise, if tests pass (not sure about those coverage and docker errors), I would like to squash a little bit before merging.

krischer:
The coverage tests are always kinda weird - ignore these. I also don't see any docker failures (http://tests.obspy.org/?pr=1659). So - looks good to me. Let me know once you deem it ready to merge.

claudiodsf:
@megies Should I worry about those failing docker tests?

claudiodsf:
If the failing tests are not important, this is good to merge.

megies:
Hmm.. so when the smoothing matrix is specified, there's practically almost nothing of the logic of the routine left.. https://github.com/obspy/obspy/pull/1659/files#diff-029c61109d64a5325db86a879e3b0e93R196

Also this overrides basically almost every argument of the function..

I'm not using this routine much so I don't want to interfere too much, but wouldn't it probably make more sense to refactor this piece as a separate function (maybe `konnoohmachismoothing.apply_smoothing_matrix()`)?

claudiodsf:
The alternative is to not touch the `konno_ohmachi_smoothing()` routine at all and just to suggest doing

    smoothed_spectrum = np.dot(spectrum, smoothing_matrix)

which, by the way, is already suggested in the [documentation](http://docs.obspy.org/packages/autogen/obspy.signal.konnoohmachismoothing.calculate_smoothing_matrix.html#obspy.signal.konnoohmachismoothing.calculate_smoothing_matrix) of `calculate_smoothing_matrix()`.

If you agree on this second option, I would remodulate this PR to just keep the improvements which avoid using `with np.errstate(all='ignore')` all over the time and I would maybe suggest to rename `calculate_smoothing_matrix()` into `konno_ohmachi_smoothing_matrix()`.

What do you think?

megies:
I think the name of `calculate_smoothing_matrix` is appropriate (as it's inside the `konnoohmachismoothing` module) and can stay as is, I would just pack the multiplying the smoothing matrix that you added here as another function `apply_smoothing_matrix` along with the `count` option maybe.. but really, I'm not working much with this, that was just my two cents..

claudiodsf:
The name change proposal was just for consistency with `konno_ohmachi_smoothing_window()`.
Otherwise, ok for `apply_smoothing_matrix()` with the `count` option.
@krischer ?

claudiodsf:
Ok, so I followed @megies approach because it seems to me more logical.

obspy/signal/konnoohmachismoothing.py
```diff
@@ -85,10 +77,13 @@ def konno_ohmachi_smoothing_window(frequencies, center_frequency,
...
-    # Calculate the bandwidth*log10(f/f_c)
-    smoothing_window = bandwidth * np.log10(frequencies / center_frequency)
-    # Just the Konno-Ohmachi formulae.
-    smoothing_window[...] = (np.sin(smoothing_window) / smoothing_window) ** 4
+    # Disable div by zero errors and return zero instead
+    with np.errstate(divide='ignore', invalid='ignore'):
+        # Calculate the bandwidth*log10(f/f_c)
+        smoothing_window = bandwidth * np.log10(frequencies / center_frequency)
+        # Just the Konno-Ohmachi formulae.
+        smoothing_window[...] =\
```
megies: What's this? Never seen slicing with `x[...]` on an array.. is this identical with `x[:]`?

obspy/signal/konnoohmachismoothing.py
```diff
@@ -210,11 +228,8 @@ def konno_ohmachi_smoothing(spectra, frequencies, bandwidth=40, count=1,
...
-        # Disable NumPy warnings due to possible divisions by zero/logarithms
-        # of zero.
-        with np.errstate(all='ignore'):
-            smoothing_matrix = calculate_smoothing_matrix(
-                frequencies, bandwidth, normalize=normalize)
+        smoothing_matrix = calculate_smoothing_matrix(
+            frequencies, bandwidth, normalize=normalize)
         new_spec = np.dot(spectra, smoothing_matrix)
         # Eventually apply more than once.
         for _i in range(count - 1):
```
megies: Unless you want to avoid that spectra dtype check you could use the new function here, to avoid code duplication and give it a bit of logical structure..

obspy/signal/konnoohmachismoothing.py
claudiodsf: http://stackoverflow.com/questions/118370/how-do-you-use-the-ellipsis-slicing-syntax-in-python

Don't think it is necessary here. This is, anyway, a piece of code from @krischer.

obspy/signal/konnoohmachismoothing.py
claudiodsf: Ok, I'll try to do this.

obspy/signal/konnoohmachismoothing.py
megies: Uhmm.. this seems to be a pretty obscure hidden feature in Python.. any reason to use this over `[:]`, @krischer??

QuLogic: I'm not sure it's that obscure, but it's not really a Python feature in this case, but a NumPy one. It means 'all elements in all dimensions'; `:` means all elements in just the first dimension only.

megies: Thanks for clarifying. So it's really something else (in general), didn't know this one..

claudiodsf: @krischer Any comments on this (or other things)? If you're ok, I'll implement the last suggestion from @megies and try to merge this PR.

megies:
@claudiodsf would you mind rebasing on master again, to get fresh CI results? But I think this is good to go..

megies:
@claudiodsf please add those minor changes I mentioned and rebase on master within the next days for inclusion in 1.1.0 if you can.. :-)

obspy/signal/konnoohmachismoothing.py
krischer: Yea - its probably not necessary and if it confuses people feel free to replace it.

krischer:
Looks pretty good to me. Good to go after the chances requested by @megies have been implemented.

megies:
Implemented my minor comments and rebased on current master. Good to go, I think, if CI gives the nod.

claudiodsf:
Great! Sorry for no reaction. Really busy in this period.

PR #1664, opened 2017-02-10
PICKLE format broken / seishub client broken (unpickling of new nanosecond based utcdatetime)

megies:
I think #1325 broke `clients.seishub`, because data is sent as pickled streams from the server and the data from the server is lacking `_ns` attribute then...

I can't see a way to fix this right now, as we can't make sure seishub server and client are on the same page UTCDateTime-wise..
```
(master) ~$ py.test ~/git/obspy-master/obspy/clients/seishub -k test_get_waveform_with_metadata
====================================================== test session starts =======================================================
platform linux2 -- Python 2.7.12, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /home/megies, inifile: pytest.ini
plugins: random-0.2
collected 16 items 

git/obspy-master/obspy/clients/seishub/tests/test_client.py F

============================================================ FAILURES ============================================================
_________________________________________ ClientTestCase.test_get_waveform_with_metadata _________________________________________

self = <obspy.clients.seishub.tests.test_client.ClientTestCase testMethod=test_get_waveform_with_metadata>

    def test_get_waveform_with_metadata(self):
        # metadata change during t1 -> t2 !
        t1 = UTCDateTime("2010-05-03T23:59:30")
        t2 = UTCDateTime("2010-05-04T00:00:30")
        client = self.client
        self.assertRaises(Exception, client.waveform.get_waveforms, "BW",
                          "UH1", "", "EH*", t1, t2, get_paz=True,
                          get_coordinates=True)
        st = client.waveform.get_waveforms("BW", "UH1", "", "EH*", t1, t2,
                                           get_paz=True, get_coordinates=True,
>                                          metadata_timecheck=False)

git/obspy-master/obspy/clients/seishub/tests/test_client.py:235: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
git/obspy-master/obspy/clients/seishub/client.py:505: in get_waveforms
    stream = _unpickle(data)
git/obspy-master/obspy/clients/seishub/client.py:49: in _unpickle
    obj = pickle.loads(data)
anaconda/envs/master/lib/python2.7/pickle.py:1388: in loads
    return Unpickler(file).load()
anaconda/envs/master/lib/python2.7/pickle.py:864: in load
    dispatch[key](self)
anaconda/envs/master/lib/python2.7/pickle.py:1223: in load_build
    setstate(state)
git/obspy-master/obspy/core/util/attribdict.py:115: in __setstate__
    self.update(adict)
git/obspy-master/obspy/core/util/attribdict.py:142: in update
    self.__setitem__(key, value)
git/obspy-master/obspy/core/trace.py:165: in __setitem__
    value = UTCDateTime(value)
git/obspy-master/obspy/core/utcdatetime.py:231: in __init__
    self._ns = value._ns
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'UTCDateTime' object has no attribute '_UTCDateTime__ns'") raised in repr()] SafeRepr object at 0x7fdcb2143710>

    def _get_ns(self):
>       return self.__ns
E       AttributeError: 'UTCDateTime' object has no attribute '_UTCDateTime__ns'

git/obspy-master/obspy/core/utcdatetime.py:360: AttributeError
=================================== 15 tests deselected by '-ktest_get_waveform_with_metadata' ===================================
============================================ 1 failed, 15 deselected in 1.65 seconds =============================================
```


barsch:
maybe issue with the int(round()) construct - can you debug it using a break point?

megies:
Let's wait a bit with merging this, I need to do more checks..

megies:
Ermmh.. today doctests get executed again for me.. (no idea why they weren't yesterday).

There really seems to be a problem with a one second offset with this PR:
http://tests.obspy.org/69881/

barsch:
```self._from_timestamp``` uses also ```int(round(value * 10**9))``` - could it be a rounding issue - unfortunatly I can't really debug it ...

**Comment by megies** [comment]
> self._from_timestamp uses also int(round(value * 10**9))

yeah, I saw that, just used that one to not duplicate code.. need to debug it when I got some time..

**Comment by megies** [comment]
Ok, so after fixing the unpickling, there is a completely separate issue. I'm a bit confused right now, it seems like it's a rounding issue, I can reproduce using `Trace.trim()`..

```python
import numpy as np
from obspy import UTCDateTime, Trace

t = UTCDateTime()
try:
    t._from_timestamp(1484308799.9749999046)
    label = 'new'
except:
    t = UTCDateTime(1484308799.9749999046)
    label = 'old'

x = np.arange(15, dtype=np.int32)
x[::5] = 0
tr = Trace(x)
tr.stats.sampling_rate = 200
tr.stats.starttime = t

t2 = UTCDateTime(2017, 1, 13, 12)

print tr

tr.trim(starttime=t2)

print t2
print tr
print '{:.9f}'.format(tr.stats.starttime.timestamp)

tr.stats.station = label
tr.write('/tmp/utcdatetime_trim_bug_{}.mseed'.format(label), 'MSEED')
```

On maintenance_1.0.x:
```
... | 2017-01-13T11:59:59.975000Z - 2017-01-13T12:00:00.045000Z | 200.0 Hz, 15 samples
2017-01-13T12:00:00.000000Z
... | 2017-01-13T12:00:00.000000Z - 2017-01-13T12:00:00.045000Z | 200.0 Hz, 10 samples
1484308800.000000000
```

On master:
```
... | 2017-01-13T11:59:59.975000Z - 2017-01-13T12:00:00.045000Z | 200.0 Hz, 15 samples
2017-01-13T12:00:00.000000Z
... | 2017-01-13T11:59:59.000000Z - 2017-01-13T12:00:00.045000Z | 200.0 Hz, 10 samples
1484308799.999999762
```

The timestamps that are output seem to differ by 0.3 microseconds but printing they differ by one second straight?

barsch:
master:
```
>>> UTCDateTime(2017, 1, 13, 11, 59, 59, 975000) + 0.005
UTCDateTime(2017, 1, 13, 11, 59, 59, 980000)
>>> UTCDateTime(1484308799.975) + 0.005
UTCDateTime(2017, 1, 13, 11, 59, 59, 979999)
```
so never trust timestamps ...

barsch:
we should use as fallback ```self._from_datetime``` instead of ```self._from_timestamp``` if the _ns attribute is missing - it should work than at least in this case

generally usage of ```UTCDateTime.timestamp``` should be discouraged as its not accurate enough ...

megies:
> we should use as fallback self._from_datetime instead of self._from_timestamp if the _ns attribute is missing - it should work than at least in this case

uhm.. old utcdatetime objects serve everything derived from the timestamp.. the timestamp is the only attribute that exists when an old utcdatetime object is unpickled

barsch:
```self._fom_datetime(datetime.datetime.fromtimestamp(value))``` works in this case - however it won't work with for timestamps before < 1970 and >2038 again ...

**Comment by megies** [comment]
> however it won't work with for timestamps before < 1970 and >2038 again ...
> 
> but as fallback it should be fine?

I guess that's fine, because it's supposed to not have worked before either..

**Comment by megies** [comment]
seishub tests are failing: http://tests.obspy.org/70224/

(I added a missing test for fetching waveforms today)

**Comment by megies** [comment]
I've added a test that shows the issue without access to a seishub server. Our core PICKLE format is broken. Reading files that were written on obspy 1.0.3 using a current master results in wrong timing information.

File: obspy/core/utcdatetime.py [hunk 1]
```diff
@@ -233,7 +233,7 @@ def __init__(self, *args, **kwargs):
                 try:
                     self._ns = value._ns
                 except AttributeError:
-                    self._from_datetime(datetime.datetime.fromtimestamp(
+                    self._from_datetime(datetime.datetime.utcfromtimestamp(
```
megies: Explains a bit the erratic passing/failing of tests (on local computer vs. CI) as the previous code depended on local computer time setting (UTC vs. some local time zone).. nice catch @barsch!

megies:
OK, so.. the timing problem is fixed, but it turns out we can not read a pickle on Py3 when it was written on Py2. (The other way around works)

Actually, I am not sure we ever were capable of that, as we only had a round-trip write-read test for PICKLE format.

krischer:
Regarding py2/py3: https://docs.python.org/3/library/pickle.html#pickle.load

The load method contains a couple options that might enable that.

Pickling is AFAIK guaranteed to be backwards compatible but I'm not sure if that's valid across python 2 and 3.

barsch:
@krischer: I already played with those - don't have all tracebacks anymore - but here some excerpts from chat with Tobi

```encoding='utf-8'``` or ```encoding='latin1'```  did not work - but with ```pickle.load(filename, encoding="raw-unicode-escape")``` I got:

``` 
  File "D:\Workspace\obspy\obspy\core\tests\test_stream.py", line 1335, in test_read_pickle
    st = read(pickle_file, format='PICKLE')
  File "<decorator-gen-32>", line 2, in read
  File "D:\Workspace\obspy\obspy\core\util\decorator.py", line 299, in _map_example_filename
    return func(*args, **kwargs)
  File "D:\Workspace\obspy\obspy\core\stream.py", line 231, in read
    st.extend(_read(file, format, headonly, **kwargs).traces)
  File "<decorator-gen-33>", line 2, in _read
  File "D:\Workspace\obspy\obspy\core\util\decorator.py", line 209, in uncompress_file
  File "D:\Workspace\obspy\obspy\core\stream.py", line 273, in _read
    headonly=headonly, **kwargs)
  File "D:\Workspace\obspy\obspy\core\util\base.py", line 463, in _read_from_plugin
  File "D:\Workspace\obspy\obspy\core\stream.py", line 3148, in _read_pickle
    return pickle.load(fp, errors="ignore")
ValueError: buffer size does not match array size
```

with ```pickle.load(filename, encoding="raw-unicode-escape") ``` -> maxrecursion error


```
File "D:\Python\virtualenv\obspy\64\Python34\lib\site-packages\future\types\newint.py", line 126, in __mul__
    return long(self) * other
  File "D:\Python\virtualenv\obspy\64\Python34\lib\site-packages\future\types\newint.py", line 126, in __mul__
    return long(self) * other
  File "D:\Python\virtualenv\obspy\64\Python34\lib\site-packages\future\types\newint.py", line 126, in __mul__
    return long(self) * other
  File "D:\Python\virtualenv\obspy\64\Python34\lib\site-packages\future\types\newint.py", line 126, in __mul__
    return long(self) * other
  File "D:\Python\virtualenv\obspy\64\Python34\lib\site-packages\future\types\newint.py", line 126, in __mul__
    return long(self) * other
```

and

```
DeprecationWarning: __int__ returned non-int (type newint).  The ability to return an instance of a strict subclass of int is deprecated, and may be removed in a future version of Python.
  value = int(value)
```

barsch:
while working on the vcr stuff I made pretty much the same experience - py2 pickles couldn't be loaded into py3 but the other way around - no idea why - thats the reason the "vcr recording" part is restricted to py3 only

krischer:
Hmm..really strange to be honest but I don't have any experience in those things so who knows.

So the solution is to port SeisHub to Python 3? ;-)

megies:
> So the solution is to port SeisHub to Python 3? ;-)

;-)  I think it's OK if we explicitly raise an exception when trying to fetch data from a seishub server, stating that it only works on Py2.

barsch:
SeisHub should better be replaced with https://github.com/krischer/jane ;)

megies:
seishub and pickle tests are running fine now locally, on py27 and py36:
 - http://tests.obspy.org/73197/
 - http://tests.obspy.org/73198/

Looks like this is resolved now.. :-)

megies:
Huhh.. don't understand that recursion error.. on Python 3, and it's not happening (locally) on Python 3.6..

krischer:
Huii...looks like a future error. Do you have the same future version installed?

megies:
I also thought it's future but I ended up trying future 0.16.0 on Py3.5 and also had that failure..

megies:
OK, this fixes it for me..

megies:
OK, so hopefully this is the last fix needed here.. (Python 3.3 Travis shows some floating point accuracy/rounding issue when converting old pickled UTCDateTime object to new integer nanosecond one..)

barsch:
looking good!

PR #1670, opened 2017-02-13
Comparing UTCDateTime with other objects (float, int, UTCDateTime, datetime.datetime)

### LINKED ISSUE #1652: Error when comparing obspy.core.utcdatetime.UTCDateTime and datetime.datetime objects
It seems that comparing an UTCDateTime and a datetime object always return False, no matter the actual date values. The following was tested on Linux Redhat 6.4 with Python 2.7.10 and obspy 1.0.2:

```
>>> from obspy.core.utcdatetime import UTCDateTime
>>> from datetime import datetime
>>> a = datetime(1990, 1, 1, 0, 0, 0)
>>> e = UTCDateTime(2000, 1, 2, 1, 39, 37)
>>>
>>> a < e
False
>>> e < a
False
>>> a <= e
False
>>> e <= a
False
>>> a > e
False
>>> e > a
False
>>> a >= e
False
>>> e >= a
False
>>> a == e
False
>>> e == a
False
```

barsch:
see #1652 - those isinstance checks seem to influence performance not very much - no change in execution time for all core tests ...

### DIFF
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1104,6 +1104,23 @@ class UTCDateTimeTestCase(unittest.TestCase):
         self.assertEqual(str(UTCDateTime("9999-12-31T23:59:59.999999")),
                          "9999-12-31T23:59:59.999999Z")
 
+    def test_1652(self):
+        """
+        Comparing UTCDateTime and datetime.datetime objects - see #1652
+        """
+        a = datetime.datetime(1990, 1, 1, 0, 0)
+        e = UTCDateTime(2000, 1, 2, 1, 39, 37)
+        self.assertTrue(a < e)
+        self.assertFalse(a > e)
+        self.assertTrue(a <= e)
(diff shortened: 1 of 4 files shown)


barsch:
why does it say: docker-testbot — Docker tests failed while all is green in details ?

megies:
> why does it say: "docker-testbot — Docker tests failed" while all is green in details ?

The commit status is determined by the combined exit code of all docker containers while that link is just (trying to) match all of the sent reports from inside the containers. There's way less reports than there should be, could be test execution didn't get to the point of sending a failure report, or something else entirely went wrong. Hard to say.

There's no way to tell the docker bot to test that commit again (since there is already a commit status of the respective context -- and commit statuses can not be deleted), the only way to get a new build would be to push a new commit (can be an empty dummy commit of course).

megies:
Rebased on current master and force-pushed. This should get rid of unrelated CI fails..

barsch:
CI passes - merging

PR #1674, opened 2017-02-15
Fixes and extends non-functional io.mseed test

### LINKED ISSUE #1555: Bad mseed test case
Just stumbled over a bad mseed test case in `obspy/io/mseed/tests/test_mseed_special_issues.py`:

``` python
    def test_wrong_record_length_as_argument(self):
        """ 
        Specifying a wrong record length should raise an error.
        """
        file = os.path.join(self.path, 'data', 'libmseed',
                            'float32_Float32_bigEndian.mseed')
        self.assertRaises(Exception, read, file, reclen=4096)
```

The exception is raised because the file is in data/encoding and the path is wrong, so the whole test case is bogus.. should be fixed.. ;-)

If I put the right path the following exception is raised which is also pretty strange because the file is there and can be opened..

```
                # XXX: Might cause problems if the data is faulty and the user
                # set start/end time. Not sure what to do in this case.
                elif not starttime and not endtime:
>                   raise Exception("Cannot open file/files: %s" % pathname)
E                   Exception: Cannot open file/files: /home/megies/git/obspy-maintenance_1.0.x/obspy/io/mseed/tests/data/encoding/float32_Float32_bigEndian.mseed

../../../core/stream.py:243: Exception
```

barsch:
fixes #1555

### DIFF
--- a/obspy/io/mseed/tests/test_mseed_special_issues.py
+++ b/obspy/io/mseed/tests/test_mseed_special_issues.py
@@ -276,9 +276,18 @@ class MSEEDSpecialIssueTestCase(unittest.TestCase):
         """
         Specifying a wrong record length should raise an error.
         """
-        file = os.path.join(self.path, 'data', 'libmseed',
+        file = os.path.join(self.path, 'data', 'encoding',
                             'float32_Float32_bigEndian.mseed')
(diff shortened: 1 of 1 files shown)


krischer:
Looks good. The CI failures are related to some errors in the master.

PR #1672, opened 2017-02-16
test_code_formatting fails with ```from cryptography.hazmat.backends import default_backend``` 

megies:
see http://tests.obspy.org/70204/#1

```
File "/obspy/obspy/core/tests/test_code_formatting.py", line 205, in test_no_pyplot_import_in_any_file
self.assertEqual(len(failures), 0, "\n" + "\n\n".join(failures))
AssertionError:
File '/obspy/obspy/clients/arclink/decrypt.py' (line 31)
matches a forbidden matplotlib import statement outside of class/def statements
(breaking matplotlib backend switching on some systems):
' from cryptography.hazmat.backends import default_backend'
File '/obspy/obspy/clients/arclink/decrypt.py' (line 31)
matches a forbidden matplotlib import statement outside of class/def statements
(breaking matplotlib backend switching on some systems):
' from cryptography.hazmat.backends import default_backend'
```

This error is essentially blocking PR https://github.com/obspy/obspy/pull/1360

### DIFF
--- a/obspy/core/tests/test_code_formatting.py
+++ b/obspy/core/tests/test_code_formatting.py
@@ -158,17 +158,75 @@ class FutureUsageTestCase(unittest.TestCase):
 
 
 class MatplotlibBackendUsageTestCase(unittest.TestCase):
+    patterns = (
+        r" *from pylab import",
+        r" *from pylab\..*? import",
+        r" *import pylab",
+        r" *from matplotlib import (.*?, *)*(pyplot|backends)",
+        r" *import matplotlib\.(pyplot|backends)")
+
+    def forbidden_match(self, line):
+        for pattern in self.patterns:
+            if re.match(pattern, line):
+                return pattern
+        return False
+
+    def test_no_pyplot_regex(self):
+        """
+        Tests that the regex patterns to match forbidden lines works
+        as expected.
+        """
+        positives = (
+            'from pylab import something',
+            '    from pylab import something',
+            'from pylab.something import something',
+            'import pylab',
+            '    import pylab',
+            'from matplotlib import pyplot',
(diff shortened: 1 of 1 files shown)


megies:
This is now a PR that should fix the regex (and I've added a test to make sure the regex is matching the expected strings).

megies:
No idea what flake8 is complaining about, since #1647 removed any failure messages from our flake8 test case.

krischer:
screen shot 2017-02-21 at 09 54 02

;-)

megies:
Those two pep8 fails are false positives from an issue in pep8 that was first reported over four years ago:

```python
179         positives = (
180             'from pylab import something',
181             '    from pylab import something',
182             'from pylab.something import something',
183             'import pylab',
184             '    import pylab',
185             'from matplotlib import pyplot',
186             '  from matplotlib import pyplot',
187             'from matplotlib import backends',
188             '  from matplotlib import backends',
189             'from matplotlib import dates, backends',
190             'import matplotlib.pyplot as plt',
191             'import matplotlib.pyplot',
192             'import matplotlib.backends',
193             '   import matplotlib.backends',
194             )
```

The indentation of the closing bracket is totally valid. [see pep8, 'indentation' at end of paragraph](https://www.python.org/dev/peps/pep-0008/#indentation)

I'm confused because I don't see any pep8 version installed in the test runner.. (supposedly it's fixed in pep8 1.4.6 in 2014)

megies:

megies:
I think this is good to go..

 - flake8 check has two false positives, but this has to be fixed upstream or by using a different flake8/pep8 version
 - some docker testbots could not run due to #1671 

Merging now, to not delay @barsch's #1360 any longer.

PR #1682, opened 2017-02-23
Removing make_pep8.py

krischer:
It is no longer needed.

+DOCS


misc/docs/source/_templates/index.html
```diff
@@ -106,16 +106,4 @@ <h1 class="hidden-xs">
     </div>
   </div>
 
-  <div class="row">
-    <div class="col-sm-4 col-sm-offset-1">
-      <p class="biglink"><a class="biglink" href="pep8/index.html">PEP 8</a><br>
-         <span class="linkdescr">Don't use tabs!</span></p>
-    </div>
-    <div class="col-sm-4 col-sm-offset-1">
-      <p class="biglink">Coverage <a class="biglink" href="https://coveralls.io/r/obspy/obspy?branch=master">Python</a> /
```
megies: Instead of removing all coverage links, we could link to [codecov](https://codecov.io/github/obspy/obspy) now maybe.. but I'm also OK if this gets scrapped entirely.

krischer: Yea but then we would have to remember to change it again next time we use a different coverage service. I'm also fine with either but would tend to leave this as it is.

megies: OK, fine!

PR #1685, opened 2017-03-01
fdsn/get_stations_bulk cross-implemented with fdsn/get_waveforms_bulk

krischer:
Hi, this is something I noticed while doing background research for the IRIS federator project. It persists in the current master branch, and it looks like it is in the maintenance branch, too.

The documentation for **obspy.clients.fdsn.client.Client.get_stations_bulk** contains the parameter list for `get_waveforms_bulk`, such as `quality`, `minimumlength`, `longestonly`, and `attach_response`.

Also, the function itself checks for the existence of "dataselect" in services instead of "station".

Cheers,
Celso


krischer:
This issue is now a PR fixing it.

PR #1688, opened 2017-03-01
Run dedicated code formatting check in CI

krischer:
Should ease discovery of code formatting changes. Code formatting checks should of course be disabled in all other runners.

See discussion here: https://github.com/obspy/obspy/commit/c3120f234392909121f3694b5a12d6262c25c9f9#commitcomment-21103422

Maybe we could even run this on a different CI service? Then it would be very easy to tell if a failure is code formatting related. Maybe on CircleCI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -192,9 +192,9 @@ script:
   - export IFS=" "
   # if not on a PR, i.e. running on a push to master or maintenance_XXX: test *all* modules
   - if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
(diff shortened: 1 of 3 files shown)


krischer:
Gave it a shot. The extra config file is really short and IMHO totally worth it. Its nothing compared to the complexity of the other test running scripts. Also the CircleCI run is done in like 30 seconds which is nice.

Travis and appveyor now no longer install flake8 and the code formatting tests should thus be skipped.

megies:
> the code formatting tests should thus be skipped.

Why not skip them explicitly?  `--no-flake8`

megies:
Same problem as in #1686: command line option `--ignore`  doesn't append to default ignores but replaces them.. therefore the CircleCI flake8 fail..

krischer:
> Why not skip them explicitly? --no-flake8
> 
> I think it'll show an error if flake8 is not installed..?

No - it will just skip the tests if flake8 is not installed.

krischer:
Or at least it should. I think we can actually remove the `--no-flake8` flag from the run_tests script.

Issue #1048, opened 2015-05-14
Redirected FDSN client URLs get double-encoding


QuLogic:
This code will fail with a 400 Bad Request:
```
usgs = Client('USGS', debug=True)

minmag_glob = 6
tstart = UTCDateTime('2000-01-01T00:00:00')
glob_ev = usgs.get_events(starttime=tstart, minmagnitude=minmag_glob)
```
The URL is `http://comcat.cr.usgs.gov/fdsnws/event/1/query?minmagnitude=6.0&starttime=2000-01-01T00%3A00%3A00.000000`. If you use the default client (IRIS), then the URL is `http://service.iris.edu/fdsnws/event/1/query?minmagnitude=6.0&starttime=2000-01-01T00%3A00%3A00.000000`.

However, if you check in wireshark, `comcat.cr.usgs.gov` redirects to `earthquake.usgs.gov`, and the parameters are encoding a second time, so that the URL is `http://earthquake.usgs.gov/fdsnws/event/1/query?minmagnitude=6.0&starttime=2000-01-01T00%253A00%253A00.000000`.

I don't seem to see any handling for 301 Location Moved status codes, though... Is this a bug lower in the stack, or maybe in `future`?

QuLogic:
OK, this looks like a bug in the USGS server:
```
GET /fdsnws/event/1/query?starttime=2000-01-01T00%3A00%3A00.000000&minmagnitude=6.0 HTTP/1.1
Accept-Encoding: identity
Connection: close
User-Agent: ObsPy 0.10.1 (Linux-3.19.5-200.fc21.x86_64-x86_64-with-fedora-21-Twenty_One, Python 3.4.1)
Host: comcat.cr.usgs.gov

HTTP/1.1 301 Moved Permanently
Date: Thu, 14 May 2015 02:54:22 GMT
Server: Apache
Location: http://earthquake.usgs.gov/fdsnws/event/1/query?starttime=2000-01-01T00%253A00%253A00.000000&minmagnitude=6.0
```

The `Location` in the response is double-encoded. I'm going to change the URL in the mapping for USGS, but we should report this bug to someone at USGS.

krischer:
Appears to be `mod_rewrite` messing things up. There are quite some resources online:

* http://tkurek.blogspot.de/2013/06/252f-instead-of-2f-in-url-apache.html
[...]

PR #1690, opened 2017-03-02
USGS web service endpoint should be updated to https

krischer:
This looks like the same underlying issue as #1048, a query to
http://earthquake.usgs.gov/fdsnws/event/1/query?starttime=2017-03-01T00%3A00%3A00.000000
gets double-encoded in the redirect, pointing to
https://earthquake.usgs.gov/fdsnws/event/1/query?starttime=2017-03-01T00%253A00%253A00.000000

I think that as with the earlier issue this is ultimately a bug on the USGS site, but it can be avoided by updating ObsPy's URL mappings to use the https URL.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -47,6 +47,7 @@ master: (doi: 10.5281/zenodo.165135)
      the more general `FDSNException`.
    * Fixing cross implementation of bulk waveform and station requests (see
      #1685).
+   * Updating some endpoint mappings to use HTTPS. (See #1690, #1048)
  - obspy.imaging:
    * The functionality behind the `obspy-scan` command line script has been
      refactored into a `Scanner` class so that it can be reused in custom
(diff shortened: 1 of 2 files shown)


krischer:
This issue is now a PR.

I tested all our URL mappings to see if they work with HTTPS and those that do now use it as the default mapping. I don't think there is a downside to this and it should work around this particular URL double encoding issue.

emartinez-usgs:
We will have the USGS servers updated later today, but good to know you will start directly requesting HTTPS in the future!

megies:
@krischer doctests have to be updated: https://travis-ci.org/obspy/obspy/jobs/208021430#L1428

krischer:
Fixed directly on master with 7fc054e2ebc0ff98db11c5ca18c727281b5286d3. Did not seem worthy of its on PR.

Issue #1072, opened 2015-06-03
rename datamark to win


ThomasLecocq:
Altough win is a stupid name, it's the real name of the data format, while datamark is the name of the datalogger...

As the CSD article is out now, and presents `obspy.win`, this might be a good time to start the name change+deprecation, no ?

xiansch:
I have tried using the datamark i/o on Hi-net win32 and it seems to work.  I am not familiar with the basic win format and the differences from win32, but have never heard of datamark, so second the name obspy.win.  Is anyone familiar with the win/win32 differences and possible conversion issues?  I'm assuming datamark = basic win and not win32?

megies:
@ThomasLecocq any insights?

ThomasLecocq:
Hi,
sorry for the delay, was on holiday :-)

Sooo.. Indeed, "Win" is the name of the format, and "Datamark" is the name of the datalogger, built by Hakusan. I named it "datamark" because it's how it was called in Indonesia by the CVGHM. 

Check the PDF and email of Hakusan corp in the io/datamark folder for more information about win.

So, from what I can tell, the current naming could be changed to win. 

On the other hand, I *think* `win` is not `win32`, the new format stored by the Datamark/Hakusan dataloggers. I have never seen win32 files so I can't tell.

HTH

krischer:
@xiansch Can you provide us with some `win32` files?

xiansch:
I put a pull request on master/docs, there is a sample win32 file and the pdf with the data format.  Normally the win files are 1 min long (100 Hz) but the one I used is downsampled to 20 Hz, and 1 hour, and I extracted only 3 channels because of the file size.  Since the win32 files sometimes contain a lot of channels that are not needed (again, I'm not sure about regular win but I think it is the same) it would also be nice to have the i/o read specific channels only as a input parameter.

xiansch:
Hi, I've contacted NIED about using win32 in the obspy docs but there has been no reply as of yet.  For the time being, is there a way to get io.datamark to do a little more than just read() for the basic win/datamark file, such as read in particular channels?
There are some native tools for doing that on the win site: http://eoc.eri.u-tokyo.ac.jp/WIN/Eindex.html
such as dewin: extracts a channel and writes to .txt file
wch: extracts channels (can do multiple channels) but writes each to a _separate_ win file
so it is possible to extract all channels before reading into obspy, but somewhat troublesome and exponentially increases the number of files in storage.
Ideally I would want to be able to do something like
[...]

PR #1692, opened 2017-03-02
support reading WIN32 format

seisman:
This PR adds basic support for reading WIN32 format. It's a little similar to obspy.io.datamark. It works under Python 3.5 but fails under Python 2.7, since I know nothing about Python 2/3 compatibility. :cry: 

It still need further revisions to make it compatible and I open this PR to see if you have any suggestions.


seisman:
As far as I can see, there are at least four differences between datamark and win32:

1. win32 has a 4 bytes file header, which is always equals to '0x00000000'; datamark seems to has no file header;
2. win32 use 8 bytes to store starttime, and datarmark use 6 bytes;
3. for each one-second block, both win32 and datamark have a block header, but the sizes of block header are different;
4. the header for each channel seems a little different, too. win32 block header contains more information;

Maybe the datamark can be extended to support win32 format by adding several if-elif-else statements.

megies:
Please compare #1072..

ThomasLecocq:
@krischer so datamark is actually the name of the datalogger and should have been called "win" since the beginning. Now win32 is a slightly evolved version. @megies and others, I could try to have a look at it in the coming months...

megies:
This PR looks pretty good already, great work @seisman!

A couple of things..
 - please check out the [adding a new submodule page](), e.g. docs skeleton and plugin integration in setup.py is missing
 - assuming that win32 is a special format of "win" maybe we should rename the new submodule to "win" and rename "core.py" to "win32.py".. just a thought, I don't know details of these formats, CC @ThomasLecocq (maybe the full pdf document from which you cropped the win32 section might give some answers as to the different flavors of win?)
 - all py files need our common "future import" header, see https://travis-ci.org/obspy/obspy/jobs/207036477#L1133 and just copy paste the top import lines from any other py file in the repo

krischer:
Should we also rename `io.datamark` to `io.win` and maybe somehow merge the win and win32 data format if they share a lot of characteristics. Would be easier to maintain in the long run.

krischer:
Sorry -- accidentially closed it by pressing the wrong button.

megies:
> Should we also rename io.datamark to io.win and maybe somehow merge the win and win32 data format if they share a lot of characteristics. Would be easier to maintain in the long run.

see #1072

barsch:
is the datamark logger the only device producing win/win32 format? if so I would almost prefer to leave it as datamark as win/win32 clashes a bit with a name of a certain os and it may be confusing ...

megies:
@seisman pasted a wrong link, but I guess you found it.. ;-)
https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule

> is the datamark logger the only device producing win/win32 format? if so I would almost prefer to leave it as datamark as win/win32 clashes a bit with a name of a certain os and it may be confusing ...

good point..

ThomasLecocq:
> is the datamark logger the only device producing win/win32 format? if so I would almost prefer to leave it as datamark as win/win32 clashes a bit with a name of a certain os and it may be confusing ...

AFAIK yes, the file format name is win and win32. And yes, it's a terrible name... 

see the docs attached to the format:
https://github.com/obspy/obspy/blob/master/obspy/io/datamark/docs/Re__DataMark_format_description.eml

megies:
Hmm.. since this format was supposedly developed by ERI, from what I understand, maybe use `ERI_WIN` or `ERI_WIN32` as names to avoid having to use `WIN` etc.?

seisman:
According to [FAQ09 on the Hi-net website](http://www.hinet.bosai.go.jp/faq/?LANG=en#Q09), win32 format is extended by NIED, not by ERI.

> NIED extended the Japanese popular seismic waveform data format "WIN" developed by the University of Tokyo. We named this extended format as "WIN32".

And according to [FAQ07](http://www.hinet.bosai.go.jp/faq/?LANG=en#Q07), it seems that I can't redistribute the win32 format datas downloaded from Hi-net website. Maybe I should remove the sample win32 files under `io/win32/tests/data`.
> You are prohibited from redistributing the data and/or graphical images provided by the NIED Hi-net website in any form. When announcing your original results obtained by using the downloaded data, this rule does not applied. In that case, however, you are obliged to indicate that the use of the NIED Hi-net (or F-net) data. For the data provided by other organizations such as the Japan Meteorological Agency or universities, you are required to use them only in accordance with each organization's rule.

megies:
> Maybe I should remove the sample win32 files under io/win32/tests/data

If you have other (free) test data better remove it (completely, also from history.. squash out, rebase+force-push), if not just ask for permission, I can't see how anybody would not allow inclusion as test data (assuming it's very short waveform snippets).

megies:
> NIED extended the Japanese popular seismic waveform data format "WIN" developed by the University of Tokyo. We named this extended format as "WIN32".

hmm.. so no idea how to best name it.. -_-

barsch:
so essentially two different formats: so if there are no efforts for merging both formats we just should roll with io.win and io.win32 - those modules shouldn't called directly anyway

barsch:
btw: we have another PR with a existing win32 sample file: https://github.com/obspy/obspy/pull/1494 in case we can't use the existing one

seisman:
I contacted Hi-net to ask for a permission one week ago. There is no reply yet...

ThomasLecocq:
just use those files: https://github.com/obspy/obspy/pull/1494/files

claudiodsf:
My two cents: I personally work with HiNet data and I know many people who use it. They all know that the file format is "win32" (even if that name may sound a terrible choice), but nobody knows about "datamark". So, I think we should stick with `io.win32`.

ThomasLecocq:
>My two cents: I personally work with HiNet data and I know many people who use it. They all know that the file format is "win32" (even if that name may sound a terrible choice), but nobody knows about "datamark". So, I think we should stick with io.win32.

@claudiodsf : I renamed datamark to `win` in #1717 and this one here should be called `io.win32` indeed.

seisman:
Two bad news:

1. My request is rejected.

>Thank you for accessing our website. 
> This is the NIED Hi-net/F-net/V-net administrator.
>
> We prohibit any redistribution of data from our webpage. No data is
currently allowed to be included in your module. We will consider releasing
a freely redistributable sample data as a request for a future update.
>
> Sincerely yours,
> NIED Hi-net/F-net/V-net administrator

2. The win32 files in #1494 have a network code of `0101`, which means these files are produced by Hi-net, too.

barsch:
after this answer from "NIED Hi-net/F-net/V-net administrator" I almost tend to not support this format at all - tztzt

anyway, @seisman can you produce some own test data you can use - where you know exactly it is working as expected - once NIED decides to release freely redistributable samples we can add them too

ThomasLecocq:
I've also asked someone I know at ERI if we can have de-classified data officially !

megies:
> Two bad news:
> 
>     My request is rejected.
> 
>>     Thank you for accessing our website.
>>     This is the NIED Hi-net/F-net/V-net administrator.
> 
>>    We prohibit any redistribution of data from our webpage. No data is

This must be a misunderstanding.. did you make clear that this is just about an example test file with ~1000 samples, i.e. a couple of seconds of data?

seisman:
I think I have made it clear that I only need a one-minute sample data of three channels. 

> Dear NIED/Hi-net administrators,
> 
> I'm a PhD student majoring in geophysics. Recently I'm trying to write a Python module which allows reading win32 format data in Python.This module will be included into ObsPy as a submodule in the future.
> 
> ObsPy (http://docs.obspy.org ) is an open-source project (licensed under the LGPL v3.0) dedicated to provide a Python framework for processing seismological data. It supports reading and writing several common file formats used in seismology, e.g., SAC, mseed and segy. However, it does not support win32 yet. I think it will benefit many users if obspy can support reading win32 format.
> 
> When writing this module, I need to include a sample win32 data into this module to test the codes. And this sample win32 data will be distributed by ObsPy licensed under the LGPL v3.0. However, according to the NIED/Hi-net policy, I have no right to redistribute any waveform data obtained from Hi-net website. 
> 
> So I'm writing this e-mail to ask if you can give a permission to allow me include a small piece of win32 data into ObsPy. This sample data will be one-minute long and include only three channels. Thank you.
> 
> Looking forward to your reply.
> 
> Best Regards,

megies:
Yep, perfectly clear in your email.. hard to understand why it should be a problem to give permission ofr inclusion of a small file sample.. they should be happy if software can read their data format and somebody else is doing the work actually..

ThomasLecocq:
I've also received the email address of two persons who would be able to answer "really" and not copy/pasting the faq... in progress.

megies:
We decided to push for 1.1.0 and since I'm assuming this will not be ready by Sunday I'm punting this to 1.2.0. Feel free to push on this if you feel it can be ready by Sunday.

megies:
I'm gonna punt this to 1.3.0 since it doesn't look like there has been progress..

megies:
Well, I don't know though, without a test file even.. is there no other converters somebody could use to make one, or ask again for a sample? It is kinda ridiculous to support a format for which we don't even have a single file

megies:
@ThomasLecocq I suppose we want this merged into the "new" `io.win` that was moved over from `io.datamark` or something?

ThomasLecocq:
> @ThomasLecocq I suppose we want this merged into the "new" `io.win` that was moved over from `io.datamark` or something?

Yes, so Win & Win32 are two different formats, but both made from the same dataloggers (datamark datalogger)/institute

megies:
Done, and over to you @ThomasLecocq

PR #1671, opened 2017-03-04
Master broken for older distros

megies:
This commit dbf73680b3e3a631f99f0c3cc88100a2394eff4d on master breaks obspy on older distros. It was discovered because for some PRs (e.g. #1670) not all docker testbot containers are able to send test reports which leads to the confusing situation that docker testbot commit status is "red" while the corresponding link to test reports at tests.obspy.org shows only "green" reports (but way fewer than the docker testbot runs containers). Please fix @krischer and next time check out docker testbot status before merging.. ;-)

### DIFF
--- a/obspy/core/util/decorator.py
+++ b/obspy/core/util/decorator.py
@@ -25,7 +25,7 @@ import warnings
 import zipfile
 
 import numpy as np
-from decorator import decorator, decorate
+from decorator import decorator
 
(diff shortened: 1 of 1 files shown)


megies:
Alright.. this should fix it hopefully. Please be more careful to not break our master @krischer..  (or at least fix it again 😜 )

megies:
All green, merging.

PR #1694, opened 2017-03-04
fix Miniconda installation in Travis for Mac OSX 64bit Py2k

megies:
Seems ContinuumIO changed Miniconda installer filename for Py2: see https://travis-ci.org/obspy/obspy/jobs/207402228#L72

.. or at least removed the alias `Miniconda-...` for Py2k


megies:
Travis is green, so merging.

PR #1697, opened 2017-03-06
MSEED: 1 microsecond rounding issue

krischer:
Seems we have a minor rounding issue on master: http://tests.obspy.org/72428/#1

### DIFF
--- a/obspy/io/mseed/tests/test_mseed_util.py
+++ b/obspy/io/mseed/tests/test_mseed_util.py
@@ -124,10 +124,10 @@ class MSEEDUtilTestCase(unittest.TestCase):
             self.assertEqual(
                 ts * 1000000, util._convert_datetime_to_mstime(dt))
         # Additional sanity tests.
-        # Today.
-        now = UTCDateTime()
-        self.assertEqual(now, util._convert_mstime_to_datetime(
-            util._convert_datetime_to_mstime(now)))
+        # Random date that previously failed.
+        dt = UTCDateTime(2017, 3, 6, 4, 12, 16, 260696)
+        self.assertEqual(dt, util._convert_mstime_to_datetime(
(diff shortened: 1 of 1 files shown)


krischer:
I did not manage to reproduce locally as the test was not reproducible as it used the current timestamp. I guess it was some kind of rounding issue with the new nanoseconds UTCDateTime that by chance manifested in that test run. I replaced the current timestamp with a hardcoded time of the test failure. If this works it should be good enough.

megies:
I saw a couple of those fails recently, see here: http://tests.obspy.org/?limit=50&show=errors

But yeah, hard to debug with a non fixed test value..

krischer:
I tried a couple million timestamps on my machine and I cannot reproduce it. @megies Can you try on debian?

barsch:
can't reproduce either - the test does not fail for me

megies:
I also can't reproduce it locally.. let's wait and see what the docker testbot has to say..

megies:
Hmm.. strange.. still think this might be a rounding issue edge case, maybe in combination with floating point accuracy of the timestamp?

The timestamp you coded in the test is only down to microseconds.. but in the test execution (the one that failed and was reported) it likely had more accuracy that was cut off? Dunno, didn't look into the new UTCDateTime details fully yet..

```python
        # Random date that previously failed.
        dt = UTCDateTime(2017, 3, 6, 4, 12, 16, 260696)
        self.assertEqual(dt, util._convert_mstime_to_datetime(
            util._convert_datetime_to_mstime(dt)))
```

krischer:
On my machine I tested it with randomly scattered timestamps in nanoseconds precision. I stopped it after a couple million iterations with no failure.

megies:
Ok.. let's handle it if we see more of it.. we have lots of tests after all.. :-)

PR #1700, opened 2017-03-06
More fixes various small issues, typos, UserWarnings, deprecations, etc.

barsch:
someone was to fast merging the previous one ;)

### DIFF
--- a/obspy/io/gse2/tests/test_core.py
+++ b/obspy/io/gse2/tests/test_core.py
@@ -10,6 +10,7 @@ from future.builtins import *  # NOQA
 import copy
 import os
 import unittest
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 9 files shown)


megies:
> someone was to fast merging the previous one ;)

ooops, who could that be.. ;-)

barsch:
imho ready to merge - dont feel like doing more :)

barsch:
@krischer: I didn't add a test just for the print(st) part - "/path/to/225051000_00008656" is used multiple times in unittest as well as in doctest - so I didn't bother adding an extra test just for this basic functionalty - but if you insist I can add something

barsch:
we could move that print(st[0].stats) also down to the second part of the doctest - than it is covered too

krischer:
Done in 3c8899e052762f1efdb83751448c3269e81c1f0f.

PR #1699, opened 2017-03-06
Raise better error message if failing to create ResourceIdentifier

krischer:
http://lists.swapbytes.de/archives/obspy-users/2017-March/002313.html

We should raise a message that is more descriptive and tells people how to fix it.


krischer:
This is now a PR.

Issue #1561, opened 2016-10-19
Release 1.0.3


megies:
Issues left to handle:
- [milestone 1.0.3](https://github.com/obspy/obspy/milestone/15)
- other issues with no action or lower priority were bumped to [1.1.1 milestone](https://github.com/obspy/obspy/milestone/17)

Things to do, when release candidates and final release are done:

#### pre release
- [x] Run docker tests (@krischer / @megies -- are done routinely now, but double-checking can't hurt)
- [x] Get a new DOI on zenodo (@megies)

#### during release
- [x] source distribution to pypi
- [x] [tag + release on github](https://github.com/obspy/obspy/releases)
- [x] conda packages via conda-forge: conda-forge/obspy-feedstock#6
- [ ] conda packages for linux 32bit via obspy channel (@krischer)
- [x] Windows pypi packages (@barsch)
- [x] debian packages (@megies)
- [x] redhat/centos packages (@qulogic)
- [ ] macports packages (@petrrr)

#### post release
- [x] Update docs once released (@megies)
- [x] News entry in the wiki
- [ ] "What's new?" page in the wiki
- [x] Email to the users and announcements mailing lists
- [x] Tweet
- [ ] update base branches for all issues bumped to 1.1.1 to a new branch `maintenance_1.1.x`

ThomasLecocq:
#1553 is done I think now.

Jollyfant:
Any more changes for the log? I will add them all at once before the release.

- obspy.io.seiscomp
  * Fixed implicit conversion of None to 'None' for empty tags. (see #1537)

- obspy.signal.quality_control
  * New module for assessing quality metrics for mSEED data. (see #1141)

megies:
The #1537 missing changelog is a nice catch.. you/we/somebody can just commit it directly on `maintenance_1.0.x`.

#1141 as a new feature was merged into `master` for 1.1.0 and it's already in the changelog: https://github.com/obspy/obspy/blob/master/CHANGELOG.txt#L62

calum-chamberlain:
If I do a bit of work today on #1517 it might be good to go - would be good to get that into a release (either this one of #1562)?

Jollyfant:
I think it fits better with 1.1.0 as it is a new feature. But if you can finish it power to you!

megies:
@obspy/developers I propose we target end of November for releasing 1.0.3 and 1.1.0.

megies:
@obspy/developers 

[...]

PR #1684, opened 2017-03-06
CatchOutput manager makes tests crash on Windows Python 3.6

megies:
See https://github.com/obspy/obspy/issues/1561#issuecomment-282266027

```
py3.6 on Windows crashes on our CatchOutput implementation
```

### DIFF
--- a/obspy/core/tests/test_response.py
+++ b/obspy/core/tests/test_response.py
@@ -26,7 +26,7 @@ from matplotlib import rcParams
 from obspy import UTCDateTime, read_inventory
 from obspy.core.inventory.response import (
     _pitick2latex, PolesZerosResponseStage)
-from obspy.core.util.misc import CatchOutput
+from obspy.core.util.capture import CCatchOutput
 from obspy.core.util.obspy_types import ComplexWithUncertainties
 from obspy.core.util.testing import ImageComparison, get_matplotlib_version
 from obspy.signal.invsim import evalresp
@@ -167,7 +167,7 @@ class ResponseTestCase(unittest.TestCase):
         t_samp = 0.05
         nfft = 256
 
-        with CatchOutput():
+        with CCatchOutput():
             self.assertRaises(ValueError,
                               inv[0][0][0].response.get_evalresp_response,
                               t_samp, nfft, output="DISP")
--- a/obspy/core/tests/test_response.py
+++ b/obspy/core/tests/test_response.py
@@ -26,7 +26,7 @@ from matplotlib import rcParams
 from obspy import UTCDateTime, read_inventory
 from obspy.core.inventory.response import (
     _pitick2latex, PolesZerosResponseStage)
-from obspy.core.util.capture import CCatchOutput
+from obspy.core.util.capture import SuppressOutput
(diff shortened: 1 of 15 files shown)


krischer:
Man - can we maybe get rid of the C-level catching? If its only for a couple of tests we could also modify the C sources. Otherwise let me know once you think you got it working and I'll test on OSX which is also problematic with these things.

QuLogic:
IIRC, one of the problems on Python 3 is that `flush` does not necessarily flush the libc cache, which is mentioned in the test. Maybe adding a `libc.flush` somewhere will help fix it.

barsch:
yeah I tried all kind of libc flushes libc.flush(c_stdout) as well as clib.flush(None) as shown in various recipies

here some recipies I tried:

http://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/
http://stackoverflow.com/questions/17942874/stdout-redirection-with-ctypes-and-python
https://github.com/Drekin/readline-hooks/blob/master/readline_hooks.py (various c level stream filepointer for windows)
https://github.com/pytest-dev/pytest/blob/master/_pytest/capture.py
https://github.com/behave/behave/blob/master/behave/capture.py
etc.

it may run for your test environment - but as soon as you start testing against all supported python versions on Windows it fails

megies:
Rebased on current master and force-pushed to update CI results.

barsch:
not working - we should just remove all C-level catching - makes our life much easier ...

megies:
In any case we need to decide what to do with this PR in the next days to not hold up 1.1.0...

Minimal solution: Short-circuit/disable CatchOutput manager on Py36+Win?

ThomasLecocq:
@barsch  It looks like the pytest guys also struggled and this https://github.com/pytest-dev/pytest/pull/2462 is their solution....

ThomasLecocq:
@barsch : found that adding ``setbuf(stdout, NULL);`` to the ``decomp_6b_buffer`` routine solves the CatchOutput issue.

(modifying this too, of course):

```python
    def test_no_dat2_null_pointer(self):
        """
        Checks that null pointers are returned correctly by read83 function
        of read. Error "decomp_6b: Neither DAT2 or DAT1 found!" is on
        purpose.
        """
        filename = os.path.join(self.path,
                                'loc_RJOB20050831023349_first100_dos.z')
        fout = io.BytesIO()
        with open(filename, 'rb') as fin:
            lines = (l for l in fin if not l.startswith(b'DAT2'))
            fout.write(b"".join(lines))
        fout.seek(0)
        # with CatchOutput() as out:
        with CatchOutput() as out:
            self.assertRaises(GSEUtiError, libgse2.read, fout)
        # XXX: CatchOutput does not work on Py3k, skipping for now
        self.assertEqual(out.stdout,
                        b"decomp_6b: Neither DAT2 or DAT1 found!\n")
```

barsch:
@ThomasLecocq will look into it - thanks for the link

barsch:
the saga continues in #1854

barsch:
@ThomasLecocq where exactly did u put ```setbuf(stdout, NULL);``` to ```decomp_6b_buffer```?

PR #1702, opened 2017-03-06
Attempting to polyfill the decorate() function

krischer:
See #1644.

#1671 did fix the test failures but the decorator in this case did not actually do anything as the lock was recreated for each function call and thus there was no lock.

This is an attempt to polyfill the `decorate()` function to older decorator modules as that functionality is crucial for a lot of things that don't really work with normal decorators like our documentation.

I honestly have no idea how to actually test this - one would need to get access to the local variables of the decorated function but `__closure__` appears to be empty. A found a solution online but that seems a bit too much hassle to actually do it: http://stackoverflow.com/questions/9186395/python-is-there-a-way-to-get-a-local-function-variable-from-within-a-decorator

I tested this with print statements and it appears to work as intended.

### DIFF
--- a/obspy/core/util/decorator.py
+++ b/obspy/core/util/decorator.py
@@ -300,18 +300,39 @@ def map_example_filename(arg_kwarg_name):
     return _map_example_filename
 
 
-@decorator
-def rlock(func, *args, **kwargs):
+def _decorate_polyfill(func, caller):
+    """
+    decorate(func, caller) decorates a function using a caller.
+    """
+    try:
+        from decorator import decorate
+        return decorate(func, caller)
+    except ImportError:
+        from decorator import FunctionMaker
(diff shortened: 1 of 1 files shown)


megies:
> #1671 did fix the test failures but the decorator in this case did not actually do anything as the lock was recreated for each function call and thus there was no lock.

Sorry.. was relying on test results there as I wasn't familiar with the details.

krischer:
No problem - and it also is a bit tricky and not immediately obvious + not tested...

You are also not the only one ;-) https://github.com/obspy/obspy/commit/18dbddb20369ad2ce3f08c628ce568d576d287be#diff-20a7c5c57de612451c7de805bc4addecR299

krischer:
Yay :) Seems to work! Even though the decorator is not directly tested it is executed quite a bit in the whole test suite so I conclude that it works as expected.

PR #1696, opened 2017-03-06
io.nordic: Settings "Time_Residual_RMS"

calum-chamberlain:
https://github.com/obspy/obspy/pull/1695/commits/4363cb6bf2fc6d61a5886e69c03a6ab4629a37a6 hides a warning that is set when `Time_Residual_RMS` timing errors are set. Our event data model cannot represent that and it should potentially be changed to something else? Support error quantifications include `uncertainty`, `lower_uncertainty`, `upper_uncertainty`, and `confidence_level`.

@calum-chamberlain: Any opinion/preference?

### DIFF
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -32,7 +32,7 @@ import io
 from obspy import UTCDateTime, read
 from obspy.geodetics import kilometers2degrees, degrees2kilometers
 from obspy.core.event import Event, Origin, Magnitude, Comment, Catalog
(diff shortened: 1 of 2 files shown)


calum-chamberlain:
Oh, very exciting - does that look right?
Also, thanks for educating me! hub looks very useful.

krischer:
I actually don't use it for anything else but it does indeed look like it has some nice features :-)

krischer:
Thanks for the quick work on this @calum-chamberlain!

PR #1703, opened 2017-03-07
Update libmseed to 2.19.2.

QuLogic:
New release of libmseed is out; see [here](https://github.com/iris-edu/libmseed/releases). (They're on GitHub now!)

Since libmseed is using git now, I don't know if we maybe want to use a submodule or subtree in the future.


krischer:
Looks good to me.

@megies Can you have a look at the docker failures? I'm not fully certain that they are unrelated.

Regarding submodule/subtree. Good idea :) In this particular case I'd personally favor a subtree because the initial cloning is easier. Also the libmseed repository is fairly small (2.4 MB; 100 commits) so the additional burden on our repository is not too big.

megies:
> @megies Can you have a look at the docker failures? I'm not fully certain that they are unrelated.

i've seen those every now and then.. never was able to pin it down, really.. at least it's unrelated

PR #1705, opened 2017-03-07
docker: adjust cronjob docker testbot script for usage outside of vbox

megies:
on a normal workstation, and also add some more docs..

CC @krischer


krischer:
As far as I understand it I have to merge this before I can try it, right?

megies:
> As far as I understand it I have to merge this before I can try it, right?

You can check out this branch in a local clone and run it there

krischer:
Oh - so you mean haveing one obspy repository with the script and a separate working dir. Got it.

megies:
Reportedly works well with @krischer's new testbot, so merging.

PR #1707, opened 2017-03-07
avoid using basemap hires data in tests, could lead to faster installation during CI

megies:
..because basemap-hires conda packages don't have to be installed.

Basemap standard build on conda-forge comes with resolution "c" and "l"
only.

If we're lucky image diffs are even still within tolerance..


megies:
Looks like this can be merged..

PR #1708, opened 2017-03-08
Appveyor improvements

barsch:
- now building Py2.7/64 bit and Py3.6/32bit
- enabled rolling builds: whenever you do a new commit to the same branch OR pull request all current queued/running builds for that branch or PR are cancelled and the new one is queued. 
- enabled fast finish:  if first job has failed the entire build is marked as failed and no further jobs will start
- do not build feature branch with open Pull Requests

todo in future: add +TESTS magic to enable building the full matrix (py2.7 - py3.6, 32 and 64bit)

IMHO ready to merge see: https://ci.appveyor.com/project/obspy/obspy/build/1.0.4368-appveyor

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,13 +39,13 @@ environment:
     #   PYTHON_VERSION: "3.5"
     #   PYTHON_ARCH: "64"
 
-    # - PYTHON: "C:\\Miniconda36"
-    #   PYTHON_VERSION: "3.6"
-    #   PYTHON_ARCH: "32"
-
-    - PYTHON: "C:\\Miniconda36-x64"
+    - PYTHON: "C:\\Miniconda36"
       PYTHON_VERSION: "3.6"
-      PYTHON_ARCH: "64"
+      PYTHON_ARCH: "32"
+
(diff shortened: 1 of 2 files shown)


megies:
Great improvement to our CI, thanks @barsch!

krischer:
Yea - looks great :)

> now building Py2.7/64 bit and Py3.6/32bit

Contrary to that statement it currently tests both for 64 bit. Also should we maybe enable 32 and 64 bit builds? Appveyor is usually faster then travis so I don't think it will slow down our whole CI turnaround time.

krischer:
For future reference: There is also this package here: https://github.com/pelson/Obvious-CI

conda-forge uses it for the appveyor builds and it would make it much easier. But as you already invested all the sweat I'd leave it as it is for now :) Also I don't think this can do the early termination and rolling builds which are really cool.

barsch:
ups commented out the wrong version ;)

on appveyor we have only one job per time for the whole project - there is nothing parallel - so increasing the build time from 10 min to 80 min per checkin is quite a difference ... so I choose to offer only min and max python version ...

barsch:
ok CI good - merging

PR #1711, opened 2017-03-10
earthworm client: slow stream cleanup

tparker-usgs:
Earthworm's waveserverV returns data in an ordered set of small chucks. The first chuck returned may start before the requested start time, and the last chuck returned may end after the requested end time. The time required to trim the data using Stream.trim() grows linearly with the number of data chunks returned by the server.

This change replaces the call to Stream.trim() with individual calls to trim only the first and the last trace in the Stream.

### DIFF
--- a/obspy/clients/earthworm/client.py
+++ b/obspy/clients/earthworm/client.py
@@ -123,7 +123,7 @@ class Client(object):
         if cleanup:
             st._cleanup()
 
-        # only traces from the first and last tracebug need to be trimmed
+        # only traces from the first and last tracebuf need to be trimmed
         st[0].trim(starttime, endtime)
         st[len(st)-1].trim(starttime, endtime)
 
--- a/obspy/clients/earthworm/client.py
+++ b/obspy/clients/earthworm/client.py
@@ -122,7 +122,11 @@ class Client(object):
             st.append(tb.get_obspy_trace())
         if cleanup:
             st._cleanup()
-        st.trim(starttime, endtime)
+
+        # only traces from the first and last tracebug need to be trimmed
+        st[0].trim(starttime, endtime)
+        st[len(st)-1].trim(starttime, endtime)
+
         return st
 
     @deprecated("'saveWaveform' has been renamed to 'save_waveforms'. Use "


krischer:
Also note that if `cleanup` is True it will merge perfectly adjacent traces in any case which (for reasonable data) should result in very few Traces that actually enter the trimming operation.

tparker-usgs:
Thanks for the quick review. Sorry I have so much to fix.

We're wanting to migrate some operational code from Matlab to Obspy and are running into performance problems using the earthworm client. My pull requests are attempting to correct problems which are preventing us from using Obspy. I administer the wave server that the earthworm client uses as its example in the docs and have no reason to suspect that what I'm seeing is an incompatibility between my server and the code.

As an example, I retrieved 24 hours of data a few different ways and timed how long it took at each stage. The code I used was modeled after the example in the earthworm client docs.

    client = Client("pubavo1.wr.usgs.gov", 16022)
    st = client.get_waveforms('AV', 'ACH', '', 'EHE', UTCDateTime(2017,3,8,0,0), UTCDateTime(2017,3,9,0,0), cleanup=False)


The times are from a single run each, but are consistent with previous testing. The request for 24 hours of data returned 86,401 traces and what I found is in the table below.

Task | Total Run (seconds)| Data Retrieval (seconds) | Build Stream Object (seconds) | Cleanup (seconds) | Trim (seconds)
-----|-----------|---------------|----------------------|---------|------
Retrieve with cleanup=True | 692.6 | 13.6 | 25.1 | 651.5 | 0
Retrieve with cleanup=False | 85.2 | 16.2 | 27.5 | 0 | 38.8
Retrieve with cleanup=False and minimal trimming | 40.5 | 12 .7 | 25.7 | 0 | 0 

I haven't looked at the cleanup code, but it appears to run in exponential time. Over 11 minutes to cleanup 24 hours of wave data is consistent with my other testing.

I'm not sure why a noop would take 39 seconds, even when called 86k times. Perhaps this is revealing another problem? If there's a better way to resolve the performance, I'm happy.  

Good call on the station wildcard. The method returns early when provided a wildcard. Wy code doesn't break, but it also doesn't change the trim behavior.

I'll rebase on master and also apply my change to the trim behavior when a wildcard is provided.

krischer:
Thanks for your measurements which uncover quite some issues! Furthermore none of the core developers regularly uses earthworm so we were not aware of the issues.

So 86k traces is really a bit of an extreme case and everything is definitely too slow and needs to be faster. ObsPy and Python are also not really designed for these kinds of things which is also the reason why we don't for example expose record-level MiniSEED to the users.

* The trimming IMHO should be faster - I'll look into it next week.
* The cleanup code path is slow as it successively adds traces together which keeps making new, larger arrays... This needs to be fixed as it also matters for traces which have way less traces.

In most practical cases data has only very few traces so we were not really aware of the problems.

Regarding the proper fix for the waveserver client: IMHO the code should be refactored quite a bit to first collecting everything in a buffer and then treating it as a file and calling a `read_file()` method on it, e.g (pseudocode):

```python
with io.BytesIO() as buf:
    while data:
        buf.write(data)

    buf.seek(0, 0)
    st =read_file(buf)
```

This (if done properly) would eliminate almost all the overhead and runtime would pretty much be the download time + some parsing time. ObsPy reads an equivalent MiniSEED file in 0.3 seconds.

Do you know the format that the waveserver servers? Judging from a quick look at the code it appears to be some kind of custom but simple binary format. This could be read quite efficiently using numpy structured arrays and then maybe going down to C for the chunk merging part.

You might have to spearhead this a bit to get it done in due time but I'm happy to help out along the way!

tparker-usgs:
I think you're right about the proper fix. I had been dodging that solution, but it is the right thing to do. I'm somewhat familiar with the earthworm WSV protocol and will give it a shot. I may need to take you up on your offer of help.

Should I close this pull request and create a new one when ready, or leave this one open?

megies:
+1 for fixing this during interpreting the data returned from the server. The current implementation clearly is quick-and-dirty in this regard, creating lots of individual Trace objects and handing them over to the user to worry about.

Would be great if you work on this, this will definitly be an improvement for our earthworm client implementation. :-)

> Should I close this pull request and create a new one when ready, or leave this one open?

You can force-push to the branch of this PR and reuse this PR. That will avoid ticket duplication. :-)

megies:
@tparker-usgs I pushed a quick commit to here: https://github.com/megies/obspy/commits/earthworm_data_parsing

This might be a small improvement, avoiding to copy data around in memory to some extent (first determining which packets are contiguous and then merging the numpy array data once). If this is still too slow you might have to go with the record array approach @krischer mentioned. You could have a look at my implementation of reftek reader based on record arrays in #1433. Most relevant pieces might be these:
 - https://github.com/obspy/obspy/pull/1433/files#diff-01bf13111af73817bce2b2a6f9c28b1dR219
 - https://github.com/obspy/obspy/pull/1433/files#diff-b5aef08f8cd81c4a607c8deeebbe3625R139

megies:
This might take some more time, so bumping to 1.1.1 (assuming that we can consider this some sort of a bug fix..?)

tparker-usgs:
Following suggestions by @krischer and @megies, I reworked the code to merge adjacent traces as the data is retrieved. This works well and performance is much better.

However, I clobbered the head fork of this pull request without understanding how it would affect the pull request. Is there a way to point the PR to a new fork, or should I close and re-create the pull request?

krischer:
Hmm - I'm not sure if you edit the origin branch/fork of a PR. You also remove the original one so feel free to open a new PR as I'm not sure how to otherwise resolve this.

megies:
Re-adding this to 1.1.0 as it sounds like this might be almost ready..

PR #1713, opened 2017-03-12
Travis: fast fail builds if newer builds for the branch/PR are already queued

megies:
This PR enables a strategy to reduce number of redundant Travis builds by reusing `travis_fastfail.sh` implemented by Julia developers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,13 @@ matrix:
     - os: osx
       env: PYTHON_VERSION=3.5
 
+# mimick travis fast fail and rolling build setup from:
+# https://github.com/JuliaLang/julia/blob/master/.travis.yml
+before_install:
+  - if [ `uname` = "Darwin" ]; then brew update; brew install -v jq; fi
+  - curl "https://raw.githubusercontent.com/JuliaLang/julia/master/contrib/travis_fastfail.sh" -o "/tmp/fastfail.sh" || echo "Failed to fetch travis_fastfail.sh"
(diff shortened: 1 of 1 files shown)


megies:
Works.. https://travis-ci.org/obspy/obspy/builds/210221363

(removed those dummy commits again that were used to test this PR: 690074a and 2486b48)

megies:
Good to merge.. I think.

QuLogic:
Travis actually has fast-fail in beta; we're using it on Matplotlib. It's a bit buggy with already-running tests though.

megies:
Good to know, we should definitely switch once it goes out of beta (or maybe even now already?)

krischer:
> Good to know, we should definitely switch once it goes out of beta (or maybe even now already?)

I'd vote for switching right now - at least we won't forget then.

megies:
I didn't find where/how this is set up in matplotlib, doing a quick search.. can you give a quick link @QuLogic?

QuLogic:
We have this setting, and I think you need to ask for it because it's in beta:
screenshot from 2017-03-14 17-12-39
As noted, it doesn't work on already-running tests, but maybe trying adding a 🎉 on travis-ci/beta-features#8 so they'll have more pressure to add it.

megies:
I'm not seeing that switch on our settings page.. I guess we can still switch once it's in public beta or out proper. Added the 🎉  in any case.. ;-)

megies:
Auto cancellation is now live as a beta feature on Travis and I've switched it on for PRs in our Travis settings (not sure if it can be set via Travis yaml, which is always a bit more trasparent..).

PR #1717, opened 2017-03-13
renaming obspy.io.datamark to obspy.io.win

ThomasLecocq:
### What does this PR do?
This PR corrects my original mistake of calling the WIN format DATAMARK.
I refactored the code and the entry points but I didn't create deprecations.

It's linked to 

- https://github.com/obspy/obspy/pull/1692
- https://github.com/obspy/obspy/pull/1494

- [x] All tests still pass.


ThomasLecocq:
I guess codecov failure is due to moving/renaming.

setup.py
```diff
@@ -142,7 +142,7 @@
         'SLIST = obspy.io.ascii.core',
         'PICKLE = obspy.core.stream',
         'CSS = obspy.io.css.core',
-        'DATAMARK = obspy.io.datamark.core',
+        'WIN = obspy.io.win.core',
```
megies: Hmm.. In general I agree that we can make an exception here and allow the module move/renaming without deprecation reroute, as people have no reason to use the low level routines.

But.. we could still break peoples' codes if they circumvent the filetype detection, i.e. if they do `read(..., format='DATAMARK')`.

Maybe we should leave the `DATAMARK` plugin in there (pointing to new `io.win`)?

Also, maybe we should wait with merging until it's clear how this one and #1692 will play together?

ThomasLecocq: would it work to use state:
`'DATAMARK = obspy.io.win.core',`
?

megies: And also the other plugin lines.. yeah.. not sure about any negative side effects, but probably it should be OK like this..

ThomasLecocq: or just a dummy obspy.io.datamark.core that answers it not longer exists and returns /dies. not a clean deprecation... but yah...

megies: > that answers it not longer exists and returns /dies

If you do a deprecation, than it shouldn't raise but rather show a warning and use the new location at obspy.io.win

ThomasLecocq: ... and stay until 1.2 ? pfff not sure...

megies: If you do a deprecation, it will be in there for 1.1.* and removed again for 1.2.0

obspy/io/win/README.txt
```diff
@@ -1,24 +1,28 @@
-package obspy.io.datamark
+package obspy.io.win
 =========================
```
megies: header underline needs adjustment

setup.py
megies: @ThomasLecocq do you want to add a deprecation or leave as is? We plan to freeze 1.1.0 on Sunday so if this PR should be in there it should be ready by then.

ThomasLecocq: I'd suggest to keep this as-is... I don't know many users of this routine anyway, and I've modified the docstring so that a simple googling should show DATAMARK in the io.win explanation

krischer:
> And the moral of the story.. alwys try to get module names right with the first try.. ;-)

Yea ;-)

Looks good to me. We just have to remember to document this in some sense breaking chance in the release message.

Issue #1493, opened 2016-08-01
Incorporate new evalresp version?


krischer:
In #1492 it was noticed that a new evalresp version has been out for quite a while:

http://ds.iris.edu/ds/nodes/dmc/software/downloads/evalresp/4-0-6/revisions/

We should probably investigate if we want these changes. The main new feature is stationxml support which we already have. If we want to use the new stationxml support in evalresp we would once again have to use temporary files which I would really rather not do...

Additionally we would then have a compile time dependency on libxml2 - we already have a dependency on `lxml` which also needs `libxml2` so in principle this is no problem but we'd probably run into lots of build issues on all the various platforms.

If we don't incorporate the new evalresp we might have to backport this changlog item:

` * MODIFIED: Handling of stages with “no data”. Now accepted, as they may still have decimation and channel gain.`

All in all I think it would be nice have an up-to-date evalresp version within ObsPy but the effort is non-trivial as our evalresp version is patched quite a bit. I was actually under the impression that evalresp is basically not developed anymore but onyl receives the most necessary patches.

What do you guys think?

ghisvail:
Now that the issue with the vendored `libmseed` is solved, the vendored `evalresp` also must be made substitutable for a system installed version in order to make `obspy` package-able on Linux distributions.

Based on your listing here, I see two issues:
1. The vendored `evalresp` version is outdated. Linux distributions would typically package the latest upstream version, unless there is a very good reason not to.
2. The vendored `evalresp` is patched, i.e. there are (significant?) deviations from the upstream code to tailor it to what `obspy` expect.

Perhaps solving 1. would make 2. a little easier?

krischer:
> Perhaps solving 1. would make 2. a little easier?

Unfortunately not - it would very likely still require quite a number of patches and it is a lot of work for us with no true benefits. The main issue is that `evalresp` is not designed to be usable as a library where one can directly pass the information as C structs and we had to make some changes to enable that.

Please see my comment in https://github.com/obspy/obspy/issues/1492#issuecomment-285618905 about the upcoming re-design of `evalresp`. I don't know how long that will take though and if it will truly work for us - but I think that is one of the goals of the re-design.

megies:
[...]

PR #1718, opened 2017-03-13
Proof of concept: Using scipy instead evalresp

krischer:
Based on the discussion in https://github.com/obspy/obspy/issues/1493 it came to me that it would be almost trivial to replace evalresp completely with scipy. I don't know why I did not think of this before. Much simpler than our current wrapping of evalresp.

This PR does this for the PAZ, decimation, and coefficient response stages which are the most common and it also handles the final assembly and sensitivity calculations so it can already do quite a lot of responses. The rest is quite easy to add.

Of course this needs a lot more testing so I'd personally not include this in the upcoming 1.1 but rather wait a release.

The amplitude response matches the evalresp one to a relative error of 1E-10 - the sensitivity scaling makes this a bit different but I think what's done in this PR is correct. The phase differs a bit more for the FIR stage mainly after nyquist but also a tiny bit before. Again I think here scipy does the correct thing as it for example has no Gibb's artifacts which evalresp has.

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -28,7 +28,7 @@ from obspy.core.util.obspy_types import (ComplexWithUncertainties, CustomFloat,
                                          FloatWithUncertainties,
                                          FloatWithUncertaintiesAndUnit,
                                          ObsPyException,
-                                         ZeroSamplingRate)
+                                         ZeroSamplingRate, Enum)
 
 from .util import Angle, Frequency
 
@@ -763,6 +763,10 @@ class Response(ComparingObject):
     """
     The root response object.
     """
+    # The various types of units.
+    core_unit_enum = Enum(["displacement", "velocity", "acceleration",
+                           "volts", "counts", "tesla", "pressure"])
+
     def __init__(self, resource_id=None, instrument_sensitivity=None,
                  instrument_polynomial=None, response_stages=None):
         """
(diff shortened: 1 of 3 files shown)


jkmacc-LANL:
+1 !

amkearns-usgs:
I would like to confirm that this pull request is ready / awaiting review.

megies:
> I would like to confirm that this pull request is ready / awaiting review.

uhm.. it handles any types of response stages out there already??

amkearns-usgs:
> > I would like to confirm that this pull request is ready / awaiting review.
> 
> uhm.. it handles all different types of response stages out there already??

I believe so. I tested it rather thoroughly with Ringler in order to make sure it did. (There was a lot of back-and-forth about how to handle digital vs. analog filter stages.)

krischer:
Removing the response is such an integral part of ObsPy that we absolutely have to make sure that we can explain every difference to the way evalresp does it. When we initially implemented the StationXML response removal and being able to read SEED files we did by comparing the output of raw evalresp to almost every response file in existance: https://github.com/obspy/sandbox/tree/master/the_great_response_test

I think something similar would be needed here in addition to all the testing you already did. I can take care of setting up a basic version - should not take too long.

That is the most up-to-date branch here, correct? https://github.com/amkearns-usgs/obspy/tree/evalresp_update

amkearns-usgs:
Yes. We haven't been rebasing it to any updates on the main branch at all -- I don't think we've touched this since like April or so, but that's where all the updates would happen.

aringler-usgs:
@krischer thanks for setting up an additional location for tests.  I am sure there will be a few oddities that come out of this.  I will work with @amkearns-usgs to deal with the various issues in the tests.  Thanks!

krischer:
Great job @amkearns-usgs & @aringler-usgs !! This already works really really well and given how much time I spent on this problem in the past I personally really appreciate this. So thanks a lot already.

I pushed a first version of the aforementioned test here: https://github.com/obspy/sandbox/tree/master/the_great_response_test_scipy

I hope it is easy enough to understand. The main idea is to run this against a significant portion of the data out there and see if the results are identical between evalresp and the new scipy implementation. I'd be fine with some differences as long as they are documented and rational.

krischer:
I realize that this might be a decent amount of work. But given that response removal is one of the absolute core features of ObsPy we cannot afford to go wrong here. I'm happy to take some of the work.

Given that you know the code better than me I think it would be most useful if you try to deal with the initial broader issues and I'll jump it at any time when you need a break.

krischer:
I also pushed a tiny commit here: https://github.com/amkearns-usgs/obspy/tree/evalresp_update

krischer:
Let's continue the discussion in https://github.com/obspy/obspy/pull/2592 - more useful when the code is directly there.

PR #1716, opened 2017-03-13
generalized linear_regression

ThomasLecocq:
This PR adds a generalized linear regression to `obspy.signal`  - this allows OLS and WLS, with/without allowing for an intercept.

### PR Checklist
- [x] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [x] First time contributors have added your name to `CONTRIBUTORS.txt` .

### TODO

- [x] Test

PR #1719, opened 2017-03-13
Feature mwcs

ThomasLecocq:
### What does this PR do?
Adding the Moving Window Cross-Spectral method. This method allows comparing two signals in the frequency domain (cross-spectrum, or cepstrum) and returns a table of time delays (offsets) vs time for each window.

### PR Checklist
- [X] it is linked to #841 and branched from #1716
- [X] All tests still pass.

- [ ] I should provide an example with some figures.


ThomasLecocq:
wow I have no idea why this fails on Travis while it's OK on appveyor ? @megies @krischer help!?

ThomasLecocq:
@barsch @megies @krischer 

The tests fail depending on the scipy version available, because the results of the MWCS are slightly different when using nextpow2 or next_fast_len ? Should this function be defined with nextpow2 by default but allow for some sort of callback in case one wants to use next_fast_len (scipy >= 0.18) ?

krischer:
Can you up the test tolerance to have it test similar enough on different scipy versions?

PR #1720, opened 2017-03-14
simple but effective speed up of test_tau.py

barsch:
reducing runtime on my computer from 116s to 14s by randomly skipping test values - set speedup factor to 1 to have the original behaviour


barsch:
don't know if wanted - but I'm kind of annoyed by the runtime of the taup tests

krischer:
Very good idea.

barsch:
merging - failing test is not related

PR #1725, opened 2017-03-20
more verbose TypeErrors in trace.__add__

ThomasLecocq:
### What does this PR do?
Just adds some verbosity in the sanity checks done in the __add__ method of traces.

It's soooooo frustrating to debug 
```TypeError("Data type differs") ```

### PR Checklist
- [x] All tests still pass.
- [X] Significant changes have been added to `CHANGELOG.txt`  (is this significant? don't think so)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -84,7 +84,7 @@ install:
   - "activate test"
   # Install default dependencies
   #   for now fix requests version, can be changed when a new requests is released (see #1599)
-  - "conda install -q --yes pip numpy scipy \"matplotlib<2\" lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema"
+  - "conda install -q --yes pip numpy scipy \"matplotlib<1.9\" lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema"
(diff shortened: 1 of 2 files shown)


ThomasLecocq:
wtf? appveyor?

QuLogic:
Change "matplotlib<2" to "matplotlib<1.9" since conda seems to think 2.0.0rc2 is not yet 2 and installs it.

ThomasLecocq:
> Change "matplotlib<2" to "matplotlib<1.9" since conda seems to think 2.0.0rc2 is not yet 2 and installs it.

I just based this on master, so you're saying we need to change master .appveyor.yml ?

megies:
> Change "matplotlib<2" to "matplotlib<1.9" since conda seems to think 2.0.0rc2 is not yet 2 and installs it.

Ah.. right, and that's actually correct according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#development-release-separators)

> I just based this on master, so you're saying we need to change master .appveyor.yml ?

Since this PR is ready to merge already, you can just add it to this PR so the change will be in master within days.

PR #1726, opened 2017-03-21
Tutorial: add example for custom basemap plot with Inventory/Catalog

megies:
I was asked how to achieve a local projection with custom extent and plot Inventory/Catalog into it, so might as well add it to Tutorial as well..

Also puts this example as the first Basemap example, as it's the first example using the canonical way of using the builtin Inventory/Catalog.plot() methods.

+DOCS

Good to go if the docs Tutorial page looks good..

### DIFF
--- a/misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
+++ b/misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
@@ -1,0 +1,36 @@
+from mpl_toolkits.basemap import Basemap
+import numpy as np
+import matplotlib.pyplot as plt
+
+from obspy import read_inventory, read_events
+
+# Set up a custom basemap, example is taken from basemap users' manual
+fig, ax = plt.subplots()
+
+# setup albers equal area conic basemap
+# lat_1 is first standard parallel.
+# lat_2 is second standard parallel.
+# lon_0, lat_0 is central point.
+m = Basemap(width=8000000, height=7000000,
+            resolution='c', projection='aea',
+            lat_1=40., lat_2=60, lon_0=35, lat_0=50, ax=ax)
+m.drawcoastlines()
+m.drawcountries()
+m.fillcontinents(color='wheat', lake_color='skyblue')
+# draw parallels and meridians.
+m.drawparallels(np.arange(-80., 81., 20.))
+m.drawmeridians(np.arange(-180., 181., 20.))
+m.drawmapboundary(fill_color='skyblue')
+ax.set_title("Albers Equal Area Projection")
+
+# we need to attach the basemap object to the figure, so that obspy knows about
+# it and reuses it
+fig.bmap = m
+
+# now let's plot some data on the custom basemap:
+inv = read_inventory()
+inv.plot(fig=fig, show=False)
+cat = read_events()
+cat.plot(fig=fig, show=False, title="", colorbar=False)
+
+plt.show()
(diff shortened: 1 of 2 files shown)


misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
```diff
@@ -0,0 +1,37 @@
...
+            resolution='l', projection='aea',
+            lat_1=40., lat_2=60, lon_0=35, lat_0=50)
+m.drawcoastlines()
+m.drawcountries()
+m.fillcontinents(color='coral', lake_color='aqua')
+# draw parallels and meridians.
+m.drawparallels(np.arange(-80., 81., 20.))
+m.drawmeridians(np.arange(-180., 181., 20.))
+m.drawmapboundary(fill_color='aqua')
+# draw tissot's indicatrix to show distortion.
```
QuLogic: That doesn't seem to be here...

misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
```diff
@@ -0,0 +1,37 @@
...
+plt.title("Albers Equal Area Projection")
+
+fig = plt.gcf()
+ax = plt.gca()
```
QuLogic: Unused, though it might be better to create the `Basemap` from the `Axes` instead of the other way around.

misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
megies: Oh, thanks, that's a part I removed from the original basemap example.. overlooked to remove the comment as well..

misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
megies: Thanks again!

megies:
Changing those ghastly colors.. but should be good to go now.. let's wait for docs buildbot once more..

megies:
http://docs.obspy.org/pull-requests/1726/tutorial/code_snippets/basemap_plot_with_beachballs.html#basemap-plot-with-custom-projection-setup

misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
```diff
@@ -0,0 +1,37 @@
+from mpl_toolkits.basemap import Basemap
+import numpy as np
+import matplotlib.pyplot as plt
+
+from obspy import read_inventory, read_events
+
+# Set up a custom basemap, example is taken from basemap users' manual
+fig = plt.figure()
+ax = fig.add_subplot(111)
```
QuLogic: fig, ax = plt.subplots()

misc/docs/source/tutorial/code_snippets/basemap_plot_custom.py
```diff
@@ -0,0 +1,36 @@
...
+from obspy import read_inventory, read_events
+
+# Set up a custom basemap, example is taken from basemap users' manual
+fig, ax = plt.subplots()
+
+# setup albers equal area conic basemap
+# lat_1 is first standard parallel.
+# lat_2 is second standard parallel.
+# lon_0, lat_0 is central point.
+m = Basemap(width=8000000, height=7000000,
```
ThomasLecocq: what are the units of width & height ?

megies: Just copy pasted from the basemap example.. in any case, width/height is in projection units. Not sure but I'm guessing meters.

PR #1728, opened 2017-03-23
MSEED: Segfault reading truncated file

krischer:
While trying to work around a problem when reasding truncated files (in SDS client while reading files that are currently being appended to by a different program), I came across a segfault when reading truncated MiniSEED files:

```python
import copy
from io import BytesIO
from obspy import read
from obspy.core.util import get_example_file

file_ = get_example_file('BW.BGLD.__.EHE.D.2008.001.first_10_records')

with open(file_, 'rb') as fh: 
    data = fh.read()

# for i in range(1, 1000):
for i in [257]:
    print(i)
    bio = BytesIO(copy.deepcopy(data[:-i]))
    read(bio, format='MSEED')
```

```
$ python read_mseed_truncated.py 
257
Segmentation fault
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -93,8 +93,9 @@ master: (doi: 10.5281/zenodo.165135)
    * Update to libmseed v2.19.2 (see #1703).
    * Correctly read MiniSEED files with a data offset of 48 bytes (see #1540).
    * InternalMSEEDReadingError now called InternalMSEEDError and
-	 InternalMSEEDReadingWarning now called of InternalMSEEDWarning as both
-	 can now also be raised in non-reading contexts (see #1658).
+     InternalMSEEDReadingWarning now called of InternalMSEEDWarning as both
+     can now also be raised in non-reading contexts (see #1658).
+   * Should no-longer segfault with arbitrarily truncated files (see #1728).
  - obspy.io.nlloc:
    * Set preferred origin of event (see #1570)
  - obspy.io.nordic:
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -257,6 +257,21 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
     else:
         bo = None
 
+    # Determine total size. Either its a file-like object.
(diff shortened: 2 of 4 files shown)


krischer:
This branch contains a fix: https://github.com/obspy/obspy/tree/mseed-fix-segfault-truncated-file

Not sure why I cannot convert this issue to a PR right now but I'll try again later tonight or tomorrow. Or maybe somebody else can try?

---

Some other types of record corruption where already caught by libmseed and correctly bubble up to the Python warnings. I'm not entirely sure why this one does not but maybe its just because its truncated fairly late in the file?

In any case: now works as expected and it raises a nice warning (but still reads all previous records).

krischer:
Hmm...looks like one of my tries did convert it to a PR in the end? Or did someone else do it?

Anyways - IMHO good to go. Feel free to review and merge :)

megies:
Thanks for the fix @krischer, checking again, there's still some truncation scenarios that end in segfaults though..

Can you maybe have a look at these two byte offset:
 - 256
 - 5066

These seem to be different issues.. the latter one I've seen in real live reading mseed files that currently also get appended to in other threads (checking data latency).

```python
import copy
from io import BytesIO
from obspy import read
from obspy.core.util import get_example_file

file_ = get_example_file('BW.BGLD.__.EHE.D.2008.001.first_10_records')

with open(file_, 'rb') as fh: 
    data = fh.read()

for i in range(1, 10000):
    # this seems to be a different issue than the already covered one:
    if i == 256:
        continue
    # these seem to be the same issue as with 256, as there just offset by 512
    # bytes..
    if i % 512 == 256:
        continue
    # this is finally the issue I was looking after: :-)
    if i == 5066:
        continue
    print(i)
    bio = BytesIO(copy.deepcopy(data[:-i]))
    read(bio, format='MSEED')
```

krischer:
All fixed, rebased and force pushed.

The 256 + 512 bytes offsets were just because I forgot the `<=` case. The larger truncation because you passed a file with less than 128 bytes - this now raises a much better error message.

megies:
Thanks for the fix(es)! 🎉

krischer:
IMHO ready to be merged.

PR #1731, opened 2017-03-27
fdsn mass downloader: skip some known failures on ancient scipy

megies:
Skip some known failures on ancient scipy < 0.12, see e.g. http://tests.obspy.org/74905/

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/__init__.py
+++ b/obspy/clients/fdsn/mass_downloader/__init__.py
@@ -513,12 +513,28 @@ Further functionality of this module is documented at a couple of other places:
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
+from future.utils import native_str
 
+import warnings
+
+import scipy
+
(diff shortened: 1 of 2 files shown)


obspy/clients/fdsn/mass_downloader/__init__.py
```diff
@@ -521,6 +526,15 @@
                      CircularDomain, GlobalDomain)  # NOQA
 
 
+__all__ = [native_str(i) for i in (
+    'MassDownloader', 'Restrictions', 'Domain', 'RectangularDomain',
+    'CircularDomain', 'GlobalDomain')]
+
+msg = ('At least some parts of FDSN Mass downloader might not '
+       'work with old scipy versions <0.12.0 (installed: {})')
+warnings.warn(msg.format(scipy.__version__))
```
krischer: Should this warning not only be raised if people actually have an old scipy versions?

megies: Oooops.. :blush:

Issue #639, opened 2013-09-23
Seishub client to raise more specific Exceptions


trichter:
The seishub client often raises a simple Exception e.g. if data is not available

```
Exception: No waveform data available
```

I think it is better to raise more specificly, e.g. something like `SeishubError` or `NoWaveformError`. (Arclink client raises `ArcLinkException`.)

megies:
Obsolete, closing

PR #1732, opened 2017-03-28
SDS client: ignore <128 bytes MSEED files

megies:
Just ignore small MSEED files, in use cases working with near-realtime data these are usually just being created right at request time, e.g. when fetching current data or checking latency right after midnight when first samples are currently being written to a new file.

### DIFF
--- a/obspy/io/mseed/headers.py
+++ b/obspy/io/mseed/headers.py
@@ -53,6 +53,9 @@ SAMPLESIZES = {'a': 1, 'i': 4, 'f': 4, 'd': 8}
 VALID_RECORD_LENGTHS = [256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536,
                         131072, 262144, 524288, 1048576]
 
+FILESIZE_TOO_SMALL_MSG = (
+    'The smallest possible mini-SEED record is made up of 128 bytes.')
+
 # allowed encodings:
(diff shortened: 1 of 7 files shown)


megies:
We already have `obspy.core.util.obspy_types.ObsPyException` and I'm already using it e.g. in the SDS report script and Reftek-130 read plugin.. 😜 

But yeah, good idea to also do it here.. since we're just now adding these new exceptions..

krischer:
Ah I did not even know that :)

If you end up changing it here - make sure the `ObsPyMSEEDReadingError` or how its called also inherits from the generic miniseed exception.

megies:
I've tried to come up with some proposal for more fine-grained exceptions, see also #639 and feel free to comment or modify..

obspy/io/mseed/headers.py
```diff
@@ -53,6 +53,9 @@
 VALID_RECORD_LENGTHS = [256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536,
                         131072, 262144, 524288, 1048576]
 
+FILESIZE_TOO_SMALL_MSG = (
+    'The smallest possible mini-SEED record is made up of 128 bytes.')
+
```
krischer: This is no longer needed and can be directly folded in the the error message.

PR #1734, opened 2017-03-28
io.wav: unskip test

megies:
Some `io.wav` test was skipped due to a wrong skip condition. Also refactoring getting dependency versions in `core.util`.


krischer:
The version getter now uses the `pkg_resources` package. Gets rid of having to rely on the `__version__` attribute being set.

megies:
Test fails on Travis OSX are due to some packaging issue with cartopy, so merging.

PR #1735, opened 2017-03-30
Exception raised while comparing stream objects read from big and little endian SEG-Y files

krischer:
Linux, Obspy 1.0.2 installed via pip

Hi trying to compare the stream objects read from two SEGY files (both were created using obspy), which only differ by endianess I get the following exception: 

    Python 2.7.12 (default, Jul 01 2016, 15:36:53) [GCC] on linux2
    >>> from obspy import read
    >>> sl = read("littleEndian.sgy")
    >>> s = read("bigEndian.sgy")   
    >>> s == sl
      File "/usr/lib64/python2.7/site-packages/obspy/core/stream.py", line 595, in __eq__
        if self_sorted.traces != other_sorted.traces:
      File "/usr/lib64/python2.7/site-packages/obspy/core/trace.py", line 298, in __eq__
        if not self.stats == other.stats:
      File "/usr/lib64/python2.7/_abcoll.py", line 426, in __eq__
        return dict(self.items()) == dict(other.items())
      File "/usr/lib64/python2.7/_abcoll.py", line 426, in __eq__
        return dict(self.items()) == dict(other.items())
      File "/usr/lib64/python2.7/_abcoll.py", line 426, in __eq__
        return dict(self.items()) == dict(other.items())
      File "/usr/lib64/python2.7/_abcoll.py", line 414, in items
        return [(key, self[key]) for key in self]
      File "/usr/lib64/python2.7/site-packages/obspy/io/segy/core.py", line 721, in __getitem__
        return super(AttribDict, self).__getitem__(name)
      File "/usr/lib64/python2.7/_abcoll.py", line 377, in __getitem__
        raise KeyError
    KeyError

Indicating that not all attributes are properly set in reading one of the files. The files read can be found in attached zip file [segy_files.zip](https://github.com/obspy/obspy/files/878843/segy_files.zip)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -81,6 +81,8 @@ master: (doi: 10.5281/zenodo.165135)
     * No longer add the (unused) time zone field to StationXML datetimes to
       follow the example of big data centers. (see #1572)
  - obspy.io.segy:
+    * Fixing an issue when comparing two still packed SEG-Y trace headers
+      (see #1735).
     * Iterative reading of large SEG-Y and SU files with
       `obspy.io.segy.segy.iread_segy` and `obspy.io.segy.segy.iread_su`.
       (see #1400).
--- a/obspy/io/segy/core.py
+++ b/obspy/io/segy/core.py
@@ -656,7 +656,7 @@ class LazyTraceHeaderAttribDict(AttribDict):
 
     This version is used for the SEGY/SU trace headers.
     """
-    readonly = ["unpacked_data", "endian"]
+    readonly = ["unpacked_header", "endian"]
 
     def __init__(self, unpacked_header, unpacked_header_endian, data={}):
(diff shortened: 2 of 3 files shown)


krischer:
Now a PR with a fix. There were a bit two many loops to jump when inheriting from AttribDict - it should now work fine. It will of course still not compare equal with two different byte orders but that is a different issue and IMHO likely not that crucial in practice.

krischer:
CI failure are unrelated.

megies:
Docker testbot says 👍

Fails are just because I had some new Debian 9 Stretch images lying around that have mpl 2.0 fails, so they can be ignored.

megies:
Travis fails are also unrelated.

krischer:
@paitor Thanks for your comment. I fixed it.

From my side this is ready to merge if all tests pass.

PR #1740, opened 2017-03-30
try to fix Travis OSX fails

megies:
Try to force conda-forge channel to top of channel priority list..


megies:
Oh great, if I use conda-forge for main dependecies as well we run into icu problems again..

PR #1738, opened 2017-03-30
missing mandatory items in SEGYFile._write_textual_header(...)

krischer:
Linux, Obspy 1.0.2 installed via pip

Hi 

The textual header in a SEGY file have two mandatory entries namely the revision number (same as in bytes 301-302 of the 400 byte binary file header) and the text "C40 END TEXTUAL HEADER" (or alternatively "C40 END EBCDIC") 

It would be neat if this were automatically set by SEGYFile._write_textual_header(...), especially in the case where the textual header is not pre-set but created blank inside the function.

Not sure if the format of the revision number is mandatory, but the documentation in SEG Y rev 1 (table 1 in http://www.seg.org/Portals/0/SEG/News%20and%20Resources/Technical%20Standards/seg_y_rev1.pdf) suggest to set it in record 39 (bytes 3040-3120) on the format "C39 SEG Y REV 1"

Actually the fact that these two pieces of information are mandatory is only mentioned in footnote 6 (see table 1 and page 6 in pdf file linked to above) hence can easily be overlooked

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -93,6 +93,8 @@ master: (doi: 10.5281/zenodo.165135)
       `obspy.io.segy.segy.iread_segy` and `obspy.io.segy.segy.iread_su`.
       (see #1400).
     * Write correct revision number (see #1737).
+    * Textual headers will now always contain the file revision number and the
+      end header mark if nothing else exists at these positions (see #1738).
     * The SEG-Y format detection now also checks the format version number
       (see #1781).
  - obspy.io.css:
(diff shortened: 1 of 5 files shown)


krischer:
This is now a PR hopefully fixing it. @paitor can you please review and check the files? I essentially checked against the files I created myself as none of the couple of text files I checked actually have them set...

megies:
A rebase on master should get rid of Travis fails..

krischer:
I implemented the requested changes - can IMHO be merged once CI passes.

@paitor: Thanks for the review!

PR #1669, opened 2017-04-01
Docstring rendering problem for `plot_radiation_pattern()` and `Event.plot()`

krischer:
Docstrings for [plot_radiation_pattern()](http://docs.obspy.org/master/packages/autogen/obspy.imaging.source.plot_radiation_pattern.html#obspy.imaging.source.plot_radiation_pattern) and [Event.plot()](http://docs.obspy.org/master/packages/autogen/obspy.core.event.event.Event.plot.html#obspy.core.event.event.Event.plot) currently render very large fonts in some parts. 

Furthermore, it looks that, for `plot_radiation_pattern()` there should be a list for the `kind` parameter, which is not rendered properly.

### DIFF
--- a/obspy/core/event/event.py
+++ b/obspy/core/event/event.py
@@ -173,17 +173,19 @@ class Event(__Event):
         :param kind: A list of strings (for a 1-row plot) or a nested list of
             strings (one list of strings per row), with the following keywords
             to generate a matplotlib figure:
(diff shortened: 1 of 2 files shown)


krischer:
Now a PR fixing it. I think our sphinx theme has issues rendering bullet lists inside the parameter descriptions so I kind of worked around it. Should be good enough for now.

+DOCS

PR #1745, opened 2017-04-01
Cleaning up the renaming of io.datamark -> io.win

krischer:
Removes a couple of stray leftovers in the code base and most importantly it should now render the new `io.win` docs.

+DOCS

### DIFF
--- a/misc/docs/source/packages/index.rst
+++ b/misc/docs/source/packages/index.rst
@@ -38,7 +38,6 @@ The functionality is provided through the following packages:
    obspy.io.ah
    obspy.io.ascii
    obspy.io.css
-   obspy.io.datamark
    obspy.io.gcf
    obspy.io.gse2
    obspy.io.kinemetrics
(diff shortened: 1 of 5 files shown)


krischer:
Here is the renaming PR: https://github.com/obspy/obspy/pull/1717

megies:
@ThomasLecocq, off topic for this PR, just as a side note:

`_read_win(filename, century='20', **kwargs)`

See century handling I use in reftek 130 plugin: Setting two-digit years 00-49 -> 20XX and 50-99 -> 19XX 
 - https://github.com/obspy/obspy/blob/master/obspy/io/reftek/core.py#L89-L96
 - https://github.com/obspy/obspy/blob/master/obspy/io/reftek/util.py#L42-L57

ThomasLecocq:
all good for me. The year handling seem good, should I add it now or later is ok ?

megies:
> all good for me. The year handling seem good, should I add it now or later is ok ?

Later or never, both fine for me. Just wanted to mention since I stumbled over this line. It's not very critical I think, people should notice right away if they're data are displayed with e.g. '2097' when it should be '1997'.. ;-)

PR #1746, opened 2017-04-01
Raise exception when attempting to read very large mini-SEED files

krischer:
The issue is already described here: https://github.com/obspy/obspy/pull/1419

It will likely be a while until we can properly fix this and multiple people ran into this issue and thus ObsPy now raises a proper exception message with instructions on how to still read these files if it encounters this problem.

Error message is: `"ObsPy can currently not directly read mini-SEED files that are larger than 2^31 bytes (2048 MiB). To still read it,  please read the file in chunks as documented here:  https://github.com/obspy/obspy/pull/1419#issuecomment-221582369"`

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -104,6 +104,8 @@ master: (doi: 10.5281/zenodo.165135)
      InternalMSEEDReadingWarning now called InternalMSEEDWarning as both
      can now also be raised in non-reading contexts (see #1658).
    * Should no-longer segfault with arbitrarily truncated files (see #1728).
+   * Will now raise an exception when attempting to directly read mini-SEED
+     files larger than 2048 MiB (#1746).
  - obspy.io.nlloc:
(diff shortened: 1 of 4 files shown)


krischer:
appveyor is acting up. Can IMHO be ignored for this PR.

Jollyfant:
Is the branch name `raise-exception-when-reading-very-large-seed-file` too long for AppVeyor to digest or something? 😋

megies:
> Is the branch name raise-exception-when-reading-very-large-seed-file too long for AppVeyor to digest or something? 😋

Looks more like it's looking in the wrong repo..

```
git clone -q --branch=raise-exception-when-reading-very-large-mseed-file https://github.com/obspy/obspy.git C:\projects\obspy
```

But branch is on @krischer's fork.. ?!

krischer:
I think its because I accidentially pushed to the obspy fork at one point because I forgot to set the remote tracking branch. Still a bug in appveyor.

Issue #1754, opened 2017-04-12
AH (Ad Hoc) write support


s-schneider:
Currently there is read support for AH files, but no write support. I just wrote a short script, based on the obspy.io.ah module to write an read-in AH file as an AH-file again, but I have issues with the header information.

[https://github.com/s-schneider/nmPy/blob/master/nmpy/util/writeah.py
](https://github.com/s-schneider/nmPy/blob/master/nmpy/util/writeah.py)

barsch:
I wrote the AH reader using those header files: https://github.com/obspy/obspy/tree/master/obspy/io/ah/docs which itself were extracted from ftp://www.orfeus-eu.org/pub/software/mirror/ldeo.columbia/ - I couldn't find any other proper documentation about AH 

also be aware there are two versions of AH - version 2 is as far as I know nowhere used and lacks proper real live example files

hope it helps,
Robert

s-schneider:
thanks! It kind of works, but only if i specify the format
`st = read('foo.AH, format='AH')`

Otherwise I have problems with the _get_ah_version where the fixed values do not match. Do you have a clue what the purpose of these is?
```
elif magic == 6:
    # AH1 has no magic variable :/
    # so we have to use some fixed values as indicators
    try:
        fh.seek(12)
        if xdrlib.Unpacker(fh.read(4)).unpack_int() != 6:
            return False
        fh.seek(24)
        if xdrlib.Unpacker(fh.read(4)).unpack_int() != 8:
            return False
        fh.seek(700)
        if xdrlib.Unpacker(fh.read(4)).unpack_int() != 80:
            return False
        fh.seek(784)
        if xdrlib.Unpacker(fh.read(4)).unpack_int() != 202:
            return False
    except:
        return False
    return '1.0'
```

barsch:
I really don't remember how I extracted this information - could be buggy ...

its probably somewhere buried deep inside ftp://www.orfeus-eu.org/pub/software/mirror/ldeo.columbia/ah_alpha.tar.Z - I rember looking especially on C files in ah2asc

also the comments on this PR https://github.com/obspy/obspy/pull/851 may help ...

barsch:
ah i remember now - those fields are length indicator for the following string - as those strings are fixed, the length values are therefore also fixed:

_ah1.f_
![ah1](https://cloud.githubusercontent.com/assets/602450/3756069/f66facd8-182a-11e4-866b-57921aa87792.png)

[...]

PR #1757, opened 2017-04-14
Ah write support

s-schneider:
### What does this PR do?
Adds write support for AH (Ad Hoc) format to stream.write()
related to issue #1754
### PR Checklist
- [x] All tests still pass.


krischer:
Please let us know once you deem it ready so we can review it.

s-schneider:
It should be ready

megies:
Restarted CI, these SSL errors were transient, I hope..

krischer:
The continuous integration tests still fail. Can you please have a look?

megies:
I'm pretty sure this magic-y filesuffix handling is to blame.
 - https://github.com/obspy/obspy/pull/1757/files#diff-573d6f56616c93ae6dbf07a7697abdd2R241
 - https://github.com/obspy/obspy/pull/1757/files#diff-573d6f56616c93ae6dbf07a7697abdd2R441

There's too much tampering of read/write functions with filename suffix. Just use what the user specifies as filename and leave it untouched, I would say..

megies:
@s-schneider we need some action on this PR otherwise we'll have to postpone it for 1.2.0.

s-schneider:
what can I do to help solving the issue with the continuous integration test? I can't figure out the link to my changes

megies:
> I can't figure out the link to my changes

In `core` we have some tests that iterate over all waveform plugins, so there might actually be a connection to your PR. You could run that file `obspy\core\tests\test_waveform_plugins.py` locally with `pytest --pdb` and debug the problem.. if that's too much I can have a look at some later point

s-schneider:
I am pretty sure I'm doing something wrong here, even if I try that test on the master branch it fails

megies:
If you switch branches that have changes in setup.py (plugin entry points), be sure to rerun pip install..

I have a debugger open and that test fails for the AH plugin

s-schneider:
I rerun pip install and ran `pytest -- pdb obspy\core\tests\test_waveform_plugins.py`.
But they still fails in both banches

megies:
Ah, problem is the calibration factor. I guess ah format doesn't save a calibration factor with the waveforms?

megies:
@s-schneider I guess `Trace.stats.calib` is just not used by ah read/write, so I guess you can just add ah format to the formats that don't use it here, so that that test doesnt complain:
https://github.com/obspy/obspy/blob/master/obspy/core/tests/test_waveform_plugins.py#L143

s-schneider:
Alright, thanks for the help!

megies:
@s-schneider please amend that last commit, removing those anaconda files again and force-push, thanks!

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -61,7 +61,7 @@ def test_read_and_write(self):
         for format in formats:
             # XXX: skip SEGY and SU formats for now as they need some special
             # headers.
-            if format in ['SEGY', 'SU', 'SEG2']:
+            if format in ['SEGY', 'SU', 'SEG2', 'AH']:
```
megies: please don't skip here, but only below where the `calib` is checked, line number as indicated by the link I posted.

megies:
Yep, that should fix that test fail.. let's see what CI says

s-schneider:
still the same error with the calibration

s-schneider:
apparently `xdrlib` messes up the float number, after the second decimal place, when using `pack_float`/`unpack_float`

megies:
Aaargs.. this format really serializes sampling interval as floating point type?? Ugly. That means that basically *every* seismological time series (OK, not if you're working on free oszillations of the Earth, necessarily ;-)) will be represented wrongly with an error in sampling frequency.

I guess we could change the sampling interval we us in our waveform plugin test case to something that can be represented as a float without error to work around this..  any objections.. @krischer?

That would mean changing [that line](https://github.com/obspy/obspy/blob/master/obspy/core/tests/test_waveform_plugins.py#L84) and accordingly [that line](https://github.com/obspy/obspy/blob/master/obspy/core/tests/test_waveform_plugins.py#L150) to something like `tr.stats.delta = 0.25`.

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -151,12 +151,12 @@ def test_read_and_write(self):
                         self.assertEqual(st[0].stats.starttime, start)
                         if format == 'AH':
                             self.assertEqual(st[0].stats.delta, 0.25)
-                            self.assertEqual(st[0].stats.endtime,
-                                                start + 499.75)
+                            cend = start + 499.75
+                            self.assertEqual(st[0].stats.endtime, cend)
```
megies: just FYI, would've also worked to just align it under the bracket.. that's what PEP8 was complainign about
```
                            self.assertEqual(st[0].stats.endtime,
                                             start + 499.75)
```

s-schneider:
There is a SchemaError left in signal

megies:
@s-schneider are you sure those fields are UTF-8 encoded? Not ASCII?

s-schneider:
The read function decodes UTF-8 here:
https://github.com/obspy/obspy/blob/master/obspy/io/ah/core.py#L115

megies:
Alright.. I guess there's no other documentation than these C header files?

s-schneider:
Not that I know of, I mainly checked the documentation @barsch recommended
ftp://www.orfeus-eu.org/pub/software/mirror/ldeo.columbia/ 
and 
https://github.com/obspy/obspy/tree/master/obspy/io/ah/docs

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -81,7 +81,10 @@ def test_read_and_write(self):
                     tr.stats.location = "00"
                     tr.stats.channel = "EHE"
                     tr.stats.calib = 0.199999
-                    tr.stats.delta = 0.005
+                    if format == 'AH':
```
krischer: Sorry for joining the discussion a bit late - any particular reason to special case this? Why not just do it for all formats? Does it break for some formats?

obspy/io/ah/core.py
```diff
@@ -228,6 +227,217 @@ def _unpack_trace(data):
...
+    """
+    Writes a Stream object to an AH v1 waveform file.
+
+    :type stream:
+    :param stream: The ObsPy Stream object to write.
+    :type filename: str
+    :param filename: open file, or file-like object
+
+    """
+    def _pack_trace_with_ah_dict(trace, packer):
```
krischer: Could you move both out of the `_write_ah1()` function? This is IMHO better style and would potentially allow their direct usage/unit testing.

setup.py
```diff
@@ -264,6 +264,7 @@
     'obspy.plugin.waveform.AH': [
         'isFormat = obspy.io.ah.core:_is_ah',
         'readFormat = obspy.io.ah.core:_read_ah',
+        'writeFormat = obspy.io.ah.core:_write_ah1'
```
krischer: I don't know anything about AH versioning. I assume there are other AH versions as well? In that case - do you think it would make sense to have an `AH1` plugin?, e.g.

```python
'obspy.plugin.waveform.AH1': [
    'writeFormat = obspy.io.ah.core:_write_ah1'
]
```

The result is that writing AH files would be with `st.write("FILENAME", format="AH1")` instead of `format="AH"` like it currently is.

obspy/core/tests/test_waveform_plugins.py
s-schneider: AH uses floats not double for the delta, therefore the test fails using a delta of 0.005

setup.py
s-schneider: there is another format: AH2, which is apparently never used. The AH files in AH1 format I saw so far all used the AH or ahx suffix

obspy/core/tests/test_waveform_plugins.py
megies: Yeah.. we could probably just switch all tests over to this float-representable delta.. might be easier on maintenance to have less if/elses.

setup.py
megies: Renaming the write part to AH1 would complicate our roundhouse test (read 'AH' but needs to write 'AH1').. maybe it can stay this way and if ever somebody decides to implement the other one we just have 'AH' and 'AH2'? But whatever you guys decide on is fine with me..

obspy/core/tests/test_waveform_plugins.py
krischer: Yes I got that from the conversion. But maybe just change the delta to 0.25 (or 0.001 which can also be exactly represented by 32 bit floats) for all formats? That would reduce the complexity of this test.

setup.py
krischer: okay - as long as ah2 is essentially never used I'm totally fine with this.

s-schneider:
I'm a little confused by the changelog conflict

megies:
Quick fix for changelog merge conflict in online editor.. quick and dirty ;-)

krischer:
Thanks a lot for your work and patience on this :) Looks good to me.

One last issue I want to raise is the fairly confusing history of this PR - should we squash it or at least condense it a bit?

s-schneider:
Thanks for helping me with all the small and big issues :)

Yes I agree, we should condense it a bit, what do you have in mind and how can we do it?

megies:
I'm cleaning up history of this PR right now..

s-schneider:
Thank you!

obspy/io/ah/core.py
```diff
@@ -228,6 +227,219 @@ def _unpack_trace(data):
...
+
+def _pack_trace_with_ah_dict(tr, packer):
+
+    # station info
+    packer.pack_int(6)
+    packer.pack_fstring(6, tr.stats.ah.station.code.encode('utf-8'))
+    packer.pack_int(6)
+    try:
+        packer.pack_fstring(6, tr.stats.channel.encode('utf-8'))
+    except:
```
megies: code smell.. that try/except was added when there was no encoding yet and it would have never triggered, but due to the encoding being added this is now catching any encoding problems that it's not supposed to catch..

I'll fix it..

obspy/io/ah/core.py
```diff
@@ -257,7 +257,10 @@ def _pack_trace_with_ah_dict(trace, packer):
         packer.pack_int(CODESIZE)
         packer.pack_fstring(CODESIZE, tr.stats.ah.station.code)
         packer.pack_int(CHANSIZE)
-        packer.pack_fstring(CHANSIZE, tr.stats.ah.station.channel)
+        try:
+            packer.pack_fstring(CHANSIZE, tr.stats.channel)
+        except:
+            packer.pack_fstring(CHANSIZE, tr.stats.ah.station.channel)
```
megies: @s-schneider what is the logic behind this? Packing/unpacking should always go via `Trace.stats.station/channel/...` so I do not see the need to lookup those codes from within `tr.stats.ah`.

obspy/io/ah/core.py
```diff
@@ -228,6 +227,219 @@ def _unpack_trace(data):
...
+    packer.pack_float(len(poles))
+    packer.pack_float(0)
+    packer.pack_float(len(zeros))
+    packer.pack_float(0)
+
+    for _i in range(1, 30):
+        try:
+            r, i = poles[_i].real, poles[_i].imag
+        except IndexError:
+            r, i = 0, 0
```
megies: @s-schneider this looks really odd.. I assume this `0, 0` is supposed to be a place holder for "no pole"/"no zero"?

obspy/io/ah/core.py
s-schneider: The AH files usually store some event information as well, therefore I wrote two functions. If the AH file was read all those meta-informations go into `trace.stats.ah`. Was not sure how to handle it in a proper way

obspy/io/ah/core.py
s-schneider: yes, that was my intention

megies:
I've cleaned up history and added some more cleanup on top.

@s-schneider can you please double-check. Especially please have a look at https://github.com/obspy/obspy/pull/1757/commits/8b8fab71680a945a66267c35ae7cfd12032ae0d1. I believe that `Trace.stats.station/channel` should always be used rather than the values from the `Trace.stats.ah` dictionary, otherwise this is working around some of our core functionality.

megies:
Also, this commit was done probably to make flake8 happy but it also removes some logic, so I really liked those constants better than all those numbers spread across the code..

But anyway, cosmetics, I guess so it can also stay as is..

https://github.com/obspy/obspy/pull/1757/commits/1b2d9d8f5cfc215ea03c1e2b4bb77adf7dba3abd

s-schneider:
same for me, but I think flake8 was just complaining about the constants all being upper case?

megies:
All upper case for constants should be fine. Judging from your commit message "remove unused variables", my guess is that it would have been enough to simply remove those few unused ones (i.e. TYPEMIN, TYPEMAX, LOGENT, NEXTRAS, NOCALPTS)..

s-schneider:
Should I put the constants back in?

megies:
> Should I put the constants back in?

Wouldn't hurt, if you don't mind.. otherwise this is good to go, I think.

megies:
OK, I'll squash the last few commits and rebase once more on master, then it's good to go.

btw @s-schneider you can use a flake8 pre-commit hook for automatic flake8 check before commiting (see http://flake8.pycqa.org/en/latest/user/using-hooks.html; hooks can be skipped with `git commit -n` which is sometimes necessary for obspy e.g. when making changes in `__init__.py` files because we use a customized flake8 config).

s-schneider:
@megies Thanks for the recommendation, I'm going to use it!

s-schneider:
I just found out, that I should change one thing in this function: currently the poles and zeros of the response are not written to the file. I already have the solution, what would be a good way to implement it? Reopen the PR?

QuLogic:
This PR is merged; please open a new one.

PR #1759, opened 2017-04-24
appveyor: upload wheels to pypi

xoviat:
You will need to set your username and password with the appveyor secure variables web tool.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,18 +33,22 @@ environment:
     - PYTHON: "C:\\Miniconda3"
       PYTHON_VERSION: "3.4"
       PYTHON_ARCH: "32"
+      SKIP_NOTAG: "true"
 
     - PYTHON: "C:\\Miniconda3-x64"
       PYTHON_VERSION: "3.4"
       PYTHON_ARCH: "64"
+      SKIP_NOTAG: "true"
 
     - PYTHON: "C:\\Miniconda35"
       PYTHON_VERSION: "3.5"
       PYTHON_ARCH: "32"
+      SKIP_NOTAG: "true"
 
     - PYTHON: "C:\\Miniconda35-x64"
       PYTHON_VERSION: "3.5"
       PYTHON_ARCH: "64"
+      SKIP_NOTAG: "true"
 
     - PYTHON: "C:\\Miniconda36"
       PYTHON_VERSION: "3.6"
@@ -72,7 +76,7 @@ install:
         Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
           throw "There are newer queued builds for this pull request, failing early." }
   - ps: |
-      If (($env:PYTHON_VERSION -eq "3.5" -or $env:PYTHON_VERSION -eq "3.4") -and ($env:APPVEYOR_REPO_TAG -ne "true")) {
+      If (($env:SKIP_NOTAG -eq "true") -and ($env:APPVEYOR_REPO_TAG -ne "true")) {
           throw "Skipping Python version, not a tag."
       }
   # - ECHO "Filesystem root:"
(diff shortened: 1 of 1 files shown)


krischer:
Having wheel's would be great.

If I understand correctly thought this will upload to pypi every single appveyor run? Should this not include some kind of test if it is a releasable version or something?

barsch:
also it enables more enviroments tests which we disabled by purpose as the whole building takes ages already - wheel should only produced once we release

xoviat:
> If I understand correctly though this will upload to pypi every single appveyor run? Should this not include some kind of test if it is a releasable version or something?

Nope. This only uploads when the tag is set and the tests pass.

megies:
So far deploying for windows on pypi is done manually by @barsch, not sure if this will mix well with manually deplyoment or get confusing..? But I'll leave this one to the windows people to decide.. ;-)

xoviat:
Also, appveyor doesn't seem to have fired for the tags. You may need to do git tag and git push --tags rather than doing the web interface.

megies:
> Also, appveyor doesn't seem to have fired for the tags. You may need to do git tag and git push --tags rather than doing the web interface.

The description of `APPVEYOR_REPO_TAG` in https://www.appveyor.com/docs/environment-variables/ honestly doesn't sound very convincing. When is it actually set? Does it work when pushing a tag at some later time after pushing the respective commit? Are we even triggering appveyor builds on tag pushes??

xoviat:
It's set for a 'tag build'. You can see an example of that [here](https://ci.appveyor.com/project/xoviat/pypiwin32/history). 

image

appveyor.yml
```diff
@@ -100,9 +117,13 @@ test_script:
...
   - "python -m obspy.scripts.runtests --no-flake8 -n appveyor-ci -r %CI_URL% %PR_URL%"
 
-matrix:
-  # enable fast fail strategy
-  fast_finish: true
+
+on_success:
+  - ps: |
+      If ($env:APPVEYOR_REPO_TAG -eq "true") {
+          python setup.py sdist bdist_wheel
```
barsch: if we already upload wheels we could also build and upload ```bdist_wininst``` and ```bdist_egg``` too

appveyor.yml
```diff
@@ -58,6 +71,10 @@ install:
         https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
         Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
           throw "There are newer queued builds for this pull request, failing early." }
+  - ps: |
+      If (($env:PYTHON_VERSION -eq "3.5" -or $env:PYTHON_VERSION -eq "3.4") -and ($env:APPVEYOR_REPO_TAG -ne "true")) {
```
barsch: ist it possible to have an environment variable e.g. ```SKIP_IF_TAG``` within the matrix instead of checking against each version we want to skip here? Would be easier for adding later other/newer environments ...

xoviat: Will do.

xoviat:
> I wonder about pre-releases e.g. 1.0.3rc1 - could you filter them out too?
Not trivially, but pip won't install pre-releases anyway.

appveyor.yml
barsch: any thoughts on the comment above - or do you really dislike egg/exe file that much so that you dont even bother to comment on it ? ^^

barsch:
oh I forgot - you should add yourself to the contributers list

appveyor.yml
```diff
@@ -1,6 +1,8 @@
 # AppVeyor.com is a Continuous Integration service to build and run tests under Windows
 # see https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml
 
+clone_depth: 50
```
megies: Shallow git clones will lead to wrong/misleading version numbers in test reports because our version number lookup depends on `git-describe`. I don't see how this change is needed for uploads to pypi. I really think this change should be reverted.

xoviat: Addressed in #1774.

megies:
@xoviat can you please unbreak our appveyor commit status again? Appveyor thinks the build is failing while you're trying to skip it. Please fix this on master. See e.g.
 - https://ci.appveyor.com/project/obspy/obspy/build/1.0.4952-constructor
 - https://ci.appveyor.com/project/obspy/obspy/build/1.0.4952-constructor/job/btx1v4xhywavirm4

xoviat:
@megies Sorry for the delayed response. That job failed becuase [this](https://ci.appveyor.com/project/obspy/obspy/build/1.0.4952-constructor/job/y2na6hb784qyd5l6) failed. That has nothing to do with skipping jobs.

PR #1761, opened 2017-04-25
build_taup_model fails with simple regional model

krischer:
Installed ObsPy v1.0.3 on Ubuntu 16.04 using Anaconda with python 3.4.5.

A simplistic 2-layer model file [cn01.tvel](https://github.com/obspy/obspy/files/955726/cn01.tvel.txt):
```
CN01 - P: Stevens, Milne, Wetmiller, & Horner (1972) Canadian Earthquakes - 1966
CN01 - S: 
    0.000      6.2000      3.5700      2.7200
   36.000      6.2000      3.5700      2.7200
   36.000      8.2000      4.7000      3.3198
 6371.000      8.2000      4.7000      3.3198
```

Produces the following error using `obspy.taup.taup_create.build_taup_model`:

``` ipython
In [1]: from obspy.taup.taup_create import build_taup_model
In [2]: build_taup_model('cn01.tvel')
Building obspy.taup model for 'cn01.tvel' ...

  File "<ipython-input-7-0f8429869a68>", line 1, in <module>
    build_taup_model('cn01.tvel')

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/taup_create.py", line 181, in build_taup_model
    mod_create.run()

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/taup_create.py", line 121, in run
    self.tau_model = self.create_tau_model(self.v_mod)

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/taup_create.py", line 86, in create_tau_model
    _DEFAULT_VALUES["slowness_tolerance"])

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/slowness_model.py", line 76, in __init__
    self.create_sample()

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/slowness_model.py", line 153, in create_sample
    self.find_critical_points()

  File "/home/nackerle/anaconda3/lib/python3.4/site-packages/obspy/taup/slowness_model.py", line 440, in find_critical_points
    self.radius_of_planet, len(self.v_mod), -1, -1)

IndexError: index 2 is out of bounds for axis 0 with size 2
```

Yet same model file works fine with java TauP-2.4.3.

``` bash
$ taup_create -tvel CN01.tvel
TauP_Create starting...
Done!
$ taup_time -mod cn01 -h 18 -deg 1

Model: CN01
Distance   Depth   Phase   Travel    Ray Param  Takeoff  Incident  Purist    Purist
  (deg)     (km)   Name    Time (s)  p (s/deg)   (deg)    (deg)   Distance   Name 
-----------------------------------------------------------------------------------
    1.00    18.0   p         18.14    17.678     98.71    80.29     1.00   = p    
    1.00    18.0   Pn        19.20    13.484     48.93    48.75     1.00   = Pn   
    1.00    18.0   PcP       19.88    16.069     63.96    63.64     1.00   = PcP  
    1.00    18.0   s         31.51    30.701     98.71    80.29     1.00   = s    
    1.00    18.0   Sn        33.39    23.525     49.24    49.05     1.00   = Sn   
    1.00    18.0   ScS       34.53    27.907     63.96    63.64     1.00   = ScS  
```
A workaround when using  `obspy.taup.taup_create.build_taup_model` is to add an extra layer at any depth between 36 km and the earth's radius. The output is then:

``` ipython
In [3]: build_taup_model('cn01.tvel')
Building obspy.taup model for 'cn01.tvel' ...

In [4]: from obspy.taup import TauPyModel

In [5]: model = TauPyModel('cn01')

In [6]: for arrival in model.get_travel_times(18, 1):
   ...:    print(arrival)
p phase arrival at 18.143 seconds
Pn phase arrival at 19.202 seconds
PcP phase arrival at 19.884 seconds
sP phase arrival at 22.054 seconds
ScP phase arrival at 23.029 seconds
PcS phase arrival at 26.410 seconds
s phase arrival at 31.509 seconds
Sn phase arrival at 33.395 seconds
ScS phase arrival at 34.533 seconds
```

My expectation, however, is that anything which works with the java TauP toolkit should work with `obspy.taup`.

Is there a better way to implement simple local and regional velocity models using `obspy.taup` and/or the java TauP toolkit? It would be nice to be able to suppress the nonsensical core phases, for example.


krischer:
@nackerley Thanks for reporting! This is now a PR fixing the issue - it was just that we created an internal array with too little entries in those cases. Making it larger does not seem to hurt anything else as the tests still pass.

Sorry for highjacking your comment...converting this to a PR somehow now makes Github think that I opened the issue - this was not intentional.

nackerley:
Just imagine if every hijacking came with a repair job. Thanks!

One outstanding question: Is there a better way to implement simple local and regional velocity models using obspy? It would be nice to be able to suppress the nonsensical core phases, for example, or to be able to specify an "infinite radius" i.e. a layered half-space. The two are probably connected (the generation of core phases and the enforced implementation of a spherical model). At this point I think we'd be going beyond the functionality of the java TauP, so maybe that's a problem.

krischer:
> One outstanding question: Is there a better way to implement simple local and regional velocity models using obspy? It would be nice to be able to suppress the nonsensical core phases, for example, or to be able to specify an "infinite radius" i.e. a layered half-space. The two are probably connected (the generation of core phases and the enforced implementation of a spherical model). At this point I think we'd be going beyond the functionality of the java TauP, so maybe that's a problem.

That will probably be very difficult as TauP works in spherical coordinates. @crotwell do you have any comment on that?

Note that you can just pass the `phase_list` argument to most of the functions and it will only return the specified phases which should do what you want it to do.

crotwell:
Humm, I would say that modifying TauP to work in a flat earth/infinite radius style is probably a decent amount of work for not much gain. Modifying the phase list to get rid of PcP for these situations is simple and probably sufficient to take care of at least that part of the issue. Another problem with flat earth is that it would happily generate arrivals well beyond the distance at which they are appropriate. 

I feel a better solution for the local-distance use case for TauP is to paste a crust/upper mantle model onto an existing global model. So even though the model only specifies a crust and mantle velocity, you could do something like this:
```
    0.000      6.2000      3.5700      2.7200
   36.000      6.2000      3.5700      2.7200
   36.000      8.2000      4.7000      3.3198
   210.000    8.3000    4.5180    3.4258   <--- paste iasp91 from here down
   210.000    8.3000    4.5220    3.4258
   260.000    8.4825    4.6090    3.4561
...
```
You have your crust for direct phases and mantle velocity for Pn, but if the model is "accidentally" used for further distances, you have iasp91 the rest of the way down giving reasonable answers for mantle turning P or for deeper phases like PcP. TauP tries to find the core mantle boundary automatically and gets confused if there is no core. In the simple model, the only discontinuity is the moho, so it ends up being the core-mantle boundary for phase parsing. Kind of weird but that is the tradeoff for dealing with a wide variety of earth (and non-earth) models. Maybe adding a way to automatically do the pasting would be useful. I have thought about doing this, but never gotten around to it.

nackerley:
Thanks for the insightful comments @crotwell. You're absolutely right; in most if not all of our applications it makes a lot of sense to prepare for "accidental" use of local/regional velocity models at large distances. Initially I was annoyed by TauP's insistence on spherical symmetry, but we don't live in flatland, so I should accept it.

@krischer phase_list is a good workaround in some cases but not others; in my case I will probably be stripping any arrival with 'c' in its name from the list of arrivals, after the fact. 

I'm still bothered that TauP (in both the obspy and java implementations) is identifying the lone discontinuity as both the moho (first discontinuity? discontinuity nearest to 30 km?) and the core (last discontinuity?). Should the outer and inner core not also be identified  by transitions from non-zero to zero S-wave velocity and back? 

I think the magic happens in  [VelocityModel.fix_discontinuity_depths](https://docs.obspy.org/_modules/obspy/taup/velocity_model.html#VelocityModel.fix_discontinuity_depths). It could make sense to leave `cmb_depth` at `radius_of_planet` when the S-wave velocity is nowhere zero, instead of setting it at the discontinuity closest to the `radius_of_planet` which in this case means it floats all the way up to `moho_depth`.

crotwell:
One point to remember is that TauP is not always used for earth models. I have had Mars and the Moon show up already in bug reports, so you can't assume a liquid outer core. It might be worth considering additional sanity checks on the major discontinuities, but it is tricky when someone actually wants to do something that is not "earth-like", asteroid seismology perhaps! And as far as TauP is concerned, your model actually looks like a "solid core" with a thin crust, or maybe a core-less planet with only a mantle and a crust. But it is kind of hard for TauP to tell. I think you can argue that it is actually doing the best it can with the limited information it has. That is actually one reason I like the "nd" file format over the "tvel" as you get to specify the discontinuities directly in the model file.

Can you give specific example of a phase that shows up that doesn't make sense? May be able to give some advice if I know the specifics. Ideally you specify the exact phases you care about and only those show up, so I would be surprised if a PcP showed up when you didn't ask for it.

nackerley:
The model that started this PR is a good example. I don't think PcP , ScP, PcS or ScS should be being generated; there's no liquid layer, so there is no outer core.

A simple regional model [cn01.tvel](https://github.com/obspy/obspy/files/955726/cn01.tvel.txt) has the moho mis-identified as the CMB, resulting in suppression of Sn and other phases which require S to propagate below the moho, as well as incorrect incident angles for ScP and PcS:
cn01a

With CN01 grafted onto IASP91 starting at the Lehmann discontinuity [cn01b.nd](https://github.com/obspy/obspy/files/962780/cn01b.nd.txt) the correct result is obtained ('c' and 'i' phases have been suppressed, for clarity):
cn01b

If the simple model [cn01.tvel](https://github.com/obspy/obspy/files/955726/cn01.tvel.txt) was treated by `obspy.taup.taup_create.build_taup_model` as having a core-mantle boundary at the radius of the planet, then I think this problem might disappear.

It makes no difference whether the model is implemented as .nd or .tvel. 
[cn01a.nd](https://github.com/obspy/obspy/files/962793/cn01a.nd.txt) gives the same result as  [cn01.tvel](https://github.com/obspy/obspy/files/955726/cn01.tvel.txt).

crotwell:
I think I have updated TauP to accommodate coreless and even crustless models. The default now is to push the cmb down in a tvel model instead of allowing it to be the same as the moho if there are no other discontinuities. TauP is also now better about respecting labels in .nd models, the moving of the boundaries now happens only on .tvel models, so if you screw up an .nd model, you are on your own. I think all this is most likely an improvement, at least all my tests pass and you can now create even weirder earth models. But "garbage in, garbage out" is still the case.

Prerelease is here, please test and let me know if you see any issues. I will probably wait a few days to let things settle before pushing out the release:
http://www.seis.sc.edu/downloads/TauP/prerelease/TauP-2.4.4-SNAPSHOT.tgz

@krischer commit is here. Kind of large and I bungled the first commit and had to fix the fix, so need to look at both if you want to port it over.
https://github.com/crotwell/TauP/commit/b712c614f9a589b3beb6524096e03230a460a914
https://github.com/crotwell/TauP/commit/5f0e9cb5bd8b9798e023fa0d09348d557587e78d

@nackerley I still strongly! recommend pasting crustal models onto a realistic global model like prem or iasp91 instead of just a constant. To do otherwise I feel is just asking for trouble at some point in the future when you forget how far the model can be pushed. That said I think you now have enough rope to shoot yourself in the foot if you would like to! :)

crotwell:
@krischer found some issues with edge cases, so one more commit you might want to look at
https://github.com/crotwell/TauP/commit/243b39c89d1501e94d5e0824e7e4b447bf948e9b

PR #1763, opened 2017-04-26
plot_response degrees option

bonaime:
Option 'deg' for plotting plot_response in degrees (radian is default)
- [x] Tests

bode_degrees


krischer:
Looks good but it does need a test case. Please generate the test image with the `v2.0.x` branch on matplotlib's github.

megies:
Hmm.. OT.. maybe we should at some point consider using a nicer style as default as well, that image up there looks better than our default right now, I think (Axes, Tick setup etc)..

krischer:
> Hmm.. OT.. maybe we should at some point consider using a nicer style as default as well, that image up there looks better than our default right now, I think (Axes, Tick setup etc)..

Our plots already look nicer on matplotlib 2 - we force the styles only for the tests.

megies:
> Our plots already look nicer on matplotlib 2 - we force the styles only for the tests.

Ah! Didn't know.. nice!

bonaime:
Image test for Response 
response_response_degrees

QuLogic:
All the other parameters use full words; I think `deg` might be too short/unclear, maybe `plot_degrees` or something?

obspy/core/inventory/channel.py
```diff
@@ -333,6 +333,8 @@ def plot(self, min_freq, output="VEL", start_stage=None, end_stage=None,
             amplitude/phase spectrum. If not specified, a new figure is opened.
         :type unwrap_phase: bool
         :param unwrap_phase: Set optional phase unwrapping using NumPy.
+        :type deg: bool
+        :param deg: if True plot bode in degrees
```
megies: two back ticks for formatting:
```
``True``
```
in docstring will be displayed by sphinx as monotyped:

``True``

obspy/core/inventory/inventory.py
```diff
@@ -728,6 +728,8 @@ def plot_response(self, min_freq, output="VEL", network="*", station="*",
             opened.
         :type unwrap_phase: bool
         :param unwrap_phase: Set optional phase unwrapping using NumPy.
+        :type deg: bool
+        :param deg: if True plot bode in degrees
```
megies: see above

obspy/core/inventory/response.py
```diff
@@ -1294,6 +1294,8 @@ def plot(self, min_freq, output="VEL", start_stage=None,
             that should be plotted times two.
         :type unwrap_phase: bool
         :param unwrap_phase: Set optional phase unwrapping using NumPy.
+        :type deg: bool
+        :param deg: if True plot bode in degrees
```
megies: and again, backticks

obspy/core/inventory/station.py
```diff
@@ -438,6 +438,8 @@ def plot(self, min_freq, output="VEL", location="*", channel="*",
             opened.
         :type unwrap_phase: bool
         :param unwrap_phase: Set optional phase unwrapping using NumPy.
+        :type deg: bool
+        :param deg: if True plot bode in degrees
```
megies: backticks..

bonaime:
back ticks and plot_degrees are in the new commit

krischer:
Travis failures are unrelated and dealt with in #1777.

krischer:
Thanks for this!

Issue #1554, opened 2016-10-13
Sporadic fails in SDS client


megies:
We're seeing sporadic fails in an SDS client test case, e.g. here: http://tests.obspy.org/54134/
Should be examined/fixed at some point..

PR #1770, opened 2017-04-29
Omit more warnings in unit/doctests

barsch:
also small bugfixes in sac and pde which resulted into warnings ...

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -74,18 +74,19 @@ class QuantityError(AttribDict):
         Boolean testing for QuantityError.
 
         QuantityError evaluates ``True`` if any of the default fields is not
-        ``None``.
+        ``None``. Setting non default fields raises also an UserWarning which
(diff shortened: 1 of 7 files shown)


barsch:
failing tests are unrelated - see #1554 - imho ready to merge

barsch:
@jkmacc-LANL if ```header.get(hdr)``` is numeric you will get an ```Unicodewarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal```

jkmacc-LANL:
Oi, that might be a typo.  The relative time headers are all floats, so I think comparison should be against the float null value, `HD.FNULL`.  Yup, I'm pretty sure it should be `HD.FNULL`.  If you make that change with the original `in ( )` syntax, do the warnings go away?

barsch:
will check once I'm at home

jkmacc-LANL:
Thanks for the catch!

PR #1773, opened 2017-05-02
Variable string encoding for SAC files 

### LINKED ISSUE #1768: SAC reading problem
Hi Obspy developer,

I tried to use the Obspy 1.0.3 to read a SAC file but it throw an error at me as shown below:

```
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-6-91703a32c963> in <module>()
----> 1 st = obspy.read('1100-5001.sac.ds.bp.sml.2152.01.sml', debug_headers=True)

<decorator-gen-144> in read(pathname_or_url, format, headonly, starttime, endtime, nearest_sample, dtype, apply_calib, **kwargs)

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/core/util/decorator.py in _map_example_filename(func, *args, **kwargs)
    297                         except IOError:
    298                             pass
--> 299         return func(*args, **kwargs)
    300     return _map_example_filename
    301

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/core/stream.py in read(pathname_or_url, format, headonly, starttime, endtime, nearest_sample, dtype, apply_calib, **kwargs)
    229         pathname = pathname_or_url
    230         for file in sorted(glob(pathname)):
--> 231             st.extend(_read(file, format, headonly, **kwargs).traces)
    232         if len(st) == 0:
    233             # try to give more specific information why the stream is empty

<decorator-gen-145> in _read(filename, format, headonly, **kwargs)

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/core/util/decorator.py in uncompress_file(func, filename, *args, **kwargs)
    207     else:
    208         # no compressions
--> 209         result = func(filename, *args, **kwargs)
    210     return result
    211

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/core/stream.py in _read(filename, format, headonly, **kwargs)
    271     """
    272     stream, format = _read_from_plugin('waveform', filename, format=format,
--> 273                                        headonly=headonly, **kwargs)
    274     # set _format identifier for each element
    275     for trace in stream:

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/core/util/base.py in _read_from_plugin(plugin_type, filename, format, **kwargs)
    404         raise TypeError(msg % (format_ep.name, ', '.join(eps)))
    405     # read
--> 406     list_obj = read_format(filename, **kwargs)
    407     return list_obj, format_ep.name
    408

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/io/sac/core.py in _read_sac(filename, headonly, debug_headers, fsize, **kwargs)
    349             return _internal_read_sac(buf=fh, headonly=headonly,
    350                                       debug_headers=debug_headers, fsize=fsize,
--> 351                                       **kwargs)
    352     else:
    353         raise ValueError("Cannot open '%s'." % filename)

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/io/sac/core.py in _internal_read_sac(buf, headonly, debug_headers, fsize, **kwargs)
    382     """
    383     # read SAC file
--> 384     sac = SACTrace.read(buf, headonly=headonly, ascii=False, checksize=fsize)
    385     # assign all header entries to a new dictionary compatible with an ObsPy
    386     tr = sac.to_obspy_trace(debug_headers=debug_headers)

/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy-1.0.3.post0+1068.g04848ed0fc.obspy.master-py3.6-macosx-10.7-x86_64.egg/obspy/io/sac/sactrace.py in read(cls, source, headonly, ascii, byteorder, checksize, debug_strings)
   1162                 if val.startswith(native_str('-12345')):
   1163                     val = HD.SNULL
-> 1164                 hs[i] = val
   1165
   1166         sac = cls._from_arrays(hf, hi, hs, data)

UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-7: ordinal not in range(128)
```

This issue has been a while since 1.0.x, so I tried to do it myself. Somehow, when I took your code and write something similar, it works properly. I am not exactly sure what happened to the Obspy version, but I will post my solution here in case anyone has the same issue as I had for a long time.

Here is my code:
```Python
"""Read SAC file into Numpy array"""
import numpy as np

# Read SAC
from future.utils import native_str
from . import header as HD
import os
import sys


def read(file_name, byteorder=None, checksize=False):
    '''Customed version of obspy sac reader

    '''

    # Check byte order
    is_byteorder_specified = byteorder is not None
    if not is_byteorder_specified:
        byteorder = sys.byteorder

    if byteorder == 'little':
        endian_str = '<'
    elif byteorder == 'big':
        endian_str = '>'
    else:
        raise ValueError("Unrecognized byteorder. Use {'little', 'big'}")

    with open(file_name, 'rb') as fh:
        # --------------------------------------------------------------
        # READ HEADER
        # The sac header has 70 floats, 40 integers, then 192 bytes
        #    in strings. Store them in array (and convert the char to a
        #    list). That's a total of 632 bytes.
        # --------------------------------------------------------------

        hf = np.array(memoryview(fh.read(4 * 70))
                      ).view(native_str(endian_str + 'f4')).copy()
        hi = np.array(memoryview(fh.read(4 * 40))
                      ).view(native_str(endian_str + 'i4')).copy()
        hs = np.array(memoryview(fh.read(24 * 8))
                      ).view(native_str('|S8')).copy()

        if not is_valid_byteorder(hi):
            if is_byteorder_specified:
                # specified but not valid. you dun messed up.
                raise ValueError("Incorrect byteorder {}".format(byteorder))
            else:
                # not valid, but not specified.
                # swap the dtype interpretation (dtype.byteorder), but keep the
                # bytes, so the arrays in memory reflect the bytes on disk
                hf = hf.newbyteorder('S')
                hi = hi.newbyteorder('S')

        # check header lengths
        if len(hf) != 70 or len(hi) != 40 or len(hs) != 24:
            hf = hi = hs = None
            raise ValueError("Cannot read all header values")

        npts = hi[HD.INTHDRS.index('npts')]

        # check file size
        if checksize:
            cur_pos = fh.tell()
            fh.seek(0, os.SEEK_END)
            length = fh.tell()
            fh.seek(cur_pos, os.SEEK_SET)
            th_length = (632 + 4 * int(npts))
            if length != th_length:
                msg = "Actual and theoretical file size are inconsistent.\n" \
                      "Actual/Theoretical: {}/{}\n" \
                      "Check that headers are consistent with time series."
                raise ValueError(msg.format(length, th_length))

        # --------------------------------------------------------------
        # READ DATA
        # --------------------------------------------------------------
        data = np.array(memoryview(fh.read(int(npts) * 4))
                        ).view(native_str(endian_str + 'f4')).copy()

        if len(data) != npts:
            fh.close()
            print("Cannot read all data points")

    fh.close()

    return data


def is_valid_byteorder(hi):
    nvhdr = hi[HD.INTHDRS.index('nvhdr')]
    return (0 < nvhdr < 20)

if __name__ == '__main__':
    pass
```

It requires this header.py which I took directly from Obspy:

```Python
# -*- coding: utf-8 -*-
                        unicode_literals)

MODULE_DOCSTRING = """
SAC header specification, including documentation.

Header names, order, types, and nulls, as well as allowed enumerated values,
are specified here.  Header name strings, and their array order are contained
in separate float, int, and string tuples.  Enumerated values, and their
allowed string and integer values, are in dictionaries.  Header value
documentation is in a dictionary, for reuse throughout the package.

"""

# header documentation is large and used in several places, so we just write
# it once here and distributed it as needed.
DOC = {'npts': 'N    Number of points per data component. [required]',
       'nvhdr': '''N    Header version number. Current value is the integer 6.
                  Older version data (NVHDR < 6) are automatically updated
                  when read into sac. [required]''',
       'b': 'F    Beginning value of the independent variable. [required]',
       'e': 'F    Ending value of the independent variable. [required]',
       'iftype': '''I    Type of file [required]:

                  * ITIME {Time series file}
                  * IRLIM {Spectral file---real and imaginary}
                  * IAMPH {Spectral file---amplitude and phase}
                  * IXY {General x versus y data}
                  * IXYZ {General XYZ (3-D) file}''',
       'leven': 'L    TRUE if data is evenly spaced. [required]',
       'delta': 'F    Increment between evenly spaced samples (nominal value).'
                ' [required]',
       'odelta': 'F    Observed increment if different from nominal value.',
       'idep': '''I    Type of dependent variable:

                  * IUNKN (Unknown)
                  * IDISP (Displacement in nm)
                  * IVEL (Velocity in nm/sec)
                  * IVOLTS (Velocity in volts)
                  * IACC (Acceleration in nm/sec/sec)''',
       'scale': 'F    Multiplying scale factor for dependent variable '
                '[not currently used]',
       'depmin': 'F    Minimum value of dependent variable.',
       'depmax': 'F    Maximum value of dependent variable.',
       'depmen': 'F    Mean value of dependent variable.',
       'nzyear': 'N    GMT year corresponding to reference time in file.',
       'nzjday': 'N    GMT julian day.',
       'nzhour': 'N    GMT hour.',
       'nzmin': 'N    GMT minute.',
       'nzsec': 'N    GMT second.',
       'nzmsec': 'N    GMT millisecond.',
       'iztype': '''I    Reference time equivalence:

                  * IUNKN (5): Unknown
                  * IB (9): Begin time
                  * IDAY (10): Midnight of reference GMT day
                  * IO (11): Event origin time
                  * IA (12): First arrival time
                  * ITn (13-22): User defined time pick n, n=0,9''',
       'o': 'F    Event origin time (seconds relative to reference time.)',
       'ko': 'K    Event origin time identification.',
       'a': 'F    First arrival time (seconds relative to reference time.)',
       'ka': 'K    First arrival time identification.',
       'f': 'F    Fini or end of event time '
            '(seconds relative to reference time.)',
       'kf': 'F    Fini or end of event time identification.',
       't0': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't1': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't2': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't3': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't4': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't5': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't6': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't7': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't8': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       't9': 'F    User defined time (seconds picks or markers relative to '
             'reference time).',
       'kt0': 'F    User defined time pick identification.',
       'kt1': 'F    User defined time pick identification.',
       'kt2': 'F    User defined time pick identification.',
       'kt3': 'F    User defined time pick identification.',
       'kt4': 'F    User defined time pick identification.',
       'kt5': 'F    User defined time pick identification.',
       'kt6': 'F    User defined time pick identification.',
       'kt7': 'F    User defined time pick identification.',
       'kt8': 'F    User defined time pick identification.',
       'kt9': 'F    User defined time pick identification.',
       'kinst': 'K    Generic name of recording instrument',
       'iinst': 'I    Type of recording instrument. [currently not used]',
       'knetwk': 'K    Name of seismic network.',
       'kstnm': 'K    Station name.',
       'istreg': 'I    Station geographic region. [not currently used]',
       'stla': 'F    Station latitude (degrees, north positive)',
       'stlo': 'F    Station longitude (degrees, east positive).',
       'stel': 'F    Station elevation (meters). [not currently used]',
       'stdp': 'F    Station depth below surface (meters). '
               '[not currently used]',
       'cmpaz': 'F    Component azimuth (degrees, clockwise from north).',
       'cmpinc': 'F    Component incident angle (degrees, from vertical).',
       'kcmpnm': 'K    Component name.',
       'lpspol': 'L    TRUE if station components have a positive polarity '
                 '(left-hand rule).',
       'kevnm': 'K    Event name.',
       'ievreg': 'I    Event geographic region. [not currently used]',
       'evla': 'F    Event latitude (degrees north positive).',
       'evlo': 'F    Event longitude (degrees east positive).',
       'evel': 'F    Event elevation (meters). [not currently used]',
       'evdp': 'F    Event depth below surface (meters). [not currently used]',
       'mag': 'F    Event magnitude.',
       'imagtyp': '''I    Magnitude type:

                  * IMB (52): Bodywave Magnitude
                  * IMS (53): Surfacewave Magnitude
                  * IML (54): Local Magnitude
                  * IMW (55): Moment Magnitude
                  * IMD (56): Duration Magnitude
                  * IMX (57): User Defined Magnitude''',
       'imagsrc': '''I    Source of magnitude information:

                  * INEIC (National Earthquake Information Center)
                  * IPDE (Preliminary Determination of Epicenter)
                  * IISC (International Seismological Centre)
                  * IREB (Reviewed Event Bulletin)
                  * IUSGS (US Geological Survey)
                  * IBRK (UC Berkeley)
                  * ICALTECH (California Institute of Technology)
                  * ILLNL (Lawrence Livermore National Laboratory)
                  * IEVLOC (Event Location (computer program) )
                  * IJSOP (Joint Seismic Observation Program)
                  * IUSER (The individual using SAC2000)
                  * IUNKNOWN (unknown)''',
       'ievtyp': '''I    Type of event:

                  * IUNKN (Unknown)
                  * INUCL (Nuclear event)
                  * IPREN (Nuclear pre-shot event)
                  * IPOSTN (Nuclear post-shot event)
                  * IQUAKE (Earthquake)
                  * IPREQ (Foreshock)
                  * IPOSTQ (Aftershock)
                  * ICHEM (Chemical explosion)
                  * IQB (Quarry or mine blast confirmed by quarry)
                  * IQB1 (Quarry/mine blast with designed shot
                    info-ripple fired)
                  * IQB2 (Quarry/mine blast with observed shot
                    info-ripple fired)
                  * IQMT (Quarry/mining-induced events:
                    tremors and rockbursts)
                  * IEQ (Earthquake)
                  * IEQ1 (Earthquakes in a swarm or aftershock
                    sequence)
                  * IEQ2 (Felt earthquake)
                  * IME (Marine explosion)
                  * IEX (Other explosion)
                  * INU (Nuclear explosion)
                  * INC (Nuclear cavity collapse)
                  * IO_ (Other source of known origin)
                  * IR (Regional event of unknown origin)
                  * IT (Teleseismic event of unknown origin)
                  * IU (Undetermined or conflicting information)
                  * IOTHER (Other)''',
       'nevid': 'N    Event ID (CSS 3.0)',
       'norid': 'N    Origin ID (CSS 3.0)',
       'nwfid': 'N    Waveform ID (CSS 3.0)',
       'khole': 'k    Hole identification if nuclear event.',
       'dist': 'F    Station to event distance (km).',
       'az': 'F    Event to station azimuth (degrees).',
       'baz': 'F    Station to event azimuth (degrees).',
       'gcarc': 'F    Station to event great circle arc length (degrees).',
       'lcalda': 'L    TRUE if DIST AZ BAZ and GCARC are to be calculated '
                 'from st event coordinates.',
       'iqual': '''I    Quality of data [not currently used]:

                  * IGOOD (Good data)
                  * IGLCH (Glitches)
                  * IDROP (Dropouts)
                  * ILOWSN (Low signal to noise ratio)
                  * IOTHER (Other)''',
       'isynth': '''I    Synthetic data flag [not currently used]:

                  * IRLDTA (Real data)
                  * ????? (Flags for various synthetic seismogram codes)''',
       'user0': 'F    User defined variable storage area 0.',
       'user1': 'F    User defined variable storage area 1.',
       'user2': 'F    User defined variable storage area 2.',
       'user3': 'F    User defined variable storage area 3.',
       'user4': 'F    User defined variable storage area 4.',
       'user5': 'F    User defined variable storage area 5.',
       'user6': 'F    User defined variable storage area 6.',
       'user7': 'F    User defined variable storage area 7.',
       'user8': 'F    User defined variable storage area 8.',
       'user9': 'F    User defined variable storage area 9.',
       'kuser0': 'K    User defined variable storage area 0.',
       'kuser1': 'K    User defined variable storage area 1.',
       'kuser2': 'K    User defined variable storage area 2.',
       'lovrok': 'L    TRUE if it is okay to overwrite this file on disk.'}

# because readable
_hdritems = sorted(DOC.items())
_hdrfmt = "{:10.10s} = {}"
HDR_DESC = '\n'.join([_hdrfmt.format(_hdr, _doc) for _hdr, _doc in _hdritems])

HEADER_DOCSTRING = """
============ ==== ===============================================
Field Name   Type Description
============ ==== ===============================================
""" + HDR_DESC + \
    "\n============ ==== ==============================================="

# Module documentation string
__doc__ = MODULE_DOCSTRING + HEADER_DOCSTRING

# ------------ NULL VALUES ----------------------------------------------------
FNULL = -12345.0
INULL = -12345
SNULL = '-12345  '

# ------------ HEADER NAMES, TYPES, ARRAY POSITIONS ---------------------------
# these are useful b/c they can be used forwards or backwards, like:
# FLOADHDRS.index('az') is 40, and FLOATHDRS[40] is 'az'.
FLOATHDRS = ('delta', 'depmin', 'depmax', 'scale', 'odelta', 'b', 'e', 'o',
             'a', 'internal0', 't0', 't1', 't2', 't3', 't4', 't5', 't6', 't7',
             't8', 't9', 'f', 'resp0', 'resp1', 'resp2', 'resp3', 'resp4',
             'resp5', 'resp6', 'resp7', 'resp8', 'resp9', 'stla', 'stlo',
             'stel', 'stdp', 'evla', 'evlo', 'evel', 'evdp', 'mag', 'user0',
             'user1', 'user2', 'user3', 'user4', 'user5', 'user6', 'user7',
             'user8', 'user9', 'dist', 'az', 'baz', 'gcarc', 'internal1',
             'internal2', 'depmen', 'cmpaz', 'cmpinc', 'xminimum', 'xmaximum',
             'yminimum', 'ymaximum', 'unused6', 'unused7', 'unused8',
             'unused9', 'unused10', 'unused11', 'unused12')

INTHDRS = ('nzyear', 'nzjday', 'nzhour', 'nzmin', 'nzsec', 'nzmsec', 'nvhdr',
           'norid', 'nevid', 'npts', 'internal3', 'nwfid', 'nxsize', 'nysize',
           'unused13', 'iftype', 'idep', 'iztype', 'unused14', 'iinst',
           'istreg', 'ievreg', 'ievtyp', 'iqual', 'isynth', 'imagtyp',
           'imagsrc', 'unused15', 'unused16', 'unused17', 'unused18',
           'unused19', 'unused20', 'unused21', 'unused22', 'leven', 'lpspol',
           'lovrok', 'lcalda', 'unused23')

STRHDRS = ('kstnm', 'kevnm', 'kevnm2', 'khole', 'ko', 'ka', 'kt0', 'kt1',
           'kt2', 'kt3', 'kt4', 'kt5', 'kt6', 'kt7', 'kt8', 'kt9', 'kf',
           'kuser0', 'kuser1', 'kuser2', 'kcmpnm', 'knetwk', 'kdatrd', 'kinst')

"""
NOTE:

kevnm also has a kevnm2 b/c it takes two array spaces.
'kevnm' lookups must be caught and handled differently.  This happens in the
SACTrace string property getters/setters, .io.dict_to_header_arrays
and .arrayio.header_arrays_to_dict.
"""
# NOTE: using namedtuples for header arrays sounds great, but they're immutable

# TODO: make a dict converter between {'<', '>', '='} and {'little', 'big'}

# ------------ ENUMERATED VALUES ----------------------------------------------
# These are stored in the header as integers.
# Their names and values are given in the mapping below.
# Some (many) are not used.
# TODO: this is ugly; rename things a bit
ENUM_VALS = {'itime': 1, 'irlim': 2, 'iamph': 3, 'ixy': 4, 'iunkn': 5,
             'idisp': 6, 'ivel': 7, 'iacc': 8, 'ib': 9, 'iday': 10, 'io': 11,
             'ia': 12, 'it0': 13, 'it1': 14, 'it2': 15, 'it3': 16, 'it4': 17,
             'it5': 18, 'it6': 19, 'it7': 20, 'it8': 21, 'it9': 22,
             'iradnv': 23, 'itannv': 24, 'iradev': 25, 'itanev': 26,
             'inorth': 27, 'ieast': 28, 'ihorza': 29, 'idown': 30, 'iup': 31,
             'illlbb': 32, 'iwwsn1': 33, 'iwwsn2': 34, 'ihglp': 35, 'isro': 36,
             'inucl': 37, 'ipren': 38, 'ipostn': 39, 'iquake': 40, 'ipreq': 41,
             'ipostq': 42, 'ichem': 43, 'iother': 44, 'igood': 45, 'iglch': 46,
             'idrop': 47, 'ilowsn': 48, 'irldta': 49, 'ivolts': 50, 'imb': 52,
             'ims': 53, 'iml': 54, 'imw': 55, 'imd': 56, 'imx': 57,
             'ineic': 58, 'ipdeq': 59, 'ipdew': 60, 'ipde': 61, 'iisc': 62,
             'ireb': 63, 'iusgs': 64, 'ibrk': 65, 'icaltech': 66, 'illnl': 67,
             'ievloc': 68, 'ijsop': 69, 'iuser': 70, 'iunknown': 71, 'iqb': 72,
             'iqb1': 73, 'iqb2': 74, 'iqbx': 75, 'iqmt': 76, 'ieq': 77,
             'ieq1': 78, 'ieq2': 79, 'ime': 80, 'iex': 81, 'inu': 82,
             'inc': 83, 'io_': 84, 'il': 85, 'ir': 86, 'it': 87, 'iu': 88,
             'ieq3': 89, 'ieq0': 90, 'iex0': 91, 'iqc': 92, 'iqb0': 93,
             'igey': 94, 'ilit': 95, 'imet': 96, 'iodor': 97, 'ios': 103}

# reverse look-up: you have the number, want the string
ENUM_NAMES = dict((v, k) for k, v in ENUM_VALS.items())


# accepted values, by header
ACCEPTED_VALS = {'iftype': ['itime', 'irlim', 'iamph', 'ixy'],
                 'idep': ['iunkn', 'idisp', 'ivel', 'ivolts', 'iacc'],
                 'iztype': ['iunkn', 'ib', 'iday', 'io', 'ia', 'it0', 'it1',
                            'it2', 'it3', 'it4', 'it5', 'it6', 'it7', 'it8',
                            'it9'],
                 'imagtyp': ['imb', 'ims', 'iml', 'imw', 'imd', 'imx'],
                 'imagsrc': ['ineic', 'ipde', 'iisc', 'ireb', 'iusgs', 'ipdeq',
                             'ibrk', 'icaltech', 'illnl', 'ievloc', 'ijsop',
                             'iuser', 'iunknown'],
                 'ievtyp': ['iunkn', 'inucl', 'ipren', 'ipostn', 'iquake',
                            'ipreq', 'ipostq', 'ichem', 'iqb', 'iqb1', 'iqb2',
                            'iqbx', 'iqmt', 'ieq', 'ieq1', 'ieq2', 'ime',
                            'iex', 'inu', 'inc', 'io_', 'il', 'ir', 'it', 'iu',
                            'iother'],
                 'iqual': ['igood', 'iglch', 'idrop', 'ilowsn', 'iother'],
                 'isynth': ['irldta']}

ACCEPTED_INT = ACCEPTED_VALS.copy()
for _hdr in ACCEPTED_INT:
    ACCEPTED_INT[_hdr] = [ENUM_VALS[_ival] for _ival in ACCEPTED_VALS[_hdr]]
```

Hope it may help.

-Lijun Zhu

lijunzh:
This is an attempt to add "encoding" flag to obspy.io.sac.satrace.read() function in order to read some SAC that has encoding other than "ASCII". 

### PR Checklist
- [x] All tests still pass.

Closes #1768.


obspy/io/sac/util.py
```diff
@@ -171,7 +171,7 @@ def _clean_str(value, strip_whitespace=True):
     Trace that the user may have manually added.
     """
     try:
-        value = value.decode('ASCII', 'replace')
+        value = value.decode(encode, 'replace')
```
megies: I think it would be a good idea to make this something like..

```python
try:
    value = value.decode(encoding, 'strict')
except UnicodeError:
    msg = ('Encountered characters that can not be '
           'decoded with encoding {}: {}\nThese characters '
           'are ignored.'.format(encoding, value))
    warnings.warn(msg)
    value = value.decode(encoding, 'replace')
except AttributeError:
    ...
```

obspy/io/sac/sactrace.py
```diff
@@ -1158,10 +1158,14 @@ def read(cls, source, headonly=False, ascii=False, byteorder=None,
...
-                val = _ut._clean_str(val, strip_whitespace=False)
+                if 'encode' in kwargs.keys():
+                    val = _ut._clean_str(val, encode=kwargs['encode'],
+                                         strip_whitespace=False)
+                else:
+                    val = _ut._clean_str(val, strip_whitespace=False)
                 if val.startswith(native_str('-12345')):
                     val = HD.SNULL
-                hs[i] = val
+                hs[i] = val.encode('ascii', 'replace')
```
megies: I can't follow this logic, why always encode to ASCII when the user might have specified a different encoding?

megies:
So right now you''re only handling reading files with a different encoding? I assume the idea is to regard files with an encoding different then ASCII as illegal files and only facilitate reading them while any output will be encoded to ASCII? I can follow that logic but then there should be an warning message I think when a user specifies a different encoding but output is ASCII encoded..

lijunzh:
Thank @megies for reviewing the pull request. I will answer your two review comments and one last comment together here.

1. As @megies mentioned previously 

> the main point is whether SAC is allowing an encoding different from ASCII. I only see it mentioned as "alphanumeric" in the SAC manual, so my guess is.. no.

I think that we should always reinforce the alphanumeric idea by using only ASCII encoding. This also helps the encoding problem between py2 and py3. (e.g. You will have an error when decoding special characters encoded in UTF-8 using ASCII, but not the other way around) Since

>A UTF-8 file that contains only ASCII characters is identical to an ASCII file.

from [Wikipedia](https://en.wikipedia.org/wiki/Comparison_of_Unicode_encodings)

2. Yes, I am kind of regarding any file encoded other than ASCII an illegal file. But by supporting the 'encode' flag, we give users the ability to read the content. However, I don't think any station/channel names with non-ASCII character is much useful (except the case when they are written in a different language, e.g. Chinese, Arabic, etc.) So, I am in favor of replacing those characters by '?' instead of passing them along to the stream object. More importantly, I don't want to let users write to a SAC that contains that kind of characters which may cause problems when they process it in a different program or pass to others. I would like Obspy only output/write to a SAC file in ASCII encoding. 

3. If you guys think that we should support "UTF-8" encoding throughout the obspy.io.sac, I would say that the solution in PR is a cheap band-aid sticker that we should avoid. However, it will involve changing a lot of codes in all places which is out of the scope of this PR. We can start a different PR to address that problem instead. 

4. The idea of the warning message is great. My initial thoughts are that when the user see a bunch of '????' in their stream object, they will know something wrong with the encoding processing. But throwing out an explicit warning message is a better way to go. I will change the PR accordingly.

megies:
I tend to agree with the general idea outlined above.

Default:
 - read and decode textual headers as ASCII, show useful error message explaining the option to specify a different encoding but that the file should be considered illegal (show that message when a UnicodeError woould get raised)
 - write and encode as ASCII

Spefifying an encoding:
 - use that encoding to read/decode textual headers
 - write and encode as ASCII and show warning message explaining the situation

lijunzh:
By the way, the current testing file `test_encode.sac` only has special characters in its textual headers. It will be nice if we can have a SAC file that has only alphanumeric characters but encoded in an encoding other than ASCII. This will allow us to test the encoding conversion though it should be any problem as far as I am reading the code.

jkmacc-LANL:
To reiterate what I said in the issue, changing a string header from unknown to '?????' by default is different behavior from the way the SAC program behaves (which seems to allow arbitrary byte strings to pass through unchanged).  This  may trip up some users, and I suspect will result in bug reports.  Also, I think our current test suite has some assumptions that a file can be read and re-written unchanged.

megies:
Right now byte strings can also not pass through unchanged.. and we never claim to be 100% sac compatible, so I don't see that as a show stopper. Showing meaningful warnings in these exotic and probably sac-format-breaching cases is enough I would say..

megies:
> I think our current test suite has some assumptions that a file can be read and re-written unchanged

I don't think so, as it looks like you're calling this string cleanup method that will always decode..

lijunzh:
> the way the SAC program behaves (which seems to allow arbitrary byte strings to pass through unchanged)

If we want to achieve that, we may need a class attribute that store the original headers in bytes. An alternative way to go is to record the user inputted encoding flag and use that when writing to SAC. Either way will involve changing the code structure which I am not comfortable to do as a newcomer to Obspy.


> our current test suite has some assumptions that a file can be read and re-written unchanged.

I double we have some testing suite that checks the integrity of SAC file (unchanged after read-write process). My tweaks to the code pass the test properly which clearly does not hold that principle.

lijunzh:
Can anyone give me some advice here? The Travis CI throw some tests fail like:

```
FAIL: test_catalog_plot_ortho (obspy.core.tests.test_event.CatalogBasemapTestCase)
```

which I didn't even mess with. All changes I made were in `obspy.io.sac` folder. I am not sure what went wrong here.

jkmacc-LANL:
@lijunzh That failure may be unrelated.  If you see it when you run the test suite on the `master` branch, than it's not from you.  I think you can look for that failure on other PRs, too, to see if it's from your changes.

megies:
My guess is that these image test fails come from slight changes in new matplotlib 2.0.1 which was finally released a few days ago. Baseline images were created some weeks ago on matplotlib 2.0.x branch..

CC @krischer @QuLogic

megies:
@lijunzh but there's also some "real" fails in io.sac: http://tests.obspy.org/79829/#1

megies:
In any case, I'm leaving this to @jkmacc-LANL to decide how to tackle this encoding issue in `io.sac`.

krischer:
> My guess is that these image test fails come from slight changes in new matplotlib 2.0.1 which was finally released a few days ago. Baseline images were created some weeks ago on matplotlib 2.0.x branch..
> 
> CC @krischer @QuLogic

Seems to be that the clipping the edges is slightly different - we can just up the tolerance a tiny bit - no need to recreate the test images IMHO. I'll do it in a separate PR.

obspy/io/sac/sactrace.py
```diff
@@ -1100,7 +1100,7 @@ def reftime(self, new_reftime):
     # --------------------------- I/O METHODS ---------------------------------
     @classmethod
     def read(cls, source, headonly=False, ascii=False, byteorder=None,
-             checksize=False, debug_strings=False):
+             checksize=False, debug_strings=False, **kwargs):
```
krischer: Instead of using `kwargs` - can you please directly specify the argument name in the function definition? This is IMHO much easier to read and understand.

lijunzh: That's actually easier for me to code. I was trying to keep all the code unchanged  until absolutely needed. If you guys have no problem with adding a new parameter to `read` in `obspy.io.sac.sactrace` and change the calling statement in `_internal_read_sac` in `obspy.io.sac.core`, I will be more than happy to do that.

krischer: Great :) Public methods (not starting with underscore) should be backwards compatible but just adding a new argument that defaults to the old behavior of course satisfies that constraint.

lijunzh: Nice. I will change the code accordingly.

lijunzh:
Can anyone give some insight what this test is for?

```Python
    def test_reftime_incomplete(self):
        """
        Replacement for SACTrace._from_arrays doctest which raises UserWarning
        """
        sac = SACTrace._from_arrays()
        self.assertTrue(sac.lcalda)
        self.assertFalse(sac.leven)
        self.assertFalse(sac.lovrok)
        self.assertFalse(sac.lpspol)
        self.assertEqual(sac.iztype, None)
        self.assertRaises(SacHeaderTimeError, getattr, sac, 'reftime')
        # raises "UserWarning: Reference time information incomplete"
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter('always', UserWarning)
            str(sac)
            self.assertEqual(len(w), 1)
            self.assertIn("Reference time information incomplete", str(w[0]))
```

My code fails at 
```Python
self.assertEqual(len(w), 1)
```
in which I don't know why we want to keep `len(w)` equals 1.

krischer:
It is probably useful to either jump to the line with a debugger or just print the caught warnings. Maybe some other warning got triggered as well? Tests usually assert the number of caught warnings as they intentionally trigger only one warning.

lijunzh:
I kind of throw multiple warnings for different content when using the "encoding" flag. It will first warn user that even they specify an encoding, it will use ASCII as output. The second one tell user that what characters are replaced by "?" when conversion happens. It that triggers this test to fail, I can merge them.

jkmacc-LANL:
This test is to verify that the user gets exceptions when they explicitly request time header information that isn't there, and to warn them with they implicitly request it (`str(sac)`).   I only expect one warning, but since you're now expecting more than one, I think it's fair to change this test.

lijunzh:
In fact, when I tested on my machine, I don't see this fail. It correctly delivers only one warning: 
```
'Reference Time = XX/XX/XX (XXX) XX:XX:XX.XXXXXX\n\tiztype not set\nlcalda     = True\nleven      = False\nlovrok     = False\nlpspol     = False'
/Users/lijun/anaconda3/lib/python3.6/site-packages/obspy/io/sac/sactrace.py:1454: UserWarning: Reference time information incomplete.
  warnings.warn(msg)
```

More importantly, Travis CI shows that it only fails on py27 and py33. I then created an env for py27 and repeat the test. Here are my results:
```Python
(py27) lijun@Lijuns-MacBook-Pro:~$ python
Python 2.7.13 |Continuum Analytics, Inc.| (default, Dec 20 2016, 23:05:08)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import obspy
>>> sac = obspy.io.sac.sactrace.SACTrace._from_arrays()
>>> sac
SACTrace(lcalda=1, leven=0, lovrok=0, lpspol=0)
>>> str(sac)
/Users/lijun/anaconda3/envs/py27/lib/python2.7/site-packages/obspy/io/sac/sactrace.py:1441: UserWarning: Reference time information incomplete.
  warnings.warn(msg)
'Reference Time = XX/XX/XX (XXX) XX:XX:XX.XXXXXX\n\tiztype not set\nlcalda     = True\nleven      = False\nlovrok     = False\nlpspol     = False'
>>> import warnings
>>> with warnings.catch_warnings(record=True) as w:
...     warnings.simplefilter('always', UserWarning)
...
>>> len(w)
0
>>> type(w)
<type 'list'>
>>> w
[]
```
Apparently, py27 treat `w` as an empty list. This leads me to think that maybe the `warning.catch_warning` behavior changes after py33. But this is not caused by the changes made in this PR. 

@jkmacc-LANL Maybe we should fix this problem in a different PR. You may have more insight in this test.

obspy/io/sac/sactrace.py
```diff
@@ -1231,7 +1248,7 @@ def _from_arrays(cls, hf=None, hi=None, hs=None, data=None):
         .. rubric:: Example
 
         >>> sac = SACTrace._from_arrays()
-        >>> print(sac)  # doctest: +NORMALIZE_WHITESPACE +SKIP
+        >>> print(sac)  # doctest: +NORMALIZE_WHITESPACE
```
jkmacc-LANL: If this test is giving the PR problems, and it was previously skipped, maybe just put the `+SKIP` back in and I'll make a new PR for it.

lijunzh:
@jkmacc-LANL Thanks a lot for reviewing the PR. I am grateful to have so much help from the Obspy community for my first PR. I will make the changes accordingly.

obspy/io/sac/core.py
```diff
@@ -381,7 +381,12 @@ def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
     :return: A ObsPy Stream object.
     """
     # read SAC file
-    sac = SACTrace.read(buf, headonly=headonly, ascii=False, checksize=fsize)
+    if 'encoding' in kwargs.keys():
```
lijunzh: I didn't know that we can do this. It will be much cleaner.

obspy/io/sac/core.py
```diff
@@ -381,7 +381,12 @@ def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
     :return: A ObsPy Stream object.
```
lijunzh: I have second thoughts here. I don't think we want to make "encoding" flag a standard parameter nor we want to encourage users to use it. Its existence is due to the fact that bad SAC files are created when people passing files from one system to another. When they are facing this encoding problem like I have (i.e. someone maybe read the file in Windows system and didn't encode it correctly before writing it to SAC. That's why I have a lot of characters in CP 1252), we have a backup plan to guide them using an "encoding" flag by throwing a user warning. I did document the "encoding" parameter in `obspy.io.sac.sactrace` which I regard as the lowest level that we can hide its existence. 

```
 :param encoding: By default, ASCII encoding is used. ASCII-characters
            encoded in a different encoding scheme will be converted to ASCII
            while other special characters will be replaced by '?'.
 :type encoding: str
```

If you guys think it is better to have it as a standard parameter, I have no problem with it. But it seems to be unnecessary changes to the documents and Obspy functionality that most people will not use nor they want to konw about.

obspy/io/sac/sactrace.py
```diff
@@ -1158,10 +1162,23 @@ def read(cls, source, headonly=False, ascii=False, byteorder=None,
                                             checksize=checksize)
         if not debug_strings:
             for i, val in enumerate(hs):
-                val = _ut._clean_str(val, strip_whitespace=False)
+                if encoding.lower() != 'ascii':     # throw warning
+                    msg = ("Using encoding other than ASCII: {}\n"
+                           "All non-ASCII characters will be replaced by "
+                           "?".format(encoding))
+                    warnings.warn(msg)
+                val = _ut._clean_str(val, encoding=encoding,
```
lijunzh: Good catch. The parameter positions need to be consistent.

jkmacc-LANL: Oops, the order it doesn't technically matter here.

obspy/io/sac/core.py
jkmacc-LANL: Good thought.  I'm OK with that.

lijunzh: That's very nice of you. I will go ahead and change the rest of the PR and make a push now.

lijunzh:
@jkmacc-LANL The PR complies with all PEP8 standard and has all test passed. There is one more red cross here:

> codecov/project — No report found to compare against

Basically, the current commit only fixes a PEP8 problem without change any code effectively. That may be the reason why there is no report generated for comparison.  I am not sure if it is important or not, but I think this PR is ready for merge.

obspy/io/sac/sactrace.py
```diff
@@ -1231,7 +1262,7 @@ def _from_arrays(cls, hf=None, hi=None, hs=None, data=None):
         .. rubric:: Example
 
         >>> sac = SACTrace._from_arrays()
-        >>> print(sac)  # doctest: +NORMALIZE_WHITESPACE +SKIP
+        >>> print(sac)  # doctest: +NORMALIZE_WHITESPACE + SKIP
```
megies: this was garbled by your auto-code formatter I assume, please revert to make it `+SKIP`

obspy/io/sac/core.py
```diff
@@ -381,7 +381,8 @@ def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
...
     """
     # read SAC file
-    sac = SACTrace.read(buf, headonly=headonly, ascii=False, checksize=fsize)
+    sac = SACTrace.read(buf, headonly=headonly, ascii=False, checksize=fsize,
+                        encoding=kwargs.get('encoding', 'ASCII'))
```
megies: This kind of hides the default encoding `ASCII` in the belly of this routine.. I don't think it has a downside to move it up as `def _internal_read_sac(..., encoding='ASCII', **kwargs)`. Even though it's an internal low level function not to be used directly by the user there's no need to hide away things, I think. This feels a bit like security by obscurity.. The docstrings should just state that the `encoding` parameter should not be changed normally and that it might produce an 'illegal' SAC file, if people want to use it then it's on their own doing..

megies:
Two minor nit picks from my side but overall it looks good to me (although I didn't play around with it interactively). Looks like it shows meaningful error/warning messages in all non-standard scenarios, and that's my only vital concern (besides using ASCII by default).

No idea why we don't have a codecov report uploaded for the commit in master that it wants to compare against.. but you can just change the first commit in the link shown above to use a different commit on current master to get a proper comparison:
https://codecov.io/gh/obspy/obspy/compare/cdfc16c70d5f8ca350f25fe0aa9b013642818db4...d92a982c3fde96f5201e968c4f9569824a87ecdd

Coverage looks good to me, all lines with the new encoding stuff are covered.

Actually, one last thing.. is it possible now to write to file with different encoding? If so, might be good to have a test for it, similar to the new encoding read test.

obspy/io/sac/sactrace.py
```diff
@@ -1158,10 +1176,23 @@ def read(cls, source, headonly=False, ascii=False, byteorder=None,
                                             checksize=checksize)
         if not debug_strings:
             for i, val in enumerate(hs):
-                val = _ut._clean_str(val, strip_whitespace=False)
+                if encoding.lower() != 'ascii':  # throw warning
+                    msg = ("Using encoding other than ASCII: {}\n"
+                           "All non-ASCII characters will be replaced by "
+                           "?".format(encoding))
```
krischer: Is this error message correct? Should it not rather be `"Using non-ASCII encoding: XXX. All characters that cannot be decoded with it will be replaced by '?'"`? Also maybe warn only when this actually happens? e.g. when it fails to decode some string.

obspy/io/sac/tests/test_core.py
```diff
@@ -915,6 +917,19 @@ def test_always_sac_reftime(self):
         self.assertAlmostEqual(tr1.stats.sac.a, a, places=5)
         self.assertEqual(tr1.stats.sac.b, b)
 
+    def test_encode_flag(self):
+        """
+        Read SAC file using ASCII encoding by default. User can specify
+        encoding through encoding flag.
+        """
```
krischer: Please catch all the warnings here - depending on what you use to execute the tests you might not see them but they will clobber the test output.

obspy/io/sac/sactrace.py
krischer: I just noticed that the `_clean_str()` method does already raise a warning in that case. So maybe just remove that warning here?

lijunzh: Those are two different warnings. The warning here tells users that they are using encoding other than "ASCII" which is not ideal and non-ASCII character will be converted.  The warning in `_clean_str()` was to tell users that the conversion actually happened. I agree that throwing two warnings here may be a bit too verbose. I will think about how to merge them.

krischer: Also the warning is IMHO not necessary as the user made a conscious choice of the encoding - so no need to inform her again.

obspy/io/sac/core.py
lijunzh: OK, if you guys all think that we should make "encoding" available at a higher level, I have no problem with that. I can make it a parameter for `_internal_read_sac` and default to 'ASCII'.

obspy/io/sac/sactrace.py
lijunzh: I was trying to guide the user using those warnings. But, if we decided to document 'encoding' flag in the higher level function `_internal_read_sac`, there is truly no need to warn a conscious user about the obvious thing. I will change it along with the docstring.

lijunzh:
@megies 

> is it possible now to write to file with different encoding? If so, might be good to have a test for it, similar to the new encoding read test.

I think it is definitely possible to add 'encoding' to `sactrace.write`. However, I still have doubt if we want to implement that for 'write'. Since SAC prefers using alphanumeric characters which are completely covered by 'ASCII', there is no much need to use any encoding other than 'ASCII'. Furthermore, it is beneficial to keep the output of Obspy's SAC file encoded only in 'ASCII' which will avoid encoding problem with other programs. But it is certainly OK to offer 'encoding' flag to write function and default it to 'ASCII'. If that is the direction we agree to go, I will add extra codes to `io.sac` module.

jkmacc-LANL:
I think it should be possible, though maybe not the default, to allow a non-ASCII encoding to be passed through without conversion.  This SAC program _does_ allow non-ASCII encodings to be read and written, and I think it's useful to do the same.  This would imply adding an encoding keyword to `SACTrace.write`, if you're willing to, @lijunzh .

lijunzh:
Sure, I can add 'encoding' flag to write funciton.

obspy/io/sac/core.py
lijunzh: @megies @jkmacc-LANL @krischer, I have a quick question for you guys. `_internal_read_sac` is called by the following two function. If we want to add a parameter 'encoding' to it, do I need to add that parameter to these two functions as well or `**kwargs` will take care it for me?

`obspy.core.stream._read`:
```Python
    stream, format = _read_from_plugin('waveform', filename, format=format,
                                       headonly=headonly, **kwargs)
```

`obspy.core.util.base._read_from_plugin`:
```Python
   # read
```

jkmacc-LANL: If I understand you correctly, `**kwargs` is intended to carry any format-specific arguments to the specific reader, so `_internal_read_sac` should get the `encoding` keyword from the `kwargs` dictionary.

lijunzh: @jkmacc-LANL  Which means if I change `_internal_read_sac` 
From 
```Python
def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
                       **kwargs):  # @UnusedVariable
```
To
```Python
def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
                       encoding='ASCII', **kwargs):  # @UnusedVariable
```
I don't need to worry about breaking those two functions in the `obspy.core` module?

krischer: Nope - that is fine - as long it is an optional argument with a default value.

krischer: Also note that the `_read_sac()` function is the one that is called by the `obspy.core` function and thus the highest level function that should have the `encoding` parameter. (all plug-in entry points are defined in the root level `setup.py`).

QuLogic:
I don't think we should add support for writing in alternative encodings so long as we believe it's against the standard. We should not proliferate bad files just because one software does something bad. There's a [saying in networking](https://en.wikipedia.org/wiki/Robustness_principle): Be conservative in what you do, be liberal in what you accept from others. If someone really needs it, they can monkey-patch the writing in some way, but so long as we have no real requirement for it, we should not add it.

jkmacc-LANL:
@lijunzh Sorry, the master branched changed recently from another merged PR.  The changes look isolated from your work, and I think you can just rebase on on top of master to resolve this, after you [sync your fork](https://help.github.com/articles/syncing-a-fork/).  Does anyone see any issue with doing this?  (Actually, the web editor looks maybe easier.  I think you just have to prefer the code blocks from `master` instead of from your branch.)
```bash
# Make sure your work is committed.
# Assumes you have obspy/master as a remote called "upstream", 
# and your remote is called "origin".
git fetch upstream
git checkout master
git merge upstream/master
# Your local master is up to date with obspy/master.
git checkout sac_string_encoding_flag
git rebase master
# Your feature branch commits are just replayed as though you branched from the up-to-date master.
# They have new commit hashes, unfortunately.
# Push your new hashes, which clobbers the old ones.
git push --force origin
```

jkmacc-LANL:
@QuLogic I can appreciate your conservatism on this issue, as you are responsible for many features in ObsPy.  My suggestion comes from knowing that colleagues just down the hall from me will be frustrated that they can't simply add origin times to their SAC files without also mangling station names.  I think the conservative choice here is actually adding the feature.  The defaults follow the spec, but there is a release valve for non-confirming files.

megies:
@jkmacc-LANL so your colleagues are really working with non-ASCII non-SEED-compatible station names? Phew.. must be tough to handle in many places..?

Is passing through non-ASCII in read-and-write situations really working right now? Surely not via `Trace` so probably via `SACTrace` only?

jkmacc-LANL:
@megies That's not quite what I mean.  Many people use the SAC program, and don't necessarily pay attention to how the strings are encoded because the program just passes them through.  ObsPy throws an exception (hence this Issue/PR).  In persuading SAC users to consider ObsPy, I think there should at least be a way to replicate the behavior they're used to (no warnings, no mangling).

megies:
@jkmacc-LANL I was mostly asking because I think that allowing arbitrary strings to pass through..

 - will never be possible when going via `read`/`Stream`
 - is not even possible right now using lower level `SACTrace`

Correct me if I'm wrong, but I think facilitating this (at least for `SACTrace`) would be a major rehaul of internal SAC reading routines and `SACTrace` as well, whereas this PR is rather minimally invasive?

megies:
@jkmacc-LANL so I understand your feeling that we should enable writing non-ASCII / passing through arbitrary bytes is not coming from an actual use case of people you know and who really work with non-ASCII encodings but rather from potential problems with it in the future? That about right?

*If that is correct*, I would propose that for short-term (i.e. 1.1.0) we stick with the read-only non-ASCII functionality of this PR and we put writing/passing-through non-ASCII on the agenda for 1.2.0.

How does that sound?

jkmacc-LANL:
@megies You're right, this PR is small in scope (which is a Good Thing ™️ ).  I don't think that supporting non-ASCII encodings is a large task, but I won't ask @lijunzh to do that here.  A later PR sounds good to me.  I'm not happy with how the string encoding/decoding is spread across different files anyway, and had some ideas about consolidating.  Thanks for humoring me 😉 

@lijunzh Did we answer all your questions so far?

lijunzh:
Mostly, I was kind of waiting for everyone to pitch in. My view of this PR is more of a bug fix which prevents my colleagues and me using obspy for reading (broken) SAC files. The thing that matters most to me is to change 
```Python
hs[i] = val
```
in ` sactrace.read` 

```Python
hs[i] = val.encode('ascii', 'replace')
```


The original line works perfectly as designed in Python 2x since the default encoding is 'ASCII'; however, it triggers problems in Python 3 whose default encoding is 'UTF-8'. 


The idea of adding an 'encoding' flag is very appealing since it gives users the choice to take the risk when reading files that may be encoded differently. However, I second @jkmacc-LANL that passing encoding flag through multiple files is not what I want to implement this flag. 

My original plan was to hide it in the deepest level so that only people need it will dig the code and find a way to do it properly. But that's very bad practice not to document an existing feature. I would love to have only that line changes in this PR so that we can make it to v1.1 and solve the issue. Further functionality can be implemented in a different PR. I have no problem doing that if you guys let me.  

I like @QuLogic 's comment about software robustness. I would argue that we want to support reading SAC files that are encoded differently from ASCII but not so much for writing to them. I, myself, am on the more conservative side of having features that may get in the way of someone's workflow. However, we can still implement the writing feature and let users make their best judgment of whether or not using it.

megies:
🎉  Alright, so let's do it like this then.. get this PR in as is in 1.1.0 and postpone eventual polishing and/or writing non-ASCII to 1.2.0

lijunzh:
@megies, @jkmacc-LANL, @krischer  If everyone is on board with this plan, I will go ahead and change/push the code.

megies:
So I guess this is ready for people to share a final review if they want..

CC @QuLogic @krischer

megies:
Oh @lijunzh maybe add a short changelog?

lijunzh:
@megies I don't like the implementation of 'encoding' flag now and will trim it off for the v1.1 release. Let's figure out a better way to do it later. The only code that will change is the line that I mentioned in the previous comment:




> change
> 
> hs[i] = val
> in sactrace.read
> 
> to
> 
> hs[i] = val.encode('ascii', 'replace')

jkmacc-LANL:
So, apparently, I have no idea how to use GitHub's review workflow.  I saw nothing troubling in the changes.  As long as @lijunzh addresses the requests from others and fixes those conflicts, this looks fine to me.

lijunzh:
I am working on a version of minimum changes now and will push it shortly. Please take a look at that. Let's postpone the 'encoding' flag to the v1.2

obspy/io/sac/sactrace.py
```diff
@@ -1092,7 +1175,7 @@ def read(cls, source, headonly=False, ascii=False, byteorder=None,
                 val = _ut._clean_str(val, strip_whitespace=False)
                 if val.startswith(native_str('-12345')):
                     val = HD.SNULL
-                hs[i] = val
+                hs[i] = val.encode('ascii', 'replace')
```
lijunzh: This is the only changes I intended to make. Somehow, there appears to be a large number of changes after merging with master branch from upstream as advised by @jkmacc-LANL .

jkmacc-LANL: Hopefully, [this comment](https://github.com/obspy/obspy/pull/1773#issuecomment-299992461) should help you update your branch.

lijunzh:
@krischer @jkmacc-LANL @megies This is the minimal version of this PR for v1.1. It only alters the behavior when non-ASCII bytes are encountered:

Previously: non-ASCII bytes are converted to UTF-8 "?" in py3 due to the default encoding. It then triggers an exception when 
```Python
val = val.encoding()
```
Now: Those characters are forced being replaced by ASCII "?" in this line:
```Python
val = val.encode('ASCII', 'replace')
```

Please review this PR. Let's implement the 'encoding' flag in the later version. This PR should fix the SAC reading problem in py3 which is of great value to my research team.

jkmacc-LANL:
I would recommend squashing these commits down to a much smaller number of logical commits, as the changes ended up being quite minimal.  Something like `git rebase -i HEAD~28` would start the process.  Other than that, this looks fine to me.  Thanks!

lijunzh:
Delete the old branch to remove merge commits.

lijunzh:
Sorry that I am not very familiar with the github process. I guess I will just submit another PR to clean up the mess.

lijunzh:
I've restored the branch. Can anyone help me figure out how to rebase the changes?

jkmacc-LANL:
Are you trying to squash commits, or something else?

megies:
> Can anyone help me figure out how to rebase the changes?

If you're working on command line, then like @jkmacc-LANL stated: `$ git rebase -i master` (or even safer `$ git rebase -i upstream/master` -- where 'upstream' would be the alias the main obspy repo has in your local git clone)
Afterwards push with `-f` for "force" (because it will overwrite existing history of this branch)

If you're not comfortable with all that just let me know, we can also do it, no problem.

lijunzh:
@jkmacc-LANL  Yes, I am trying to squash the commits. 
@megies Thanks for offering the help. I tried it my best and solved tons of conflicts. Here is the best I can get. Let me know if this is what you guys expected. I'd love to learn to use git properly.

lijunzh:
@jkmacc-LANL All tests are passed. I think we can merge the PR now.

jkmacc-LANL:
I don't think this can be merged until @krischer 's review is cleared.

QuLogic:
Something is wrong with the rebase here; there are commits that were authored by other people, but committed by @lijunzh. I think you may have merged with latest `master`, but then rebased against some old commit. If you want to rebase again, make sure that both `master` and `upstream/master` are up to date and in sync, then rebase against it.

krischer:
@jkmacc-LANL You should be able to overwrite missing reviews but as @QuLogic mentioned: something went wrong with the rebase.

megies:
@lijunzh it might be easiest to start again from the previous state of the branch, you might be able to find out the hash of the former tip of the branch by looking at `$ git reflog`.

If you need help, maybe post the newest 20 lines lines from that output and we can help..

jkmacc-LANL:
(@krischer By "overwrite missing reviews", do you mean "Dismiss review"?)

lijunzh:
@QuLogic You are totally right. I did merge with master and then took @jkmacc-LANL's advice to rebase my commits. I think I may mess something up during the process so I asked help previously. 

@megies Attached is the output of that command. I did rebase multiple times so that log is messed with all kinds of "rebase" history. 

[git_reflog.txt](https://github.com/obspy/obspy/files/993265/git_reflog.txt)

megies:
Looking at the reflog you posted, this is the state before the rebase that went wrong, I think:
https://github.com/obspy/obspy/commits/637696bec

Personally, I would recommend to just raze your branch, start again from current master and manually apply the diff of this page (i.e. the final state):
https://github.com/obspy/obspy/compare/master...637696bece2bfbee2648aeb42f186ba8eecb6deb

lijunzh:
@megies I did try to erase the branch and start all over again. But that automatically close my PR and I asked how to reopen it before. I was only able to reopen my PR once I restore my old branch and then the rebase problem came back.

megies:
> @megies I did try to erase the branch and start all over again. But that automatically close my PR and I asked how to reopen it before. I was only able to reopen my PR once I restore my old branch and then the rebase problem came back.

just do..

```
git checkout master
git pull
git branch -D sac_string_encoding_flag
git checkout -b sac_string_encoding_flag
# do cahnges and commit them
git push -f sac_string_encoding_flag
```

lijunzh:
@megies I did what you said here:

> git checkout master
> git pull
> git branch -D sac_string_encoding_flag
> git checkout -b sac_string_encoding_flag
> # do changes and commit them
> git push -f sac_string_encoding_flag

And it says


> Pull request successfully merged and closed
> You’re all set—the lijunzh:sac_string_encoding_flag branch can be safely deleted.

However, I can't see my changes on `master` branch. Could you please let me know what happened?

megies:
I'm on the move now.. maybe somebody on our gitter channel can give further advice, otherwise I'll do that change tomorrow.

lijunzh:
@megies @jkmacc-LANL I think this PR is now ready for merge. There is only one commit from it which capture all changes I made for the minimal-change version.

obspy/io/sac/sactrace.py
```diff
@@ -1092,7 +1092,7 @@ def read(cls, source, headonly=False, ascii=False, byteorder=None,
                 val = _ut._clean_str(val, strip_whitespace=False)
                 if val.startswith(native_str('-12345')):
                     val = HD.SNULL
-                hs[i] = val
+                hs[i] = val.encode('ASCII', 'replace')
```
QuLogic: This doesn't make sense to me; when reading, why would you want `bytes`?

lijunzh: Thanks for your comment. `hs[i]` was an array of bytes. So, the previous statement will make the implicit conversion which is equivalent to `hs[i] = val.encode('ASCII', 'strict')`. However, when header string is not encoded in 'ASCII', it causes a problem and thus raises an exemption. This PR was trying to let the non-ASCII characters pass the `read` function for now as '?' and leave the implementation of a 'encoding' flag for the later version as we discussed before.

QuLogic: `_ut._clean_str` produces a `str`; `encode` goes from `str` to `bytes`; I think you might have something backwards. Are you testing on Python 3?

QuLogic:
The latest Travis build is [here](https://travis-ci.org/obspy/obspy/builds/231204290) but for some reason it's not linking it right.

QuLogic:
So even GitHub is confused here: `git fetch upstream pull/1773/head && git log -1 FETCH_HEAD`:
commit 85167d07adc14c823b9a951eb60eeea70ddb98c2 (that's not the latest commit here!)

Maybe a reopen will fix it...

QuLogic:
Bah, still wrong, but at least the merge commit (which Travis and AppVeyor test) is correct:
```
$ git fetch upstream pull/1773/head && git log -1 FETCH_HEAD
From github.com:obspy/obspy
 * branch            refs/pull/1773/head -> FETCH_HEAD
commit 85167d07adc14c823b9a951eb60eeea70ddb98c2
Merge: a203b5a 46ca3c5
Author: Tobias Megies <megies@users.noreply.github.com>
Date:   Thu May 11 14:32:27 2017 +0200

    Merge pull request #1667 from obspy/docker_add_debian_stretch
    
    Debian stretch / Ubuntu Zesty 17.04

$ git fetch upstream pull/1773/merge && git log -1 FETCH_HEAD
From github.com:obspy/obspy
 * branch            refs/pull/1773/merge -> FETCH_HEAD
commit bffbcac2d63eaf981308fef95eedacddc97e976a
Merge: 85167d0 865fd6b
Author: Lijun Zhu <lijunzh@users.noreply.github.com>
Date:   Thu May 11 22:56:03 2017 +0000

    Merge 865fd6b0588bae8c093beab6a32f9a61e236df77 into 85167d07adc14c823b9a951eb60eeea70ddb98c2
```

obspy/io/sac/sactrace.py
QuLogic: OK, upon further investigation, this works because SAC I/O is not very good at defining boundaries and does encode/decode a bit too much.

QuLogic: The other thing I didn't know is that NumPy does an implicit `encode` here when assigning to `hs[i]` (even on Python 3, probably for backwards compatibility reasons, but it's kind of unfortunate.)

QuLogic:
I'm still not sure the existing architecture of `bytes`/`str` makes sense, but this looks fine for now.

obspy/io/sac/sactrace.py
lijunzh: Totally agree with you. This has given us a lot problems in the past.

QuLogic:
PS, both docker bots and CircleCI are running on the PR head, whereas Travis and AppVeyor are using the PR merge head. I'm not sure how GitHub got out of sync like this, but that's why they're failing.

lijunzh:
Thank @QuLogic for trying to clear the git problem here. Hopefully, we can merge with those red crosses. The code itself is actually good. It passed those tests before I trying to rebase the git changes.

jkmacc-LANL:
So, if I "dismiss" @krischer's review, will that make merging possible?  Is everything else good to go?

obspy/io/sac/sactrace.py
megies: @QuLogic Yep.. it's a bit strange that we store it internally as bytes again (in numpy) after first decoding it. But we kind of decided to make this PR a minimal fix and postpone the more major cleanup for later..

megies:
Rest of the failing CI is unrelated, so I'm gonna merge this now. Thanks @lijunzh for working on this and uncovering what the actual problem was/is exactly!

krischer:
> So, if I "dismiss" @krischer's review, will that make merging possible? Is everything else good to go?

You can just press "merge" and then use your admin privileges to merge it even if some reviews are not yet accepted.

Sorry for being a bit slow this week.

lijunzh:
@megies @jkmacc-LANL @krischer @QuLogic Thank you all for helping me through this PR. It was a delightful experience contributing to Obspy community. 

@jkmacc-LANL Do you want me to start another PR for actually adding 'encoding' flag to SAC read/write functions or you want to do it yourself?

megies:
> You can just press "merge" and then use your admin privileges to merge it even if some reviews are not yet accepted.

I can't confirm (obviously), but I think this override switch is only available for members of the "admins" team (which is rather small) but not for the "developers" team (which is huge).. which makes sense kind of..

jkmacc-LANL:
@megies Correct, "merge" wasn't highlighted, so I think I needed the help.  Thanks for managing that.  
@lijunzh Sorry this ultimately simple change become rather complicated, but the hard part is over:-)  Welcome to the contributors!  Thanks for your efforts.

lijunzh:
@jkmacc-LANL Since we only access the header field by field, why not store it in a list during the read/write process which can easily hold a byte or string. In this way, we only need to encode/decode once during read/write process.

> I'll do a new PR for adding the write encoding.

Regarding the 'encoding' flag, I don't think we now have it for either read or write. This PR was only a minimal fix for the current unexpected exception. I can work with you to set a uniform read/write 'encoding' flag.

megies:
> @lijunzh Sorry this ultimately simple change become rather complicated, but the hard part is over:-) Welcome to the contributors! Thanks for your efforts.

I can second that, thanks for your work and.. welcome! :-)

PR #1774, opened 2017-05-03
appveyor: optimize building distributions.

xoviat:
- skip builds early in init rather than install 
- suppress all twine output to avoid printing username and/or password
- actually install twine in case it is not installed
- fetch entire history on distribution

This PR is in response to @megies comments, along with some other improvements. I don't anticipate any further adjustments to appveyor.yml.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,7 +65,7 @@ matrix:
     # enable fast fail strategy
     fast_finish: true
 
-install:
+init:
   # If there is a newer build queued for the same PR, cancel this one.
   # The AppVeyor 'rollout builds' option is supposed to serve the same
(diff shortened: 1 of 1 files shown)


megies:
> fetch entire history on distribution

euuhm.. just saw that it's still doing a shallow git clone.. like i said that will report wrong version numbers for us

xoviat:
> euuhm.. just saw that it's still doing a shallow git clone.. like i said that will report wrong version numbers for us

Yes, because there's no need to fetch the entire history for no distribution.

megies:
Like I said, version numbers reported by the test runner (and reported to http://tests.obspy.org/) will be wrong if we don't fetch the full history.

PR #1776, opened 2017-05-04
SDS client: add option to fetch available seed ids for specific net/sta only

megies:
This PR enables fetching available SEED IDs in the SDS archive for specific network/station codes. Getting a list of *all* available SEED IDs might take a very long time on slow file systems because a huge amount folders/files need to be checked. This speeds up requests when only SEED IDs of specific stations are of interest.

Not yet ready.. needs test etc. just posting it here so I don't forget later.

### DIFF
--- a/obspy/clients/filesystem/sds.py
+++ b/obspy/clients/filesystem/sds.py
@@ -486,7 +486,8 @@ class Client(object):
         else:
             return False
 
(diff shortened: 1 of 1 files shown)


Jollyfant:
Might as well ask for location and channel too. If they are `None` just put them to `*`.

megies:
Yeah.. just also used that locally, so putting it here too.. ;-)

PR #1767, opened 2017-05-05
Docs and tutorial state that output of paz_to_freq_resp should be manipulated to get the phase - really?

trichter:
[Docs](https://docs.obspy.org/master/packages/autogen/obspy.signal.invsim.paz_to_freq_resp.html)
[Tutorial](https://docs.obspy.org/tutorial/code_snippets/frequency_response.html)

The doc states:

Note
In order to plot/calculate the phase you need to multiply the complex part by -1. This results from the different definition of the Fourier transform and the phase. The numpy.fft is defined as A(jw) = int_{-inf}^{+inf} a(t) e^{-jwt}; where as the analytic signal is defined A(jw) = | A(jw) | e^{jphi}. That is in order to calculate the phase the complex conjugate of the signal needs to be taken. E.g. phi = angle(f,conj(h),deg=True) As the range of phi is from -pi to pi you could add 2*pi to the negative values in order to get a plot from [0, 2pi]: where(phi<0,phi+2*pi,phi); plot(f,phi)

But `paz_to_freq_resp` never uses `numpy.fft`, rather `scipy.freqs`. Also the phase in the tutorial plot should go the other way round from `+` to `-`, shouldn't it?

### DIFF
--- a/misc/docs/source/tutorial/code_snippets/frequency_response.py
+++ b/misc/docs/source/tutorial/code_snippets/frequency_response.py
@@ -25,6 +25,7 @@ plt.ylabel('Phase [radian]')
 plt.yticks(
     [0, np.pi / 2, np.pi, 3 * np.pi / 2, 2 * np.pi],
     ['$0$', r'$\frac{\pi}{2}$', r'$\pi$', r'$\frac{3\pi}{2}$', r'$2\pi$'])
+plt.ylim(-0.2, 2 * np.pi + 0.2)
 # title, centered above both subplots
 plt.suptitle('Frequency Response of LE-3D/1s Seismometer')
 # make more room in between subplots for the ylabel of right plot
--- a/misc/docs/source/tutorial/code_snippets/frequency_response.py
+++ b/misc/docs/source/tutorial/code_snippets/frequency_response.py
@@ -17,8 +17,7 @@ plt.xlabel('Frequency [Hz]')
 plt.ylabel('Amplitude')
 
 plt.subplot(122)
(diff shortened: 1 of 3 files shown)


trichter:
Nobody commented - so probably I am right. I pushed a fix.

+DOCS

megies:
You could also try to squash it into a `Response` object with `PolesAndZerosResponseStage` and see what the plot for it looks like..

krischer:
> Also the phase in the tutorial plot should go the other way round from + to -, shouldn't it?

I agree. Could you also change the ylim of the plot? Looks a bit funny now. http://docs.obspy.org/pull-requests/1767/tutorial/code_snippets/frequency_response.html

megies:
Here's a plot for a LE-3D lite:
```python
client = Client('BGR')
inv = client.get_stations(station='TMO20', level='response', channel='EHZ')
inv.plot_response(0.001)
```

figure_3

.. so I guess the correction is right..

http://docs.obspy.org/pull-requests/1767/tutorial/code_snippets/frequency_response.html
http://docs.obspy.org/master/tutorial/code_snippets/frequency_response.html

trichter:
Yes, I also checked it against the phase of 1 / ( 1 + j*2*pi*f). (pole at -1, phase=arctan2(-2*pi*f)). Should be OK.

I try to make it look a bit better.

trichter:
Is ready now if docs build is OK.

krischer:
Looks good to me. Travis failures are unrelated.

PR #1779, opened 2017-05-05
Add support for the IRIS Federated Catalog Service

CelsoReyes:
IRIS has a  [Federated Catalog Webservice](https://service.iris.edu/irisws/fedcatalog/1/) that I have been tasked with integrating into ObsPy.

From the webpage:

> The irisws-fedcatalog web service returns selected time series channels from across multiple FDSN data centers. The combined catalog of channels is updated daily from each contributing data center.

The use of this module is similar to the existing fdsn Client.
```python
from obspy.clients.fdsn import FederatedClient
client = FederatedClient()
data = client.get_waveforms( ... ) #  just like normal
data = client.get_waveforms_bulk( ... ) #  just like normal
inv = client.get_stations( ... ) #  just like normal
inv = client.get_stations_bulk( ... ) #  just like normal
```

These would behave much like the normal versions, except that they would retrieve data from any number of data centers (providers), depending on the response of the fedcatalog service

one can also get the routing information without downloading the data
```
routes = client.get_routing( ... )
routes = client.get_routing_bulk( ... )
```

I have attached a diagram that shows how the parts interact:
[FederatedClient.pdf](https://github.com/obspy/obspy/files/979481/FederatedClient.pdf)


CelsoReyes:
apveyor seems to befailing on  test_future_imports_in_every_file, even though the imports exist...

megies:
This file needs to to be squashed out from this PR (its 4MB btw): https://github.com/CelsoReyes/obspy/blob/eb7187237740e6cd27dcd59e76c1c4508f7ff968/obspy/clients/fdsn/routers/.ipynb_checkpoints/Untitled-checkpoint.ipynb

Please move `obspy/clients/fdsn/routers/tests/test_federatedclient.py` over to `obspy/clients/fdsn/tests/test_federatedclient.py` and get rid of `fdsn/routers/tests`.

These descritpitve files either need to be moved to `fdsn/doc` or removed:
 - fdsn/routers/FederatedClient.svg
 - fdsn/routers/fedcatalog_parser.yuml
 - ...

You also should add the nwe submodule and important classes/functions to our docs skeletons, see https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule for more info.

megies:
> apveyor seems to be failing on test_future_imports_in_every_file, even though the imports exist...

Those imports need to be at top of files and grouped together otherwise our regex checks dont pick it up properly. I've added some commits that should fix the future imports, @CelsoReyes please remember to pull in changes to your local branch.

megies:
Looks like this is might be ready-ish already, so I'll add it to 1.1.0 for now..

krischer:
Something just went wrong here ;-) Did you merge `master` into the branch? You'll probably want to rebase `fcf5736eaef9c50fb0de1a077c17515f886f9cbb` on top of master. I can do it you want but it would result in a force push to your fork so I'd better ask beforehand.

CelsoReyes:
Hi Tobias, Lion,
Thanks for the review.  This is where I guess I'm getting tripped up now.  I had attempted to update my local branches to get them synced up before making any changes. But I guess now my version and this are intimately tied together?  This is a part of Git I've not had much experience in.

I haven't made any changes other than trying to get things synced, so feel free to do whatever you need to to untangle me (sorry!).  Then, once we're all in a good place, I will make any still-needed changes to my branch

megies:
Rebased on current master and force-pushed..

CelsoReyes:
Hi, Just checking on the status.  It looks like it's tripping over some core stuff?  I'm not making too much sense out of a couple of these reports.  Moving forward, issues regarding this feature will  be handled by Chad Trabant and the guys at IRIS.  Of course, I'm still here and at your service for questions

krischer:
I'm still working on one other thing - once that is done I'll have a look at this PR. Sorry for taking so long but this is a big PR and I need to find a big chunk of time to do it.

megies:
The files that were removed in this commit (0f39d1a4d3bd51aa9a491de5e5a21d9dca1b52ad) need to be removed from history for good, just removing them with another commit doesn't cut it, this will need some history cleanup/squashing.

nick-iris:
The files removed in commit https://github.com/obspy/obspy/commit/0f39d1a4d3bd51aa9a491de5e5a21d9dca1b52ad have been completely removed from the history.

megies:
@nick-iris can you please fix the `future` imports in all files (see CI logs)? Each python file('s code block) has to start strictly with the following lines (as we do some regex check on this):

```python
                        unicode_literals)
from future.builtins import * # NOQA
```

For example [this file](https://github.com/CelsoReyes/obspy/blob/74154377fa0da517029b4db9ba43a2db69921bd0/obspy/clients/fdsn/routers/fedcatalog_client.py#L36) is missing the second part and causes our CI checks to fail (other files might be failing as well).

nick-iris:
@megies Sure I'll add the `future` imports this morning.

nick-iris:
@megies The `future` import issue has been addressed. The test was failing because the fedcatalog_client.py, fedcatalog_parser.py, and routing_client files used carriage return line feeds (CRLF) instead of just line feeds (LF). I rewrote each of the files to use only line feeds and the test is now passing. You may wish to look into this "issue" further. Thanks.

megies:
@nick-iris just noticed that CI hasn't been running FDSN client tests.. so green CI is rather meaningless so far.. so:  +TESTS:clients.fdsn

nick-iris:
Good catch. I'll work on fixing the remaining issues.

megies:
Seems there's a Python 2/3 issue left to tackle..

nick-iris:
Thanks @megies. I'm working on fixing the Python 2/3 issues now.

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -109,7 +109,7 @@
 ...                             starttime='2015-05-24T12:00:00')
 
   or, using obspy.core.UTCDateTime ...
->>> t1 = UTCDateTime(2015,05,24,12,0,0)
+>>> t1 = UTCDateTime(2015,5,24,12,0,0)
```
megies: @nick-iris please do also adhere to PEP8 in docstrings/doctests, e.g. here add spaces after commas.

obspy/clients/fdsn/routers/fedcatalog_client.py
```diff
@@ -593,7 +593,8 @@ def get_waveforms_bulk(self, bulk, quality=None, minimumlength=None,
              :meth:`~obspy.fdsn.clients.Client.get_stations_bulk`
 
         >>> client = FederatedClient()
-        >>> bulkreq = "IU ANMO * ?HZ 2010-02-27T06:30:00 2010-02-27T06:33:00"
+        >>> bulkreq = str("IU ANMO * ?HZ 2010-02-27T06:30:00 "
```
megies: I hope the `str` is not necessary here? The parantheses alone should be sufficient to allow the line break..

obspy/clients/fdsn/client.py
```diff
@@ -1078,7 +1078,10 @@ def _get_bulk_string(self, bulk, arguments):
                 msg = ("Unrecognized input for 'bulk' argument. Please "
                        "contact developers if you think this is a bug.")
                 raise NotImplementedError(msg)
-        return bulk.encode('ASCII')
+        if PY2:
+            return bulk
+        else:
+            return bulk.encode('ASCII')
```
nick-iris: @megies For Python 3 I had to explicitly encode the bulk data string as ASCII (to be `<class 'bytes'>`). For Python 2 the bulk request "string" is of type `future.types.newstr.newstr` and worked as is.

Do you know of a better way to handle this?

nick-iris:
The Travis-CI tests are now passing. Sorry for the delay in getting this working for you. The Python 2/3 issues were a little tricky to track down.

obspy/clients/fdsn/client.py
megies: I can't say for sure without digging through the code and investigating which test fail you're fixing here (and what that test is doing), the main question is where this return is getting used..

But, if you're running into a situation that you have to do an encoding or not, depending on Py2/3 maybe the problem is further up.. it's best practice to always do explicit de/encoding on I/O.

megies:
Don't know why, but Travis only tested fdsn module on Python 2.7.. http://tests.obspy.org/?pr=1779&node=travis-ci

I'll trigger a new Travis run for this PR.

megies:
+DOCS

megies:
I had a quick glimpse on this PR.. the main complaints i have from this quick glimpse are

 -  that there is no example, I think there should be one example added to `fdsn/__init__.py`, similar to the examples for the regular client.
 - the docstring for `FederatedClient` initialization is rather obscure. as this is exposed to the users this should be improved, I would strongly vote for explicitly repeating those args and their docstrings that are handed over to `RoutingClient.__init__` (`arguments destined for either Fedcatalog or Client` is just really too vague and actually wrong, since it's actually going to `RoutingClient`)
 - usage of `FederatedClient.exclude_provider` is rather obscure. Please transform this either into an `arg` of `FederatedClient.__init__` or add it as a kwarg to all the request methods. It's also totally unclear what values are acceptable for `exclude_provider`, those strings seem to come from nowhere ("IRISDMC","IRIS","IRIS-DMC"??).

```
Exclude a provider from being queried

>>> keep_out = ["IRISDMC","IRIS","IRIS-DMC"]
>>> fclient.exclude_provider = keep_out
>>> INV2 = fclient.get_stations(network="I?", station="A*",
...                           starttime="2013-01-01", level="network")
```

megies:
- docstring for `FederatedClient.get_routing` also seems messy to me. Are these so volatile that we can't explicitly state them? Users should be able to make use of the software without doing internet searches for parameters to use IMHO..

```python
 382     def get_routing(self, routing_file=None, **kwargs):                                                                           
 383         """                                                                                                                       
 384         send query to the fedcatalog service as param=value pairs (GET)                                                           
 385                                                                                                                                   
 386         Retrieves and parses routing details from the fedcatalog service,                                                         
 387         which takes a query, determines which datacenters/providers hold                                                          
 388         the appropriate data, and then returns information about the holdings                                                     
 389                                                                                                                                   
 390         :type routing_file: str                                                                                                   
 391         :param routing_file: filename used to write out raw fedcatalog response                                                   
 392         :type **kwargs: various                                                                                                   
 393         :param **kwargs: arguments to be passed to the fedcatalog service                                                         
 394         as GET parameters.  eg ... http://.../query?param1=val1&param2=val2&...                                                   
 395         :rtype: :class:`~obspy.clients.fdsn.routers.FederatedRoutingManager`                                                      
 396         :returns: parsed response from the FedCatalog service                                                                     
```

nick-iris:
Thanks for the review.

While working on these fixes I believe I found a potential problem. The ObsPy `Clients` currently use a hardcoded list of datacenters that implement the FDSN Web Service specification.

i.e. 
```
URL_MAPPINGS = {
    "BGR": "http://eida.bgr.de",
    "ETH": "http://eida.ethz.ch",
    "EMSC": "http://www.seismicportal.eu",
    "GEONET": "http://service.geonet.org.nz",
    "GFZ": "https://geofon.gfz-potsdam.de",
    "ICGC": "http://ws.icgc.cat",
    "INGV": "http://webservices.rm.ingv.it",
    "IPGP": "http://eida.ipgp.fr",
    "IRIS": "http://service.iris.edu",
    "ISC": "http://isc-mirror.iris.washington.edu",
    "KOERI": "http://eida.koeri.boun.edu.tr",
    "LMU": "http://erde.geophysik.uni-muenchen.de",
    "NCEDC": "http://service.ncedc.org",
    "NIEP": "http://eida-sc3.infp.ro",
    "NOA": "http://eida.gein.noa.gr",
    "ODC": "http://www.orfeus-eu.org",
    "ORFEUS": "http://www.orfeus-eu.org",
    "RESIF": "http://ws.resif.fr",
    "SCEDC": "https://service.scedc.caltech.edu",
    "USGS": "https://earthquake.usgs.gov",
    "USP": "http://sismo.iag.usp.br”}
```

The issue with this design is that every time a datacenter is added to the IRISWS FedCatalog service, the source code will need to be updated. 

For example, the new 'IRISPH5' datacenter endpoint was recently added to the FedCatalog, however it is not in the hardcoded URL_MAPPINGS list. Because of this, the FederatedClient code raises - `ValueError: The FDSN service shortcut 'IRISPH5' is unknown.` since 'IRISPH5' is not in the URL_MAPPINGS dictionary.

To remedy this issue, the service endpoints need to be dynamic. I think that If the service shortcut is not found in the URL_MAPPINGS dictionary, then all FDSN `Clients` should also check if the service shortcut exists in the result fetched from https://service.iris.edu/irisws/fedcatalog/1/datacenters before raising an exception.

I was planning on adding the additional check here: https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/client.py#L205-L211.

Does this seem like a reasonable solution to you?

nick-iris:
@megies I believe that I made all of the changes that you wanted, with the exception of removing the `PROVIDERS` global variable from the `fedcatalog_client` module. I wasn't sure of the best way to do this, but perhaps you have some ideas.

I also modified the `FederatedClient` to explicitly use the endpoints returned by the FedCatalog Service, instead of the hardcoded URL_MAPPINGS shortcuts in the `header` module.

megies:
Some more comments (sorry not enough time to do complete review in one single run.. plus on vacation):

@nick-iris I added some commits, please make sure to pull in those before adding more commits

 - [ ] `FederatedClient` should have a `_repr_pretty_()` method for pretty printing in IPython (compare e.g. normal FDSN client)
 - [ ] HTTPS warning on initialization (and `get_waveforms`) needs to be handled: ` /home/megies/anacond /envs/default/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:852: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings`
 - [ ] verbose stdout logging on requests should be deactivated by default, with optional `verbose` flag (if you think it should be `verbose=True` so that first time users see logging, then at least show a message how to disable verbose logging as the first line when verbose is `True`)
 - [ ] streams should be trimmed to requested start/endtime (returned data has some additional samples at start end)
 - [ ] the url of the service should not be hard coded as a global, it's hard to access for the user on URL changes, please change this in line with the normal client, also major version has to be accessible on init; to sum up e.g. `def __init__(self, url='https://service.iris.edu/irisws/fedcatalog/', major_version=1, use_parallel=False, ...)` and then assemble the url (checking for single trailing slash)
 - [ ] if i see this right, individual access clients to single data centers are created from scratch on each request. this doesnt seem ideal, i think the clients should be cached as an attribute of `RoutingClient` (for `parallel_query_machine` this might need a lock for the client initialization phase, or do it sequentially and then only the requests in parallel)
 - [ ] please check the changed docs pages  for interlink errors (i fixed a few but did not check systematically)
 - [ ] assert statements should be changed to explicit `if not ...: raise ...` (assert statements are skipped when running `python -O`, e.g. try `python -O -c 'assert 0 == 1'`)
 - requesting `'--'` as location code for actually `''` location codes in the python layer should not be encouraged IMHO (by examples), this is merely a helper mapping within FDSNWS (actually using `''` wasn't even working, see below)
 - [ ] add regression test for empty location code (e.g. `get_waveforms(location='')`), wasnt working

megies:
> @megies I believe that I made all of the changes that you wanted, with the exception of removing the PROVIDERS global variable from the fedcatalog_client module. I wasn't sure of the best way to do this, but perhaps you have some ideas.

I'd just set it during `FederatedClient.__init__()` as `self._providers` and update it there accordingly when needed. This also avoids surprising interactions when using multiple `FederatedClient`s.

megies:
- [ ] this also needs checking (during `obspy-runtests clients.fdsn`):
```
[2017-08-17 21:41:05,566] - obspy.clients.fdsn.routing_client - INFO: Retrieved 3 items from GFZ
/home/megies/git/obspy/obspy/clients/fdsn/client.py:1446: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  elif wadl == "timeout":

```

nick-iris:
> streams should be trimmed to requested start/endtime (returned data has some additional samples at start end)

@megies  In my opinion, by default the ObsPy client should not clean up the waveform data that is returned by the web services. I think that it should be the responsibility of the web service to properly cut the waveform data to the correct start and end times. Perhaps we should make this an optional feature.

obspy/clients/fdsn/routers/fedcatalog_client.py
```diff
@@ -56,15 +57,18 @@
     inventory_to_bulkrequests, stream_to_bulkrequests)
 
 
+# ignore InsecureRequestWarning
+# https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```
megies: hmm.. so if we don't do certificate verification, does it make sense to bother requesting from https://?  @QuLogic maybe you can comment on this..? I don't have a lot of experience with this.

markcwill:
Why is SSL verification disabled in 'requests' calls?

megies:
> @megies In my opinion, by default the ObsPy client should not clean up the waveform data that is returned by the web services. I think that it should be the responsibility of the web service to properly cut the waveform data to the correct start and end times. Perhaps we should make this an optional feature.

I tend to disagree. The user is making a request to the obspy client and specifies start and end time, so to the user it doesnt matter where the data is cut to these times, but it should be in any case.

megies:
> Why is SSL verification disabled in 'requests' calls?

From what I'm reading [here](https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl) this would mean additional dependencies? Dunno..

markcwill:
@megies, you are right for `urllib3`, but `requests` is a wrapper and does this for you, `certifi` is a hard dep for `requests`, so all the deps should be met? The only reason I can think of for doing this is that the FDSN certs aren't valid, or are self-signed, the solution for this would be to change the endpoint to not be `https` if it isn't actually supported, or offer their self-signed cert publicly so users can add it to their trusted bundle somehow. This looks like a security issue to me -- ignoring invalid certs deep in the code and swallowing all the user warnings?

I might be missing something. Curious as to the reason @CelsoReyes or @nick-iris or @adam-iris have for doing this.

megies:
> This looks like a security issue to me -- ignoring invalid certs deep in the code and swallowing all the user warnings?

I agree, going to https and ignoring warnings is not really an option. If https doesn't work properly we should either figure out why or fall back to http.

nick-iris:
> I tend to disagree. The user is making a request to the obspy client and specifies start and end time, so to the user it doesnt matter where the data is cut to these times, but it should be in any case.

The normal (non-federated) obspy.clients.fdsn.Client does not trim samples outside of the requested start and end times. I think this is also an issue with the normal Client and not the FederatedClient exclusively.

For example, the following dataselect request for 10 seconds of data returned a result with containing data before and after the requested start and end times.

**GET Version:**
http://eida.ipgp.fr/fdsnws/dataselect/1/query?net=G&sta=PEL&cha=LHZ&starttime=2016-11-01T00:00:00&endtime=2016-11-01T00:00:10
```
   Source                Start sample             End sample        Gap  Hz  Samples
G_PEL_00_LHZ      2016,305,23:43:16.000000 2016,306,00:00:27.000000  ==  1   1032
Total: 1 trace(s) with 1 segment(s)
```

**ObsPy Client Version:**
i.e. running
```python
from obspy import UTCDateTime
from obspy.clients import fdsn
c = fdsn.client.Client('http://service.iris.edu',
                        service_mappings={
                            'dataselect': \
                                'http://eida.ipgp.fr/fdsnws/dataselect/1'
                        },
                        debug=True
                    )

stream = c.get_waveforms('G', 'PEL',
                         '*', 'LHZ',
                         UTCDateTime('2016-11-01T00:00:00'),
                         UTCDateTime('2016-11-01T00:00:10'))

print stream
```
produces this output
```
1 Trace(s) in Stream:
G.PEL.00.LHZ | 2016-10-31T23:43:16.000000Z - 2016-11-01T00:00:27.000000Z | 1.0 Hz, 1032 samples
```

markcwill:
Per  #1254, I think we are trying to move toward `requests` rather than away. @nick-iris your existing code looked good (other than disabling SSL). I was kind of hoping it would be model for fixing `fdsn.client`, not the other way around... Was there a test failure related to `requests`?

adam-iris:
I think moving to `requests` is a good idea, but I feel like this isn't the right place to spearhead that work. (Mainly, it seems like a very large change in scope.) The latest changes have `FederatedClient` now mostly reusing components from `fdsn.client`; as a result, if `fdsn.client` does move to use `requests` I think it should be relatively simple for `FederatedClient` to follow.

megies:
We're trying to finally get to a 1.1.0 release, so I think it's OK for now if it's not using requests exclusively. (CC @krischer)



> For example, the following dataselect request for 10 seconds of data returned a result containing data before and after the requested start and end times.

Interesting. This should indeed then also be fixed in the regular FDSN client. At least for the normal request (not for the direct output request using `filename=...` which directly saves the MSEED data to file without modification).
I guess it's OK for now in here as well and we can handle it in a follow up issue for both clients.

obspy/clients/fdsn/__init__.py
```diff
@@ -134,8 +134,69 @@
                                         endtime=endtime)
         inventory.plot()
 
+
+Basic FDSN FedCatalog Client Usage
+----------------------------------
+
+The
+:mod:`FDSN fedcatalog_client <obspy.clients.fdsn.routers.fedcatalog_client>`
```
megies: This link seems [broken in the docs build](http://docs.obspy.org/pull-requests/1779/packages/obspy.clients.fdsn.html#basic-fdsn-fedcatalog-client-usage), I think the reason is that the new submodules are not added to the sphinx package skeleton: https://github.com/obspy/obspy/blob/master/misc/docs/source/packages/obspy.clients.fdsn.rst

See also the build log for respective warnings/errors: http://docs.obspy.org/pull-requests/1779/log.txt

obspy/clients/fdsn/__init__.py
```diff
@@ -134,8 +134,69 @@
...
+module provides federated
+access to multiple web servers that implement the
+`FDSN Station and Dataselect web service definitions
+<https://www.fdsn.org/webservices/>`_.
+
+The first step is always to initialize a :class:`FederatedClient` object.
```
megies: This link sees broken as well.. this should work, hopefully: ```:class:`~obspy.clients.fdsn.FederatedClient` ``` if not you have to specify full submodule path..

obspy/clients/fdsn/__init__.py
```diff
@@ -134,8 +134,69 @@
...
+
+>>> from obspy.clients.fdsn import FederatedClient
+>>> client = FederatedClient()
+
+(1) :meth:`~obspy.clients.fdsn.routers.fedcatalog_client.
+FederatedClient.get_waveforms()`: The following
```
megies: broken link here too, likely due to aforementioned missing submodule entries in sphinx skeleton document (`misc/docs/source/packages/obspy.clients.fdsn.rst`)

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+implement the FDSN web service definitions (https://www.fdsn.org/webservices/).
+The holdings of many of these services are indexed by the FedCatalog service
+ (https://service.iris.edu/irisws/fedcatalog/1/). Users searching for waveforms
+ or metadata can then query the FedCatalog service to learn which provider
+ holds which data. Furthermore, the results from a FedCatalog query are easily
+ turned into POST requests for each service.
+
+ This FederatedClient first queries the FedCatalog service to determine where
+ the data of interest reside. It then queries the individual web services from
+ each provider using the FDSN Client routines to retrieve the resulting data.
```
megies: Remove leading spaces for all these lines, it's interpreted as a block quote by sphix: http://docs.obspy.org/pull-requests/1779/packages/autogen/obspy.clients.fdsn.routers.html#module-obspy.clients.fdsn.routers

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+    >>> st.plot()  # doctest: +SKIP
+
+    .. plot::
+
+        from obspy import UTCDateTime
+        from obspy.clients.fdsn import FederatedClient
+        client = Client()
+        t = UTCDateTime("2010-02-27T06:45:00.000")
+        st = client.get_waveforms("IU", "ANMO", "00", "LHZ", t, t + 60 * 60)
+        st.plot()
```
megies: Code is broken as it's importing FedClient and then trying to use normal client. Also, I think it's not necessary in this example to do a plot.. I would just have it print the `st` after fetching the data in the doctest example above.

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+
+Submitting a GET request to the federated catalog service. The service
+recognizes parameters that are normally accepted by the station web service.
+
+>>> inv = client.get_stations(station="AN*", channel="BHZ", level="station")
+
+
+Retrieving Waveform Metadata
+---------------------------
+Submitting a GET request to the federated catalog service. The service
```
megies: A lot of times I'm seeing these notes about doing `GET` or `POST` requests (FDSN API front page, fdsn/routers API front page etc)... this should be removed from the docs I think, that's a low level implementation aspect that's not interesting for the user.

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+        from obspy import UTCDateTime
+        from obspy.clients.fdsn import FederatedClient
+        client = Client()
+        t = UTCDateTime("2010-02-27T06:45:00.000")
+        st = client.get_waveforms("IU", "ANMO", "00", "LHZ", t, t + 60 * 60)
+        st.plot()
+
+caveats
+----
+* duplicated metadata remains duplicated.
```
megies: I think this could use a bit more information

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+suggestions for future
+----------------------
+* bulk requests for a certain level of metadata, eg. network, station
+  will return an egregious amount of data that is very likely duplicated
+  Perhaps the fedcatalog web service could recognize it and return pared
+  down data or the fdsn.client.Client  class could recognize and reduce
+  the duplications. example:
+    >>> INV2 = fclient.get_stations(network="I?", station="AN*",
+    ...                           level="network", includeoverlaps="true")
+    ...                           #doctest: +SKIP
```
megies: This (deduplication) sounds like a (valid) feature request for `Inventory` class. I'm not sure though if this is the right place to add this notice..

obspy/clients/fdsn/routers/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+  or, using obspy.core.UTCDateTime ...
+>>> t1 = UTCDateTime(2015, 5, 24, 12, 0, 0)
+>>> itemB = FDSNBulkRequestItem(network='IU', station='ANMO', starttime=t1)
+
+Basic comparisons can be made between these items.
+A < B : alphabetically compares net, stations, locations, then chans,
+        and then starttime by date.
+A contains B : This takes wildcards and time ranges into account to denote
+               whether a request for A would include B's data too.
+A == B : all fields are the same (wildcards only match wildcards, for example)
```
megies: This whole section renders pretty awkward in sphinx, can you please try to fix it? http://docs.obspy.org/pull-requests/1779/packages/autogen/obspy.clients.fdsn.routers.html#module-obspy.clients.fdsn.routers

screenshot from 2017-10-04 17-43-14

nick-iris:
@megies Is there a script in place for creating the Sphinx docs on my local computer? Thanks.

jkmacc-LANL:
@nick-iris There's a [Makefile](https://github.com/obspy/obspy/blob/master/misc/docs/Makefile) in `misc/docs`.  You can type `make html`, as long as you have Sphinx and any other needed extensions installed.  There's a dependencies file next to the Makefile that lists doc dependencies.

jkmacc-LANL:
[Here's a relevant link in the Wiki.](https://github.com/obspy/obspy/wiki/Sphinx-documentation)

megies:
Also see https://gist.github.com/megies/f3f639f575e0ca97fdfd9b41348da802 for the exact conda env currently used on our docs buildbot. I would recommend to start out with the same sphinx at least. Also, I fear building docs takes a long time, unfortunately.. :-/

krischer:
Just a quick heads up so we don't do duplicate work: I'm currently working on finalizing this and I'll push my things early next week.

krischer:
Has been merged as part of #1919. I don't know why it did not automatically close it.

Issue #1769, opened 2017-04-28
libmseed 2.19.4 released, small change


chad-iris:
A new version of libmseed (2.19.4) released:
https://github.com/iris-edu/libmseed/releases/tag/v2.19.4

The only change is the addition of a global variable LM_SIZEOF_OFF_T that is set to the size of the off_t data type determined at compile time.

As off_t is not defined in `ctypes`, this could be used to determine which integer should be used instead.  Currently ObsPy [maps off_t usage in libmseed to `ctypes.c_long`](https://github.com/obspy/obspy/blob/master/obspy/io/mseed/headers.py#L466), which is a good guess but will be wrong for 32-bit builds (long is 32-bit) where the library is compiled with large file support (off_t is 64-bit).

Even if ObsPy is not modified right away to take advantage of this is would be useful to include the new libmseed in the next release so that information could be used by others from within ObsPy.

krischer:
Thanks for the heads-up! Done in #1780.

Out of curiosity: What happens on Windows? `off_t` is AFAIK not part of the C standard but defined by POSIX.

chad-iris:
> Out of curiosity: What happens on Windows? off_t is AFAIK not part of the C standard but defined by POSIX.

The `off_t` type is defined on Windows, for a long time now:
https://msdn.microsoft.com/en-us/library/323b6b3k.aspx

Simple C code, limited to C standards and common POSIX usage, is really quite portable between most modern systems nowadays.

QuLogic:
> Thanks for the heads-up! Done in #1780.

Like I said before, add "Fixes #1769" in the description and GitHub will close this one automatically. No need to close it early before it's actually fixed.

PR #1780, opened 2017-05-05
Update to Libmseed 2.19.4

krischer:
Continuation of  #1769 as I currently don't really trust the issue to PR conversion of Github.

The corresponding `ctypes` datatype for `off_t` is now derived by testing the size of a couple of candidates and chosing one that is equal to the size in `libmseed` indicated by the new global `LM_SIZEOF_OFF_T` variable.


chad-iris:
Nice catch 400e2a3e2a80049fad500bd33e4afbe4c6d5c032. I've added this upstream so it'll be in the next release.

krischer:
> Nice catch 400e2a3. I've added this upstream so it'll be in the next release.

Great. I would have sent a PR in any case but I wanted to wait what our CI says and you were faster.

PR #1781, opened 2017-05-08
is_segy() now explicitly checks the version number

krischer:
Fixes a bug noticed on the mailing list where a GCF files was format-detected as a SEG-Y file: http://lists.swapbytes.de/archives/obspy-users/2017-May/002419.html

This PR fixes it by asserting the SEG-Y version number. I don't think an extra test is necessary for this as it would be very constructed and it is already tested by the other tests cases.

### DIFF
--- a/obspy/io/segy/core.py
+++ b/obspy/io/segy/core.py
@@ -110,7 +110,7 @@ def _is_segy(filename):
     # Test the version number. The only really support version number of
     # ObsPy is 1.0 which is encoded has 0100_16. Many file have version
     # number zero which is used to indicate "traditional SEG-Y" conforming
-    # to the1975 standard.
+    # to the 1975 standard.
     # Also allow 0010_16 and 0001_16 as the definition is honestly awkward
     # and I image many writers get it wrong.
     if _format_number not in (0x0000, 0x0100, 0x0010, 0x0001):
--- a/obspy/io/segy/core.py
+++ b/obspy/io/segy/core.py
@@ -105,15 +105,23 @@ def _is_segy(filename):
     _number_of_data_traces = unpack(fmt, _number_of_data_traces)[0]
     _number_of_auxiliary_traces = unpack(fmt,
                                          _number_of_auxiliary_traces)[0]
+
     _format_number = unpack(fmt, _format_number)[0]
+    # Test the version number. The only really support version number of
+    # ObsPy is 1.0 which is encoded has 0100_16. Many file have version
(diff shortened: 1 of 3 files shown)


obspy/io/segy/core.py
```diff
@@ -105,15 +105,23 @@ def _is_segy(filename):
     _number_of_data_traces = unpack(fmt, _number_of_data_traces)[0]
     _number_of_auxiliary_traces = unpack(fmt,
                                          _number_of_auxiliary_traces)[0]
+
     _format_number = unpack(fmt, _format_number)[0]
+    # Test the version number. The only really support version number of
+    # ObsPy is 1.0 which is encoded has 0100_16. Many file have version
+    # number zero which is used to indicate "traditional SEG-Y" conforming
+    # to the1975 standard.
```
Jollyfant: space

Jollyfant: support => supported
has => as

krischer:
Thanks for the review!

> So version 1.0 is encoded as 0x0100 and version 2.0 is encoded as 0x0010? Pretty confusing stuff.. just found a type in the docstring.

Yea it is really strange - the spec is also not entirely clear on how it is supposed to be interpreted... I think it is literally just two hexadecimal numbers and they are interpreted as written in non-hex?

e.g.
* 1.0 = `0x0100`,
* 2.0 = `0x0200`
* 1.1 = `0x0101`
* 10.15 = `0x1015`

Thankfully we don't have to worry too much as there is only version 1.0 ;-)

paitor:
Just a comment on the revision number encoding. According to the official documentation (http://www.seg.org/Portals/0/SEG/News%20and%20Resources/Technical%20Standards/seg_y_rev1.pdf, page 8, bytes 3501-3502):

"This is a 16-bit unsigned value with a Q-point between the first and second bytes." 

I would interpret this as the number format UQ8.8 (see https://en.wikipedia.org/wiki/Q_(number_format)) but note that this may be wrong given that the resolution is 1/256, hence in terms of revision numbers this is a bit impractical as only x.0 and x.5 can be exactly represented (0x0101 = 1.00390625 and 1.5 = 0x0180)

krischer:
Thanks for the comment! I did not know about Q-numbers before but maybe that is what the spec is trying to say. Thankfully we don't have to worry about it as long as there is no SEG-Y 1.1 ;-)

paitor:
Well, if I'm correct in my assumptions that'll be version 1.00390625,  lol

PR #1782, opened 2017-05-08
Mseed attributes now per trace

### LINKED ISSUE #1450: Trace.stats.mseed number_of_records showing info of full file.
Hi all, is there a reason that for all traces loaded from an mSEED file that every Trace.stats.mseed dictionary shows info for the entire file, and not the single trace itself?

```
for tr in st:
    print tr.stats.mseed
```

Will always give the following result

```
{
    'record_length': 512,
    'encoding': 'STEIM2', 
    'filesize': 3684864, 
    'dataquality': 'R', 
    'number_of_records': 7197, 
    'byteorder': '>'
}
```

IMHO it make more sense to change the number of records to the number that is actually contained within the trace and not the entire file. Thoughts?

krischer:
Fixed #1450.

'record_length', 'encoding', 'dataquality', 'number_of_records', and 'byteorder' are now per-trace and no longer global. Includes a test.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -111,6 +111,8 @@ master: (doi: 10.5281/zenodo.165135)
    * Should no-longer segfault with arbitrarily truncated files (see #1728).
    * Will now raise an exception when attempting to directly read mini-SEED
      files larger than 2048 MiB (#1746).
+   * `.stats.mseed` attributes are no longer per-file but per-trace where
+     applicable (see #1782).
  - obspy.io.nlloc:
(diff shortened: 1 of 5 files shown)


Jollyfant:
What about the record data quality indicator? I'm not sure if segments are split on that field at all.

krischer:
> What about the record data quality indicator? I'm not sure if segments are split on that field at all.

Segments are already split on that field (and it also was set correctly for each trace even before this PR). Internally it first sorts everything based on SNCL + dataquality indicator and later merges adjacent records in a second pass.

PR #1760, opened 2017-04-24
Change nlloc signature format

ChrisBail:
### What does this PR do?

fix the problem to read the "signature" in nlloc phase-hyp file line. It allows to recognize between old version (<6) and new nlloc version 

### PR Checklist
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [ ] Any new or changed features have are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #1783, opened 2017-05-09
nlloc: fix reading post-version-6 hypocenter files

megies:
Fixes reading NonLinLoc hypocenter files output by NonLinLoc versions after version 6 (6.0.3 beta is affected).

For whatever reason the structure of the SIGNATURE line was changed and `'run:'` is now prepended to the last part of the SIGNATURE line which holds the date and time of the location run.

Closes #1760.


megies:
Ready if CI gives the nod.

PR #1785, opened 2017-05-09
fixes UTCDateTime not compatible with date separator `/`

### LINKED ISSUE #1784: UTCDateTime not compatible with date separator `/`
`UTCDateTime` takes nearly everything, but no `'/'` as date separator.

```
from obspy import UTCDateTime as UTC
UTC('2005/01/01 20:00:10')

  File "<ipython-input-35-0ccfc701b0c1>", line 1, in <module>
    UTC('2005/01/01 20:00:10')
  File "/home/eule/dev/obspy/obspy/core/utcdatetime.py", line 323, in __init__
    dt = datetime.datetime(*args, **kwargs)
TypeError: an integer is required (got type str)
```

barsch:
fixes #1784

### DIFF
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -20,7 +20,7 @@ class UTCDateTimeTestCase(unittest.TestCase):
         """
         Tests initialization from a given time string not ISO8601 compatible.
         """
-        # some strange patterns
+        # some supported non ISO8601 patterns
         dt = UTCDateTime("1970-01-01 12:23:34")
         self.assertEqual(dt, UTCDateTime(1970, 1, 1, 12, 23, 34))
         dt = UTCDateTime("1970,01,01,12:23:34")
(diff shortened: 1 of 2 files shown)


obspy/core/tests/test_utcdatetime.py
```diff
@@ -31,7 +31,9 @@ def test_from_string(self):
         self.assertEqual(dt, UTCDateTime(2009, 7, 1, 12, 12, 12))
         dt = UTCDateTime("19700101")
         self.assertEqual(dt, UTCDateTime(1970, 1, 1, 0, 0))
-        # non ISO8601 strings should raise an exception
+        dt = UTCDateTime("1970/01/01 12:23:34")
```
megies: Using `1970/01/01` for this test feels a bit unsafe as it's an edge case date..

Otherwise PR looks good.

barsch: we only support YMD never YDM or other combinations (except ISO8101 stuff) - otherwise things get complicated, but maybe we should state that somewhere

megies: I think 1970/01/01 should in general not be used for these tests (at least not solely) because in principle it could mask severe problems parsing the whole date (as it is the 0-date kind of)

PR #1786, opened 2017-05-12
Unify list of obspy dependencies (hard and soft)

megies:
This PR proposes to finally manage the list of all ObsPy dependencies (hard and soft) in one unified place (together with all other important 'headers').

Since this file is used at install time (by setup.py) it can't have other obspy/future imports in it, so we an't use the existing `base.py` for this. But I think it might be cleaner this way anyway.

 - [ ] adapt https://github.com/obspy/obspy/wiki/How-to%3A-add-a-new-submodule


obspy/core/util/headers.py
```diff
@@ -0,0 +1,484 @@
...
+# XXX TODO when bumping to numpy 1.9.0: replace bytes() in io.reftek with
+# np.tobytes()
+HARD_DEPENDENCIES = {
+    "future": ">=0.12.4",
+    "numpy": ">=1.6.1",
+    "scipy": ">=0.9.0",
+    "matplotlib": ">=1.1.0",
+    "lxml": "",
+    "setuptools": "",
+    "sqlalchemy": "",
```
krischer: If we do this separation of hard and soft dependencies we should also use the opportunity and make `sqlalchemy` optional - its only used for `obspy.db` but I think this is a rarely used module in practice for most users.

obspy/core/util/headers.py
```diff
@@ -0,0 +1,484 @@
...
+        "cartopy": ""},
+    'mseed': {
+        "jsonschema": ""},
+    'shapefile': {
+        "gdal": ""},
+    'tests': {
+        "flake8": ">=3.0",
+        "mock": "",  # Py2 only
+        "pep8-naming": "",
+        "pyflakes": "",
```
krischer: already part of `flake8`.

obspy/core/util/headers.py
```diff
@@ -0,0 +1,484 @@
...
+        "pyimgur": "",
+        "pyproj": ""},
```
krischer: Its only needed for the tests of a few specific modules. So maybe rather add it to those?

obspy/core/util/headers.py
```diff
@@ -0,0 +1,484 @@
...
+    }
+if sys.version_info.major == 3:
+    OPTIONAL_DEPENDENCIES['tests'].pop('mock')
```
krischer: Adding it if on Python 2 is IMHO clearer especially if Python 2 will sunset in the foreseeable future.

obspy/core/util/headers.py
```diff
@@ -0,0 +1,484 @@
...
+    'arclink': {
+        "cryptography": "",
+        "m2crypto": "",
+        "pycrypto": ""},
+    # for docs, also see conda environment file under
+    # misc/docs/py3-docs-master.yml
+    # other versions might work as well but these are currently used in our
+    # docs buildbot and therefore confirmed to work
+    'docs': {
+        "sphinx": "==1.2.2",
```
QuLogic: Not so sure we should hard-code these. Makes it annoying to work with other versions.

megies:
> I'm not so sure I prefer moving the setup related things out of setup.py - it is currently a bit long but most things are in one place at least.

I know, I'm torn too, but I don't see a different way to avoid duplication here, since `setup.py` contents are not available in an installation, only in the source code..

> On the other hand the library names for example are already determined by obspy.core.utils.

Yeah, we already have some stuff from `core/util` that gets used during installation, so I think it's pretty clear where to look for things that used to be in `setup.py` directly.



> So I guess I'm not really sure what I'm trying to say here but I'd like to at least discuss this a bit.

Sure, I definitely wanted to hear opinions/concerns about this one. But I feel it's a good thing to do, actually I think it clusters more of these "header" type stuff in one common place now.

obspy/core/util/headers.py
megies: I'm not sure that people actually use those optional dependency specifiers with pip. I'm mostly thinking of this as a place to summarize all hard/soft dep versions and to keep track of things. With building docs it can be really tough to get the correct versions installed, so I'd always happily try to use versions that are confirmed to work.

On the other hand, there's the conda environment file with exact versions that our docs buildbot uses, so that's enough I guess and people could also look it up there.

megies:
Looking at `pkg_resources` documentation, we might also be able to replace version comparison logic (comparing lists of split version number) by just using `pkg_resources.SetuptoolsVersion` (I guess we won't move away from `setuptools` anytime soon)

E.g. replace
```python
from obspy.core.util import to_int_or_zero
GDAL_VERSION_SUFFICIENT = [1, 7, 3] < list(map(
    to_int_or_zero, __gdal_version__.split(".")))
```

by:
```python
from pkg_resources import SetuptoolsVersion, get_distribution
GDAL_VERSION_SUFFICIENT = (
    SetuptoolsVersion('1.7.3') <
    get_distribution('gdal').parsed_version)
```

That way, all in-between version numbers (post/pre-release) etc would be taken into account properly by just handing over version comparison logic to `setuptools`.

QuLogic:
There's also `distutils.version.LooseVersion` (from the stdlib); I never knew why we didn't use it.

megies:

QuLogic:
On `master`, dependencies are solely defined in `setup.py`. I don't think we should move metadata into the runtime.

PR #1787, opened 2017-05-13
Skip test if jsonschema version is too old

krischer:
Also raise a warning message if that is encountered during run time. Fixes the issue mentioned in https://github.com/obspy/obspy/pull/1141#issuecomment-301049970


krischer:
Looks good to me! Thanks. Can IMHO be merged if CI passes.

PR #1797, opened 2017-05-24
docker: fixes for running ARM images on other machines

megies:
Seems like there's some additional commands that need to be executed when building/testing ARM docker images (I didn't need them on my machine when first building those images, somehow..).


krischer:
Cannot really check as it appears to work after I've set-up qemu once. While at it: Can you make it so that the ARM images are not tested by default but require an extra flag? They just take too long to run them all the time.

megies:
> Cannot really check as it appears to work after I've set-up qemu once.

Yes, from what I read, it looks like this has to be executed only once. Or in my case, not at all..

megies:
Euhh.. we should really switch to docker python API at some point.. added a switch for ARM now and in the process tried to unify both the testing and deb packaging scripts (hope it'll help when we switch).

I think it's cleanest to explicitly define the lists of images to run on and not use directory listings, because local directory can always contain untracked files etc.

megies:
I'm closing this one as I was struggling with deb packaging on this branch. So just leave the current messy situation as is.. never touch a running system

PR #1804, opened 2017-06-01
locations2degree numpy compatible (closes #1803)

### LINKED ISSUE #1803: Make locations2degree numpy compatible?
Hi all,

This is not an issue but a sort of enhancement request/ question:
I would like to ask if you plan to make `locations2degree` (https://github.com/obspy/obspy/blob/master/obspy/geodetics/base.py#L309) numpy compatible. From the code it seems straighforward as all python `math` functions have a corresponding `numpy` equivalents (I implemented by myself such a method with a bunch of tests which confirm it's just a matter of string substitution).

Thank you in advance

rizac:
### What does this PR do?

Makes locations2degrees function numpy compatible. That is, it accepts scalars or numpy numeric arrays

Need to have faster calculations of distances over arrays

### PR Checklist
- [X] All tests still pass.
- [] Significant changes have been added to `CHANGELOG.txt` .

### DIFF
--- a/obspy/geodetics/base.py
+++ b/obspy/geodetics/base.py
@@ -333,6 +333,10 @@ def locations2degrees(lat1, long1, lat2, long2):
     >>> locations2degrees(5, 5, 10, 10)
     7.0397014191753815
     """
+    # broadcast explicitly here so it raises once instead of somewhere in the
+    # middle if things can't be broadcast
+    lat1, lat2, long1, long2 = np.broadcast_arrays(lat1, lat2, long1, long2)
+
     # Convert to radians.
     lat1 = np.radians(np.asarray(lat1))
     lat2 = np.radians(np.asarray(lat2))
--- a/obspy/geodetics/base.py
+++ b/obspy/geodetics/base.py
@@ -315,15 +315,15 @@ def locations2degrees(lat1, long1, lat2, long2):
     Earth. For more accurate values use the geodesic distance calculations of
     geopy (https://github.com/geopy/geopy).
 
-    :type lat1: ``float or :class:`numpy.ndarray```
+    :type lat1: float or :class:`numpy.ndarray`
     :param lat1: Latitude(s) of point 1 in degrees
-    :type long1: ``float or :class:`numpy.ndarray```
+    :type long1: float or :class:`numpy.ndarray`
     :param long1: Longitude(s) of point 1 in degrees
-    :type lat2: ``float or :class:`numpy.ndarray```
+    :type lat2: float or :class:`numpy.ndarray`
     :param lat2: Latitude(s) of point 2 in degrees
-    :type long2: ``float or :class:`numpy.ndarray```
+    :type long2: float or :class:`numpy.ndarray`
     :param long2: Longitude(s) of point 2 in degrees
-    :rtype: ``float or :class:`numpy.ndarray```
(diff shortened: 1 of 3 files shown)


obspy/geodetics/base.py
```diff
@@ -315,16 +315,17 @@ def locations2degrees(lat1, long1, lat2, long2):
     Earth. For more accurate values use the geodesic distance calculations of
     geopy (https://github.com/geopy/geopy).
 
-    :type lat1: float
-    :param lat1: Latitude of point 1 in degrees
-    :type long1: float
+    :type lat1: float or numpy numeric array
+    :param lat1: Latitude(s) of point 1 in degrees
+    :type long1: float or numpy numeric array
     :param long1: Longitude of point 1 in degrees
```
krischer: You added the `(s)` for the `lat1` parameter. Please also do it for all others.

megies:
My only criticism is that you seem to allow variable lengths in the input data arrays or even mixing with scalars, right?
You're even allowing an array for `lon1` and a scalar for `lat1`?

I think it's really not very intuitive that way.. I would rather restrict it to equal length of all input lat/lons. After all that's not a speed impediment and can easily set up by the user.

rizac:
Well the function (to me) should allow everything that numpy does (without breaking the previous/pre-defined "scalar" behaviour).
I agree that the "mixing type" is not very intuitive, but I assume that a user passing numpy values knows numpy: implementing a "length check" is not free from criticisms too, in my opinion: it forces a developer to manually tune arrays of same lengths before calling the function, writing more code and probably letting him/her think "what the benefits of making the function numpy compatible then?"

That said, if you have other arguments based on your experience, we can discuss it. After all, I can implement a check in the function

QuLogic:
How about we just make it more explicit? Call `np.broadcast_arrays` on the input and use the result for the calculation. This gets you a) consistent shapes for the duration of the rest of the function and b) an error message in one location instead of somewhere in the middle if things can't be broadcast.

obspy/geodetics/base.py
```diff
@@ -315,37 +315,42 @@ def locations2degrees(lat1, long1, lat2, long2):
...
-    :param lat1: Latitude of point 1 in degrees
-    :type long1: float
-    :param long1: Longitude of point 1 in degrees
-    :type lat2: float
-    :param lat2: Latitude of point 2 in degrees
-    :type long2: float
-    :param long2: Longitude of point 2 in degrees
-    :rtype: float
-    :return: Distance in degrees as a floating point number.
+    :type lat1: ``float or :class:`numpy.ndarray```
```
krischer: Minor nitpick: Can you please remove the double-backticks around the whole type? This page describes the syntax of the sphinx info field lists: http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

rizac: I wrote it like that following your previous suggestion which had double backticks, I am confused but I assume they where intended for the formatting here on the page but they were rendered unformatted.
Whatever, hope to have understood what you mean, let me know

krischer: Ah sorry - my bad - I intended to use single backticks so github would render it nicely but I screwed up and used two backticks...sorry about that.

krischer:
Looks good to me. I'll merge if CI passes.

krischer:
Thanks a bunch!

PR #1805, opened 2017-06-01
Fix tau model load from depth cache (closes #1802) 

### LINKED ISSUE #1802: Bug in `obspy.taup.tau_model.TauModel.load_from_depth_cache` ?
Sorry if the problem has been reported (I didn't find any on the issues list), but it seems inconsistent
the behaviour of this line:
https://github.com/obspy/obspy/blob/master/obspy/taup/tau_model.py#L238
where 
```
self._load_from_depth_cache()
```
is called without arguments (by providing the argument in the constructor `cache=False`)

Regardless of whether I am using the wrong `cache` value (which from the code it does not seem the case, but any help is appreciated in case), this seems to be a bug as the method expects an argument which is not given

Thank you very much in advance

rizac:
### What does this PR do?
Fixes the bug in TauModel.load_from_depth_cache
~~NOTE: this branch has already been merged with PR #1804~~ (shouldn't be the case anymore)

Bug

### PR Checklist
- [] This PR is not directly related to an existing issue (which has no PR yet).
- [X] All tests still pass.

### DIFF
--- a/obspy/taup/tests/test_tau.py
+++ b/obspy/taup/tests/test_tau.py
@@ -180,7 +180,7 @@ class TauPyModelTestCase(unittest.TestCase):
 
     def test_p_iasp91(self):
         """
-        Test P phase arrival against TauP output in in model IASP91
+        Test P phase arrival against TauP output in model IASP91
         with different cache values to test `TauModel.load_from_depth_cache`
         """
         for cache in self.caches:
@@ -193,7 +193,7 @@ class TauPyModelTestCase(unittest.TestCase):
 
     def test_p_ak135(self):
         """
-        Test P phase arrival against TauP output in in model AK135
+        Test P phase arrival against TauP output in model AK135
         with different cache values to test `TauModel.load_from_depth_cache`
         """
         for cache in self.caches:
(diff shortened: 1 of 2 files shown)


krischer:
Can you please remove the commits from #1804? Both PRs are unrelated and pull requests in general should be as atomic as reasonable. You could either do an interactive rebase or just branch again and cherry pick the commits of interest (there are probably other ways to achieve the same goal). Let us know if you need help with that.

rizac:
Of course (I don't know why I had the idea to merge by myself). Anyway: I checked out to a next-to-last commit, as the last one as the 'merge' command. Should work

megies:
Looks like this is ready? (I'm not too much into the taup details..)

obspy/taup/tests/test_tau.py
```diff
@@ -172,82 +180,96 @@ def test_p_iasp91_geo_fallback_manual(self):
 
     def test_p_iasp91(self):
         """
-        Test P phase arrival against TauP output in in model IASP91.
+        Test P phase arrival against TauP output in in model IASP91
```
QuLogic: Not sure that all of these docstrings needed to change, but if you're going to do so, can you remove the duplicate 'in in'?

megies: Done.

megies:
Good to go, I guess. CI was green and reviewed as well.

megies:
Changes look like they make sense to me as well, also it seems like existing test cases were not modified but instead expanded (using different values for `cache` option of `TauPyModel()`). So, merging..

PR #1808, opened 2017-06-06
Fix sc3ml event unit

Brtle:
This small PR fix the origin uncertainty units in the seiscomp event module.

In the QuakeML format, the origin uncertainty is in meter but seiscomp use kilometers.


Brtle:
No, usually an XSLT isn't reversible. However, an XSLT is given to convert SC3ML to QuakeML (http://geofon.gfz-potsdam.de/schema/0.9/). I didn't check it but I think that there is some mistakes, like unit conversion for origin uncertainty. With some changes, it should be possible to add it to ObsPy.

krischer:
Can we use that license wise? It validity should be pretty easy to check by just inverting the current test suite for writing seiscomp event files.

PR #1817, opened 2017-06-19
SC3ML: Set depth to 0 if missing, fix issue with finding publicIDs (fixes #1816)

### LINKED ISSUE #1816: Problem reading ObsPy SC3ML inventory (FW: mailing list)
Hi

I’ve made a station inventory using Gempa’s SMP tool. I’ve ensured the I export the xml as SC3XML/0.7, in line with Obspy.

However, when I try to read the inventory in Obspy, I get the following error:

    File "./test.py", line 4, in <module>
    inv = read_inventory('sph1r17_VOiLA-onshore-8.xml')
    ...
    File "/Users/sph1r17/anaconda/lib/python3.5/site-packages/obspy/io/seiscomp/sc3ml.py", line 472, in _read_channel
     response_element, cha_element,
    UnboundLocalError: local variable 'response_element' referenced before assignment

The xml file is here if anyone would like to see it:  https://www.dropbox.com/s/nso0fdtf5uj3yeo/sph1r17_VOiLA-onshore-8.xml?dl=0.

The smp repository can be found here:  https://smp.gempa.de/sph1r17/VOiLA-onshore

The sensor and digitiser response seems to be correctly attached in the xml file.

Can anyone assist me here with trying to get to the root of the problem?

Thanks in advance for your help
Steve

Jollyfant:
* Response is now found even if the publicID is not prepended with the type.
* Depth set to 0 if missing from channel element

- [ ] Send fix around in mailing list

Still needs tests.


krischer:
Thanks for the quick action! Someone should remember to write to the mailing list once this is merged.

trichter:
:+1: Great, this fixes a similar issue I had with reading SC3ML. I thought the XML was not valid and inserted the missing depth manually into the XML file.

obspy/io/seiscomp/sc3ml.py
```diff
@@ -383,6 +383,9 @@ def _read_channel(inventory_root, cha_element, _ns):
                                 unit=True)
     depth = _read_floattype(cha_element, _ns("depth"), Distance,
                             unit=True)
+    # Set depth to 0 if it is missing (#1816)
+    if depth is None:
+        depth = 0
```
megies: Same goes for longitude/latitude/elevation, they're all not mandatory in sc3ml according to the schema..

Also, please show a warning message, e.g.
```python
    if depth is None:
        msg = "Channel is missing depth information, using '0.0'"
        warnings.warn(msg)
        depth = 0
```

Jollyfant:
@krischer I tried mailing to the list but it wouldn't let me do it. Said I needed to sub first. I pointed Steve who reported this to the thread in a private mail instead. I'll add a to-do for it anyway.

obspy/io/seiscomp/sc3ml.py
```diff
@@ -398,11 +401,11 @@ def _read_channel(inventory_root, cha_element, _ns):
...
-            resp_type = response_id.split("#")[0]
-            if resp_type == 'ResponsePAZ':
-                search = "responsePAZ[@publicID='" + response_id + "']"
-                response_element = inventory_root.find(_ns(search))
-            elif resp_type == 'ResponsePolynomial':
+            resp_type = 'ResponsePAZ'
+            search = "responsePAZ[@publicID='" + response_id + "']"
+            response_element = inventory_root.find(_ns(search))
+            if response_element is None:
+                resp_type == 'ResponsePolynomial'
```
megies: There's also `ResponseFIR` tags.. they should also be searched, no?

I'd also propose to sanitize the lookup a bit so that problems with one publicID being multiplied in the file are caught:

```python
            response_elements = []
            for resp_type in ['responsePAZ', 'responsePolynomial', 'responseFIR']:
                search = "{}[@publicID='{}']".format(resp_type, response_id)
                response_elements += inventory_root.findall(_ns(search))
            if len(response_elements) == 0:
                msg = ("Could not find response tag with public ID "
                       "'{}'.".format(response_id))
                raise ObsPyException(msg)
            elif len(response_elements) > 1:
                msg = ("Found multiple matching response tags with the same "
                       "public ID '{}'.".format(response_id))
                raise ObsPyException(msg)
            response_element = response_elements[0]
```

krischer:
@Jollyfant You are registered with `LASTNAME[at]knmi.nl` in the mailing list - make sure you use this email as the sender - then it should work.

megies:
> I tried mailing to the list but it wouldn't let me do it. Said I needed to sub first.

You need to make sure that the sender address you're using is the one subscribed to the list..

Jollyfant:
I will fix and push this next week including the comments by @megies

shicks-seismo:
Hi guys, thanks so much for working on this. I just found this thread.

@Jollyfant - I didn't get your private mail, but cheers for requesting the changes.

obspy/io/seiscomp/sc3ml.py
Jollyfant: The sensor response attribute can only be either responzePAZ or responsePolynomial. ResponseFIR does not need to be searched because it can only be defined in a `digitalFilterChain` element inside a `datalogger` element and by then you already know the type.

megies: Ah OK, thanks for looking into it.

megies:
Just a few flake8 fails that need to be fixed (lines too long) otherwise good to go, I think.

Jollyfant:
These lines introduced a regression and would raise if the response was missing (i.e. channel level is requested) but has been fixed in the last commits.

    if len(response_elements) == 0:
        msg = ("Could not find response tag with public ID "
                    "'{}'.".format(response_id))

I will do some more stability testing later and merge this by the end of this week.

krischer:
Thanks a bunch!

Are you going to write the mailing list?

Jollyfant:
Yeah but it's moderated so someone needs to approve it.

krischer:
Just did. Everything is moderated now as some spammers faked emails coming from us to post to the mailing list. A bit inconvenient that every message has to be approved but better than having the spam. I actually don't know how other mailing lists deal with this.

megies:
> A bit inconvenient that every message has to be approved but better than having the spam. I actually don't know how other mailing lists deal with this.

That would be a good use case for email signing with PGP.. in principle.. ;-)

PR #1821, opened 2017-06-21
rel_calib_stack: fix bug when using non-default windowing overlap fraction

megies:
In `signal.calibration.rel_calib_stack` there's a bug in spectrum scaling when using a non-default window overlap fraction (default is `0.5`, i.e. 50% overlap). This bug has been in there from the very beginning. So, **all calculations ever made using this function with non-default overlap fraction are wrong.**

Here's a plot from the regression test I added, showing how spectra are offset for different choices of window overlap fraction:

figure_1-1

The reason was that computation of how many windows will be computed did not account for used overlap. Fix is trivial, as the FFT helper routine returns all individual FFTs, we can just look at the shape of the FFT result.

I'll add a regression test and a fix.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -142,6 +142,8 @@ master: (doi: 10.5281/zenodo.165135)
      alongside gaps (see #1366)
    * obspy-plot now has option to disable min/max plot (see #1583)
  - obspy.signal:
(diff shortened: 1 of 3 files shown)


megies:
CC @jwassermann @mbdb  since you're using this..

PR #1829, opened 2017-06-28
Added check for response function

s-schneider:
### What does this PR do?
Adds write support for AH (Ad Hoc) format to stream.write(), It is an addition to PR #1757
related to issue #1754
### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/ah/core.py
+++ b/obspy/io/ah/core.py
@@ -382,11 +382,33 @@ def _pack_trace_wout_ah_dict(tr, packer, codesize, chansize,
     packer.pack_float(0)
 
     # Poles and Zeros are not provided by stream object, are set to 0
-    for _i in range(0, 30):
+    if hasattr(tr.stats, 'response'):
+        poles, zeros = tr.stats.response.get_paz()
+        packer.pack_float(len(poles))
         packer.pack_float(0)
+        packer.pack_float(len(zeros))
         packer.pack_float(0)
-        packer.pack_float(0)
-        packer.pack_float(0)
+
+        for _i in range(1, 30):
+            try:
+                r, i = poles[_i].real, poles[_i].imag
+            except IndexError:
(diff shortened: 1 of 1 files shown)


obspy/io/ah/core.py
```diff
@@ -382,12 +382,34 @@ def _pack_trace_wout_ah_dict(tr, packer, codesize, chansize,
     packer.pack_float(0)
 
     # Poles and Zeros are not provided by stream object, are set to 0
-    for _i in range(0, 30):
-        packer.pack_float(0)
-        packer.pack_float(0)
+    if hasattr(tr.stats, 'response'):
+        poles, zeros = tr.stats.response.get_paz()
```
megies: In general we nowadays don't encourage "magic uses" of additional items in `Trace.stats` anymore.. but if it's important for you we might still accept this change I guess..

s-schneider: Would there be a better way to check if there is a response attached?

krischer:
I'd personally either like to see this as `tr.stats.ah.response` or `Stream.write(..., format="AH", response=response)`. Adding important things to `tr.stats` is a hack in any case and we should not keep doing this - there are already enough places in our code base that do that.

s-schneider:
I use the information provided for instance by the `get_waveforms(attach_response=True)` of `obspy.clients.fdsn`. I didn't add that information manually

krischer:
Yea- this is one of the instances in our code base where it does already happen. I'd be fine with a construct like this which offers a cleaner interface but also falls back to using the existing `tr.stats.response`:

```python
def _write_ah(..., response=None, ...):
    ...
    # First use passed response, otherwise look in trace.stats.
    if response is None and "response" in tr.stats:
        response = tr.stats.response

    if response:
        ...
```

PR #1832, opened 2017-06-30
utcdatetime: loading a numpy binary with UTCDateTime objects fails

megies:
Loading a numpy binary with UTCDateTime objects that was saved on ObsPy < 1.1 (i.e. before the nanosecond switch, see #1325) fails..

This is likely a very similar issue to #1664.

This adds a failing test to illustrate the issue.


krischer:
Do we really care? These things are bound to break and I'm not sure we even have to guarantee stability for these cases.

The numpy binary format is mainly intended to offer a light-weight and fast storage for numpy arrays. Object arrays I think are just pickled so I'm actually kind of surprised it does not work already.

megies:
> Do we really care? These things are bound to break and I'm not sure we even have to guarantee stability for these cases.

I'm not serializing UTCDateTime objects myself, would always go for the underlying timestamps (or matplotlib date representation floats). But I got sent a numpy binary .npy file from a student who saved it on obspy 1.0. So I would argue we should work around this, and I'm fairly sure that the fix is pretty simple, just didn't have time to work on it yet.

krischer:
If its easy to do I don't mind adding it but IMHO we should not put a lot of effort into this.

trichter:
I am not familiar with pickle. Could something along these lines work (methods for `UTCDateTime`)? (reminds me of the hash discussion)

```
    def __getstate__(self):
        return (self.timestamp, self.precision)

    def __setstate__(self, state):
        self.timestamp = state[0]
        self.precision = state[1]
```

megies:
Punting to 1.1.1.. otherwise we'll never even get to 1.1.0

megies:
Changed base branch to `maintenance_1.1.x`.

d-chambers:
Is this student's work still relevant @megies ? Maybe we can close this now.

megies:
Yeah, let's close it

PR #1842, opened 2017-07-13
attach_coordinates() method for Trace and Stream

claudiodsf:
Similar to `attach_response()`, this method search for and attaches coordinates from an `Inventory` object.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -2949,6 +2949,50 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         return new_stream
 
     @map_example_filename("inventories")
+    def attach_coordinates(self, inventories):
+        """
+        Search for and attach channel coordinates to each trace as
+        trace.stats.coordinates. Does not raise an exception but shows a
+        warning if channel coordinates can not be found for all traces.
+        Returns a list of traces for which no coordinates could be found.
+
+        >>> from obspy import read, read_inventory
+        >>> st = read()
+        >>> inv = read_inventory()
+        >>> st.attach_coordinates(inv)
+        []
+        >>> tr = st[0]
+        >>> print(tr.stats.coordinates.latitude)  \
+                # doctest: +NORMALIZE_WHITESPACE
+        47.737167
+        >>> print(tr.stats.coordinates.longitude)  \
+                # doctest: +NORMALIZE_WHITESPACE
+        12.795714
+        >>> print(tr.stats.coordinates.elevation)  \
+                # doctest: +NORMALIZE_WHITESPACE
+        860.0
+
+        :type inventories: :class:`~obspy.core.inventory.inventory.Inventory`
+            or :class:`~obspy.core.inventory.network.Network` or a list
+            containing objects of these types.
+        :param inventories: Station metadata to use in search for coordinates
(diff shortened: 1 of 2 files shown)


claudiodsf:
Is dockets enough? Or do you want explicit testing?

claudiodsf:
+DOCS please

megies:
Actually, @krischer and me too, we've kind of arrived at the opinion that inventing `attach_response()` was a design mistake. At a first glance it seems super convenient, but when you look at it in detail it..

 - adds non-transparent magic, by shifting weight to essentially undocumented (mis-?)use of non-default keys in `Trace.stats`
 - adds problematic scenarios, mostly stemming from the above non-default-ness of additional keys in `Trace.stats` that are not handled canonically when making transformations on traces (e.g. what happens when merging two traces with different responses/coordinates?)

I think I'm the one to blame for adding `attach_response()`, not putting enough thought, following along with the old use of `Trace.stats.paz` that we used sloppily already from very old times. But.. we already did some (only initially) slightly painful backtracking by introducing `inventory` as first `arg` in `Trace.remove_response()` to get rid of `Trace.stats` misuse, and I fear this PR will not get a nod, I hope the above was enough to convince you.

As I said, we made the same mistake several times but recognized now that this is not the right way to go. Wherever you need coordinates, you should put `inventory` as first `arg` to avoid non-transparent `stats` magic. And finally.. we discussed this many times, at the end of the day what we want here is a canonical new object structure, working title "DataSet", that combines waveform+station[+event] data.

claudiodsf:
@megies I'm fine with that. Just pay attention to the fact that in some cases `trace.stats.coordinates` seems required, like for instance in `Stream.plot()` (https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream.plot.html)

claudiodsf:
...or in https://docs.obspy.org/packages/autogen/obspy.signal.array_analysis.get_geometry.html

megies:
> Just pay attention to the fact that in some cases trace.stats.coordinates seems required, like for instance in Stream.plot()

> ...or in https://docs.obspy.org/packages/autogen/obspy.signal.array_analysis.get_geometry.html

Yes, I know. And all of these instances have to be considered bad legacy practice.. :-/

krischer:
I second @megies view. All these instances are design mistakes in retrospect and we should rather add `inventory` arguments to all of these to augment them.

I'm also not sure if the `_get_coordinates()` is needed as `inv.get_coordinates(tr.id)` is IMHO pretty simple to use.

megies:
@claudiodsf if you want to put some effort into this, you could channel your energy towards the long planned `DataSet` object. It's supposed to eventually circumvent all these dirty helper stuff, like attaching response to traces and so on..

d-chambers:
should we close this?

megies:
Yeah, I think we should close this. Lookups should always be dynamic from an Inventory and not static from some info attached to a Trace. Having `attach_response()` was clearly a flaw in design introduced by us.

megies:
Sorry @claudiodsf that you put some work into this and we kinda decide not to merge.. :worried:

claudiodsf:
:cry:

... just kidding: as I told you, I'm fine with that ;)

PR #1845, opened 2017-07-23
implemented load_entry_point cache and check_compression param of read

d-chambers:
### What does this PR do?

This PR is intended to speed up reading mseed files through two methods:

1. Caching outputs of pkg_resources.load_entry_point which can be rather slow 

2. Add the check_compression parameter to obspy.read function for skipping potentially expensive compression checks (from tar and zip libraries).  

This has also reduced the run time of the entire obspy test suit by 5% to 10%, and since obspy uses plugins heavily I suspect it will speed up many other obspy functions. 

See [this gist](https://gist.github.com/d-chambers/1eabd9c2147b11db0e68cf57ac3a8ad1)

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_stream.py
+++ b/obspy/core/tests/test_stream.py
@@ -2484,6 +2484,26 @@ class StreamTestCase(unittest.TestCase):
             self.assertEqual(arg[1]["order"], 2)
             self.assertEqual(arg[1]["plot"], True)
 
+    def test_read_no_check_compression(self):
+        """
+        Test to ensure calling read with check_compression=False does not 
+        call expensive tar or zip functions  
+        """
+        # write a file to disk
+        file_name = 'temp.mseed'
+        st = read()
+        st.write(file_name, 'mseed')  # maybe read from test data instead?
+
+        with mock.patch("tarfile.is_tarfile") as tar_patch:
+            with mock.patch("zipfile.is_zipfile") as zip_patch:
+                _ = read(file_name, check_compression=False)
+
+        # delete temp file
+        os.remove(file_name)
+        # assert neither compression check function was called.
+        self.assertEqual(tar_patch.call_count, 0)
+        self.assertEqual(zip_patch.call_count, 0)
+
 
 def suite():
     return unittest.makeSuite(StreamTestCase, 'test')
--- a/obspy/core/tests/test_stream.py
+++ b/obspy/core/tests/test_stream.py
@@ -2486,8 +2486,8 @@ class StreamTestCase(unittest.TestCase):
 
     def test_read_no_check_compression(self):
         """
-        Test to ensure calling read with check_compression=False does not 
-        call expensive tar or zip functions  
+        Test to ensure calling read with check_compression=False does not
+        call expensive tar or zip functions.
         """
         # write a file to disk
         file_name = 'temp.mseed'
@@ -2496,7 +2496,7 @@ class StreamTestCase(unittest.TestCase):
 
         with mock.patch("tarfile.is_tarfile") as tar_patch:
             with mock.patch("zipfile.is_zipfile") as zip_patch:
-                _ = read(file_name, check_compression=False)
+                read(file_name, check_compression=False)
 
         # delete temp file
         os.remove(file_name)
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -10,26 +10,24 @@ Base utilities and constants for ObsPy.
 """
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
-from future.builtins import *  # NOQA @UnusedWildImport
-from future.utils import native_str
+from future.builtins import *  # NOQA
 
 import doctest
 import functools
 import inspect
 import io
 import os
-import pkg_resources
 import sys
 import tempfile
 from collections import OrderedDict
 
-from pkg_resources import iter_entry_points, load_entry_point
 import numpy as np
-
+import pkg_resources
 import requests
+from future.utils import native_str
+from pkg_resources import iter_entry_points, load_entry_point
 
 from obspy.core.util.misc import to_int_or_zero
-
 
 # defining ObsPy modules currently used by runtests and the path function
 DEFAULT_MODULES = ['clients.filesystem', 'core', 'db', 'geodetics', 'imaging',
@@ -61,22 +59,28 @@ _sys_is_le = sys.byteorder == 'little'
 NATIVE_BYTEORDER = _sys_is_le and '<' or '>'
 
 
-# apply simple memoization cache to load_entry_points
-def decorate_load_entry_point(func):
+# apply simple memoization cache on load_entry_points
+# this function cannot go in decorator.py due to circular import issues
+def _load_entry_point_decorator(func):
+    """
+    Decorate pkg_resources' load_entry_point function to cache outputs.
+    """
     cache = {}
-    sig = inspect.signature(func)
+    expected_keys = ('dist', 'group', 'name')
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        b = sig.bind(*args, **kwargs).args
-        if not b in cache:
-            cache[b] = func(*args, **kwargs)
-        return cache[b]
+        # make a str key from the inputs to func for hashing in cache
+        input_dict = inspect.getcallargs(func, *args, **kwargs)
+        hash_str = '/'.join([input_dict[x] for x in expected_keys])
+        if hash_str not in cache:
+            cache[hash_str] = func(*args, **kwargs)
+        return cache[hash_str]
 
     return wrapper
 
 
-load_entry_point = decorate_load_entry_point(load_entry_point)
+load_entry_point = _load_entry_point_decorator(load_entry_point)
 
 
 class NamedTemporaryFile(io.BufferedIOBase):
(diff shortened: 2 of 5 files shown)


megies:
Looks good from a first glance..

obspy/core/util/base.py
```diff
@@ -10,26 +10,25 @@
...
 import sys
 import tempfile
 from collections import OrderedDict
 
-from pkg_resources import iter_entry_points, load_entry_point
 import numpy as np
-
+import pkg_resources
 import requests
+from future.utils import native_str
```
megies: Only now saw this @d-chambers, why did you move this down?

obspy/core/util/base.py
```diff
@@ -60,6 +59,30 @@
...
+        input_dict = inspect.getcallargs(func, *args, **kwargs)
+        hash_str = '/'.join([input_dict[x] for x in expected_keys])
+        if hash_str not in cache:
+            cache[hash_str] = func(*args, **kwargs)
+        return cache[hash_str]
+
+    return wrapper
+
+
+load_entry_point = _load_entry_point_decorator(load_entry_point)
```
megies: Hmm.. is this patching it at `pkg_resources` directly, actually? If so, it might affect other installed python packages.. and I'd vote for only patching it inside ObsPy (i.e. just replace `load_entry_point` inside obspy with a cached version)..

megies:
I didn't really check, have to admit.. also I guess with the heavy fluctuations in CI runtime, you'd really have to do heavy statistics to see such a change.

For me locally it reduces testsuite runtime by 8% (only did a single run though).

obspy/core/tests/test_stream.py
```diff
@@ -2484,6 +2484,26 @@ def test_passing_kwargs_to_trace_detrend(self):
...
+        call expensive tar or zip functions.
+        """
+        # write a file to disk
+        file_name = 'temp.mseed'
+        st = read()
+        st.write(file_name, 'mseed')  # maybe read from test data instead?
+
+        with mock.patch("tarfile.is_tarfile") as tar_patch:
+            with mock.patch("zipfile.is_zipfile") as zip_patch:
+                read(file_name, check_compression=False)
```
megies: You could also add the opposite test, i.e. that the unzip check routines get called if `check_compression` is not overridden.

obspy/core/util/base.py
d-chambers: I am not sure, it might have been my IDE.

obspy/core/util/base.py
d-chambers: I will double check but I am fairly sure the wrapped function is only going to be bound to the base module, so unless another program accesses  it through obspy.core.util.base.load_entry_point it should be unaffected.

krischer: Yes this will only affect `obspy.core.util.base.load_entry_point` so all should be good.

megies:
> d-chambers restored the d-chambers:read-optimization branch a day ago

Sorry, as I already merged this PR you'll have to do a new PR..

krischer:
Pretty cool :) Did you check if the runtime reduction comes from the entry point memoization or from the archiving format checks?

A side note: If we ever get around to refactor our waveform format plugins: We could just manually check for the magic numbers of zip and tar files and the archiving format check would be almost free - right now it is fairly expensive because it might have to open the file one or more additional times.

obspy/core/util/base.py
megies: > Yes this will only affect obspy.core.util.base.load_entry_point so all should be good.

Still a bit confused about this, as we import `from pkg_resources` throughout the code base wherever this function is used and not `from obspy.core.util.base`..

```
obspy/core/event/catalog.py
32:from pkg_resources import load_entry_point
496:            write_format = load_entry_point(

obspy/core/inventory/inventory.py
20:from pkg_resources import load_entry_point
270:            write_format = load_entry_point(

obspy/core/util/base.py
28:from pkg_resources import iter_entry_points, load_entry_point
62:# apply simple memoization cache on load_entry_points
64:def _load_entry_point_decorator(func):
66:    Decorate pkg_resources' load_entry_point function to cache outputs.
73:        print("in _load_entry_point_decorator")
84:load_entry_point = _load_entry_point_decorator(load_entry_point)
347:    func = load_entry_point(entry_point.dist.key, 'obspy.plugin.%s' % (group),
396:            is_format = load_entry_point(
426:        read_format = load_entry_point(
495:        func = load_entry_point(ep.dist.key,

obspy/core/stream.py
25:from pkg_resources import load_entry_point
1436:            write_format = load_entry_point(

obspy/core/tests/test_waveform_plugins.py
14:from pkg_resources import load_entry_point
49:                is_format = load_entry_point(
205:            is_format = load_entry_point(
462:            fct_name = module_name + "." + load_entry_point(
```

But at least in an interactive IPython session I don't see this affecting `pkg_resources`, so it indeed seems to be fine, I guess.

krischer: Ah sorry - I did not read the PR carefully enough - yea then this is not used by ObsPy. It will really only set in `obspy.core.util.base.load_entry_point` so this would also have to be imported everywhere. It would of course be possible to monkey patch the `load_entry_point` function in `pkg_resources` but this is (i) dangerous for other packages and (ii) it would need to happen before it is imported everywhere else in ObsPy so we would introduce some hard-to-debug import order logic...

I'd personally favor to just import the modified `load_entry_point` everywhere in ObsPy - and maybe also just give it a different name and just code it into a function without the decorator as that would be easier to understand.

Can someone open a new PR to fix and also add a test for this?

megies: > yea then this is not used by ObsPy

It actually is, that's what puzzles me.. (In the example I've added a print statement inside the decorator)

```python
In [1]: cd git/obspy/obspy/io/mseed/tests/data/
/home/megies/git/obspy/obspy/io/mseed/tests/data

In [2]: st = read('two_channels.mseed')
in _load_entry_point_decorator
in _load_entry_point_decorator

In [3]: st = read('two_channels.mseed')
in _load_entry_point_decorator
in _load_entry_point_decorator

In [4]: st = read('two_channels.mseed')
in _load_entry_point_decorator
in _load_entry_point_decorator

In [5]: from pkg_resources import load_entry_point

In [6]: load_entry_point('obspy', 'obspy.plugin.waveform.MSEED', 'readFormat')
Out[6]: <function obspy.io.mseed.core._read_mseed>

In [7]: from obspy.core.util.base import load_entry_point

In [8]: load_entry_point('obspy', 'obspy.plugin.waveform.MSEED', 'readFormat')
in _load_entry_point_decorator
Out[8]: <function obspy.io.mseed.core._read_mseed>

In [9]: load_entry_point('obspy', 'obspy.plugin.waveform.MSEED', 'readFormat')
in _load_entry_point_decorator
Out[9]: <function obspy.io.mseed.core._read_mseed>

In [10]: from pkg_resources import load_entry_point

In [11]: load_entry_point('obspy', 'obspy.plugin.waveform.MSEED', 'readFormat')
Out[11]: <function obspy.io.mseed.core._read_mseed>
```

d-chambers:
I should have time for this (hopefully) this weekend. I will open a new PR and define a function to do this. Which module should such a function live in? I don't want to get hung up with a bunch of circular imports.

d-chambers:
@krischer the speed up in the test execution should be exclusively from the entry point caching, the default behavior is still to check for compression on streams, and there is only one test, which I added, that skips the compression.

megies:
> Which module should such a function live in?

Assuming that you're talking about the zip magic number check, I would probably put it in `obspy.core.util.misc`.

> I don't want to get hung up with a bunch of circular imports.

Well, sometimes these are a problem, but we can think about it if we really run into one. At least we'll notice right away when running the tests.

d-chambers:
> Assuming that you're talking about the zip magic number check
Sorry I ment the obspy load entry point function. Probably the same answer applies though.

megies:
> Sorry I ment the obspy load entry point function. Probably the same answer applies though.

Uhm.. well, you already placed it in `obspy/core/util/base.py`?? Fits in there just fine, I'd say. I'm just puzzled because the new decorated `load_entry_point` function is defined in there, but across the code base it's used by doing a `from pkg_resources import load_entry_point` (which kind of implies to me that pkg_resources might globally be affected) and it's still using the supposedly local `load_entry_point` in `obspy.core.util.base`.

d-chambers:
I am planning to open a new PR to undo this one. In the new PR I will follow  @krischer 's advice:

> I'd personally favor to just import the modified load_entry_point everywhere in ObsPy - and maybe also just give it a different name and just code it into a function without the decorator as that would be easier to understand.

So I will:
1. Get rid of the decorator
2. define a new function in obspy.util.core.misc which does the caching
3. make sure all other obspy functions use the new function to rather than from `pkg_resources.load_entry_point` to load entry points

This should be easier to understand and clear up the confusion.

obspy/core/util/base.py
krischer: I can confirm this behavior but I am honestly a bit puzzled by it. Does anyone understand why Python behaves like this here?

krischer: Ah I got it - `load_entry_points` for reading is always only called in `base.py` which of course uses the decorated ones. All other calls in the code base only use it for writing (and some tests) but these would still use the normal one.

krischer:
Sounds good! Maybe call it `buffered_load_entry_points()` or something along the line to make the distinction clear.

PR #1847, opened 2017-08-07
Stream.select() for components now also works for components with less than 3 letters

krischer:
A fair number of data (synthetics, active source data, ...) does not have three letter channel names so `Stream.select(component="X")` did not work as it only worked for traces that has three letters. With this PR it now just takes the last letter.

@megies I've see that this is intentional behavior (including a test case) from you. Why is this needed? I've been bitten by this quite a number of times so I've written this PR but if there is a good reason not to have this I'm also fine with dropping it.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -44,6 +44,8 @@ master: (doi: 10.5281/zenodo.165135)
    * Order of extra tags for event type classes serialized to QuakeML can now
      be controlled by using an OrderedDict (see #1617)
    * Bode plots can now optionally plot the phase in degrees (see #1763).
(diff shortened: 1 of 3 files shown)


Jollyfant:
> With this PR it now just takes the last letter.

What do you mean? Because `channel=Z` should not match `BHZ`. That's when you put `*Z`.

krischer:
Sorry - that was a type in the desription - I just fixed it. I meant for the `component` level filtering - not channel.

This PR essentially removes this line: https://github.com/obspy/obspy/blob/fc0d0cb8c23e7b9f1aa83b31587e5a1934b56241/obspy/core/stream.py#L1812-L1813

PR #1638, opened 2017-01-10
Add SC3ML event write support

Brtle:
Hi,

I did an XSLT to convert quakeml 1.2 into sc3ml 0.8. I think it could be valuable to obspy so I put it in a small module. It writes sc3ml from a Catalog.

I didn't do the read module but it shouldn't be too hard to add it since an XSLT is already given (http://geofon.gfz-potsdam.de/schema/0.8/).

I'm using the QuakeML unit test data but there are still 126Ko of data. Should I remove some of them?

PR #1848, opened 2017-08-08
Add SC3ML event read module

Brtle:
I did the [SC3ML event write module](https://github.com/obspy/obspy/pull/1638) some time ago, here is the read module!

Once again, an XSLT is used to convert SC3ML file to a QuakeML file. This QuakeML file is then read with the QuakeML module. The original XSLT file was created by [Stephan Herrnkind](http://geofon.gfz-potsdam.de/schema/) who kindly changed its license from GPL to LGPL to be compatible with the ObsPy license, thank you very much to him! I added some modifications which are indicated in the file itself.

I also bring some modifications to the write module. The version attribute have been added to the root node (optional but used in the [`_is_sc3ml`](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/sc3ml.py#L44) method). The unit tests have been modified a bit to be more homogeneous with the tests of the read module. The [`_is_sc3ml`](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/sc3ml.py#L44) method can now receive string parameter in addition to filename or file-like object ([this test](https://github.com/obspy/obspy/blob/master/obspy/io/quakeml/tests/test_quakeml.py#L658) was failing randomly since the [`_is_sc3ml`](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/sc3ml.py#L44) didn't accept string parameter). For that, I used the [_xml_doc_from_anything](https://github.com/obspy/obspy/blob/master/obspy/io/quakeml/core.py#L64) method from the QuakeML module. Let me know if you prefere that I don't use this method since it's in another module.


megies:
Looks good from a first glance. @Jollyfant you're more into this topic I think, maybe you can give this a review when you have some free time?

obspy/io/seiscomp/tests/test_event.py
```diff
@@ -34,69 +38,192 @@ def setUp(self):
...
+    def test_read_example(self):
+        filename = os.path.join(self.path, 'qml-example-1.2-RC3.sc3ml')
+        catalog = _read_sc3ml(filename)
+
+        self.assertEqual(len(catalog.events), 1)
+        self.assertEqual(len(catalog.events[0].origins), 1)
 
-    def test_event(self):
-        self.cmp_file('quakeml_1.2_event.xml', 'quakeml_1.2_event.sc3ml')
+    def test_read_ID_prefix(self):
```
megies: Guess you'll have to change it to `test_read_id_prefix` to make flake8 happy (or add `  # NOQA` but renaming is not a problem so better not circumvent flake8)

Brtle: Done!

Brtle:
Trigger docs build +DOCS

megies:
Docs buildbot is done: http://docs.obspy.org/pull-requests/1848/

obspy/io/seiscomp/event.py
```diff
@@ -59,6 +66,47 @@ def _validate_sc3ml(path_or_object, verbose=False):
...
+    This function should NOT be called directly, it registers via the
+    the :meth:`~obspy.core.event.catalog.Catalog.write` method of an
+    ObsPy :class:`~obspy.core.event.catalog.Catalog` object, call this
+    instead.
+
+    :type filename: str
+    :param filename: SC3ML file to be read.
+    :type id_prefix: str
+    :param id_prefix: ID prefix. SC3ML does not enforce any particular ID
+        restriction, this ID prefix allow to convert the IDs to a well
```
krischer: s/allow/allows

obspy/io/seiscomp/sc3ml.py
```diff
@@ -55,34 +56,39 @@ def _is_sc3ml(path_or_file_object):
...
+    if hasattr(xmldoc, "getroot"):
+        root = xmldoc.getroot()
+    else:
+        root = xmldoc
+
+    try:
+        match = re.match(
+            r'{http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/[-+]?'
+            r'[0-9]*\.?[0-9]+}', root.tag)
+        assert match is not None
```
krischer: Best remove the `assert` statement and replace it with an explicit exception as assert calls are removed when running Python is run with `-O`. Admittedly: few people do this but better be safe then sorry.

obspy/io/seiscomp/sc3ml.py
```diff
@@ -55,34 +56,39 @@ def _is_sc3ml(path_or_file_object):
...
+                pass
+
+    if hasattr(xmldoc, "getroot"):
+        root = xmldoc.getroot()
+    else:
+        root = xmldoc
+
+    try:
+        match = re.match(
+            r'{http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/[-+]?'
```
krischer: The dots should probably be escaped with a backslash.

Brtle:
I just wanted to fix the regex and the assert and I finally refactored a little more…

I create a new file `core.py` where I put the common files to both event and inventory module. I fixed the `_is_sc3ml` method and I also factorize the `validate` method.

The `_is_sc3ml` function also no longer uses the optional version attribute but only the namespace.

Jollyfant:
That's a nice refactor, thanks. I'll try and review a bit more soon.

obspy/io/seiscomp/tests/data/iris_events.sc3ml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<seiscomp xmlns="http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.9">
+<seiscomp xmlns="http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.9" version="0.9">
   <EventParameters publicID="smi:www.iris.edu/ws/event/query">
     <origin publicID="smi:www.iris.edu/ws/event/query?originId=7680412">
       <time>
-        <value>2011-03-11T05:46:24.120000Z</value>
+        <value>2011-03-11T05:46:24.1200</value>
```
Jollyfant: Why is the time being changed?

obspy/io/seiscomp/event.py
```diff
@@ -20,43 +20,71 @@
...
+    xslt_filename = os.path.join(os.path.dirname(__file__), 'data',
+                                 'sc3ml_0.9__quakeml_1.2.xsl')
+    transform = etree.XSLT(etree.parse(xslt_filename))
+    sc3ml_doc = _xml_doc_from_anything(filename)
+    quakeml_doc = transform(sc3ml_doc,
+                            ID_PREFIX=etree.XSLT.strparam(id_prefix))
+    return Unpickler().load(io.BytesIO(quakeml_doc))
 
 
 def _write_sc3ml(catalog, filename, validate=False, verbose=False,
```
Jollyfant: Instead of using `open()` it is usually better practise to use `with open()`. See at the bottom of this paragraph:

https://docs.python.org/2/tutorial/inputoutput.html#methods-of-file-objects

obspy/io/seiscomp/tests/data/iris_events.sc3ml
Brtle: The write module was tested by reading QuakeML files and writing SC3ML files using ObsPy [read_events](https://docs.obspy.org/master/packages/autogen/obspy.core.event.read_events.html) method. Now, the unit tests don't use the [read_events](https://docs.obspy.org/master/packages/autogen/obspy.core.event.read_events.html) method, the QuakeML file is directly converted in SC3ML with the XSLT file. I think it's more accurate since what we really want to test is the XSLT file. However, the result isn't exactly the same. The date format of the original file isn't modified when using the XSLT file.

By the way, sorry for the late answer!

obspy/io/seiscomp/event.py
Brtle: I don't see any `open()` in your quote. Moreover, the [lxml FAQ](http://lxml.de/FAQ.html#can-lxml-parse-from-file-objects-opened-in-unicode-text-mode) states that:

```text
if possible, prefer passing the file path directly into parse() instead of an opened Python file object.
```

Maybe you're speaking of [this](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/event.py#L107) `open()`, which was copied from the [QuakeML module](https://github.com/obspy/obspy/blob/master/obspy/io/quakeml/core.py#L1832). I can eventually use a `with open()` but I'll also modifie the QuakeML module then.

Jollyfant: @Brtle yeah that's the one I was referring to. Sorry, I commented on the wrong function.

Brtle: Done, I put a `with open()` in both SC3ML and QuakeML module.

Brtle:
I don't have the right to merge it.

Jollyfant:
Looks like Travis CI got stuck on the callback. But it looks like tests passed. @megies is it OK to ignore Travis and merge?

PR #1854, opened 2017-08-12
Python only CatchOutput

barsch:
had to remove C level support due to http://bugs.python.org/issue30555
(_io._WindowsConsoleIO breaks in the face of fd redirection)

continues #1684 - had to close the other one as merging current master became a real pain ;/

### DIFF
--- a/obspy/core/tests/test_util_misc.py
+++ b/obspy/core/tests/test_util_misc.py
@@ -2,6 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
+from future.utils import PY2
 
 import os
 import sys
@@ -49,6 +50,18 @@ class UtilMiscTestCase(unittest.TestCase):
             print("test_catch_output #5", file=sys.stderr)
         self.assertEqual(out.stdout, 'test_catch_output #4\n')
         self.assertEqual(out.stderr, 'test_catch_output #5\n')
+
+    def test_catch_output_bytes(self):
+        with CatchOutput() as out:
+            if PY2:
+                sys.stdout.write(b"test_catch_output_bytes #1")
+                sys.stderr.write(b"test_catch_output_bytes #2")
+            else:
+                # PY3 does not allow to write directly bytes into text streams
+                sys.stdout.buffer.write(b"test_catch_output_bytes #1")
+                sys.stderr.buffer.write(b"test_catch_output_bytes #2")
+        self.assertEqual(out.stdout, 'test_catch_output_bytes #1')
+        self.assertEqual(out.stderr, 'test_catch_output_bytes #2')
 
     def test_catch_output_io(self):
         """
--- a/obspy/core/tests/test_util_misc.py
+++ b/obspy/core/tests/test_util_misc.py
@@ -4,12 +4,9 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 
 import os
-import platform
 import sys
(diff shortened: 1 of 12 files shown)


barsch:
imho ready for review - at least Win 3.6 does not crash anymore ...

Return values from the ```CatchOutput``` context manager are now normalized (unicode & same newline char) - hopefully no more encoding issues in tests itself ...

looks also like some C level messages are suppressed some not, maybe we can handle the following messages within C to bring in in line ...:

* ```obspy.core.tests.test_response.ResponseTestCase.test_segfault_after_error_handling``` with
   ```
   EVRESP ERROR (... [File: <stdin>; Start date: ; Stage: 2]):
	check_channel; units mismatch between stages,
	skipping to next response now
   ```
* ```obspy.io.gse2.tests.test_libgse2.LibGSE2TestCase.test_no_dat2_null_pointer``` with
   
   ```
   decomp_6b: Neither DAT2 or DAT1 found!
   ```

barsch:
btw: why does codecov/patch complain - about all those files which I didn't even touch ... - I really would opt for removing this discouraging check!

Jollyfant:
I don't think anyone really knows why codecov does that.. I always pretty much ignore it.

megies:
> btw: why does codecov/patch complain about all those files which I didn't even touch ... - I really would opt to remove this discouraging check!

@barsch I'm assuming you're referring to [this tab labelled "changes"](https://codecov.io/gh/obspy/obspy/compare/5dc0cb774242e6065f7fc8d4f39f607d777ba7e6...d83048b79683766fabbbfdc2ae7fe6fd1508b703/changes)?

This option is [already disabled in the settings](https://github.com/obspy/obspy/blob/master/codecov.yml#L28-L30), and I agree that it's very confusing and not very helpful. We never could figure out why it goes so utterly wrong. While codecov does not send a commit status for it anymore, it still computes this result, so you can find it if you go that page "manually".
But.. the fail for "codecov/patch" has nothing to do with the strange "changes" section, but rather with the [coverage percentage in the diff](https://codecov.io/gh/obspy/obspy/compare/5dc0cb774242e6065f7fc8d4f39f607d777ba7e6...d83048b79683766fabbbfdc2ae7fe6fd1508b703/diff) (in the new lines): `"85.92% of diff hit (target 90%)"`

If you look at the diff, it's clear that the missed lines are in Windows specific parts that won't be reached on Travis (Linux), e.g. [these lines in `_py36_windowsconsoleio_workaround()`](https://codecov.io/gh/obspy/obspy/compare/5dc0cb774242e6065f7fc8d4f39f607d777ba7e6...d83048b79683766fabbbfdc2ae7fe6fd1508b703/diff#D8-362).

So, this is caused by the Win-specific nature of the PR, not much that can be done, but disabling "codecov/patch" is not an option, IMHO. It's necessary to ensure proper coverage levels of new code coming in in PRs.

QuLogic:
> If you look at the diff, it's clear that the missed lines are in Windows specific parts that won't be reached on Travis (Linux), e.g. these lines in _py36_windowsconsoleio_workaround().

Is codecov not uploading on AppVeyor?

megies:
Ideally we should have a test to prevent new additions that add verbosity during tests.. because it's a real pain to fix it later on. But not sure how to check that.. :-/

megies:
> Is codecov not uploading on AppVeyor?

Nope.

https://github.com/obspy/obspy/blob/master/appveyor.yml#L122-L134

megies:
Thanks @barsch for fixing this one on Win!

PR #1855, opened 2017-08-14
Separate freq resp. calc from plotting

Jollyfant:
Hey, I needed ObsPy to return frequency response without plotting it. So I considered separating the calculation from the plotting in a new function. However, I'm not sure of the best approach to forward so many variables in the function. I'm using **locals() but myeh..

@megies ideas?

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1499,7 +1499,6 @@ class Response(ComparingObject):
             raise ZeroSamplingRate(msg)
 
         t_samp = 1.0 / sampling_rate
-        nyquist = sampling_rate / 2.0
         nfft = sampling_rate / min_freq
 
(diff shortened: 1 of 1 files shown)


megies:
Uhm.. we already have `Response.get_evalresp_response_for_frequencies()`..? https://github.com/obspy/obspy/blob/master/obspy/core/inventory/response.py#L1365

Jollyfant:
Hmmmmmmmmmmm

Jollyfant:
@megies this is not in version 1.0.3 right? :think:

megies:
Nope, sorry: https://github.com/obspy/obspy/blob/master/CHANGELOG.txt#L45

When me and @krischer are back from vacation we really need to release 1.1.0..  :tired_face:

PR #1857, opened 2017-08-22
Fix station format in FDSN client

Brtle:
The [get_stations](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/client.py#L406) method of the FDSN client has an optional parameter `format` with `xml` as default value. This parameter is consequently always given to the url of the FDSN webservice. However, this parameter is optional for the FDSN webservice. If the FDSN webservice doesn't implement the `format` parameter, the [get_stations](https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/client.py#L406) method will crash. Just giving a `None` default value solves the problem. The FDSN webservice returns a StationXML file by default anyway.

Since it's only a small change, I didn't make a unit test for that. Does that suit you?

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -411,7 +411,7 @@ class Client(object):
                      minradius=None, maxradius=None, level=None,
                      includerestricted=None, includeavailability=None,
                      updatedafter=None, matchtimeseries=None, filename=None,
-                     format="xml", **kwargs):
+                     format=None, **kwargs):
         """
         Query the station service of the FDSN client.
 


megies:
I don't see harm in this.. I'm triggering CI again, there was a build error somehow.

megies:
Hmm.. seems something changed at CircleCI.. :-( looks like we now have to manually prevent it from trying to install obspy..

QuLogic:
There's no Circle config on the maintenance branch, so it does its default thing of guessing.

megies:
Oh.. nice catch! @Brtle we won't do a 1.0.4 release, so currently everything goes into master. Only once 1.1.0 is out we will use another maintenance branch again.

I'll change the base branch for this PR.

Brtle:
I rebased from master and CircleCI seems to be fine.

PR #1867, opened 2017-08-29
FDSN mass downloader: sanitize_download: check file existence doing anything on it

paride:
### What does this PR do?

When using the mass downloader to download, for example, one year of data in 1-day chunks, sometimes some days are missing. If, later, the StationXML is missing too, ObsPy tries to do several things, but it fails when the miniseed was not existing in the first place.

This simple fix check that the file exists before trying anything.

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -42,6 +42,7 @@ Käufl, Paul
 Köhler, Andreas
 Lecocq, Thomas
 Leeman, John
+Legovini, Paride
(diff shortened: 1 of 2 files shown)


krischer:
I guess adding this does not hurt. I think it can only happen if a file is deleted from the outside, while the mass downloader is running? But this is a fair enough occurrence to catch.

Thanks a bunch for this and welcome to contributing to ObsPy :) Sorry for the delay in getting this reviewed.

PR #1876, opened 2017-09-05
Additional formatting when writing ASCII

### LINKED ISSUE #1869: ASCII SLIST uses only float32
Is there any particular reason why in the obspy.io.ascii module there is assumed that any float written or read by it has to be float32? 

I'm currently working on a script that involves saving signal with response removed to an ASCII file and unfortunately the accuracy that is used by default in ObsPy is not high enough for me. 

BTW: Comparing sample to sample with same signal but saved and read to SAC gives me a 10883 errors from numpy.testing.assertion_almost_equal for a signal 12221 samples long.

heavelock:
This PR adds possibility to use a custom formatter when writing to ASCII file formats. It is supposed to solve issue #1869 

### PR Checklist
- [x] All tests still pass.

Closes #1869


heavelock:
I've tried to add this feature with lowest impact on code possible. How does it fit the project's vision?

obspy/io/ascii/core.py
```diff
@@ -326,8 +328,12 @@ def _write_slist(stream, filename, **kwargs):  # @UnusedVariable
             elif trace.data.dtype.name.startswith('float'):
                 dtype = 'FLOAT'
                 fmt = '%f'
+
             else:
                 raise NotImplementedError
+            # fmt
+            if not custom_fmt is None:
```
barsch: ```if custom_fmt is not None``` is IMHO better readable

heavelock:
I might add this. The question is what to add to the header in that case? "Float"? IRIS does not define use of "DOUBLE" keyword in the header.

barsch:
the ```ASCII2MSEED``` writer just writes FLOAT too:

https://github.com/iris-edu/mseed2ascii/blob/f117564953ae6c1e0330483d3f176b334f841608/src/mseed2ascii.c#L233-L236

so I assume ```FLOAT``` is fine - but maybe @chad-iris can comment on that ...

barsch:
btw ```ASCIIMSEED``` uses also a different ```fmt``` string for ```float32``` - maybe we should fix this while you are on it ;)

https://github.com/iris-edu/mseed2ascii/blob/f117564953ae6c1e0330483d3f176b334f841608/src/mseed2ascii.c#L312-L325

megies:
> shouldn't it automatically write double values if we have np.float64 data?

I think the concept of float/double is a bit mixed up here. What is currently in the code is `%f` which IMHO doesn't have much to do with single/double precision and the actual precision in the output ascii depends on the value that is to be stored. It's simply using a fixed number of decimal places in text representation and that's the problem.

```
In [7]: '%f' % 1e-10
Out[7]: '0.000000'
```

As I pointed out before, the main gotcha is to use scientific notation, which will always yield the same number of valid decimals in text output.

```
In [8]: '%+.10e' % 1e-10
Out[8]: '+1.0000000000e-10'
```

As @barsch mentioned the format specs don't say anything about valid formatting, but since `%g` is in that `mseed2ascii` program, I guess we can also safely use it.

I'm not sure we even need the custom fmt option if we use a safe default, but if you want to add it (and since you added a test), I guess that's fine with me as well.

obspy/io/ascii/core.py
```diff
@@ -434,8 +443,9 @@ def _write_tspair(stream, filename, **kwargs):  # @UnusedVariable
             elif trace.data.dtype.name.startswith('float'):
                 dtype = 'FLOAT'
                 fmt = '%f'
```
megies: I'd change the default to `%+.10e` or at least to something that dynamically switches to scientific notation like `%.10g`.

heavelock: Okay!. Should I discriminate between float32('%+.8e') and float64('%+.10e') just like mseed2ascii?

heavelock:
Okay, so I decided to use a scientific notation as @megies suggested and discriminate between float32('%+.8e') and float64('%+.10e') since the mseed2sac does that. What do you think about that discrimination? Should I also modify the reading method to take into account that the might occur doubles?

I think that I would opt to leave the custom formatter argument since it doesn't harm and might be useful for somebody.

barsch:
distinction between ```float32```/```float64``` and using ```%+.8g```/```%+.10g``` is definitiv the way to go

concerning ```custom_fmt``` - I would leave it in - but we should maybe use the new keyword ```CUSTOM``` instead of ```FLOAT``` as one could provide any other string formating string ... e.g. ```%e```, ```%x```

barsch:
@heavelock ```mseed2ascii``` uses ```%g``` not ```%e```

barsch:
> Should I also modify the reading method to take into account that the might occur doubles?
yes please

files with ```CUSTOM``` data will raise a ```NotImplementedError``` - which is fine I guess - i wouldn't try to cover that in read ...

heavelock:
Okay, so since we are discussing adding a CUSTOM, maybe we should also add DOUBLE? Afterwards reading of DOUBLE would create a float64 and CUSTOM would raise an error.

I forgot about updating tests, will be in next commit.

megies:
> @heavelock mseed2ascii uses %g not %e

Just a matter of taste, I'd say. `%g` might save some space, `%e` aligns all columns equally (but also only with `+` modifier)

megies:
Adding a new `CUSTOM` keyword to SLIST/TSPAIR header is IMHO a breach of the format spec. I don't think we want to do that.

http://ds.iris.edu/ds/nodes/dmc/data/formats/simple-ascii/
```
Type | Sample type ‘INTEGER’, ‘FLOAT’ or ‘ASCII’
```

barsch:
as ```ascii2mseed``` is pretty much our only reference I would suggest to stay with ```FLOAT``` for double values ..., which also could cover future usage of ```np.float128``` - ofc not using ```DOUBLE``` complicates the read method ...

megies:
> distinction between float32/float64 and using %+.8g/%+.10g is definitiv the way to go

For the default case of no custom specifier given: Why not just always stick with `FLOAT` (as per format spec) and always use `%+.10g` when writing and always deserialize to `numpy.float64`?

heavelock:
The only reason for that distinction was a mseed2ascii, but using  %+.10g and np.float64 as default would make things easier.

barsch:
> Adding a new CUSTOM keyword to SLIST/TSPAIR header is IMHO a breach of the format spec. I don't think we want to do that.

I disagree - I like ```custom_fmt``` to be used for people exporting in whatever format they like - in this case its not standard any more and that alone is a reason to not write any of the 3 types defined above - we should not allow reading this format anymore within ObsPy (its complicated anyway as we don't save the ```custom_fmt``` string anywhere) also we should not advertise it by documenting in the main ```write_ascii```  method - but we could leave it as it is as an additional option in ```_write_tspair``` - its already implemented, its useful and tested ...

obspy/io/ascii/core.py
```diff
@@ -419,6 +430,7 @@ def _write_tspair(stream, filename, **kwargs):  # @UnusedVariable
         2003-05-29T02:13:22.318400  2767
         ...
     """
+    custom_fmt = kwargs.get('custom_fmt', None)
```
barsch: ```custom_format``` is already a function argument and not part of ```kwargs```

megies:
I think it's strange that we should write something that we can't read afterwards.

But if you want to have it in, I don't care much, so go ahead.

heavelock:
To conclude:
- IMO, switching to `np.float64` as default is a reasonable choice.
- I think we should stick to `+.10e` as default for floats. The reason for `e` instead of `g` is  aesthetics of outputs and it's fixed form.
 I'm attaching screenshots of both types opened in a simple text editor. 
`+.10e` 
slist_e
testfile_e
`+.10g` 
slist_g
testfile_g
- I would rather add the `CUSTOM` to header when the `custom_fmt` is being used. It will allow us to keep read function simple. User who will manage to find `custom_fmt` option will also have manage consequence of using it. I've added additional information about this in docstrings.

What do you think?

obspy/io/ascii/core.py
```diff
@@ -325,17 +327,15 @@ def _write_slist(stream, filename, custom_fmt=None,  **kwargs):  # @UnusedVariab
...
-            elif trace.data.dtype.name.startswith('float64'):
-                dtype = 'FLOAT'
-                fmt = '%+.10g'
+                fmt = '%+.10e'
 
             else:
                 raise NotImplementedError
             # fmt
             if  custom_fmt is not None:
+                dtype = 'CUSTOM'
```
megies: I really don't care about this `CUSTOM` case a lot (I still don't see the point writing a file that can not be read afterwards), but I think if the custom format is something that relates to float (`%....f`, `%...g`, ...) than the type should also be `FLOAT`, so that it can be read again (same goes for custom integer types obviously).

Just my two cents. Leave as is if you like.

obspy/io/ascii/tests/data/tspair_float_custom_fmt.ascii
```diff
@@ -1,13 +1,13 @@
...
-2008-01-15T00:00:00.100000  +189.03999
-2008-01-15T00:00:00.125000  +194.05
-2008-01-15T00:00:00.150000  +205.06
-2008-01-15T00:00:00.175000  +209.07001
-2008-01-15T00:00:00.200000  +214.08
-2008-01-15T00:00:00.225000  +222.09
-2008-01-15T00:00:00.250000  +225.98
-2008-01-15T00:00:00.275000  +226.99001
-2008-01-15T00:00:00.300000  +219
+TIMESERIES XX_TEST__BHZ_R, 12 samples, 40 sps, 2008-01-15T00:00:00.025000, TSPAIR, CUSTOM, Counts
```
megies: See above comment, I think it's really strange to specify `CUSTOM`, when the string formatting clearly is of float type..

megies:
> What do you think?

I agree on the first two points, but I think it's really, really strange to specify `CUSTOM` when writing a float or integer string format.. anyway, I will never use this so, feel free to ignore these concerns.

obspy/io/ascii/core.py
barsch: @megies: this argument is invalid - you can provide any string format - such as ```%s```, ```%04d``` or ```%x``` - so ```FLOAT``` is not useful at all

barsch:
we have within ObsPy all kind of special functionality hidden in the privat reader/writers functions - and now suddenly for this useful functionality it is a big issue - its a special case for people to export in whatever they want - we don't promote it but its there if someone needs it and we can point to it if it occurs again - and as I said before its already implemented and tested

obspy/io/ascii/core.py
heavelock: What about creating a little method that's output will be FLOAT for `%..e f g` and INTEGER for `%..d i u`?

heavelock:
What about creating a method for determining whether the `custom_fmt` outputs `FLOAT`, `INTEGER` or `CUSTOM`? 
```

def _determine_dtype(custom_fmt):
    floats = ('e', 'E', 'f', 'F', 'g', 'G')
    ints = ('d', 'i')
    if any((c in floats) for c in custom_fmt):
        return 'FLOAT'
    elif any((c in ints) for c in custom_fmt):
        return 'INTEGER'
    else:
        return 'CUSTOM'

```

barsch:
sure works for me

megies:
Yep, that's what I had in mind, just a slight change: detect by last character only

```python
def _determine_dtype(custom_fmt):
    floats = ('e', 'f', 'g')
    ints = ('d', 'i')
    if custom_fmt[-1].lower() in floats:
        return 'FLOAT'
    elif custom_fmt[-1].lower() in ints:
        return 'INTEGER'
    else:
        return 'CUSTOM'
```

obspy/io/ascii/core.py
```diff
@@ -470,6 +470,24 @@ def _write_tspair(stream, filename, custom_fmt=None,
...
+    """
+    :type custom_fmt: str
+    :param custom_fmt: Python string formatter.
+    :rtype: str
+    :return: Datatype string for writing in header. Currently supported
+        are 'INTEGER', 'FLOAT' and `CUSTOM`.
+    """
+    floats = ('e', 'f', 'g')
+    ints = ('d', 'i')
+    if custom_fmt[-1].lower() in floats:
```
barsch: will raise exception for ```costum_fmt = ''``` - maybe have an ```try/except``` around it and raise a ```ValueError``` if it fails?

megies: Or you can replace it with 
```
if custom_fmt[-1:].lower() in floats:
```

in any case an empty format string will fail at some point anyway..

heavelock: Adding `try/except` clause gives a change to provide some more information to user.
But you are right, most of formatters that are left will raise an exception on file writing because of mismatch of np.dtype and string formatter.

megies:
That fail in appveyor seems unrelated.. looks like a change in numpy maybe and should affect master equally, I think

heavelock:
Great, thanks fro approval!

megies:
Thanks for the nice PR, @heavelock! And welcome to the dev team.  🚀

heavelock:
@megies Yay, thanks! :D

PR #1877, opened 2017-09-06
taup: added a time plot -- analogue to the existing ray path plot

### LINKED ISSUE #1330: Restore travelTimePlot() method
For convenience we should restore the old `travelTimePlot()` method. See #1276

### LINKED ISSUE #1501: travelTimePlot() function in obspy version 1.0.1 deprecated?
Using Obspy 1.0.1, installed through Anaconda 4.1.1 on Linux 64-bit. 
Trying to duplicate example 29.1 in the obspy tutorials: 
https://docs.obspy.org/tutorial/code_snippets/travel_time.html
One receives the warning: 
The travelTimePlot() function is deprecated. Please use the obspy.taup.TauPyModel class directly.

However, the obspy.taup.TauPyModel class does not seem to have a plot function in the same way, only a possibility to calculate travel times. For student teaching purposes, it would be great to have this function still. And it would also be nice to have an option for model PREM. 

Many thanks!

kaspervanwijk:
### What does this PR do?

create a method on the arrivals object for a travel time plot, analogue to the ray path plot method. It replaces the old (and removed `travelTimePlot` function, see https://docs.obspy.org/archive/1.0.3/tutorial/code_snippets/travel_time.html#travel-time-plot)
 
Fixes #1501, fixes #1330

### PR Checklist
- [x] All tests still pass.

### Todos
- [x] add convenience wrapper function
- [x] add example usage to convenience wrapper function docstring (with example plot)
- [x] add image test
- [x] add example plot to gallery

```python
import numpy as np

# set the velocity model you want to use:
model = TauPyModel(model='iasp91')

# set epicentral distances, the phases you want, and a source depth:
DISTANCES=range(0,100,1)
PHASES = ['S','P']
sourcedepth=5

plt.figure(figsize=(10, 10))
ax = plt.subplot(111)

# plot arrival times for all arrivals in arrivals object:
for distance in DISTANCES:
    arrivals = model.get_ray_paths(sourcedepth, distance, phase_list=PHASES)
    if not len(arrivals):
        print('No arrival time for:', PHASES, distance)
        continue
    arrivals.timeplot(PHASES,show=False,ax=ax)
ax.set_xlim(DISTANCES[0],DISTANCES[-1])
plt.show()
```


megies:
@kaspervanwijk Can you post an example image?

obspy/taup/tau.py
```diff
@@ -165,11 +165,55 @@ def copy(self):
...
+            will be created. Must be a polar axes for the spherical plot and
+            a regular one for the Cartesian plot.
+        :type ax: :class:`matplotlib.axes.Axes`
+        :param show: Show the plot.
+        :type show: bool
+
+        :returns: The (possibly created) axes instance.
+        :rtype: :class:`matplotlib.axes.Axes`
+
         """
```
megies: Please reformat docstring to (to be in line with rest of codebase):

```
"""
Method to plot the travel times of arrivals class, if any have
been calculated.

:param ax: Axes to plot to. If not given, a new figure with an axes
    will be created. Must be a polar axes for the spherical plot and
    a regular one for the Cartesian plot.
:type ax: :class:`matplotlib.axes.Axes`
:param show: Show the plot.
:type show: bool
:returns: The (possibly created) axes instance.
:rtype: :class:`matplotlib.axes.Axes`
"""
```

kaspervanwijk:
The output of the example test code with this is:
figure_1-9

megies:
Ah OK, so this is basically a replacement for the old plot that is not working in master anymore?

https://docs.obspy.org/tutorial/code_snippets/travel_time.html#travel-time-plot

kaspervanwijk:
That's right. A replacement (hopefully, this is my first PR), but the same way the ray paths are currently plotted: as a method on "arrivals."

megies:
Having this as a method on arrivals is very nice API-wise, but I think also a convenience wrapper function for the whole process would also be a real handy addition. Something like this..

```python
def traveltime_plot(source_depth, phases=['P', 'S'], model='iasp91',
                    distance_range=(0, 180)):
    ...
```

I'll add some todos in the OP.

megies:
For distances over 180 degrees, this could optionally be done in the plot doing a wraparound like here: in "Jeffreys/Bullen style" but it might be a bit tricky to get the duplicated x-axis labels right..

kaspervanwijk:
Could make the wrapper function you mention, but that would be very similar to the old way (except it would not call deprecated code, but properly use TauPyModel.get_travel_times(). Would you want both the method and the wrapper function?

As is, the plot handles >180 correctly, I believe.
 
(Yes, I will need some tips on how to do an image test and gallery add)

megies:
> Would you want both the method and the wrapper function?

Yes, I would implement both.

megies:
Requesting a docs build for this PR: +DOCS

obspy/taup/tau.py
```diff
@@ -663,3 +713,91 @@ def create_taup_model(model_name, output_dir, input_dir):
     else:
         model_file_name = model_name + ".tvel"
     TauPCreate.main(model_file_name, output_dir, input_dir)
+
+
+def traveltime_plot(min_degree=0, max_degree=360, npoints=1000,
+                    phases=['P', 'S'], source_depth=10, model='iasp91',
```
megies: I would vote for making `source_depth` an argument, not a `kwarg` with default value, such that people have to explicitly set their source depth of choice. Otherwise people might overlook it's importance for the output and I feel like there's no generally useful default value for it.

obspy/taup/tau.py
```diff
@@ -663,3 +713,91 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+                    ax=None, show=True):
+        """
+        Method to plot the travel times of arrivals class, if any have
+        been calculated.
+
+        :param min_degree: minimum distance from the source (in degrees) to 
+            plot travel times Defaults to ``0``.
+        :type min_degree: float, optional
```
megies: I think we nowadays usually leave away the `, optional` part (I know we still have it in older parts of the docs).

It's just plain redundant to add it for just every `kwarg` out there and only clutters the docs, IMHO.

obspy/taup/tau.py
```diff
@@ -663,3 +713,91 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        :type max_degree: float, optional
+        :param npoints: Number of points to plot. Defaults to ``1000``.
+        :type npoints: int, optional
+        :param phases: List of phase names which should be used within the 
+            plot. Defaults to ``['P', 'S']``.
+        :type phases: list of str, optional
+        :param source_depth: Source pepth in kilometers. Defaults to ``10``.
+        :type source_depth: float, optional
+        :param model: string containing the model to use.
+        :type model: str. Defaults to 'iasp91'
```
megies: The default value should be mentioned in the `:param ..:` section below, I think, rather than the `:type :` section

obspy/taup/tau.py
```diff
@@ -663,3 +713,91 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        :type ax: :class:`matplotlib.axes.Axes`
+        :param show: Show the plot.
+        :type show: bool
+
+        :returns: The (possibly created) axes instance.
+        :rtype: :class:`matplotlib.axes.Axes`
+
+        .. rubric:: Example
+
+        >>> from obspy.taup.tau import traveltime_plot
```
megies: This plot will be very useful, I'd vote to add it's import to `obspy/taup/__init__.py` so that it can be imported as `from obspy.taup import traveltime_plot`. Accordingly then all examples should also use that shorter import notation.

megies:
This is looking real nice already. :-)

To add it to the gallery, just replace the old example that's still in there (broken of course). I.e. adapt `obspy/misc/docs/source/tutorial/code_snippets/travel_time_plot.py` and `obspy/misc/docs/source/tutorial/code_snippets/travel_time_plot.rst` and also adapt it in `obspy/misc/docs/source/gallery.rst` (line 141).

For the image test, just copy/paste/adapt one of the test routines in `obspy/obspy/taup/tests/test_plotting.py`. You can then run `$ obspy-runtests taup --keep-images` (or alternatively `$ OBSPY_KEEP_IMAGES= python taup/tests/test_plotting.py` if you want to omit the other taup tests) and move the newly created image in `obspy/obspy/taup/tests/images/testrun` up one directory to drop it in as the base line image. If I'm not mistaken, base line images are still based on matplotlib 1.3.1.. (@krischer, @QuLogic, that right?), so that's the version you should use when creating the base line image.

megies:
If you want to see what flake8 thinks of your commits right when doing then, you could set up a local git hook, see first paragraph on this page: https://docs.obspy.org/master/coding_style.html#obspy-coding-style-guide

It's really useful.

megies:
Docs build for the PR is up (will be automatically updated on new commits): http://docs.obspy.org/pull-requests/1877/packages/obspy.taup.html#module-obspy.taup

Example plot's not working..: http://docs.obspy.org/pull-requests/1877/packages/autogen/obspy.taup.tau.traveltime_plot.html#obspy.taup.tau.traveltime_plot

See http://docs.obspy.org/pull-requests/1877/log.txt, searching for `traveltime_plot`:
```
/private-backup/home/obspy/update-docs-pr/src/obspy/obspy/taup/tau.py:docstring of obspy.taup.tau.traveltime_plot:36: WARNING: Exception occurred in plotting obspy-taup-tau-traveltime_plot-1
 from /private-backup/home/obspy/update-docs-pr/src/obspy/misc/docs/source/packages/autogen/obspy.taup.tau.traveltime_plot.rst:
  File "/home/obspy/anaconda3/envs/tmp_build_docs-pr/lib/python3.4/site-packages/matplotlib/sphinxext/plot_directive.py", line 517, in run_code
    six.exec_(code, ns)
  File "<string>", line 2, in <module>
NameError: name 'travetime_plot' is not defined
```

Can you also add a short paragraph with plot to the taup landing page (i.e. `taup/__init__.py` docstring)?

obspy/taup/tau.py
```diff
@@ -716,30 +716,29 @@ def create_taup_model(model_name, output_dir, input_dir):
 
 
 def traveltime_plot(min_degree=0, max_degree=360, npoints=1000,
-                    phases=['P', 'S'], source_depth=10, model='iasp91',
+                    phases=['P', 'S'], source_depth, model='iasp91',
```
megies: you'll have to move `source_depth` to the front..

kaspervanwijk:
Running into a problem with creating the image, using the timeplot() method. Just like the plot() method, there is an optional creating of an ax, but in the case of timeplot(), the test in not expecting an ax:

[kvan637@sc-phy-324060 tests]$ obspy-runtests taup --keep-images
Running /home/kvan637/obspy/obspy/scripts/runtests.py, ObsPy version '1.0.3.post0+1502.gca18fa85fc.dirty.kaspervanwijk.timeplot'
.F..s..................E..................s.......s..s.....

ERROR: test_travel_time_plot (obspy.taup.tests.test_plotting.TauPyPlottingTestCase)
  File "/home/kvan637/obspy/obspy/taup/tests/test_plotting.py", line 124, in test_travel_time_plot
    plt.savefig(ic.name)
  File "/home/kvan637/obspy/obspy/core/util/testing.py", line 423, in __exit__
    msg = self.compare()
  File "/home/kvan637/obspy/obspy/core/util/testing.py", line 523, in compare
    return compare_images(self.baseline_image, self.name, tol=self.tol)
  File "/home/kvan637/obspy/obspy/core/util/testing.py", line 195, in compare_images
    raise IOError('Baseline image %r does not exist.' % expected)
OSError: Baseline image '/home/kvan637/obspy/obspy/taup/tests/images/traveltimeplot.png' does not exist.

FAIL: traveltime_plot (obspy.taup.tau)
Doctest: obspy.taup.tau.traveltime_plot
  File "/home/kvan637/anaconda3/lib/python3.5/doctest.py", line 2190, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for obspy.taup.tau.traveltime_plot
  File "/home/kvan637/obspy/obspy/taup/tau.py", line 720, in traveltime_plot

File "/home/kvan637/obspy/obspy/taup/tau.py", line 752, in obspy.taup.tau.traveltime_plot
Failed example:
    traveltime_plot(source_depth=120,min_degree=0, max_degree=50,
             phases=['P', 'S', 'PP'], model='iasp91', npoints=1000)
Expected nothing
Got:
    <matplotlib.axes._subplots.AxesSubplot object at 0x7fe5fc0c25f8>


Ran 59 tests in 99.274s

megies:
Oh, I think you can just copy any other base line image to the expected base line image location.. we should eventually implement some canonical way to create an initial new base line image..

kaspervanwijk:
Ok, I hope I got those changes in correctly... thanks for the flake8 tip!

potential issues:
- I failed to downgrade to matplotlib 1.3.1 for the image. If it is crucial, I will try harder than 
conda install matplotlib=1.3.1 (which clashed with other things I have)
- My baseline image is made by hand, and not from the test plotting
- gallery.rst seemed ok already. I was not seeing any changes to be made

obspy/taup/tau.py
```diff
@@ -165,6 +165,56 @@ def copy(self):
...
+        import matplotlib.pyplot as plt
+        cmap = plt.cm.Dark2
+        colors = cmap(np.linspace(0, 1, len(phases)))
+        if not self:
+            raise ValueError("Can only plot calculated travel times.")
+
+        # create an axis/figure, if there is none, yet:
+        if not ax:
+            plt.figure(figsize=(10, 10))
+            ax = plt.subplot(111)
```
megies: Ideally, this should all be done with clean API, not relying on current figure / current axes. I.e...

```python
if ax:
    fig = ax.figure
else:
    fig, ax = plt.subplots()
```

And then in the following all plot commands on `ax` not using pylab (current axes), so avoiding things like `plt.plot(..)` and using `ax.plot()` instead.

obspy/taup/tau.py
```diff
@@ -165,6 +165,56 @@ def copy(self):
...
+
+        # extract the time/distance for each phase, and for each receiver:
+        for arrival in self:
+            phase = arrival.name
+            if phase in phases:
+                plt.plot(arrival.distance, arrival.time / 60, '.',
```
megies: ax.plot()

obspy/taup/tau.py
```diff
@@ -165,6 +165,56 @@ def copy(self):
...
+                         label=phase, color=colors[phases.index(phase)])
+
+        # merge all arrival labels of a certain phase:
+        handles, labels = plt.gca().get_legend_handles_labels()
```
megies: ax.get_legend_...()

obspy/taup/tau.py
```diff
@@ -165,6 +165,56 @@ def copy(self):
...
+        labels, ids = np.unique(labels, return_index=True)
+        handles = [handles[i] for i in ids]
+
+        plt.legend(handles, labels, loc=2, numpoints=1)
+
+        plt.grid()
+        plt.xlabel('Distance (degrees)')
+        plt.ylabel('Time (minutes)')
```
megies: ax.legend(), ax.grid(), ax.set_xlabel(), ...

obspy/taup/tau.py
```diff
@@ -663,3 +713,88 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        :type show: bool
+
+        :returns: The (possibly created) axes instance.
+        :rtype: :class:`matplotlib.axes.Axes`
+
+        .. rubric:: Example
+
+        >>> from obspy.taup import traveltime_plot
+        >>> traveltime_plot(source_depth=120, min_degree=0, max_degree=50,
+        ...          phases=['P', 'S', 'PP'], model='iasp91', npoints=1000)
```
megies: Need to add the return value to the doctest, see https://ci.appveyor.com/project/obspy/obspy/build/1.0.5123-master/job/gjp7dcd9yjp7vq5u#L719:

```

        >>> traveltime_plot(source_depth=120, min_degree=0, max_degree=50,
        ...          phases=['P', 'S', 'PP'], model='iasp91',
        ...          npoints=1000)  # doctest: +ELLIPSIS
        <matplotlib.axes._subplots.AxesSubplot object at 0x...>

**Comment by megies** [comment]
> I failed to downgrade to matplotlib 1.3.1 for the image. If it is crucial, I will try harder than
> conda install matplotlib=1.3.1 (which clashed with other things I have)

Usually it's easiest to create a new conda environment specifically for this purpose.

> My baseline image is made by hand, and not from the test plotting

Hmm.. that can and likely will make a difference, as we set some matplotlib defaults in our test runner

> gallery.rst seemed ok already. I was not seeing any changes to be made

Ah, right, since it's just linking to that code snippet file in the tutorial that you're reusing, the gallery entry is fine already.

File: obspy/taup/tau.py (hunk 10 above)
Comment (kaspervanwijk) [reply obspy/taup/tau.py]: On it. How about existing instances in tau.py? For example, line 341. Do you want me to find and replace all of them?

File: obspy/taup/tau.py (hunk 11 above)
Comment (kaspervanwijk) [reply obspy/taup/tau.py]: The return value *is* in the doc string, but should it be at the very end of the string? There was an example after the return value. I am changing the order, but is that enough?

File: obspy/taup/tau.py (hunk 10 above)
Comment (megies) [reply obspy/taup/tau.py]: You don't have to touch stuff outside of the PR if you don't want to. But it looks like all lines that contain `plt.`-calls are covered (btw. [codecov github browser extensions](https://docs.codecov.io/docs/browser-extension) are really useful), so go ahead if you also want to fix those.

File: obspy/taup/tau.py (hunk 11 above)
Comment (megies) [reply obspy/taup/tau.py]: > The return value is in the doc string

Sorry, I don't understand what you're saying. `traveltime_plot` is called without assigning it's return value and so the interactive prompt will print it's representation after the call and the result of this interactive print is missing in your doctest example. Just change it like I proposed and it should work.

> that's the same error I got when creating the test image!

That's strange, because in the unit test no return values are automatically printed/asserted, only explicit unit test assertions. But of course, if you run `$ obspy-runtests taup --keep-images` then doctests are executed as well (as opposed to running `$ OBSPY_KEEP_IMAGES= python taup/tests/test_plotting.py` which really only runs the unit tests in that file).

File: misc/docs/source/tutorial/code_snippets/ray_path_plot.py [hunk 12]
```diff
@@ -0,0 +1,7 @@
+from obspy.taup.tau import raypath_plot
+import matplotlib.pyplot as plt
+
+ax = plt.subplot(111, polar=True)
+fig = ax.figure
```
megies: just nitpicking.. but matplotlib is nowadays advertising this for most fig/ax setup steps:

```python
fig, ax = plt.subplots(polar=True)
```

See e.g. https://matplotlib.org/examples/pylab_examples/subplots_demo.html

kaspervanwijk: Sure! Will change this (also in tau.py)

krischer:
> If I'm not mistaken, base line images are still based on matplotlib 1.3.1.. (@krischer, @QuLogic, that right?), so that's the version you should use when creating the base line image.

No they should actually be create with the latest `matplotlib 2.0.x`. So that should actually be quite a bit easier to install.

kaspervanwijk:
Should I redo the figs then? I ended up making these with an environment
with matplotlib1.3.1...

Kasper van Wijk
Department of physics
University of Auckland
https://www.physics.auckland.ac.nz/people/kvan637

QuLogic:
It doesn't look like tests are passing, so that's probably what needs to be done. If you can, please amend a previous commit or rebase so that only the latest copy of these (large-ish) images is added.

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
     else:
         model_file_name = model_name + ".tvel"
     TauPCreate.main(model_file_name, output_dir, input_dir)
+
+
+def traveltime_plot(source_depth, min_degree=10, max_degree=90, nrecs=1000,
```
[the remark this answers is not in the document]
krischer: Also any particular reason to go from 10-90? I feel like it should go the full distance by default.

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        :type min_degree: float
+        :param max_degree: maximum distance from the source (in degrees) to
+            plot travel times. Defaults to ``90``.
+        :type max_degree: float
+        :param nrecs: Number of points to plot. Defaults to ``1000``.
+        :type nrecs: int
+        :param phases: List of phase names which should be used within the
+            plot. Defaults to ``['P', 'PKP']``.
+        :type phases: list of str, optional
+        :param model: string containing the model to use. Defaults to 'iasp91'.
```
krischer: \``"iasp"\`` to get it to render correctly. Also maybe just document it as "Model name."

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+
+        >>> from obspy.taup import traveltime_plot
+        >>> import matplotlib.pyplot as plt
+        >>> fig, ax = plt.subplots()
+        >>> ax = traveltime_plot(source_depth=10, ax=ax, fig=fig)
+        >>> plt.show()
+
+        .. plot::
+
+        from obspy.taup import traveltime_plot
```
krischer: I think this has to intended to correctly render.

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        import matplotlib.pyplot as plt
+
+        fig, ax = plt.subplots()
+
+        ax = traveltime_plot(source_depth=10, ax=ax, fig=fig)
+
+        plt.show()
+
```
krischer: superfluous space

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+        """
+
```
krischer: superfluous space

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+                print(err)
+            else:
+                plotted = True
+
+        if not plotted:
+            raise ValueError("No travel times to plot")
+        return ax
+
+
+def raypath_plot(source_depth, min_degree=10, max_degree=90, nrecs=20,
```
krischer: Same comments as before. I think `plot_raypaths()` is a better name and the arguments might benefit from different names.

obspy/taup/tau.py
```diff
@@ -663,3 +728,174 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+
+        # calculate the arrival times and plot vs. epicentral distance:
+        degrees = np.linspace(min_degree, max_degree, nrecs)
+        for degree in degrees:
+            try:
+                arrivals = model.get_ray_paths(source_depth, degree,
+                                               phase_list=phases)
+                arrivals.plot(show=False, ax=ax, plot_type=plot_type,
+                              plot_all=plot_all,
+                              legend=legend)
```
krischer: This might also have the problem of phases changing colors - I think there should be way to resolve this although it might not be trivial here.

obspy/taup/tests/test_plotting.py
```diff
@@ -24,39 +24,44 @@ def setUp(self):
         self.image_dir = os.path.join(os.path.dirname(__file__), 'images')
         self.model = TauPyModel(model="iasp91")
 
-    def test_spherical_many_phases(self):
+    def test_spherical_all_phases(self):
```
krischer: TauP in this implementation has no concept of "all" phases - phase names are dynamically evaluated so "many" is IMHO not a bad name for the default set.

obspy/taup/tests/test_plotting.py
```diff
@@ -24,39 +24,44 @@ def setUp(self):
...
+        """
+        Spherical plot of the ray paths for all phases for a single
+        epicentral distance, but both ways around the globe.
+
+        """
```
krischer: Please remove the empty line.

obspy/taup/__init__.py
```diff
@@ -185,6 +184,63 @@
...
+>>> arrivals = model.get_ray_paths(source_depth_in_km=500,
+...                                distance_in_degree=130)
+>>> arrivals.timeplot()  # doctest: +SKIP
+
+.. plot::
+    :width: 50%
+    :align: center
+
+    from obspy.taup import TauPyModel
+    TauPyModel().get_ray_paths(500, 130).timeplot()
```
krischer: Could you use only a couple of phases here? Otherwise the plot's legend just get really big and the plot not as pretty as it could be.

obspy/taup/__init__.py
```diff
@@ -185,6 +184,63 @@
...
+
+Alternatively, convenience wrapper functions are in place to plot the
+travel times and the ray paths.
```
krischer: Maybe replace with "... to plot travel times (maybe uses arrival times instead?) and ray paths across many distances."

kaspervanwijk:
I will work my way through the stylistic changes, no problem, but I feel
the method for both ray paths (already there) and travel times (new) is the
way to go. This way, both rays and times work identically:

- the methods plot ray paths, or arrival times, for a single epicentral
distance (for as many phases as one defines)
- the wrapper functions plot ray paths or arrival times for as many
epicentral distances (calling the method each time) as one wants.

(to me, the only issue with the travel time plot method is that the output
is not very informative: dots at one epicentral distance)

I think the methods are the key part of the code, and the wrapper functions
are just suggestions on usage. Anyone can do their own, really.

kaspervanwijk:
There is only a small problem in the legend, as far as I am aware. Bunching
all labels with the same phase name works, EXCEPT when we have multiple
arrivals of the same phase at the same epicentral distance. I am not sure
why.

Bunching is currently done in the method, and not in the function, which is
a mistake. It will be fixed in the next version I will commit.

I think I should do the same bunching in the ray path function. Try turning
on the legend in the current version of the wrapper function. Chaos! Every
instance of a phase for every distance will show in the legend!

So if you guys like the way the legend is done in the time plot, I will add
the same trick to the ray path plot:

# merge all arrival labels of a certain phase:
handles, labels = ax.get_legend_handles_labels()
labels, ids = np.unique(labels, return_index=True)
handles = [handles[i] for i in ids]
ax.legend(handles, labels, loc=2, numpoints=1)

krischer:
Yes - this might solve that part of the problem (if the legend is only called for the last distance - I'm not sure how matplotlib behaves when calling `.legend()` multiple times but maybe it just work.

There is the other (and harder to solve) problem of how colors are assigned - the first phase gets the first color, the second one the second, and so on. This means that different phases can have the same color when changing distances. This has to be resolved somehow I feel. Or did I overlook something and this is already dealt with?

kaspervanwijk:
I see your point. But this problem already exists for the ray paths plot!
Each call to the plot() method may color different phases with different
colors (depending on what phases are in the arrivals object). What if I
assign colors based on the index of the phases argument? I think that would
do the trick.

Kasper van Wijk
Department of physics
University of Auckland
https://www.physics.auckland.ac.nz/people/kvan637

kaspervanwijk:
OK, almost have the legend thing sorted. One question: is there a good trick to get the  list of phase names that make up phase_list="ttall"? (I see versions of it in utils.py.)

krischer:
Yea - `utils.parse_phase_list()` does pretty much that.

I don't think this will work in general though. I think it would be a totally valid use case (and an actual use for having it as a separate method) to plot P waves for a certain distance range and then switch to another wave type - in that case the `phase_list` argument will change.

One possibility is to just store the color <-> phase_name assignment as an attribute on the `Axis` object. A bit ugly but there is no downside that I can immediately think of. And it might fully solve the issue.

megies:
> I don't think the figures need to be regenerated - this adds a fairly significant burden to the repository (especially considering we just regenerated most figures to support matplotlib 2) as any change will forever be in ObsPy's history. I think the main reason why you regenerated is that the default figure size is changed?

If image updates are kept in single commits with no code changes, then squashing them before merging is pretty simple.

krischer:
> If image updates are kept in single commits with no code changes, then squashing them before merging (to keep history size down) is pretty simple.

True - but in this particular case I think its just not necessary and worth the trade-off.

obspy/taup/tau.py
```diff
@@ -165,10 +165,57 @@ def copy(self):
         return self.__class__(super(Arrivals, self).copy(),
                               model=self.model)
 
+    def timeplot(self, show=True, fig=None, ax=None):
```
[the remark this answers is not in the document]
kaspervanwijk: I agree it should be plot_times(), but can I change the existing plot() to plot_rays() then? I like the consistency in that, but worry about compatibility with people's existing scripts.

obspy/taup/tau.py
```diff
@@ -165,10 +165,57 @@ def copy(self):
...
+            will be created.
+        :type ax: :class:`matplotlib.axes.Axes`
+
+        :returns: The (possibly created) axes instance.
+        :rtype: :class:`matplotlib.axes.Axes`
+
+        """
+        import matplotlib.pyplot as plt
+
+        if not self:
```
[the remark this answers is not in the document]
kaspervanwijk: this I copied from the existing plot() method for the rays. Happy to change both or neither. Let me know.

obspy/taup/tau.py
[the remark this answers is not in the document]
kaspervanwijk: happy to make it min_degrees. how about existing distance_in_degree? distance_in_degrees? Or can we just make it distance and explain in docs it is in degrees? Same question about  sourcedepth_in_km, receiver_depth_km. Personally, I would drop the units in all names.

megies:
@kaspervanwijk I'm really sorry, that was wrong information about the mpl version to use for baseline images.. so please use mpl 2.0 for them (see https://github.com/obspy/obspy/pull/1616/commits/9602597a52bf9e05c7c634fb52746f3e352f48c2 and #1616)

kaspervanwijk:
Agreed with all the suggested changes by @megies and @krischer, and hope I got them all in there now.

megies:
Seems the color cycle was different when you did the baseline images (links from appveyor run):
```
	Expected:  https://i.imgur.com/X8pTfLP.png
	Actual:    https://i.imgur.com/mDAkvZf.png
	Diff:      https://i.imgur.com/sgWUcrx.png
```
Did you generate them via `obspy-runtests`?

kaspervanwijk:
I do get these from the obspy-runtests, but there is something wrong with my code; two runs of the tests in a row do not lead to the same color scheme; I don't get it!

megies:
Image test fails seem to be only minimal differences now.. what mpl are you using for base line images?

kaspervanwijk:
2.0.0

On 23 Sep 2017 12:39 am, "Tobias Megies" <notifications@github.com> wrote:

Image test fails seem to be only minimal differences now.. what mpl are you
using for base line images?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/obspy/obspy/pull/1877#issuecomment-331435191>, or mute
the thread

krischer:
You'll have to use at least 2.0.1 if I remember correctly.

megies:
@kaspervanwijk, you're using `tab20` and it's not in mpl < 2.0.1 it seems, according to Travis runs. You'll need to work around this fact, same as we do for viridis. You'll have to copy the color dict for `tab20` from [mpl source code](https://github.com/matplotlib/matplotlib/blob/95d58f46f22d10649eb87de4a3570055911ea46b/lib/matplotlib/_cm.py#L1299-L1320) and paste it in `obspy/imaging/cm.py` and set up the color map with `from matplotlib.colors import ListedColormap; tab20 = ListedColormap(_tab20_data)` and then use it accordingly in your plotting code.. `from obspy.imaging.cm import tab20`.

megies:
I'm not sure what's going on, but the docs build is garbled: http://docs.obspy.org/pull-requests/1877/
 - link to ’obspy.taup’ is broken: http://docs.obspy.org/pull-requests/1877/packages/index.html
 - and no taup plots show up in gallery: http://docs.obspy.org/pull-requests/1877/gallery.html

Also.. your workaround to switch back to "Paired" doesn't work for ancient matplotlib, unfortunately, as it returns a ’LinearSegmentedColormap’ instead: https://travis-ci.org/obspy/obspy/jobs/279784993#L1130

It's easiest to just hard code those values and create a ’ListedColorMap’ like I recommended, I think..

kaspervanwijk:
guys, I was stubborn on the colors and defined them now the way the master tau.py had them, but I now realize that may be foolish. Let me know, and I will actually do as @megies suggested by embedding an explicit  colormap.

megies:
It seems to work across versions, so OK for me.

megies:
There's still some broken paths in the docs build, looks like, please check all the changed paged..

E.g. http://docs.obspy.org/pull-requests/1877/tutorial/code_snippets/ray_path.html#ray-path-plot

kaspervanwijk:
Ok, I got those (I think), but are there more? You mention multiple problems with the docs, but the doc test passed.

megies:
> Ok, I got those (I think), but are there more? You mention multiple problems with the docs, but the doc test passed.

Unfortunately, the doc test passing only means that *some* output was generated, it's not saying that there are no "minor" warnings/errors within the docs build (e.g. images that cant be created, broken links etc.).

It would be really good to at some point get the zillions of warnings in the docs build fixed and add a test that no new problems get introduced by new PRS.. but i doubt that somebody will volunteer for work like this, given the fact that we even continuously fail to get the release done.

If you want a glimpse at the PR build log, it's here: http://docs.obspy.org/pull-requests/1877/log.txt

misc/docs/source/tutorial/code_snippets/travel_time.rst
```diff
@@ -1,24 +1,23 @@
 ========================
-Travel Time Calculations
+Travel Time and Ray Path Plotting
 ========================
```
megies: headline indicators should have same length as headline

obspy/taup/__init__.py
```diff
@@ -140,50 +140,106 @@
 --------
 
 If ray paths have been calculated, they can be plotted using the
-:meth:`~obspy.taup.tau.Arrivals.plot` method:
+:meth:`~obspy.taup.tau.Arrivals.plot_rays` method:
 
 >>> arrivals = model.get_ray_paths(source_depth_in_km=500,
 ...                                distance_in_degree=130)
->>> arrivals.plot()  # doctest: +SKIP
+>>> arrivals.plot_rays()  # doctest: +SKIP
```
megies: Do these plots take a lot of time? In principle we should skip as few commands as possible.. any big downsides of having these doctests (here and below) executed?

obspy/taup/__init__.py
```diff
@@ -140,50 +140,106 @@
...
+Alternatively, convenience wrapper functions plot the arrival times
+and the ray paths for a range of epicentral distances.
+
+The travel times wrapper function is:
+
+>>> from obspy.taup import plot_travel_times
+>>> import matplotlib.pyplot as plt
+>>> fig, ax = plt.subplots()
+>>> ax = plot_travel_times(source_depth=10, phase_list=["P","S","PP"],
+... ax=ax, fig=fig)
```
megies: this should be indented to match indentation of `source_depth=...`

obspy/taup/__init__.py
```diff
@@ -140,50 +140,106 @@
...
+    ... fig=fig)
+
+The ray path plot wrapper function is:
+
+>>> from obspy.taup import plot_ray_paths
+>>> import matplotlib.pyplot as plt
+>>> ax = plt.subplot(111, polar=True)
+>>> fig = ax.figure
+>>> ax = plot_ray_paths(source_depth=100, ax=ax, fig=fig)
+There were rays for all but the following epicentral distances:
```
megies: See above, please add an empty line below the code block, I think that's nicer to look at in an editor.

obspy/taup/__init__.py
```diff
@@ -382,8 +438,9 @@
 
 
 # Convenience imports.
-from .tau import TauPyModel  # NOQA
-
+from .tau import TauPyModel # NOQA
+from .tau import plot_travel_times # NOQA
+from .tau import plot_ray_paths # NOQA
```
megies: not sure why flake8 is not complaining, but according to PEP8, there should be two spaces before the inline comment hash sign (compare the deleted line above) (https://www.python.org/dev/peps/pep-0008/#inline-comments)

paul-freeman:
> It would be really good to at some point get the zillions of warnings in the docs build fixed and add a test that no new problems get introduced by new PRS.. but **i doubt that somebody will volunteer for work like this**, given the fact that we even continuously fail to get the release done.

@megies I have a bit of Sphinx experience and would be happy to take a crack at this in a new branch. I'd probably learn a tremendous amount in the process. Even if I only get the low hanging fruit, it'll could still be useful.

krischer:
That would be really great! Don't bother with backwards compatibility too much - it only really has to work for the latest Sphinx.

The docs currently also take a long time to build. A part of that is due to the arclink doc downloading lots of things which could be improved. I also don't really understand why it rebuilds things quite so often when one runs it a number of times with small changes in between.

obspy/taup/__init__.py
kaspervanwijk: I didn't think that was possible; the doc string needs to match the output of the code block. An empty line would throw a wrench into that idea. But I must be missing something.

kaspervanwijk:
Please note the gcc errors  in travis-ci for all but one version of python (which is, curiously, not the version I successfully build this branch with locally; 3.5.4)

megies:
@paul-freeman here's the current conda env used on our docs buildbot, as I know it can be tricky to get correct package versions for building docs: https://gist.github.com/f3f639f575e0ca97fdfd9b41348da802

obspy/taup/__init__.py
megies: Blank lines that are really in the output are specified as `<BLANKLINE>` in doctests. So adding a blank line for nicer look is totally safe. See http://www.sphinx-doc.org/en/stable/ext/doctest.html?highlight=doctest#directive-doctest

obspy/taup/__init__.py
```diff
@@ -140,50 +140,109 @@
...
+    fig, ax = plt.subplots()
+    ax = plot_travel_times(source_depth=10, ax=ax, phase_list=["P","S","PP"],
+    ...                    fig=fig)
+
+The ray path plot wrapper function is:
+
+>>> from obspy.taup import plot_ray_paths
+>>> import matplotlib.pyplot as plt
+>>> ax = plt.subplot(111, polar=True)
+>>> fig = ax.figure
```
megies: please compress this as `fig, ax = plt.subplots(...)` like above and to be consistent across all examples

obspy/taup/__init__.py
```diff
@@ -140,50 +140,109 @@
...
+
+.. plot::
+    :width: 50%
+    :align: center
+
+    from obspy.taup import plot_ray_paths
+    import matplotlib.pyplot as plt
+
+    ax = plt.subplot(111, polar=True)
+    fig = ax.figure
```
megies: please compress this as `fig, ax = plt.subplots(...)` like above and to be consistent across all examples

megies:
This looks good to me now. Only thing that's changed is that we have more paths in this one Tutorial plot (maybe due to this change: https://github.com/obspy/obspy/pull/1877/files#diff-4343e81d5f94dd2b0905e41df4cae7c2L54):

(compare e.g. rays arriving from both sides at `SSS` arrival near south pole)

Also arrival labels seem to need more space, but I guess that's just due to the [reduced figure size ](https://github.com/obspy/obspy/pull/1877/files#diff-4343e81d5f94dd2b0905e41df4cae7c2R32) in the tutorial example (I guess you reduced size to make plots align nicer in the bootstrap gallery mosaic?)?

### This PR


### master



---

One more thing: The logic to use same color across a single phase now causes this plot to use the same color for both ray paths: https://github.com/kaspervanwijk/obspy/blob/ddbbc459eb40651d234836cf776591062b78bbfc/obspy/taup/tests/images/spherical_more_then_360.png
For plots with just a few rays of same phase but different distance it might still be nice to have the possibility to color each ray differently?

obspy/taup/__init__.py
kaspervanwijk: OK, consider that done.

obspy/taup/__init__.py
kaspervanwijk: done

kaspervanwijk:
I think the color options are not ideal, but hopefully acceptable. As is the size of the pot (I wanted to keep the default figure size to match most of the gallery).

These last comments did point out a bug in the current code, though, which I don't know how to fix.

```
import numpy as np

phase_list = ['SKKS']
model = TauPyModel()
arrivals = model.get_ray_paths(500, -120,
                               receiver_depth_in_km=200,phase_list=phase_list)
arrivals.plot(plot_type="spherical", plot_all=True,
                                `show=True,label_arrivals=True)`
```
results in:
figure_1-2

This goes wrong here in tau.py:

for arrival in self:
            if arrival.path is None:
                continue
            dist = arrival.purist_distance % 360.0
            distance = arrival.distance
            if abs(dist - distance) / dist > 1E-5:
                if plot_all is False:
                    continue
                # Mirror on axis.
                arrival = copy.deepcopy(arrival)
                arrival.path["dist"] *= -1.0
            arrivals.append(arrival)

but I cannot follow the logic. If you call the above code with a positive 120 degrees, all is well. We want to "mirror" that behavior for -120.

Can the author of tau.py in the master chime in?

kaspervanwijk:
the new version of the plot above is much cleaner, and only a fraction of the coding. I also found a bug in tau.py for negative epicentral distances, as a bonus.  I added some color to the plot in the latest version, as requested.

Rays of the same phase and same distance, just a different path, will have the same color. Rays of the same phase different distance, or same distance, different phase, **will** have different colors (if in ttall; if not, they will be black unless a phase_list is explicitly defined) 

I should probably increase the distance between label and symbol: I wanted to keep the plot the standard figure() size.

megies:
@kaspervanwijk are you adding more changes to this PR? I think I can take over for last changes and rebasing and then it can be merged..

kaspervanwijk:
@megies: if you are happy with the PR, I am happy. Thanks for letting me contribute. I learned heaps!

obspy/taup/tau.py
```diff
@@ -663,3 +833,199 @@ def create_taup_model(model_name, output_dir, input_dir):
...
+
+        if show:
+            plt.show()
+        return ax
+
+
+def plot_ray_paths(source_depth, min_degrees=0, max_degrees=360, npoints=10,
+                   plot_type='spherical', phase_list=['P', 'S', 'PP'],
+                   model='iasp91', plot_all=True, legend=False,
+                   label_arrivals=False, fig=None, show=True, ax=None):
```
megies: @kaspervanwijk I think there's a caveat about `plot_type`/`ax` that is not covered yet. What's your take on the following point?

 - if `ax` is specified, then depending on what kinds of plots have been done before this could get a garbled mixed plot with both spherical and cartesian plots overlain? I think we should raise an `ValueError` if `plot_type='spherical'` and provided axes (if any) `isinstance(ax, matplotlib.projections.polar.PolarAxes)` (and vice versa for polar plot)?


Just wanted to hear if this sounds about right, I'm already at some minor changes to finalize this PR and can do it.

kaspervanwijk: Not sure if I follow. Do you mean that if someone calls this helper function in a script, and had already plotted something (maybe in cartesian coordinates), before a plt.show() and then this would send a polar plot to that instance?

megies: Yep, something like that..

megies: In any case, I do my final change (push them maybe in an hour or so) and then you can have a look and comment on them later.

megies:
Wow.. this takes a really long time on my machine:

```python
from obspy.taup import plot_travel_times
plot_travel_times(source_depth=10)
```

I think we should lower the default number of points.. maybe to 50, that finishes in a reasonable amount of time for me. People can always crank it up if they need more.

obspy/taup/tau.py
megies: Ah.. I see, that wasn't in your changes, really.. it's just in the diff because it was moved over from the old `plot` routine to `plot_rays`..

megies:
@kaspervanwijk I just pushed some changes, would be good if you breifly check them. I'll start rebasing intermediate images out of this now and will then force-push over this branch. For reference, current tip is at 0a0b8acf72bbdda8ef0130173baf0c0b1a89beee.

kaspervanwijk:
thank you, @megies, your edits are all improvements. I agree with making the default number of travel times to calculate a smaller number to speed up the code, but hope we can keep a nice dense graph in the gallery.

megies:
> hope we can keep a nice dense graph in the gallery.

fair enough, i'll crank it up

megies:
Thanks a bunch @kaspervanwijk, great work on this PR!

megies:
see 9bd3a80ea96e1209a1d7de88b9b343fc1ff1305d

PR #1873, opened 2017-09-12
SC3ML event read support for all schema versions

megies:
Read support for SC3ML event data was added in #1848 by @Brtle. However it restricts read support to schema version 0.9 currently. Since stylesheets provided by upstream maintainers (http://geofon.gfz-potsdam.de/schema/) are used for the conversion to QuakeML, it is trivial to support all schema versions, [like also done in the SC3ML inventory read routine](https://github.com/obspy/obspy/blob/5838487227913be0ab5e382a0fe115706928a576/obspy/io/seiscomp/inventory.py#L69-L76). I suppose there's a wide variety of SeisComp installations out there and thus also SC3ML files at different schema versions.

Since you did #1848 already, can I task you with this trivial extension, @Brtle?
I'd like to have this in 1.1.0 since it might be useful for quite a few people. Do you have some time to work on this in the next weeks? (I've added you to developers, so ideally work in a branch in the main repo for easier collaboration)

### DIFF
--- a/obspy/io/seiscomp/event.py
+++ b/obspy/io/seiscomp/event.py
@@ -18,34 +18,14 @@ from future.builtins import *  # NOQA
 import io
 import os
 import re
-import warnings
 
 from lxml import etree
 
 from obspy.io.quakeml.core import Pickler, Unpickler, _xml_doc_from_anything
-from obspy.io.seiscomp.core import _is_sc3ml as _is_sc3ml_version
 from obspy.io.seiscomp.core import validate as validate_sc3ml
 
 
 SCHEMA_VERSION = ['0.5', '0.6', '0.7', '0.8', '0.9']
-
-
-def _is_sc3ml(path_or_file_object):
-    """
-    Simple function checking if the passed object contains a valid sc3ml file.
-    Returns True of False.
-
-    The test is not exhaustive - it only checks the root tag but that should
-    be good enough for most real world use cases. If the schema is used to
-    test for a StationXML file, many real world files are false negatives as
-    they don't adhere to the standard.
-
-    :type path_or_file_object: str
-    :param path_or_file_object: File name or file like object.
-    :rtype: bool
-    :return: ``True`` if SC3ML file is valid.
-    """
-    return _is_sc3ml_version(path_or_file_object, SCHEMA_VERSION)
 
 
 def _read_sc3ml(filename, id_prefix='smi:org.gfz-potsdam.de/geofon/'):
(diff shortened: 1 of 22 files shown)


QuLogic:
> I'll make this issue into a PR from your branch.

As this now changes the author and date, I don't think this is a good idea any more. You can just put "Fixes #1873" in the description and have it close the issue automatically.

megies:
> As this now changes the author and date, I don't think this is a good idea any more

I agree that this new behavior is now confusing and strange. I still don't like the idea of having twice as many tickets for the same amount of topics. In any case, this issue's OP was by me anyway, so no glitch was introduced at least in this issue by converting to a PR.

Jollyfant:
> Finally, I also modify the minimum usable version of the inventory read module. There is absolutely no change between version 0.5 and 0.7 so it's straightforward.

That's interesting. Thanks for reviewing and editing the inventory too.

Jollyfant:
@megies is there a way to reset Travis CI?

obspy/io/seiscomp/event.py
```diff
@@ -78,12 +80,29 @@ def _read_sc3ml(filename, id_prefix='smi:org.gfz-potsdam.de/geofon/'):
...
+        r'{http://geofon\.gfz-potsdam\.de/ns/seiscomp3-schema/([-+]?'
+        r'[0-9]*\.?[0-9]+)}', sc3ml_doc.tag)
+
+    try:
+        version = match.group(1)
+    except AttributeError:
+        warnings.warn('The file does not appear to be a SC3ML file. Proceed '
+                      'with caution.')
+        version = SCHEMA_VERSION[-1]
+    else:
```
Jollyfant: Should we raise a warning in this branch? e.g. `Schema version %s is not supported, using the most recent version`.

Brtle: A warning is already raised in the [_is_sc3ml](https://github.com/obspy/obspy/blob/ee71bd14278d4153befb2aac044d89edc7242d88/obspy/io/seiscomp/core.py#L82) method. If we also put a warning here, there will be 2 warning messages.

Jollyfant: Right!

megies: Maybe the warning should be moved from the `_is` method to the `_read` method? So that people that explicitly specify the file type during reading also see it?

megies:
> @megies is there a way to reset Travis CI?

Yep. But it's not necessary, the single fail is due to #1879 and it won't pass if I restart the build.

Just curious, if you login to travis with github, don't you get the "restart build" buttons? If not, then being part of the admin team isn't enough for it, I guess.. :-/

screenshot from 2017-09-14 12-18-49

megies:
Would be nice if you added a test that checks the warning message. Then you would also get green light from codecov which is complaining about to high percentage of uncovered lines in the diff, see https://codecov.io/gh/obspy/obspy/compare/f9a8a3fcdaca4c5a50e23155c001db76f2e61867...ee71bd14278d4153befb2aac044d89edc7242d88/diff#D14-89.

Here's an example for a test that checks warning message: https://github.com/obspy/obspy/blob/91a3f2904c54df330fa0a9f4468ea1c234b8b528/obspy/core/tests/test_trace.py#L1338

Jollyfant:
> Just curious, if you login to travis with github, don't you get the "restart build" buttons? If not, then being part of the admin team isn't enough for it, I guess.. :-/

Sorry for offtopic. But actually there is, I just haven't ever noticed it before. Ty!

Brtle:
I changed warnings to exceptions when the namespace was wrong since it always generates an empty file:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<q:quakeml xmlns:q="http://quakeml.org/xmlns/quakeml/1.2" xmlns="http://quakeml.org/xmlns/bed/1.2"/>
```

There isn't an `EventParameters` node so such file raises an exception in the QuakeML module. I also added a small unit test with a wrong namespace.

obspy/io/seiscomp/event.py
Brtle: Done. I have also changed the warning to an exception.

krischer:
I just realized this does not need a new changelog entry as SC3ML support was not in 1.0.3 and the new support is already in the changelog. Thus I hit the big merge button. Thanks a bunch for this!

megies:
Thanks a lot, @Brtle!

PR #1895, opened 2017-09-15
fixing reading underline "_" character from dataless file

### LINKED ISSUE #1893: Failure to read underline "_" character from dataless file
In the attached dataless file ([UP_BACU_HH.dataless.zip](https://github.com/obspy/obspy/files/1303020/UP_BACU_HH.dataless.zip)) the site_name field in blockette 50 holds the string "T3930_b A6689 3930" as can be verified by opening the file using [pdcc](http://ds.iris.edu/ds/nodes/dmc/software/downloads/pdcc/), however running 

    from obspy.io.xseed import Parser
    print Parser("UP_BACU_HH.dataless").blockettes[50][0].site_name

outputs "T3930 b A6689 3930". 

Note that also writing strings with the underline character to a dataless file also fails (sorry, no simple code-snippet available here). 

(running obspy 1.0.3 on open suse installed via pip a few months ago)

barsch:
fixes #1893


obspy/io/xseed/fields.py
```diff
@@ -81,25 +81,28 @@ def _format_string(self, s):
...
         if '_' in self.flags:
-            rx_list.append("_")
+            # underline symbol
+            rx_list.append(r"_")
+        # auto-format
         if 'U' in self.flags and 'L' not in self.flags:
             sn = sn.upper()
         elif 'L' in self.flags and 'U' not in self.flags:
             sn = sn.lower()
-        if 'S' in self.flags and 'X' not in self.flags:
```
barsch: there is no such thing as 'X' in the mseed documentation - so this has been blindly copied (probably by me) from the original source code ...

obspy/io/xseed/fields.py
```diff
@@ -81,25 +81,28 @@ def _format_string(self, s):
...
+            # any punctuation characters (including “_”)
+            rx_list.append(r"[^A-Za-z0-9 ]")
         if 'S' in self.flags:
-            rx_list.append(" ")
+            # spaces between words
+            rx_list.append(r" ")
         if '_' in self.flags:
-            rx_list.append("_")
+            # underline symbol
+            rx_list.append(r"_")
```
megies: These raw string literal prefixes don't have any effect.. or am I missing something?

barsch: I started to generally write regex patterns as raw strings since Python 3.6 started throwing ```DeprecationWarning: invalid escape sequence``` at me - is it needed here? probably not

PR #1896, opened 2017-09-19
QuakeML: fix a bug in reading QuakeML (empty stub MomentTensor added)

megies:
Reading an event with focal mechanism without moment tensor introduces an empty stub moment tensor (instead of keeping it None), which in turn can result in invalid QuakeML files when saving again later.

### DIFF
--- a/obspy/io/quakeml/tests/test_quakeml.py
+++ b/obspy/io/quakeml/tests/test_quakeml.py
@@ -1020,7 +1020,7 @@ class QuakeMLTestCase(unittest.TestCase):
         fm = FocalMechanism()
         event = Event(focal_mechanisms=[fm])
         cat = Catalog(events=[event])
-        cat.write(memfile, format="QUAKEML")
+        cat.write(memfile, format="QUAKEML", validate=True)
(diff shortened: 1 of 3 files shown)


obspy/io/quakeml/tests/test_quakeml.py
```diff
@@ -1009,6 +1009,24 @@ def test_read_same_file_twice_to_same_variable(self):
...
+        # create virtually empty FocalMechanism
+        fm = FocalMechanism()
+        event = Event(focal_mechanisms=[fm])
+        cat = Catalog(events=[event])
+        cat.write(memfile, format="QUAKEML")
+        # now read again, and make sure there's no stub MomentTensor, but
+        # rather `None`
+        memfile.seek(0)
+        cat = read_events(memfile, format="QUAKEML")
+        self.assertEqual(cat[0].focal_mechanisms[0].moment_tensor, None)
```
krischer: Maybe also write it out again with `validate=True` to test the behavior because of which you discovered the bug?

megies:
CI was green, only added changelog now. Merging.

Issue #1874, opened 2017-09-02
https FDSN web service end points not working


megies:
When checking the `InsecureRequestWarning` in #1779, I noticed that those FDSNWS end points we have that go to https addresses don't seem to work..

```python
In [1]: from obspy.clients.fdsn import Client

In [2]: client = Client('GFZ')
---------------------------------------------------------------------------
FDSNException                             Traceback (most recent call last)
<ipython-input-2-1d7f32e7dd73> in <module>()
----> 1 client = Client('GFZ')

/home/megies/git/obspy-master/obspy/clients/fdsn/client.pyc in __init__(self, base_url, major_versions, user, password, user_agent, debug, timeout, service_mappings, force_redirect)
    260             print("Request Headers: %s" % str(self.request_headers))
    261 
--> 262         self._discover_services()
    263 
    264     def get_events(self, starttime=None, endtime=None, minlatitude=None,

/home/megies/git/obspy-master/obspy/clients/fdsn/client.pyc in _discover_services(self)
   1479                    "be due to a temporary service outage or an invalid FDSN "
   1480                    "service address." % self.base_url)
-> 1481             raise FDSNException(msg)
   1482 
   1483         # Cache.

FDSNException: No FDSN services could be discovered at 'https://geofon.gfz-potsdam.de'. This could be due to a temporary service outage or an invalid FDSN service address.

In [3]: client = Client('SCEDC')
---------------------------------------------------------------------------
FDSNException                             Traceback (most recent call last)
<ipython-input-3-7c19ea139f48> in <module>()
----> 1 client = Client('SCEDC')

/home/megies/git/obspy-master/obspy/clients/fdsn/client.pyc in __init__(self, base_url, major_versions, user, password, user_agent, debug, timeout, service_mappings, force_redirect)
    260             print("Request Headers: %s" % str(self.request_headers))
    261 
--> 262         self._discover_services()
    263 
    264     def get_events(self, starttime=None, endtime=None, minlatitude=None,

/home/megies/git/obspy-master/obspy/clients/fdsn/client.pyc in _discover_services(self)
   1479                    "be due to a temporary service outage or an invalid FDSN "
   1480                    "service address." % self.base_url)
-> 1481             raise FDSNException(msg)
   1482 
   1483         # Cache.

FDSNException: No FDSN services could be discovered at 'https://service.scedc.caltech.edu'. This could be due to a temporary service outage or an invalid FDSN service address.

In [4]: client = Client('USGS')
[...]

PR #1902, opened 2017-09-20
Adding the ICGC mapping

krischer:
Adds the ICGC mapping discovered by @megies in https://github.com/obspy/obspy/issues/1874#issuecomment-330902354.


megies:
Oh, thanks @krischer.. would've overlooked this one after closing the other HTTPS issue.

PR #1903, opened 2017-09-20
Directly return response objects in NRL

### LINKED ISSUE #1897: get_datalogger_resp() and get_sensor_resp()
I appreciate the new NRL class in the Obspy master.. great job!

It would be valuable to be able to pull separate sensor and datalogger response stages into a Obspy Response object, as get_response() does.

In some of my use cases, my sensor is home-made not part of the NRL library while the datalogger is. After this, I could append the stages using response.response_stages.append()

krischer:
Two helper functions used to return RESP strings - they now return fully formed `Response` objects. I also renamed the functions a bit.
 
CC: @nick-iris @dsentinel @andreww @jdassink 

Feel free to veto.

Fixes #1897

### DIFF
--- a/obspy/clients/nrl/client.py
+++ b/obspy/clients/nrl/client.py
@@ -138,29 +138,37 @@ class NRL(object):
     def _clean_str(self, string):
         return string.strip('\'"')
 
-    def get_datalogger_resp(self, datalogger_keys):
+    def get_datalogger_response(self, datalogger_keys):
         """
-        Get the RESP string of a datalogger by keys.
+        Get the datalogger response.
 
         :type datalogger_keys: list of str
-        :rtype: str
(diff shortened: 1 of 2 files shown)


krischer:
Renaming is safe as the NRL client is not part of the last release so its not yet public API.

krischer:
+TESTS:clients.nrl

krischer:
I added a test - can be merged if CI is alright.

megies:
Uhm.. @krischer NRL tests were not run during CI likely since you added that magic comment after the push..

krischer:
Yea I noticed as well - but I ran the tests locally on Python 2 and 3 and decided I'm confident enough in it to merge it ;)

megies:
Fair enough.

jdassink:
Thanks y'all!

PR #1915, opened 2017-10-07
docs build(bot): add overview page with sphinx build warnings

megies:
.. and set status to "failure" if docs build has warnings or errors.

Right now this will fail with zillions of warnings but eventually if we fix current problems this will help keep docs sane when merging PRs.


megies:
+DOCS

megies:
Works in principle.. commit status now points to a page with info about warnings/errors. But, those pages seem to only have very few warnings.. when i build docs locally I usually get ~3000 warnings.. -_-

screenshot from 2017-10-07 18-47-55
screenshot from 2017-10-07 18-45-46
screenshot from 2017-10-07 18-45-26


In any case, merging, as having the docs build warning info in the commit status is still very valuable.

PR #1911, opened 2017-10-08
"Supported Formats" table missing in docs for Inventory.write()

megies:
Should be done same as for `Catalog.write()`

+DOCS


megies:
Any objections to the change of exception type? Otherwise will merge this in a day or so.

PR #1185, opened 2015-12-03
SEED -> Inventory read support + NRL client

dsentinel:
I'm developing a tool for temporary network operators to create metadata for their deployments. I had started with the intent of creating Dataless SEED, but am considering using the modern StationXML and converting to Dataless for downstream processing, or updating the downstream processes to handle StationXML.
 The idea is to scan the mseed recorded, use that data to create net, sta, chan and other info in those headers to populate. The user points to the correct datalogger and sensor from the NRL, and fills in the remaining info such as lat long.

+TESTS:clients.nrl

PR #1899, opened 2017-10-09
Extend "Create StationXML from scratch" tutorial to include Response from NRL

megies:
Would be great to extend the tutorial page on ["Creating StationXML file from scratch"](http://docs.obspy.org/master/tutorial/code_snippets/stationxml_file_from_scratch.html) to include the new powerful NRL client to even add Response information. 

Most of the [information is already on the NRL client API landing page](http://docs.obspy.org/master/packages/obspy.clients.nrl.html), so it's a low hanging fruit.. and it will increase visibility of the new functionality.

CC @dsentinel @krischer

+DOCS

### DIFF
--- a/misc/docs/source/tutorial/code_snippets/stationxml_file_from_scratch.rst
+++ b/misc/docs/source/tutorial/code_snippets/stationxml_file_from_scratch.rst
@@ -1,6 +1,6 @@
-=============================================================================
-Creating a StationXML file (without instrument responses) from Scratch
-=============================================================================
+=======================================
+Creating a StationXML file from Scratch
+=======================================
 
 Creating a custom StationXML file is a task that sometimes comes up in
 seismology. This section demonstrates how to it with ObsPy. Please note that
@@ -14,9 +14,6 @@ StationXML file and many arguments are optional. ObsPy will validate the
 resulting StationXML file against its schema upon writing so the final file is
 assured to be valid against the StationXML schema.
 
-Please note that this tutorial does not cover the authoring of instrument
-responses which would be a larger endeavour.
-
 The following illustration shows the basic structure of ObsPy's internal
 representation.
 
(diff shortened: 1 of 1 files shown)


jdassink:
Looks good y'all. I was working on it (too slowly) this weekend; noticed that you get a few "empty" stages (gain 1) when requesting a datalogger response, that appear as a placeholder for the sensor stages, for example stages 1 and 2 in the following request:

In [9]: nrl.get_datalogger_response(['Quanterra','Q412x (default cascade only)','40'])
Out[9]:
Channel Response
	From M/S (Velocity in Meters per Second) to COUNTS (Digital Counts)
	Overall Sensitivity: 424507 defined at 0.050 Hz
	6 stages:
		**Stage 1: PolesZerosResponseStage from M/S to V, gain: 1
		Stage 2: ResponseStage from V to V, gain: 1**
		Stage 3: CoefficientsTypeResponseStage from V to COUNTS, gain: 427000
		Stage 4: CoefficientsTypeResponseStage from COUNTS to COUNTS, gain: 1
		Stage 5: CoefficientsTypeResponseStage from COUNTS to COUNTS, gain: 0.997077
		Stage 6: CoefficientsTypeResponseStage from COUNTS to COUNTS, gain: 0.997077

Similarly, when requesting a sensor response, you get an "empty" stage 2:

In [4]: nrl.get_sensor_response(['Streckeisen','STS-2','20000','3 - installed 04/97 to present'])
Out[4]:
Channel Response
	From M/S (Velocity in Meters per Second) to COUNTS (Digital Counts)
	Overall Sensitivity: 20000 defined at 1.000 Hz
	2 stages:
		Stage 1: PolesZerosResponseStage from M/S to V, gain: 20000
		**Stage 2: CoefficientsTypeResponseStage from V to COUNTS, gain: 1**


Can this be changed? It would be confusing to have to go in after getting a NRL response to cut out the stages that are not needed.

megies:
> Looks good y'all. I was working on it (too slowly) this weekend; noticed that you get a few "empty" stages (gain 1) when requesting a datalogger response, that appear as a placeholder for the sensor stages, for example stages 1 and 2 in the following request:

> Can this be changed? It would be confusing to have to go in after getting a NRL response to cut out the stages that are not needed.

We also discussed this and actually already had them "dummy" stages removed, but @dsentinel brought up absolutely valid points in keeping those "dummy" stages from NRL and they really do no harm, so we agreed that it's better to not tamper with what comes out of NRL. See discussion in https://github.com/obspy/obspy/pull/1185#discussion_r130435665.

megies:
> I was working on it (too slowly) this weekend;

Hope I didn't get in the way too much, it's just that I want this release out finally, and soon (i.e. rather yesterday then tomorrow). ;-)

jdassink:
No problem at all. Happy you all helped. If needed, I can provide an example where I make a composite of a home-made sensor with known Poles and Zero's and a manufactured datalogger that is part of the NRL (which is 'my case').

megies:
Sure, feel free to add more complex cases to the tutorial. At least for the release we have a minimum example with the simplest use case.

jdassink:
Here's a snippet of a simple calculation for an IMS infrasound sensor, with some modifications for the datalogger.

from obspy.clients.nrl import NRL
nrl = NRL()

response = nrl.get_response(
	sensor_keys=['CEA-DASE','MB2000 Microbarometer (infrasound)'],
	datalogger_keys=['Generic','Unity'])

# modify some of the datalogger parameters, e.g. sampling at 20.0 Hz instead of 40.0 Hz.
response.response_stages[2].decimation_input_sample_rate = 20.0
# it's a 24 bit datalogger, digitizing range over -16 V to 16 V (= 32 V), so 2**24 counts for 32 V.
response.response_stages[2].stage_gain = 2**24 / 32.0

response.recalculate_overall_sensitivity()
print response
response.plot(1e-3)

megies:
Nice example, would be cool if you made a PR to include it in the "StationXML from scratch" tutorial.

Issue #1389, opened 2016-05-03
Use FDSN service to fetch citations/DOIs by network code


megies:
Just learned of this [service by FDSN that fetches citations for given network codes](http://www.fdsn.org/networks/citation/?networks=GE).. might be worth a thought to add a method to `Inventory` (and/or `Network`, or also `Stream`..) that uses this service.

PR #1919, opened 2017-10-10
FDSNWS Routing

krischer:
This PR builds up straight from the last changes in #1779. I made this a different PR to enable some extra discussion as it changes things. The main user visible change is that this now also adds support for the EIDAWS routing service (thus this also fixes #1397 - to some part at least; the token authentication is another discussion in my opinion).

I slimmed it down a lot by making one simplifying assumption: There is not need for ObsPy to understand the return values of the routing services in any detail - the results can directly be used and send as POST requests to each FDSNWS implementation. But the whole thing including base router class, parallelization, and both concrete implementations is only about 240 statements long - I feel fairly comfortable maintaining this in the long run. It also has full test coverage for all new parts.

Furthermore it also does not care about restarting or retrying downloads from failing FDSNWS nodes - it behaves more or less exactly like the single provider FDSNWS client. We still have the mass downloader for people who really need to download a lot of data.

Usage:

```python
>>> from obspy.clients.fdsn import RoutingClient
>>> import obspy
>>> c = RoutingClient("federator")  # or "eidaws"
>>> st = c.get_waveforms(channel="LHZ", starttime=obspy.UTCDateTime(2017, 1, 1),
                         endtime=obspy.UTCDateTime(2017, 1, 1, 0, 5),
                         latitude=10, longitude=10, maxradius=35)  
>>> print(st)
```
```
284 Trace(s) in Stream:

8A.NEPA..LHZ | 2017-01-01... - 2017-01-01T... | 1.0 Hz, 300 samples
...
(282 other traces)
...
WM.SFS..LHZ | 2016-12-31T... - 2017-01-01T... | 1.0 Hz, 197 samples

```

It works the same for stations and in a similar way across both routing services. Both route on different parameters of course so this has to be taken into account. Most parameters and any additional kwargs are also passed on in the requests to each individual data center (if a data center claims support for a parameter in their WADL).

Also all data returned by one of the FDSNWS clients will now have a `Trace.stats._fdsnws_dataselect_url` to improve provenance.

CC: @CelsoReyes @nick-iris @chad-iris @andres-h

+DOCS +TESTS:clients.fdsn


Jollyfant:
I would avoid using the term "federator" for IRIS. EIDA refers to it as being a "federator" too. How about just using "IRIS" and "EIDA"?

krischer:
> I would avoid using the term "federator" for IRIS. EIDA refers to it as being a "federator" too. How about just using "IRIS" and "EIDA"?

I'm fine with renaming to something else but I would consider renaming to "IRIS" and "EIDA" confusing - this would give the impression that both are implementations of the same thing (especially because that is how our normal fdsnws client behaves)  - whereas they are actually pretty different in what they can do. I'm open for other suggestions though.

Jollyfant:
Sorry I am a little confused -- does this class only do routing for EIDA? And as an extra feature it can collect waveform data from IRIS too?

megies:
`RoutingClient` has two subclasses, one handling IRIS fedcatalog and the other handling EIDAWS.

Jollyfant:
> RoutingClient has two subclasses, one handling IRIS fedcatalog and the other handling EIDAWS.

I understand that much but what is the "fedcatalog" and what is "EIDAWS"?

megies:
Uhmm..?
 - https://service.iris.edu/irisws/fedcatalog/1/
 - http://www.orfeus-eu.org/data/eida/webservices/routing/

megies:
Unless I'm overlooking anything the `get_waveforms()` of the fedcatalog flavor can be also used by the eidaws flavor, so I'll push it to the base class.

Jollyfant:
If the fedcatalog returns waveforms and our routing service only returns routes I don't understand why they share a class.. Unless the EIDA part now also returns waveforms after the change by @megies -- but then my initial point about calling only IRIS a "federator" is confusing.

krischer:
Both return routes - the data still has to be fetched from the individual data centers.

Jollyfant:
Oh ok.. your example in the OP calls the `get_waveforms` method so I assumed it was for that.

krischer:
> Oh ok.. your example in the OP calls the get_waveforms method so I assumed it was for that.

It is - internally ObsPy first gets the routes and then gets the data from each individual data center.

Jollyfant:
I see, thanks. The name should be `eida` as  we reserve the `eidaws` keyword for services unrelated to FDSN (e.g. Routing/WFCatalog). And this class routes FDSN web services. IMHO the "federator" should definitely have IRIS in its name -- maybe `iris-federator`.

krischer:
okay- so how about `eida-routing` and `iris-federator`?

Jollyfant:
I think that would be more descriptive. I know it sounds like bike shedding but EIDA is introducing a "federator" too. Just want to avoid future confusion when this is merged in to the API.

krischer:
> I think that would be more descriptive. I know it sounds like bike shedding but EIDA is introducing a "federator" too. Just want to avoid future confusion when this is merged in to the API.

Will this also be some kind of routing service? If both EIDA and IRIS have one I feel like it should just become an FDSN standard.

Jollyfant:
> Will this also be some kind of routing service? If both EIDA and IRIS have one I feel like it should just become an FDSN standard.

It is a web service that automatically federates (routes) requests across EIDA and returns mSEED or StationXML. Users will have a single HTTP entry point for all EIDA data.

krischer:
> It is a web service that automatically federates (routes) requests across EIDA and returns mSEED or StationXML. Users will have a single HTTP entry point for all EIDA data.

Okay that is different then. Any chance to give this a different name? Otherwise there are two federators that do different things.

Also will it be fully fdsnws compatible? Then this would be extremely easy for us to support.

Jollyfant:
> Okay that is different then. Any chance to give this a different name? Otherwise there are two federators that do different things.

No :-( that is why I am lobbying against calling this a "federator". Of course IRIS is free to call it whatever,  but it needs their namespace.

It will be FDSN compatible.

krischer:
> No :-( that is why I am lobbying against calling this a "federator". Of course IRIS is free to call it whatever, but it needs their namespace.

Haha :) I see. Who can we petition to change it? This will be extremely confusing for users. And while IRIS does not own the name they used it first for a public service and it would IMHO be in the interest of all to have different names for both if they don't do the same thing.

> It will be FDSN compatible.
Great :)

megies:
> This will be extremely confusing for users

We just need an ObsPy client on top that combines everything and then it's a one-liner for our users.. ;-)

krischer:
The `eida-router` got a couple of enhancements: For `get_waveforms(_bulk)()` it will now first submit a station query with the routing service and then call each data center with `station` request (including geographical and other constraints). This information is then used to assemble the final bulk request to the router to actually download the data.

This results both routing services to behave almost identical from a user's point of view, e.g.

```python
c.get_waveforms(starttime=UTCDateTime(...), endtime=UTCDateTime(...),
                latitude=10, longitude=20, maxradius=30)
```

and similar queries work as expected, also for the eida service.

The initialization strings have also been renamed to `eida-routing` and `iris-federator`, respectively.

megies:
> This results both routing services to behave almost identical from a user's point of view, e.g.

🚀 🎉

krischer:
In case someone ever needs it: Here is a prototype of a router client that combines the routes of multiple routing services.

https://github.com/krischer/obspy/blob/fdsnws-routing-combined-client/obspy/clients/fdsn/routing/combined_routing_client.py

Right now it seems pretty useless to have as we only have 2 routing services and the IRIS federator in our tests seems to always return some superset of what the EIDA router returns.

krischer:
I also think this is pretty much done. Test failures are due to an expired certificate on our examples web page.

Let's wait one more day for give other people a chance to chime in and then we can merge this if no objections arise.

CTrabant:
> I would avoid using the term "federator" for IRIS. EIDA refers to it as being a "federator" too. How about just using "IRIS" and "EIDA"?
> 
On the contrary, our system is called the IRIS Federator.  Please retain our branding whenever appropriate.

chad-iris:
One of the original desires when the IRIS portion of this work started was to be able to annotate the returned data with some sort of structure or tag to identify which data center it was retrieved from.  Exposing, or at a minimum retaining this information, is important.  Data centers should not become totally opaque in the data collection process, they deserve recognition and credit along with the network operators/owners for data use.  In my quick review I did not see anything like this in the current code.

Questions: would it be possible to add a "data center" attribute somewhere in a Stream object?  I don't know the ObsPy internals well enough, but is there a Stats dictionary or something where this annotation could be accessible to the user?

Excuses if I've missed this someone in the PR.

krischer:
Last line in the initial post ;)

> Also all data returned by one of the FDSNWS clients will now have a Trace.stats._fdsnws_dataselect_url to improve provenance.

chad-iris:
> Last line in the initial post ;)

Ha!  Thanks, sorry to have not read far enough...

megies:
> Data centers [..] deserve recognition and credit along with the network operators/owners for data use

Eventually also #1389 should be implemented to help provide a means for users to assemble DOIs for proper attribution/citation of data. In any case, right now being able to tell what FDSN server the data comes from is already a minimum solution.

So everybody happy now with the wording?
 - `RoutingClient('iris-federator')`
 - `RoutingClient('eida-routing')`

We want to get 1.1.0 out within a day or two, so any wishes for changes have to be voiced right now.

Jollyfant:
> On the contrary, our system is called the IRIS Federator.  Please retain our branding whenever appropriate.

You're right, Chad - I hope the branding `iris-federator` works for you guys too.

krischer:
Rebased on latest master to get CI and doc building working again (see #1921 ).

krischer:
If somebody still wants to have a look at this: please do it ASAP. Otherwise we'll merge this tomorrow at the latest.

krischer:
No further responses. I'm merging.

Thanks a lot for everyone contributing to this long-term-in-the-making feature :) 🎉

PR #1920, opened 2017-10-10
Fix in mass downloader test case

krischer:
Fixes a regression introduced in #1867. Only affected the tests - I could not happen in a real run.


krischer:
+TESTS:clients.fdsn

PR #1921, opened 2017-10-10
HTTP for the examples

krischer:
As we currently have an expired certificate and this results in most CI failing I changed all the examples to use the HTTP version of our examples server. I don't see any harm in that but maybe I am mistaken.


megies:
This currently also crashes our docs builds, I'm also 👍 for HTTP..

megies:
Gonna merge this for now, so that current docs can be built again.. if anybody's opposing this change please open a follow-up.

QuLogic:
I don't particularly like that the documentation links were changed too.

krischer:
Fair enough - now that it works again I'd be fine with reverting this whole PR. We had to change this as doc building also no longer works.

PR #1922, opened 2017-10-11
Return empty inventory if read_inventory on http:// returns 204

Jollyfant:
ObsPy throws when `read_inventory` is called with a 204 response. I think it would be good to return an empty inventory instead a `TypeError`.

    TypeError: Unknown format for file /tmp/obspy-cg1Kog.tmp

Can add a test using our webservice, but what is the best place to put it? And is it OK to include tests over http?

    www.orfeus-eu.org/fdsnws/station/1/query?network=BOGUS

### DIFF
--- a/obspy/core/tests/test_util_base.py
+++ b/obspy/core/tests/test_util_base.py
@@ -9,7 +9,7 @@ import unittest
 
 from obspy.core.compatibility import mock
 from obspy.core.util.base import (NamedTemporaryFile, get_dependency_version,
-                                 download_to_file)
+                                  download_to_file)
 from obspy.core.util.testing import ImageComparison, ImageComparisonException
 
 from requests import HTTPError
--- a/obspy/core/tests/test_util_base.py
+++ b/obspy/core/tests/test_util_base.py
@@ -8,7 +8,8 @@ import shutil
 import unittest
 
 from obspy.core.compatibility import mock
-from obspy.core.util.base import NamedTemporaryFile, get_dependency_version, download_to_file
+from obspy.core.util.base import (NamedTemporaryFile, get_dependency_version,
+                                 download_to_file)
 from obspy.core.util.testing import ImageComparison, ImageComparisonException
 
 from requests import HTTPError
--- a/obspy/core/tests/test_util_base.py
+++ b/obspy/core/tests/test_util_base.py
@@ -107,9 +107,9 @@ class UtilBaseTestCase(unittest.TestCase):
                 mocked_get.return_value.reason = reason
                 with self.assertRaises(HTTPError) as e:
                     download_to_file(url, None)
-                    self.assertEqual(e.exception.args[0],
-                                     "%s HTTP Error: %s for url: %s" %
-                                     (code, reason, url))
+                self.assertEqual(e.exception.args[0],
+                                 "%s HTTP Error: %s for url: %s" %
+                                 (code, reason, url))
 
 
 def suite():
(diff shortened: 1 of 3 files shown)


krischer:
Hmm - this is a bit tricky and your change relies on FDSNWS semantics. I would rather check the error code (and maybe change the `download_to_file()` routine to also return the failure reason) and make sure all the `read_X()` methods raise a nice error message if the return code is not 200.

This would make it independent of some assumption in FDSNWS and other webservices might return 204 for a number of reasons.

Jollyfant:
Pretty sure the `download_to_file()` method already returns the failure reason during `r.raise_for_status()`. It's just 204 that needs to be added manually. Either by another raise or an empty `Stream` or `Inventory`.

krischer:
Ah i see. How about replacing `r.raise_for_status()` with

```python
if r.status_code != 200:
    raise requests.HTTPError("HTTP return code %i: %s" % (r.status_code, r.reason))
```

or something along this line?

Or is there any status aside from 200 where we would not want to raise? I'm personally not a big fan of just returning empty objects as users might not understand what is happening.

Jollyfant:
Fair enough!

krischer:
Thanks! Can you fix the code formatting and maybe add a small (mock?) test?

Jollyfant:
Ok tried using mock. Could you please the  check the test carefully because I haven't used it before..

obspy/core/tests/test_util_base.py
```diff
@@ -90,6 +92,24 @@ def test_image_comparison(self):
...
+        """
+        Tests HTTP Error on 204, 400, and 500
+        """
+        url = "http://obspy.org"
+        for response_tuple in [("204", "No Content"), ("400", "Bad Request"),
+                               ("500", "Internal Server Error")]:
+          code = response_tuple[0]
+          reason = response_tuple[1]
+          with mock.patch("requests.get") as mocked_get:
+            mocked_get.return_value.status_code = code
```
megies: please 4 spaces for indentation..

obspy/core/tests/test_util_base.py
```diff
@@ -90,6 +93,24 @@ def test_image_comparison(self):
...
+            code = response_tuple[0]
+            reason = response_tuple[1]
+            with mock.patch("requests.get") as mocked_get:
+                mocked_get.return_value.status_code = code
+                mocked_get.return_value.reason = reason
+                with self.assertRaises(HTTPError) as e:
+                    download_to_file(url, None)
+                    self.assertEqual(e.exception.args[0],
+                                     "%s HTTP Error: %s for url: %s" %
+                                     (code, reason, url))
```
krischer: Can you move the asserting statement out of the context manager? In this case it probably does not matter but the `assertRaises()` would also be satisfied if the exception is raised in the `assertEqual()` call if it is part of the context manager.

obspy/core/util/base.py
```diff
@@ -573,7 +573,10 @@ def download_to_file(url, filename_or_buffer, chunk_size=1024):
     except TypeError:
         r = requests.get(url)
 
-    r.raise_for_status()
+    # Raise anything except for 200
+    if r.status_code != 200:
+        raise requests.HTTPError(u'%s HTTP Error: %s for url: %s'
```
krischer: It the `u` prefix necessary here?

Jollyfant: Nah probably not. I copied it in from `requests.raise_for_status()`.

krischer:
Thanks! I'll merge after CI has run.

Issue #1909, opened 2017-10-04
read_nordic - maximum azimuthal gap


shicks-seismo:
The current version of read_nordic doesn't appear to read in maximum azimuthal gap and populate originquality, which is written on the second line of the header.

Example:

>  2016  313 0718 21.8 L  -1.997 -14.342 35.7  SOT 16 1.1 6.0LSOT                1
>  GAP=195        1.03      27.7     7.6 33.6 -0.1871E+03 -0.2096E+03  0.8193E+03E
>                                                                                6
>  STAT SP IPHASW D HRMM SECON CODA AMPLIT PERI AZIMU VELO AIN AR TRES W  DIS CAZ7
>  S06D     P   1    718 29.40                             129    0.66 9 37.4 345
>  S06D     S   2    718 34.10                             129    0.06 6 37.4 345

calum-chamberlain:
I can probably have a go at that next week if @shicks-seismo hasn't done it already?  There are quite a few things in the nordic format that I didn't implement, some of them because I didn't find an obvious place for them in events (but that is likely because I didn't look too hard), so if you notice any other things that could be read in (if there) happy to include them too.

shicks-seismo:
Hi Calum,

That would be great if you could. I don’t quite have the time just now, and it would take me a bit of time to delve into the reading of the second line header.

calum-chamberlain:
Currently not included are line types:
- [ ] E (hyp error estimates)
- [ ] F (Fault plane solution)
- [ ] H (High accuracy hypocenter line) - *note, this should probably be preferred_origin if included*
- [ ] M (Moment tensor solution)
- [ ] E13 and EC3 line (explosion info)
- [ ] MACR03 line

I will go ahead and try and implement a load of these where I can find appropriate space, and update the docs to highlight what I don't end up implementing.  The current implementation is a bit bare-bones and should be better.

calum-chamberlain:
On the Error line note, before I write a function to convert from an error ellipse and co-variance in x, y, z to one with semi-major and semi-minor axes lengths and orientation angles, does ObsPy already have one, and if not, where would you like me to put it?

megies:
> one with semi-major and semi-minor axes lengths and orientation angles, does ObsPy already have one, and if not, where would you like me to put it?

Not sure what you mean, the `ConfidenceEllipsoid` should have everything according to QuakeML angle notations: https://github.com/obspy/obspy/blob/master/obspy/core/event/base.py#L1219-L1226

[...]

PR #1924, opened 2017-10-12
WIP: Add more features to nordic IO

calum-chamberlain:
# Work in progress, do not merge yet!
### What does this PR do?

This PR attempts to add more functionality to Nordic format IO operations. This will add in:
~~- [x] Read *hyp* error estimates and convert to confidenceEllipsoid;~~
~~- [x] Write confidenceEllipsoid to *hyp* uncertainty estimates;~~
- [x] Read focal mechanism solutions;
- [x] Write focal mechanism solutions;
- [x] Read moment tensor solutions;
- [x] Write moment tensor solutions.
- [x] latin1 encoded reading (currently raises a `UnicodeDecodeError` in Py3 and other errors in py2) - how should format be detected, `chardet`, or should the user specify if there is an error? Could ask Jens what formats are accepted and try then?
~~-  [ ] More `xyz_to_confidence_ellipsoid` and `confidence_ellipsoid_to_xyz` to `obspy.core.event.origin`.~~
- [x] Review comments.

The ability to read *Hyp* derived errors was requested by @shicks-seismo in issue #1909 

### PR Checklist
- [ ] All tests still pass.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .~~


calum-chamberlain:
This PR is more to document progress, and ask for help on converting between covariance matrix and confidenceEllipsoid (and back).  I have made a stab at this in [`obspy.io.nordic.core.xyz_to_confidence_ellipsoid`](https://github.com/obspy/obspy/pull/1924/files#diff-4bc6ea61b90645c5ad3a04a2f843d8c8R249) and the sister function: [`obspy.io.nordic.core.confidence_ellipsoid_to_xyz`](https://github.com/obspy/obspy/pull/1924/files#diff-4bc6ea61b90645c5ad3a04a2f843d8c8R293), but I'm failing at the moment - welcome any changes here!

I don't think this PR is urgent.

calum-chamberlain:
Reading now implemented, all tests pass expect `confidenceEllipsoid` conversion, which I will work to fix at some point in the next couple of weeks.

calum-chamberlain:
@shicks-seismo, I think the reading in is working, do you have any expected results for confidence ellipsoids from x, y, z and covariance? I'm still stuck on writing (going back the other way), I can get the right numbers but not the right order.  Any help appreciated.

megies:
> This PR is more to document progress, and ask for help on converting between covariance matrix and confidenceEllipsoid (and back). I have made a stab at this in obspy.io.nordic.core.xyz_to_confidence_ellipsoid and the sister function: obspy.io.nordic.core.confidence_ellipsoid_to_xyz, but I'm failing at the moment - welcome any changes here!

I think these helper functions should eventually be moved to `obspy.core.event.origin` for general use. No hurry though, can also be refactored at the end when everything is ready, and we won't do 1.2.0 until march/april.

shicks-seismo:
Hi @calum-chamberlain, sorry for the late reply. I am not really using the error ellipsoids so I don't really have any expectations here.
Note my new PR @ #1991 regarding longitude and depth reading.

calum-chamberlain:
Brief update on this: I have been in touch with Jens Havskov and he thinks they (the SEISAN team) have conversions from x, y, z to ellipse (oid?), but not the reverse, which is still the part I am struggling with. He also tried out a nordic file with some interesting things that I didn't account for, but have included here.

One other thing his file highlighted is that this module doesn't currently account for different encodings - his file was encoded as latin1, which breaks these functions.  Do we have a good/standard way of checking encoding ([`chardet`](https://chardet.github.io/) seems like the tool, but would slow things down most of the time)? As a prefred way of doing it, I have asked Jens what the standard encodings are and then I can add in a `try: .. except:` clause to try different encodings. Thoughts @krischer @megies ?

megies:
> One other thing his file highlighted is that this module doesn't currently account for different encodings - his file was encoded as latin1, which breaks these functions. Do we have a good/standard way of checking encoding (chardet seems like the tool, but would slow things down most of the time)? As a prefred way of doing it, I have asked Jens what the standard encodings are and then I can add in a try: .. except: clause to try different encodings. Thoughts @krischer @megies ?

without looking at the details.. if there really are different encodings used in the wild, the easiest option would be to at least add an `encoding` kwarg that can be specified by the user (maybe defaulting to `ASCII`), so at least users can read such files without hacking the code

krischer:
I think trying a few default encodings would be good and simple first step. Otherwise you could raise an error and tell users that they have to specify the encoding and maybe even point them to the `chardet` library to figure out which encoding they have.

calum-chamberlain:
The word from Jens is that nordic files in seisan are read in as bytes and decoded as single-byte characters, which seems to be consistent with everything being `latin-1` encoded.  This appears to only have been an issue for some of the less-used characters (like: Ø, in the included `dos-file.sfile` test file).

Reading everything using:
```python
f = open(sfile, 'r', encoding='latin-1')
```
seems to work fine and pass all my tests...

krischer:
Okay good. I still think an optional `encoding` argument would be useful for example when somebody manually edits a file and accidentally saves it in a different encoding. Also its really trivial to implement.

obspy/io/nordic/core.py
```diff
@@ -57,19 +57,22 @@ def __init__(self, value):
...
+def _is_sfile(sfile, encoding='latin-1'):
     """
     Basic test of whether the file is nordic format or not.
 
     Not exhaustive, but checks some of the basics.
 
     :type sfile: str
     :param sfile: Path to sfile
+    :type encoding: str
+    :param encoding: Encoding for file, used to decode from bytes to string
```
krischer: This sounds a bit too technical. How about just "Encoding of the file"?

CHANGELOG.txt
```diff
@@ -151,6 +151,11 @@ master:
     * Also parse author information and COMMENT line (see #1484)
     * Fix reading hypocenter files created by NonLinLoc versions of the 6.0.x
       beta branch (see #1760 and #1783)
+ - obspy.io.nordic:
+    * Add ability to read and write focal mechanism, hypocentral error, moment
+      tensor information
+    * Add explicit warnings about lines not yet able to read.
```
krischer: I don't think that's a complete sentence.

obspy/io/nordic/tests/test_nordic.py
```diff
@@ -412,6 +432,8 @@ def test_write_select(self):
             with warnings.catch_warnings():
                 warnings.simplefilter('ignore', UserWarning)
                 write_select(cat, filename=tf.name)
+            if not _is_sfile(tf.name):
+                raise(NordicParsingError("format is wrong somehow"))
```
krischer: Replace with `self.assertTrue(_is_sfile(tf.name))`.

obspy/io/nordic/tests/test_nordic.py
```diff
@@ -546,6 +587,18 @@ def test_distance_conversion(self):
...
                  p.split()[1] == 'HZ'][0].split()[-1]), 30)
 
+    def test_convert_xyz_error(self):
+        """
+        Test that we are able to map between xyz errors and ellipsoid
+        """
+        errors = {'x_err': 1., 'y_err': 1.5, 'z_err': 2.0,
+                  'xy_cov': 0.0, 'xz_cov': 0.0, 'yz_cov': 0.0}
+        confidence_ellipsoid = xyz_to_confidence_ellipsoid(errors)
+        print(confidence_ellipsoid)
```
krischer: Please remove the print statement.

obspy/io/nordic/tests/test_nordic.py
```diff
@@ -546,6 +587,18 @@ def test_distance_conversion(self):
...
+        errors_back = confidence_ellipsoid_to_xyz(confidence_ellipsoid)
+        self.assertEqual(errors, errors_back)
+        # TODO: Test mapping for known result.
```
krischer: This should really be done as its otherwise hard to really trust the calculations - so the test just shows that one function is the inverse of the other which is necessary but not sufficient IMHO. For simple cases it should also be fairly easy to do.

obspy/io/nordic/core.py
```diff
@@ -28,19 +28,26 @@
...
-mag_mapping = {"ML": "L", "MLv": "L", "mB": "B", "Ms": "S", "MW": "W",
-               "MbLg": "G", "Mc": "C"}
+mag_mapping = {"ML": "L", "MLv": "L", "mB": "B", "Ms": "s", "MS": "S",
+               "MW": "W", "MbLg": "G", "Mc": "C"}
 
 onsets = {'I': 'impulsive', 'E': 'emergent'}
 
+accepted_tags = ['1', '6', '7', 'E', ' ', 'F', 'M', '3']
+# List of currently implemented line-endings, which in Nordic mark what format
+# info in that line will be.
```
krischer: Should this go above the code?

obspy/io/nordic/core.py
```diff
@@ -50,51 +57,74 @@ def __init__(self, value):
...
             return False
     else:
         return False
 
 
-def _get_headline(f):
+def _get_line_tags(f):
+    f.seek(0)
+    line = f.readline()
+    if len(line.rstrip('\n').rstrip('\r')) != 80:
```
krischer: Just `.rstrip('\n\r')` should be enough. Stripping always removes as many of those. Just using `.rstrip()` should actually also be enough as it by default will remove any ASCII whitespace I believe.

obspy/io/nordic/core.py
```diff
@@ -50,51 +57,74 @@ def __init__(self, value):
...
+        try:
+            line_id = line[79]
+        except IndexError:
+            line_id = ' '
+        if line_id in tags.keys():
+            tags[line_id].append((line, i))
+        elif line_id in accepted_tags:
+            tags.update({line_id: [(line, i)]})
+        else:
+            UserWarning("Lines of type %s have not been implemented yet, "
```
krischer: You have to use `warnings.warn()` to actually raise a warning.

obspy/io/nordic/core.py
```diff
@@ -50,51 +57,74 @@ def __init__(self, value):
...
-    else:
-        return None
+        try:
+            line_id = line[79]
+        except IndexError:
+            line_id = ' '
+        if line_id in tags.keys():
+            tags[line_id].append((line, i))
+        elif line_id in accepted_tags:
+            tags.update({line_id: [(line, i)]})
```
krischer: This would be about the same number of lines but it feels a bit more elegant to me:

```python
tags = collections.defaultdict(list)
...
if line_id in accepted_tags:
    tags[line_id].append((line, i))
```

obspy/io/nordic/core.py
```diff
@@ -322,40 +414,113 @@ def _readheader(f):
...
+                        ('longitude', slice(30, 38)),
+                        ('depth', slice(38, 43))]:
+        try:
+            new_event.origins[0].__dict__[key] = float(line[_slice])
+        except ValueError:
+            new_event.origins[0].__dict__[key] = None
+    if new_event.origins[0].depth:
+        new_event.origins[0].depth *= 1000
+    # new_event.depth_ind = line[44]
+    # new_event.loc_ind = line[45]
```
krischer: Why are those commented?

obspy/io/nordic/core.py
```diff
@@ -924,41 +1221,61 @@ def _write_nordic(event, filename, userid='OBSP', evtype='L', outdir='.',
...
-                'I'.rjust(6) + '\n')
-    # Write line 3 of s-file
+    # Write a second line if more than three magnitudes...
+    line_one_begin = (
+        ' ' + str(evtime.year) + ' ' + str(evtime.month).rjust(2) +
+        str(evtime.day).rjust(2) + ' ' + str(evtime.hour).rjust(2) +
+        str(evtime.minute).rjust(2) + ' ' + str(evtime.second).rjust(2) + '.' +
+        str(evtime.microsecond).ljust(1)[0:1] + ' ' + evtype.ljust(2) +
+        lat.rjust(7) + ' ' + lon.rjust(7) + depth.rjust(5) + agency.rjust(5) +
+        ksta.rjust(3) + timerms.rjust(4))
```
krischer: This is for one really pretty slow as each `+` generates an intermediate string object. Could this not just be done with the `.format()` method of string objects?

obspy/io/nordic/core.py
```diff
@@ -973,6 +1290,124 @@ def _write_nordic(event, filename, userid='OBSP', evtype='L', outdir='.',
         return
 
 
+def _write_moment_tensor_line(focal_mechanism):
+    """
+    Generate the two lines required for moment tensor solutions in Nordic.
+    """
+    # First line contains hypocenter info, second contains tensor info
+    lines = [list(' ' * 79 + 'M'), list(' ' * 79 + 'M')]
```
krischer: I've not seen this before but its actually kind of a nice way to get a mutable string :smile:

obspy/io/nordic/tests/test_nordic.py
calum-chamberlain: Agree - sorry, this still isn't ready - the confidence ellipsoid stuff still doesn't get the right result for the reverse (confidence_ellipsoid_to_xyz), so I haven't written more tests yet.

obspy/io/nordic/tests/test_nordic.py
calum-chamberlain: Will do once its working!

obspy/io/nordic/core.py
calum-chamberlain: It definitely could and should! Will do.

obspy/io/nordic/tests/test_nordic.py
krischer: Yea I should have read the first post 🙈 Sorry about that!

obspy/io/nordic/core.py
calum-chamberlain: Commented because I don't know where to put them in the Event format, any ideas? They are single character identifies that the user defines.
depth_ind is usually (AFAIK) one of `(' ', 'F', 'S')` where `'F'` marks fixed depth, `'S'` marks that it is a starting value (for inversion I presume).
loc_ind looks like it is either `(' ', '*')` where `*` means "do not locate".

calum-chamberlain:
Thanks @krischer I will get on those before merging (as well as actually finished the confidence ellipsoid functions).

obspy/io/nordic/core.py
krischer: The origins have a `depth_type` attribute:

screen shot 2018-01-25 at 23 08 13

But I'm not sure how that would map to the values you mentioned. There is also an `origin_type` field but that does not seem to be applicable here.

screen shot 2018-01-25 at 23 10 08


`loc_ind` make little sense to be to be honest. If there is a location it must have been located somehow.

calum-chamberlain: Hmm, I'm not sure the `depth_ind` maps well onto those either.

I think `loc_ind` is used to ensure that the location is not re-calculated, e.g. it was calculated once and the user is happy with it, or it is a known location (like a blast) and should be fixed...

krischer: I also don't really know how to map these things. I guess there just is no fully suitable mapping. The origin's also have an `epicenter_fixed` boolean field which might be suitable.

calum-chamberlain: :+1: that sounds like where it might go. Excellent, thanks!

calum-chamberlain:
I *think* I might have got it working - I think I was using the wrong form of the [rotation matrix](https://en.wikipedia.org/wiki/Euler_angles#Conversion_to_other_orientation_representations) for Tait-Bryon angles.  I was using the ZYX form (which I was led to from the [QuakeML description](https://quake.ethz.ch/quakeml/QuakeML2.0/BasicEventDescriptionTypesDiscussion#class_ConfidenceEllipsoid)), but it looks like I was going the wrong-way, so changing to XYZ appears to give the correct answers for the simple cases I have tested.

obspy/io/nordic/core.py
```diff
@@ -77,13 +77,16 @@ def _is_sfile(sfile, encoding='latin-1'):
...
+            tags = _get_line_tags(f=sfile)
+        except Exception:
+            return False
     if tags is not None:
         # Note that there can be two origin lines, but only to allow more
         # magnitudes, not more origins.
         try:
             head_line = tags['1'][0][0]
-        except KeyError:
+        except IndexError:
```
megies: without knowing details.. there's three nested `get` operations in one single try/except.. looks fragile

calum-chamberlain: Thanks for having a look @megies, but I'm sorry, I don't get this comment, could you explain a bit more (sorry!).

calum-chamberlain:
@megies and @krischer, is there anything else you would like me to change in this? From my side it looks good to go, but I haven't moved the confidence-ellipsoid functions to somewhere more general - I didn't see any functions in core.event.Event, so thought I might have been looking in the wrong place.

@megies, I didn't understand your comment on the nested `get` operations, happy to change how that is implemented though if needed?

krischer:
Sorry for the delay. Can be merged from my point of view. @megies what do you think?

obspy/io/nordic/core.py
```diff
@@ -213,94 +241,162 @@ def _nortoevmag(mag_type):
...
-        mag = inv_mag_mapping[mag_type.upper()]
+        mag = inv_mag_mapping[mag_type]
     except KeyError:
         warnings.warn(mag_type + ' is not convertible')
         return ''
     return mag
 
 
-def readheader(sfile):
+def xyz_to_confidence_ellipsoid(errors):
```
megies: Maybe this should be moved to `core.event.base`, so that it has more visibility and other people can reuse it? It could even be a staticmethod on `ConfidenceEllipsoid` to create a new such object?

megies:
I didn't look at every detail but I guess it's OK.. personally, I'd move the confidence ellipsoid routine to core.event so that it can be reused more easily..

krischer:
I agree with moving the confidence ellipsoid routine as it might be useful to have around in a more accessible place.

calum-chamberlain:
Hi all.  Moved confidence ellipsoid things and added more tests, which showed that for more complicated cases it isn't working.  I have failed to fix it, and haven't been able to get help.  I suggest that I take out the confidence ellipsoid conversions and leave this with all the other little patches to nordic.
Sorry for taking up so much time.

megies:
The confidence ellipsoid stuff would've been nice, but yeah, whatever works for you.. you're doing the work in here.. ;-)

calum-chamberlain:
I've removed the confidence ellipsoid things - I will work on them outside of a PR for a while and see where I get, but in the mean-time I think the things in here should be included in obspy.  Looks like appveyor fails are not from my things. Good to go AFAIK, sans confidenceEllipsoid conversion.

krischer:
> I think the things in here should be included in obspy

I agree - this has been ready for a while aside the confidence ellipsoid. Let's get this merged. As always: thanks a bunch!



> Looks like appveyor fails are not from my things.

Yep - this being worked on here: https://github.com/obspy/obspy/pull/2089

PR #1927, opened 2017-10-12
fix rotate to ZNE

### LINKED ISSUE #1913: Bug in rotate2zne with non-orthogonal components
obspy version: 1.0.3
installed with anaconda
According to the documentation, rotate2zne works explicitly also for non orthogonal systems. Looking into the source code, I realized, that for the basis transformation from an arbitrary system (u v w) to (z n e), the **transposed** basis vector matrix is taken. In my opinion, the **inverse** matrix must be taken.
For the transformation from one orthogonal system to another orthogonal system, the transposed matrix is equal to the inverse matrix. But in general, for non orthogonal systems, this is not the case. 
I will run some test on this issue and let you know about the results, soon.

megies:
seems it did not work correctly for non-orthogonal input data

code change by @jwassermann

fixes #1913


obspy/signal/rotate.py
```diff
@@ -124,8 +125,8 @@ def rotate_lqt_zne(l, q, t, ba, inc):
         raise ValueError("Back Azimuth should be between 0 and 360 degrees!")
     if inc < 0 or inc > 360:
         raise ValueError("Inclination should be between 0 and 360 degrees!")
-    ba *= 2 * pi / 360
-    inc *= 2 * pi / 360
+    ba = radians(ba)
+    inc = radians(inc)
```
megies: Thanks for reverting that part.. also wanted to do that but brain wasn't working anymore yesterday.. ;-)

megies:
Gonna add the test case highlighted in https://github.com/obspy/obspy/issues/1913#issuecomment-336155562 in a bit..

krischer:
So this is now one the best tested pieces of code in ObsPy given that it is only like 10 lines of logic... I cannot think of anything else to test and I would be pretty surprised if this is still wrong.

The good news is that it previously was also correct for orthogonal components which probably limits the damage this bug has done to some rare cases.

Thanks a lot to everyone who contributed!

megies:
Yeehaw. Looks good to me. All tests (old and new) pass, so this should be good. Also @fbernauer stated that he checked the code change by @jwassermann locally and it works as expected.

I think this can be merged.

megies:
Merging this now. If there's really problems left (which I doubt) then we can handle it in a follow-up.

PR #1928, opened 2017-10-13
FDSNWS: implement EIDA token authentication and add authentication to routed Clients

megies:
This is a simplistic implementation of the EIDA token authentication mechanism, see #1397 and http://geofon.gfz-potsdam.de/waveform/archive/auth/auth-example.php.

~~Didn't add the tests yet, because I don't get the expected data back, see https://github.com/obspy/obspy/pull/1397#issuecomment-336423830.~~

I think we should include it in 1.1.0, if bugs turn up in the long run we can still fix them in point releases. ~~While this doesn't add a mechanism for token auth in the routed client flavors, at least users can use their tokens when directly talking to the respective data center then.~~

+TESTS:clients.fdsn
+DOCS

CC @krischer @andres-h


krischer:
If `eida_token` is given it should check the WADL file if the `/auth` route is available - and only then use it and otherwise raise an error.

megies:
At this point the services are not yet discovered.. I would have to drop in a temporary dummy opener and headers to `self`, then discover services and then replace the dummy opener with the real opener at the end of `__init__`. I can do that if you want..

krischer:
> At this point the services are not yet discovered.. I would have to drop in a temporary dummy opener and headers to self, then discover services and then replace the dummy opener with the real opener at the end of __init__. I can do that if you want..

Any reason this can not just be moved to the end of the init method?

What would also be nice for later usage in the routing services: `Client.eida_token` as a property that automatically downloads the credentials - then this could be added after the clients have been intialized.

megies:
Right now it's just before the openers are set up using user/password auth mechanism. If I'd want to move it to the bottom I'd have to refactor the normal auth mechanism.. also doable, if you want

andres-h:
> If |eida_token| is given it should check the WADL file if the |/auth| 
> route is available - and only then use it and otherwise raise an error.
Yes, WADL should be checked, because not all nodes implement the /auth 
method (yet). If /auth is not implemented, I'd suggest to fall back to 
no auth instead of raising an error.

krischer:
> I'd suggest to fall back to no auth instead of raising an error.

When using the routing clients: yes. Otherwise we should definitely raise an error so users are aware something is wrong.

megies:
Only thing missing is tests for providing credentials to `RoutingClient`..

megies:
@krischer, actually the WADL parser currently mostly looks  up `method`s in the wadl and to determine if `auth` is supported it needs to check whether it's in  `resources`..

i guess we need a tiny method `has_eida_auth()` for that on the wadl parser..

screenshot from 2017-10-13 19-38-19

There really should be an official discussion whether this whole mechanism can be added to official FDSN WS specs, this is really quite an unofficial hijack add-on to FDSNWS.

megies:
@andres-h some things i noticed while working on this:
 - the `auth` endpoint is documented in the [dataselect wadl](https://geofon.gfz-potsdam.de/fdsnws/dataselect/1/application.wadl) but it's not listed on the dataselect landing page (see https://geofon.gfz-potsdam.de/fdsnws/dataselect/1)
 -  there's "localhost" in the wadl, I think we've seen it before (and it should've been fixed by seiscomp?) (assuming there's a seiscomp behind it, can't confirm right now as it's down currently)
 - next: it seems that the GEOFON fdsnws is down quite regularly for a short (but very noticeable) period of time, on the order of minutes  (actually it's been down for around an hour right now), see screenshot below. I've encountered this twice now, and I only made some sparse requests for testing throughout the day

One more thing @andres-h, is the `auth` endpoint exclusiv to `dataselect`?

screenshot from 2017-10-13 19-53-27

megies:
I've added passing in a per-base-url credentials dictionary to the routed client (also some basic tests for it).

Hopefully CI agrees and then thisshould be good to go. Please review @krischer .

andres-h:
>There really should be an official discussion whether this whole mechanism can be added to official
>FDSN WS specs, this is really quite an unofficial hijack add-on to FDSNWS.

Token authentication relies on trusting digital signatures, so I don't 
think this would work globally (we don't trust non-EIDA signatures and 
probably non-EIDA datacentres do not trust EIDA signatures).

We carefully designed token authentication as an add-on, rather than 
modifying existing FDSNWS standard.

In addition to providing access to restricted data, an important aspect 
is identifying users. For example, if we correct a dataset, we may want 
to contact users who have downloaded the invalid data. At least until 
full version control is implemented...

If anybody wants to download data anonymously for privacy reasons, it is 
still possible, of course.

PS. There can be other attributes in the token, like project membership, 
etc., but the client just treats the token as opaque data, so this is 
not relevant.

andres-h:
>@andres-h some things i noticed while working on this:
>
> *  the auth endpoint is documented in the dataselect wadl but it's not listed on the dataselect
>    landing page (see https://geofon.gfz-potsdam.de/fdsnws/dataselect/1)
> *  there's "localhost" in the wadl, I think we've seen it before (and it should've been fixed by
>    seiscomp?) (assuming there's a seiscomp behind it, can't confirm right now as it's down currently)

Yes, we must fix those.


> *  next: it seems that the GEOFON fdsnws is down quite regularly for a short (but very noticeable)
>    period of time, on the order of minutes (actually it's been down for around an hour right now), see
>    screenshot below. I've encountered this twice now, and I only made some sparse requests for testing
>    throughout the day

We are regularly updating the metadata. During the update (ca. 1 
minute), fdsnws is not able to serve requests, which may trigger some 
timeouts. Frustratingly, there is an unfound bug that sometimes causes 
fdsnws to hang during update, so it is restarted by watchdog after 10 
minutes.

Outages may be also caused by too many users. I think we have a limit of 
20 concurrent users, which may be too low with long-running massive 
downloads.

The 1 hour outage was caused by me stupidly forgetting to revert some 
changes that I made to debug the issue with KMBO...


>One more thing @andres-h, is the auth endpoint exclusiv to dataselect?

Currently yes. Our client does check /auth with "station", but I don't 
think this will be implemented.

Thanks for this work!

megies:
> timeouts. Frustratingly, there is an unfound bug that sometimes causes fdsnws to hang during update, so it is restarted by watchdog after 10 minutes.

> The 1 hour outage was caused [...]

Alright, that's probably what I was experiencing earlier. Thanks for the information.

megies:
> CI is green. Please review @krischer .

Gonna merge this soon, to finally be able to do the RC.

krischer:
Feel free to merge from my point of view if CI passes.

The routers credentials now work like this:

```python
c = RoutingClient(..., credentials={
    # Uses ObsPy's URL mappings + user/pw combination.
    'IRIS': {'user': 'bla', 'password': 'bla2'},
    # per fdsnws url token.
    'geofon.gfz-postdam.de': {'eida_token': '/path/to/token'},
    # Global token - will be tried for every datacenter that has the 
    # /auth route and does not have data center specific settings
    'EIDA_TOKEN': '/path/to/other/token'}
```

Currently the `/auth` route times out for every data center except the GFZ one. So this might take a while before it is an actually useful feature.

obspy/clients/fdsn/client.py
```diff
@@ -347,15 +352,19 @@ def _resolve_eida_token(self, token):
...
             raise ValueError(msg)
+
         # force https so that we don't send around tokens unsecurely
-        url = 'https://{}/fdsnws/dataselect/1/auth'.format(re.sub(
-            pattern='^http://', repl='', string=self.base_url, count=1))
+        url = 'https://{}/fdsnws/dataselect/1/auth'.format(
+            urlparse(self.base_url).netloc + urlparse(self.base_url).path)
         # paranoid: check again that we only send the token to https
-        if not url.startswith('https://'):
+        if urlparse(url).scheme != "https":
```
megies: nice.. didn't know that one. :-)

obspy/clients/fdsn/client.py
```diff
@@ -368,15 +367,12 @@ def _resolve_eida_token(self, token):
...
-            msg = ("Failed to resolve EIDA token from URL '{}', server "
-                   "replied with HTTP status code '{!s}' and message "
-                   "'{}'.").format(url, response.status_code,
-                                   response.reason)
-            raise FDSNException(msg)
-        user, password = response.content.decode().split(':')
+
+        # Already does the error checking with fdsnws semantics.
+        response = self._download(url=url, data=token.encode(),
+                                  use_gzip=True, return_string=True)
```
megies: Ah, yeah.. didn't use it earlier because this was done pretty early in `__init__` initially and `self._download` couldn't be used yet.

krischer:
> Currently the /auth route times out for every data center except the GFZ one. So this might take a while before it is an actually useful feature.

@andres-h Is this how the tokens are supposed to be used? And how are the datacenters supposed to behave?

If a global `EIDA_TOKEN` key is given as part of a routing client's credentials it will be applied to every datacenter that claims to implement the `/auth` route. But only the GFZ node does not time out.

megies:
I will merge this right now.. looks good in general with me and there's only the question left if we should use the token globally for all EIDA datacenters or not. If we're on the wrong track in that regard, we can still change it on master before the release.

obspy/clients/fdsn/client.py
```diff
@@ -198,10 +198,10 @@ def __init__(self, base_url="IRIS", major_versions=None, user=None,
         :type eida_token: str
         :param eida_token: Token for EIDA authentication mechanism, see
             http://geofon.gfz-potsdam.de/waveform/archive/auth/index.php. If a
-            token is provided, options ``user`` and ``password`` will be
-            overridden. This mechanism is only available on select EIDA nodes.
-            The token can be provided in form of the PGP message as a string,
-            or the filename of a local file with the PGP message in it.
+            token is provided, options ``user`` and ``password`` must be given.
```
megies: I think this should read `If a token is provided, options ``user`` and ``password`` must **not** be given.`

andres-h:
On 10/16/2017 10:44 AM, Lion Krischer wrote:
>     Currently the /auth route times out for every data center except the
>     GFZ one. So this might take a while before it is an actually useful
>     feature.
> 
> @andres-h <https://github.com/andres-h> Is this how the tokens are 
> supposed to be used? And how are the datacenters supposed to behave?

Looks OK to me. Any datacenter that declares /auth is supposed to handle 
it via https (I don't know if it is possible to tell in WADL that it is 
https-only). Only GFZ and RESIF seem to have opened https so far. I will 
take up this issue with EIDA.

krischer:
Thanks for the clarification. Then we will leave it as it currently is and change it if required in the future.

PR #1701, opened 2017-03-06
io.shapefile: add support for writing origin uncertainties to shapefile

megies:
This PR adds functionality to write event origin uncertainties to shapefile format (e.g. to use in GIS applications).

ToDo:
- [x] example / tutorial
- [ ] test
- [x] write uncertainties into shapefile table for reuse in GIS applications
- [ ] other information on the event would be useful to have in the shapefile database table as well, especially origin quality metrics, e.g. number of stations used in location, azimuthal gap, ...

+DOCS

Usage example:
```python
from obspy import read_events
cat = read_events('my_catalog.xml')
cat.write('/tmp/tmp_my_events.shp', format='SHAPEFILE',
          error_ellipses={'filename': '/tmp/tmp_my_events_errors.shp', 'epsg': 31468})
```

Potential improvement:
 - [x] also write origin uncertainties to feature tables (so that filtering by error margins in GIS applications is possible)
 - [ ] automatic lookup of appropriate UTM zone, based on event locations in catalog (or raise exception if catalog is too wide-spread so that a single geocentric zone will not work)

PR #1933, opened 2017-10-14
io.shapefile: replace gdal -> pyshp and add horizontal uncertainties to table

megies:
This replaces gdal in io.shapefile with pure-Python `pyshp` module. Also adds horizontal uncertainties to the shapefile table, so covers part of #1701. This will also make testing much easier, as we can just read the written shapefile with `pyshp` again and check the contents.

Will adjust the test later today or tomorrow..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,6 +129,7 @@ install:
         lxml
         mock
         nose
+        pyshp
         requests
         sqlalchemy
         "freetype<2.8"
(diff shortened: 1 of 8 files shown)


megies:
oh my god, not that one again: https://travis-ci.org/obspy/obspy/jobs/288145509#L1849

i'm starting to get nightmares from that one.

megies:
Tests pass, merging.

PR #1756, opened 2017-10-14
arclink.Client.get_inventory: useless routing request / fails for StationGroups (virtual networks)

petrrr:
Obspy version 1.0.2 (no change for module .clients.arclink documented)
Python version 3.5.3
Platform: macOX 10.10.5 (14F2315)
Package manager: MacPorts

I realized this because I tried to get an inventory for a Virtual network (Station Group in Seiscomp3 terminology). 

Apparently, the `get_inventory` method requests routing information first, before it requests the inventory itself. (1) To my understanding the routing request is useless, because in an Arclink federation (EIDA) the Inventory of all nodes should have been synchronized. If synchronization fails, it also does for the routing information. (2) For Station Groups no routing information is available so this step fails completely.

I therefore suggest to eliminate the routing request completely. It is useful only for data requests (`get_waveforms`), while `get_inventory` and `get_routing` are at the same level. 

If someone really thinks routing for `get_inventory` might be somehow useful, we should set the default to `route=False`. But honestly, I'd rather opt for complete removal because do not see any utility.

Example:
```
from obspy.clients.arclink import Client

client = Client(user="peter.danecek", institution="INGV", debug=True)
client.get_inventory(network="_NFOVRANC")
```

Error:
```
ArcLinkException: Could not find route to _NFOVRANC.*. If you think the data should be there, you might want to retry with manually connecting to a different ArcLink node (see docstring of Client) to see if there is a problem with a routing table at a specific ArcLink node (and contact the ArcLink node operators).
```

Fix:
```
client.get_inventory(network="_NFOVRANC", route=False)

### DIFF
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -1115,7 +1115,7 @@ class Client(object):
             #  (1) not network id is given,
             #  (2) virtual network (Station group) is requested
             if route:
-                msg = ("Routing was requested but parameter 'netork' is '{}' "
+                msg = ("Routing was requested but parameter 'network' is '{}' "
                        "and therefore routing is disabled.").format(network)
                 warnings.warn(msg)
             result = self._fetch(rtype, rdata, route=False)
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -1103,15 +1103,19 @@ class Client(object):
             rdata.append('lonmin=%f' % min_longitude)
         if max_longitude:
             rdata.append('lonmax=%f' % max_longitude)
+
         # fetch plain XML document
-        if network == '*':
-            # set route to False if not network id is given
+        if network == '*' or network[0] == '_':
+            # To not use routing if
+            #  (1) not network id is given,
+            #  (2) virtual network (Station group) is requested
             result = self._fetch(rtype, rdata, route=False)
         else:
             result = self._fetch(rtype, rdata, route=route)
+
         # parse XML document
         xml_doc = etree.fromstring(result)
-        # get routing version
+        # get inventory version
         if _INVENTORY_NS_1_0 in xml_doc.nsmap.values():
             xml_ns = _INVENTORY_NS_1_0
             stream_ns = 'sensorLocation'
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -1114,6 +1114,10 @@ class Client(object):
             # To not use routing if
             #  (1) not network id is given,
             #  (2) virtual network (Station group) is requested
+            if route:
+                msg = ("Routing was requested but parameter 'netork' is '{}' "
+                       "and therefore routing is disabled.").format(network)
+                warnings.warn(msg)
             result = self._fetch(rtype, rdata, route=False)
         else:
             result = self._fetch(rtype, rdata, route=route)
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -450,7 +450,7 @@ class Client(object):
         # fetching PAZ and coordinates
         if metadata:
             # fetch metadata only once
-            # NOTE: The outing step here is not strictly necessary
+            # NOTE: The routing step here is not strictly necessary
             # TODO: Routing should preferably be done only once
             inv = self.get_inventory(network=network, station=station,
                                      location=location, channel=channel,
--- a/obspy/clients/arclink/client.py
+++ b/obspy/clients/arclink/client.py
@@ -210,7 +210,7 @@ class Client(object):
         self._write_ln('BYE')
         self._client.close()
 
-    def _fetch(self, request_type, request_data, route=True):
+    def _fetch(self, request_type, request_data, route=False):
         # skip routing on request
         if not route:
             # always use initial node if routing is disabled
@@ -450,6 +450,8 @@ class Client(object):
         # fetching PAZ and coordinates
         if metadata:
             # fetch metadata only once
+            # NOTE: The outing step here is not strictly necessary
+            # TODO: Routing should preferably be done only once
             inv = self.get_inventory(network=network, station=station,
                                      location=location, channel=channel,
                                      starttime=starttime, endtime=endtime,
@@ -616,6 +618,7 @@ class Client(object):
         # request data
         rdata = [starttime, endtime, network, station]
         # fetch plain XML document
+        # TODO: A simpler "routing-less" methods, check _request
         result = self._fetch(rtype, rdata, route=False)
         # parse XML document
         xml_doc = etree.fromstring(result)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by petrrr** [comment]
Okay, fixed this problem the test `test_get_inventory` fails on case 5:
https://github.com/obspy/obspy/blob/26a985d5bf04b440a77809a0f76594e279c00880/obspy/clients/arclink/tests/test_client.py#L180-L182

The expected Exception however occurs in the routing step, so IMHO it should not be tested here. I (re)move this case, see if it is need in the routing test.

> ArcLinkException: Could not find route to 00.. If you think the data should be there, you might want to retry with manually connecting to a different ArcLink node (see docstring of Client) to see if there is a problem with a routing table at a specific ArcLink node (and contact the ArcLink node operators).

**Comment by petrrr** [comment]
I am looking a bit deeper in the logic of this module. It is probably not the most readable and linear piece of code in ObsPy. On the other, hand we probably do not want to vast to much energy on refactoring, as we hope all that Arclink will go away sooner than later.

Still, I'd like to propose few changes along with this PR. In particular, would like to avoid the routing step where it is not necessary. The Inventory and the routes should be synchronized within the federation and there available locally at each node. In principle it may be true that some metadata might be slightly outdated, but I guess this is a rather minor issue. If synchronization fails the system breaks in a much worse way.

**Comment by megies** [comment]
@petrrr If you think the default for routing in get_stations should be changed to `False` I would accept that change. I guess you're knowing arclink much more than me or most other devs. @krischer any complaints to merging this for 1.1.0?

**Comment by petrrr** [comment]
> @petrrr If you think the default for routing in get_stations should be changed to False I would accept that change. I guess you're knowing arclink much more than me or most other devs.

Not sure if I really know it, but had to deal with all these nasty metadata details. Andres is the guru. Anyway, routing for Inventory requests is kind of redundant, because these metadata are supposed to be already present locally. Anyway, the possibility to force routing remains. Should be mostly transparent to users, except **very** special cases.

> @krischer any complaints to merging this for 1.1.0?

Would be great!

* First, the possibility to deal with Virtual nets is becoming quite relevant for AlpArray, which still depends on this protocol. 
* Second, it would also make somewhat more sense for full release, as the behavior changes slightly.

But of cause I'd admit that this comes in rather late.

**Comment by megies** [comment]
> But of cause I'd admit that this comes in rather late.

Doesn't matter for me, it's an overseeable change and if it makes sense, then it should go in. Oh ,but please add a small changelog entry, since this is changing default behavior (although hopefully to the better).

CCing @andres-h, would be great if you could comment on this proposed change.

**Comment by andres-h** [comment]
Although route=True is the most general case, I agree with @petrrr that 
(Arclink) routing should be false for metadata requests by default.

Note that this is not the case with FDSNWS, where metadata requests must 
be routed and virtual nets are resolved by the routing service (by the 
EIDA routing service, at least -- don't know about the IRIS Federator).

**Comment by megies** [comment]
> Note that this is not the case with FDSNWS, where metadata requests must

Yep, this PR is exclusively about our arclink client. Thanks for the feedback.

File: obspy/clients/arclink/client.py [hunk 1]
```diff
@@ -210,7 +210,7 @@ def _bye(self):
         self._write_ln('BYE')
         self._client.close()
 
-    def _fetch(self, request_type, request_data, route=True):
+    def _fetch(self, request_type, request_data, route=False):
```
megies: Maybe this one should stay `route=True` as a safe default? We can still manually use `route=False` in all metadata requests below, like you propose.

obspy/clients/arclink/client.py
```diff
@@ -450,6 +450,8 @@ def get_waveforms(self, network, station, location, channel, starttime,
         # fetching PAZ and coordinates
         if metadata:
             # fetch metadata only once
+            # NOTE: The outing step here is not strictly necessary
```
megies: outing -> routing

megies:
@petrrr Two minor comments, otherwise I'd be fine with merging this for 1.1.0. Please also add a short changelog entry, it's changing behavior slightly after all.

megies:
@petrrr, I will now merge this and implement my minor comments on master directly. In the future it would be nice if you push to a branch in the main repo when working on things with a high probability of getting merged, so it's easier for other devs to add commits.

megies:
+TESTS:clients.arclink

petrrr:
> @petrrr, I will now merge this and implement my minor comments on master directly. In the future it would be nice if you push to a branch in the main repo when working on things with a high probability of getting merged, so it's easier for other devs to add commits.

Sorry for the complication. I was not too sure about this, especially because it was not evident to me that this can be included into the release.

obspy/clients/arclink/client.py
```diff
@@ -1114,6 +1114,10 @@ def get_inventory(self, network, station='*', location='*', channel='*',
             # To not use routing if
             #  (1) not network id is given,
             #  (2) virtual network (Station group) is requested
+            if route:
+                msg = ("Routing was requested but parameter 'netork' is '{}' "
```
petrrr: There is a small typo! 
Is this something which can be fixed directly in master?

megies:
> Sorry for the complication. I was not too sure about this, especially because it was not evident to me that this can be included into the release.

Ah, no worries.. actually it's not a big deal to add another remote.. just clutters the local git repo a bit, but it can be removed later. So, no problem!

obspy/clients/arclink/client.py
megies: Yep. Just fix it in master directly. 👍

obspy/clients/arclink/client.py
petrrr: @megies: Sorry for following up late on this. Just a note for the record.

I had/have not a too strong opinion on this, especially because in practice this really should not have any relevance. All calls now specify the parameter explicitly. 

However, there was some reasoning on changing `route=False`, because `route=True` implies recursion. I judged that for this lower level routine `route=False` was safer wrt. to avoiding infinite recursions. 

Now, propose not to touch it anymore. I may look into this part at some point again to get it cleaner and more linear.

QuLogic:
Strictly speaking, you do not need another remote. To just download the code use `git fetch upstream pull/1756/head:fix_arclink`. To push, you can specify the remote url without adding it `git push git@github.com:petrr/obspy fix_arclink`.

I also like to set an `insteadOf` config like:
```
$ git config --list | grep instead
url.git://github.com/.insteadof=github:
url.git@github.com:.pushinsteadof=github:
```
so that I can do `git push github:petrr/obspy fix_arclink`.

megies:
Slick. Now I only need to remember this piece.. 😊

obspy/clients/arclink/client.py
petrrr: Nevermind! You already corrected this in a later commit.

megies:
Added your info [to the wiki](https://github.com/obspy/obspy/wiki/Interesting-Reading-on-Git,-GitHub) thanks again!

PR #1935, opened 2017-10-14
fix #1934 header won't be changed in Trace.init

### LINKED ISSUE #1934: trace.Trace should not change the header
Basic info
  -  ObsPy version 1.0.3
  - Python version: 3.6
  - Platform: Windows 10

I found a problem when I was using trace.Trace to initialize a trace object. Codes are listed bellow

```python
    from obspy.core import trace
    import numpy as np
    data = [0] * 20
    header = {
        'station': 'MS',
        'starttime': None
    }
    header['starttime'] = 1
    my_trace1 = trace.Trace(data=np.array(data), header=header)
    print(my_trace1)
    data = [1] * 60
    my_trace2= trace.Trace(data=np.array(data), header=header)
    print(my_trace2)
    print(header)
```
Then I got:

```
.MS.. | 1970-01-01T00:00:01.000000Z - 1970-01-01T00:00:20.000000Z | 1.0 Hz, 20 samples
.MS.. | 1970-01-01T00:00:01.000000Z - 1970-01-01T00:00:20.000000Z | 1.0 Hz, 20 samples
{'station': 'MS', 'starttime': 1, 'npts': 20}
```

I used this script to initialize two sets of traces. What I expected of `my_trace2` is a trace of 60 samples, but I got 20. I browsed the source code of trace and found a possible reason. When Trace is initialized, the `header` would be set a default key `npts`, therefore, the sample number of `my_trace2` is not correct when I use `header` the second time.

```python
    def __init__(self, data=np.array([]), header=None):
        # make sure Trace gets initialized with suitable ndarray as self.data
        # otherwise we could end up with e.g. a list object in self.data
        _data_sanity_checks(data)
        # set some defaults if not set yet
        if header is None:
            header = {}
        header.setdefault('npts', len(data))
        self.stats = Stats(header)
        # set data without changing npts in stats object (for headonly option)
        super(Trace, self).__setattr__('data', data)
```

In my opinion, an appropriate solution is to insert a line `header = header.copy()` before `header.setdefault('npts', len(data))`.

Harlan-pku:
### What does this PR do?
fix #1934. The header won't be change during Trace.__init__
Issue #1934 
### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -70,3 +70,4 @@ Andrew Walker
 Marcus Walther
 Joachim Wassermann
 Mark C. Williams
+Li, Yulin
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -272,6 +272,7 @@ class Trace(object):
         # set some defaults if not set yet
         if header is None:
             header = {}
+        header = header.copy()
         header.setdefault('npts', len(data))
         self.stats = Stats(header)
         # set data without changing npts in stats object (for headonly option)


megies:
This should go into `master` branch as we discontinued 1.0.x, I'll change the base branch for this PR. But change looks good to me. Thanks for reporting and fixing @Harlan-pku!

megies:
Looks like there's a merge conflict due to the change of base branch.. I'll open a new PR for this and will cherry-pick your commit in there. I think that's easiest and fastest. I'll also add you to developers, so you can push to branches in the main repo, that's usually easier for collaboration (unless it's about experimental features that might not be merged for a long time, in which case it's better to push to your fork to keep number of branches in main repo low)

Harlan-pku:
I thought it was a buf for 1.0.3 so I chose the maintenance branch as base branch. Thank you for your timely reply and work.

Harlan-pku:
Hi @megies, I think I was a little cofilicted by the page of  [ObsPy Git Branching Model](https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model). In that page, a bugfix PR should be started from the current maintenance branch. The latest maintenance branch I found is `maintennace_1.0.x`, so I chose that branch as base branch of this PR. Maybe some instructions should be added to the  [Branching Model Page](https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model) so it will be more clearly for develop beginners like me :)

megies:
Yeah.. in principle you did the exact right thing by opening this PR against the maintenance branch.. the special situation right now is just that we decided to not do any more bug fix releases for 1.0.x since we originally intended 1.1.0 to be out moths ago already.. sorry about the confusion and thanks again for reporting and fixing this! :-)

PR #1936, opened 2017-10-14
Avoid clashes in Trace header init: deepcopy provided dictionary

### LINKED ISSUE #1934: trace.Trace should not change the header

megies:
Supersedes #1935 and fixes #1934.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -45,6 +45,7 @@ Lecocq, Thomas
 Leeman, John
 Legovini, Paride
 Lesage, Philippe
+Li, Yulin
 Lomax, Anthony
 Lopes, Rui L.
 MacCarthy, Jonathan
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2700,8 +2700,14 @@ class TraceTestCase(unittest.TestCase):
         dictionary for multiple Trace inits)
         """
         header = {'station': 'MS', 'starttime': 1}
+        original_header = deepcopy(header)
+
+        # Init two traces and make sure the original header did not change.
         tr1 = Trace(data=np.ones(2), header=header)
+        self.assertEqual(header, original_header)
         tr2 = Trace(data=np.zeros(5), header=header)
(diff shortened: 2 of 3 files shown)


megies:
I'll merge this tomorrow if nobody opposes.

megies:
Looks good. Merging.

PR #1939, opened 2017-10-17
Fix warnings when building extensions / Update libmseed to 2.19.5

QuLogic:
Various small fixes within `evalresp`, plus updating `libmseed` to 2.19.5, which is mostly a bugfix release for warnings as well.

There is one small issue though...


obspy/signal/src/evalresp/calc_fctns.c
```diff
@@ -526,8 +526,9 @@ double r, i;
 *=================================================================*/
 void norm_resp(struct channel *chan, int start_stage, int stop_stage,
                int hide_sensitivity_mismatch_warning) {
+  // TODO - NULL assignments below made blindly to fix compiler warning.  bug?
```
QuLogic: Backported from `evalresp` 4.

obspy/signal/src/evalresp/parse_fctns.c
```diff
@@ -1040,11 +1040,11 @@ void parse_ref(FILE *fptr, struct blkt *blkt_ptr, struct stage *stage_ptr) {
                    (i.e. that the next non-comment line starts the actual response information) */
 
 int parse_channel(FILE *fptr, struct channel* chan) {
-  int blkt_no, read_blkt = 0, fld_no, no_units;
+  // TODO - assigments for no_units and tmp_stage2 made blindly to fix compiler warning.  bug?
```
QuLogic: Also backported.

obspy/signal/src/evalresp/resp_fctns.c
```diff
@@ -219,6 +220,7 @@ void check_channel(struct channel *chan) {
     j = 0;
     next_stage = stage_ptr->next_stage;
     stage_type = gain_flag = deci_flag = ref_flag = 0;
+    filt_blkt = deci_blkt = gain_blkt = ref_blkt = NULL;
```
QuLogic: I reset the pointers here because the flags are reset here...

obspy/signal/src/evalresp/resp_fctns.c
```diff
@@ -401,25 +403,57 @@ void check_channel(struct channel *chan) {
       if(ref_flag && deci_flag) {
         stage_ptr->first_blkt = ref_blkt;
         ref_blkt->next_blkt = filt_blkt;
+        if (filt_blkt == NULL) {
```
QuLogic: ... and added these checks to be sure there were no `NULL`-dereferences, but it seems to break `test_response_54_without_58`. It doesn't make sense to me that blockettes should be re-used for stages, but maybe I misunderstood something.

megies:
There's a failing test in Travis.. @QuLogic   CC @krischer

krischer:
I just removed this test. I stared at it way too long and this is one of those things that might work on evalresp with certain compilers by chance - in this case due to pointers being shared between responses... This makes no sense and test also previously failed on Windows.

krischer:
Maybe add the libmseed update and partial evalresp 4 backport to the changelog?

PR #1943, opened 2017-10-18
image test uploads to imgur: make it possible to do full authentication

megies:
This adds possibility to set imgur client secret and imgur refresh token in env variables to get higher `X-RateLimit-UserLimit` when trying to upload to imgur.


megies:
CC @petrrr

megies:
No harm in merging this.

PR #1946, opened 2017-10-19
read ISF ISM1.0 bulletin event format

megies:
See http://www.isc.ac.uk/standards/isf/


shicks-seismo:
Good that you're working on this, but since ISC bulletin searches output QuakeML, do we really need an ISF reader? Maybe it is good for legacy catalogues...?

megies:
Believe me, this is not a fun project.. ;-)
I'm going to a workshop, and other lecturers there use a lot of legacy ISF catalog files in their tutorials. It'd just be so embarrassing to talk about how convenient obspy is and then not be able to read 90% of the files that were handed to users as example data.

krischer:
One of our raisons d`être is also to be able to read as many seismological file formats as possible so I'd always vote for adding more!

megies:
Current implementation is slow..  but so what, this is just to support some legacy format, not something that's supposed to be a work horse for number crunching.

Also, extensions to the standard are not covered fully yet, see "ISF Extensions": http://www.isc.ac.uk/standards/isf/#PB.
If I get a nod from CI I'd still vote for merging into master tomorrow. It can still be improved later on, but would be nice to have in 1.1.0.

megies:
I've run the current state of this PR through 200 real world test files locally, amounting to 10MB of ascii, all with multiple magnitudes and origins and some with several thousand phases in them. So I'm pretty confident this PR works.

If there's any more problems with this, we can still address it in bug fixes. And if somebody really wants to spend time in speeding this up, this can also be done later.

krischer:
I changed a couple of minor things and read the rest. Should IMHO be good to go as a first version - any issues that will undoubtedly show up once people feed it slightly different data can be fixed once we encounter them.

megies:
Circle CI seems to use a new flake8 that has some new warnings e.g. about bare try/except.. guess we can work on those after 1.1.0

obspy/io/iaspei/tests/test_core.py
```diff
@@ -28,17 +28,14 @@ def _assert_catalog(got):
     data, num_subs = re.subn(expected_id_prefix, got_id_prefix, data)
     # 49 resource id replacements should be done in the QuakeML file we compare
     # against
-    if num_subs != 49:
-        raise AssertionError()
+    assert num_subs == 49
```
megies: So `assert` is OK for test code I guess?

megies:
@krischer tests are failing now

obspy/io/iaspei/tests/test_core.py
krischer: Yea I would say so. `pytest` bases everything on top of assert statements and I really know nobody who runs python with `-O`.

megies: Alright.. just thought I'd remembered something about better avoiding it.. (from you ;-P but probably just imagining things ;-))

krischer:
CI is "green". Merging.

PR #1947, opened 2017-10-20
Remove clients.seishub doctests

krischer:
They currently always cause `obspy-runtests --all` to fail (and they cannot be skipped like the normal seishub unit tests).

Two possible alternatives:

* Move `obspy.clients.seishub` to a separate package and remove it from ObsPy itsself. I feel like this might be an acceptable solution.
* Find a way to conditionally skip the doctests.


krischer:
This is now a PR that just `+SKIP`s all `obspy.clients.seishub` doctests. I checked and everything is still covered in the unit tests. As both the server and client are stable the documentation should also be stable so I feel like this is an acceptable solution.

megies:
OK for me.

PR #1948, opened 2017-10-20
Compatibility for old requests and lxml

krischer:
This PR fixes a couple of problems with very old versions of the `requests` and `lxml` packages. Mainly Debian Wheezy is affected.


megies:
+TESTS:clients.fdsn

will retrigger Travis to see FDSN results at least on recent systems

PR #1950, opened 2017-10-21
fix docker Linux32 conda package builder

megies:
One problem seems to be current Miniconda3, it ends up in GLIBC version errors. But even with this change, inside the build container, conda-build errors out.


krischer:
I'm removing this from the milestone - this is purely a packaging aspect and I don't want it to hold off 1.1.0 - we can figure out the best way to do this after the release as it is a bit time intensive.

megies:
Build seems to work again, locally, after some minor changes.

For future reference, there's also this: https://github.com/ContinuumIO/docker-images/tree/master/conda_builder_linux

It has some additional settings for environment variables and cflags, so if we run into problems at some point, we might wanna have a look there as well (although it doesn't look like it's used anymore).

PR #1959, opened 2017-10-23
Updating to latest flake8

krischer:
Hopefully fixes the CircleCI failures.

There are a bunch of cases where I feel like we could use more specific exceptions but I'm not too comfortable doing this right before a release.


krischer:
Seems to fix circleci. Please don't merge until the other CI passed and it also changes some pieces of logic.

PR #1961, opened 2017-10-23
try to fix regex check for future imports, avoid decoding and use byte strings

megies:


krischer:
Superseded by #1963 I think.

PR #1949, opened 2017-10-20
Enable means of creating response object from pole/zero list

amkearns-usgs:
Thank your for contributing to ObsPy!

Before submitting a PR, please review the pull request guidelines:
https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Also, please make sure you are following the ObsPy branching model:
https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model

### What does this PR do?

Adds a means of creating a response object from lists of poles and zeros, using default parameters as appropriate. (Feature request)

### Why was it initiated?  Any relevant Issues?

This was initiated from discussion in issue #1824.

### PR Checklist
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.
- [*] Any new features or fixed regressions are be covered via new tests.
- [*] Any new or changed features have are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [*] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #1962, opened 2017-10-23
Create response files from Poles and Zeros lists

amkearns-usgs:
### What does this PR do?

Enables the ability to construct response objects from pole/zero lists, along with a series of optional parameters with appropriate defaults chosen.

Yes. See issue #1949.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1891,17 +1891,11 @@ class Response(ComparingObject):
         return paz_to_sacpz_string(paz, self.instrument_sensitivity)
 
     @staticmethod
-    def from_paz(zeros, poles, stage_sequence_number=1, stage_gain=1,
-                 stage_gain_frequency=1, input_units='COUNTS',
-                 output_units='M/S', normalization_frequency=1,
+    def from_paz(zeros, poles, stage_gain,
+                 stage_gain_frequency=1.0, input_units='VOLTS',
+                 output_units='M/S', normalization_frequency=None,
                  pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
-                 normalization_factor=1.0, resource_id=None, resource_id2=None,
-                 name=None, input_units_description=None,
-                 output_units_description=None, description=None,
-                 decimation_input_sample_rate=None, decimation_factor=None,
-                 decimation_offset=None, decimation_delay=None,
-                 decimation_correction=None, response_id=None,
-                 response_sensitivity=None, response_polynomial=None):
+                 normalization_factor=1.0):
         """
         Takes in lists of complex poles and zeros and returns a Response with
         those values defining its only stage.
@@ -1909,34 +1903,26 @@ class Response(ComparingObject):
         :param zeros: All zeros of the response to be defined.
         :type poles: list of complex
         :param poles: All poles of the response to be defined.
+        :type stage_gain: double
+        :param stage_gain: The gain value of the response [sensitivity]
 
         Most of the optional parameters defined here are from
         :class:`~obspy.core.inventory.response.PolesZerosResponseStage`
-        The exceptions are found in
-        :class:`~obspy.core.inventory.response.Response`
-        :type response_id: str
-        :param response_id: alias for returned response's resource_id parameter
-        :type response_sensitivity:
-            :class:`~obspy.core.inventory.response.InstrumentSensitivity`
-        :param response_sensitivity: alias for response's
-            instrument_sensitivity parameter
-        :type response_polynomial:
-            :class:`~obspy.core.inventory.response.InstrumentPolynomial`
-        :param response_polynomial: alias for response's instrument_polynomial
+        Note that if no value is specified for the normalization_frequency
+        parameter, the value assigned to stage_gain_frequency will be used.
         :returns: new Response instance with given P-Z values
         """
+        if normalization_frequency is None:
+            normalization_frequency = stage_gain_frequency
+        sequence = 1  # must be stage 1 because paz defines entire response
         pzstage = PolesZerosResponseStage(
-                    stage_sequence_number, stage_gain,
+                    sequence, stage_gain,
                     stage_gain_frequency, input_units, output_units,
                     pz_transfer_function_type, normalization_frequency, zeros,
-                    poles, normalization_factor, resource_id, resource_id2,
-                    name, input_units_description, output_units_description,
-                    description, decimation_input_sample_rate,
-                    decimation_factor, decimation_offset, decimation_delay,
-                    decimation_correction)
-        resp = Response(resource_id=response_id,
-                        instrument_sensitivity=response_sensitivity,
-                        instrument_polynomial=response_polynomial,
+                    poles, normalization_factor)
+        sensitivity = InstrumentSensitivity(stage_gain, stage_gain_frequency,
+                                            input_units, output_units)
+        resp = Response(instrument_sensitivity=sensitivity,
                         response_stages=[pzstage])
         return resp
 
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1890,6 +1890,56 @@ class Response(ComparingObject):
         paz = self.get_paz()
         return paz_to_sacpz_string(paz, self.instrument_sensitivity)
 
+    @staticmethod
+    def from_paz(zeros, poles, stage_sequence_number=1, stage_gain=1,
+                 stage_gain_frequency=1, input_units='COUNTS',
+                 output_units='M/S', normalization_frequency=1,
+                 pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
+                 normalization_factor=1.0, resource_id=None, resource_id2=None,
+                 name=None, input_units_description=None,
+                 output_units_description=None, description=None,
+                 decimation_input_sample_rate=None, decimation_factor=None,
+                 decimation_offset=None, decimation_delay=None,
+                 decimation_correction=None, response_id=None,
+                 response_sensitivity=None, response_polynomial=None):
+        """
+        Takes in lists of complex poles and zeros and returns a Response with
+        those values defining its only stage.
+        :type zeros: list of complex
+        :param zeros: All zeros of the response to be defined.
+        :type poles: list of complex
+        :param poles: All poles of the response to be defined.
+
+        Most of the optional parameters defined here are from
+        :class:`~obspy.core.inventory.response.PolesZerosResponseStage`
+        The exceptions are found in
+        :class:`~obspy.core.inventory.response.Response`
+        :type response_id: str
+        :param response_id: alias for returned response's resource_id parameter
+        :type response_sensitivity:
+            :class:`~obspy.core.inventory.response.InstrumentSensitivity`
+        :param response_sensitivity: alias for response's
+            instrument_sensitivity parameter
+        :type response_polynomial:
+            :class:`~obspy.core.inventory.response.InstrumentPolynomial`
+        :param response_polynomial: alias for response's instrument_polynomial
+        :returns: new Response instance with given P-Z values
+        """
+        pzstage = PolesZerosResponseStage(
+                    stage_sequence_number, stage_gain,
+                    stage_gain_frequency, input_units, output_units,
+                    pz_transfer_function_type, normalization_frequency, zeros,
+                    poles, normalization_factor, resource_id, resource_id2,
+                    name, input_units_description, output_units_description,
+                    description, decimation_input_sample_rate,
+                    decimation_factor, decimation_offset, decimation_delay,
+                    decimation_correction)
+        resp = Response(resource_id=response_id,
+                        instrument_sensitivity=response_sensitivity,
+                        instrument_polynomial=response_polynomial,
+                        response_stages=[pzstage])
+        return resp
+
 
 def paz_to_sacpz_string(paz, instrument_sensitivity):
     """
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1897,22 +1897,21 @@ class Response(ComparingObject):
                  pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
                  normalization_factor=1.0):
         """
-        Convert poles and response lists into a one-stage response.
+        Convert poles and zeros lists into a single-stage response.
 
         Takes in lists of complex poles and zeros and returns a Response with
-        those values defining its only stage.
+        those values defining its only stage. Most of the optional parameters
+        defined here are from
+        :class:`~obspy.core.inventory.response.PolesZerosResponseStage`.
+
         :type zeros: list of complex
         :param zeros: All zeros of the response to be defined.
         :type poles: list of complex
         :param poles: All poles of the response to be defined.
         :type stage_gain: float
         :param stage_gain: The gain value of the response [sensitivity]
-
-        Most of the optional parameters defined here are from
-        :class:`~obspy.core.inventory.response.PolesZerosResponseStage`.
         :returns: new Response instance with given P-Z values
         """
-
         pzstage = PolesZerosResponseStage(
                     stage_sequence_number=1, stage_gain=stage_gain,
                     stage_gain_frequency=stage_gain_frequency,
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1890,8 +1890,8 @@ class Response(ComparingObject):
         paz = self.get_paz()
         return paz_to_sacpz_string(paz, self.instrument_sensitivity)
 
-    @staticmethod
-    def from_paz(zeros, poles, stage_gain,
+    @classmethod
+    def from_paz(cls, zeros, poles, stage_gain,
                  stage_gain_frequency=1.0, input_units='M/S',
                  output_units='VOLTS', normalization_frequency=1.0,
                  pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
@@ -1924,8 +1924,7 @@ class Response(ComparingObject):
                                      frequency=stage_gain_frequency,
                                      input_units=input_units,
                                      output_units=output_units)
-        resp = Response(instrument_sensitivity=sens,
-                        response_stages=[pzstage])
+        resp = cls(instrument_sensitivity=sens, response_stages=[pzstage])
         resp.recalculate_overall_sensitivity()
         return resp
 
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1924,6 +1924,7 @@ class Response(ComparingObject):
                                             input_units, output_units)
         resp = Response(instrument_sensitivity=sensitivity,
                         response_stages=[pzstage])
+        resp.recalculate_overall_sensitivity()
         return resp
 
 
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1892,37 +1892,40 @@ class Response(ComparingObject):
 
     @staticmethod
     def from_paz(zeros, poles, stage_gain,
-                 stage_gain_frequency=1.0, input_units='VOLTS',
-                 output_units='M/S', normalization_frequency=None,
+                 stage_gain_frequency=1.0, input_units='M/S',
+                 output_units='VOLTS', normalization_frequency=1.0,
                  pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
                  normalization_factor=1.0):
         """
+        Convert poles and response lists into a one-stage response.
+
         Takes in lists of complex poles and zeros and returns a Response with
         those values defining its only stage.
         :type zeros: list of complex
         :param zeros: All zeros of the response to be defined.
         :type poles: list of complex
         :param poles: All poles of the response to be defined.
-        :type stage_gain: double
+        :type stage_gain: float
         :param stage_gain: The gain value of the response [sensitivity]
 
         Most of the optional parameters defined here are from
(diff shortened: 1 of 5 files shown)


krischer:
Something is still off. Would you mind if I force push to your branch to fix it?

amkearns-usgs:
That would be fine. Apparently I did my last commit on autopilot and didn't
think about the weirdness going on with whatever must have changed in
libmseed.

*Aaron Kearns*

KBRwyle  |  Software Engineer
Target Road 10002 Isleta SE  |  Kirtland AFB, NM 87117  |  USA
Office: +1 505.853.2582  |  Mobile: +1 304.997.0148  |  akearns@usgs.gov  |
 aaron.kearns@kbrwyle <aaron.kearns@kbrwyle.com>.com
<aaron.kearns@honeywell.com>

krischer:
Fixed now.

amkearns-usgs:
Thanks. Sorry for the headaches.

*Aaron Kearns*

KBRwyle  |  Software Engineer
Target Road 10002 Isleta SE  |  Kirtland AFB, NM 87117  |  USA
Office: +1 505.853.2582  |  Mobile: +1 304.997.0148  |  akearns@usgs.gov  |
 aaron.kearns@kbrwyle <aaron.kearns@kbrwyle.com>.com
<aaron.kearns@honeywell.com>

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,56 @@ def get_sacpz(self):
...
+                    stage_sequence_number, stage_gain,
+                    stage_gain_frequency, input_units, output_units,
+                    pz_transfer_function_type, normalization_frequency, zeros,
+                    poles, normalization_factor, resource_id, resource_id2,
+                    name, input_units_description, output_units_description,
+                    description, decimation_input_sample_rate,
+                    decimation_factor, decimation_offset, decimation_delay,
+                    decimation_correction)
+        resp = Response(resource_id=response_id,
+                        instrument_sensitivity=response_sensitivity,
```
krischer: This must be an `InstrumentSensitivity` object which can be constructed given a sensitivity value and a frequency.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,56 @@ def get_sacpz(self):
...
+                        instrument_polynomial=response_polynomial,
```
krischer: Same here but this simple function should IMHO just not allow settings the polynomial as this a lot more special purpose.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,20 @@ def test_response_with_no_units_in_stage_1(self):
             out, [0 + 9869.2911771081963j, 0 + 19738.582354216393j,
                   0 + 39477.164708432785j])
 
+    def construct_response_from_paz_values(self):
+        poles = [1+2j, 1-2j, 2+2j, 2-2j]
+        zeros = [0, 0, 5]
+        resp = Response.from_paz(zeros, poles)
```
krischer: Please add a bit more to this test, e.g. make some assertions about the response object and maybe also calculate the response and compare with a response from some known-good source.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,20 @@ def test_response_with_no_units_in_stage_1(self):
...
+
+    def construct_new_response_paz(self):
+        zeros = [0., 0., -9., -160.7, -3108.]
+        poles = [-0.03852+0.03658j, -0.03852-0.03658j, -178., -135.+160.j,
+                -135.-160.j, -671.+1514.j, -671.-1514.j]
+        gain = 3.08*(10**5)
+        sensitivity = 1201.*(2**26/40.)
+        resp = Response.from_paz(zeros, poles, stage_gain=gain,
+                                 response_sensitivity=sensitivity)
```
krischer: Same comment as above.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,56 @@ def get_sacpz(self):
         paz = self.get_paz()
         return paz_to_sacpz_string(paz, self.instrument_sensitivity)
 
+    @staticmethod
+    def from_paz(zeros, poles, stage_sequence_number=1, stage_gain=1,
```
krischer: The gain should really be a required parameter.

krischer: And the stage sequence number should IMHO not as this will return a complete response object and you already know that it will be stage 1.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,56 @@ def get_sacpz(self):
...
+                 stage_gain_frequency=1, input_units='COUNTS',
+                 output_units='M/S', normalization_frequency=1,
+                 pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
+                 normalization_factor=1.0, resource_id=None, resource_id2=None,
+                 name=None, input_units_description=None,
+                 output_units_description=None, description=None,
+                 decimation_input_sample_rate=None, decimation_factor=None,
```
krischer: in my opinion: No need to set these - this is really just a helper function to construct the simplemost PAZ response.

amkearns-usgs:
I believe I have made the relevant changes. Please let me know if there are any more things you need.

krischer:
There is still something wrong with your `git`. Does it maybe somehow autoreplace symlinks or something? In any case - not important right now - I can fix it before merging.

This still lacks a test in terms of the actual response being calculated, e.g. something like

```python
resp = Response.from_paz(zeros, poles)
r = get_evalresp_response_for_frequencies(frequencies, output="VEL")
np.testing.assert_all_close(r, known_good_r_from_some_other_source)
```

This would then also test that all the normalization constants and what not are set correctly.

megies:
@amkearns-usgs you commited some files by accident, we can handle it later, but you might want to give `$ git add -p` a try and add ne files manually with `$ git add <filename>` and avoid using `$ git commit --all` in the future.

amkearns-usgs:
Thanks. (Apologies, silly force of habit on my part.)

*Aaron Kearns*

KBRwyle  |  Software Engineer
Target Road 10002 Isleta SE  |  Kirtland AFB, NM 87117  |  USA
Office: +1 505.853.2582  |  Mobile: +1 304.997.0148  |  akearns@usgs.gov  |
 aaron.kearns@kbrwyle <aaron.kearns@kbrwyle.com>.com
<aaron.kearns@honeywell.com>

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
 
+    @staticmethod
+    def from_paz(zeros, poles, stage_gain,
+                 stage_gain_frequency=1.0, input_units='VOLTS',
+                 output_units='M/S', normalization_frequency=None,
+                 pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
+                 normalization_factor=1.0):
+        """
+        Takes in lists of complex poles and zeros and returns a Response with
+        those values defining its only stage.
```
krischer: If possible, please rephrase the docstring into
```python
"""
Single line description
   <- Empty Line
Longer description.

...
"""
```

Then it will render nicer in the sphinx documentation.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
+        :type zeros: list of complex
+        :param zeros: All zeros of the response to be defined.
+        :type poles: list of complex
+        :param poles: All poles of the response to be defined.
+        :type stage_gain: double
```
krischer: All floating point numbers are double precision in Python - just call it `float`.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
+        :param stage_gain: The gain value of the response [sensitivity]
+
+        Most of the optional parameters defined here are from
+        :class:`~obspy.core.inventory.response.PolesZerosResponseStage`
+        Note that if no value is specified for the normalization_frequency
+        parameter, the value assigned to stage_gain_frequency will be used.
+        :returns: new Response instance with given P-Z values
+        """
+        if normalization_frequency is None:
```
krischer: This is very unlikely to happen it defaults to 1.0 so people would have to explicitly set it to `None`. In that case I guess they are just more likely to directly set it to whatever value they want. I would remove this "magic" behavior and just assume people are very explicit in what they want.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
+            normalization_frequency = stage_gain_frequency
+        sequence = 1  # must be stage 1 because paz defines entire response
+        pzstage = PolesZerosResponseStage(
+                    sequence, stage_gain,
+                    stage_gain_frequency, input_units, output_units,
+                    pz_transfer_function_type, normalization_frequency, zeros,
+                    poles, normalization_factor)
```
krischer: Just a style preferences: Could you please change it `PolesZerosResponseStage(stage_sequence_number=sequence, stage_gain=stage_gain, ...)`? The main reason for this is that it is very easy to make a mistake for functions/objects with a lot of parameters, especially if at one point we change the signature of the object's `__init__()` method.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
+        sensitivity = InstrumentSensitivity(stage_gain, stage_gain_frequency,
+                                            input_units, output_units)
```
krischer: Same comment here.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+
+    def test_resp_loading_vs_evalresp(self):
+        zeros = [0., 0.]
+        poles = [-4.443+4.443j, -4.443-4.443j]
+        stime = UTCDateTime('2017-001T00:00:00.0')
+        filename = 'RESP.XX.NS306..SHZ.GS13.1.2180'
+        evalresp_resp = evalresp(.1, 2**6, filename, stime, units='VEL')
+        resp = Response.from_paz(zeros, poles, 2180.,
+                                 normalization_frequency=5.,
+                                 normalization_factor=1.)
```
krischer: Small nitpick: Can you please set this to anything but `1.0`? We had so many test cases in the past passing just fine but the calculations were actually wrong. But testing with `1.0` in many cases just means it either does not really have to do anything or it could also be the inverse. I think here it is actually fine but better safe than sorry and we've been bitten by this too many times.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+        resp = Response.from_paz(zeros, poles, sensitivity)
+        r_sens = resp.instrument_sensitivity.value
+        np.testing.assert_equal(sensitivity, r_sens)
+
+    def test_resp_loading_vs_evalresp(self):
+        zeros = [0., 0.]
+        poles = [-4.443+4.443j, -4.443-4.443j]
+        stime = UTCDateTime('2017-001T00:00:00.0')
+        filename = 'RESP.XX.NS306..SHZ.GS13.1.2180'
+        evalresp_resp = evalresp(.1, 2**6, filename, stime, units='VEL')
```
krischer: Please use something like:
```python
resp_er = obspy.read_inventory(filename)[0][0][0].response
evalresp_resp = resp_er.get_evalresp_response(.1, 2**6, output='VEL')
```

It is fairly new that `obspy.read_inventory()` can deal with RESP files but i'm honestly not sure how much longer we'll keep evalresp around in ObsPy.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+        resp = Response.from_paz(zeros, poles, 2180.,
+                                 normalization_frequency=5.,
+                                 normalization_factor=1.)
+        response_resp = resp.get_eval_resp_response(.1, 2**6, output='VEL',
```
krischer: It must be `get_evalresp_response()`. Note that you can run the tests locally just fine by either just executing the test file or using `pytest`.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
         paz = self.get_paz()
         return paz_to_sacpz_string(paz, self.instrument_sensitivity)
 
+    @staticmethod
+    def from_paz(zeros, poles, stage_gain,
+                 stage_gain_frequency=1.0, input_units='VOLTS',
+                 output_units='M/S', normalization_frequency=None,
+                 pz_transfer_function_type='LAPLACE (RADIANS/SECOND)',
```
krischer: Input and output units are mixed up here. In term's of response what comes in is ground motion and this is converted to volts, the output units.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+                                                    start_stage=1, end_stage=1)
```
krischer: Please use all stages - there is only one here but who knows. Just leave out the `start_stage` and `end_stage` kwargs.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+        sensitivity = 1201.*(2**26/40.)
+        resp = Response.from_paz(zeros, poles, sensitivity)
+        r_sens = resp.instrument_sensitivity.value
+        np.testing.assert_equal(sensitivity, r_sens)
+
+    def test_resp_loading_vs_evalresp(self):
+        zeros = [0., 0.]
+        poles = [-4.443+4.443j, -4.443-4.443j]
+        stime = UTCDateTime('2017-001T00:00:00.0')
+        filename = 'RESP.XX.NS306..SHZ.GS13.1.2180'
```
krischer: This must be `os.path.join(self.data_dir, 'RESP.XX...')` - otherwise the test will only work if executed from the data directory.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
                   0 + 39477.164708432785j])
 
+    def test_paz_zero_values(self):
+        poles = [1+2j, 1-2j, 2+2j, 2-2j]
+        zeros = [0, 0, 5]
+        resp = Response.from_paz(zeros, poles, 1.0)
+        r_zeros = resp.response_stages[0].zeros
+        np.testing.assert_equal(zeros, r_zeros)
+
+    def test_paz_pole_values(self):
```
krischer: This can be combined with the preceding test case.

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
             out, [0 + 9869.2911771081963j, 0 + 19738.582354216393j,
                   0 + 39477.164708432785j])
 
+    def test_paz_zero_values(self):
```
krischer: Can you please rename your test cases to something like `test_resp_from_paz_settings_zeros()` and similar more descriptive names?

obspy/core/tests/test_response.py
```diff
@@ -298,6 +298,42 @@ def test_response_with_no_units_in_stage_1(self):
...
+        poles = [1+2j, 1-2j, 2+2j, 2-2j]
+        zeros = [0, 0, 5]
+        resp = Response.from_paz(zeros, poles, 1.0)
+        r_zeros = resp.response_stages[0].zeros
+        np.testing.assert_equal(zeros, r_zeros)
```
krischer: Please also test the other set parameters.

obspy/core/tests/data/RESP.XX.NS306..SHZ.GS13.1.2180
```diff
@@ -0,0 +1,86 @@
...
+#		
+B058F03     Stage sequence number:                 1
+B058F04     Gain:                                  2.180000e+03
+B058F05     Frequency of gain:                     5.000000e+00 HZ
+B058F06     Number of calibrations:                0
+#		
+#		+               +-------------------------------------------+                 +
+#		+               |   Response (Coefficients), NS306 ch SHZ   |                 +
+#		+               +-------------------------------------------+                 +
+#
```
krischer: You will probably have to manually remove all non-PAZ stages from this files to get the test to work.

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,42 @@ def get_sacpz(self):
...
+        sequence = 1  # must be stage 1 because paz defines entire response
+        pzstage = PolesZerosResponseStage(
+                    sequence, stage_gain,
+                    stage_gain_frequency, input_units, output_units,
+                    pz_transfer_function_type, normalization_frequency, zeros,
+                    poles, normalization_factor)
+        sensitivity = InstrumentSensitivity(stage_gain, stage_gain_frequency,
+                                            input_units, output_units)
+        resp = Response(instrument_sensitivity=sensitivity,
+                        response_stages=[pzstage])
```
krischer: Please add a `resp.recalculate_overall_sensitivity()` at the end to make sure the overall sensitivity is actually correct at the end.  If `stage_gain_frequency` and `normalization_frequency` differ this is not necessarily the case.

amkearns-usgs: I've made this change (added the recalculate_overall_sensitivity() call).

I would like to make sure that my test captures the desired behavior for this -- for some reason, even if stage_gain_frequency and normalization_frequency are set to the same value, I'm getting a different result for the response's sensitivity compared to the gain for the poles-zeros. This seems like this shouldn't be the case, since the PZ stage is the only one in the response. 

To be clear, when I set the stage gain to 2014943641.6, that IS the stage gain for the PZ stage. However, the overall gain looks something like 432115652.548. The frequencies for stage gain and normalization are both kept at 1.0, as is the normalization factor. Is this an issue with the units, or is there a step that I'm missing somewhere?

obspy/core/inventory/response.py
```diff
@@ -1663,6 +1663,46 @@ def get_sacpz(self):
...
+        :param poles: All poles of the response to be defined.
+        :type stage_gain: float
+        :param stage_gain: The gain value of the response [sensitivity]
+
+        Most of the optional parameters defined here are from
+        :class:`~obspy.core.inventory.response.PolesZerosResponseStage`.
+        :returns: new Response instance with given P-Z values
+        """
+
+        pzstage = PolesZerosResponseStage(
```
amkearns-usgs: I made the requested changes for these lines of code and just wanted to make sure the comment I set has been seen, since I replied to the request for the change rather than as a main comment in the thread.

I would like to make sure that my test captures the desired behavior for the `recalculate_overall_sensitivity()` line -- for some reason, even if stage_gain_frequency and normalization_frequency are set to the same value, I'm getting a different result for the response's sensitivity compared to the gain for the poles-zeros. This seems like this shouldn't be the case, since the PZ stage is the only one in the response.

To be clear, when I set the stage gain to 2014943641.6, that IS the stage gain for the PZ stage. However, the overall gain looks something like 432115652.548. The frequencies for stage gain and normalization are both kept at 1.0, as is the normalization factor. Is this an issue with the units, or is there a step that I'm missing somewhere?

krischer: I'll have to check this on monday. The `recalculate_overall_sensitivity()` function just takes the result of evalresp at a given/derived frequency and uses that as the total sensitivity. I think this is correct. But I'll have to look at it in a bit more detail next week to really answer your question.

amkearns-usgs: OK, I think we've figured out the issue with this toy test case with the help of someone else here at the lab. I didn't actually make the numbers out of anything but thin air, so the gains of the instrument aren't just the one sensitivity parameter, but also the values associated with the poles and zeros themselves. So I could probably counteract that by making sure the A0 value for the PZ stage would set the response curve to 1 at that point, whereas instead right now the A0 *defaults* to 1, and thus the overall sensitivity is affected by the numbers I apply to the poles and zeros.

aringler-usgs: It looks like the response file is using blockette 58 field 4 to state the gain of stage 1.  I think we want to put that as the normalization so that blockette 53 field 7 has the normalization value (that isn't unity) and then put a sensitivity value for a GS-13 (blockette 58 field 4).

amkearns-usgs:
In addition to previous changes and getting the test cases in the previous form to pass, we have now changed the evalresp vs. from-paz resp test to use an A0 normalization factor different from 1.0. Locally, the test cases are passing.

amkearns-usgs:
All requested changes have now been made, including removing the stage-2 data from the response file used in testing (since the values in question were unit, they shouldn't have affected the data, but now we can say so with certainty).

krischer:
Thanks a bunch - I'll try to find some time this week to properly review this!

megies:
This PR is deleting symlinks and replacing them with a copy of the real file (CONTRIBUTORS). This needs some hands-on love. Too late for me to rebase properly and safely..

amkearns-usgs:
How should this be fixed? Move CONTRIBUTORS.txt to the ./obspy/ subdirectory and have a symbolic link at the top level like what has been done for LICENSE.txt?

megies:
I cleaned up this PR, rebased on current master and force pushed.

megies:
CI test fails are unrelated, so this is ready for a final review..

krischer:
Sorry for the crazy slow response on this one. I rebased on the latest master and changed the static method to a class method. Will merge this if CI passes as especially the evalresp comparision test IHMO looks really trustworthy.

megies:
CI fails are unrelated

PR #1963, opened 2017-10-23
Rewrite tests that failed because of line-end chars.

ThomasLecocq:

### What does this PR do?
Resolves some line-end chars failures in tests. The errors might only arise when git cloning the repo.

e.g. http://tests.obspy.org/89313/

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/seiscomp/tests/test_event.py
+++ b/obspy/io/seiscomp/tests/test_event.py
@@ -58,11 +58,7 @@ class EventTestCase(unittest.TestCase):
             if validate:
                 self.assertTrue(_validate_quakeml(tf.name))
             filepath_cmp = os.path.join(self.path, quakeml_file)
-            tf.seek(0)
-            dat1 = tf.read().decode("utf-8").splitlines()
-            with open(filepath_cmp, 'rb') as f:
-                dat2 = f.read().decode("utf-8").splitlines()
-            self.compare_list_of_lines(dat1, dat2)
+            self._compare(tf, filepath_cmp)
 
     def cmp_write_xslt_file(self, quakeml_file, sc3ml_file, validate=True,
                             path=None):
@@ -82,11 +78,7 @@ class EventTestCase(unittest.TestCase):
             if validate:
                 self.assertTrue(validate_sc3ml(tf.name))
             filepath_cmp = os.path.join(self.path, sc3ml_file)
-            tf.seek(0)
-            dat1 = tf.read().decode("utf-8").splitlines()
-            with open(filepath_cmp, 'rb') as f:
-                dat2 = f.read().decode("utf-8").splitlines()
-            self.compare_list_of_lines(dat1, dat2)
+            self._compare(tf, filepath_cmp)
 
     def test_sc3ml_versions(self):
         """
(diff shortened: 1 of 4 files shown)


obspy/io/seiscomp/tests/test_event.py
```diff
@@ -79,7 +84,13 @@ def cmp_write_xslt_file(self, quakeml_file, sc3ml_file, validate=True,
...
                 self.assertTrue(validate_sc3ml(tf.name))
             filepath_cmp = os.path.join(self.path, sc3ml_file)
-            self.assertTrue(filecmp.cmp(filepath_cmp, tf.name))
+            tf.seek(0)
+            dat1 = tf.read().decode("utf-8").splitlines()
+            with open(filepath_cmp, 'rb') as f:
+                dat2 = f.read().decode("utf-8").splitlines()
+            self.assertEqual(len(dat1), len(dat2))
+            for d1, d2 in zip(dat1, dat2):
+                self.assertEqual(d1, d2)
```
krischer: Can you wrap this in a short helper method as its repeated a couple of times?

ThomasLecocq: no :)

obspy/io/seiscomp/tests/test_event.py
```diff
@@ -59,7 +58,11 @@ def cmp_read_xslt_file(self, sc3ml_file, quakeml_file, validate=True):
             if validate:
                 self.assertTrue(_validate_quakeml(tf.name))
             filepath_cmp = os.path.join(self.path, quakeml_file)
-            self.assertTrue(filecmp.cmp(filepath_cmp, tf.name))
+            tf.seek(0)
+            dat1 = tf.read().decode("utf-8").splitlines()
+            with open(filepath_cmp, 'rb') as f:
+                dat2 = f.read().decode("utf-8").splitlines()
+            self.compare_list_of_lines(dat1, dat2)
```
krischer: Sorry to nitpick here - how about a _compare_files(filename, tf) function? That would even further deduplicate the code.

ThomasLecocq: indeed :) I think you have access to the repo, don't you ? :)

krischer: fair enough ;)

obspy/io/seiscomp/tests/test_event.py
```diff
@@ -304,15 +303,24 @@ def test_write_remove_events(self):
...
+            self._compare(tf, filename)
+
+    def _compare(self, tf, filename):
+        tf.seek(0)
+        dat1 = tf.read().decode("utf-8").splitlines()
+        with open(filename, 'rb') as f:
+            dat2 = f.read().decode("utf-8").splitlines()
+        self.assertEqual(len(dat1), len(dat2))
+        for d1, d2 in zip(dat1, dat2):
+            self.assertEqual(d1, d2)
```
megies: This might be better placed under `core/utils/testing.py` since it seems to be generic to ASCII text file comparison and might be useful in other tests as well?

barsch:
I feel this shouldn't be merged - this is working around an issue which should be solved on the client/user sides by syncing/downloading the repo with the original line endings and text encodings, or?

allowing the format test to not fail at this point opens up a can of worms once someone uses this copy to extend ObsPy ...

krischer:
For the encodings I would tend to agree but this is purely to deal with CRLF line endings so I'm not so sure. We do know the encoding of the test files so decoding them should be safe enough. I just checked but neither the cnv format nor the nlloc format appear to specify which line endings to use so I'm not even sure its wrong to use CRLF.

krischer:
Superseded by #1964.

PR #1964, opened 2017-10-24
Adding .gitattributes to force LF line endings in text files

krischer:
This is an alternative to  #1963. It basically forces git to treat all text files with LF line endings. Requires git >= 2.10 to really work but should not do any harm with older gits.

To activate it:

```bash
git rm --cached -r .  # Remove every file from git's index.
git reset --hard      # Rewrite git's index to pick up all the new line endings.
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


krischer:
@ThomasLecocq Any chance you can check if this works?

ThomasLecocq:
running tests now... looks OK !

krischer:
Would be great - this would also prevent any Python CRLF files to get committed in the future. But maybe this was already not possible - I'm not entirely sure how github set things up.

ThomasLecocq:
looks OK
http://tests.obspy.org/89411/

except you based on the tobi's branch ? :)

krischer:
No - buts its already merged. You just forgot to run `pip install` again to register the plug-in I think.

krischer:
In any case - I'm merging this. Thanks for testing!

PR #1965, opened 2017-10-24
fix io.shapefile for pyshp version 1.2.11

megies:
Should fix io.shapefile for pyshp version 1.2.11, at least works locally now for 1.2.10, 1.2.11 and current 1.2.12


megies:
Also works locally, so merging..

PR #1972, opened 2017-10-25
Updating the issue template

krischer:
This is an attempt to steer users to use the mailing list or gitter for generic questions.


barsch:
maybe stress that the mailig list is more for scientific or questions to the general obspy user community (==users) while gitter is more direct communication with the developers/contributers of obspy (==devs, e.g. technical questions)

krischer:
Good point and done.

PR #1973, opened 2017-10-25
Less fragile and actual image test for obspy.taup

krischer:
It previously tested a failure which is really quite fragile for various matplotlib versions. It is now an actual image test.

This should fix the last failures in a couple of the docker test runners (I tested locally).


krischer:
I tested this on a lot of images and it appears to work. Merging to move the release forward.

PR #1979, opened 2017-11-01
Remove deprecated xcorr function

trichter:
This PR removes the old xcorr function which became obsolete with #1585.

Edit: I just realized, that in the mentioned PR it was suggested to wait with the complete removal of xcorr until version 2.0. So maybe wait a bit more before merging this PR ... didn't want to rush.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -23,6 +23,9 @@ Changes:
      encountered (see #2668)
  - obspy.clients.fdsn:
    * introduce fine-grained FDSN client exceptions (see #2653)
+ - obspy.signal.cross_correlation:
+   * Remove deprecated xcorr function, remove deprecated domain keyword
+     argument in correlate function (see #1979)
 
 maintenance_1.2.x
 =================
--- a/obspy/signal/cross_correlation.py
+++ b/obspy/signal/cross_correlation.py
@@ -81,8 +81,7 @@ def _xcorr_slice(a, b, shift, method):
     return cc[mid - shift:mid + shift + len(cc) % 2]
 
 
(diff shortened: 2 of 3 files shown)


krischer:
Yea - I would also wait a bit longer - no harm in leaving the old version around for a bit longer as I'm sure lots of codes use it. Especially given that we want to increase our release frequency.

trichter:
I guess it is OK to remove `xcorr()` in version 2.0?

krischer:
> I guess it is OK to remove `xcorr()` in version 2.0?

Yea should be okay - as long as we clearly document the migration path but that is easy in this case.

trichter:
I rebased this branch on latest master and updated it accordingly. Should be ready for merging?

trichter:
Rebased again on latest master.

heavelock:
I guess it's good to go. This was already agreed to be deleted in 2.0 and it was mentioned as deprecated. 

Or maybe, it would be a good place to start a document `Migration manual` or something similar indicating what could break after update and how to fix it?

trichter:
I agree ;)

IMHO, the what's new document (or similar) could be created shortly before the actual release from the changelog file.

heavelock:
If this would be a personal project of mine, I would not trust myself to leave it for later :P Anyway, merge it in if you feel like that.

heavelock (approved):

megies:
I agree that usually for most changes its OK to just look up things from the Changelog at time of release when putting the "What's New" together but in these cases were people need to know how to change code I would propose to put it now already in a draft for the "Whats New". At least if I were to do the whats new, I wouldnt want to spend time looking up how to migrate code for this one.

trichter:
OK, agreed and done! I am merging...

PR #1980, opened 2017-11-01
Travis: try to reenable 32bit build

megies:
Maybe it  was broken earlier only due to syntax changes in Travis yml, looks like specifying apt package addons in build matrix was once supported but it's now ignored.


megies:
See f4c8f5dcffc02fec5ba0c8eae4ba31c30080486b which disabled 32bit in Travis.

megies:
Ah well, doesn't seem like this works.. so closing.

PR #1984, opened 2017-11-03
minor doc correction of lanczos_interpolation

trichter:
Old [doc](http://docs.obspy.org/master/packages/autogen/obspy.signal.interpolation.lanczos_interpolation.html).


krischer:
Nice catch!

PR #1988, opened 2017-11-06
mseed: Avoid invalid warning when guessing endianness of timestamps / UTCDateTime: proper exceptions on invalid 'julday'

megies:
mseed: avoid showing invalid warnings when guessing endian during
    parsing time stamps
    
    so far, if during parsing a timestamp the endian is not known beforehand
    and if the wrong endian is tried first, a warning message about
    invalid fractional seconds gets shown which is a bogus warning
    
    also, internally we simply rely on UTCDateTime.__init__() raising any
    exception whatsoever for endian detection. this adds some sanity by
    checking if the julday is in between 1-366 (which is the range accepted
    by UTCDateTime.__init__)

---

    utcdatetime: add proper sanity checks for UTCDateTime(..., julday=..)
    
    right now passing invalid juldays can result in misleading error messages, because
    julday is silently discarded when encountering out-of-bounds values that
    internally in datetime.datetime.strptime(..., '%Y%j') raise "ValueError:
    unconverted data remains:"
    
    In [1]: UTCDateTime(year=2017, julday=367)
    <ipython-input-1-0e1465633d82> in <module>()
    ----> 1 UTCDateTime(year=2017, julday=367)
    
    /home/megies/git/obspy-master/obspy/core/utcdatetime.py in __init__(self, *args, **kwargs)
        357             kwargs['microsecond'] = int(round(_frac * 1e6))
        358             args = args[0:5]
    --> 359         dt = datetime.datetime(*args, **kwargs)
        360         self._from_datetime(dt)
        361
    
    TypeError: Required argument 'month' (pos 2) not found
    
This changes this to:
    
    In [1]: UTCDateTime(year=2017, julday=367)
    <ipython-input-1-0e1465633d82> in <module>()
    ----> 1 UTCDateTime(year=2017, julday=367)
    
    /home/megies/git/obspy-master/obspy/core/utcdatetime.py in __init__(self, *args, **kwargs)
        342             if not (0 <= int(kwargs['julday']) <= 366):
        343                 msg = "'julday' out of bounds: {!s}".format(kwargs['julday'])
    --> 344                 raise ValueError(msg)
        345             if 'year' in kwargs:
        346                 # year given as kwargs
    
    ValueError: 'julday' out of bounds: 367

no time to add tests right now, sorry..

### DIFF
--- a/obspy/io/mseed/util.py
+++ b/obspy/io/mseed/util.py
@@ -20,6 +20,7 @@ import numpy as np
 from obspy import UTCDateTime
 from obspy.core.compatibility import from_buffer
 from obspy.core.util.decorator import ObsPyDeprecationWarning
+from . import InternalMSEEDParseTimeError
 from .headers import (ENCODINGS, ENDIAN, FIXED_HEADER_ACTIVITY_FLAGS,
                       FIXED_HEADER_DATA_QUAL_FLAGS,
                       FIXED_HEADER_IO_CLOCK_FLAGS, HPTMODULUS,
@@ -641,6 +642,10 @@ def _get_record_information(file_object, offset=0, endian=None):
         return native_str('%sHHBBBxHHhhBBBxlxxH' % s)
 
     def _parse_time(values):
+        if not (0 <= values[1] <= 366):
+            msg = 'julday out of bounds (wrong endian?): {!s}'.format(
+                values[1])
+            raise InternalMSEEDParseTimeError(msg)
         # The spec says values[5] (.0001 seconds) must be between 0-9999 but
         # we've  encountered files which have a value of 10000. We interpret
         # this as an additional second. The approach here is general enough
@@ -663,7 +668,7 @@ def _get_record_information(file_object, offset=0, endian=None):
             endian = ">"
             values = unpack(fmt(endian), data)
             starttime = _parse_time(values)
-        except Exception:
+        except InternalMSEEDParseTimeError:
             endian = "<"
             values = unpack(fmt(endian), data)
             starttime = _parse_time(values)
@@ -671,7 +676,7 @@ def _get_record_information(file_object, offset=0, endian=None):
         values = unpack(fmt(endian), data)
         try:
             starttime = _parse_time(values)
-        except Exception:
+        except InternalMSEEDParseTimeError:
             msg = ("Invalid starttime found. The passed byte order is likely "
                    "wrong.")
             raise ValueError(msg)
--- a/obspy/io/mseed/util.py
+++ b/obspy/io/mseed/util.py
@@ -642,7 +642,7 @@ def _get_record_information(file_object, offset=0, endian=None):
         return native_str('%sHHBBBxHHhhBBBxlxxH' % s)
 
     def _parse_time(values):
-        if not (0 <= values[1] <= 366):
+        if not (1 <= values[1] <= 366):
             msg = 'julday out of bounds (wrong endian?): {!s}'.format(
                 values[1])
             raise InternalMSEEDParseTimeError(msg)
--- a/obspy/io/mseed/util.py
+++ b/obspy/io/mseed/util.py
@@ -658,10 +658,15 @@ def _get_record_information(file_object, offset=0, endian=None):
                 "the maximum strictly allowed value is 9999. It will be "
                 "interpreted as one or more additional seconds." % values[5],
                 category=UserWarning)
-        return UTCDateTime(
-            year=values[0], julday=values[1],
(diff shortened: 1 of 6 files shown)


Jollyfant:
Shouldn't the minimum be `julday=1` instead of 0? Considering that the last index is 366.

megies:
> Shouldn't the minimum be julday=1 instead of 0? Considering that the last index is 366.

You're right, we should be checking for `1 <= julday <= 366`  :+1:

megies:
Fixed that point and rebased.

krischer:
needs to be rebased again

megies:
Good to merge I think @krischer

obspy/io/mseed/util.py
```diff
@@ -653,25 +658,30 @@ def _parse_time(values):
...
-            hour=values[2], minute=values[3], second=values[4],
-            microsecond=msec % 1000000) + offset
+        try:
+            t = UTCDateTime(
+                year=values[0], julday=values[1],
+                hour=values[2], minute=values[3], second=values[4],
+                microsecond=msec % 1000000) + offset
+        except TypeError:
+            msg = 'Problem decoding time (wrong endian?)'
+            raise InternalMSEEDParseTimeError(msg)
```
krischer: Maybe pass on the message of the exception which would ease debugging and improve understanding of whatever problem is encountered?

obspy/io/mseed/util.py
```diff
@@ -653,25 +658,30 @@ def _parse_time(values):
...
             starttime = _parse_time(values)
     else:
         values = unpack(fmt(endian), data)
         try:
             starttime = _parse_time(values)
-        except Exception:
+        except InternalMSEEDParseTimeError:
             msg = ("Invalid starttime found. The passed byte order is likely "
                    "wrong.")
             raise ValueError(msg)
```
krischer: again pass on exception message.

obspy/io/seisan/tests/test_core.py
```diff
@@ -195,13 +194,8 @@ def test_read_obspy(self):
         # 1 - little endian, 32 bit, version 7
         st1 = read(os.path.join(self.path,
                                 '2011-09-06-1311-36S.A1032_001BH_Z'))
-        # raises "UserWarning: Record contains a fractional seconds" - ignore
-        with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter('always', UserWarning)
-            st2 = read(os.path.join(self.path,
-                                    '2011-09-06-1311-36S.A1032_001BH_Z.mseed'))
-            self.assertEqual(len(w), 1)
-            self.assertEqual(w[0].category, UserWarning)
```
krischer: why is this no longer necessary?

obspy/io/mseed/util.py
megies: I don't think the message would be very helpful, at least not in the case that seems to be what is normally encountered (when trying the wrong endian):

```
$ python -c 'from obspy import UTCDateTime; UTCDateTime(year=444444, julday=10)'
Version 7.000kness 20eader in AFM: UnderlinePosition -133
  File "/home/megies/git/obspy/obspy/core/utcdatetime.py", line 377, in __init__
    dt = datetime.datetime(*args, **kwargs)
TypeError: Required argument 'month' (pos 2) not found
```

I'm not sure it would help to see that message to be honest.

obspy/io/seisan/tests/test_core.py
megies: Reading this file triggers the invalid warning that I got rid of in mseed reading (i.e. potentially trialing the wrong endian in the first attempt).

The only thing that changes in here is that the warning-catching was removed, nothing else.

obspy/io/mseed/util.py
megies: In any case this PR is not making things worse. Before it there was a bare try/except, now at least I'm only catching an exception that is explicitly raised inside the subroutine that parses the timestamp.

megies:
> also add a test for the MiniSEED part

What exactly do you want tested? Outwards facing, the only thing that changes in mseed is that the invalid warning went away, see the change in that seisan test. Sure, I could add a test that checks that no warnings a shown. But I don't see a lot of value in that.. it's kinda implicitly tested because if you run the tests you don't see a warning cluttering stdout.. ;-)

Other than that there's only changes in `_parse_time()` that could be tested, so you want a test for that internal routine (which didn't have tests yet and isn't even exposed on module level)?

megies:
@krischer do you really want more changes in here or can i merge and do 1.1.1RC1?

krischer:
You are right - I misinterpreted what this PR is doing. Merging :-)

Issue #1989, opened 2017-11-07
pickling SOH traces


dsentinel:
These raise a division by zero Exception:
Which prevents pickling traces.

```
from obspy.core import Stats, Trace
import pickle

s = Stats()
s.delta = 0    

t = Trace()
t.stats.sampling_rate = 0
pickle.loads(pickle.dumps(t))
```
```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-7-c105e6dd98fb> in <module>()
----> 1 pickle.loads(pickle.dumps(t))

~/code/obspy/obspy/core/util/attribdict.py in __setstate__(self, adict)
    113         self.__dict__.update(self.defaults)
    114         # update with pickle dictionary
--> 115         self.update(adict)
    116 
    117     def __getattr__(self, name, default=None):

~/code/obspy/obspy/core/util/attribdict.py in update(self, adict)
    140             if key in self.readonly:
    141                 continue
--> 142             self.__setitem__(key, value)
    143 
    144     def _pretty_str(self, priorized_keys=[], min_label_length=16):

~/code/obspy/obspy/core/trace.py in __setitem__(self, key, value)
    159             if key == 'delta':
    160                 key = 'sampling_rate'
--> 161                 value = 1.0 / float(value)
    162             elif key == 'sampling_rate':
    163                 value = float(value)

ZeroDivisionError: float division by zero
```

krischer:
Fur future reference: No need to explicitly close this - github can do it automatically if the correct triggers are used: https://help.github.com/articles/closing-issues-using-keywords/

PR #1990, opened 2017-11-07
Fix bug setting stats.delta to zero. Tests. Pickling tests.

dsentinel:
Addresses bug #1989
This bug prevented pickling of SOH channels with SR = 0.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -158,9 +158,9 @@ class Stats(AttribDict):
             # ensure correct data type
             if key == 'delta':
                 key = 'sampling_rate'
-                if value > 0.0:
+                try:
                     value = 1.0 / float(value)
-                else:
+                except ZeroDivisionError:
                     value = 0.0
             elif key == 'sampling_rate':
                 value = float(value)
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -158,7 +158,10 @@ class Stats(AttribDict):
             # ensure correct data type
             if key == 'delta':
                 key = 'sampling_rate'
-                value = 1.0 / float(value)
+                if value > 0.0:
+                    value = 1.0 / float(value)
+                else:
+                    value = 0.0
             elif key == 'sampling_rate':
                 value = float(value)
             elif key == 'starttime':
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,7 @@
+1.1.x:
+ -obspy.core:
+   * Allow pickling of traces with sr=0 / setting delta =0 (see #1990)
+
 1.1.0: (doi: 10.5281/zenodo.165135)
  - General:
    * Read support for Guralp Compressed Format (GCF) waveform data,
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,6 @@
 1.1.x:
  -obspy.core:
-   * Allow pickling of traces with sr=0 / setting delta =0 (see #1990)
+   * Fix pickling of traces with a sampling rate of 0 (see #1990)
 
 1.1.0: (doi: 10.5281/zenodo.165135)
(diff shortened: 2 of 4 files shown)


obspy/core/trace.py
```diff
@@ -158,7 +158,10 @@ def __setitem__(self, key, value):
             # ensure correct data type
             if key == 'delta':
                 key = 'sampling_rate'
-                value = 1.0 / float(value)
+                if value > 0.0:
+                    value = 1.0 / float(value)
+                else:
+                    value = 0.0
```
krischer: Can you use the same `try/except` construct that is used for this check for the sampling rate a couple of lines down? Or change the other one as well? Better to have things consistent.

dsentinel:
> At first I was a bit skeptical about this but then I realized we've always been doing this for the sampling rate and thus we cannot easily change it as some people probably rely on it. So this PR improves consistency which it good! Thanks for this!

I don't think I understand, but glad your on board now. :)
Happy to help.

> A side note and just FYI: We also have waveform reading/writing plugins so tr.write(buf, format="pickle", protocol=2) works and it can also be detected while reading.

Thanks, good to know, but I'm passing stream objects around multiprocessing trying to save wall time. I'm not doing any pickling manually, but multiprocessing lib pickles to pipes automagically and crashes on sr = 0 traces. Pickling sets all the attributes, including derived: sr, delta, endtime, which is silly as only sr, starttime and npts  would need to be set. We could define `__getstate__()` to return `self.__dict__` minus all derived values. But I would think we would still want to allow setting of delta = 0 as we return delta=0 if sr=0 (maybe?)

- [ ] Can you please also add a short changelog entry? As this would be the first entry for the 1.1.x line best add a new section at the beginning and just call it 1.1.x.

Surely.

obspy/core/trace.py
dsentinel: I'll go with `try/except` on zerodivision. This will let whatever the behavior of negative SR and deltas to remain.

dsentinel:
Had to force push (because of reabasing?) 

Note that I also created a test for setting `trace().delta = 0`  is this wanted?

I also created a bug #1989, should I create an issue if I'm also going to create a PR, I thought you could take an issue and click turn issue into a PR. 

I think it's ready to merge.

krischer:
Thanks for this!

> I also created a bug #1989, should I create an issue if I'm also going to create a PR, I thought you could take an issue and click turn issue into a PR.

No - just the PR is enough.

It is also possible to convert issues to PRs - best google how to do it - but Github changed it like a year ago so that the original author of the issue would get changed if the PR is from a different person. So nowadays we only do it if the issue is from the same person that also does the PR.

> Had to force push (because of reabasing?)

Yes.

> Note that I also created a test for setting trace().delta = 0 is this wanted?

Yes. Especially as this is part of the underlying reason why it failed.


> Thanks, good to know, but I'm passing stream objects around multiprocessing trying to save wall time. I'm not doing any pickling manually, but multiprocessing lib pickles to pipes automagically and crashes on sr = 0 traces.

You might consider serializing to uncompressed miniseed or SAC to transfer from one processor to the next - this might actually be faster.

If you are serious about this and don't mind a bit of additional engineering work: It is possible to actually share memory and thus numpy arrays across processors by using `multiprocessing.Array`. Then you don't have to serialize the actual data at all but just pass a pointer to each forked process.

`joblib` has AFAIK also a pretty fast way to serialize numpy arrays to share across processors.

> Pickling sets all the attributes, including derived: sr, delta, endtime, which is silly as only sr, starttime and npts would need to be set. We could define __getstate__() to return self.__dict__ minus all derived values. But I would think we would still want to allow setting of delta = 0 as we return delta=0 if sr=0 (maybe?)

Might be nice. On the other hand: with this PR it seems to work fine so why add the additional complexity and maintenance overhead? Up to you.

PR #1993, opened 2017-11-09
Bug in reading SC3ML (assumes optional maxClockDrift is always present)

megies:
There is a bug in reading SC3ML station metadata into Inventory. It is assumed that `maxClockDrift` is always present, but it's optional in SC3ML. First reported on users mailing list: http://lists.swapbytes.de/archives/obspy-users/2017-November/002578.html

I changed the test data so that the bug shows with current tests.

@Jollyfant can you have a look since you implemented SC3ML -> Inventory?

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,8 @@
 1.1.x:
  -obspy.core:
    * Fix pickling of traces with a sampling rate of 0 (see #1990)
+ -obspy.io.seiscomp:
+   * Fix inventory read when maxClockDrift is unset in SC3ML (see #1993)
 
 1.1.0: (doi: 10.5281/zenodo.165135)
  - General:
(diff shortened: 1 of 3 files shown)


Jollyfant:
Added a simple check for None if the tag is missing. Should be enough?

Jollyfant:
Hm did I screw up? There's some commits in there that I don't think should be

krischer:
Yea something went astray - I fixed by manually cherry-picking the commits. Please check if I got everything before you delete your local branch.

krischer:
Also please add a changelog entry.

megies:
> Looks good to me but please change the target branch and rebase to maintenance_1.1.x.

Ahhh, my bad, opening this against `master`, sorry!

Would be good if github would not use a default base branch but force users to manually select a base branch (or if we could specify some options for most used base branches to chose from)

Issue #2009, opened 2017-11-17
NOR to CNV


shicks-seismo:
io.nordic reads in pick weight from a cnv catalogue (i.e. 0 to 4) as the arrival.timeweight variable.
However, when writing the same catalogue to CNV format, the weights are not written. This is because CNV is trying to write pick.time_errors.uncertainty.
I'm not sure which is the correct convention here, but they should be consistent ...
Perhaps a fix needed for a future update ...?

krischer:
I assigned @megies and @calum-chamberlain because I think they are originally responsible for the `io.cnv` and `io.nordic` modules ;-)

megies:
Uhm.. I haven't used CNV export in ages, so I'll hapilly accept PRs. Argumentation seems reasonable, do you have examples for real world use of this pick weight stuff?

calum-chamberlain:
I definitely used my discretion for where to put the weight in `io.nordic`... Those values are weights used in computing locations, and should be a vague representation of pick uncertainty. But, they are not in any kind of time units, when I transfer to NonLinLoc for large catalogues I do some stats to try and relate those weights to pick uncertainty in time (which is what I would think pick.time_errors.uncertainty should be).

Without knowing the CNV format, I would say that Nordic is doing what I think it should, and I don't think the two values are interchangeable. My take is that it should be up to the user to convert in some sensible way between the weights and physical uncertainty values.

I have many examples for pick-weights in Nordic format if needed - also weighted picks are included in the tests for `io.nordic`.

megies:
Is this still an issue or should we close it (I'm tempted to). Has been stale for 1.5 years now.

PR #2010, opened 2017-11-18
bugfix CNV write, fixes #2001

### LINKED ISSUE #2001: CNV writing phase_mapping fails
This line:
https://github.com/obspy/obspy/blob/master/obspy/io/cnv/core.py#L63

doesn't work because 

``values.update(("P", "S"))`` returns ``None``

This works:
```python
values = set(phase_mapping.values())
values.update(("P","S"))
    if values != set(("P", "S")):

```

ThomasLecocq:
Passing ``phase_mapping`` to ``Catalog.write(format="CNV", phase_mapping=...)`` resulted in an exception.

Bugfix #2001 

This https://github.com/obspy/obspy/issues/2009 is not linked but could be part of the same PR?


ThomasLecocq:
How do I disable the warnings when S phases don't have a mapping in the last test?

krischer:
`warnings.catch_warnings()` - you can also use this to test the raised warning and if it is an expected warning it is a good idea to do so.

megies:
Oh.. @ThomasLecocq should be rebased on `maintenance_1.1.x`!

ThomasLecocq:
@megies did you rebase or just changed the "base branch" of the PR ?

ThomasLecocq:
I _think_ I rebased on ``maintenance_1.1.x`` but then why does it show stuff from you... dunno...

megies:
> I think I rebased on maintenance_1.1.x but then why does it show stuff from you... dunno...

I don't think this worked.. those commits are from `master` and they're not even the original ones but re-commited.. 😜 

I can rebase your branch if you like, but you'll need to check it out fresh from here then..?

ThomasLecocq:
Pleeeaaase do!!!

megies:
@ThomasLecocq please double check. :-)

megies:
Rebased.. can we merge this @ThomasLecocq?

ThomasLecocq:
Yep, if test pass, I guess it's good to go.

megies:
Yeah some conda packaging issues for proj on win it seems

PR #2011, opened 2017-11-20
Allowing singular COUNT unit in evalresp

### LINKED ISSUE #2003: Evalresp doesn't recognise IRIS response units
Running the example [here](https://docs.obspy.org/tutorial/code_snippets/seismometer_correction_simulation.html#using-a-resp-file) raises an `EVALRESP ERROR` as follows:
```
 EVRESP ERROR (BFZ.NZ.10.HHZ [File: /tmp/obspy-2vae215y.tmp; Start date: 2009,323,01:00:01; Stage: 2]):
	check_units; units found ('COUNT - DIGITAL COUNTS') are not supported,
	skipping to next response now
```

This is running in a fairly clean anaconda env with obspy v. 1.1.0 - tested with both py3.5 and 3.6 (3.6 for the clean environment).  Digging deeper, it looks like it may be a change with the IRIS client - this worked yesterday, but not today, so I'm a little confused...?

If I replace `COUNT - Digital Counts` with `COUNTS` in the output file then there is no error. To me it looks like something changed at IRIS in how they write response units, but without a file from yesterday to hand it is hard to check.

krischer:
Fixes #2003 

Small thing and I did not add a test because we don't really use this internally but it might help some people.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,6 +3,8 @@
    * Fix pickling of traces with a sampling rate of 0 (see #1990)
  -obspy.io.seiscomp:
    * Fix inventory read when maxClockDrift is unset in SC3ML (see #1993)
+ -obspy.signal:
+   * Allow singular COUNT units in evalresp (see #2003, #2011).
 
 1.1.0: (doi: 10.5281/zenodo.165135)
  - General:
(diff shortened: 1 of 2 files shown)


megies:
@krischer wrong base branch..

krischer:
Hehe - but only on github - locally I used the correct one :)

PR #2015, opened 2017-11-22
StationXML write dates before 1900

### LINKED ISSUE #2013: Inventory write fails when network starts before year 1900, due to datetime strftime
inventory.write fails  when the network start_date is before 1900  ( which it is for the New Zealand national network, ==1884), due to the datetime strftime(), which requires year >= 1900.

[ Using obspy 1.1 through anaconda, and python 2.7 ] 

Fails through line 320 in inventory.py, in write   ; line 864 in core.py in _write_stationxml ; line 928 in core.py, in _write_network ; line 928 in core.py in _get_base_node_attributes, ; line 1490 in core.py in _format_time,  and line 1213 in utcdatetime.py in strftime.
Final line in the traceback is "ValueError: year=1884 is before 1900; the datetime strftime() methods require year >= 1900"

krischer:
Fixes #2013 also on Python 2.7.

### DIFF
--- a/obspy/io/stationxml/core.py
+++ b/obspy/io/stationxml/core.py
@@ -855,7 +855,7 @@ def _write_stationxml(inventory, file_or_file_object, validate=False,
         etree.SubElement(root, "Module").text = inventory.module
         etree.SubElement(root, "ModuleURI").text = inventory.module_uri
 
-    etree.SubElement(root, "Created").text = _format_time(inventory.created)
+    etree.SubElement(root, "Created").text = str(inventory.created)
 
     if level not in ["network", "station", "channel", "response"]:
         raise ValueError("Requested stationXML write level is unsupported.")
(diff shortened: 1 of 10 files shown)


obspy/io/stationxml/core.py
```diff
@@ -1482,10 +1482,12 @@ def _obj2tag(parent, tag_name, tag_value):
...
 
 def _format_time(value):
+    # Piggy back on the __str__ implementation of our UTCDateTime objects as
+    # they work well also for years before 1900.
     if value.microsecond == 0:
-        return value.strftime("%Y-%m-%dT%H:%M:%S")
+        return str(value).split(".")[0]
     else:
-        return value.strftime("%Y-%m-%dT%H:%M:%S.%f")
+        return str(value).rstrip("Z")
```
barsch: XML Schema DateTime allows time zones (as well as the Z terminator for UTC) - e.g. https://www.w3schools.com/xml/schema_dtypes_date.asp or http://books.xmlschemata.org/relaxng/ch19-77049.html

so a plain ```str(value)``` should be enough

krischer:
> XML Schema DateTime allows time zones (as well as the Z terminator for UTC) - e.g. https://www.w3schools.com/xml/schema_dtypes_date.asp or http://books.xmlschemata.org/relaxng/ch19-77049.html
> 
> so a plain str(value) should be enough

We discussed this in the past I feel and I guess there is also a reason to not always write the microseconds? In any case - I think this is a good idea so it now always writes the microseconds as well as the UTC time zone terminator.

Please speak up if this is has some kind of negative implications.

Also should we rather merge this to `master` instead of the maintenance branch as it affects the way stationxml files are written?

barsch:
IMHO maintance is fine - any StationXML parser should be able to handle timezone information according to the StationXML specifications (==provided XML schema)

btw it needs now also a test for reading datetime w/o time zones ...

krischer:
That's implicitly tested as most of our test files don't have time zones. I only added them to the few test files where we test strict round-tripping.

PR #2016, opened 2017-11-23
Ndk bugfix

jsaul:
Line 3 must be treated as fixed-width fields because otherwise parsing fails for a few ndk's where there is no space between the centroid time and its error, like e.g. in event C110397G.

To test this:
```
wget 'http://www.ldeo.columbia.edu/~gcmt/projects/CMT/catalog/jan76_dec13.ndk'
python -c 'import obspy; obspy.read_events("jan76_dec13.ndk")'
```
This produced several errors but now works.

### DIFF
--- a/obspy/io/ndk/core.py
+++ b/obspy/io/ndk/core.py
@@ -515,11 +515,15 @@ def _read_lines(line1, line2, line3, line4, line5):
     #         events, the date and time of the analysis. This is useful to
     #         distinguish Quick CMTs ("Q-"), calculated within hours of an
     #         event, from Standard CMTs ("S-"), which are calculated later.
+    if line3[0:9] != "CENTROID:":
+        raise ObsPyNDKException("parse error")
+    numbers = [ line3[10:18], line3[18:22], line3[22:29], line3[29:34], \
+        line3[34:42], line3[42:47], line3[47:53], line3[53:58] ]
     rec["centroid_time"], rec["centroid_time_error"], \
         rec["centroid_latitude"], rec["centroid_latitude_error"], \
         rec["centroid_longitude"], rec["centroid_longitude_error"], \
         rec["centroid_depth_in_km"], rec["centroid_depth_in_km_error"] = \
-        map(float, line3[:58].split()[1:])
+        map(float, numbers)
     type_of_depth = line3[59:63].strip().upper()
 
     if type_of_depth == "FREE":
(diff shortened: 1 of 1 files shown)


krischer:
Can you please rebase on the `maintenance_1.1.x` branch and force push to get rid of all the extra commits?

krischer:
Can you fix the formatting issues and add a test case?

I remember encountering this before but somehow came to the conclusion that there was some problem with just hard coding the offset but I cannot recall why. Maybe I was mistaken.

krischer:
So the offending events look like this:

```
PDE  1996/11/20 19:42:56.1  10.30  127.43  33.0 5.6 5.3 PHILIPPINE ISLANDS REGIO
B112096E         B: 35   58  45 S:  0    0   0 M:  0    0   0 CMT: 1 BOXHD: 22.0
CENTROID:      1.060.0  10.47 0.03  127.42 0.03  15.0  0.0 BDY  O-00000000000000
24 -4.983 0.133 -0.092 0.153  5.075 0.173 -0.433 0.443  0.715 0.469 -2.898 0.141
V10   6.430  4 246  -1.390  2 156  -5.050 86  45   5.740 338 41  -88 154 49  -92
```

The third lines merges the first and second values (being the time and the time error). This is problematic if the time error is larger than 9.9 which occurs in only 2 events in the GCMT catalogue. But in this case they could be interpreted as `1.0` and `60.0` or `1.06` and `0.0`. I personally would favor the first interpretation which this PR implements but its not really defined and IMHO just a faulty file, especially as these are the only non-whitespace separated values in ndk files.

I'm fine with this PR once it has tests but would be happy to hear other opinions as well.

jsaul:
If there are still any doubts left then please try
```
awk 'NR%5==3' jan76_dec13.ndk
awk 'NR%5==3 && substr($0,19,1)!=" "'  jan76_dec13.ndk
```
It is clear that in the example the error is 60.0. Such a large value is almost certainly an error in the original file. But do you want the parser to fail here? In the other example (C110397G) the error is 10.0 and while certainly very rare it is not a totally unrealistic value.

With the PR you are on the safe side.

megies:
Would be good to ask the GCMT guys about this?

jsaul:
It's not a question about semantics. As I wrote above, the 60 s uncertainty is most probably a mistake and it would be good to correct it upstream. Not so sure about the 10 s in the other example, though. In any case the examples show that uncertainty values >= 10 s do occur. They must not cause the parser to fail. Just like there are many time strings in the NDK files with the seconds field being 60.0, which is rejected by the UTCDateTime constructor but which the ObsPy NDK parser fortunately does accept via a workaround.

megies:
Ideally a regression test should be added with one or two of these offending examples, otherwise fine for me.

megies:
I think this should fix PEP8.

megies:
Alright.. regression test still lacking but I'm gonna merge this now, otherwise it'll stay pending forever..

PR #1987, opened 2017-11-23
Raise Arclink EOL message

krischer:
Arclink will "soon" be EOL. The EIDA management board asked us to raise a warning message when users are importing the Arclink client.

They'll come back to us with a concrete suggestion for this warning message.

Including this in the next bugfix release should be fine as it does not change behavior.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,9 @@
 1.1.x:
  - obspy.core:
    * Fix pickling of traces with a sampling rate of 0 (see #1990)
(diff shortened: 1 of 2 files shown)


krischer:
This issue is now a PR raising a warning at arclink import time. Looks like this (not pretty but visible and it gets the job done):

screen shot 2017-11-23 at 17 37 02

That wording has been proposed by the EIDA management board.

megies:
Doesnt have a test, but I think it's really not necessary, as it's happening in main code block of `__init__.py`.

Issue #1822, opened 2017-06-23
Inventory (and Network) should have plot option with station/channel epochs


megies:
`Inventory` and `Network` (and maybe `Station`) should have a plot option like `obspy-scan`, basically showing the station/channel epochs defined in the metadata.

amkearns-usgs:
I am interested in helping with this feature request. A few questions (as I've not looked too deeply into the `obspy-scan` script):

Is this to show data availablity? Is the goal mainly to have timeseries plots that show, like `obspy-scan`, something like red bars over ranges that don't correspond to an active epoch? Do we want to have a marker for the boundary of two contiguous epochs? Do we want to also demarcate gaps in traces where no data has been collected?

krischer:
Not really to show data availability (as it would be based purely on station metadata) but to plot the start and end times of the networks, stations, and channel epochs in that inventory. StationXML files might have an additional data availability tag that could also be overlayed somehow. I did not give this too much thought yet but I envision something like a gant chart that shows when each entity in the inventory objects claims to be active.

krischer:
Quick mockup:

![screen shot 2017-11-09 at 08 59 51 1](https://user-images.githubusercontent.com/800487/32594474-7205a274-c52c-11e7-8d3c-3c8ce634c7ec.png)

QuLogic:
Something like this graph that appears just before the linked section? http://earthpy.org/tag/matplotlib.html#Visualizing-Data (there's no nearby anchor except this one right after the plot I mean)

krischer:
That is indeed very nice :)

megies:
We could group all channels of a station by not using a line plot for the station level but rather a rectangle with colored fill and only plot the channels as lines (or fine rectangles) on top of it..

amkearns-usgs:
I like this grouping idea. Would we want to do something similar to group all the stations in a network, or would it be better to make separate plots for them?

megies:
i think it should be one plot for everything (in one Inventory).

and, not sure myself, channel level as lines (or thin rectangles) seems pretty clear, otherwise we could use `alpha`d rectangles for all of the higher levels, but i guess we'd need to try it out, how it looks..

megies:
Dunno.. I guess the "group by background shading" really only works for one level, either network or station.. or maybe we could use the colored shade for station level and use a thick box outline for network level?

amkearns-usgs:
[...]

PR #2023, opened 2017-12-06
Epoch plots

amkearns-usgs:

!! Please check that you select the **correct base branch** (details see below link) !!

Before submitting a PR, please review the pull request guidelines:
https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Also, please make sure you are following the ObsPy branching model:

### What does this PR do?

This allows for inventory objects to have their epochs plotted as 

See issue #1822 

### PR Checklist
- [*] Correct base branch selected? `master` for new fetures, `maintenance_...` for bug fixes
- [*] All tests still pass.
- [*] Any new or changed features have are fully documented.

### DIFF
--- a/obspy/core/inventory/channel.py
+++ b/obspy/core/inventory/channel.py
@@ -368,7 +368,7 @@ class Channel(BaseNode):
             unwrap_phase=unwrap_phase, plot_degrees=plot_degrees, show=show,
             outfile=outfile)
 
-    def get_epoch_plottable_struct(self):
+    def _get_epoch_plottable_struct_(self):
         plot_dict = {};
         name = str(self.code)
         if self.start_date is not None:
@@ -380,6 +380,12 @@ class Channel(BaseNode):
         return plot_dict
 
     def plot_epochs(self, outfile=None):
+        """
+        Plot the epochs of this given inventory object.
+        :param outfile: If included, the plot will be saved to a file with the
+            given filename. (Otherwise it will be displayed in a window)
+        :type outfile: str
+        """
         plot_dict = self.get_epoch_plottable_structure()
         plot_inventory_epochs(plot_dict, outfile)
 
--- a/obspy/core/inventory/channel.py
+++ b/obspy/core/inventory/channel.py
@@ -14,10 +14,12 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 from future.utils import python_2_unicode_compatible
 
+from obspy import UTCDateTime
 from obspy.core.inventory.util import plot_inventory_epochs
 from obspy.core.util.obspy_types import FloatWithUncertainties
 from . import BaseNode
-from .util import Azimuth, ClockDrift, Dip, Distance, Latitude, Longitude
+from .util import (Azimuth, ClockDrift, Dip, Distance, Latitude, Longitude,
+                   plot_inventory_epochs)
 
 
 @python_2_unicode_compatible
@@ -366,13 +368,15 @@ class Channel(BaseNode):
             unwrap_phase=unwrap_phase, plot_degrees=plot_degrees, show=show,
             outfile=outfile)
 
-    def get_epoch_plottable_struct(self, y_offset=0):
+    def get_epoch_plottable_struct(self):
         plot_dict = {};
+        name = str(self.code)
         if self.start_date is not None:
-            end = self.end_date
-            if end is None:
-                end = obspy.core.utcdatetime.now()
-            plot_dict[y_offset] = (self.start_date, end, 0, self.location_code)
+            if self.end_date is None:
+                end = UTCDateTime.now()
+            else:
+                end = min(self.end_date, UTCDateTime.now())
+            plot_dict[name] = [(self.start_date, end, {})]
         return plot_dict
 
     def plot_epochs(self, outfile=None):
--- a/obspy/core/inventory/channel.py
+++ b/obspy/core/inventory/channel.py
@@ -14,6 +14,7 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 from future.utils import python_2_unicode_compatible
 
+from obspy.core.inventory.util import plot_inventory_epochs
 from obspy.core.util.obspy_types import FloatWithUncertainties
 from . import BaseNode
 from .util import Azimuth, ClockDrift, Dip, Distance, Latitude, Longitude
@@ -373,6 +374,10 @@ class Channel(BaseNode):
                 end = obspy.core.utcdatetime.now()
             plot_dict[y_offset] = (self.start_date, end, 0, self.location_code)
         return plot_dict
+
+    def plot_epochs(self, outfile=None):
+        plot_dict = self.get_epoch_plottable_structure()
+        plot_inventory_epochs(plot_dict, outfile)
 
 
 if __name__ == '__main__':
--- a/obspy/core/inventory/channel.py
+++ b/obspy/core/inventory/channel.py
@@ -365,6 +365,15 @@ class Channel(BaseNode):
             unwrap_phase=unwrap_phase, plot_degrees=plot_degrees, show=show,
             outfile=outfile)
 
+    def get_epoch_plottable_struct(self, y_offset=0):
+        plot_dict = {};
+        if self.start_date is not None:
+            end = self.end_date
+            if end is None:
+                end = obspy.core.utcdatetime.now()
+            plot_dict[y_offset] = (self.start_date, end, 0, self.location_code)
+        return plot_dict
+
 
 if __name__ == '__main__':
     import doctest
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -21,7 +21,6 @@ import textwrap
 import warnings
 
 import obspy
-from obspy.core.inventory.util import plot_inventory_epochs
 from obspy.core.util.base import (ENTRY_POINTS, ComparingObject,
                                   _read_from_plugin, NamedTemporaryFile,
                                   download_to_file)
@@ -30,7 +29,7 @@ from obspy.core.util.misc import buffered_load_entry_point
 from obspy.core.util.obspy_types import ObsPyException, ZeroSamplingRate
 
 from .network import Network
-from .util import _unified_content_strings, _textwrap
+from .util import _unified_content_strings, _textwrap, plot_inventory_epochs
 
 # Make sure this is consistent with obspy.io.stationxml! Importing it
 # from there results in hard to resolve cyclic imports.
@@ -894,20 +893,28 @@ class Inventory(ComparingObject):
         return fig
 
     def get_epoch_plottable_struct(self):
-        height = 0
         plot_dict = {}
-        ch_y = 1;
-        # get height of objects
-        for network in self.networks:
-            sta_dict = network.get_epoch_plottable_struct(y_offset=ch_y)
-            # +2 to represent value after top of bounding rectangle
-            height += len(sta_dict) + 2
-            plot_dict.update(sta_dict)
-            ch_y += height
+        sub_dict = {}
+        for network in set(self.networks):
+            eps = network.get_epoch_plottable_struct()
+            for key in eps.keys():
+                if key not in sub_dict.keys():
+                    sub_dict[key] = []
+                sub_dict[key]+=eps[key]
+        if hasattr(self, 'start_date'):
+            start = self.start_date
+            if self.end_time is None:
+                end = UTCDateTime.now()
+            else:
+                end = min(self.end_time, UTCDateTime.now())
+        else:
+            start = -1
+            end = -1
+        plot_dict[str('')] = [(start, end, sub_dict)]
         return plot_dict
 
     def plot_epochs(self, outfile=None):
-        plot_dict = self.get_epoch_plottable_structure()
+        plot_dict = self.get_epoch_plottable_struct()
         plot_inventory_epochs(plot_dict, outfile)
 
 
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -892,6 +892,19 @@ class Inventory(ComparingObject):
 
         return fig
 
+    def get_epoch_plottable_struct(self):
+        height = 0
+        plot_dict = {}
+        ch_y = 1;
+        # get height of objects
+        for station in self.stations:
+            sta_dict = s.get_epoch_plottable_struct(y_offset=ch_y)
+            # +2 to represent value after top of bounding rectangle
+            height += len(sta_dict) + 2
+            plot_dict.update(sta_dict)
+            ch_y += height
+        return plot_dict
+
 
 if __name__ == '__main__':
     import doctest
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -892,11 +892,14 @@ class Inventory(ComparingObject):
 
         return fig
 
-    def get_epoch_plottable_struct(self):
+    def _get_epoch_plottable_struct_(self):
+        # get structure of inventory epoch's data in format
+        # {inventory: (start, end, subdictionary)}
+        # where subdictionary is recursively the same for netwk, stn, ch data
         plot_dict = {}
         sub_dict = {}
         for network in set(self.networks):
-            eps = network.get_epoch_plottable_struct()
+            eps = network._get_epoch_plottable_struct_()
             for key in eps.keys():
                 if key not in sub_dict.keys():
                     sub_dict[key] = []
@@ -914,6 +917,12 @@ class Inventory(ComparingObject):
         return plot_dict
 
     def plot_epochs(self, outfile=None):
+        """
+        Plot the epochs of this given inventory object.
+        :param outfile: If included, the plot will be saved to a file with the
+            given filename. (Otherwise it will be displayed in a window)
+        :type outfile: str
+        """
         plot_dict = self.get_epoch_plottable_struct()
         plot_inventory_epochs(plot_dict, outfile)
 
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -21,6 +21,7 @@ import textwrap
 import warnings
 
 import obspy
+from obspy.core.inventory.util import plot_inventory_epochs
 from obspy.core.util.base import (ENTRY_POINTS, ComparingObject,
                                   _read_from_plugin, NamedTemporaryFile,
                                   download_to_file)
@@ -897,13 +898,17 @@ class Inventory(ComparingObject):
         plot_dict = {}
         ch_y = 1;
         # get height of objects
-        for station in self.stations:
-            sta_dict = s.get_epoch_plottable_struct(y_offset=ch_y)
+        for network in self.networks:
+            sta_dict = network.get_epoch_plottable_struct(y_offset=ch_y)
             # +2 to represent value after top of bounding rectangle
             height += len(sta_dict) + 2
             plot_dict.update(sta_dict)
             ch_y += height
         return plot_dict
+
+    def plot_epochs(self, outfile=None):
+        plot_dict = self.get_epoch_plottable_structure()
+        plot_inventory_epochs(plot_dict, outfile)
 
 
 if __name__ == '__main__':
--- a/obspy/core/inventory/network.py
+++ b/obspy/core/inventory/network.py
@@ -18,11 +18,13 @@ import copy
 import fnmatch
 import warnings
 
-from obspy.core.inventory.util import plot_inventory_epochs
+import obspy.core.utcdatetime as utc
+
 from obspy.core.util.obspy_types import ObsPyException, ZeroSamplingRate
 
 from .station import Station
-from .util import BaseNode, _unified_content_strings, _textwrap
+from .util import (BaseNode, _unified_content_strings, _textwrap,
+                   plot_inventory_epochs)
 
 
 @python_2_unicode_compatible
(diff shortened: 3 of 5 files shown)


megies:
Mind posting an example image?

amkearns-usgs:
figure_1

obspy/core/inventory/util.py
```diff
@@ -859,6 +862,139 @@ def _seed_id_keyfunction(x):
...
+    plt.xticks(x_ticks, x_tick_labels, rotation='45')
+    plt.grid()
+
+    if outfile:
+        plt.savefig(outfile)
+    else:
+        plt.show()
+
+
+def _plot_traversal_helper_(plot_dict, y_dict, offset=0, prefix=''):
```
megies: no trailing underscore with method/function names, please, to be in line with rest of code base

obspy/core/inventory/util.py
```diff
@@ -859,6 +862,139 @@ def _seed_id_keyfunction(x):
...
+            (start, end, sub_dict) = epoch_tuple
+            offset = _plot_traversal_helper_(sub_dict, y_dict, offset=offset,
+                                             prefix=label)
+        if height == 0:
+            height = offset - current_offset
+        y_dict[label] = (current_offset, height)
+    return offset
+
+
+def _plot_builder_(ax, plot_dict, y_dict, xmin, xmax, clrs, pfx=''):
```
megies: see above..

obspy/core/inventory/station.py
```diff
@@ -507,6 +508,34 @@ def plot(self, min_freq, output="VEL", location="*", channel="*",
 
         return fig
 
+    def _get_epoch_plottable_struct_(self):
```
megies: see below.. no trailing underscore

obspy/core/inventory/network.py
```diff
@@ -675,6 +676,38 @@ def plot_response(self, min_freq, output="VEL", station="*", location="*",
 
         return fig
 
+    def _get_epoch_plottable_struct_(self):
```
megies: see below.. no trailing underscore

obspy/core/inventory/inventory.py
```diff
@@ -892,6 +892,40 @@ def plot_response(self, min_freq, output="VEL", network="*", station="*",
 
         return fig
 
+    def _get_epoch_plottable_struct_(self):
```
megies: see below.. no trailing underscore

obspy/core/inventory/channel.py
```diff
@@ -365,6 +367,27 @@ def plot(self, min_freq, output="VEL", start_stage=None, end_stage=None,
             unwrap_phase=unwrap_phase, plot_degrees=plot_degrees, show=show,
             outfile=outfile)
 
+    def _get_epoch_plottable_struct_(self):
```
megies: see below.. no trailing underscore

megies:
This looks promising.. here's a few comments:

 - Looks like you're doing manual ticking on the x-Axis? This is in general not advisable as it might not update as expected when people zoom into plots etc etc. Can you have a look at how this is handled in `obspy/imaging/waveform.py` in `WaveformPlotting.__plot_set_x_ticks()` which is using helper routine `_set_xaxis_obspy_dates(ax)`. Should be a one-liner to reuse that helper.
- I'm not sure what units you are using on the x-Axis, but it looks like you might simply be converting `UTCDateTime` objects with `float(..)`. If that's correct, it would mean you're using POSIX timestamps on the x-Axis. But ideally you should be using matplotlib's time-stamping conventions in plots, i.e. do something like (assumgin `my_time` is a `UTCDateTime` instance): `from matplotlib.dates import date2num; x = date2num(my_time.datetime)`. Obviously this is also needed for automatic x-ticking like mentioned above

I think it would be nice to have additional logical grouping with colors to guide the eyes..
 - you could use something like `matplotlib.patches.Rectangle(fill=True, zorder=-5, alpha=0.2, facecolor=...)` to give a slightly transparent background color to e.g. all stations from the same network
- line widths should be increased I think, maybe to `lw=2` or even `lw=3`
- arbitrary color-cycling through each individual line does not bring a big advantage I think, maybe it might be an idea to use pre-defined color depending on sampling rate of channel? That way one could immediately see if a station has a high-sampling-rate channel available or not.

Let me know if above comments are unclear.

megies:
See circle CI log for things that need fixing for PEP8 compliance:

```
obspy/clients/httpproxy.py:115:9: E741 ambiguous variable name 'l'
obspy/clients/httpproxy.py:116:12: E741 ambiguous variable name 'l'
obspy/clients/earthworm/waveserver.py:242:13: E741 ambiguous variable name 'l'
obspy/clients/filesystem/tests/test_sds.py:222:17: E722 do not use bare except'
obspy/clients/seedlink/tests/test_basic_client.py:44:13: E722 do not use bare except'
obspy/core/trace.py:2415:9: E722 do not use bare except'
obspy/core/inventory/inventory.py:398:13: E722 do not use bare except'
obspy/core/inventory/inventory.py:910:23: F405 'UTCDateTime' may be undefined, or defined from star imports: future.builtins
obspy/core/inventory/inventory.py:912:42: F405 'UTCDateTime' may be undefined, or defined from star imports: future.builtins
obspy/core/inventory/network.py:688:30: E225 missing whitespace around operator
obspy/core/inventory/network.py:692:23: F405 'UTCDateTime' may be undefined, or defined from star imports: future.builtins
obspy/core/inventory/network.py:694:42: F405 'UTCDateTime' may be undefined, or defined from star imports: future.builtins
obspy/core/inventory/response.py:819:17: E722 do not use bare except'
obspy/core/util/testing.py:465:9: E722 do not use bare except'
obspy/imaging/cm.py:321:5: E741 ambiguous variable name 'l'
obspy/io/arclink/inventory.py:75:9: E722 do not use bare except'
obspy/io/arclink/inventory.py:89:9: E722 do not use bare except'
obspy/io/arclink/inventory.py:176:5: E722 do not use bare except'
obspy/io/arclink/inventory.py:192:5: E722 do not use bare except'
obspy/io/arclink/inventory.py:885:5: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:145:9: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:171:9: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:228:13: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:254:13: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:265:13: E722 do not use bare except'
obspy/io/mseed/scripts/recordanalyzer.py:278:13: E722 do not use bare except'
obspy/io/xseed/utils.py:113:5: E741 ambiguous variable name 'l'
obspy/signal/freqattributes.py:61:28: E741 ambiguous variable name 'l'
obspy/signal/freqattributes.py:81:9: E741 ambiguous variable name 'l'
obspy/signal/freqattributes.py:82:9: E741 ambiguous variable name 'l'
obspy/signal/rotate.py:106:5: E741 ambiguous variable name 'l'
```

obspy/core/inventory/util.py
amkearns-usgs: I was doing that as a means of marking the functions that weren't intended to be callable by the user. Is the leading underscore still acceptable, or should I remove that as well?

megies: A single leading underscore is correct Python practice for private routines, just get rid of the trailing one please.

megies:
Another thought.. instead of colors for sampling rate, we could also use linestyles to represent sampling rates.. e.g. dotted line for high sampling rates, dashed line for intermediate rates etc.  (it's possible to [customize line stroking pattern](http://matplotlib.org/1.2.1/examples/pylab_examples/dash_control.html)), that might be even more intuitive

obspy/core/inventory/util.py
dsentinel: Leading underscore denotes "private" function. Leave it.

amkearns-usgs:
I've made most of the changes requested here so far, though I wouldn't be surprised if some of the specific parameters such as dash length will call for additional tuning.

How do we want to handle the question of coloring the lines and rectangles? While the plotter gives a new color to each line by default, all the rectangles have the same shade of blue to them.

dsentinel:
I would avoid broken lines. May be interpreted as gaps. Even though this is from an inventory/metadata, people may carry over from time series coverage plots. Maybe, maybe it works.

megies:
@amkearns-usgs Can you post an updated image?

amkearns-usgs:
Here's a cap of what it looks like now. I may still change the per-station colors to all match, so GR.WET.* all has the same color, for example. 
figure_1

obspy/core/inventory/inventory.py
```diff
@@ -911,8 +912,8 @@ def _get_epoch_plottable_struct(self):
             else:
                 end = min(self.end_time, UTCDateTime.now())
```
megies: I think we should plot what really is given for the endtime of Inventories etc. and not cut them off at current time. I'm assuming you did this to get the right end of the plot at current time? This can also be achieved by just setting the right x limit to current time at the very end of all plotting before popping up the plot window. But when people zoom out interactively, the actual end time can be seen.

For open ended epochs we could just leave away the end marker and have them end at current time

megies:
This is going in a good direction I think, I like the marker symbols at the epoch start/ends.

Here's some more comments, might be worth to give them a try how they look:
 - I'd propose to use different markers for start/end of epoch, maybe triangles pointing left/right? (see https://stackoverflow.com/a/23345696)
 - colored lines are sometimes bad for visibility, maybe it's enough to color the markers by station and use all black for epoch lines?
 - I'm not sure I like the grid, it interferes with the epoch lines a bit.. maybe we could try without
 - using a monospaced font for y tick labels with SEED codes would be good
 - also.. it seems that location code is omitted from the SEED labels? At least there should be one more dot in the labels (all of those stations have empty location codes)

The plot looks pretty good already, but we might need a different strategy for larger inventories. Probably if number of lines in the plot reaches a certain threshold we have to get rid of individual lines per channel and might have to group by component codes e.g. Z/N/E (or Z/1/2, etc. etc.) if they have matching epoch times (which they usually will have, I'd wager).

amkearns-usgs:
Does the default data when I call read_inventory() have channel location codes? They seem to be blank.

krischer:
Blank location codes are also valid location codes.

amkearns-usgs:
Yeah, black lines on colored rectangles is definitely a lot more visible

figure_1

amkearns-usgs:
For replacing the caps with pointed arrows, I'm guessing we'd want the tips to be at the ends of the lines so they don't overlap, and we want them pointing to the ends of the ranges <|----|>?

amkearns-usgs:
Still need to add merging for large plots where multiple channels exist on the same epoch lines. Otherwise I think all the current requests are being fulfilled in the code right now.

megies:
I like how the plot is coming along, good work so far. It's much improved already, I'd say. 👍 

> For replacing the caps with pointed arrows, I'm guessing we'd want the tips to be at the ends of the lines so they don't overlap, and we want them pointing to the ends of the ranges <|----|>?

I would recommend a triangle marker pointing right for start of epoch and pointing left for end of epoch (i.e. pointing inwards). If they point outwards that could be easily misinterpreted, I think.

> Still need to add merging for large plots where multiple channels exist on the same epoch lines. Otherwise I think all the current requests are being fulfilled in the code right now.

I would start with grouping component codes with identical epoch time ranges. This will already give a reduction of lines by a factor of three. You could label such combined lines with e.g. `GR.FUR..HH[ZNE]`

amkearns-usgs:
Channels with matching epochs and sample rates are now by default plotted on the same y-axis line; this is controlled with a boolean that can be set to false to get each channel on a separate line.

megies:
Sounds good.. can you post an updated image?

amkearns-usgs:
figure_1

(Without the grouping enabled it looks the same as the previous plot above)

krischer:
Starting to look pretty nice :) I think grouping on the band code (first letter of the channel code) would also be a good idea. Most channels are just decimated versions of some other channel so they will most like share epochs. So maybe grouping by `GR.FUR..[VLHB]H[ENZ]` in the above example would work and it would make very large inventories significantly easier to read.

megies:
> So maybe grouping by GR.FUR..[VLHB]H[ENZ] in the above example would work

Maybe this could be incorporated into the existing `combine` option..
 - `combine=None`  every channel separately
 - `combine='component'`  (Default?)  combine e.g. `GR.FUR..HH[ENZ]` etc.
 - `combine='station'`   combine e.g. `GR.FUR..[VLBH]H[ENZ]`

Other things I noticed while actually trying the plot myself now:
 - the code is still buggy, you need to try with a real world inventory
 - these lines raise an exception for me: https://github.com/obspy/obspy/pull/2023/files#diff-77b7d2658a061118d2c7c776b3388c61R1011 you need to cast the sampling rate objects to str or do something else

screenshot from 2017-12-21 10-31-36
--
screenshot from 2017-12-21 10-31-04
--
screenshot from 2017-12-21 10-28-43

amkearns-usgs:
I'd like to have more data to test against. Right now I'm just looking at the default data that gets loaded in when I call obspy.read_inventory(). It looks like this is the only built-in data that obspy has, right? If not I can certainly go looking for some sXML files to test on.

krischer:
You can assemble arbitrary large inventory objects very easily using ObsPy's fdsnws client: https://docs.obspy.org/packages/obspy.clients.fdsn.html

amkearns-usgs:
I have pushed a few changes that should hopefully fix at least a couple of these issues. As I am about to leave on holiday I assume several of them are still not fixed. Also, for a channel where the sample rate changes between epochs, the plot still uses the sample rate for the first epoch in all of the dashing.

amkearns-usgs:
Text for large inventory objects was overlapping, so I added a couple lines to resize the plot based on the length of the labels and how many labels there are. Graphs can get potentially very tall even with the merging of equivalent epochs.

I've also fixed some of the issues plaguing the merging code. It behaves much better now.

plot

amkearns-usgs:
Just wanted to check to see whether or not the current plotting result is meeting your needs or not.

krischer:
Yea I really like the current looks to be honest! The different line styles might no longer be truly necessary as they don't really convey any information (or am I missing something) but that is also to some extent a stylistic choice.

For this to be integrated into ObsPy it still requires tests. Have a look at the existing image tests (https://github.com/obspy/obspy/blob/5762fb5097404dbb04bf91fba6ecacb1075f0bdf/obspy/core/tests/test_network.py#L142) to get an idea for how that is done.

Please make sure to use the latest matplotlib to generate the baseline images.

megies:
> they don't really convey any information (or am I missing something?)

they're grouping channels based on sampling rates. dotted for high sampling rates, dashed for intermediate and solid for low sampling rates

I think it's gonna be tricky to select appropriate figure sizes and axes sizes depending on how many items will be shown in the plot.. but in general I also like the looks a lot right now.

amkearns-usgs:
For these image comparison tests, I would just be saving an image with some given data and making sure that the saved image matches what's being produced by the program, then?

amkearns-usgs:
For these image comparison tests, I would just be saving an image with some given data and making sure that the saved image matches what's being produced by the program, then?

megies:
> For these image comparison tests, I would just be saving an image with some given data and making sure that the saved image matches what's being produced by the program, then?

You can copy paste another image comparison test, should be self-explaining. After you run the test, you'll find the image that was created during tests in `.../tests/images/testrun/` and can move it down one directory to use it as the initial/new base line image.

When you add/update image files, it's a good idea to do it in a separate commit (without other code changes). That way, if images get changed during the PR, we can easily drop older image commits before merging the PR (to keep the git history small).

amkearns-usgs:
I've written up a test case, but I'm having trouble getting the tests to run. Getting an error like this when I try to do `python3 runtests.py`:

  File "[obspy directory]/obspy/core/util/libnames.py", line 94, in _load_cdll
    cdll = ctypes.CDLL(str(libpath))
  File "/usr/lib/python3.5/ctypes/__init__.py", line 347, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: [obspy directory]/obspy/core/util/../../lib/libmseed_Linux_64bit_py35.cpython-35m-x86_64-linux-gnu.so: cannot open shared object file: No such file or directory

obspy/core/tests/test_inventory.py
```diff
@@ -513,6 +513,19 @@ def test_combined_station_event_plot(self):
...
+        """
+        Tests the plotting of epochs defined in inventory objects as time series
+        """
+        inv = read_inventory()
+        with ImageComparison(self.image_dir,
+                             'inventory-epoch-plot.png',
+                             reltol=reltol) as ic:
+            rcParams['savefig.dpi'] = 72
+            fig = inv.plot_epochs(show=False)
+            cat.plot(outfile=ic.name, fig=fig)
```
megies: I guess you want to replace this line with `fig.savefig(ic.name, ...)`

megies:
> I've written up a test case, but I'm having trouble getting the tests to run. Getting an error like this when I try to do python3 runtests.py:

Looks like your installation is broken..? I guess you did a dev install using pip? You could try to reinstall in that case: `$ pip uninstall obspy && pip install -e /path/to/obspy/source/tree`. Also, you can run those tests doing `$ obspy-runtests core`

megies:
Some unrelated Flake8 fails should go away if you rebase on current master..

amkearns-usgs:
Looking over this with the changes you've mentioned (and running the `obspy-runtests core` command) I find that I'm failing a code formatting test due to the line `import matplotlib.pyplot as plt` that I have near the start of util.py. What is the preferred practice for dealing with this sort of import?

megies:
@amkearns-usgs these imports (that potentially set the matplotlib backend) should be done inside the routines that need that import. See e.g. https://github.com/amkearns-usgs/obspy/blob/fbb1ad7c04e1935d48e6fa4175fb6e745ddac7e4/obspy/imaging/waveform.py#L222-L230

amkearns-usgs:
I have the test case set up and the image being compared against in the testing images subdirectory, but for some reason the test is failing on my end. Not sure why. (That said, I wouldn't be surprised to find that it was something simple like the generated plot shown in a window has a grey background, but the saved image has a white one.)

megies:
Uhm.. did you attempt a rebase? Hmm, or rather it looks like you just merged master branch? In any case, the history of your branch is utterly messed up, it really needs a clean up..

megies:
Regarding the test fails, there seems to be a problem with your baseline images: http://tests.obspy.org/92654/#1 and see https://raw.githubusercontent.com/amkearns-usgs/obspy/a0c2998d3c437e08a8d6466123db1ac51d16a104/obspy/core/tests/images/inventory-epoch-plot.png which also shows an error

amkearns-usgs:
OK. I wasn't able to totally fix the history pollution but I have managed to get the png upload working properly and the test now passes.

obspy/core/inventory/util.py
```diff
@@ -864,7 +863,7 @@ def _seed_id_keyfunction(x):
     return x
 
 
-def plot_inventory_epochs(plot_dict, outfile=None, colorspace=None, show=True,
+def plot_inventory_epochs(plot_dict, outfile=None, colorspace=None, show=False,
```
megies: I think usually we have `show=True` in all plots? I.e. having interactive plots pop up on default..

megies:
@amkearns-usgs I wanted to try this one with a real world inventory and ran into an exception..

```python
client = Client()
inv = client.get_stations(network='IU')
inv.plot_epochs()
```

---

```
  File "/home/megies/git/obspy/obspy/core/inventory/inventory.py", line 953, in plot_epochs
    combine)
  File "/home/megies/git/obspy/obspy/core/inventory/util.py", line 896, in plot_inventory_epochs
    mg_dict = _combine_same_epochs(plot_dict)
  File "/home/megies/git/obspy/obspy/core/inventory/util.py", line 963, in _combine_same_epochs
    return _create_same_epochs_string(merge_dict, epochs_dict)
  File "/home/megies/git/obspy/obspy/core/inventory/util.py", line 1011, in _create_same_epochs_string
    loc = split[2]
```
---

Also, when I do the plot with empty location codes, they show up in the axis labels as `_` it seems..

---

One more thing.. it can be sometimes hard to see relatively short interruptions in channels, maybe it would be better to see if we changed the plot to one shaded box per epoch, rather than one box with overall start/end of channels?

megies:
This might yet take a bit of discussion and polishing, I think, so I will bump it to 1.3.0. Feel free to change back to 1.2.0 if you want to push this PR hard in the coming days @amkearns-usgs.

megies:
Alright, no worries we can slate it for 1.3.0

amkearns-usgs:
It's been a while since I've had time to look at this. Has the underlying structure for hardware epochs changed at all?

amkearns-usgs:
Oh, here's a question related to a recent comment -- I wanted to use "_" to represent empty location codes to make it clear that an empty location field might exist in the case where we were plotting multiple channels' data on the same line. While it makes sense to use an empty string when only one channel is plotted to a line, it becomes ambiguous in the second case where some channels might have non-empty location codes, in which case having some symbol makes it easier to tell where locationless channels have been placed. If you have any recommendations on this I'd be interested to hear them.

krischer:
> It's been a while since I've had time to look at this. Has the underlying structure for hardware epochs changed at all?

No.


> Oh, here's a question related to a recent comment -- I wanted to use "_" to represent empty location codes to make it clear that an empty location field might exist in the case where we were plotting multiple channels' data on the same line. While it makes sense to use an empty string when only one channel is plotted to a line, it becomes ambiguous in the second case where some channels might have non-empty location codes, in which case having some symbol makes it easier to tell where locationless channels have been placed. If you have any recommendations on this I'd be interested to hear them.

I don't fully understand the question in the context of the plots to be honest. Can you clarify a bit?

The community in general seems to have arrived at the conclusion that an empty location code is still valid and `_` would be different from an empty one. So I'd recommend to leave the empty code and in case multiple location codes are present in a single line, just use `*`?

Issue #2004, opened 2017-11-16
Add support to sc3ml 0.10 schema for stations


claudiodsf:
Any plans to add support to [sc3ml 0.10 schema](http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.10/)?

Jollyfant:
Looks like the only change in the 0.10 schema is the introduction of a `ResponseIIR` stage. But it's not really used so it should be trivial almost?

@andres-h

megies:
also see #1873 for event data

claudiodsf:
To give some context, we had to move to a development version of SC3, since they [recently introduced](https://github.com/SeisComP3/seiscomp3/commit/135092b96055f4f419618bebe30907de60ba50e7) decimation attributes for PAZ stages, which we do have in our metadata.

Jollyfant:
@claudiodsf can you post an example of metadata in the 0.10 format with the decimation attributes? Also, do you have one with the newly introduced responseIIR element?

claudiodsf:
@Jollyfant Sorry for reeeeaaaly long delay. Attached is an example.
[OVPF.BOR1.dataless.fdsnwml2inv.xml.zip](https://github.com/obspy/obspy/files/1889152/OVPF.BOR1.dataless.fdsnwml2inv.xml.zip)

Jollyfant:
We can represent IIR filters using the `PolesZerosResponseStage` right? Something like this for the example you provided as a mapping.. anyone confirm?

    <responseIIR publicID="ResponseIIR/20171214124152.521971.13" name="ResponseIIR/20171214124152.521971.13">
      <type>D</type>
      <gain>1</gain>
      <gainFrequency>1</gainFrequency>
      <decimationFactor>0</decimationFactor>
      <delay>0</delay>
      <correction>0</correction>
      <numberOfNumerators>1</numberOfNumerators>
      <numberOfDenominators>1</numberOfDenominators>
      <numerators>1</numerators>
      <denominators>0.999999</denominators>
    </responseIIR>

	ResponseIIR/20171214124152.521971.13 
	From COUNTS to COUNTS
	Stage gain: 1.0, defined at 1.00 Hz
	Decimation:
		Input Sample Rate: 100.00 Hz
		Decimation Factor: 0
		Decimation Offset: 0
		Decimation Delay: 0.00
		Decimation Correction: 0.00
	Transfer function type: DIGITAL (Z-TRANSFORM)
[...]

PR #2024, opened 2017-12-08
Add event support to SC3ML 0.10 schema

Brtle:
Hey! I looked a bit to the SC3ML 0.10 event support (see https://github.com/obspy/obspy/issues/2004) and it should be pretty good. There are some changes on the `Arrival/weight` mapping and some fields have been added. Event types have also been added to better match QuakeML. The rest concerns the inventory part that I didn't touch.

I made some changes to the XSLT files provided by SeisComP3 and as they added them to their [Github repository](https://github.com/SeisComP3/seiscomp3/tree/master/src/trunk/libs/xml), I did a [pull request](https://github.com/SeisComP3/seiscomp3/pull/158). It is still open so changes can still be made to this PR.

Morevover, the SC3ML files are always written in SC3ML 0.10. I didn't let the possibility to write files in SC3ML 0.9. What do you think about that? Is it worth to keep the choice of the version at the cost of a slightly greater complexity? I do not know if many people want to write in an old SC3ML version.

### DIFF
--- a/obspy/io/seiscomp/core.py
+++ b/obspy/io/seiscomp/core.py
@@ -24,7 +24,7 @@ from obspy.io.quakeml.core import _xml_doc_from_anything
 
 
 # SC3ML version for which an XSD file is available
-SUPPORTED_XSD_VERSION = ['0.3', '0.5', '0.6', '0.7', '0.8', '0.9']
+SUPPORTED_XSD_VERSION = ['0.3', '0.5', '0.6', '0.7', '0.8', '0.9', '0.10']
 
 
 def _is_sc3ml(path_or_file_object):
@@ -73,7 +73,7 @@ def _is_sc3ml(path_or_file_object):
     return match is not None
 
 
-def validate(path_or_object, version='0.9', verbose=False):
+def validate(path_or_object, version=None, verbose=False):
     """
     Check if the given file is a valid SC3ML file.
 
@@ -87,6 +87,9 @@ def validate(path_or_object, version='0.9', verbose=False):
     :rtype: bool
     :return: `True` if SC3ML file is valid.
     """
+    if version is None:
+        version = SUPPORTED_XSD_VERSION[-1]
+
     if version not in SUPPORTED_XSD_VERSION:
         raise ValueError('%s is not a supported version. Use one of these '
                          'versions: [%s].'
--- a/obspy/io/seiscomp/core.py
+++ b/obspy/io/seiscomp/core.py
@@ -87,20 +87,6 @@ def validate(path_or_object, version=None, verbose=False):
     :rtype: bool
     :return: `True` if SC3ML file is valid.
     """
-    if version is None:
-        version = SUPPORTED_XSD_VERSION[-1]
-
-    if version not in SUPPORTED_XSD_VERSION:
-        raise ValueError('%s is not a supported version. Use one of these '
-                         'versions: [%s].'
-                         % (version, ', '.join(SUPPORTED_XSD_VERSION)))
-
-    # Get the schema location.
-    xsd_filename = 'sc3ml_%s.xsd' % version
-    schema_location = os.path.join(os.path.dirname(__file__), 'data',
-                                   xsd_filename)
-    xmlschema = etree.XMLSchema(etree.parse(schema_location))
-
     if hasattr(path_or_object, "tell") and hasattr(path_or_object, "seek"):
         current_position = path_or_object.tell()
 
@@ -117,6 +103,27 @@ def validate(path_or_object, version=None, verbose=False):
                 path_or_object.seek(current_position, 0)
             except Exception:
                 pass
+
+    # Read version number from file
+    if version is None:
+        match = re.match(
+            r'{http://geofon\.gfz-potsdam\.de/ns/seiscomp3-schema/([-+]?'
+            r'[0-9]*\.?[0-9]+)}', xmldoc.tag)
+        try:
+            version = match.group(1)
+        except AttributeError:
+            raise ValueError("Not a SC3ML compatible file or string.")
+
+    if version not in SUPPORTED_XSD_VERSION:
+        raise ValueError('%s is not a supported version. Use one of these '
+                         'versions: [%s].'
+                         % (version, ', '.join(SUPPORTED_XSD_VERSION)))
+
+    # Get the schema location.
+    xsd_filename = 'sc3ml_%s.xsd' % version
+    schema_location = os.path.join(os.path.dirname(__file__), 'data',
+                                   xsd_filename)
+    xmlschema = etree.XMLSchema(etree.parse(schema_location))
 
     valid = xmlschema.validate(xmldoc)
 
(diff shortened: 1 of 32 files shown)


obspy/io/seiscomp/core.py
```diff
@@ -87,6 +87,9 @@ def validate(path_or_object, version='0.9', verbose=False):
     :rtype: bool
     :return: `True` if SC3ML file is valid.
     """
+    if version is None:
+        version = SUPPORTED_XSD_VERSION[-1]
```
megies: Could we not also look up the version from the file that is to be validated?

obspy/io/seiscomp/event.py
```diff
@@ -116,7 +116,7 @@ def _write_sc3ml(catalog, filename, validate=False, verbose=False,
     nsmap_ = getattr(catalog, "nsmap", {})
     quakeml_doc = Pickler(nsmap=nsmap_).dumps(catalog)
     xslt_filename = os.path.join(os.path.dirname(__file__), 'data',
-                                 'quakeml_1.2__sc3ml_0.9.xsl')
+                                 'quakeml_1.2__sc3ml_0.10.xsl')
```
megies: No idea.. do we want to be able to write different versions of SC3ML? Probably not and writing the most recent version is OK to keep maintenance down, I guess..

megies:
Looks like you're changing all of the test files to SC3ML 0.10 in their header.. I think we might want to keep at least some of them at 0.9 so that reading that older version stays tested?

obspy/io/seiscomp/core.py
Brtle: You're right! I changed that.

Brtle:
I think most of the unit tests should be done on the latest version. Note that all XSLT files are exactly the same, except the number version. However, it's true that there is a lack of tests for older versions. I added some unit tests to check the [qml-example-1.2-RC3.sc3ml](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/tests/data/qml-example-1.2-RC3.sc3ml) file with older versions. To avoid adding test files (which are already numerous) I modified the version dynamically for each test. Let me know what you think about it and if the old versions are enough tested.

krischer:
Sorry the slow response time - this looks pretty reasonable to me. Is this ready - then I'd give it a more detailed review.

Brtle:
You can take a look! There are 2 things which can eventually be improved. The new SC3ML files are always written in SC3ML 0.10. Do you think we need to add the possibility to write 0.9 files? Moreover, I updated all the unit tests with the 0.10 schema. I added some unit tests to check old versions but it may not be enough. What do you think about it?

krischer:
> You can take a look! There are 2 things which can eventually be improved. The new SC3ML files are always written in SC3ML 0.10. Do you think we need to add the possibility to write 0.9 files? Moreover, I updated all the unit tests with the 0.10 schema. I added some unit tests to check old versions but it may not be enough. What do you think about it?

In the tests for reading the old versions you basically already write old SC3ML files. How about exposing that functionality to the `_write_sc3ml()` function and adding a `version` argument. Then you can also use this for the same tests and already have some reasonable coverage. All xsd schemas for all SC3ML versions are also already present and if you also use these in the tests I would say we could have quite some confidence that SC3ML writing works correctly. All of this seems to pretty easy to do with the XSLT approach so I'd say go for it!



The new fields in SC3ML 0.10 (I guess its mainly the `pdf` fields) could be to custom tags in a separate namespace (see e.g. here for an example: http://docs.obspy.org/tutorial/code_snippets/quakeml_custom_tags.html). At that point ObsPy could directly read/write it. This would require some modification to the XSLT files and I don't have enough experience with these files to judge the effort.

Brtle:
> In the tests for reading the old versions you basically already write old SC3ML files. How about exposing that functionality to the _write_sc3ml() function and adding a version argument. Then you can also use this for the same tests and already have some reasonable coverage. All xsd schemas for all SC3ML versions are also already present and if you also use these in the tests I would say we could have quite some confidence that SC3ML writing works correctly. All of this seems to pretty easy to do with the XSLT approach so I'd say go for it!

The point is that I juste change the namespace. The [qml-example-1.2-RC3.sc3ml](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/tests/data/qml-example-1.2-RC3.sc3ml) is actually the same for every SC3ML version, except the namespace part. I just wanted to test that old versions conversion didn't crash without testing every distinctive feature of each version. If we add the possibility to change the namespace in the `_write_sc3ml` function, this can generate unvalid SC3ML files. This can be done but a warning message should be displayed.

What we can do is maybe test all the versions that a unit test can run, for instance:

```python
def test_read_xslt_origin(self, version=['0.9', '0.10']):
    …

def test_read_xslt_magnitude(self, version=['0.8', '0.9', '0.10']):
    …
```

So old versions should be relatively well tested without exploding the number of unit tests.

> The new fields in SC3ML 0.10 (I guess its mainly the pdf fields) could be to custom tags in a separate namespace (see e.g. here for an example: http://docs.obspy.org/tutorial/code_snippets/quakeml_custom_tags.html). At that point ObsPy could directly read/write it. This would require some modification to the XSLT files and I don't have enough experience with these files to judge the effort.

The namespace is hardcoded in the XSLT files so we can't use the same XSLT file without templating the file itself. The easiest solution seems to keep one file per version. However, we will need to maintain several XSLT files when most people will likely use the latest version. It's easier to maintain in the read plugin since only the namespace change between all sc3ml_0.*__quakeml_1.2.xsl files.

Brtle:
Sorry, it's been a little too long since I did not look into this PR!

I did some modifications to the unit tests. For each read conversion tests, every SC3ML versions for which the file is valid are tested. So old versions are not completly tested but I think this is the best compromise we can do.

I didn't change the `_write_sc3ml` method. I think it can be misleading if we add a parameter `version` which only change the namespace as I did in the unit tests. Invalid SC3ML files could be generated.

> The new fields in SC3ML 0.10 (I guess its mainly the pdf fields) could be to custom tags in a separate namespace (see e.g. here for an example: http://docs.obspy.org/tutorial/code_snippets/quakeml_custom_tags.html). At that point ObsPy could directly read/write it. This would require some modification to the XSLT files and I don't have enough experience with these files to judge the effort.

I had answered a little quickly in my last reply and I did not quite understand it. But you're right, these custom tags should be used during the conversion! When I made the plugin the first time I was not aware of that. Actually there are [other nodes](https://github.com/obspy/obspy/blob/master/obspy/io/seiscomp/data/sc3ml_0.9__quakeml_1.2.xsl#L92) for which it could be used. However, this require some upstream modifications to the XSLT files. Since this doesn't seem completely necessary and it does not only concern the 0.10 schema, it would be perhaps more judicious to make another PR?

megies:
> and it does not only concern the 0.10 schema, it would be perhaps more judicious to make another PR?

yeah, sounds like it can be done in a separate PR if you prefer not to work on it right now

megies:
This looks ready @Brtle? Anything left to do in here?

krischer:
Thanks for this! I'll add a changelog entry directly to the master branch.

PR #2040, opened 2018-01-09
PPSD Fix exact trace cutting; includes test

### LINKED ISSUE #2039: Minor correction in obspy.signal.PPSD
Hi, was trying to do very precise PSD processing of our archive. Unfortunately I need to work with exact samples and expect 48 segments (30 min. each) per day. The PPSD calculation always uses a single sample per segment more than it should and results in to the final daily segment being dropped.

Here is some demo code:

    import obspy
    import numpy as np

    header = {
        "starttime":  obspy.UTCDateTime("2017-01-01T00:00:00"),
        "network": "GB",
        "station": "MCH1",
        "channel": "BHE"
    }

    # 24.5 hours to account for 50% overlap through the next day
    print tr = obspy.Trace(data=np.arange(30 * 60 * 49, dtype=np.int32), header=header)
    >>> GB.MCH1..BHE | 2017-01-01T00:00:00.000000Z - 2017-01-02T00:29:59.000000Z | 1.0 Hz, 88200 samples

    ppsd = obspy.signal.PPSD(
        tr.stats,
        obspy.read_inventory("http://www.orfeus-eu.org/fdsnws/station/1/query?sta=MCH1&cha=BHE&level=response"),
    )

    ppsd.add(tr)

    print len(ppsd._times_processed)
    >>> 47

When we add a single second extra to the trace it will give 48 PSDs as expected but that is incorrect.

Here is a list of the segments it calculates:

    GB.MCH1..BHE | 2017-01-01T00:00:00.000000Z - 2017-01-01T01:00:00.000000Z | 1.0 Hz, 3601 samples
    GB.MCH1..BHE | 2017-01-01T00:30:00.000000Z - 2017-01-01T01:30:00.000000Z | 1.0 Hz, 3601 samples
    ...
    GB.MCH1..BHE | 2017-01-01T22:30:00.000000Z - 2017-01-01T23:30:00.000000Z | 1.0 Hz, 3601 samples
    GB.MCH1..BHE | 2017-01-01T23:00:00.000000Z - 2017-01-02T00:00:00.000000Z | 1.0 Hz, 3601 samples

It shows that the final sample used per segment overlaps and we are using 3601 samples per segment, not 3600 as we should. IMHO It can be fixed easily by subtracting tr.stats.delta from the end time of each segment.

I think it could be changed, what do you reckon @megies?

Jollyfant:
Fixes #2039

### DIFF
--- a/obspy/signal/tests/test_spectral_estimation.py
+++ b/obspy/signal/tests/test_spectral_estimation.py
@@ -677,6 +677,29 @@ class PsdTestCase(unittest.TestCase):
             ppsd.plot_temporal([0.1, 1, 10], filename=ic.name, show=False,
                                **restrictions)
 
+    def test_exclude_last_sample(self):
+
+        header = {
+            "starttime":  UTCDateTime("2017-01-01T00:00:00"),
+            "network": "GR",
+            "station": "FUR",
+            "channel": "BHZ"
+        }
+
+        tr = Trace(data=np.arange(30 * 60 * 49, dtype=np.int32), header=header)
+
+        ppsd = PPSD(
+            tr.stats,
+            read_inventory()
+        )
+        
+        ppsd.add(tr)
+        
+        self.assertEqual(48, len(ppsd._times_processed))
+
+        for i, time in enumerate(ppsd._times_processed):
+            self.assertTrue(UTCDateTime(time) == UTCDateTime("2017-01-01T00:00:00") + (i * 30 * 60))
+
     def test_ppsd_spectrogram_plot(self):
         """
         Test spectrogram type plot of PPSD
--- a/obspy/signal/tests/test_spectral_estimation.py
+++ b/obspy/signal/tests/test_spectral_estimation.py
@@ -692,13 +692,14 @@ class PsdTestCase(unittest.TestCase):
             tr.stats,
             read_inventory()
         )
-        
+
         ppsd.add(tr)
-        
+
         self.assertEqual(48, len(ppsd._times_processed))
 
         for i, time in enumerate(ppsd._times_processed):
-            self.assertTrue(UTCDateTime(time) == UTCDateTime("2017-01-01T00:00:00") + (i * 30 * 60))
+            current = UTCDateTime("2017-01-01T00:00:00") + (i * 30 * 60)
+            self.assertTrue(UTCDateTime(time) == current)
 
     def test_ppsd_spectrogram_plot(self):
         """
--- a/obspy/signal/tests/test_spectral_estimation.py
+++ b/obspy/signal/tests/test_spectral_estimation.py
@@ -687,7 +687,7 @@ class PsdTestCase(unittest.TestCase):
         }
 
         # 49 segments of 30 minutes to allow 30 minutes overlap in next day
-        tr = Trace(data=np.arange(30 * 60 * 49, dtype=np.int32), header=header)
+        tr = Trace(data=np.arange(30 * 60 * 4, dtype=np.int32), header=header)
 
         ppsd = PPSD(
             tr.stats,
@@ -696,7 +696,7 @@ class PsdTestCase(unittest.TestCase):
 
         ppsd.add(tr)
 
-        self.assertEqual(48, len(ppsd._times_processed))
+        self.assertEqual(3, len(ppsd._times_processed))
         self.assertEqual(3600, ppsd.len)
 
         for i, time in enumerate(ppsd._times_processed):
--- a/obspy/signal/tests/test_spectral_estimation.py
+++ b/obspy/signal/tests/test_spectral_estimation.py
@@ -686,6 +686,7 @@ class PsdTestCase(unittest.TestCase):
             "channel": "BHZ"
         }
 
+        # 49 segments of 30 minutes to allow 30 minutes overlap in next day
         tr = Trace(data=np.arange(30 * 60 * 49, dtype=np.int32), header=header)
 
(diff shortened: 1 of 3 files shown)


megies:
Would be cool if you added a test that checks the length of the time series going into the individual PSD calculations.

Jollyfant:
Is there a private keeping track of this..? I can't find it.

megies:
Probably not.. 

Actually, there's a property that keeps the expected number of samples in a slice though..

```python
 348         # Trace length for one psd segment.
 349         self._len = int(self.sampling_rate * self.ppsd_length)
 350         # make an initial dummy psd and to get the array of periods
 351         _, freq = mlab.psd(np.ones(self.len), self.nfft,
 352                            self.sampling_rate, noverlap=self.nlap)
```

..and it's also checked when the individual PSDs are calculated:

```python
 746     def __process(self, tr):
 747         """
 748         Processes a segment of data and save the psd information.
 749         Whether `Trace` is compatible (station, channel, ...) has to
 750         checked beforehand.
 751        
 752         :type tr: :class:`~obspy.core.trace.Trace`
 753         :param tr: Compatible Trace with data of one PPSD segment
 754         :returns: `True` if segment was successfully processed,
 755             `False` otherwise.
 756         """
 757         # XXX DIRTY HACK!!
 758         if len(tr) == self.len + 1:
 759             tr.data = tr.data[:-1]
 760         # one last check..
 761         if len(tr) != self.len:
 762             msg = "Got a piece of data with wrong length. Skipping"
 763             warnings.warn(msg)
 764             print(len(tr), self.len)
 765             return False
```

..and there's also some code that was probably used to handle such 1-sample-too-long slices..

Jollyfant:
So we can check if `ppsd.len` is equal 3600 in my test case? That should be OK.

megies:
> So we can check if ppsd.len is equal 3600 in my test case? That should be OK. (fyi it is 👍)

Ok, so the problem is reduced to later slices (e.g. at the end of daily stream object) having slightly later starttime (by a few samples) due to the malformed slicing.. and thus the last slice being excluded since it's starting some milliseconds later then expected? Nice catch!

Jollyfant:
Yeah that is basically just it, nothing worrisome.

Jollyfant:
Can this go in maintenance_1.1.x since this is essentially a bug fix? Or maybe that's later than 1.1.0.. idk!

obspy/signal/tests/test_spectral_estimation.py
```diff
@@ -677,6 +677,32 @@ def test_ppsd_temporal_plot(self):
...
+
+        header = {
+            "starttime":  UTCDateTime("2017-01-01T00:00:00"),
+            "network": "GR",
+            "station": "FUR",
+            "channel": "BHZ"
+        }
+
+        # 49 segments of 30 minutes to allow 30 minutes overlap in next day
+        tr = Trace(data=np.arange(30 * 60 * 49, dtype=np.int32), header=header)
```
megies: not sure how long this test runs.. but maybe using a smaller number of slices (3? 4?) is equally fit as a regression test? the problem should show right away after a single slice, I guess?

megies:
Personally, I'm OK if this goes into maintenance.

obspy/signal/tests/test_spectral_estimation.py
Jollyfant: Good point! I will do three segments.

Jollyfant:
What going on with CI? It's bailing on unrelated floating point accuracies.

krischer:
> What going on with CI? It's bailing on unrelated floating point accuracies.

Oh this is nasty :( They changed the string representation of arrays with numpy 1.14.0.

krischer:
After looking at it in a bit more detail its maybe not that bad: The printing style can be chosen and has support for a "legacy" mode. I'll open a new PR and then we can rebase the others on it.

Jollyfant:
> After looking at it in a bit more detail its maybe not that bad: The printing style can be chosen and has support for a "legacy" mode. I'll open a new PR and then we can rebase the others on it.

Can we stick that in one central place or do all scripts need updating?

krischer:
I'll push later today - it took a bit longer because there are also a ton of new warnings in numpy 1.14. I put it in the central `obspy-runtests` script and also added configuration files for pytest so it works for those without a lot of effort.

krischer:
Done in #2044.

krischer:
The failures should be gone now if you rebase.

Jollyfant:
Okay! Rebased and force pushed with a changelog entry. We pick `maintenance_1.1.x` or `master`?

krischer:
I would tend to do this in `master` as it does change the calculations (however slightly) but I'll leave it up to @megies .

Jollyfant:
Nah master is fine if we go for biannual releases 🐇

obspy/signal/tests/test_spectral_estimation.py
```diff
@@ -677,6 +677,32 @@ def test_ppsd_temporal_plot(self):
...
+        )
+
+        ppsd.add(tr)
+
+        self.assertEqual(3, len(ppsd._times_processed))
+        self.assertEqual(3600, ppsd.len)
+
+        for i, time in enumerate(ppsd._times_processed):
+            current = UTCDateTime("2017-01-01T00:00:00") + (i * 30 * 60)
+            self.assertTrue(UTCDateTime(time) == current)
```
megies: interesting.. so I guess start times of individual PSD slices were never tested so far.. :-/

Issue #2035, opened 2018-01-07
Correlations not normalised


calum-chamberlain:
The recent correlation implementation does not correctly normalise for moving window correlations. Currently the total data window standard deviations are accounted for, rather than the standard deviations at each step.  A result of this is that, in the example [here](https://gist.github.com/calum-chamberlain/50b80dd0be71916906e33e532654ebe1) correlations are suppressed, with a perfect correlation given a value of ~0.55 rather than 1.0.

The example is a little extensive, but I see two options, either [zero-normalised](https://en.wikipedia.org/wiki/Cross-correlation#Zero-normalized_cross-correlation_(ZNCC)) (as used generally in template matching, where both the mean and standard deviation at each step are accounted for), or [normalised](https://en.wikipedia.org/wiki/Cross-correlation#Normalized_cross-correlation_(NCC)) (where only the standard deviation is accounted for). In the example I adapted the correlate function (or rather a hack thereof) to generate zero-normalised cross-correlations using bottleneck for moving window standard deviations.  I guess we wouldn't want to add bottleneck as a dependency though?

For interest sake, [EQcorrscan](http://eqcorrscan.readthedocs.io/en/latest/submodules/utils.correlate.html) uses zero-normalised cross-correlations and [FastMatchedFilter](https://github.com/beridel/fast_matched_filter/blob/master/fast_matched_filter/src/matched_filter.c#L115) uses the other type.

## System info
* Tested on Ubuntu and MacOS
* conda install, Python 3.5

trichter:
Thanks for raising this issue. Coming from a noise background, where generally the data is much longer than the interesting shifts in the correlation function, this difference didn't play a mayor role. I think, we refrained from implementing the rolling standard deviation, because we feared a negative impact on the performance.

A possibility would be to include `demean='moving'`, `normalize='moving'` options to add support for this normalization. We should compare performance of a plain numpy appoach (e.g. [this one](http://www.rigtorp.se/2011/01/01/rolling-statistics-numpy.html)) against your bottleneck version.

Another kind of ugly work-around for your use case: Tentatively "correct" the normalization by adding a factor (5**0.5, square of quotient of window length) and calculate the correct normalizations only for interesting shifts.

calum-chamberlain:
Ah, fair enough (re noise background) - sorry I didn't weigh in on the original PR that made this change, I think I missed it.

[...]

PR #2042, opened 2018-01-12
Add correlate_template() function with 'full' normalization; deprecate 'domain' keyword in correlate() in favor of 'method' keyword

calum-chamberlain:
### What does this PR do?

This PR adds full normalisation to cross-correlations as discussed in #2035.

I have based this on Master given that it was flagged for 1.2.0 in #2035, and it isn't strictly a bug-fix.

I still need to work out a good way to test for rounding errors - the usual case I use is to download some data from the 2016 M7.8 Kaikoura earthquake as this has enough range to result in floating point errors for many correlation implementations when using a day of data. How do we feel about adding an extra network test that is quite memory intensive...?

## TODO list:
- [x] Add tests for all versions of normalize/demean arguments;
- [x] Find source of differences between EQcorrscan result and obspy
- [x] Allow zero-normalized and normalized in `_normxcorr` using the `demean` argument;
~~- [ ] *(Possibly)* Add network test with Kaikoura dataset (passes locally currently, but unsure how to do it non-locally - currently comparing to EQcorrscan result computed on-the-fly, which would require an EQcorrscan install too...)~~

### PR Checklist
- [x] All tests still pass.


obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
     return c[mid - shift:mid + shift + len(c) % 2]
 
 
-def correlate(a, b, shift, demean=True, normalize=True, domain='freq'):
+def _normxcorr(a, b, shift, domain='freq'):
```
krischer: Please document the arguments.

obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
...
+    if shift is None:
+        shift = mid
+    # Normalize short window before correlation
+    b = (b - b.mean()) / (b.std() * len(b))
+    # We HAVE to pad in this case
+    a = np.pad(a, pad_width=len(b), mode='constant', constant_values=0)
+    if domain == 'time':
+        c = scipy.signal.correlate(a, b, 'valid')
+    else:
+        c = scipy.signal.fftconvolve(a, b[::-1], 'valid')
```
krischer: In this case I'd prefer the construct

```python
if domain == "time":
    ...
elif domain == "freq":
   ...
else:  # pragma: no cover
    raise NotImplementedError("...")
```

This is a bit safer as accidental misspellings will be very obvious.

obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
...
+    denominator = denominator[len(b):-1] - denominator[:-len(b) - 1]
+    np.multiply(move_mean, move_mean, out=move_mean)
+    np.divide(move_mean, len(a), out=move_mean)
+    denominator -= move_mean
+    denominator *= b.std()
+    np.maximum(denominator, 0, out=denominator)
+    np.sqrt(denominator, out=denominator)
+    # Cope with zero divisions? - scikit uses a mask
+    response = np.zeros_like(c, dtype=np.float64)
+    mask = denominator > np.finfo(np.float64).eps
```
krischer: I don't really know a lot about this particular problem so take this comment with a grain of salt. This will be a really small number and even slightly larger numbers might results in numerically unstable calculations. Could you use something like the water level we use for instrument responses? Also is settings those values that are not in the mask to zero the correct approach?

obspy/signal/cross_correlation.py
```diff
@@ -88,8 +125,10 @@ def correlate(a, b, shift, demean=True, normalize=True, domain='freq'):
         The cross-correlation will consist of ``2*shift+1`` or
         ``2*shift`` samples. The sample with zero shift will be in the middle.
     :param bool demean: Demean data beforehand.
-    :param bool normalize: Normalize cross-correlation. A perfect
-        correlation will correspond to the value 1.
+    :param bool normalize: Method for normalization of cross-correlation.
+        Any one of 'naive', 'full', None, or True or False (True == 'naive').
```
krischer: Please use ``` ``'naive'`` ```, ``` ``True`` ```... to get it rendered correctly in the docs.

obspy/signal/tests/test_cross_correlation.py
```diff
@@ -271,6 +271,81 @@ def test_xcorr_pick_correction_images(self):
...
+        np.testing.assert_allclose(cc, result, atol=0.01)
+        shift, corr = xcorr_max(cc)
+        self.assertAlmostEqual(corr, 1.0, 2)
+        self.assertEqual(shift, 0)
+
+    """
+    We need a test that does this, but the only way to do it with long duration
+    data - EQcorrscan tests against Kaikoura data, which we know causes issues
+    for a lot of cross-correlation implementations.
+    """
```
krischer: How big is the data?And could you not just use some random noise with a large spike in the middle?

trichter:
This looks also good to me. Here are my comments after a first superficial review:
* In the test you added, values of cc are all lower than results from EQcorrscan. Also the maximum correlation is a bit lower than 1 which is not expected. You compensate this by setting a very high `atol`. Maybe you missed something in the algorithm? Could you also change the test a bit, so that the shift does not equal zero.
* If len(a) == len(b) the output of correlate should be the same for `demean=True, normalize='naive'` and `normalize='full'`. Am I correct? If yes, you could add a simple test checking it.
* The algorithm only works for len(a) >= len(b). Could you please generalize by switching variables and mirroring the result for len(a) < len(b)? (+test)
* I think it would be nice, if the demean parameter could be passed to _normxcorr. Depending on its value _normxcorr could perform zero-normalized or normalized cross-correlation
* The test for rounding errors should definitely be added. If you need the real data, the test could be skipped by default and only run if triggered. (How to implement this exactly I don't know. We have to ask @krischer or @megies)

obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
...
+    """
+    Normalized cross-correlation
+    """
+    mid = (len(a) + len(b) - 1) // 2
+    if shift is None:
+        shift = mid
+    # Normalize short window before correlation
+    b = (b - b.mean()) / (b.std() * len(b))
+    # We HAVE to pad in this case
+    a = np.pad(a, pad_width=len(b), mode='constant', constant_values=0)
```
trichter: np.pad is new in numpy 1.7. Obspy supports numpy down to 1.6. You could use the _pad_zeros function and later we can delete this and use np.pad everywhere.

obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
...
+    # Calculate moving mean and moving standard deviation
+    # using cumulative sums.
+    c = c[1:-1]
+    move_mean = np.cumsum(a)
+    move_mean = move_mean[len(b):-1] - move_mean[:-len(b) - 1]
+    numerator = c - move_mean * b.mean()
+    denominator = np.cumsum(a ** 2)
+    denominator = denominator[len(b):-1] - denominator[:-len(b) - 1]
+    np.multiply(move_mean, move_mean, out=move_mean)
+    np.divide(move_mean, len(a), out=move_mean)
```
trichter: Found it! This should be len(b).

obspy/signal/cross_correlation.py
```diff
@@ -76,7 +76,44 @@ def _xcorr_slice(a, b, shift, domain='freq'):
...
+    if domain == 'time':
+        c = scipy.signal.correlate(a, b, 'valid')
+    else:
+        c = scipy.signal.fftconvolve(a, b[::-1], 'valid')
+    # Calculate moving mean and moving standard deviation
+    # using cumulative sums.
+    c = c[1:-1]
+    move_mean = np.cumsum(a)
+    move_mean = move_mean[len(b):-1] - move_mean[:-len(b) - 1]
+    numerator = c - move_mean * b.mean()
```
trichter: You took a short-cut at line 87. b.mean() is 0 now. This line does not do anything. Not sure of line 105.

obspy/signal/cross_correlation.py
calum-chamberlain: Can do, I was following the `_xcorr_slice` and `_xcorr_pad` precedent, shall I document them while I'm there (copy and paste)?

obspy/signal/cross_correlation.py
calum-chamberlain: My take on this is that the only time you get zeros in the denominator here is when the input data are zeros - I think it is the assumption that is made in the `naive` approach too, when the `stdev` is set to 1 when it equals zero:
```python
stdev = (np.sum(a ** 2)) ** 0.5 * (np.sum(b ** 2)) ** 0.5
if stdev == 0:
    # set stdev to 1 to prevent division by 0
    # cross-correlation function will have only zeros anyway
    stdev = 1
```

It's what we do in EQcorrscan too, but we raise a warning if it happens when we don't expect it to.

By setting the limit for the mask based on the smallest resolvable 64-Bit float on the machine (the line: `mask = denominator > np.finfo(np.float64).eps`, which was a new function to me) I think we are effectively setting the appropriate water-level. I imagine there might be some cases where there is an instability, but I haven't managed to generate any so far. I also can't find a useful page to state where the limits of those divisions might be.

obspy/signal/tests/test_cross_correlation.py
calum-chamberlain: In all honesty, I have tried with random data with spikes, but I haven't been able to usefully reproduce the amplitude ranges in the data, so gave up.  With float 64 in C the issues only occur after about 16 hours of data during the sequence (sampled at 100Hz), so it is quite a chunk of data required to get this issue. Unfortunately these issues occur exactly when (in one case) you want correlations to be robust: for the detection of early aftershocks or triggered seismicity.
The dataset that I use in previous tests is multi-channel (6-channels, 8640001 samples each) and is 474MB on disk as MiniSeed - looking at top it looks like that inflates to about 1.3GB in RAM.
Testing that locally suggests that we aren't running into floating point rounding errors here.

obspy/signal/cross_correlation.py
calum-chamberlain: Indeed, for some reason I had this condition in an earlier version, but I reworked it and removed it - added the switch back in! (to be pushed!)

obspy/signal/cross_correlation.py
calum-chamberlain: Again, something I removed from scikit, then added back in - you are of course correct, unnecessary and removed (again, to be pushed).

obspy/signal/cross_correlation.py
calum-chamberlain: :+1:  thanks! I did wonder why `np.pad` wasn't used, but will remove.

calum-chamberlain:
Thanks for those reviews @trichter and @krischer, I'm making some changes down here and have fixed  the things I have replied to. Agree that results are not quite as expected and I need to dig into this a little more.
For now I will push up the changes that I have made with tests deliberately failing on the comparison of normalized results - I can ping you @trichter again when I have found the cause of the differences if you want?

trichter:
I meant something else with the one comment. In line 155
```    
np.divide(move_mean, len(long_data), out=move_mean)
```
you have to substitute `len(long_data)` with `len(short_data)` and some problems will go away.

Also, I think I am not correct with the point, that the correlations with different normalizations are the same for len(a)==len(b). But you can check that the full_xcorr == norm_xcorr for the mid sample and that full_xcorr > norm_xcorr for all other samples.

And last but not least here is another input: `correlate` was designed to return a cross-correlation "around lag time 0".  I suspect when you do template matching you are a bit surprised to specify the shift parameter and also you want to have a correlation which has the same length as your data vector. I have two suggestions to deal with this:
1. make `shift` optional and/or accepting keywords like `'same'` and `'valid'`.
2. introduce a completely new function `correlate_template(data, template, mode=, demean=, normalize=)`
What are your thoughts/opinions?

obspy/signal/cross_correlation.py
krischer: Parts of our code base are not yet up to the standard and even for private functions it definitely helps if they are properly documented especially with not immediately obvious parameters. This improves maintainability for the next person touching the code.

obspy/signal/cross_correlation.py
krischer: Sounds reasonable to me. Worst case I guess it flips over to an `inf` and it will be immediately obvious that something is wrong - but looking at the code a bit more I guess that cannot really happen.

obspy/signal/tests/test_cross_correlation.py
krischer: Where does the instability come from? To my this line:

```python
denominator = np.cumsum(long_data ** 2)
```

seems to be a candidate but even with data still in counts and very long time series and a short template I have a hard time imagining this getting into a regime where double precision floating point arithmetics gets inaccurate enough to notice.

The whole problem (if its due to this line) could be mitigated by not using a single cumulative sum but many of them with some special handling of the boundary regions - not sure if its worth the effort to be honest.

krischer: Some short test data that might work: Use some data you know and prepend some very high amplitude random data to it. Then do the CC with a fairly short template and only compare the result for the later part to a known good one. This should "preload" the cumulative sum so rounding errors might become noticeable. To test if this is actually a problematic data set you could run it against a CC implementation that you know has problems with these kind of things.

```python
data = np.concatenate([np.random.random(10) * 1e200, obspy.read()[0].data])
```

krischer:
> 1. make shift optional and/or accepting strings like 'same' and 'valid'.
> 2. introduce a completely new function correlate_template(data, template, mode=, demean=, normalize=)
> What are your thoughts/opinions?

I don't have strong opinions about this but I slightly prefer the first version. But if you think this use case warrants a second function I'd personally also be with it.

trichter:
I think I slightly prefer the second option.

OK, I've been doing some coding on the _normxcorr function and I finally came up with:

```
def _pad_zeros(a, num, num2=None):
    """Pad num zeros at both sides of array a"""
    if num2 is None:
        num2 = num
    hstack = [np.zeros(num), a, np.zeros(num2)]
    return np.hstack(hstack)

def _window_sum(data, window_len):
    window_sum = np.cumsum(data)
    window_sum = window_sum[window_len:] - window_sum[:-window_len]
    return window_sum

def correlate_template(data, template, mode='same', demean=True, normalize='full', domain='freq'):
    N = len(template)
    assert len(data) >= N
    if demean:
        template = template - np.mean(template)
        data = data - np.mean(data)
    if normalize is not None:
        template = template / (np.sum(template ** 2)) ** 0.5
    if domain == 'time':
        c = scipy.signal.correlate(data, template, mode=mode)
    else:
        c = scipy.signal.fftconvolve(data, template[::-1], mode=mode)
    if normalize is None:
        denominator = 1
    elif normalize == 'naive':
        denominator = (np.sum(data ** 2)) ** 0.5
        if denominator == 0:
            denominator = 1
    elif normalize == 'full':
        pad = len(c) - len(data) + N
        if mode == 'same':
            pad1, pad2 = (pad + 2) // 2, (pad-1) // 2
        else:
            pad1, pad2 = (pad + 1) // 2, pad // 2
        data = _pad_zeros(data, pad1, pad2)
        denominator = _window_sum(data ** 2, N)
        if demean:
            denominator -= _window_sum(data, N) ** 2 / N
        denominator = np.sqrt(denominator, out=denominator)
        denominator[denominator == 0] = 1
    return c / denominator

```

Did not push to this brach, because I didn't want to overwrite @calum-chamberlain 's code.

krischer:
> I think I slightly prefer the second option.

Then go for it.

I'll let you and @calum-chamberlain sort out the details of this branch :)

obspy/signal/tests/test_cross_correlation.py
calum-chamberlain: That is exactly the line that would cause the instability - and you are right, one way to mitigate would be to break the data down, but I'm not sure it is worth the effort either. It doesn't seem to be an issue with the Kaikoura data.
By the way, I agree with the issue being less likely with data in counts, but often in template matching the data are filtered before correlation, yielding data in floats.

calum-chamberlain:
> I meant something else with the one comment. In line 155
>
> `np.divide(move_mean, len(long_data), out=move_mean)`
>
> you have to substitute len(long_data) with len(short_data) and some problems will go away.

So that would work, but *should* be less memory efficient for the `domain='time'` option, because the short data would be padded to the length of the long data - for correlations in the frequency domain this is done anyway I think, so it would make no difference there.  I would prefer to maintain it the way I have it now (with setting the shorter of `(a, b)` as `short_data`, but I should probably document this behaviour and say something like:
```python
"""
    :note:
         For ``normalization='full'`` the shorter of ``(a, b)`` will be slid through the long 
         data, such that the zeroth element of the returned array corresponds to the
         correlation of the short input array with the first n elements of the long input 
         array where n is the length of the short input array.
"""
```
*Opinions welcome on that, sounds a little opaque to me.*

---

> Also, I think I am not correct with the point, that the correlations with different normalizations are the same for len(a)==len(b). But you can check that the full_xcorr == naive_xcorr for the mid sample and that full_xcorr > naive_xcorr for all other samples.

Yup, have changed that test to `shift=0` locally, happy days.

---

> 1.    make shift optional and/or accepting strings like 'same' and 'valid'.
> 2.   introduce a completely new function correlate_template(data, template, mode=, demean=, normalize=)
>    What are your thoughts/opinions?

I mildly prefer option 2 as well, but with some doc-string in `signal.cross_correlation.correlate` pointing users to the right place.  I'm not keen on giving the user the option of `naive` normalization with template-matching though, mostly I think it opens up ways for people to get very wrong answers.

---

@trichter how do you want to proceed? Happy to reformat my stuff here with the `correlate_template` function you suggested, or do you want me to push the minor changes I made here so you can add that change?

trichter:
> > I meant something else with the one comment. In line 155
np.divide(move_mean, len(long_data), out=move_mean)
you have to substitute len(long_data) with len(short_data) and some problems will go away.

>So that would work, but should be less memory efficient for the domain='time' option, because the short data would be padded to the length of the long data

Still, we are talking about different issues , here. It just has to be replaced in this one line to get the algorithm correct.

> @trichter how do you want to proceed? Happy to reformat my stuff here with the correlate_template function you suggested, or do you want me to push the minor changes I made here so you can add that change?

OK, I am going to refactor everything a bit and push my changes to your branch.

trichter:
OK, I pushed my changes.

I had to resolve a strange merge conflict in the changelog. Hope it is OK now.

I deleted the doc for the `_xcorr_...` functions again, because its all documented in the correlate function and these two functions are internal. 

@calum-chamberlain can you have a thorough look, if you agree on this implementation?
I left the naive normalization also for the correlate_template function, but the default is full.

It would be nice to have a test against FastMatchedFilter, @calum-chamberlain is it much work for you to get a cross-correlation from this package?

+DOCS

calum-chamberlain:
>         I meant something else with the one comment. In line 155
>         np.divide(move_mean, len(long_data), out=move_mean)
>         you have to substitute len(long_data) with len(short_data) and some problems will go away.
> 
>     So that would work, but should be less memory efficient for the domain='time' option, because the short data would be padded to the length of the long data
> 
> Still, we are talking about different issues , here. It just has to be replaced in this one line to get the algorithm correct.

Oh man, sorry, that was dim, for some reason I misread that and was thinking about the padding line, I don't know why, so apologies (and thanks)!

Also apologies for the need to merge master into this branch, I thought I had updated my master and based on that before creating this branch, obviously not.

What was the reason for removing the docs for the `_xcorr...` functions?

I do agree with this implementation, it's much closer to how I originally had it (but wanted to try the scikit type way just to see), I think this way makes more sense than the way I was going. Thanks! I usually divide the template by `template.std() * len(template)`, but testing a couple of things it doesn't seem to make any difference. I thought it was required to cope with large differences in amplitude between the template and the data, but doing things like:
```python
import numpy as np
from obspy import read
from obspy.signal.cross_correlation import correlate_template

data = read()[0].data
template = data[300:400]
template_large = template * 10e10
template_small = template * 10e-10

cc = correlate_template(data, template)
cc_large = correlate_template(data, template_large)
cc_small = correlate_template(data, template_small)
assert np.allclose(cc, cc_large)
assert np.allclose(cc, cc_small)
```
seems fine.

Will generate a test against FMF and push.

calum-chamberlain:
So, I have a test against FastMatchedFilter, but I was somewhat wrong, FastMatchedFilter does remove the total mean of the templates prior to computing the correlations, but they do not remove the mean of the data from individual windows. If you don't remove the mean of the template in the test examples all correlations are positive. In contrast, if you do, the `demean=False` option gives the same result as FastMatchedFilter, and a different one to EQcorrscan.

My understanding is that FastMatchedFilter has it right, and the template mean should be removed in either the ZeroNormalised (currently `normalize='full', demean=True`) and the Normalized correlations. Can we remove the `demean` argument, enfore removing the mean of the template, and change the `normalize` options to `("zero-normalized", "normalized", "naive")` which `"zero-normalized"` as the default?

calum-chamberlain:
[This commit](https://github.com/obspy/obspy/pull/2042/commits/9ba05f9ebc0aa53e9b53f9dd3a9e1842f5e37d85) has all the changes I think are needed for my previous suggestion, to demonstrate how it could work, if we decide otherwise we should drop this commit.

An aside, I found an issue in my previous test and fixed it in that commit too: doing an inline `template -= template.mean()` was editing the data array too, which I had also done for the FastMatchedFilter result. If we drop the previous commit I should reapply that fix.

trichter:
> This commit has all the changes I think are needed for my previous suggestion, to demonstrate how it could work, if we decide otherwise we should drop this commit.

I am against including this commit. It is for example not possible to calculate a "normal not-normalized, not-demeaned" cross-correlation anymore. Other reason see below.

> So, I have a test against FastMatchedFilter, but I was somewhat wrong, FastMatchedFilter does remove the total mean of the templates prior to computing the correlations, but they do not remove the mean of the data from individual windows. If you don't remove the mean of the template in the test examples all correlations are positive. In contrast, if you do, the demean=False option gives the same result as FastMatchedFilter, and a different one to EQcorrscan.My understanding is that FastMatchedFilter has it right, and the template mean should be removed in either the ZeroNormalised (currently normalize='full', demean=True) and the Normalized correlations. Can we remove the demean argument, enfore removing the mean of the template, and change the normalizeoptions to ("zero-normalized", "normalized", "naive") which "zero-normalized" as the default?

I do not understand that normalization approach of FMF. If you demean the template it does not matter for the nominator if you demean the data (but for the normalization). E.g., consider a demeaned template of  (1, 0, -1) and a part of the data vector (2, 4, 4). The cc is -2. If you subtract any number from the data vector (e.g. the mean) the cc will stay the same, simply because the sum over the template is 0. Consequently, the cc should be normalized with the locally demeaned data to get a cc of 1 for a perfect match. The best example is the test you added. We only have a cc of 0.7 for a perfect fit? This is introduced by demeaning the template, but not normalizing with locally demeaned data.
I tend to say the correct way of calculating normalized cc is ours with demean=False, normalize='full'. If the user really wants it the way FMF does it, he can demean the template beforehand and call the the function with demean=False (as you did for the test).

trichter:
> What was the reason for removing the docs for the _xcorr... functions?

Imho a one-liner for these functions is enough, because all arguments are documented in the public functions. (And it was not entirely correct anyway, just copy-pasted.)

> I do agree with this implementation, it's much closer to how I originally had it (but wanted to try the scikit type way just to see)

This numpy ufunc approach from scikit will probably save some memory. We can implement it later if we like. First let's get the basics correct.

calum-chamberlain:
> I am against including this commit. It is for example not possible to calculate a "normal not-normalized, not-demeaned" cross-correlation anymore. Other reason see below.

Fair enough, but that option in itself is just returning the output from scipy functions, so it is readily accessible.

> I do not understand that normalization approach of FMF... etc.

It is a slightly odd approach; they rely on the assumption that when you have a decent length template of a few hundred samples, the local mean of the data should be zero.  This isn't strictly the case (and we have had a lot of chats about this), but it doesn't change the correlations significantly, as long as you use it as intended (which I'm not doing in this test because I'm using such a short data window).

I also totally agree that the zero-normalized form of cross-correlation is *the* way to do template-matching (hence it being the only option in EQcorrscan).  I'm therefore a little concerned that we are cooking up use-cases that potentially do not exist.  I also think that, if we are going to include "normalized" correlation alongside "zero-normalized" correlation we should just stick to those terms.

I have reverted the commit locally and corrected the FMF comparison result - let me know if you want me to push.

trichter:
>>I am against including this commit. It is for example not possible to calculate a "normal not-normalized, not-demeaned" cross-correlation anymore. Other reason see below.

>Fair enough, but that option in itself is just returning the output from scipy functions, so it is readily accessible.

Yeah, that's true.

>>I do not understand that normalization approach of FMF... etc.

>It is a slightly odd approach; they rely on the assumption that when you have a decent length template of a few hundred samples, the local mean of the data should be zero. This isn't strictly the case (and we have had a lot of chats about this), but it doesn't change the correlations significantly, as long as you use it as intended (which I'm not doing in this test because I'm using such a short data window).

Thanks for this explanation. How I interpret your statement: What they want to do is zero-normalized cross-correlation as EQcorrscan. But because they do it not entirely correct, it is just an approximation. They do not claim to do "normalized cc", too. I could not find anything on the web stating only one time series should be demeaned - be it for zero-normalized or normalized cc.

> I also totally agree that the zero-normalized form of cross-correlation is the way to do template-matching (hence it being the only option in EQcorrscan). I'm therefore a little concerned that we are cooking up use-cases that potentially do not exist. I also think that, if we are going to include "normalized" correlation alongside "zero-normalized" correlation, we should just stick to those terms.
> I have reverted the commit locally and corrected the FMF comparison result - let me know if you want me to push.

If you don't mind: Please push it. In this way we support zero-normalized cc with `demean=True, normalize='full'` (the default) and normalized cc with `demean=False, normalize='full'`(as far as I understand normalized cc, and I think we agree?) In this way the arguments are consistent with the existing `correlate` function. In the docstring we can explain which combination is zero-normalized and which normalized. Also, the demean keyword is pretty self-explanatory.

calum-chamberlain:
> They do not claim to do "normalized cc", too. I could not find anything on the web stating only one time series should be demeaned - be it for zero-normalized or normalized cc.

Agree, its some kind of halfway-house between the two. Maybe someone will find a use for the "normalized" implementation we have here.

I still prefer to use the "normalized" and "zero-normalized", but take the point that keeping function signatures somewhat consistent would be good, so go for it. I think we just have to add docs to document which is "normalized" and which is "zero-normalized" and we might be good to go (apart from below)?

I still haven't added a test to highlight rounding errors, mostly because I don't see any with the Kaikoura data, which makes it hard to make a test that we expect would fail if there were rounding errors. We can add some large amplitude synthetic dataset if we want, but I don't have a good way of knowing if it would be expected to fail when rounding errors are present.

trichter:
> I still haven't added a test to highlight rounding errors, mostly because I don't see any with the Kaikoura data, which makes it hard to make a test that we expect would fail if there were rounding errors. We can add some large amplitude synthetic dataset if we want, but I don't have a good way of knowing if it would be expected to fail when rounding errors are present.

Well, it's just good if it works out of the box. :smile: We don't need that test then. If somebody finds a bug in the future we can deal with it later.

Another thing: While working on this issue I realized that scipy's [correlate](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.correlate.html) function itself now supports a `method` keyword which does the same as our `domain` keyword. I therefore think, we should depreciate `domain` parameter for correlate and add `method` instead. This can just be passed further for scipy>=0.19. This is backwards incompatible but will lead to less pain in the future... pinging @megies, @kirscher?

krischer:
> Another thing: While working on this issue I realized that scipy's correlate function itself now supports a method keyword which does the same as our domain keyword. I therefore think, we should depreciate domain parameter for correlate and add method instead. This can just be passed further for scipy>=0.19. This is backwards incompatible but will lead to less pain in the future... pinging @megies, @krischer?

I'm really against breaking backwards compatibility if there is not a really good reason to do it as we go through quite some pain trying to be as compatible as possible. But given that this is a really new function (just introduced with 1.1.0) and we want to release the next ObsPy April 1st in any case I would be somewhat fine with it here if you think this is a strong enough reason. But please not that ObsPy still has to support Scipy down to 0.11.

trichter:
I am finished for today. It's ready from my point of view. Let's see what CI and the docs buildbot say. Feel free to tune docs and code.

trichter:
Ready for review.

obspy/signal/cross_correlation.py
```diff
@@ -151,22 +187,128 @@ def correlate(a, b, shift, demean=True, normalize=True, domain='freq'):
...
+            data = data - np.mean(data)
+    cc = _call_scipy_correlate(data, template, mode, method)
+    if normalize is None:
+        norm = 1
     else:
-        raise ValueError("domain keyword has to be one of ('freq', 'time')")
-    return _xcorr(a, b, shift, domain=domain) / stdev
+        norm = np.sum(template ** 2)
+        if normalize == 'naive':
+            norm = (norm * np.sum(data ** 2)) ** 0.5
```
krischer: Some things here could happen in place to make it faster at the cost of more ugly code. I'm not sure this is a worthy trade-off here but it would be possible. For many things an optional `numexpr` dependency would also be really useful but I'm not sure we want to go down this road.

obspy/signal/cross_correlation.py
```diff
@@ -33,52 +33,70 @@
...
+    hstack = [np.zeros(num), a, np.zeros(num2)]
     return np.hstack(hstack)
 
 
-def _xcorr_padzeros(a, b, shift, domain='freq'):
+def _call_scipy_correlate(a, b, mode, method):
+    """
+    Call the correct correlate function depending on Scipy version and method
+    """
+    if scipy.__version__ >= '0.19':
```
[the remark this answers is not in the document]
megies: see e.g. https://github.com/obspy/obspy/blob/35d2a0c4d775c6b817b807dc35169590f8d4eb00/obspy/core/util/testing.py#L375

trichter:
I implemented the hints. Because the test case is quite extensive, I dared to switch to in-place operations as much as possible.

obspy/signal/cross_correlation.py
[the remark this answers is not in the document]
trichter: Thanks for the code snippet

obspy/signal/cross_correlation.py
trichter: I changed it to in-place operations with the original lines as comments.

obspy/signal/cross_correlation.py
```diff
@@ -103,9 +130,9 @@ def correlate(a, b, shift, demean=True, normalize=True, domain='freq'):
 
     .. note::
 
-        For most input parameters cross-correlation in frequency domain is much
+        For most input parameters cross-correlation using the FFT is much
         faster.
-        Only for small values of ``shift`` (``⪅100``) time domain
+        Only for small values of ``shift`` (``⪅100``) direct time domain
```
megies: I'm usually in favor of only using ASCII whenever possible.. that symbol is basically ``<=100``?

trichter: Sure, will replace this fancy symbol.

trichter: Ah, well, it is not exactly ``<=100``. It is approx. less than 100. We can also easily delete this note? New scipy has the `auto` method and users can look there.

megies: or just elaborate with words

megies:
> One other small request I have: Could you add something about this PR to this page here (https://github.com/obspy/obspy/wiki/What's-New-in-ObsPy-1.2) - then we don't have to scramble to create this page at release time.

we might want to add this to the github PR template checklist, for major changes to include themselves in the "whats new" page..

calum-chamberlain:
Looks good to me, thank @trichter for your efforts here.

trichter:
Thank you @calum-chamberlain

krischer:
Thanks a lot @calum-chamberlain and @trichter for all your work on this :)

trichter:
@calum-chamberlain using your [notebook](https://gist.github.com/calum-chamberlain/40978a9e4e7cf7d8d1fcd4581416e9de) I compared the performance of the new implementation with bottleneck. It is only marginally slower. The original numpy method failed with a memory error using data corresponding to 1 day sampled at 100Hz.

```python
indeces = np.arange(0, 24*3600, 0.01) # A day of data sampled at 100 Hz
data = signal.gausspulse(indeces - 2000, fc=0.5)
data += 0.3 * np.random.randn(len(data))
template_start, template_stop = (198000, 202000) # Lets use a little longer template too, which means a longer rolling window
template = data[template_start: template_stop]
```
```python
# no normalization
%timeit correlate_template(data, template, normalize=None)
%memit correlate_template(data, template, normalize=None)
```
    1.07 s ± 4.18 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    peak memory: 967.45 MiB, increment: 532.99 MiB

```python
# normalization with bottleneck
%timeit normxcorr(data, template, (len(data) // 2) - (len(template) // 2), demean=True, domain='freq', norm_method='bn')
%memit normxcorr(data, template, (len(data) // 2) - (len(template) // 2), demean=True, domain='freq', norm_method='bn')
```

    1.16 s ± 7.52 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    peak memory: 1000.42 MiB, increment: 565.96 MiB

```python
# normalization with np.cumsum
%timeit correlate_template(data, template)
%memit correlate_template(data, template)
```

    1.36 s ± 7.16 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    peak memory: 901.53 MiB, increment: 467.07 MiB

PR #2044, opened 2018-01-12
Support for numpy 1.14

krischer:
This PR makes the test pass with numpy 1.14 and also mostly gets rid of newly introduced deprecation warnings.

The main issue is that numpy 1.14 changed the way arrays are printed - but they do offer a legacy printing setting so the doctests don't have to be changed. This setup is performed in the main obspy-runtests routine as well as for pytest. Thus some doctests currently cannot be directly executed on numpy 1.14 but I'm not sure if anyone does this and we can fix it as we encounter it.

### DIFF
--- a/obspy/io/cmtsolution/__init__.py
+++ b/obspy/io/cmtsolution/__init__.py
@@ -26,7 +26,7 @@ It works by utilizing ObsPy's :func:`~obspy.core.event.read_events` function.
 
 The event will contain a couple of origins and magnitudes.
 
->>> print(cat[0])  # doctest: +NORMALIZE_WHITESPACES +ELLIPSIS
+>>> print(cat[0])  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
 Event:  2003-12-26T01:56:58.129999Z | +29.100,  +58.240 | 6.54 mw
             resource_id: ResourceIdentifier(id="...")
              event_type: 'earthquake'
--- a/obspy/io/cmtsolution/__init__.py
+++ b/obspy/io/cmtsolution/__init__.py
@@ -22,19 +22,24 @@ It works by utilizing ObsPy's :func:`~obspy.core.event.read_events` function.
 >>> cat = obspy.read_events("/path/to/CMTSOLUTION")
 >>> print(cat)
 1 Event(s) in Catalog:
-2003-12-26T01:56:58.129999Z | +29.100,  +58.240 | 6.54 mw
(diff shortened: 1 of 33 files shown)


krischer:
We should also really setup a test runner against the dev versions of at least Python and numpy so we have a bit more advance warning for these things. Maybe also for scipy. Does anyone know the best way to do this?

Jollyfant:
> We should also really setup a test runner against the dev versions of at least Python and numpy so we have a bit more advance warning for these things. Maybe also for scipy. Does anyone know the best way to do this?

Sounds like something that the docker-testbot would be suitable for if it still works.

megies:
> Sounds like something that the docker-testbot would be suitable for if it still works.

;-) @krischer [nudge nudge, wink wink](https://www.youtube.com/watch?v=AGrvQ1c5khU)

> Does anyone know the best way to do this?


But should equally well be possible in Travis, no? I don't see why pip-installing master of numpy etc. shouldn't be possible in one bleeding-edge job in the matrix..?

obspy/io/cmtsolution/__init__.py
```diff
@@ -22,19 +22,24 @@
 >>> cat = obspy.read_events("/path/to/CMTSOLUTION")
 >>> print(cat)
 1 Event(s) in Catalog:
-2003-12-26T01:56:58.129999Z | +29.100,  +58.240 | 6.54 mw
+2003-12-26T01:56:58.130000Z | +29.100,  +58.240 | 6.54 mw
```
megies: that's funny that new numpy makes this strange rounding-error-in-string-representation go away..

krischer: Yea I think the doctests were actually not executed at all for a while as there was a typo in the pragma of the next doctest. I've only seen a verbose error message when using pytest's doctest runner.

krischer:
Ready for review from my point of view if CI passes.

Jollyfant:
Not sure why CI is failing on that test. My python 2.7 version seems to be doing OK..

Python 2.7.14

    >>> from obspy.core.util.base import create_empty_data_chunk
    >>> out = create_empty_data_chunk(3, 'int', 10)
    >>> out.dtype
    dtype('int64')

Searched through files for `np.fromstring` but it seems like you got most (all) of them. Still some 1.14 deprecations left upstream..

krischer:
Seems to be due to C `long` being different on Windows: https://stackoverflow.com/a/36279549

I'll adjust the test case to allow both variants.

krischer:
CI looks good (OSX builds still have to pass but those take a long time). Someone please review!

Issue #2034, opened 2018-01-05
UTCDateTime oddities in obspy 1.1.0


d-chambers:
version = 1.1.0 on Ubuntu 16.04 with python 3.6 installed via Anaconda

I have run into a few surprising behaviours dealing with UTCDateTime objects that are very close in time . The issues are primarily due to rounding, floating point number weirdness, and a few potential inconsistencies in the UTCDateTime object. The result is some rather surprising behaviours. For example, check out [this gist](https://gist.github.com/d-chambers/e1469318f79a84f5cc829545a60ad218).

Although there may be several issues to address here, the behaviour I rely on that is no longer consistent is this: 
Given the default precision of 6, any datetime object should be equal to a new datetime object created using the year, month, day, hour, minute, second, and microsecond attributes. This would mean the following function should always return True (unless the precision > 6)

```python
import obspy

TIME_NAMES = ('year', 'month', 'day', 'hour', 'minute', 'second',
              'microsecond')


def can_serialize_with_time_params(arg):
    """ return True if using UTC time values (year, month, day, etc.) 
    provide enough information to reconstruct the object. """
    utc = obspy.UTCDateTime(arg)
    time_dict = {x: getattr(utc, x) for x in TIME_NAMES}
    utc2 = obspy.UTCDateTime(**time_dict)
    return utc == utc2
```
 
However, this is not the case. For example

```python
can_serialize_with_time_params(1515174511.1984465)  # False
can_serialize_with_time_params(1515174511.1984463)  # True
```

Am I wrong to assume UTCDateTime objects should behave in this way? I don't mind working on a fix but I wanted to make sure everyone who cares is on the same page.

krischer:
Without looking at this in a lot of detail I would say you are right: This is a bug. I'll delegate it to @barsch for now who has the deepest understanding of our `UTCDateTime` objects.

Another issue with some weirdness due to limited precision in the `UTCDateTime` objects: #1766

megies:
Two comments/observations..

---
First, since you're initializing with a POSIX timestamp, floating point accuracy hits when converting to integer nanoseconds (the new internal representation), since floating point resolution is `1e-15` and thus not reaching down to nanoseconds for the large values of order `~1e9` (notice the last two digits `...92` in the integer nanosecond representation):

```python
In [1]: x = 1515174511.1984465

In [2]: t = UTCDateTime(x)

In [3]: t._ns
Out[3]: 1515174511198446592

```
---

[...]

PR #2045, opened 2018-01-12
PPSD: change internal representation of timestamps to integer nanosecond

megies:
This changes all internally used timestamps in PPSD to POSIX timestamps as integer nanoseconds, eliminating all potential floating point inaccuracies when e.g. comparing time stamps etc.

The only drawback is that npz files written on newer obspy versions will result in exceptions when reading on older obspy versions (the other way around should work fine). But I think the positive effect outweighs this drawback.

### PR Checklist
- [x] ~~If the PR is making changes to documentation, docs pages can be built automatically.
- [x] All tests still pass.
- [x] ~~Any new features or fixed regressions are be covered via new tests.~~ no new features
- [x] ~~Any new or changed features have are fully documented.~~ no new features

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,10 @@ master:
      objects to shapefile (see #2012)
  - obspy.signal.PPSD:
    * Fixed exact trace cutting for PSD segments (see #2040).
+   * Timestamp representations internally and in npz I/O were changed to use
+     integer nanosecond POSIX timestamps to avoid any potential floating point
+     inaccuracies and since this is also what UTCDateTime is based on nowadays
+     (see #2045)
  - obspy.signal.cross_correlation:
    * Add new `correlate_template()` function with 'full' normalization option,
      required for correlations in template-matching
@@ -14,6 +18,7 @@ master:
    * 'domain' parameter in correlate function is deprecated in favour of new
      'method' parameter to be consistent with recent SciPy versions
      (see #2042).
+
 
 1.1.x:
  - General:
--- a/obspy/signal/spectral_estimation.py
+++ b/obspy/signal/spectral_estimation.py
@@ -326,7 +326,7 @@ class PPSD(object):
             raise ValueError(msg)
 
         # save version numbers
-        self.ppsd_version = 1
+        self.ppsd_version = 2
         self.obspy_version = __version__
         self.matplotlib_version = ".".join(map(str, MATPLOTLIB_VERSION))
         self.numpy_version = np.__version__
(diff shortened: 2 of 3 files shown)


megies:
Also see #2034 for examples of unexpected behavior associated with floating point issues.

megies:
Good idea..

megies:
@krischer, I've added raising an exception in this case and will also do a PR for maintenance branch that at least shows a warning, so that the 1.1.x line is also covered with this..

obspy/signal/spectral_estimation.py
```diff
@@ -1231,6 +1233,16 @@ def load_npz(filename, metadata=None):
...
         # the information regarding stats is set from the npz
         ppsd = PPSD(Stats(), metadata=metadata)
+        # add some future-proofing and raise an exception if older obspy
+        # versions should read a more recent ppsd npz file, since this is very
+        # like problematic
+        if data['ppsd_version'].item() > ppsd.ppsd_version:
+            msg = ("Trying to read a PPSD npz with 'ppsd_version={}'. This "
+                   "file was written on a more recent ObsPy version and can "
+                   "not safely be read with this ObsPy version (current "
+                   "'ppsd_version' is {}).").format(
```
krischer: Maybe add "Please update your ObsPy installation."?

CHANGELOG.txt
```diff
@@ -5,6 +5,10 @@ master:
  - obspy.io.shapefile:
    * Add possibility to add custom database columns when writing catalog
      objects to shapefile (see #2012)
+ - obspy.signal:
+   * PPSD: for safety reasons, raise an ObsPyException if trying to read a PPSD
+     npz file that was written with a newer ObsPy version than is currently
```
krischer: That's not really correct. Best rephrase to "written with a newer internal file format" or something like it. I guess the format does not change that often.

megies: Yeah, you're right, that's misleading..

megies:
Made those wording changes and force-pushed.

megies:
Rebased and removed the exception for newer npz, was delegated to #2051. Hope I didn't screw up the merge conflict resolution, always a bit tricky when indent levels change (due to the context manager used now)

hash before rebasing: 7d684ba0127146c2a1ae94cb3dbdb09ce93d3c45

krischer:
Some tests fail now.

Jollyfant:
> Some tests fail now.

One of these is the test included in the PPSD fix we merged some days ago.

megies:
> One of these is the test included in the PPSD fix we merged some days ago.

Ah OK.. I rebased but didn't check that new test.. and somehow must've messed up the conflict resolution during rebasing, that npz version check somehow got lost.. -_-

Jollyfant:
It could be because of the change in `._times_processed` to nanoseconds representation.

megies:
OK, looks like this one's ready now. I'm a bit anxious about the internal time representation change.. but I think it's better in the long term and hopefully I thought about all issues that might arise when reading/adding older npz files..

krischer:
I fixed a conflict in the changelog and rebased. Can be merged from my point of view if CI is green.

megies:
Alright, let's do it.. 🙈

Issue #2036, opened 2018-01-08
Exception occurring suring reading a Reftek130 file


tafro:
I got following exception while trying to read the Reftek130 raw file (
[112828970_0036EE80.zip](https://github.com/obspy/obspy/files/1611645/112828970_0036EE80.zip)
)

```
from obspy import read
read('112828970_0036EE80',dtype='float32')`

```

```
/opt/obspy/src/obspy/obspy/io/reftek/core.py:157: UserWarning: Detected permuted packet sequence, sorting.
  warnings.warn(msg)
/opt/obspy/src/obspy/obspy/io/reftek/core.py:322: UserWarning: No channel code specified in the data file and no component codes specified. Using stream label and number of channel in file as channel codes.
  warnings.warn(msg)
Traceback (most recent call last):
  File "/opt/obspy/src/obspy/obspy/io/reftek/core.py", line 333, in to_stream
    t_last + (npts_last - 1) * delta)
AssertionError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./ref2segy.py", line 130, in <module>
    stadd=read(file,dtype='float32')
  File "<decorator-gen-31>", line 2, in read
  File "/opt/obspy/src/obspy/obspy/core/util/decorator.py", line 299, in _map_example_filename
    return func(*args, **kwargs)
  File "/opt/obspy/src/obspy/obspy/core/stream.py", line 231, in read
    st.extend(_read(file, format, headonly, **kwargs).traces)
  File "<decorator-gen-32>", line 2, in _read
  File "/opt/obspy/src/obspy/obspy/core/util/decorator.py", line 209, in uncompress_file
    result = func(filename, *args, **kwargs)
  File "/opt/obspy/src/obspy/obspy/core/stream.py", line 273, in _read
    headonly=headonly, **kwargs)
  File "/opt/obspy/src/obspy/obspy/core/util/base.py", line 410, in _read_from_plugin
    list_obj = read_format(filename, **kwargs)
  File "/opt/obspy/src/obspy/obspy/io/reftek/core.py", line 102, in _read_reftek130
    verbose=verbose)
  File "/opt/obspy/src/obspy/obspy/io/reftek/core.py", line 342, in to_stream
    raise Reftek130Exception(msg)
obspy.io.reftek.core.Reftek130Exception: Reftek file has a trace with an inconsistent endtime or number of samples. Please open an issue on GitHub and provide your file fortesting.


```

megies:
[...]

PR #2046, opened 2018-01-12
UTCDateTime: add public property to access nanosecond integer POSIX timestamp

megies:
Since this is now the internal representation of UTCDateTime, I think it makes sense for user code to use that to serialize UTCDateTime objects, and we should therefore offer a public property to access this. Also implementing `__int__` for this would be handy, but I fear that might be way too confusing alongside the existing `__float__` which returns the timestamp in seconds..

e.g. #2045, #2036

### PR Checklist
- [ ] All tests still pass.
- [x] maybe should add `doc` to new property

### DIFF
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1123,6 +1123,20 @@ class UTCDateTimeTestCase(unittest.TestCase):
         self.assertFalse(a == e)
         self.assertFalse(e == a)
 
+    def test_ns_public_attribute(self):
+        """
+        Basic test for public ns interface to UTCDateTime
+        """
+        t = UTCDateTime('2018-01-17T12:34:56.789012Z')
+        # test getter
+        self.assertEqual(t.ns, 1516192496789012000)
+        # test setter
+        x = 1516426162899012123
+        t.ns = x
+        self.assertEqual(t.ns, x)
+        self.assertEqual(t.day, 20)
+        self.assertEqual(t.microsecond, 899012)
+
 
 def suite():
     return unittest.makeSuite(UTCDateTimeTestCase, 'test')
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -404,6 +404,7 @@ class UTCDateTime(object):
         self.__ns = value
 
     _ns = property(_get_ns, _set_ns)
+    ns = property(_get_ns, _set_ns)
 
     def _from_datetime(self, dt):
         """
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -23,10 +23,6 @@ import numpy as np
 TIMESTAMP0 = datetime.datetime(1970, 1, 1, 0, 0)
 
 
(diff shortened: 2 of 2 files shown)


d-chambers:
While you are working with the nanoseconds, might I also suggest rounding the ns to the specified precision upon init'ing the UTCDateTime object? If the precision is specified, anything below that is just noise that may cause surprising behavior. It also would mean everything required to fully represent the data contained in the object will be given by the string representation. This would help alleviate some of the surprising behaviors in described in #2034.

You should be able to change these lines:
https://github.com/obspy/obspy/blob/7585cc1819343fc41509ba1a396019c6b9f4eaf0/obspy/core/utcdatetime.py#L385-L388 

to this:
```python
 def _set_ns(self, value): 
     if not isinstance(value, int): 
         raise TypeError('nanoseconds must be set as int/long type') 
self.__ns = round(value, self.precision - 9) 
```
Also, why not replace all instances of `_ns` with `ns` ?  It would be a bit redundant to have a public and a private method that do exactly the same thing.

barsch:
```round``` will return a ```float`` under Python 2.7:
```
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)] on win32
>>> aa=34234234234324
>>> round(aa, -3)
34234234234000.0
```
I agree with changing ```_ns``` to ```ns```.

megies:
> While you are working with the nanoseconds, might I also suggest rounding the ns to the specified precision upon init'ing the UTCDateTime object?

That would drastically change the current logic, which is to always retain full info on the timestamp and only respect precision during comparison operations. This is definitely something that would have to be discussed in detail in a dedicated issue (and I expect mixed opinions).

> Also, why not replace all instances of _ns with ns ? It would be a bit redundant to have a public and a private method that do exactly the same thing.

Yeah, but I know that code is already around that uses `._ns` (I'm using it so far, for lack of the public method), so this would need a deprecation cycle, and to be honest, I'm not sure it's worth it to remove it.

d-chambers:
> That would drastically change the current logic, which is to always retain full info on the timestamp and only respect precision during comparison operations.

Point taken, I will open a dedicated issue when I get a few minutes so discussion can take place.

d-chambers:
Hey @barsch, using the round function with a negative number works in python 2.7 (just returns a float as you observed), but doesn't work with obspy because the future.builtins wild card import replaces the built-in round with new_round, which has the following few lines:
https://github.com/PythonCharmers/python-future/blob/39a066ed8c29a0b3a32adac2ffd407119fe9ea6d/src/future/builtins/newround.py#L32-L33.

megies:
@d-chambers if you want to put work into this (`future` related stuff), it's probably better spent in #1613.. (hoping that it's not too ridiculously outdated already.. :-/)

megies:
Rebased on current master

megies:
Added test + rebased. Good to merge on nod from CI.

megies:
Curious if properties get documented by our Sphinx.. so +DOCS

megies:
CI looks good, merging

PR #2049, opened 2018-01-15
UTCDateTime rich comparison operators and documentation update

d-chambers:
### What does this PR do?

This PR does the following:

1. Eliminates the surprising `UTCDateTime` behavior observed in #2034 
2. Fixes some of the issues with sorting, min, max, etc. outlined by #1765 
3. Updates some of the docstrings to indicate the internal representation of time is an integer (no longer a float). 
4. Reformatted some examples in the doc strings that change the precision to favor instantiating a new UTCDateTime object with the new precision rather than modifying one in place. This will be useful if UTCDateTime ever becomes immutable.  
5. Created py3_round function in `obspy.core.util.misc` to approximate the behavior of python 3's built-in round function on python 2, particularly allowing ndigits to be negative for rounding ints. 
6. Added a warning when a precision above 9 is selected (which is no longer possible/meaningful).

This PR supersedes #1766, which can be closed/discarded when this merges.    

Note: I decided to base from master rather than a maintenance branch as this PR implements a new function (py3_round) and slightly different behavior of the comparison operators and microsecond attribute of `UTCDateTime`. 

issues: #1765, #2034

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -3,7 +3,7 @@ master:
    * Casting FDSN identifiers to strings upon setting in the stats dictionary
      (see #1997).
    * UTCDateTime objects will now always evaluate equal if their string
-     representations are equal.
+     representations are equal (see #2049).
  - obspy.io.shapefile:
    * Add possibility to add custom database columns when writing catalog
      objects to shapefile (see #2012)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,8 @@ master:
  - obspy.core:
    * Casting FDSN identifiers to strings upon setting in the stats dictionary
      (see #1997).
+   * UTCDateTime objects will now always evaluate equal if their string
+     representations are equal.
  - obspy.io.shapefile:
    * Add possibility to add custom database columns when writing catalog
      objects to shapefile (see #2012)
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1188,8 +1188,6 @@ class UTCDateTimeTestCase(unittest.TestCase):
         """
         see #1008
         """
-        utc = UTCDateTime("9999-12-31T23:59:59.9999")
-        str(utc)
         self.assertEqual(str(UTCDateTime("9999-12-31T23:59:59.9999")),
                          "9999-12-31T23:59:59.999900Z")
         self.assertEqual(str(UTCDateTime("9999-12-31T23:59:59.999999")),
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -2,14 +2,13 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-import warnings
-
 from future.builtins import *  # NOQA @UnusedWildImport
 
 import copy
 import datetime
 import itertools
 import unittest
+import warnings
 
 import numpy as np
 
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1271,17 +1271,20 @@ class UTCDateTimeTestCase(unittest.TestCase):
                 yield start
 
         # convert to UTCDateTime objects
-        close_timestamps = list(yield_close_floats(1515174511.1984458, 10))
-        close_utc = [UTCDateTime(x) for x in close_timestamps]
+        float0 = 1515174511.1984458
+        for precision in range(1, 10):
+            close_timestamps = list(yield_close_floats(float0, 10))
+            close_utc = [UTCDateTime(x, precision=precision)
+                         for x in close_timestamps]
 
-        # if str are equal then objects should be equal and visa versa
-        for num in range(len(close_utc) - 1):
-            utc1 = close_utc[num]
-            utc2 = close_utc[num + 1]
-            if utc1 == utc2:
-                self.assertEqual(str(utc1), str(utc2))
-            if str(utc1) == str(utc2):
-                self.assertEqual(utc1, utc2)
+            # if str are equal then objects should be equal and visa versa
+            for num in range(len(close_utc) - 1):
+                utc1 = close_utc[num]
+                utc2 = close_utc[num + 1]
+                if utc1 == utc2:
+                    self.assertEqual(str(utc1), str(utc2))
+                if str(utc1) == str(utc2):
+                    self.assertEqual(utc1, utc2)
 
 
 def suite():
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1227,7 +1227,7 @@ class UTCDateTimeTestCase(unittest.TestCase):
     def test_timestamp_can_serialize_with_time_attrs(self):
         """
         Test that the datetime attrs can be used to serialize UTCDateTime
-        objects inited from floats (given default precision of 6) - see 2034
+        objects inited from floats (given default precision of 6) - see #2034
         """
         time_attrs = ('year', 'month', 'day', 'hour', 'minute', 'second',
                       'microsecond')
@@ -1242,7 +1242,7 @@ class UTCDateTimeTestCase(unittest.TestCase):
     def test_str_ms_equal_ms(self):
         """
         Test that the microseconds in the str representation are equal to
-        the microseconds attr - see 2034
+        the microseconds attr - see #2034
         """
         close_timestamps = [1515174511.1984465, 1515174511.1984463,
                             1515174511.1984460, 1515174511.1984458]
@@ -1262,7 +1262,7 @@ class UTCDateTimeTestCase(unittest.TestCase):
         we can no longer guarantee equality based on the difference in
         nanoseconds. This trade-off was made to ensure UTCDateTime objects
         are always equal to their string representation when precision <= 6.
-        See issue #2034 and pr # 1766.
+        See issue #2034.
         """
         # get an array of floats as close together as possible
         def yield_close_floats(start, length):
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1301,6 +1301,25 @@ class UTCDateTimeTestCase(unittest.TestCase):
             self.assertEqual(len(w), 1)
             self.assertIn('different precision', str(w[-1].message))
 
+    def test_string_representation_various_precisions(self):
+        """
+        Ensure string representation works for many different precisions
+        """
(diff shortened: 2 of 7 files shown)


obspy/core/util/misc.py
```diff
@@ -639,6 +639,34 @@ def buffered_load_entry_point(dist, group, name):
     return _ENTRY_POINT_CACHE[hash_str]
 
 
+if sys.version_info[0] < 3:
```
krischer: Please move it to this file as we already have a version of this function for the other way around: https://github.com/obspy/obspy/blob/master/obspy/core/compatibility.py#L99

CHANGELOG.txt
```diff
@@ -2,6 +2,8 @@ master:
  - obspy.core:
    * Casting FDSN identifiers to strings upon setting in the stats dictionary
      (see #1997).
+   * UTCDateTime objects will now always evaluate equal if their string
+     representations are equal.
```
krischer: Please add the issue number (but I guess you'll do that anyway in a subsequent commit).

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1123,6 +1210,68 @@ def test_issue_1652(self):
         self.assertFalse(a == e)
         self.assertFalse(e == a)
 
+    def test_timestamp_can_serialize_with_time_attrs(self):
+        """
+        Test that the datetime attrs can be used to serialize UTCDateTime
+        objects inited from floats (given default precision of 6) - see 2034
```
krischer: `#` missing in front of the issue number.

krischer: This also happens in the next two test cases.

obspy/core/utcdatetime.py
```diff
@@ -943,7 +948,8 @@ def __sub__(self, value):
             # see datetime.timedelta.total_seconds
             value = (value.microseconds + (value.seconds + value.days *
                      86400) * 10**6) / 1e6
-        return UTCDateTime(ns=self._ns - int(round(value * 1e9)))
+        # return UTCDateTime(ns=self._ns - int(round((value * 1e9))))
+        return UTCDateTime(ns=self._ns - int(round((value * 1e9))))
```
krischer: This is not any different than before, right?

obspy/core/utcdatetime.py
```diff
@@ -981,63 +987,63 @@ def __unicode__(self):
         """
         return str(self.__str__())
 
+    def _operate(self, other, op_func):
```
krischer: I'm really not that deep into the fuzzy comparison problem but would it make sense to first convert anything to a `UTCDateTime` object and then just compare those? Might be easier to keep the logic straight in that fashion.

obspy/core/utcdatetime.py
```diff
@@ -981,63 +987,63 @@ def __unicode__(self):
...
+        if isinstance(other, UTCDateTime):
+            a = py3_round(self._ns, self.precision - 9)
+            b = py3_round(other._ns, self.precision - 9)
```
krischer: Should this both `self.precision` arguments be replaced with `min(self.precision, other.precision)`? Or maybe raise a warning if two UTCDateTime objects of differing precision are used as that operation is not really defined?

obspy/core/utcdatetime.py
```diff
@@ -981,63 +987,63 @@ def __unicode__(self):
...
+
     def __eq__(self, other):
         """
         Rich comparison operator '=='.
 
         .. rubric: Example
 
-        Comparing two UTCDateTime object will always compare timestamps rounded
+        Comparing two UTCDateTime object will compare timestamps rounded
         to a precision of 6 digits by default.
```
krischer: That phrasing is not really accurate as the default precision just happens to be 6. Might be a good opportunity to change it.

obspy/core/utcdatetime.py
d-chambers: No I changed this then ended up changing it back to better state intent. I will clean this up.

obspy/core/utcdatetime.py
d-chambers: Good thought, I like the idea of raising a warning, maybe an error eventually. I will use the min of the precisions and raise a warning if they are different.

obspy/core/util/misc.py
d-chambers: I dont see a test file for the compatibility module, and I cant find any direct tests for any of those functions. Should I create a new test module or leave the tests for py3_round in test_util_misc.py ?

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1123,6 +1210,68 @@ def test_issue_1652(self):
...
+    def test_close_utc_are_equal(self):
+        """
+        Ensure UTCs init'ed with floats that are very close together are
+        equal - see 2034
+
+        Note: Due to the rounding nanosecond attribute before comparision
+        we can no longer guarantee equality based on the difference in
+        nanoseconds. This trade-off was made to ensure UTCDateTime objects
+        are always equal to their string representation when precision <= 6.
+        See issue #2034 and pr # 1766.
```
barsch: # 1766 -> #1766

obspy/core/utcdatetime.py
barsch: or to have it maybe more lightweight using nanoseconds (```long``` datatype)?

obspy/core/util/misc.py
barsch: create a new test module + move the test - tests should be in the same module as the tested function

obspy/core/utcdatetime.py
d-chambers: I modified the function to always use the nanoseconds comparison as suggested. Look ok to you guys?

d-chambers:
So it seems most (hopefully all) the failures are related to the Numpy printing issue as I have all tests passing on 2.7 and 3.6 locally. I suppose I should just rebase with master to keep this moving forward?

krischer:
> So it seems most (hopefully all) the failures are related to the Numpy printing issue as I have all tests passing on 2.7 and 3.6 locally. I suppose I should just rebase with master to keep this moving forward?

Yes - the numpy errors should go away if you rebase on master.

d-chambers:
Any insights on the Travis or docs failures? The Travis report just says stalled build on 3.4, It looks like it died trying to setup an Anaconda environment.

I am not sure what is wrong with the docs, stderr just shows a lot of warnings about documents not being included in toctree, although all this PR did was update some docstrings so I doubt if the changes in this PR are to blame.

krischer:
I just restarted the failing travis build - this oftentimes fixes these spurious problems. No clue what is wrong with the docs-buildbot. @megies - any ideas?

megies:
> I am not sure what is wrong with the docs, stderr just shows a lot of warnings about documents not being included in toctree, although all this PR did was update some docstrings so I doubt if the changes in this PR are to blame.



> No clue what is wrong with the docs-buildbot. @megies - any ideas?

Docs buildbot will always show a red X since we always have lots of warnings/errors that nobody ever got round to fix. But you can still have a look and make sure the affected pages render as expected, that's the main purpose anyway.

In a perfect world, if we ever manage to get one build with Sphinx errors and warnings, we could keep that clean state by really relying on the docs-buildbot commit status.. well.. daydreaming, I guess.

d-chambers:
> you can still have a look and make sure the affected pages render as expected, that's the main purpose anyway.

I skimmed through the affected pages, everything looks correct.

krischer:
This is ready to go from my point of view. @barsch Can you give this a final approval?

megies:
I'm a bit torn about the change of string representation of `UTCDateTime`, which apparently now depends on currently set `.precision`? It could hurt people that rely on how `UTCDateTime.__str__` looks (e.g. when outputting to custom files etc.).

I know, it's rather unlikely that this affect many people.. just mentioning.

Also, when setting precision to zero, the string rep has a funny trailing dot, and when setting precision to something negative, str rep is erroring out..

```
In [16]: t.precision = 0

In [17]: t
Out[17]: 2018-01-24T10:17:08.Z

In [18]: t.precision = -1

In [19]: t
Out[19]: ---------------------------------------------------------------------------
/home/megies/anaconda/envs/default/lib/python2.7/site-packages/IPython/core/formatters.pyc in __call__(self, obj)
    697                 type_pprinters=self.type_printers,
    698                 deferred_pprinters=self.deferred_printers)
--> 699             printer.pretty(obj)
    700             printer.flush()
    701             return stream.getvalue()

[...]

/home/megies/anaconda/envs/default/lib/python2.7/site-packages/obspy/core/utcdatetime.pyc in __str__(self)
    960         dt = self.datetime
    961         pattern = "%%.%dlf" % (self.precision)
--> 962         ns = pattern % ((self._ns % 10**9) / 1e9)
    963         return "%04d-%02d-%02dT%02d:%02d:%02d.%sZ" % (
    964             dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,

ValueError: unsupported format character '-' (0x2d) at index 2
```

d-chambers:
That does look like a problem, I will dig into it. The modification of the string behavior was something I copied over from the work already done on  #1766, so we will need to ask @barsch about his intent here. I don't have strong feelings about it either way but I will see if I cant fix the issues you highlighted.

It would be nice, in theory, for the string representation to give an indication of the precision, but I am not sure how it should handle negative precisions; we usually think of time in months/years rather than mega/giga seconds. 

It could also break some existing code, but I agree this probably would be rare. As long as this stays at the 1.2.0 milestone I don't think we need to worry about it because major version ticks allow for some (small) breaking changes no?

krischer:
> It would be nice, in theory, for the string representation to give an indication of the precision, but I am not sure how it should handle negative precisions; we usually think of time in months/years rather than mega/giga seconds.

Negative precisions did not really work before this PR here so there would be a possibility to maybe add some `[+-10e5 seconds]` or something to the string representation.


> It could also break some existing code, but I agree this probably would be rare. As long as this stays at the 1.2.0 milestone I don't think we need to worry about it because major version ticks allow for some (small) breaking changes no?

Only if its really justified and there is some tangible benefit- otherwise the costs of breakage are too high. On the other hand I think that almost everything just uses the default precision in which case nothing would change as far as I follow the discussion.

barsch:
uhm maybe I'm blind but I dont see me changing the __str__ representation in #1766 - so if backwards compatibility for the < 0.1% users using precision != 6 is needed than I guess this needs to be reverted ...

otherwise looking good for me

d-chambers:
Hey @barsch you are right, it looks like the change in the `__str__` method occurred in #1325, which was merged into master a while ago. I should have checked the git blame before assuming.

krischer:
Looks good to me! I'll merge as soon as CI turns green if there are no objections. Thanks a bunch for this!

PR #2051, opened 2018-01-16
PPSD: raise exception when trying to load/add a npz that was written with a higher 'ppsd_version'

megies:
.. than is currently used (to be more precise, raise
when reading a npz file that has a higher 'ppsd_version' which is used
for tracking changes in PPSD internal structure and serialization)

also has a test

with obspy >=1.2.0 this will be upgraded to an exception

### PR Checklist
- [x] ~~If the PR is making changes to documentation, docs pages can be built automatically.
- [x] All tests still pass.
- [x] ~~Any new or changed features have are fully documented.~~
- [x] ~~First time contributors have added your name to `CONTRIBUTORS.txt`~~

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -18,6 +18,10 @@
    * Write the UTC time zone specifier to all times (see #2015).
  - obspy.signal:
    * Allow singular COUNT units in evalresp (see #2003, #2011).
+   * PPSD: for safety reasons, raise an ObsPyException if trying to read a PPSD
+     npz file that was written with a newer version of the npz representation
+     than is used by current ObsPy version (see #2051)
+
 
 1.1.0: (doi: 10.5281/zenodo.165135)
  - General:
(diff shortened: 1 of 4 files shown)


krischer:
Alright  - can we merge this first, then merge maintenance in master, and then merge #2045?

krischer:
and of course rebase #2045 first.

megies:
Yeah, we'll need to manually resolve merge conflicts, but not a big deal and having that warning in 1.1.x is worth it.

megies:
Final wording changes and rebased

krischer:
Any reason not to already raise an exception here instead of `1.2.0`? The nanosecond change is the first change of the internal format, right?

megies:
> Any reason not to already raise an exception here instead of 1.2.0? The nanosecond change is the first change of the internal format, right?

Well yeah.. I guess it should be OK to add it as an exception here already.. will do

megies:
OK, final changes.. now also checks `ppsd_version` when using `PPSD.add_npz()`. And upgraded from warning to raising an exception.

megies:
OK, CI passing (Travis OSX taking forever).. merging

Issue #1981, opened 2017-11-01
 problem reading MiniSEED file with 512 blank bytes in it


gposada:
Hello obspy developers,

When i execute the command obspy.read() for read a seed file, obspy show an error:

**_InternalMSEEDReadingWarning: readMSEEDBuffer(): Record starting at offset 2813440 is not valid SEED. The rest of the file will not be read.
  warnings.warn(*i)_**

This error is due to what? Is an error of SEED file or of package?

thank you

krischer:
I'm fairly certain that this is a wrong SEED file. Just open the file in a hex-editor and have a look at the specified byte position?

Or upload the file somewhere and I can have a look.

gposada:
okay, i too think is an error in SEED file. 

You can download the file in [http://siata.gov.co/sismografos/temporal/](url)

Thanks!

krischer:
So this file has one 512 byte chunk in it which is completely empty:

![screen shot 2017-11-01 at 15 22 36](https://user-images.githubusercontent.com/800487/32279268-8a0da822-bf18-11e7-99e1-af81c2e655e1.png)

If you remove this chunk the file reads fine. Note that this is different from so called shadow or noise records which we deal with in #1495. These are also empty but they still have at least the sequence number.

It would be fairly straightforward to teach ObsPy how to read this file but I don't really want to do it if its not valid in any case.

@chad-iris Do you have any additional comments/experiences with this?

krischer:
I'm reopening this as I'm not sure if this happens more often and we should thus deal with it. Pinging @chad-iris again for a comment ;)

megies:
I agree, first thing to do is to find out why and where this happened and how often this might be encountered.

@gposada can you trace the data back to it's origin, i.e. who recorded and with what equipment?

gposada:
Hello, the equipment is a RaspberryShake, and the error only happened two times in three months.

I spoke with people from OSOP, they say the error it's due to the Raspberry turns itself off.

megies:
OK, given this piece of information, I would argue that this is a bug in the RaspberryShake implementation. If this is an "unclean shutdown" problem, this should be fixed upstream with RaspberryShake. It should not write an all-blank record.

I'll close this as an upstream issue, feel free to reopen, everybody.

chad-iris:
[...]

PR #2057, opened 2018-01-22
Read blank miniSEED records

krischer:
As discussed in #1981 this PR enables ObsPy to read MiniSEED files that have chunks (at some multiple of the minimal record size) filled with zero bytes. This can go in maintenance as it does not really change behavior (it just reads some files it previously did not).

Also contains an unrelated fix to make the test cases more independent.

### DIFF
--- a/obspy/core/util/testing.py
+++ b/obspy/core/util/testing.py
@@ -706,5 +706,51 @@ def get_all_py_files():
     return sorted(py_files)
 
 
+class WarningsCapture(object):
+    """
+    Try hard to capture all warnings.
+
+    Aims to be a reliable drop-in replacement for built-in
+    warnings.catch_warnings() context manager.
+
+    Based on pytest's _DeprecatedCallContext context manager.
+    """
+    def __enter__(self):
+        self.captured_warnings = []
+        self._old_warn = warnings.warn
+        self._old_warn_explicit = warnings.warn_explicit
+        warnings.warn_explicit = self._warn_explicit
+        warnings.warn = self._warn
+        return self
+
+    def _warn_explicit(self, message, category, *args, **kwargs):
+        self.captured_warnings.append(
+            warnings.WarningMessage(message=category(message),
+                                    category=category,
+                                    filename="", lineno=0))
+
+    def _warn(self, message, category=None, *args, **kwargs):
+        if isinstance(message, Warning):
+            self.captured_warnings.append(
+                warnings.WarningMessage(
+                    message=category(message), category=category or Warning,
+                    filename="", lineno=0))
+        else:
+            self.captured_warnings.append(
+                warnings.WarningMessage(
+                    message=category(message), category=category,
+                    filename="", lineno=0))
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        warnings.warn_explicit = self._old_warn_explicit
+        warnings.warn = self._old_warn
+
+    def __len__(self):
+        return len(self.captured_warnings)
+
+    def __getitem__(self, key):
+        return self.captured_warnings[key]
+
+
 if __name__ == '__main__':
     doctest.testmod(exclude_empty=True)
--- a/obspy/io/mseed/tests/test_mseed_special_issues.py
+++ b/obspy/io/mseed/tests/test_mseed_special_issues.py
@@ -1163,6 +1163,31 @@ class MSEEDSpecialIssueTestCase(unittest.TestCase):
             "read the file in chunks as documented here: "
             "https://github.com/obspy/obspy/pull/1419#issuecomment-221582369")
 
+    def test_read_file_with_empty_blocks_in_between(self):
+        """
+        Test reading MiniSEED files that have some empty blocks.
+        """
+        # This file has two 4096 bytes records.
+        filename = os.path.join(self.path, 'data', 'test.mseed')
+        with io.open(filename, "rb") as fh:
+            rec1 = fh.read(4096)
+            rec2 = fh.read(4096)
+
+        reference = _read_mseed(filename)
+        del reference[0].stats.mseed
+
+        for length in (128, 256, 512, 1024, 2048, 4096, 8192):
+            with io.BytesIO() as buf:
+                buf.write(rec1)
+                buf.write(b'\x00' * length)
+                buf.write(rec2)
+                buf.seek(0, 0)
+                st = _read_mseed(buf)
+
+            # Remove things like file-size and what not.
+            del st[0].stats.mseed
+            self.assertEqual(reference, st)
+
 
 def suite():
     return unittest.makeSuite(MSEEDSpecialIssueTestCase, 'test')
--- a/obspy/io/mseed/tests/test_mseed_special_issues.py
+++ b/obspy/io/mseed/tests/test_mseed_special_issues.py
@@ -143,7 +143,10 @@ class MSEEDSpecialIssueTestCase(unittest.TestCase):
             warnings.simplefilter("always")
             data_record = _read_mseed(file)[0].data
 
-        self.assertEqual(len(w), 1)
+        # This will raise 18 (!) warnings. It will skip 17 * 128 bytes due
+        # to it not being a SEED records and then complain that the remaining
+        # 30 bytes are not enough to constitute a full SEED record.
+        self.assertEqual(len(w), 18)
         self.assertEqual(w[0].category, InternalMSEEDWarning)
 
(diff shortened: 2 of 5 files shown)


krischer:
Ready to review if CI passes.

dsentinel:
Apologies for being late to the discussion, after @krischer developed this solution. This just caught my eye.

I have found the best approach for data readers such as this, is to skip unreadable / unparsable / non-valid blocks, as Chad does with libmseed, and continue reading for valid blocks. Regardless of what is in the block (zeros , random bits, low-level I/O Error). Issuing a warning on skip.

Although rare, It happens that digitizers make a garbage block (especially in the event of power loss during a write), or a block, or several, corrupt on media. But whatever the cause, and regardless of the contents of the bad blocks, the user always wants all the good data that can be retrieved.

Fortunately dropping a sequence of mseed blocks is robust and only manifests as a gap.

Something along the lines of changing the `Except: ... break` to `Except: issue_waring, continue` in the `while` of  io/mseed/core._read_mseed here:
[https://github.com/obspy/obspy/blob/32feb03f781d78c837cd1211d2c811a9eaebdf9d/obspy/io/mseed/core.py#L487](url)

Althought it may take a bit more work and some restructuring than this edit
(unexpected EOF will need a `break`)
..I don't think it'd be much, and it will be much more robust in reading mseed with problems, and the code won't have to be adjusted for the contents of of the bad blocks that may crop up in the future.

Also fortunate, most of these corruption issues leave the remainder of the file block aligned, so this type of solution will hold. I often deal with misalignments after bad blocks in another format, but I can't recall ever with mseed and digitizers that write directly to mseed. This is more complicated, but still solvable. 

I'll try to find some time to implement the above and see how the tests fair, or what kind of rabbit hole it creates.

krischer:
@dsentinel Thanks for your suggestion - I implemented it and ObsPy now keeps skipping 128 bytes ahead if it does not detect a miniSEED record. This still assumes somewhat aligned records but checking every single byte would be a bit too expensive.

@barsch @ThomasLecocq Any chance you can have a look at the windows failure? If I read the error correctly it does not jump in this conditional for faulty MiniSEED blocks.

https://github.com/krischer/obspy/blob/91f59c9daad1ee1d6f88034d1aac1706ded9c1e3/obspy/io/mseed/src/obspy-readbuffer.c#L380-L387

Executing `test_read_file_with_non_valid_blocks_in_between()` in `test_mseed_special_issues.py` should trigger it. Can you let me know what it returns otherwise? It also happens on the linux travis bots but I cannot for the life of myself reproduce it on several linux machines nor with docker.

ThomasLecocq:
@krischer 
am not good at C, what is the meaning of 
```c++
if (retcode == MS_NOTSEED) {
```

ThomasLecocq:
catching verbosity ... @barsch 's passion...

krischer:
Exactly. Assuming the above mentioned test fails for you - can you just add

```C
printf("%i\n", retcode);
```

after the call to `msr_parse()` and give me the full output while running the one test?

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -1176,13 +1179,45 @@ def test_read_file_with_empty_blocks_in_between(self):
...
+                "readMSEEDBuffer(): Not a SEED record. Will skip bytes "
+                "4096 to 4223.")
+
+            # Remove things like file-size and what not.
+            del st[0].stats.mseed
+            self.assertEqual(reference, st)
+
+        # Try the same thing but fill with random bytes.
+        # The seed is not really needed but hopefully guards against the
+        # very very rare case of random bytes making up a valid SEED record.
```
Jollyfant: Haha 😆 reminds me of this: https://en.wikipedia.org/wiki/Infinite_monkey_theorem

krischer:
Can anyone locally reproduce this error? Otherwise I'll just move to remote travis debugging tomorrow...

megies:
@krischer  I can reproduce: http://tests.obspy.org/91957/
.. and strangely this seems to depend on test execution order. I get fails on ’obspy-runtests’ and ’pytest io/mseed’ but no fail on e.g. ’pytest io/mseed/tests/test_mseed_special_issues.py’.

krischer:
ah okay - so I guess its just once again related to the problematic catching of warnings in Python 2.7.

megies:
Dunno, there might be an issue.. this is comparing the warnings in both runs (one failing, one passing):

---
```
(Pdb++) for w_ in w: print str(w_.message)
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4096 to 4223.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4224 to 4351.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4352 to 4479.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4480 to 4607.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4608 to 4735.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4736 to 4863.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4864 to 4991.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 4992 to 5119.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5120 to 5247.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5248 to 5375.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5376 to 5503.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5504 to 5631.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5632 to 5759.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5760 to 5887.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 5888 to 6015.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 6016 to 6143.
readMSEEDBuffer(): Not a SEED record. Will skip bytes 6144 to 6271.
readMSEEDBuffer(): Last record only has 30 byte(s) which is not enough to constitute a full SEED record. Corrupt data? Record will be skipped.
```
---
```
(Pdb++) for w_ in w: print str(w_.message)
readMSEEDBuffer(): Last record only has 30 byte(s) which is not enough to constitute a full SEED record. Corrupt data? Record will be skipped.
```

megies:
dunno.. i know issues like these can come from uncaught warnings in doctests.. but doesnt look like thats the case here

krischer:
I wrote a custom warnings capture context manager which should be finally able to reliably capture warnings in tests - I think this should come in handy in the future as we've already had a bunch of cases where we could not reliably capture them across Python/OS versions.

Let's hope this works and also fixes the test failures for this PR.

krischer:
CI looks good now. Someone please review and merge!

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -1163,6 +1166,61 @@ def test_reading_file_larger_than_2048_mib(self, getsize_mock):
...
+            with io.BytesIO() as buf:
+                buf.write(rec1)
+                buf.write(b'\x00' * length)
+                buf.write(rec2)
+                buf.seek(0, 0)
+                # This will raise 1 warning per 128 bytes.
+                with WarningsCapture() as w:
+                    st = _read_mseed(buf)
+                self.assertEqual(len(w), length // 128)
+
```
Jollyfant: Maybe confirm the expected number of records in the stream too in this test.

obspy/core/util/testing.py
```diff
@@ -706,5 +706,51 @@ def get_all_py_files():
     return sorted(py_files)
 
 
+class WarningsCapture(object):
```
Jollyfant: I'm a little bit confused why this monkey patching of `warnings` is even required.. but the code itself looks fine.

megies: Indeed, has this to do with C??

krischer: No - its just how the warnings filtering in Python works. It is really hard to get it to behave identical across all Python versions. There are have been a number of occasions in which we could not test warnings because they did not work on all platforms.

This ctx manager just monkey patches the warnings module to get it to behave consistently. Seems to work pretty nice.

obspy/io/mseed/tests/test_mseed_special_issues.py
krischer: Done

PR #2066, opened 2018-02-02
fix axes kw in plot_response

trichter:
Fixes the following problem:

```
<ipython-input-2-bb1fc9308dda> in <module>()
      1 fig, axes = plt.subplots(2, 1, figsize=(12, 12))
----> 2 inv.plot_response(0.01, axes=axes)
      3 plt.show()

~/anaconda/envs/notebooks/lib/python3.6/site-packages/obspy/core/inventory/inventory.py in plot_response(self, min_freq, output, network, station, location, channel, time, starttime, endtime, axes, unwrap_phase, plot_degrees, show, outfile)
    851         import matplotlib.pyplot as plt
    852 
--> 853         if axes:
    854             ax1, ax2 = axes
    855             fig = ax1.figure

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### DIFF
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -850,7 +850,7 @@ class Inventory(ComparingObject):
         """
         import matplotlib.pyplot as plt
 
-        if axes:
+        if axes is not None:
             ax1, ax2 = axes
             fig = ax1.figure
         else:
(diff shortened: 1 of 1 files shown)


megies:
Looks good to me. Regression test would be nice but I guess this is pretty exotic..

trichter:
Works of course the same as before. I realized, that it is only a problem with `plt.subplots` because it returns a numpy array! I am merging...

megies:
> problem with plt.subplots because it returns a numpy array

oh.. i was already curious why you would use an numpy array for Axes instances... if that's the case we might want to check if we could have such problems elsewhere.. since it's not at all exotic then

PR #2068, opened 2018-02-06
add .idea to .gitignore

d-chambers:
### What does this PR do?

Adds the project folder associated with intellij IDE (includes pycharm) to .gitignore file. 

Pycharm is a common IDE, we don't want developers accidentally committing their IDE settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.pyo
 .cache
 .DS_Store
+.idea
 .project
 .pydevproject
 /.settings


krischer:
I guess this cannot hurt especially given that we already have a bunch of IDE specific files in there. Please not that you can also set a global `.gitignore` file that `git` will just always apply:

```bash
$ git config --global core.excludesfile '~/.gitignore_global'
```

This is especially useful if your IDE of choice has some project files and you don't want to tkeep adding the same files to each project's `.gitignore`.

d-chambers:
@krischer, thanks for the tip, I didn't know about the global gitignore feature.

PR #2069, opened 2018-02-07
mseed.utils.get_start_and_end_time returning sr=1 when sr=0.

dsentinel:
### get_start_and_end_time() returns sr of 1 when sr is 0
Another little piece that doesn't handle sr=0 channels such LOGs correctly


krischer:
Thanks for this. I think it makes sense to handle it the way you coded it. Can you please also add a test as well as a changelog entry? A simple way to test this is to just use one of the existing test files, load it to a `io.BytesIO` object and replace the two bytes in question with zeros and then read the modified object - that way no new test files would be required.

dsentinel:
>Thanks for this. I think it makes sense to handle it the way you coded it. Can you please also add a test as well as a changelog entry?

Happy to.

I have a LOG file that tripped this up (ASCII LOG from an RT-130) . It's 61Kb. I can add it, or do as you suggested, or cut it down to a few blocks.  A quick scan didn't reveal any samplerate=0 mseed in test data. May be worth having one?

krischer:
Please cut it down to one or two records if you choose to add. Personally I would prefer if you just generate the required test data on the fly but I'm okay with small test files as well if you prefer it.

dsentinel:
Should be ready to merge.

CHANGELOG.txt
```diff
@@ -12,6 +12,7 @@
  - obspy.io.mseed:
    * Ability to read files that have embedded chunks of non SEED data.
      (see #1981, #2057).
+   * Fix util.get_start_and_end_time returning sr=0 when sr=1 (see #2069)
```
krischer: Can you write out sampling rate to make it easier to understand?

dsentinel:
Cheers.
Of course flake bit me. :)

megies:
You can also see #2076 for a test file.

dsentinel:
Thanks @megies! I'll give it a go.
Are you suggesting that we include it in tests? If so I can add it as a test case, and #2076 can use as well.

krischer:
> Are you suggesting that we include it in tests? If so I can add it as a test case, and #2076 can use as well.

Sounds like a good plan. Especially if we have a couple more bugs related to traces with a sampling rate of 0.

dsentinel:
I added the LOG file to tests/data, and a generic reading test for it as well as a test case for this PR.
There was also #1990 related to sr=0, but this was limited to stats.

dsentinel:
Should be ready to merge if tests pass.

krischer:
Looks good to me. Thanks a bunch!

Issue #1794, opened 2017-05-19
Pathlib aware functions


krischer:
All of our I/O functions, mainly the various `read_X()` variants should be made aware of `pathlib` objects as these are getting more and more common.

markcwill:
While we're at it, can I throw in request for support for URL-type strings, or `urlparse.ParseResult` objects? It'd be really nice to at least have some interface to patch in a custom way to turn a string into a file or IO object. The world is moving toward massively parallel stateless apps, and more of our files are at an `http://`, or `s3://` endpoint, and I'm starting to pass local file locations like this: `file:///path/to/file.txt`. 

Just thinking out loud, not 100% sure how this would look... also don't know how much this steps on the `Client` implementations for remote requests...

krischer:
I'm not entirely sure what you mean so excuse me ig I misunderstood your request. `obspy.read("http://example.com/file")` already works. We could add support for a couple more protocols I guess but at a certain point it would just be simpler to do:

```python
def read_s3(path):
    with io.BytesIO() as buf:
        buf.write(s3.read_storage(path))
        buf.seek(0, 0)
        return obspy.read(buf)
```

markcwill:
My thought was, if you're type-checking `filename_or_url` for `Path`, why not also check for a urllib `ParseResult`, and have a hook or method or plugin for dealing with various URL schemes. I didn't realize you had support for http, but it looks like you just check for `://`, and then pass to `requests`, so there's not really any place to add other schemes, maybe monkey patch `obspy.core.util.download_to_file`. Right now, passing it any URL string that `requests` doesn't understand will throw an Exception I imagine.

So yes, I agree wrapping `read` is the best option for now. Just wondering if there isn't a better one via a plugin system. Maybe not.

megies:
If I understand correctly this would mean adding an entry point for all (or at least the most important) routines like `read`, `read_inventory` or `read_events`. Essentially replacing the current `if '://' in filename: # use requests for http downloading` with a plugin point for the ["scheme" part of the URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)..

```
    'obspy.plugin.uri': [
        'file = obspy.core.util.base:open_file_stub',
        'http = obspy.core.util.base:download_to_file',
        'https = obspy.core.util.base:download_to_file',
    ]
```

[...]

PR #2070, opened 2018-02-11
receiver gather 1.6 read support

d-chambers:
I created a [new branch](https://github.com/obspy/obspy/tree/read_rg16) for adding support for receiver gather 1.6 waveform format. This is the continuous data format output by Fairfield Nodal systems, although I believe SEGD is also supported. I am raising an issue to discuss a few potential sticking points before needlessly triggering a bunch of CI builds.

So a few issues I might need some guidance on:

1. Right now channel codes output are only 2, 3, 4, which corresponds to the rg1.6 spec describing how the instrument was deployed (https://imgur.com/a/4aneG) so there isn't a universally correct mapping to E,N. I suppose I should take a stab at adding the band code, and instrument code to make it seed compliant? The problem here is that the [supported orientation codes](https://ds.iris.edu/ds/nodes/dmc/data/formats/seed-channel-naming/) are 1,2,3 so it may be confusing to map channels 2,3,4 to it. 

2. I have the format spec in a pdf, but it has proprietary written all over it (literally). Fairfield has been very liberal in distributing it, however, so if it is necessary I will ask if it can be included. The pdf may take a few mb of space though.

3. Any thoughts in general on the code? Does it look approachable enough?

Any input @ThomasLecocq, you have probably given these type of issues some thought.


megies:
+DOCS so you can check automated docs build

d-chambers:
Any thoughts on the mac 3.6 failure in "test_sds_report"?  I don't have a mac to setup the environment and am not so keen on using a VM or docker.

megies:
> Any thoughts on the mac 3.6 failure in "test_sds_report"? I don't have a mac to setup the environment and am not so keen on using a VM or docker.

Just ignore.. that one's showing up every now and then and I never was able to figure out why..

obspy/io/rg16/__init__.py
```diff
@@ -1,20 +1,88 @@
 """
-Functions to read waveform data from Receiver Gather 1.6-1 foramt.
-This format is used for continuous data by Farfield's (fairfieldnodal.com)
-Zland product line (http://fairfieldnodal.com/equipment/zland).
+obspy.io.rg16 - receiver gather v1.6 read support for ObsPy
+=============================================================
```
megies: Same length as the text please, sphinx will show a warning otherwise (adding on top of hundreds or thousands of warnings currently.. ;-)

obspy/io/rg16/__init__.py
```diff
@@ -1,20 +1,88 @@
...
+    can be found `here <https://imgur.com/a/4aneG>`_.
+
+    2. Because there is not a standard deployment orientation, channel codes
+    contained in the file format are 2, 3, and 4. See the above link for the
+    orientation in respect to instrument orientation. By default these are
+    mapped to Z, 1, 2 but a custom mapping can be passed to
+    :func:`obspy.io.rg16.read_rg16`.
+
+    3. The default polarity of channel 2 (vertical) is down. This is opposite
+    of what one might expect.
```
megies: Hmm.. maybe there should be a kwarg to `read_rg16` for switching the polarity so that this point is more prominently visible? And if it's not set but the channel is mapped to `Z` maybe show a warning?

obspy/io/rg16/__init__.py
```diff
@@ -1,20 +1,88 @@
...
+The rg16 format can be read using two methods:
+
+1. Using the standard :func:`~obspy.core.stream.read` function. Optionally,
+   the format parameter can be specified as "rg16" in order to avoid spending
+   time checking all the other foramts.
+
+2. Using the :mod:`obspy.io.rg16` specific function
+   :func:`obspy.io.rg16.read_rg16` function.
+
+There are two noteworthy parameters to the read_rg16 function:
```
megies: .. and also when using `read()`. 

You show it in the examples alright but it might not hurt to mention it also here right away.

ThomasLecocq:
I'm not sure we really want to modify the "inline" & "crossline" to E and N components... Some nodal users just use them the in-line component to orient the nodes, independent of the North. Practically, it makes sense for many to use the mapping to seed default, but not sure it's really desired for a "standard" rg16 file reader.

megies:
The default right now is to map to `Z12`, see https://github.com/obspy/obspy/pull/2070/files#diff-cc114ae5e1ec3ab959cf675b253a0707R67.

But.. reading above that Z might be positive down, maybe the default should even be `123`? So that the user is immediately aware that orientation of components has to be taken very seriously and likely figured out for each setup?

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,236 @@
...
+
+# since UTCDateTime cannot be compared to np.inf in py27 get a large timestamp
+# after which I will be dead (somebody else's problem)
+BIG_TS = UTCDateTime('3000-01-01').timestamp
+
+# default channel mapping
+COMPONENT_MAP = {2: 'Z', 3: '1', 4: '2'}
+
+# map sampling rate to band code according to seed standard
+BAND_MAP = {2000: 'G', 1000: 'G', 500: 'D', 250: 'D'}
```
megies: we have something similar already, maybe it's a good idea to put it there in `core.util`.. a helper function to get appropriate band code for sampling rate and broadband yes/no..

see https://github.com/obspy/obspy/blob/5762fb5097404dbb04bf91fba6ecacb1075f0bdf/obspy/core/util/misc.py#L42 and https://github.com/obspy/obspy/blob/5762fb5097404dbb04bf91fba6ecacb1075f0bdf/obspy/core/util/misc.py#L66

d-chambers:
@megies,
I like your idea of using 1,2,3 better than adding an option to reverse polarity. It will always make sense to reverse the Z component, but only sometimes to reverse the others, depending on deployment orientation. Good point @ThomasLecocq, I am sure the nodes are getting deployed/oriented in many different ways. Talking to my contacts at the University of Utah, they indicated that they, and most people they know, always deploy the 3C instruments with the gold contact terminals facing north. Is that consistent with your experience? 

Here is the cleanest approach I can think of:

1. Remove the component map option. Simply map to 1, 2, 3 as suggested. This will force the user to think about the issue if they want to map back to ZNE. 

2. Add an argument called `standard_deployment` (or something like that) that will be False by default. If True, it will imply the files contain only Z component data, or if 3C are used, that the gold contact terminals were deployed facing North. The vertical trace can then be multiplied by -1 and the channels can be correctly mapped to 2 -> Z, 3 -> N, and 4 -> E.

obspy/io/rg16/__init__.py
d-chambers: I think the solution I proposed would be a bit cleaner. I don't want the function to get cluttered up with too many parameters.

d-chambers:
Thinking about this more, it might be better to leave the default component codes as 2, 3, 4. This way when the user looks at the digram the docs link to it will match the component codes in the stream. I know this will break slightly with seed convention, which prescribes 1, 2, 3, but removing a potential stumbling block is probably worth a slight deviation from convention.

obspy/io/rg16/core.py
d-chambers: Looking closer at this I don't think it would be a great idea to do this in the read_rg16 function, primarily for efficiency reasons. The rg16 files can be quite large (sometimes 10 GB or more) with thousands to hundreds of thousands of traces. Reading these files is already slower than I would like. 

Since the nodes can only have sampling_rates  of 2000, 1000, 500, and 250 getting the band code is a simple dictionary lookup as it stands now. If I created a general function, it would incur the cost of a function call, as well as bounds checking to figure out which bin the sampling rate falls into, then additional cost for selecting between instruments of different corner period.

d-chambers: I could be over optimizing and the cost might be negligible, I am not sure.

megies:
> that the gold contact terminals were deployed facing North. The vertical trace can then be multiplied by -1 and the channels can be correctly mapped to 2 -> Z, 3 -> N, and 4 -> E.

No idea, really, but my feeling is that it might not be safe enough to assume there really is an agreed-upon "standard" deployment..?

> but removing a potential stumbling block is probably worth a slight deviation from convention.

Yeah, maybe you're right that it's best left at the original component codes. But the component_map option can still be left in I think, just set `None` and not doing anything by default

d-chambers:
> No idea, really, but my feeling is that it might not be safe enough to assume there really is an agreed-upon "standard" deployment..?

That is fair. I don't think there is an agreed upon standard, but from the few people I have talked to (in the western US) this is how they are deployed, so it certainly isn't uncommon. 

I guess my reason for wanting this parameter is a bit selfish; I would prefer to always deploy the instrument my agency has the same way, then be able to simply pass one parameter whenever reading these files to avoid adding the complexity of component mapping and polarity flipping in every file that reads them, or having to create another package to import that does so. However, I know I am not the only one who will use this feature.

In light of that, maybe if we renamed it to "terminals_north" or something like that it would be acceptable to include? 

>  But the component_map option can still be left in I think, just set None and not doing anything by default

If you see value in having it there I will reintroduce it.

megies:
> If you see value in having it there I will reintroduce it.

Ah, yeah.. maybe we can just leave it out, that might be clearest to force people to really think about the components twice and then manually map them..

megies:
> In light of that, maybe if we renamed it to "terminals_north" or something like that it would be acceptable to include?

There should be no disadvantages for people putting together a canonical PR instead of keeping their plugins local, and since you're the main user of this plugin, I think we can live with such a dedicated option for your purposes..

d-chambers:
I will try restarting the job to see if it happens again.

megies:
OK, looks like this can use a final review.. volunteers to the front. ;-)

ThomasLecocq:
Ok I'll do one

obspy/io/rg16/util.py
```diff
@@ -0,0 +1,234 @@
...
+import numpy as np
+
+from obspy import Trace
+
+BITMAP = np.array([1, 2, 4, 8])[::-1].reshape(4, 1)
+
+TENS = np.power(10, range(12))[::-1]
+
+
+def open_file(func):
```
krischer: We should really really have something like this in `obspy.core`. Unfortunately it would require touching every single `.io` plugin which is a lot of work :(

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+    :type merge: bool
+    :param contacts_north:
+        Setting this parameter to True indicates the file either contains 1C
+        traces or that the instruments were deployed with the gold contact
+        terminals facing north. If this parameter is used, it will map the
+        components to Z, N, and E (if 3C) as well as correct the polarity for
+        the vertical component.
+    :type contacts_north: bool
+    :return: An ObsPy :class:`~obspy.core.stream.Stream` object.
+
```
krischer: unnecessary newline

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    if not is_rg16(filename):
+        raise ValueError('read_fcnt was not passed a Fairfield RG 1.6 file')
```
krischer: I would remove this. We mostly expect users to to use the core `read()` function which performs this test before passing a file to any underlying plug-in. If users are sure of the format they can specify it directly resulting in a moderate speed-up (as you also mention above). Removing this check would slightly increased the speed-up. It is IMHO the user's responsibility in this case.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+def is_rg16(filename, **kwargs):
+    """
+    Determine if a file or buffer contains an rg16 file.
+
+    :param filename: A path to the file or a buffer of an opened file.
+    :type filename: str, buffer
+    :return: bool
+    """
+    try:
+        filename.seek(0)
```
krischer: I'd vote against seeking here as it would then enable passing a file-pointer anywhere in a file or byte stream. I'm not sure if this is useful for this particular format but at least it would be consistent with the rest of ObsPy.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+    trace_position = data_block_start
+    while True:  # read traces until parser falls of the end of file
+        try:
+            theader = _read_block(fi, trace_header_block, trace_position)
+        except ValueError:  # this is the end, my only friend, the end
+            break
+        # get stats
+        stats = _make_stats(theader, gheader, standard_orientation)
+        # expected jump to next start position
+        jumps = stats.npts * 4 + theader['num_ext_blocks'] * 32 + 20
```
krischer: Any chance this could be zero with corrupt data? It would then be trapped in an infinite loop.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+# map sampling rate to band code according to seed standard
+BAND_MAP = {2000: 'G', 1000: 'G', 500: 'D', 250: 'D'}
+
+# geophone instrument code
+INSTRUMENT_CODE = 'P'
+
+# mapping for "standard_orientation"
+STANDARD_COMPONENT_MAP = {'2': 'Z', '3': 'N', '4': 'E'}
+
+# ------------------- read and format check functions
```
krischer: Please remove this line.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+    return con1 and manufacturer_code == 20
+
+
+# ------------ helper functions for formatting specific blocks
+
+
+def _make_traces(fi, data_block_start, gheader, head_only=False,
+                 starttime=None, endtime=None, merge=False,
+                 standard_orientation=False):
+    """ make obspy traces from trace blocks and headers """
```
krischer: Please place triple quotes on their own lines for consistency.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+        sample_format = _read(filename, 2, 2, 'bcd')
+        manufacturer_code = _read(filename, 16, 1, 'bcd')
+        version = _read(filename, 42, 2, None)
+    except ValueError:  # if file too small
+        return False
+    con1 = version == b'\x01\x06' and sample_format == 8058
+    return con1 and manufacturer_code == 20
+
+
+# ------------ helper functions for formatting specific blocks
```
krischer: please remove.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+                data = -data
+        traces.append(Trace(data=data, header=stats))
+        trace_position += jumps
+    if merge:
+        traces = _quick_merge(traces)
+    return traces
+
+
+def _make_stats(theader, gheader, standard_orientation):
+    """ make Stats object """
```
krischer: Please place triple quotes on their own lines for consistency.

obspy/io/rg16/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    return BytesIO(byte_str)
+
+
+class TestRG16Util(unittest.TestCase):
+    """
+    Tests for the read function, which should read all the weird binary
+    formats used in rg16.
+    """
+
+    # --- tests bcd
```
krischer: Please remove.

obspy/io/rg16/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_ff_raises(self):
+        """
+        Ensure FF raises. BCD values for any half byte past 9 should raise.
+        """
+        with self.assertRaises(ValueError) as e:
+            _read(byte_io(b'\xFF'), 0, 1, 'bcd')
+        assert 'invalid bcd values' in str(e.exception)
+
+    # --- test half byte reads
```
krischer: Please remove.

obspy/io/rg16/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    ]
+
+    def test_read_half_bit(self):
+        """
+        Ensure reading half bytes (4 bit) works.
+        """
+        for byte, format, answer in self.halfsies:
+            self.assertEqual(_read(byte_io(byte), 0, 1, format), answer)
+
+    # --- test 24 bit (3 byte) reads
```
krischer: Please remove.

obspy/io/rg16/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+
+    why_use_3_bytes = [  # seriously, how expensive is one extra byte!?
```
krischer: Hehe :-) One of the format (forgot which one) in ObsPy has some kind of 6 byte number representation...

obspy/io/rg16/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    ]
+
+    def test_read_3_bytes(self):
+        """
+        Ensure 3 byte chunks are correctly read.
+        """
+        for byte, format, answer in self.why_use_3_bytes:
+            self.assertEqual(_read(byte_io(byte), 0, 3, format), answer)
+
+    # --- test backup
```
krischer: Please remove.

obspy/io/rg16/util.py
krischer: Please use the decorator package. ObsPy already depends on it and function signatures including documentation are correctly preserved when it is used. I'm actually kind of confused why documentation seems to be correct here. @megies Do you know? It is a thing with Python 3?

obspy/io/rg16/util.py
```diff
@@ -0,0 +1,234 @@
...
+                return func(*args, **kwargs)
+        except TypeError:  # assume we have been passed a buffer
+            if not hasattr(args[0], 'read'):
+                raise  # type error was in function call, not in opening file
+            return func(*args, **kwargs)
+
+    return _wrap
+
+
+# -------------------- functions for byte chunk reads
```
krischer: please remove.

obspy/io/rg16/util.py
```diff
@@ -0,0 +1,234 @@
...
+
+@_register_read_func('<i.')
+def _read_4_bit_right(fi, length):
+    """ read the four bits on the right """
+    assert length == 1, 'half byte reads only support 1 byte length'
+    ints = np.fromstring(fi.read(length), dtype='<u1')[0]
+    return np.bitwise_and(ints, 0x0f)
+
+
+# ---------------------- stream manipulation stuff
```
krischer: please remove.

setup.py
```diff
@@ -99,7 +99,7 @@
     'seismology', 'seismogram', 'seismograms', 'shapefile', 'signal', 'slink',
     'spectrogram', 'StationXML', 'taper', 'taup', 'travel time', 'trigger',
     'VERCE', 'WAV', 'waveform', 'WaveServer', 'WaveServerV', 'WebDC',
-    'web service', 'Winston', 'XML-SEED', 'XSEED']
+    'web service', 'Winston', 'XML-SEED', 'XSEED', 'RG-1.6']
```
krischer: Please keep the alphabetical order.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+INSTRUMENT_CODE = 'P'
+
+# mapping for "standard_orientation"
+STANDARD_COMPONENT_MAP = {'2': 'Z', '3': 'N', '4': 'E'}
+
+# ------------------- read and format check functions
+
+
+@open_file
+def read_rg16(filename, headonly=False, starttime=None, endtime=None,
```
krischer: Please make this a private function.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+    theader_start = eheader_start + (ex_headers * 32)
+    # get traces and return stream
+    traces = _make_traces(filename, theader_start, gheader, head_only=headonly,
+                          starttime=time1, endtime=time2, merge=merge,
+                          standard_orientation=contacts_north)
+    return Stream(traces=traces)
+
+
+@open_file
+def is_rg16(filename, **kwargs):
```
krischer: Please make this a private function.

obspy/io/rg16/util.py
krischer: Also please make this a private function so we don't buy into preserving it long-term.

obspy/io/rg16/core.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+# header block, combines header block one and two
+general_header_block = [
+    ('channel_sets', 28, 1, 'bcd'),
+    ('num_additional_headers', 11, 1, '>i.'),
+    ('extended_headers', [30, 37], [1, 2], ['bcd', '>i2']),
+    ('external_headers', [31, 39], [1, 3], ['bcd', '>i3']),
+    ('record_length', 46, 3, '>i3'),
+    ('base_scan', 22, 1, '>i1'),  # https://imgur.com/a/4aneG
```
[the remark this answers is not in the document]
d-chambers: The two images not included show how base scan intervals are mapped to sampling rate, and the appropriate datatype for the time series. Both of these are handled by the code so it is not very important the users have access to them. Maybe I should remove  the link since the important figure is now included?

obspy/io/rg16/util.py
d-chambers: functools.wraps (in the std lib) handles copying the `__doc__` and `__name__` attributes. On python 3.4+ it also copies signature info.

obspy/io/rg16/core.py
d-chambers: Good point, I will add an assert that jumps != 0. It shouldn't ever happen but if it does it would cause a big problem.

setup.py
d-chambers: There are some other ordering issues in this list, want me to fix them?

d-chambers:
> At least the _is_XXX() function should set the file/buffer pointer to the original position after it has terminated - otherwise all other _is_XXX() would no longer function correctly if it fails. I guess this is debatable for the _read_XXX() function.

I added a seek to the _open_file decorator to reset the position to 0 if a file object is passed. If a path is passed it wont do this but that is a non-issue as each other _is_whatever function will have to open the file.

d-chambers:
> We should really really have something like this in obspy.core. Unfortunately it would require touching every single .io plugin which is a lot of work :( 

I agree, having an open file decorator for each io method would be cleaner, and allow for solving issues like  #1794 trivially. I looked at doing this but quickly got discouraged when the magnitude of the task became clear. 

Also, there should probably be an io.utils module with functions for things like reading strangely shaped data structures (3, 6, 7.5 bytes, etc.). Then we could profile different approaches and make sure all the IO methods benefit from speeding up some of these functions. This is probably the next step for implementing support for seg-D (very similar to rg16) mentioned in #1058. It would essentially use the same utils as RG16.

derick-hess:
I haven't fully reviewed the code yet, but one thing to consider with Fairfield rg16 is that the data is not output in counts but in mV. It's not really much of a problem, it's just not what many people expect. 

At PASSCAL when we archive rg16 data we go ahead and convert it to counts so it is consistent with other archived data and the response information we provide to users assumes the data is in counts. 

We had a few users convert the rg16 data themselves to miniseed (as written in mV) with their own scripts then try to use our response information and had issues.

d-chambers:
@derick-hess that is an issue to consider. Maybe there should be an option in the read function to convert to counts?

derick-hess:
@d-chambers That could be a good way to handle it. 

The issue first popped up on us when I was helping to review a paper a group was doing on characterizing the geophone response of the nodes. They used their own script and obspy to convert the data to miniseed, but assumed the data was in counts so some of their calculations were wrong, but once that was sorted out everything was fine.

It should be fine in mV or counts as long as the user is aware of what the output is so they can correctly remove the response.

d-chambers:
Ok, I am thinking the default output should be Mv and counts are optional. That way the output matches what one might find on the node spec sheet.

Did you want to take a stab at implementing the counts option? it would be a good excuse to review the code :wink:

d-chambers:
Hey @derick-hess, do you have any interest/time to add the counts option we were discussing? If not we need to look at merging this in order to avoid holding up the next Obspy release. If you don't have time right now we can always add it later.

derick-hess:
I think go ahead and merge it for now. It's not a big deal. I"m pretty busy right now, but when I have time I can always come back and add the feature.

obspy/io/rg16/core.py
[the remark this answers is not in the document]
megies: You could replace the link to point to the image in the docs..

Also, if you feel the information in the other two pics might be important, I think it could easily be included in a ASCII comment right here

Issue #2072, opened 2018-02-13
The future of time


d-chambers:
No, this is not the title of a book written by Stephen Hawking. This issue aims to outline a proposal for future developments to the `UTCDateTime` object and to focus discussion on the topic that has been taking place in a few other threads.

Here are the problems I see, as well as a proposal for fixing the issues. I may be overlooking something so please don't hesitate to disagree. 

# The problem

1. UTCDateTime instances cannot (and should not) be used as keys in a dictionary or members in a set because they are mutable (ie they cannot be safely hashed).

2. UTCDateTime fuzzy comparisons still cause some strange behavior in sorting, min/max, etc. despite some improvements made in #2049. Some of the oddities outlined in #1765 still exists, eg the second half of [this example](https://github.com/obspy/obspy/issues/1765#issuecomment-297705550).

3. Internal representation of nanoseconds results in storing larger numbers than necessary (though the used memory might be the same) for precisions less than 9, and precisions greater than 9 are not possible.

# Proposed changes to `UTCDateTime`

## 1. Make UTCDateTime immutable
There has already been some discussion on #568, #1452, and probably others. Immutability will allow the safe hashing `UTCDateTime` instances as well as protect against a subtle class of bugs (eg unintended mutations). 

This would, of course, mean that instance attributes would no longer be settable. So rather than doing something like this:
```python
import obspy
utc = obspy.UTCDateTime()
utc.year = 1989
utc.month = 5
utc.day = 4
utc.hour = 12
utc.precision  = 3
```  
The user would be required to do this:
```python
import obspy
utc = obspy.UTCDateTime(year=1989, month=5, day=4, hour=12, precision=3)
```

## 2. Use the smallest possible int to represent time

Upon initialization, `UTCDateTime` would round the inputs to the nearest unit according to precision and only store the smallest possible integer to represent the desired precision. For example, for the default  precision of 6, the object would store an int representing microseconds. If precision was set to 0 it would store an int representing seconds, and so on. 

Arguably, if a precision is specified any significant figures exceeding the specified precision are just noise anyway, no reason to store them.  

Moreover, according to [this](https://docs.python.org/2/library/stdtypes.html#numeric-types-int-float-long-complex), longs have no limit on precision, at least until memory is full, so an arbitrary precision could be represented as efficiently as possible by `UTCDateTime`. 

## 3. Raise an error when UTCDateTime objects of different precisions are compared 

[...]

PR #2077, opened 2018-02-16
UTCDateTime immutability warning

d-chambers:
### What does this PR do?

This PR does two things:

1. Add depreciation warnings when attributes are set on instances of `UTCDateTime` outside of the `__init__` method. 

2. Add new method to `UTCDateTime ` class called `replace`. 

The replace method is modeled after the [replace method](https://docs.python.org/3/library/inspect.html#inspect.Signature.replace) on `inspect.Signature` from the python 3.6 standard library. It allows keeping most the state from the initial object, making a copy, and changing a small amount of state before initialization the copy. 

The replace method allows some behavior that was possible by setting attributes of `UTCDateTime` instances that was not simple without mutation using the previous methods. For example:

```python
from obspy import UTCDateTime


# --- set day to 15th of month that a timestamp belongs to

# using mutation (now raises depreciation warning)
t1 = UTCDateTime(999_999_999)
t1.day = 15
print(t1)

#using replace to create new object
t1 = UTCDateTime(999_999_999)
t2 = t1.replace(day=15)
print(t2)
```

This is the first phase in migrating to an immutable UTCDateTime class. See #2072 

### PR Checklist
- [X] Correct base branch selected? `master` for new fetures, `maintenance_...` for bug fixes
- [X] All tests still pass.


d-chambers:
+DOCS

obspy/core/utcdatetime.py
```diff
@@ -1204,6 +1157,16 @@ def __hash__(self):
...
         return None
 
+    def __setattr__(self, key, value):
+        # raise a warning if overwriting previous ns (see #2072)
+        if self._initialized and not self._has_warned:
+            msg = ('Setting attributes on UTCDateTime instances will raise an'
+                   ' Exception in a future version of Obspy.')
+            warnings.warn(msg, ObsPyDeprecationWarning)
+            # only issue the warning once per object
+            self.__dict__['_has_warned'] = True
```
krischer: Is it not Python's default behavior to only raise a warning message once? But I'm also sometimes a bit lost with how Python determines if a warning should be shown or not and it also behaves differently across versions.

obspy/core/utcdatetime.py
```diff
@@ -1502,6 +1458,53 @@ def _set_precision(self, value=6):
 
     precision = property(_get_precision, _set_precision)
 
+    def replace(self, **kwargs):
+        """
+        Return a new UTCDateTime object with one or more parameters replaced.
+
+        The following parameters are supported: year, month, day, julday,
+        hour, minute second, microsecond.
```
krischer: I feel like this should also support nanoseconds and the precision parameter.

obspy/core/utcdatetime.py
megies: Yeah, I also think we should rely on Python managing how often this warning is shown.. it should only show it once by default anyway (it's keeping that magic per-module warning registry)?

d-chambers: I thought the same thing but for some reason multiple warning were being issued from the same instance. I will look more into it.

obspy/core/utcdatetime.py
d-chambers: I thought about supporting nanoseconds but it may introduce some confusion into the API for two reasons:

1. The `ns` public attribute means nanoseconds since 1970, but the other parameters in the `replace` method are relative to the next higher unit of time (eg hour means hour in current day, not hours since 1970). So should nanoseconds be relative to microseconds, or seconds? If so, the usage of nanosecond and the `ns` attribute are inconsistent. If it means nanoseconds since 1970 there is no need to use it in replace, just create a new instance.

2. The default precision does not support nanoseconds, and the vast majority of users do not change the precision. What should replace do if the nanosecond parameter is used but the precision is the default value? Raise a warning and do nothing?

As for replace with the precision parameter: I think this is doable but there are two hangups. Under the proposal outlined in #2072 the internal integer representation of time is rounded according to precision upon init. This is necessary to remove the "fuzzy" behaviour and all the problems it brings. Because of this, however, if `replace` is used to go from a higher precision to a lower one it may cause some confusing behaviour. For example:

```python
from obspy import UTCDateTime
utc = UTCDateTime(10.222, precision=0)
utc.timestamp  # -> 10  (10.222 was rounded to 10 based on a precision of 0)
utc2 = utc.replace(precision=3)
utc2.timestamp  # -> 10.000 (but the user might expect 10.222) 
```
I suppose we could issue a blanket warning whenever a precision greater than the original is set? 

Also, `replace` was intended to avoid messy inits that could produce the same behaviour. For example:
```python
from obspy import UTCDateTime
utc = UTCDateTime()
# get same utc with hour 10
# using replace
utc2 = utc.replace(hour=10)
# using init
utc2 = UTCDateTime(year=utc.year, month=utc.month, day=utc.day, hour=10, ...)  # and so on
``` 
With replacing a precision, however, it is not unreasonable to simply init a new object. 
```
from obspy import UTCDateTime
utc = UTCDateTime()
utc2 = UTCDateTime(utc, precision=9)
```

krischer: You clearly gave this a lot more thought than I did and I agree with your reasoning. Thanks a lot for the very comprehensive explanations. Can you add one or two sentences to the docstring to state this in a condensed form?

d-chambers: stating this clearly in one or two sentences is the trick... will do my best

obspy/core/utcdatetime.py
d-chambers: Ok, I can confirm that python 3.6 (on ubuntu 16)  raises 4 warnings for `test_change_precision_raises_warning` and two warnings for `test_compare_utc_different_precision_raises_warning` when line the `_has_warned` attr is not used. 

I have no idea why this is the case, but I am struggling to see a reason to be too concerned with leaving `_has_warned` in the code because it is only a temporary patch until `UTCDateTime` is immutable. If someone feels it is really important I will spend more time on it, but I would prefer to leave it alone.

krischer: > I have no idea why this is the case, but I am struggling to see a reason to be too concerned with leaving _has_warned in the code because it is only a temporary patch until UTCDateTime is immutable. If someone feels it is really important I will spend more time on it, but I would prefer to leave it alone.

I agree. Let's leave it how you currently implemented it.

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1320,6 +1321,107 @@ def test_zero_precision_doesnt_print_dot(self):
...
+            day=18,
+            hour=18,
+            minute=30,
+            second=11,
+            microsecond=122255,
+        )
+
+        utc = UTCDateTime(**test_dict)
+
+        # iterate over each settable paramter and change
```
barsch: typo: parameter

CHANGELOG.txt
```diff
@@ -4,6 +4,10 @@ master:
      (see #1997).
    * UTCDateTime objects will now always evaluate equal if their string
      representations are equal (see #2049).
+   * UTCDateTime objects now issue depreciation warnings when setting any
+     attributes outside of init, or comparing UTCDateTime objects with
+     different precisions (see  #2077).
```
barsch: extra space before issue number, next line too

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1320,6 +1321,107 @@ def test_zero_precision_doesnt_print_dot(self):
...
+            # make sure only the settable parameter changed in utc2
+            for time_attribute in test_dict:
+                default = getattr(utc, time_attribute)
+                current = getattr(utc2, time_attribute)
+                if time_attribute == attr:
+                    self.assertEqual(current, default + 1)
+                else:
+                    self.assertEqual(current, default)
+
+        # test juldain day
```
barsch: typo: julian

obspy/core/utcdatetime.py
barsch: ```_has_warned``` mechanic is really ugly - but if its temporary please mark it in the code as temporary by adding an appropriate comment

obspy/io/sac/sactrace.py
```diff
@@ -1004,22 +1004,20 @@ def reftime(self, new_reftime):
...
             old_reftime = self.reftime
 
-            # find the milliseconds and leftover microseconds for new reftime
-            _, rem_microsecs = _ut.split_microseconds(new_reftime.microsecond)
-
             # snap the new reftime to the most recent milliseconds
             # (subtract the leftover microseconds)
-            new_reftime.microsecond -= rem_microsecs
+            ns = new_reftime.ns
+            time = UTCDateTime(ns=(ns - ns % 1000000))
```
barsch: time is a python module - you should avoid it as parameter/variable name if possible

krischer:
Thanks for your work on this!

krischer:
https://www.python.org/dev/peps/pep-0515/

megies:
Interesting..

Issue #2083, opened 2018-03-01
A problem with Obspy on Raspbery PI -- mseed write -- solved (ish)


starfishprime101:
Hi
I have had problems running obspy (python3) on a raspberry pi leading to error messages such as
ValueError: /usr/lib/python3/dist-packages/obspy/core/util/../../lib/libmseed_Linux_32bit_py34.cpython-34m.so: undefined symbol: LM_SIZEOF_OFF_T. Here I am trying to write in mseed format

This can be solved by renaming all the files in /usr/lib/python3/dist-packages/obspy/lib/ to omit the "-arm-linux-gnueabihf" element of their filenames

i.e. 
"sudo cp /usr/lib/python3/dist-packages/obspy/lib/libmseed_Linux_32bit_py34.cpython-34m-arm-linux-gnueabihf.so                                                                                                                                           /usr/lib/python3/dist-packages/obspy/lib/libmseed_Linux_32bit_py34.cpython-34m.so "
 

and so on for the others in the directory

Trust this is of some help

Ian

megies:
I trust it you installed the current deb packages from our apt repo? Are you on Raspbian? What version?

starfishprime101:
Good point about versions

I am running Jessie (debian)  8 kept fully up to date on stable. Obspy is version 3, agin kept up to date with apt-get.

I had to do a similar thing last summer which may have been a previous version???

megies:
Hrrmmm.. I can confirm this problem. Strange how this did not surface during packaging since it also involves running the tests on the built packages..

krischer:
Does this file exist in the distribution prior to renaming? ` /usr/lib/python3/dist-packages/obspy/core/util/../../lib/libmseed_Linux_32bit_py34.cpython-34m.so`

megies:
Nah, I have a shell script that renames built libs to match our naming standards. During deb packaging the Debian canonical naming is applied and I have to rename all these files.

krischer:
Where does this come from then? If ObsPy could not find the shared library it would raise a different error but it complains about undefined symbols which might mean its an old version of the library. Do you think its possible that this is some kind of remnant from a previous obspy installation?

megies:
Oh, overlooked that detail.. but for me locally on a clean ARM docker image it was a normal `ImportError: Could not load shared library "libmseed_Linux_32bit_py34.cpython-34m.so".` so yeah, I guess it was some remnant from an old local monkey-patch renaming for the OP..

[...]

PR #2085, opened 2018-03-06
docker: Some changes for Deb packaging

megies:
- add infos about qemu use for ARM
- remove one EOL Ubuntu
- use `deb__build_debs.sh` from target branch instead of current local repo version, that way it's cleaner to set the `DEBVERSION` and overall version of the package in the deb building branch, as opposed to just locally


megies:
Used this branch to package new deb for raspbian (#2083) so works and merging

PR #2088, opened 2018-03-15
add Inventory.remove(...) method

megies:
### What does this PR do?

Adds a method `Inventory.remove(...)` to get a view of the inventory object with matching items removed (kind of complementary to `Inventory.select(...)` and also using it internally).

See obspy users mailing list: http://lists.swapbytes.de/archives/obspy-users/2018-March/002699.html

### PR Checklist
- [x] If the PR is making changes to documentation... +DOCS
- [x] ~~If any network modules should be tested for the PR..~~
- [ ] All tests still pass.
  - [x] doctests
  - [x] unit tests

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -8,6 +8,7 @@ master:
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
    * Added replace method to UTCDateTime class (see #2077).
+   * Added remove method to Inventory class (see #2088).
  - obspy.io.reftek:
    * Implement reading reftek encodings '16' and '32' (uncompressed data,
      16/32bit integers, see #2058 and #2059)
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -556,6 +556,143 @@ class Inventory(ComparingObject):
         inv.networks = networks
         return inv
 
+    def remove(self, network=None, station=None, location=None, channel=None,
+               keep_empty=False):
+        """
+        Returns the :class:`Inventory` object that excluding the
+        :class:`~obspy.core.inventory.network.Network`\ s /
+        :class:`~obspy.core.inventory.station.Station`\ s /
+        :class:`~obspy.core.inventory.channel.Channel`\ s that match the given
+        criteria (e.g. all channels with ``channel="EHZ"``).
+
+        .. warning::
+            The returned object is based on a shallow copy of the original
+            object. That means that modifying any mutable child elements will
+            also modify the original object
+            (see https://docs.python.org/2/library/copy.html).
+            Use :meth:`copy()` afterwards to make a new copy of the data in
+            memory.
+
+        .. rubric:: Example
+
+        >>> from obspy import read_inventory, UTCDateTime
+        >>> inv = read_inventory()
+        >>> inv_new = inv.remove(network='BW')
+        >>> print(inv_new)  # doctest: +NORMALIZE_WHITESPACE
+        Inventory created at 2014-03-03T11:07:06.198000Z
+            Created by: fdsn-stationxml-converter/1.0.0
+                    http://www.iris.edu/fdsnstationconverter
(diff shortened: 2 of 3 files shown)


megies:
I think this is pretty much ready but only misses unit tests, if somebody feels like adding some, go ahead

megies:
CC @ewolin-usgs

krischer:
Yea we definitely need something like this. I wonder if it would be nicer to do it as part of the `.select()` method. Either

```python
inv.select(network_exclude="BW")
```

or 

```python
inv.select(network="!BW")
```

d-chambers:
I would vote on leaving it a separate method. 

Often it makes the interface more intuitive to have more methods (each with a descriptive name) than to have fewer methods that take more parameters or parameters that can do multiple things. There might also be some undefined behaviour, or at least added complexity, to account for inclusive and exclusive selectors in the same function call.

krischer:
Consider me convinced.

ewolin-usgs:
Wow, that was quick!  Thank you so much :)

megies:
Added unit tests. Ready for review.

megies:
> I would vote on leaving it a separate method.
> 
> Often it makes the interface more intuitive to have more methods (each with a descriptive name) than to have fewer methods that take more parameters or parameters that can do multiple things. There might also be some undefined behaviour, or at least added complexity, to account for inclusive and exclusive selectors in the same function call.

Agree 100%.
Simple is better than complex.

d-chambers:
What do you all think about making this a `pop` method? Similar to `dict.pop`; it would remove the selected items from the inventory and return a new inventory that contains the removed items. If you don't want the removed items simply don't assign the output to anything and it gets garbage collected.

It would have two advantages over `remove`:

1. It shares the name and behaviour with widely used method of the most common python data structure (I guess that makes it "intuitive" and "pythonic").
2. It can be used for deleting and selecting in one step.  

Maybe too complicated though? Up to you @megies, you are doing the work :smirk:

krischer:
I'm not in favor of using `.pop()`. Basically all subsetting/selection methods on the ObsPy objects return modified shallow copies while leaving the original objects intact. I feel like we should keep these semantics; otherwise it might be confusing if similar methods behave differently.

> It can be used for deleting and selecting in one step.

This can also be done with (a tiny bit more code):

```python
inv = inv.remove(...)
```

d-chambers:
@krischer, good point. Modifying the inventory in place is not the right thing to do if other selectors return shallow copies.

krischer:
This looks good to me. During review I got thinking that it would be really useful to have many of the set methods available on inventories, e.g.

```python
inv.union(inv2)
inv.intersection(inv2)
inv.difference(inv2)
...
```

But this is a larger effort to implement.

dsentinel:
I implemented remove and some other helpful functions for general inventory building and 
 manipulation such as:

- smart merging of 2 inventories (union?)
- propagating epoch changes up (such as from chan to station and net)
- get_parent (chan gives you sta, sta gives net)
- create_responses for a station where  NRL keys for sr and gain get sub'd for appropriate channels
- dip and azimuth population from defaults
- building a partial inventory from mseed header scan.
etc...  Off the top of my head

(no diff or intersect, I haven't had such use cases)

Most of this is in subclasses and mixin's of obspy objects, and some unfortunately creeped into my business logic.

I would be happy to push upstream if there is interest, but It will take a little time to organize and present for cherry picking and ensure works for general use and not just my needs.

Related to the NRL feature above: I have implemented and have a dream of standardizing how NRL and custom responses as well as other equipment information such as dip azimuth and SOH chans can be added and used with StationXML.
If there is interest and traction, we can start a discussion somewhere appropriate, create an implementation and possibly get FDSN to extend StationXML.

Apologies for babbling on an already merged PR

krischer:
These do sound pretty useful - and even if they are not suitable to be directly included into ObsPy they could go into the tutorial section as a sort of cookbook for other people to follow. So yes - please open a new issue/PR where we can discuss the details :-)

Issue #1910, opened 2017-10-06
Request/question: Slow QuakeML reading


calum-chamberlain:
QuakeML reading is quite slow - in my [naive test](https://gist.github.com/calum-chamberlain/a455a6f38a8dc6360ed0e52498d7799e#file-obspy_event_io-ipynb), about 10x slower than writing. For large catalog files this becomes a bit of an issue.

The main point of this issue is to ask:
1. Is there a faster IO alternative already implemented?
2. It looks to me like the main slow-down is the loop over events in obspy.io.quakeml.core.Unpickler._deserialize, I wonder if this can be a threaded operation?
3. If none of the above, does anyone have any ideas for speeding up the reading of QuakeML files to ObsPy Events?

If any suggestions are made I'm happy to undertake the work to do it.

megies:
> Is there a faster IO alternative already implemented?

Not to my knowledge.



> It looks to me like the main slow-down is the loop over events in obspy.io.quakeml.core.Unpickler._deserialize, I wonder if this can be a threaded operation?

Sounds like a pretty good idea. The FDSN `client.py` has threading for the individual endpoints and might be a good starting point for this.

calum-chamberlain:
Thanks for that @megies, I will have a look there and crack on with trying to speed it up.

megies:
No idea if there's hidden pitfalls I don't see right now with multiple threads working on the xml object, but I guess it shouldn't be too much work to set it up and see what happens..

d-chambers:
Hey @calum-chamberlain,

If you haven't used it, I recommend checking out [snakeviz](https://jiffyclub.github.io/snakeviz/), it makes profiling things like this a delightful experience. 

I did something similar trying to speed up reading miniseed files (https://github.com/obspy/obspy/pull/1845) and was able to find some low hanging fruit very quickly. [This notebook](https://gist.github.com/d-chambers/1eabd9c2147b11db0e68cf57ac3a8ad1) is a simple example.

calum-chamberlain:
Thanks for those comments guys - I thought I could get around the GIL with more direct access to lxml (which I think releases the GIL?), but it would require some strange in-between stages that I'm not that keen on.

Thanks @d-chambers I haven't used snakeviz - will do and do something similar with a notebook and share results here!

calum-chamberlain:
[...]

PR #2091, opened 2018-03-20
Resource id refactor

d-chambers:
This PR represents a few days of work trying to track down a particularly hard to find bug with the "event scoped" object referrals of `ResourceIdentifier` class. The issues were related to premature deletions of the `__resource_id_weak_dict` due to bugs in the resource_id's manual reference counting. While debugging the code I ended up performing some major refactoring and adding some additional features, namely:

1. Moved the `ResourceIdentifier` class from `obspy.core.event.base` to its own module `obspy.core.event.resourceid` as the class definition was getting quite large. 

2. Added a context manager for safely (and temporarily) replacing the `ResourceIdentifier` class state
called `_debug_class_state` (a private method for debugging only). This allowed the removal of
all manual clearing of the `__resource_id_weak_dict` (which is explicitly warned against doing in the docs). 

3. Removed the manual id reference counting by `ResourceIdentifier ` and replace it instead with a semi-singleton class (only one instance ever exists per id at one time) called  `_ResourceSingleton`. The resource singleton reference are kept in a weak value dict, and are used as keys to the resource_id_weak dict. This results in the same behavior as before, but pushes the burden of reference counting back to the standard library/python implementation.

4. Added several more tests for the event-scoped behavior of the `ResourceIdentifier`. 

5. Added the `MegaCatalog` class to the testing module of utils for creating a Catalog object with most, if not all, the classes contained in obspy.event.

6. Added a helper function for allowing context managers to be used in the 'setUp` method of  `unittest.TestCase` instances to the testing module.  

### PR Checklist
- [X] Correct base branch selected? `master` for new fetures, `maintenance_...` for bug fixes
- [X] All tests still pass.


obspy/core/event/event.py
```diff
@@ -292,6 +292,11 @@ def __deepcopy__(self, memodict=None):
         result.resource_id.bind_resource_ids()  # bind all resource_ids
         return result
 
+    @rlock
+    def __setstate__(self, state_dict):
+        super(Event, self).__setstate__(state_dict)
+        ResourceIdentifier.bind_resource_ids()
```
krischer: Just writing down my thoughts: There are now two separate thread locks for the Event objects but I think this if fine as `__setstate__()` and `__deepcopy__()` cannot really be called at the same time so there are no race conditions.

krischer: Should `bind_resource_ids()` also be called for all the other event objects like `Origin`, `Magnitude`, ...?

obspy/core/tests/test_resource_identifier.py
```diff
@@ -0,0 +1,569 @@
...
+        ResourceIdentifier class level and binds the new dicts to self with
+        keys 'r_dict' and 'unbound'.
+        """
+        # setup temporary id dict for tests in this case and bind to self
+        context = ResourceIdentifier._debug_class_state()
+        state = setup_context_testcase(self, context)
+        self.r_dict = state['rdict']  # the weak resource dict
+        self.unbound = state['unbound']  # the ubound resource ids
+
+    def print_state(self):
```
krischer: Can you add a nocover pragma here, e.g.:

```python
def print_state(self):  # pragma: nocover
    ...

File: obspy/core/tests/test_resource_identifier.py [hunk 3]
```diff
@@ -0,0 +1,569 @@
...
+    def setUp(self):
+        """
+        Setup code to run before each test. Temporary replaces the
+        resource_id_weak_dict and unbound_resource_id on the
+        ResourceIdentifier class level and binds the new dicts to self with
+        keys 'r_dict' and 'unbound'.
+        """
+        # setup temporary id dict for tests in this case and bind to self
+        context = ResourceIdentifier._debug_class_state()
+        state = setup_context_testcase(self, context)
```
krischer: Just for my understanding: `__exit__()` of the context manager would be call once the instance is garbage collected?

obspy/core/tests/test_resource_identifier.py
```diff
@@ -0,0 +1,569 @@
...
+        object_a = UTCDateTime()
+        ref_count = sys.getrefcount(object_a)
+        _res_id = ResourceIdentifier(referred_object=object_a)
+        self.assertEqual(sys.getrefcount(object_a), ref_count)
+        self.assertTrue(bool(_res_id))
+
+    def test_id_without_reference_not_in_global_list(self):
+        """
+        This tests some internal workings of the ResourceIdentifier class.
+        NEVER modify the __resource_id_weak_dict!
```
krischer: This sentence is obsolete now.

obspy/core/tests/test_resource_identifier.py
```diff
@@ -0,0 +1,569 @@
...
+            e.exception.args[0],
+            "The id 'http://example.org' is not a valid QuakeML resource "
+            "identifier. ObsPy tried modifying it to "
+            "'smi:local/http://example.org' but it is still not valid. Please "
+            "make sure all resource ids are either valid or can be made valid "
+            "by prefixing them with 'smi:<authority_id>/'. Valid ids are "
+            "specified in the QuakeML manual section 3.1 and in particular "
+            "exclude colons for the final part.")
+
+    def test_dubug_class_state(self):
```
krischer: typo

obspy/core/tests/test_resource_identifier.py
d-chambers: I am using Ned Bachelder's recipe for using [context managers with unit test](https://nedbatchelder.com/blog/201508/using_context_managers_in_test_setup.html), which calls the setup and cleanup function before and after every test. So the context manager enters and exits for every test thanks to [this helper function](https://github.com/obspy/obspy/blob/ae3c7036dc5fec2d3102124177df5d181edbb8ad/obspy/core/util/testing.py#L1040). 

Pytest has an elegant solution for kind of thing with yield fixtures, but this was the best I could find for unittest.

obspy/core/event/event.py
d-chambers: Tbh I am not really sure if we need an rlock on `__setstate__` or not. When I am feeling really ambitious I should setup a nasty multi-threaded test case to make sure it all still works with many threads.

d-chambers:
This is still a work in progress, I may would like to think about it a bit more before it gets merged.

krischer:
Just a quick ping on this. No need to hurry but it would be nice if this does not get lots in a stable PR.

d-chambers:
Hey @krischer,
I am still working on this. I have a solution that is robust and thread-safe but I need a bit longer to clean it up and to look at ways of speeding it up (reading/copying catalogs currently take ~15% performance hit).

d-chambers:
I am struggling a bit with this one. I ended up abandoning the implementation form #1644 that depended on event creation order because it was causing all sorts of hard to debug problems, and multi-threading didn't work despite rlocking key functions.  IMO it was not a good design. 

To replace it, I implemented a method of scoping resource ids based on using a parent id (eg the id of the event object) as a namespace of sorts for resource_ids and their referred objects. This works by calling a recursive function to find all resource ids in an objects attribute tree and assigning them to the parent namespace. This approach is thread-safe and seems to be much more reliable. 

However, recursion is not particularly efficient in python. This new method slows catalog creation operations (read, copy, unpickle) down by around 20%. This will hamper other efforts to improve efficiency (eg #1910).

Perhaps it might be better to leave the resource_ids un-scoped (ie they will behave how they originally did before #1644) but provide a public method on the catalog and event object in order to scope the resource ids? I imagine many catalog/event interactions don't even make use of the resource_ids' so it would just be a waste of computation in those cases, but it would allow the user the option to ensure correct resource_id behavior when it is needed. It could be rolled up into a general function called something like "validate" that will also try to detect/fix other issues that might occur in event objects.  

I will push the work I have so far, but it still will require a bit of cleanup. If anyone has any suggestions they would be much appreciated.

d-chambers:
I spent some time revisiting this and optimizing and I am pleased with the results. Doing the event scoping of the resource_ids still has a cost, but now it is not all that much. From my profiling on python 3.6 and Ubuntu 16.04 I get the following (no event_scoping, event_scoping):

read quakeml: 13.3 / 13.6 ms
copy catalog: 4.8 / 5.07 ms
unpickle 1.71 / 2.21 ms

[Here](https://gist.github.com/d-chambers/03088018405039a4acd81a1f995a256b) is the ipython notebook I used to do the profiling. In order to run in you need to be on this branch with your current obspy installation. 

It is interesting that unpickling an event object takes less than half the time that copying does, but that is an issue for a different day.

obspy/core/tests/test_resource_identifier.py
```diff
@@ -0,0 +1,611 @@
...
+        self.assertEqual(len(r_dict), r_dict_len - 1)
+        # Now create two equal references.
+        r1 = ResourceIdentifier(rid, referred_object=t1)
+        r2 = ResourceIdentifier(rid, referred_object=t1)
+        self.assertEqual(r1.get_referred_object(), t1)
+        # Deleting one should not remove the reference.
+        del r1
+        self.assertEqual(r2.get_referred_object(), t1)
+        self.assertIn(r2._resource_key, r_dict)
+        # Deleting the second one should (r_dict should not be empty)
```
krischer: s/not/now

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+        The parent_id_tree will be scanned first to see if any object_ids
+        have been assigned to the same resource_id in the parent scope. If
+        not simply use the last object assigned the resource_id. If no object
+        is found return None.
+        """
+        # Warn if resource_id was bound but its object got gc'ed.
+        if self._object_id is not None:
+            msg = ("The object with identity of: %d no longer exists, "
+                   "returning the most recently created object with a"
+                   " resource id of: %s") % (self._object_id, self.id)
```
krischer: This warning message is not fully correct as it would also be triggered if the object is found in the parent scope.

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+        :param parent:
+            An object or int (id) to which the resource_id should be scoped.
+            Used, for example, to ensure the all resource ids belonging to an
+            event object are event-scoped.
+        :type parent: object, int
+        """
+        self._object_id = id(referred_object)
+        # Use parent-scoping feature if parent is not None.
+        if parent is not None:
+            self._parent_key = parent
```
krischer: What would happen if this is called multiple times with different parent objects?

obspy/core/event/event.py
```diff
@@ -311,6 +319,23 @@ def write(self, filename, format, **kwargs):
...
+        """
+        gen = _yield_resource_id_parent_attr(self)
+
+        for resource_id, parent, attr in gen:
+            if attr == 'resource_id':
+                resource_id.set_referred_object(parent, parent=self,
+                                                warn=False)
+            else:
+                resource_id._parent_key = self
+                resource_id._object_id = None
```
krischer: How are things like the derived origin ids scoped after this function has been called? `_parent_key` would be set but `_parent_id_tree` and things like this might not be set. But I just be misunderstanding some things.

obspy/core/util/misc.py
```diff
@@ -639,6 +638,111 @@ def buffered_load_entry_point(dist, group, name):
     return _ENTRY_POINT_CACHE[hash_str]
 
 
+def yield_obj_parent_attr(obj, cls=None, is_attr=None, has_attr=None):
```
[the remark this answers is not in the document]
d-chambers: If there are not any issues I would like to keep it; I find this function very useful for transversing Catalog and Inventory instances and am planning to use it in #2154. I can always re-implement it if that would be cleaner. It is used several places in the new tests anyhow.

obspy/core/util/misc.py
```diff
@@ -639,6 +638,111 @@ def buffered_load_entry_point(dist, group, name):
...
+    from obspy.core.event import ResourceIdentifier
+
+    ids = set()  # id cache to avoid circular references
+
+    def func(obj, attr=None, parent=None):
+
+        if obj is not None and (hasattr(obj, '__dict__') or
+                                isinstance(obj, (list, tuple))):
+            id_tuple = (id(obj), id(parent))
+            if id_tuple not in ids:
```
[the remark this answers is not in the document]
d-chambers: Good idea but this is a bit tricky; this line is not in the scope of a loop so the `continue` keyword isn't valid. Since this is a generator we can just return (which is the new way of raising StopIteration according to PEP 479). I think this will work in 2.7, let's see what the CI says.

obspy/core/util/misc.py
```diff
@@ -639,6 +638,111 @@ def buffered_load_entry_point(dist, group, name):
...
+    Specialized form of yield_obj_parent_attr for getting ResourceIdentifiers.
+
+    This function makes some assumptions because only resource_identifiers are
+    being sought in order to improve efficiency.
+    """
+    from obspy.core.event import ResourceIdentifier
+
+    ids = set()  # id cache to avoid circular references
+
+    def func(obj, attr=None, parent=None):
```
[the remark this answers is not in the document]
d-chambers: This function is only used in the `_yield_resource_id_parent_attr` scope, but I agree it would be more consistent to switch the order.

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+        self.name = name + '__'
+        self.default = default
+
+    def __get__(self, instance, owner):
+        return getattr(instance, self.name, self.default)
+
+    def __set__(self, instance, value):
+        # if an object was passed, use the id of the object for hash
+        if value is not None:
+            if not isinstance(value, (int, str, native_str)):
```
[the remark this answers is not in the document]
d-chambers: Yes, some of the objects used as parents are (rightfully) not hashable. For example:

```python
import obspy.core.event as ev
hash(ev.Event())  # raises TypeError
```
 This little hack gets around it. In this case, we are actually just interested in ensuring the key objects are unique (in memory) so the mutability considerations that normally should guide if an object is hashable are not really applicable.

You bring up a good point about possible collisions, although I think it would have to be deliberate or just a very very bad day. I guess if we are concerned about it I could also check if `value` is an int and cast it to a str or float so that the only way an int gets into the hash table is through this if clause? It may incur a slight hit to performance though.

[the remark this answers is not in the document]
d-chambers: Also, looking at the flow of the program, I would be surprised if an int ever makes it here, but it wouldn't be impossible.

obspy/core/tests/test_resource_identifier.py
d-chambers: Sorry, I am not sure what you mean.

obspy/core/event/resourceid.py
d-chambers: I am trying to remember (I had this really well thought out at one point) but I think this does happen when an event is copied. I don't think this will cause issues since the resource_id instance will now point to the correct object in the new scope.

obspy/core/event/event.py
d-chambers: Any object that is referenced by a resource_id should have the attribute `resource_id`. Lines 332-334 take care of this case, which would involve logging the identities of all the objects with `resource_id` attributes into the proper place in the `_parent_id_tree`. Then, resource ids that refer to other objects in the same event (but not their parent) such as the `pick_id` of the `Arrival` object will find the correct object when `get_referred_object` is called because it was previously put into `_parent_id_tree`. Because `_object_id` on has been set to `None` it will not raise a warning. I added another test to specifically check for warnings being raised in this case.

obspy/core/event/resourceid.py
d-chambers: fixed

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+    >>> # The reference does not changed the reference count of the object.
+    >>> print(ref_count == sys.getrefcount(event))
+    True
+    >>> # It actually is the same object.
+    >>> print(event is res_id.get_referred_object())
+    True
+    >>> # Deleting it, or letting the garbage collector handle the object will
+    >>> # invalidate the reference.
+    >>> del event
+    >>> print(res_id.get_referred_object())  # doctest: +SKIP
```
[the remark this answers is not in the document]
d-chambers: unskipped

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+    >>> obj_a = UTCDateTime(10)
+    >>> obj_b = UTCDateTime(10)
+    >>> ref_a = ResourceIdentifier(res_id, referred_object=obj_a)
+    >>> ref_b = ResourceIdentifier(res_id, referred_object=obj_b)
+    >>> assert ref_a.get_referred_object() == ref_b.get_referred_object()
+    >>> assert ref_a.get_referred_object() is not ref_b.get_referred_object()
+    >>> assert ref_a.get_referred_object() is obj_a
+    >>> assert ref_b.get_referred_object() is obj_b
+    >>> del obj_b  # obj_b gets garbage collected
+    >>> assert ref_b.get_referred_object() is obj_a  # doctest: +SKIP
```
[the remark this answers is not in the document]
d-chambers: unskipped

obspy/core/event/resourceid.py
```diff
@@ -0,0 +1,653 @@
...
+    >>> del obj_a  # now no object with res_id exists
+    >>> assert ref_b.get_referred_object() is None  # doctest: +SKIP
```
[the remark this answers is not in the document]
d-chambers: unskipped

d-chambers:
No problem, thanks for taking a look at it. I really hope this finally fixes all the issues (it seems to have worked so far).

>>> The only downside I see is that this is really hard to understand.

I agree. It is more complicated than I would like, but I couldn't come up with something simpler that still solved all the edge cases. The basic gist, which I am sure you know, is to have each ID scoped to some parent, in this case the event object. The only way I found to reliably do this is wait until an event and all related attributes are created, then recurse and set each resource_id in the event to the correct parent.

obspy/core/util/misc.py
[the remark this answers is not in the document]
krischer: That argument is good enough for me. I'll prefix is with an underscore to mark it as private just so we don't commit to maintain it and are free to change its API.

obspy/core/event/resourceid.py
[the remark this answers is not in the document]
krischer: > This little hack gets around it. In this case, we are actually just interested in ensuring the key objects are unique (in memory) so the mutability considerations that normally should guide if an object is hashable are not really applicable.

I was more thinking to just always hash `id(value)`. I agree that we don't really have to worry about this minuscule chance and it could still happen with any hash.

obspy/core/util/misc.py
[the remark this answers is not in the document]
krischer: Ah yea sorry you are right. CI seems to like in on 2.7 (which I think works only when one does not return a value).

obspy/core/tests/test_resource_identifier.py
krischer: Sorry I should have been more clear. `s/SEARCH/REPLACE` is `ack` syntax to search and replace strings. It should be `r_dict should now be empty`.

obspy/core/event/event.py
krischer: :+1: Thanks for the explanation. After spending some more time I think I now fully understand your changes and how everything works. Nice job!

krischer:
I cleaned up a bit more and fixed two things that got lost during my rebase. Double checking could not hurt but I think I found everything missing.

Also all event plugin's now scope the resource identifiers just to make sure everything works.

I still have trouble understanding the whole `_ResourceKey` + descriptor business. So I see that the `_ResourceKey` object is a necessary wrapper object so it can be used in the `weakref` dictionaries. But why does it have to be a singleton? Its only used as keys for the dictionary so why does a deterministic hash like `hash(id(value))` not work? Also where do the object live? They don't seem to get garbage collected as I would expect, see e.g.

```python
In [1]: import obspy

In [2]: cat = obspy.read_events()

In [3]: cat
Out[3]:
3 Event(s) in Catalog:
2012-04-04T14:21:42.300000Z | +41.818,  +79.689 | 4.4 mb | manual
2012-04-04T14:18:37.000000Z | +39.342,  +41.044 | 4.3 ML | manual
2012-04-04T14:08:46.000000Z | +38.017,  +37.736 | 3.0 ML | manual

In [4]: dict(cat[0].resource_id._parent_key._singleton_cache)
Out[4]:
{112059039136: <obspy.core.event.resourceid._ResourceKey at 0x1a173d6908>,
 112059050304: <obspy.core.event.resourceid._ResourceKey at 0x1a173f52e8>,
 112059175544: <obspy.core.event.resourceid._ResourceKey at 0x1a173f5748>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261017/782476/796628': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6ba8>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261018/782478/796631': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5438>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261020/782484/796646': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7b00>,
 'quakeml:eu.emsc/event/20120404_0000038': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5320>,
 'quakeml:eu.emsc/event/20120404_0000039': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5b00>,
 'quakeml:eu.emsc/event/20120404_0000041': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6240>,
 'quakeml:eu.emsc/origin/rts/261017/782476': <obspy.core.event.resourceid._ResourceKey at 0x1a173d66a0>,
 'quakeml:eu.emsc/origin/rts/261018/782478': <obspy.core.event.resourceid._ResourceKey at 0x1a173f53c8>,
 'quakeml:eu.emsc/origin/rts/261020/782484': <obspy.core.event.resourceid._ResourceKey at 0x1a173d3b00>,
 'smi://eu.emsc/unid': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5b38>,
 'smi:eu.emsc-csem/origin_method/NA': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7da0>,
 'smi:smi-registry/organization/DDA': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6eb8>,
 'smi:smi-registry/organization/EMSC': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7978>,
 'smi:smi-registry/organization/NNC': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6a58>}

In [5]: del cat[0]; del cat[1]

In [6]: cat
Out[6]:
1 Event(s) in Catalog:
2012-04-04T14:18:37.000000Z | +39.342,  +41.044 | 4.3 ML | manual

In [7]: import gc; gc.collect()
Out[7]: 249

In [8]: dict(cat[0].resource_id._parent_key._singleton_cache)
Out[8]:
{112059039136: <obspy.core.event.resourceid._ResourceKey at 0x1a173d6908>,
 112059050304: <obspy.core.event.resourceid._ResourceKey at 0x1a173f52e8>,
 112059175544: <obspy.core.event.resourceid._ResourceKey at 0x1a173f5748>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261017/782476/796628': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6ba8>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261018/782478/796631': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5438>,
 'quakeml:eu.emsc/NetworkMagnitude/rts/261020/782484/796646': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7b00>,
 'quakeml:eu.emsc/event/20120404_0000038': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5320>,
 'quakeml:eu.emsc/event/20120404_0000039': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5b00>,
 'quakeml:eu.emsc/event/20120404_0000041': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6240>,
 'quakeml:eu.emsc/origin/rts/261017/782476': <obspy.core.event.resourceid._ResourceKey at 0x1a173d66a0>,
 'quakeml:eu.emsc/origin/rts/261018/782478': <obspy.core.event.resourceid._ResourceKey at 0x1a173f53c8>,
 'quakeml:eu.emsc/origin/rts/261020/782484': <obspy.core.event.resourceid._ResourceKey at 0x1a173d3b00>,
 'smi://eu.emsc/unid': <obspy.core.event.resourceid._ResourceKey at 0x1a173f5b38>,
 'smi:eu.emsc-csem/origin_method/NA': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7da0>,
 'smi:smi-registry/organization/DDA': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6eb8>,
 'smi:smi-registry/organization/EMSC': <obspy.core.event.resourceid._ResourceKey at 0x1a173c7978>,
 'smi:smi-registry/organization/NNC': <obspy.core.event.resourceid._ResourceKey at 0x1a173d6a58>}
```

d-chambers:
**Why does the _ResouceKey have to be a singleton? Why not use hash(id(value))?**

1. The `_ResourceKey` is a trick to enable the python gc to handle some of the work the resource_id was doing (unreliably) on its own  before in the `__del__` method, so we need to use the `_ResourceKey` instances in the wekref dicts.  The  `_ResourceKey.get_resource_key` method handles two distinct cases:

a. An object that is not an int/str is passed. The `_ResourceKey` instances needs to be unique to this object only (using its identity) and return a different `_ResourceKey` instance if a similar (even equal) object is passed to `_ResourceKey.get_resource_key`. This enables the parent scoping of equal yet different `Event` objects for example. 

b. An object that is an int/str is passed (note `ResourceIdentifier.id` will always be a str or a type error is raised by the setter). The same `_ResourceKey` instance should be used with equal int/str but that have different identities. This enables different `ResourceIdentifier` instances with the same id to refer the same object. The strings in the following example should return the same instance of '_ResourceKey' if passed to `_ResourceKey.get_resource_key`. If we used the id in the hash this would not work. 
 
```python
a = "'lion roars at stale pull requests'"
b = "'lion roars at stale pull requests'"
assert a == b
assert id(a) != id(b)
```

2. I suppose `_ResourceKey` could be a non-singleton as long as it implemented a hash based on the correct input (the identity of an object or the str/int) but this could greatly increase the number of `_ResourceKey` instances that python has to create/manage in the weakvalue dict, as well as require an addition setattr. All of this could result in a performance/memory usage hit. In this case I figured it is better to do more with less.

I should note too, I don't think singleton is exactly the right term (but I don't know a better one) since a true singleton should only have one instance, not just one instance per unique input. 

**Where do the `_ResourceKey` objects live?**

The only strong reference to these should be attached to the `ResourceIdentifier` instances as the `_parent_key` or `_resource_key` attributes. In your example I think there is some caching going on in the `read_events` call as the following simpler case does work as expected:

```python
import gc

import obspy.core.event as ev

event = ev.Event()
print(dict(ev.ResourceIdentifier._parent_id_tree))  # dict with one entry

del event
gc.collect()

print(dict(ev.ResourceIdentifier._parent_id_tree))  # empty dict
```
In fact, if we count the references there seem to be many in the default catalog that are being held onto by something (which is why the `del` isn't working as you expected):

```python
import gc
import sys

import obspy
import obspy.core.event as ev

def print_ref_counts(obj):
    rc = sys.getrefcount(obj)
    # subtract 3 due to temporary reference created in this function/getrefcount
    print(f'obj has {rc -3} refcounts')

event = ev.Event()
print_ref_counts(event.resource_id._parent_key)  # prints "obj has 1 refcounts"

cat = obspy.read_events()
print_ref_counts(cat[0].resource_id._parent_key)  # prints "obj has 13 refcounts"
```

**What is the deal with the descriptor?**

Since there is some logic required to figure out the key for caching the `_ResourceKey` instances (selecting the object identity or the str/int) the descriptor just lets me put that logic in one place rather than every time I set a value to the `_parent_key` or `_resource_key`. Of course it could be implemented as a getter/setter (which I think is the same thing in the end) but since there are two of them I figured it would be better to have an external descriptor to avoid writing the same code twice.

d-chambers:
Another thing to perhaps do in this PR, maybe in a new one, is to rework `convert_id_to_quakeml_uri` to return a copy of the resource_id with the new `id` rather than modifying it in place. Since the resource_ids are hashed on the `id` attr we should treat it as immutable.

krischer:
Thanks a bunch for the explanations! I especially did not get the distinction between 1 a) and 1 b) but now everything is clear.

> All of this could result in a performance/memory usage hit. In this case I figured it is better to do more with less.

I actually tested this and it is indeed a small bit but measurably (~3%) slower when running `test_events.py`.

> Of course it could be implemented as a getter/setter (which I think is the same thing in the end) but since there are two of them I figured it would be better to have an external descriptor to avoid writing the same code twice.

Yea that's the same thing. Python internally implements properties as descriptors. I do like the descriptor approach as it (as you say) forces a single implementation.


> Another thing to perhaps do in this PR, maybe in a new one, is to rework convert_id_to_quakeml_uri to return a copy of the resource_id with the new id rather than modifying it in place. Since the resource_ids are hashed on the id attr we should treat it as immutable.

:+1: for dealing with in a separate PR. We have to be a bit careful with this and maybe raise a few new warnings. Some people might use `ResourceIdentifiers` as dictionary keys and then write out a QuakeML file which (in your proposal) would create new objects and the dictionary keys might thus become meaningless. But lets discuss it elsewhere.

Regarding the references: I investigated a bit more (`gc.get_referrers()` is actually pretty helpful) and it turns out that all the spurious references are coming from `ipython` and especially all its magic like the `jedi` code-completion. Just running it in the normal interpreter completely fixes all my issues and the high refcount in your example comes internal references (mostly `_parent_key`s). So all is well.

```python
import gc
import obspy
from obspy.core.event.resourceid import _ResourceKey

cat = obspy.read_events()
print(len(_ResourceKey._singleton_cache))

del cat[0]
gc.collect()
print(len(_ResourceKey._singleton_cache))

del cat
gc.collect()
print(len(_ResourceKey._singleton_cache))
```

Output:
```
17
12
0
```

@d-chambers If you have no objections I would merge this now.

d-chambers:
Actually hold off merging for now. I think I found another (small) issue.

d-chambers:
Ok, I made a slight change to the logic that governs when a warning is issued and added a test for it. If you approve of the change go ahead and merge. 

Thanks again @krischer for all the time you spent reviewing this!

krischer:
:tada:

Thanks @d-chambers!

Issue #2090, opened 2018-03-19
Invalid resource identifier for QuakeML produced by SC3


claudiodsf:
This is a follow-up of the [discussion](https://lists.swapbytes.de/archives/obspy-users/2018-March/002702.html) in the mailing list, which I report here:

The attached QuakeML file ([ipgp2016afqtcu.xml.gz](https://github.com/obspy/obspy/files/1827019/ipgp2016afqtcu.xml.gz)) has been produced by SeisComP3.

It is correctly read by `read_events()`, but when trying to write it back to disk, I get the following error:

```python
ValueError: The id 'smi:org.gfz-potsdam.de/geofon/RMHP(60)>>ITAPER(3)>>BW(4,5,15)' is not a valid QuakeML resource identifier.
```

This is related to the presence of the sign `>` which is encoded in the original xml file as `&gt;` but gets decoded as `>`.

This symbol is part of the SeisComP3 filter grammar: https://www.seiscomp3.org/doc/jakarta/2015.040/base/filter-grammar.html

I think that, as long as this symbol is not decoded from its HTML representation, it should be acceptable. 

So my impression is that this is an ObsPy bug.

claudiodsf:
The same file is not valid according to `jing`, but due to a different error:

```
ipgp2016afqtcu.xml:61:127: error: value of attribute "publicID" is invalid; must be a URI
```

That's because an URI cannot have more than one `#` sign.

By manually fixing this problem (see [ipgp2016afqtcu_fix.xml.gz](https://github.com/obspy/obspy/files/1827033/ipgp2016afqtcu_fix.xml.gz)), the resulting file is valid XML, according to `jing`, even with those `&gt;` in it.

claudiodsf:
This has been discussed in the mailing list with @krischer, but I'm not sure whether it is an ObsPy bug or a SeisComP3 bug, so I call here @andres-h and @gempa-jabe as well.
Thanks!

megies:
See also #1872

krischer:
@andres-h @gempa-jabe pinging you again so we can discuss this. Its getting a bit lengthy on the mailing list.

[...]

PR #2093, opened 2018-03-22
Allow "&", ">" and "<" in ResourceID

claudiodsf:
This PR is a possible fix for #2090.

The characters `&`, `>` and `<` can IMHO be accepted for Resource ID, since they are HTML-escaped when serializing to XML, i.e. they become on writing `&amp;`, `&gt;` and `&lt;`, respectively.

I validated a QuakeML file resulting from this PR using `jing` and it is valid QuakeML.

Note that I changed `&amp;` (as reported on the QuakeML documentation) with simply `&`, since it has no sense in a regex string (matching `&amp;` means separately matching `&`,  `a`, `m`, `p` and `;`).

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -779,8 +779,10 @@ class ResourceIdentifier(object):
         if str(id).strip() == "":
             id = str(uuid4())
 
+        # Note: the characters "&", ">" and "<" are allowed here since
+        # they are HTML-escaped when serializing to XML
         regex = r"^(smi|quakeml):[\w\d][\w\d\-\.\*\(\)_~']{2,}/[\w\d\-\." + \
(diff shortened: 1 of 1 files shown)


krischer:
I'm personally fine with treating things a bit laxer in ObsPy so I don't have a big problem with including this. But we should first clarify if this is actually valid according to the QuakeML spec - I still think its not as any validation IMHO has to happen on the decoded/unescaped XML files.

Seiscomp appears to be the only package currently producing ids like this and this message on the mailing list seems to indicate that they changed things: https://lists.swapbytes.de/archives/obspy-users/2018-March/002712.html

claudiodsf:
@krischer, based on the discussion in #2090, shall we go ahead with this PR?
Do you want me to add a warning in case `<` or `>` appear in a ResourceID?

claudiodsf:
Pinging again @krischer. What should we do here?

krischer:
Dealt with (in same manner at least) in #2104.

Issue #1872, opened 2017-09-01
QuakeML verification problem (related to SC3ML converted using stylesheets)


megies:
I think there is a problem with the QuakeML RelaxNG schema (upstream). It seems to fail validation on URIs that do not contradict the regular expression pattern specified in the QuakeML documentation.

Here's a minimal example QuakeML:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<q:quakeml xmlns:q="http://quakeml.org/xmlns/quakeml/1.2" xmlns="http://quakeml.org/xmlns/bed/1.2">
  <eventParameters publicID="smi:org.gfz-potsdam.de/geofon/NA">
    <event publicID="smi:org.gfz-potsdam.de/geofon/gof2017qznj">
      <stationMagnitude publicID="smi:org.gfz-potsdam.de/geofon/Origin#20170830154555.615079.1673910#staMag.MLv#GR.WET">
        <originID>smi:org.gfz-potsdam.de/geofon/Origin#20170830154555.615079.1673910</originID>
        <mag>
          <value>1.597395494</value>
        </mag>
        <type>MLv</type>
        <amplitudeID>smi:org.gfz-potsdam.de/geofon/Amplitude#20170830154537.481105.2138874</amplitudeID>
        <waveformID networkCode="GR" stationCode="WET" channelCode="HHZ"/>
        <creationInfo>
          <agencyID>GOF</agencyID>
          <author>scmag@kelut</author>
          <creationTime>2017-08-30T15:45:55.803111Z</creationTime>
        </creationInfo>
      </stationMagnitude>
    </event>
  </eventParameters>
</q:quakeml>
```

Validation against the RelaxNG schema fails:
```
$ jing ~/git/obspy/obspy/io/quakeml/data/QuakeML-1.2.rng quakeml.xml
/tmp/xslt/quakeml.xml:5:121: error: value of attribute "publicID" is invalid; must be a URI
```

The problem are the additional `#` characters in the local identifier part (which itself is denoted by the first `#` character). If these are removed, then the file validates just fine. Also, if the RelaxNG schema is changed from type `anyURI` to `string` for the `ResourceIdentifier` type (that includes the regex pattern for verification anyway), than it also verifies OK. So the problem seems that the `anyURI` is adding additional constraints and disallows multiple `#` characters.

[...]

PR #2104, opened 2018-04-10
Allow invalid ids when writing QuakeML files

### LINKED ISSUE #2090: Invalid resource identifier for QuakeML produced by SC3

krischer:
Fixes #2090, #2093, and also #1872 I think. It least it fixes the problem from ObsPy's side.

This PR allows writing invalid QuakeML resource ids and can thus round-trip these ids. The rational for this is that external systems might depend on these ids so we don't want to mess with them. It does raise a warning though. I think that is the best way to handle this issue.

Before this PR ObsPy created a random (but valid) resource id for each element's `publicID` attribute. For all cross-referential ids it failed (see e.g. #2090). Now it will always work, but raise a warning.

CC: @claudiodsf @megies @gempa-jabe @andres-h @Brtle

### DIFF
--- a/obspy/core/event/base.py
+++ b/obspy/core/event/base.py
@@ -780,7 +780,7 @@ class ResourceIdentifier(object):
             id = str(uuid4())
 
         regex = r"^(smi|quakeml):[\w\d][\w\d\-\.\*\(\)_~']{2,}/[\w\d\-\." + \
-                r"\*\(\)_~'][\w\d\-\.\*\(\)\+\?_~'=,;#/&amp;]*$"
+                r"\*\(\)_~'][\w\d\-\.\*\(\)\+\?_~'=,;#/&]*$"
         result = re.match(regex, str(id))
         if result is not None:
             return id
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,6 +12,9 @@ master:
  - obspy.io.reftek:
    * Implement reading reftek encodings '16' and '32' (uncompressed data,
      16/32bit integers, see #2058 and #2059)
+ - obspy.io.quakeml:
+   * Allow writing invalid ids but raise a warning
+     (see #2104, #2090, #2093, #1872).
  - obspy.io.shapefile:
    * Add possibility to add custom database columns when writing catalog
      objects to shapefile (see #2012)
--- a/obspy/io/pde/mchedr.py
+++ b/obspy/io/pde/mchedr.py
@@ -1013,7 +1013,8 @@ class Unpickler(object):
     def _deserialize(self):
         catalog = Catalog()
         res_id = '/'.join((res_id_prefix,
-                           self.filename.replace(':', '/'))).replace("//", "/")
+                           self.filename.replace(':', '/')))\
+            .replace('\\', '/').replace('//', '/')
         catalog.resource_id = ResourceIdentifier(id=res_id)
         catalog.description = 'Created from NEIC PDE mchedr format'
         catalog.comments = ''
--- a/obspy/io/pde/mchedr.py
+++ b/obspy/io/pde/mchedr.py
@@ -1012,7 +1012,8 @@ class Unpickler(object):
 
     def _deserialize(self):
         catalog = Catalog()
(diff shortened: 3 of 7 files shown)


krischer:
> The problem I see here is that invalid_id.xml is actually valid QuakeML, since > gets escaped to gt;.

The conclusion in #2090 to me seems to be that it is not really valid (even thought `jing` claims it is). Any validation has to happen at the decoded level; XML validation is independent of the used text encoding so from my point of view it should also be independent of other forms of encoding (HTML escapes are another encoding). They are only escaped because they can easily mess with the XML but `>` is what is actually in the data and what thus should be validated. This is even more true if one considers QuakeML as a data model and the XML serializations as just one possibility.

In any case - this PR is to some extent not related to this particular discussion as I do think we want to allow writing invalid URIs for the rational laid out in the initial message of this PR.

claudiodsf:
Ok.

My other comment is that the `amp;` part of `&amp;` in the [regex](https://github.com/obspy/obspy/blob/master/obspy/core/event/base.py#L783) is not necessary (see my comments in #2093).

krischer:
> My other comment is that the amp; part of &amp; in the regex is not necessary (see my comments in #2093).

Good point. Done in c176dde.

megies:
> My other comment is that the amp; part of &amp; in the regex is not necessary (see my comments in #2093).

I'm pretty sure that regex was just copy/pasted from the QuakeML reference.

megies:
> external systems might depend on these ids so we don't want to mess with them. It does raise a warning though. I think that is the best way to handle this issue.

Looks like a minimally invasive and feasible solution to many problems. :+1:

krischer:
> I'm pretty sure that regex was just copy/pasted from the QuakeML reference.

Yes. It is needed for the XSD schema as `&amp;` is the character reference to `&`. But we specify the regex for Python and thus don't need to escape it.

megies:
Not sure where this is supposed to go.. many of the underlying issues are filed against maintenance branch and milestone 1.1.x but this one is against master and thus 1.2.0

krischer:
> Not sure where this is supposed to go.. many of the underlying issues are filed against maintenance branch and milestone 1.1.x but this one is against master and thus 1.2.0

I chose the `master` branch as this has the potential to produce different output files. Before this PR ObsPy replaced invalid `publicID` attributes with randomized ids when writing QuakeML files.

obspy/core/event/base.py
```diff
@@ -780,7 +780,7 @@ def get_quakeml_uri(self, authority_id="local"):
             id = str(uuid4())
 
         regex = r"^(smi|quakeml):[\w\d][\w\d\-\.\*\(\)_~']{2,}/[\w\d\-\." + \
-                r"\*\(\)_~'][\w\d\-\.\*\(\)\+\?_~'=,;#/&amp;]*$"
+                r"\*\(\)_~'][\w\d\-\.\*\(\)\+\?_~'=,;#/&]*$"
```
claudiodsf: Should be corrected also in the docstring (line 441)

krischer: Done directly in master: 4039c69973106b340477ab23495be4e8270a3f84

PR #2106, opened 2018-04-11
Skip invalid enums when reading QuakeML files

### LINKED ISSUE #2095: can't read Double Difference catalog
I got errors trying to read QuakeML file generated by web-request

cat_dd = obspy.core.event.read_events('./foo.xml')

Single origin is generated by the following web request:
http://service.ncedc.org/fdsnws/event/1/query?start=2018-01-04T00:00:00.0&end=2018-01-05T00:00:00.0&minmag=3&catalog=DD&includeallmagnitudes=false&includearrivals=false&includemechanisms=false&orderby=time&format=xml

Reports error:
ValueError: Setting attribute "origin_type" failed. Value "Double Difference" could not be converted to type "Enum(["hypocenter", "centroid", "amplitude", "macroseismic", "rupture start", "rupture end"])"

This breaks some code that used to work

Note: first time bug report

### LINKED ISSUE #2098: EvaluationMode in ISC Bulletin
Hi, when downloading events (including arrivals) from the ISC Bulletin using read_events with Client("ISC"), I got the following error:

> ValueError: Setting attribute "evaluation_mode" failed. Value "a" could not be converted to type "Enum(["manual", "automatic"])".

It turns out that the convention for some contributors to the ISC is to use "a" and "m" instead of "automatic"/"manual" for their picks and amplitudes. The Reviewed ISC Bulletin doesn't seem to homogenise this either.

I got around this by adding "a" and "m" to the accepted enumeration types for EvaluationMode in event/header.py, but I'm not sure if this is the most sustainable solution for the future - hence no PR.

Is there a way to make the switch to the full EvaluationMode convention upon encountering the single-letter version when reading the XML file?

krischer:
This PR enables ObsPy to read slightly invalid QuakeML files. Any time it encounters an invalid enum value it will raise a warning and not read it. But it continues reading the rest of the files.

I'm basing this on top of `master` as it might have some unintended consequences when reading QuakeML files.

Fixes #2098 
Fixes #2095

### DIFF
--- a/obspy/io/quakeml/core.py
+++ b/obspy/io/quakeml/core.py
@@ -189,7 +189,8 @@ class Unpickler(object):
         try:
             setattr(obj, key, value)
         except ValueError as e:
-            msg = "%s. Will be ignored." % e.args[0]
+            msg = ('%s. The attribute "%s" will not be set and will be missing'
+                   ' in the resulting object.' % (e.args[0], key))
             warnings.warn(msg)
 
     def _xpath(self, xpath, element=None, namespace=None):
--- a/obspy/io/quakeml/tests/test_quakeml.py
+++ b/obspy/io/quakeml/tests/test_quakeml.py
@@ -1101,8 +1101,9 @@ class QuakeMLTestCase(unittest.TestCase):
             w[0].message.args[0],
             'Setting attribute "depth_type" failed. Value "randomized" could '
             'not be converted to type "Enum(["from location", "from moment '
-            'tensor inversion", ..., "operator assigned", "other"])". Will be '
-            'ignored.')
+            'tensor inversion", ..., "operator assigned", "other"])". The '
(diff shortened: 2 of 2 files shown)


obspy/io/quakeml/core.py
```diff
@@ -181,6 +181,17 @@ def _xpath2obj(self, xpath, element=None, convert_to=str, namespace=None):
...
 
+    def _set_enum(self, xpath, element, obj, key):
+        obj_type = obj._property_dict[key]
+        if not isinstance(obj_type, Enum):  # pragma: no cover
+            raise ValueError
+        value = self._xpath2obj(xpath, element)
+        try:
+            setattr(obj, key, value)
+        except ValueError as e:
+            msg = "%s. Will be ignored." % e.args[0]
```
megies: Will be ignored -> 'The attribute %s will not be set and will be missing in the resulting object' % key ?

PR #2109, opened 2018-04-19
Support for SeismicHandler Evt files

trichter:
This PR adds SeismicHandler EVT read support.
It works already well for the 1000+ evt files I have.

+DOCS

TODO:

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -38,6 +38,8 @@ master:
       information.
     * Add explicit warnings regarding unsupported sections of Nordic files.
     * Fix mapping of magnitude-types between MS to S and Ms to s.
+ - obspy.io.sh:
+    * Add read support for SeismicHandler EVT event files
 
 
 1.1.x:
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -39,7 +39,7 @@ master:
     * Add explicit warnings regarding unsupported sections of Nordic files.
     * Fix mapping of magnitude-types between MS to S and Ms to s.
  - obspy.io.sh:
-    * Add read support for SeismicHandler EVT event files
+    * Add read support for SeismicHandler EVT event files (see #2109)
 
 
 1.1.x:
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -9,6 +9,11 @@ master:
      different precisions (see #2077).
    * Added replace method to UTCDateTime class (see #2077).
    * Added remove method to Inventory class (see #2088).
+ - obspy.io.nordic:
+   * Add ability to read and write focal mechanisms and moment tensor
+     information.
+   * Add explicit warnings regarding unsupported sections of Nordic files.
+   * Fix mapping of magnitude-types between MS to S and Ms to s.
  - obspy.io.reftek:
    * Implement reading reftek encodings '16' and '32' (uncompressed data,
      16/32bit integers, see #2058 and #2059)
@@ -17,6 +22,8 @@ master:
      (see #2104, #2090, #2093, #1872).
    * Skip invalid enumeration values during reading but raise a warning.
      (see #2106, #2098, #2095)
+ - obspy.io.sh:
+   * Add read support for SeismicHandler EVT event files (see #2109)
  - obspy.io.shapefile:
    * Add possibility to add custom database columns when writing catalog
      objects to shapefile (see #2012)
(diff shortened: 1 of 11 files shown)


megies:
Since you're working on this heavily and mostly only tests are missing, I'll mark it for 1.2.0

trichter:
DOC links

http://docs.obspy.org/pull-requests/2109/packages/autogen/obspy.io.sh.evt.html
http://docs.obspy.org/pull-requests/2109/packages/autogen/obspy.core.event.read_events.html

trichter:
This PR is ready for review.

obspy/io/sh/evt.py
```diff
@@ -153,23 +128,26 @@ def _mags(obj, evid, stamag=False):
...
     if len(mags) == 1:
         pm = mags[0].resource_id
     return mags, pm
 
 
 def _seed_id_map(inventory=None, id_map=None, id_default='.{}..{}'):
-    ret = {}
     if id_map is None:
         id_map = {}
+    ret = copy(id_map)
```
megies: Looks like `id_map` needs to be a dictionary if provided, so you could use `id_map.copy()` to get a shallow copy as well?

Or, maybe you rather even want to go for `deepcopy` instead? Just to be 100% sure you don't get unwanted side-effects (although it seems you're only ever expecting strings, i.e. non-modifiable types, in there)

CHANGELOG.txt
```diff
@@ -38,6 +38,8 @@ master:
       information.
     * Add explicit warnings regarding unsupported sections of Nordic files.
     * Fix mapping of magnitude-types between MS to S and Ms to s.
+ - obspy.io.sh:
+    * Add read support for SeismicHandler EVT event files
```
megies: Please add issue number

obspy/io/sh/core.py
```diff
@@ -27,6 +27,9 @@
 MONTHS = ['JAN', 'FEB', 'MAR', 'APR', 'MAY', 'JUN', 'JUL', 'AUG', 'SEP',
           'OCT', 'NOV', 'DEC']
 
+MONTHS_DE = ['JAN', 'FEB', 'MAR', 'APR', 'MAI', 'JUN', 'JUL', 'AUG', 'SEP',
+             'OKT', 'NOV', 'DEZ']
```
megies: I'd switch to use a dictionary, that could combine everything and the lookup done below wouldn't be as ugly with the `.index() + 1` call..

```python
MONTHS = {
    ...
    'DEC': 12,
    'DEZ': 12,
    }

Comment (trichter) [reply obspy/io/sh/core.py]: Nice idea, but I'd rather stick with the current state, because

1. it would need some extra logic to define the default for writing in `from_utcdatetime`
2. the German names are a fallback and the try except clause makes this clear

File: CHANGELOG.txt [hunk 4]
```diff
@@ -38,6 +38,8 @@ master:
       information.
     * Add explicit warnings regarding unsupported sections of Nordic files.
     * Fix mapping of magnitude-types between MS to S and Ms to s.
+ - obspy.io.sh:
+    * Add read support for SeismicHandler EVT event files (see #2109)
```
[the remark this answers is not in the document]
trichter: OK, done.

obspy/io/sh/evt.py
```diff
@@ -0,0 +1,282 @@
...
+    :return: An ObsPy Catalog object.
+
+    .. note::
+        The following fields are supported by this function: %s.
+
+        Compare with http://www.seismic-handler.org/wiki/ShmDocFileEvt
+    """
+    seed_map = _seed_id_map(inventory, id_map, id_default)
+    with open(filename, 'r') as f:
+        temp = f.read()
```
megies: As long as we do Python 2 and 3 it might be safer to read bytes and explicitly decode.. even for pure Python 3 it would be much safer to explicitly specify ASCII or UTF-8 whatever is appropriate? Otherwise we might run into strange issues on some systems (we've seen garbled default system encodings set many times in the past, esp. on Mac ;-P)

obspy/io/sh/evt.py
```diff
@@ -0,0 +1,282 @@
...
+                key, value = line.split(':', 1)
+            except ValueError:
+                continue
+            key = key.strip().lower()
+            value = value.strip()
+            if key == 'event id':
+                evid = value
+            elif value != '':
+                phase[key] = value
+    assert evid is None
```
megies: i don't understand this line, but if you think it's correct..

trichter:
I don't know if the docker tests are still working. Otherwise this is ready to merge from my point of view.

trichter:
I forgot adding waveform_id to the StationMagnitude. Now it's really ready if tests pass. Other fields can be added later if anyone needs them.

obspy/io/sh/evt.py
megies: :+1:

PR #2112, opened 2018-04-24
Fix Trace.times(type='matplotlib') being slow

ahhbjoe:
### What does this PR do?

The current version of obspy.core.trace.Trace.times is very slow for type="matplotlib". This PR speeds it up by ~1000 times.

I don't see relevant issues.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2497,7 +2497,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         elif type == "matplotlib":
             from matplotlib.dates import date2num
             time_array = (date2num(self.stats.starttime.datetime)
-                          + time_array / 86400)
+                          + time_array / 86400.0)
         else:
             msg = "Invalid `type`: {}".format(type)
             raise ValueError(msg)
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2496,11 +2496,8 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
                 [self.stats.starttime + t_ for t_ in time_array])
         elif type == "matplotlib":
             from matplotlib.dates import date2num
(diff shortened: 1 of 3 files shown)


obspy/core/trace.py
```diff
@@ -2496,8 +2496,11 @@ def times(self, type="relative", reftime=None):
                 [self.stats.starttime + t_ for t_ in time_array])
         elif type == "matplotlib":
             from matplotlib.dates import date2num
-            time_array = date2num([(self.stats.starttime + t_).datetime
-                                   for t_ in time_array])
+            SECONDSPERDAY = 86400
```
krischer: Having this as an extra variable is probably not necessary as most people should recognize that numbers and its only used once in the next name.

megies:
Nice first PR there!  :clap:

megies:
Dunno, maybe a short changelog entry might be good to have, but not thaaat important.

megies:
> Please install flake8 and pep8-naming locally and just run $ flake8 filename.py to make sure the formatting corresponds to our rules

Pro-tip: Use a pre-commit hook as a safe guard in the future.. :sunglasses: 
See http://docs.obspy.org/coding_style.html#obspy-coding-style-guide right at the top..

obspy/core/trace.py
ahhbjoe: Thanks for your suggestions. Requested changes have been made.

ahhbjoe:
It seems that CircleCI was checking an older version of my codes. The error messages are "variable 'SECONDSPERDAY' in function should be lowercase" and "trailing whitespace". But both problems have been fixed in the current version.

krischer:
Thanks! AppVeyor fails are unrelated.

megies:
Thanks for your PR, @ahhbjoe. Great job!  :tada:

PR #2118, opened 2018-04-25
try to fix appveyor by using pip<10

megies:
### What does this PR do?

Fix appveyor

### Why was it initiated?  Any relevant Issues?

Appveyor is failing to build our PRs currently

### PR Checklist
- [x] Correct base branch selected? `master` for new fetures, `maintenance_...` for bug fixes
- [ ] All tests still pass.

PR #2120, opened 2018-04-25
Merge Maintenance 1.1.x

calum-chamberlain:
As requested by @megies #2118 

Will create this, then fix the conflicts.


calum-chamberlain:
Looks like there are a few fails coming from travis regarding `obspy.clients.fdsn.tests.test_federator_routing_client.FederatorRoutingClientTestCase.test_get_stations_integration_test`, which I don't think is to do with this. @megies merge when you are happy.

megies:
> Merge branch 'master' into maintenance_1.1.x

That commit message created by git is strange, since it looks like indeed maintenance was merged into master correctly..

Anyway, looks good, so merging, thanks @calum-chamberlain!

PR #2128, opened 2018-05-01
Infer sampling rates for each response stage + Support for PAZ without decimation attributes

### LINKED ISSUE #2121: EVRESP ERROR when using tr.remove_response
Hello,

I get this error message using obspy 'tr.remove_response()' command:

EVRESP ERROR (... [File: <stdin>; Start date: ; Stage: 9]):
     check_channel; required decimation blockette for IIR or FIR filter missing,
     skipping to next response now
*** ValueError: check_channel: Illegal RESP format 

This happens only when I use the permanent stations from the DK network archived at GEOFON where I have also downloaded the StationXML file from.
The GEOFON team assume an obspy bug....

I am using Python 2.7.6 and obspy 1.1.0

I appreciate your help.

krischer:
This PR adds a new method `.get_sampling_rates()` to calculate the input and output sampling rates of each response stage. For most responses it will just parse the decimation attributes. For those with missing information it will infer the sampling rates from the surrounding stages.

```python
>>> import obspy
>>> inv = obspy.read_inventory("AU.MEEK.xml")
>>> inv[0][0][0].response.get_sampling_rates()
{1: {'decimation_factor': 1,
     'input_sampling_rate': 600.0,
     'output_sampling_rate': 600.0},
 2: {'decimation_factor': 1,
     'input_sampling_rate': 600.0,
     'output_sampling_rate': 600.0},
 3: {'decimation_factor': 1,
     'input_sampling_rate': 600.0,
     'output_sampling_rate': 600.0},
 4: {'decimation_factor': 3,
     'input_sampling_rate': 600.0,
     'output_sampling_rate': 200.0},
 5: {'decimation_factor': 10,
     'input_sampling_rate': 200.0,
     'output_sampling_rate': 20.0}}
```

Additionally it will now add dummy decimation attributes to PAZ stages without set decimation attributes as evalresp does not like that. Fixes #2121.

I realize that this adds a new method but it also fixes a but so I guess the maintance branch is okay here.

@Permian Can you also give this a quick review?


obspy/core/inventory/response.py
```diff
@@ -767,6 +767,114 @@ def __init__(self, resource_id=None, instrument_sensitivity=None,
...
+                "input_sampling_rate": input_sr,
+                "output_sampling_rate": output_sr,
+                "decimation_factor": factor}
+
+        # Find the first set input sampling rate. The output sampling rate
+        # cannot be set without it. Set all prior input and output sampling
+        # rates to it.
+        for i in stages:
+            sr = sampling_rates[i]["input_sampling_rate"]
+            if sr:
```
Jollyfant: Can you do `if sr:`? I always thought it had to be: `if sr is not None:`

obspy/core/inventory/response.py
```diff
@@ -1243,6 +1362,39 @@ def get_unit_mapping(key):
...
+                            "input_sampling_rate"]:
+                    blockette.decimation_input_sample_rate = \
+                        self.get_sampling_rates()[
+                            blockette.stage_sequence_number][
+                            "input_sampling_rate"]
+                # This branch get's large called for responses that only have a
+                # a single stage.
+                else:
+                    blockette.decimation_input_sample_rate = 1.0
+
```
Jollyfant: This solves the issue about missing decimation attributes on evalresp right? Is this set called during plotting only or when reading of the inventory?

Jollyfant:
Maybe I am overlooking something but this seems like a convoluted way of doing things. Can you not take the channel sample rate and then go over the stages in reverse order, multiply by the decimation factor to get the stage sample rates?

obspy/core/inventory/response.py
krischer: This is on purpose and a safety measure so the branch is not executed for neither `None` nor `0`.

obspy/core/inventory/response.py
krischer: Its only set when the response information is passed to evalresp (which is also used for plotting). Its done on a copy of the object so users' data is not modified.

krischer:
> Maybe I am overlooking something but this seems like a convoluted way of doing things. Can you not take the channel sample rate and then go over the stages in reverse order, multiply by the decimation factor to get the stage sample rates?

Response objects unfortunately don't have access to the attributes of their parent channel objects and they could also exist as independent objects. I agree that what I implemented here is a bit complicated but for "nice" response objects it is almost identical to your proposal - it just walks the response stages (from the other end) and sets the sampling rates. The rest is for dealing with all kinds of special cases.

The nice thing with the current solution that it should be able to infer the input and output sampling rates of most response stages as long as at least one stage has it set.

obspy/core/inventory/response.py
```diff
@@ -767,6 +768,124 @@ def __init__(self, resource_id=None, instrument_sensitivity=None,
...
+            if not si["decimation_factor"]:
+                si["decimation_factor"] = int(round(
+                    si["input_sampling_rate"] / si["output_sampling_rate"]))
+            output_sr = si["output_sampling_rate"]
+
+        def is_close(a, b):
+            return abs(a - b) < 1e-5
+
+        # Final consistency checks.
+        sr = sampling_rates[stages[0]]["input_sampling_rate"]
```
Jollyfant: First check is against itself right? `sampling_rates[stages[0]]` seems identical to `sampling_rates[i]` on its first iteration. Doesn't really matter anyway

obspy/core/inventory/response.py
```diff
@@ -1243,6 +1362,39 @@ def get_unit_mapping(key):
...
+                blockette = copy.deepcopy(blockette)
+                blockette.decimation_correction = 0.0
+                blockette.decimation_delay = 0.0
+                blockette.decimation_factor = 1
+                blockette.decimation_offset = 0
+                sr = self.get_sampling_rates()
+                if sr and blockette.stage_sequence_number in sr and \
+                        sr[blockette.stage_sequence_number][
+                            "input_sampling_rate"]:
+                    blockette.decimation_input_sample_rate = \
```
Jollyfant: You don't wanna set the decimation_factor too?

obspy/core/inventory/response.py
Jollyfant: Gotcha

obspy/core/inventory/response.py
krischer: Yea I thought about this as well but if no decimation attribute is set it must be 1 in any case. Just seems a tiny bit safer in case there is some response file out there which manages to confuse the `get_sampling_rates()` method.

obspy/core/inventory/response.py
krischer: Yes the first check is against itself - but later loop iterations are against the previous stage's output sampling rate. Is there a cleaner way to code this?

Jollyfant: Nah it's not gonna make it easier but I was just trying to follow the code paths

Permian:
From an user point of view this looks well-conceived. It would be nice if a warning could be raised in case the decimation attributes have to be filled with dummy values.

krischer:
> From an user point of view this looks well-conceived. It would be nice if a warning could be raised in case the decimation attributes have to be filled with dummy values.

I don't think that this is necessary. It does not change the response at all - evalresp just complains when it is not set.

Issue #2130, opened 2018-05-01
Fixing maintenance branch


krischer:
877791d0028c87729236f4c610bd4f66ccf10087 caused `master` to be merged into `maintenance_1.1.x` and I guess it has been accidentally pushed.

I attempted to fix this in the `maintenance-fixed` branch here: https://github.com/obspy/obspy/tree/maintenance-fixed

I used a mixture of rebasing and cherry-picking and it seems to be correct as far as I can tell. We have to force push this on the `maintenance_1.1.x` branch before we can continue. Can someone please double check that I found everything? It should include everything in the current `maintenance_1.1.x` minus the merge of `master` and minus https://github.com/obspy/obspy/pull/2120 plus the one commit immediately following. These will have to be redone after it is fixed.

The damage is limited but after we force push to `maintenance_1.1.x` we should protect the branch and disallow direct pushes.

**Don't merge anything into the maintenance branch until this is resolved!**

calum-chamberlain:
Sorry @krischer this is me :disappointed: definitely accidental, just meant to merge maintenance into master not the other way around (all done on github, rather than on my local, so the push happened immediately).

I started going through checking that commits match, but realised comparing with the current maintenance 1.1.x is stupid. Comparing with the commits that were supposed to be merged in #2120 looks good up until 8d17d997c9065a386b4802217b105686dbc10054 which in #2120 has no commit between that and 0ab5a99f9e828af0a39d4cee99544e2627e4fb76, whereas [your fixed branch](https://github.com/obspy/obspy/tree/maintenance-fixed) has quite a few - it looks strange to have had such a big gap in maintenance 1.1.x though?

krischer:
> Sorry @krischer this is me 😞 definitely accidental, just meant to merge maintenance into master not the other way around (all done on github, rather than on my local, so the push happened immediately).

No problem :-) Stuff like this happens. And we discovered it pretty quickly.


> I started going through checking that commits match, but realised comparing with the current maintenance 1.1.x is stupid. Comparing with the commits that were supposed to be merged in #2120 looks good up until 8d17d99 which in #2120 has no commit between that and 0ab5a99, whereas your fixed branch has quite a few - it looks strange to have had such a big gap in maintenance 1.1.x though?

Yea comparing with the current `maintenance_1.1.x` does not work. Some graphical tool makes it a lot easier.  If you are on OSX you can use this: http://gitup.co/ But there are a bunch of others out there.

[...]

PR #2127, opened 2018-04-30
String representation for empty origin objects

d-chambers:
### What does this PR do?
Fixes issue #2119

PR #2132, opened 2018-05-02
String representation for empty origin objects

krischer:
Supersedes #2127.

I took the liberty to rebase #2127 after #2130. Accidentially I force pushed the wrong local branch onto it and then lost permission to push again due to github's automatic push permission rules 🙈 Sorry about that.

In any case - here are the commits.


krischer:
Thanks for this!

PR #2135, opened 2018-05-03
minor fix in taup docs

megies:
### What does this PR do?

Fix an inaccurate docstring in taup

### PR Checklist


megies:
I don't care, to be honest, and I don't think it makes a big difference.. however, i think the variable is likely to get expanded during docs building, so if I'm right the docs would then sport an absolute path (from our docs buildbot server) which would be really confusing for people

PR #2129, opened 2018-05-01
Fallback to linear interpolation for instable bullen slowness evaluation

krischer:
It works in coordinates relative to the whole planet and the power low might have instabilities for very shallow layers near the surface. This PR just catches all numpy errors and falls back to linear interpolation in the case of shallow layers.

Fixes #2126 

@CovMat Can you have a look?

PR #2136, opened 2018-05-03
taup: py2 compat for #2129

megies:
### What does this PR do?

Make #2129 Python 2 compatible. `math.nan` Does not exist on Python < 3.5

### PR Checklist
- [ ] All tests still pass.


krischer:
Thanks! Maybe we should add a test for this? Problem is that this requires construction of a taup model which might be fairly slow.

PR #2139, opened 2018-05-07
fix map plots for old mpl 1.2

megies:
### What does this PR do?

Fix test fails on matplotlib 1.2, see e.g. http://tests.obspy.org/94743/#1

see #2089 

### PR Checklist
- [x] All tests still pass.


megies:
This makes all plots work on mpl 1.2 / centos7, besides one image test that has bad tick formatting, but I cant really care that much about that ancient mpl on centos7.

PR #2140, opened 2018-05-07
Remove some urllib and future.standard_library usages

### LINKED ISSUE #1254: Use requests instead of urllib

megies:
### What does this PR do?

- remove remaining instances of `with future.standard_library.hooks():` with either importing in a `if/else`  depending on Python 2/3 or by using `requests.compat` (which is internally doing just the same as well)
- remove some `urllib` usages with `requests`

Closes #1254. Even if not all urllib usage is replaced, I think getting rid of `future.standard_library` should resolve the problem reported in #1254 (e.g. seen here: http://tests.obspy.org/35446/)

### PR Checklist
- [ ] +DOCS because some changes might affect the gallery
- [x] +TESTS:ALL
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/httpproxy.py
+++ b/obspy/clients/httpproxy.py
@@ -13,14 +13,12 @@ J. MacCarthy, modified from https://gist.github.com/frxstrem/4487802
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA
-from future import standard_library
 from future.utils import native_str
 
 from base64 import b64encode
 import socket
-with standard_library.hooks():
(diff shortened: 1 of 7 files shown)


megies:

megies:
Bump to 2.0.0

trichter:
Obsolete.

PR #2141, opened 2018-05-07
Higher tolerance for an image test that likes acting up

krischer:
Makes the one remaining image tests that fails with CentOS 7 pass. See #2139.


krischer:
Beautiful green CI.

PR #2148, opened 2018-05-14
Change dummy time.strptime call to use locale-independent format #2147

### LINKED ISSUE #2147: FDSN routing client has a locale dependency
There's a dummy call to `time.strptime` in the module init that uses locale-specific formatting, which fails under locales that don't use the same names (ie. "Nov" for the 11th month of the year).

```
>>> import locale
>>> locale.setlocale(locale.LC_TIME, ('zh_CN', 'UTF-8'))
'zh_CN.UTF-8'
>>> from obspy.clients.fdsn.routing.routing_client import RoutingClient
  File "/workspace/anaconda/envs/django/lib/python2.7/site-packages/obspy/clients/fdsn/__init__.py", line 242, in <module>
    from .routing.routing_client import RoutingClient  # NOQA
  File "/workspace/anaconda/envs/django/lib/python2.7/site-packages/obspy/clients/fdsn/routing/__init__.py", line 25, in <module>
    time.strptime("30 Nov 00", "%d %b %y")
  File "/workspace/anaconda/envs/django/lib/python2.7/_strptime.py", line 478, in _strptime_time
    return _strptime(data_string, format)[0]
  File "/workspace/anaconda/envs/django/lib/python2.7/_strptime.py", line 332, in _strptime
    (data_string, format))
ValueError: time data u'30 Nov 00' does not match format u'%d %b %y'
```

I believe switching this to an ISO8601-like string would be locale-agnostic:

    time.strptime("2000/11/30", "%Y/%m/%d")

adam-iris:
Fixes #2147

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -22,6 +22,7 @@
      and/or `location` are set (see #1810, #2031, #2047).
    * A few fixes and stability improvements for the mass downloader
      (see #2081).
+   * Fixed routing startup error when running under certain locales (see #2147)
  - obspy.imaging:
    * Normalize moment tensors prior to plotting in the mopad wrapper to
      stabilize the algorithm (see #2114, #2125).
--- a/obspy/clients/fdsn/routing/__init__.py
+++ b/obspy/clients/fdsn/routing/__init__.py
@@ -22,7 +22,7 @@ from future.builtins import *  # NOQA
 #
 # See https://bugs.python.org/issue7980
 import time
-time.strptime("30 Nov 00", "%d %b %y")
+time.strptime("2000/11/30", "%Y/%m/%d")
 
 
 if __name__ == '__main__':  # pragma: no cover


megies:
@adam-iris, it'd be nice if you added a short changelog entry for this. Also, in the future I'd recommend working on dedicated branches for fixes/features, because now the `maintenance_1.1.x` branch on your fork is out of sync with the one one the main repo. You should avoid using those branches (`master`, `maintenance...`, ...) on your fork, in 99% of cases.

krischer:
> I don't think there's a chance we can add a test for this, as during our unittests obspy is already imported..

We could set the locale to something not UTF8 for on test case runners. That might be ale to find some extra errors.

Otherwise I'm also fine with this PR. Thanks a bunch!

megies:
> We could set the locale to something not UTF8 for on test case runners. That might be ale to find some extra errors.

Good idea..

adam-iris:
@megies sorry about the branch, I realized after submitting the PR that I hadn't followed the branching guidelines correctly. 

I didn't test non-UTF8 locales, that does seem like a case that might still break. I agree that explicitly setting the locale (and then resetting it) around this call might be a safer option, but localization confuses and frightens me so I didn't dig too hard into this :P

megies:
No worries @adam-iris . Just remember to either ignore your maintenance branch on the fork or resync it with the main repo one. Or well.. maintenance_1.1.x will not be used after the upcoming 1.1.1 release, so.. nevermind I guess.. ;-)

This had a pretty high likelyhood to hurt people so thanks again for letting us know. It's really good to have this fixed. non-UTF8 is rather esoteric though, I can't see how many people would use something like that..

PR #2157, opened 2018-05-23
fix a bug in AR picker

megies:
### What does this PR do?

Presumably fixes buggy AR picker C code. Fix is done by @jwassermann.

I'll open it against master for inclusion into 1.2.0 since 1.1.1 is already in release candidate stage and also because this might change results significantly (just guessing, really).

~~It's a bit scary that the AR picker test doesn't even need to be adapted, but hey that's life.~~ Test was adapted/improved.

### PR Checklist
- [x] All tests still pass.

CC @jwassermann

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -33,6 +33,8 @@ master:
      objects to shapefile (see #2012)
  - obspy.io
     * added read support for receiver gather format v. 1.6 (see #2070)
+ - obspy.signal.trigger:
+    * fix a bug in AR picker (see #2157)
  - obspy.signal.PPSD:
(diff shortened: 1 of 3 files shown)


krischer:
Any chance to add a regression test? @jwassermann do you have some data that originally triggered the bug?

megies:
> Any chance to add a regression test? @jwassermann do you have some data that originally triggered the bug?

Sure we can, the change is pretty subtle though, the result is changed for the S pick only and also only slightly. In the test case from 32.2s to 32.1s or so. I'll make the test more fine-grained again, there's a strange comment in there that the result is very machine dependent and therefore it's rounding to whole seconds in the test..

PR #2164, opened 2018-05-29
Import Directive from stdlib in sphinx extension.

QuLogic:
We already do this in the `gallery.py` extension.

This fixes build against Sphinx 1.7 which removed the deprecated compatibility shims.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/misc/docs/source/_ext/autosummary/__init__.py
+++ b/misc/docs/source/_ext/autosummary/__init__.py
@@ -62,13 +62,12 @@ import inspect
 import posixpath
 
 from six import text_type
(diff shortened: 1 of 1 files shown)


QuLogic:
I don't think the doc warnings are my fault? But maybe.

megies:
> I don't think the doc warnings are my fault? But maybe.

No idea, nobody ever got round to fix all those warnings (I started at some point but couldnt get through all of it)

megies:
And even so.. the docs buildbot is still running on spinx 1.2 so it doesnt say much about the proposed change.

PR #2167, opened 2018-06-05
UTCDateTime.strftime() raises ValueError with year <1900

megies:
### What does this PR do?

Trying to fix `UTCDateTime.strftime()` which raises `ValueError` in `datetime.datetime.strftime()` if year is <1900.

This currently fixes the problem for e.g. `.strftime('%Y-%m-%d')` by using simple string formatting instead of `datetime.datetime.strftime()`.

Plotting catalogs that have events with origin time <1900 currently does not work.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,7 @@
    * Fix Stream.get_gaps() when a trace is completely overlapping another trace
      (see #2218).
    * Fig Exception when comparing ComparingObjects (see #2220).
+   * Fix UTCDateTime.strftime() when year is <1900 on Python 2 (see #2167)
  - obspy.clients.arclink:
    * Raise a warning at import time that the ArcLink protocol will be
      deprecated soon (see #1987).
--- a/obspy/core/tests/test_event.py
+++ b/obspy/core/tests/test_event.py
@@ -6,6 +6,7 @@ from future.utils import PY2, native_str
 
 import builtins
 import copy
+import io
 import os
 import sys
 import unittest
@@ -23,6 +24,7 @@ from obspy.core.utcdatetime import UTCDateTime
 from obspy.core.util import (
     BASEMAP_VERSION, CARTOPY_VERSION, PROJ4_VERSION, MATPLOTLIB_VERSION)
 from obspy.core.util.base import _get_entry_points
+from obspy.core.util.misc import MatplotlibBackend
 from obspy.core.util.testing import ImageComparison
 from obspy.core.event.base import QuantityError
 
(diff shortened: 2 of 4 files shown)


krischer:
Just using `str(dt)` should fix this as implemented in https://github.com/obspy/obspy/pull/1325.

Also see #2013, #2015, #1318.

krischer:
We cannot really fix `.strftime()` on Python 2 aside from reimplementing it from scratch which we really don't want I think. How about just raising a nice error on Python 2 when calling `UTCDateTime.strftime()` with a year < 1900 and make the user's aware of the issue?

megies:
I started fixing that inside the catalog plot routine by using string formatting, but then I thought it's a low hanging fruit at least for some cases and it might be useful for users as well to implement some easy workarounds inside `UTCDateTime.strftime`.. dunno

megies:
> Just using str(dt) should fix this as implemented in #1325.

I just had a quick look at this, and actually I think the fix proposed here is the way to go.
 - it's really only superseding `datetime.datetime.strftime` when that particular exception is hit
 - replacing the `strftime` commands in all relevant places (mostly where origin times are converted to strings) would introduce roughly 10 changes scattered across the code base, which will be obsolete after dropping python 2
 - this change can easily be reverted if necesssary when dropping Python 2
 - this change also enables users to use `t.strftime()` with times before 1900AD in user land codes on Python 2

krischer:
Aside from the CI failures this seems to be a in a pretty good state. Can you maybe add a few more tests for the `_strftime_replacement()` method? This is easy enough to do just to make sure it works as intended even though is rarely actually called.

ThomasLecocq:
```
File "/home/travis/miniconda/envs/test-environment/lib/python3.7/site-packages/obspy/core/tests/test_utcdatetime.py", line 1136, in test_strftime_with_years_less_than_1900
self.assertEqual(t.strftime('%Y-%m-%d'), '0998-11-09')
AssertionError: '998-11-09' != '0998-11-09'
- 998-11-09
+ 0998-11-09
? +
```
only happens on Linux...

the MacOS eerrors with Taup are unrelated.

megies:
Upstream bug in Python: https://bugs.python.org/issue32195

krischer:
Not a bug according to the Python devs: https://bugs.python.org/issue32195

So this works on linux and python 3:
```python
datetime.datetime(1, 1, 1).strftime("%04Y")
```

So I guess what we have to do it test for linux and Python >= 3.5 and do:

```python
format = format.replace("Y", "%04Y")
```

ThomasLecocq:
side note, this failed 
```ipython
In [2]: UTCDateTime("998-01-01")
Out[2]: 9980-10-10T00:00:00.000000Z

In [3]: UTCDateTime("0998-01-01")
Out[3]: 0998-01-01T00:00:00.000000Z
```

and is corrected by a regex check, considering that the first part of a -, / or comma separated string is the year.

ThomasLecocq:
we regex check and reject the string years not being 4 digits

megies:

megies:
CI is green, only a minor flake8 thing and changelog missing which I'll add and then merge.

PR #2169, opened 2018-06-07
Update some FDSN WS urls

claudiodsf:
IPGP has migrated its EIDA node to RESIF, it however still maintains an FDSN WS.
The current address is still http://eida.ipgp.fr, but it will soon change to http://fdsnws.ipgp.fr.

This pull request is a reminder of the fact that the URL mapping for `IPGP` should be changed as soon as the URL changes.

I hope that IPGP will be able to change the URL before the next maintenance release of ObsPy.

I also updated the link in the comments pointing to EIDA node list.

### DIFF
--- a/obspy/clients/fdsn/header.py
+++ b/obspy/clients/fdsn/header.py
@@ -46,7 +46,7 @@ URL_MAPPINGS = {
     "GFZ": "http://geofon.gfz-potsdam.de",
     "ICGC": "http://ws.icgc.cat",
     "INGV": "http://webservices.ingv.it",
-    "IPGP": "http://eida.ipgp.fr",
+    "IPGP": "http://fdsnws.ipgp.fr",
     "IRIS": "http://service.iris.edu",
     "ISC": "http://isc-mirror.iris.washington.edu",
     "KOERI": "http://eida.koeri.boun.edu.tr",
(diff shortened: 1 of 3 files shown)


claudiodsf:
The new URL is live (http://fdsnws.ipgp.fr) and I pushed the change of the alias.

I also changed the arclink url in the `arclink.Client()` documentation to `arclink.ipgp.fr:18001`.

This is good merge from my point of view.

obspy/clients/fdsn/header.py
```diff
@@ -46,7 +46,7 @@ class FDSNNoDataException(FDSNException):
     "GFZ": "http://geofon.gfz-potsdam.de",
     "ICGC": "http://ws.icgc.cat",
     "INGV": "http://webservices.ingv.it",
-    "IPGP": "http://eida.ipgp.fr",
+    "IPGP": "http://fdsnws.ipgp.fr",
```
megies: @claudiodsf can you please also adapt the doctest in `__init__.py`, thanks

megies:
+TESTS:clients.fdsn

claudiodsf:
Done. Squashed into the previous commit and force-pushed.

obspy/clients/fdsn/header.py
claudiodsf: Done. I updated the commit and force-pushed.

claudiodsf:
Thanks 👍

PR #2176, opened 2018-06-19
Optional user-supplied marker styles

jwellik:
Allows the user to specify the marker style, color, and size for events. Defaults to a yellow star with size 12 if nothing is provided.

!! Please check that you select the **correct base branch** (details see below link) !!

Before submitting a PR, please review the pull request guidelines:
https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Also, please make sure you are following the ObsPy branching model:

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/imaging/waveform.py
+++ b/obspy/imaging/waveform.py
@@ -548,6 +548,9 @@ class WaveformPlotting(object):
         else:
             time = event["time"]
             text = event["text"] if "text" in event else None
+            marker = event["marker"] if "marker" in event else "*"
+            marker_color = event["marker_color"] in event else "yellow"
+            marker_size = event["marker_size"] in event else 12
 
         # Nothing to do if the event is not on the plot.
         if time < self.starttime or time > self.endtime:
@@ -616,8 +619,8 @@ class WaveformPlotting(object):
                             relpos=relpos, shrinkB=7),
                         zorder=10)
         # Draw the actual point. Use a marker with a star shape.
-        ax.plot(x_pos, y_pos, "*", color="yellow",
-                markersize=12, linewidth=self.linewidth)
+        ax.plot(x_pos, y_pos, marker, color=marker_color,
+                markersize=marker_size, linewidth=self.linewidth)
 
         for pick in getattr(event, 'picks', []):
             # check that network/station/location matches


obspy/imaging/waveform.py
```diff
@@ -548,6 +548,9 @@ def _plot_event(self, event):
         else:
             time = event["time"]
             text = event["text"] if "text" in event else None
+            marker = event["marker"] if "marker" in event else "*"
+            marker_color = event["marker_color"] in event else "yellow"
+            marker_size = event["marker_size"] in event else 12
```
megies: If we make modifications here..
 - a) we need to stick to matplotlib kwarg naming conventions of the `Axes` plotting method that is used (e.g. `markersize` in `Axes.plot()` -- not `marker_size`)
 - b) we might want to right away allow any modifications in the used matplotlib plotting method, i.e. pass through arbitrary kwargs (i.e. pass through anything left in `event` after extracting `time` and `text` to matplotlib)
 - c) if there's a chance that eventually we need to pass through some other kwargs to some other underlying routine, we should group the matplotlib kwargs (like `markersize`) into a subdictionary in `event`, e.g. `event['plot_kwargs`] = {'markersize': 15}` (e.g. other people might want to modify the `annotate()` call or the plot call for picks..)
 - d) maybe we should check if it makes sense to switch the plotting routine from `plot()` to `scatter()` (but maybe that's not really necessary, just thought that scatter might be in principle more appropriate)

a) is a must from my point of view
b) is also strongly recommended I think
c) dunno.. maybe not that important but it's a decision we need to consider right now
d) dunno.. maybe not important

megies:
One more thing.. we need to add a default for marker edge color, the stars look bad with default mpl settings: http://docs.obspy.org/master/tutorial/code_snippets/waveform_plotting_tutorial.html#creating-a-one-day-plot

waveform_plotting_tutorial_5

obspy/imaging/waveform.py
jwellik: 'a' makes sense to me. I think it also makes sense to implement 'c'. I definitely would like to be able to customize the marker, but I think it might be useful to people in the future to customize other things like the text. 'd' It probably does make sense in principal but is less of a priority to me personally.

jwellik: @megies I'm new to crowd-sourced coding an making edits to larger projects on Github. I've only used Github as version control for my own code. Nevertheless, I can take a crack at implementing changes a & c. Is that a good path forward?

megies: Sure, go ahead. Might be too big of a change to make it in online editor though.. maybe you should switch to working on your local computer in whatever IDE you normally use.

Issue #2174, opened 2018-06-16
[MassDownloader] String Template: do not require all the fields


paride:
Currently the "String Template" option for `mseed_storage` requires all the fields to be present in the string (`{network}`, `{station}`, `{location}`, `{channel}`, `{starttime}`, and `{endtime}`). Sometimes it can be useful to define only some of them. Practical case: while working with a single station I didn't define `{network}` and `{station}`, and the substitution didn't work.

What about changing the ANDs to ORs here: https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/mass_downloader/utils.py#L545 ?

The same goes for `stationxml_storage`: https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/mass_downloader/utils.py#L489.

megies:
Sounds totally reasonable to me.

I'd make it..

```python
elif any('{%s}' % key  in str_or_fct for key in ('network', 'station', ...)):
```

paride:
This will have to be tackled differently: https://github.com/obspy/obspy/pull/2178#issuecomment-401127129.

ThomasLecocq:
Are we still waiting for this for 1.2.0 ? @paride @krischer

krischer:
I'd argue its not urgent - if we have time we do it for 1.2.0, otherwise we punt it to the next release.

paride:
I am moving away from seismology, so I doubt I will work on this any soon. Feel free to close the issue if you prefer.

megies:
I think I'll close this. It's not urgent and users should just stick to the default usage of having all items in the string.

PR #2178, opened 2018-06-20
MassDownloader: flexible string template

paride:
### What does this PR do?

Lets the mass downloader string template work without specifying all the fields.
Fixes: #2174. 

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.fdsn"
- [ ] All tests still pass.
- [x] First time contributors have added your name to `CONTRIBUTORS.txt`.

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/utils.py
+++ b/obspy/clients/fdsn/mass_downloader/utils.py
@@ -542,8 +542,9 @@ def get_mseed_filename(str_or_fct, network, station, location, channel,
     if callable(str_or_fct):
         path = str_or_fct(network, station, location, channel, starttime,
                           endtime)
-    elif any("{%s}" % key in str_or_fct for key in ("network", "station", \
-                "location", "channel", "starttime", "endtime")):
+    elif any("{%s}" % key in str_or_fct for key in ("network", "station",
+                                                    "location", "channel",
+                                                    "starttime", "endtime")):
         path = str_or_fct.format(
             network=network, station=station, location=location,
             channel=channel, starttime=starttime.strftime(strftime),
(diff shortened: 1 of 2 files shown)


obspy/clients/fdsn/mass_downloader/utils.py
```diff
@@ -542,9 +542,8 @@ def get_mseed_filename(str_or_fct, network, station, location, channel,
     if callable(str_or_fct):
         path = str_or_fct(network, station, location, channel, starttime,
                           endtime)
-    elif ("{network}" in str_or_fct) and ("{station}" in str_or_fct) and \
-            ("{location}" in str_or_fct) and ("{channel}" in str_or_fct) and \
-            ("{starttime}" in str_or_fct) and ("{endtime}" in str_or_fct):
+    elif any("{%s}" % key in str_or_fct for key in ("network", "station", \
+                "location", "channel", "starttime", "endtime")):
```
megies: please get rid of redundant backslash (inside brackets) and fix indent of continuation line, see circle CI with code check results

paride:
@krischer thanks for your feedback. If you don't think this is a good idea we can rediscuss it. I did't think this as a problematic change: an ambiguity in the file name should be pretty clear to spot, but you certainly have a broader vision.

krischer:
> @krischer thanks for your feedback. If you don't think this is a good idea we can rediscuss it. I did't think this as a problematic change: an ambiguity in the file name should be pretty clear to spot, but you certainly have a broader vision.

The problem I see is that the filename is used for a lot of things. The mass downloader is for example designed so that it can run multiple times and will only download missing data. If there is now ambiguity in the filenames it might wrongly assume that some data already exist and will not attempt to download it in the first place. This would be really hard to spot for users.

Now that I've thought a bit more about it: I think a better solution would be to (right at the start of the `.download()` method) check if the passed string is supposed to be a string template (e.g. check for the curly braces), and if yes: make sure all required fields and no extra fields are present. If that test fails, raise a nice and instructive errors message.

This is potentially a bit more hassle to use than your solution but I feel like it would avoid a lot of potential confusion.

paride:
I'm closing this for the reasons given in @krischer's last comment.

Issue #2165, opened 2018-05-29
miniseed to ascii conversion via stream.write not writing some time stamps correctly


SCEDC:
It looks like some time values are not being written correctly using `stream.write(format='TSPAIR')` where the stream was `read(<miniseed file>)`.  I've attached a miniseed file, the ascii output from `stream.write`, and also the ascii output from `mseed2ascii`, (https://github.com/iris-edu/mseed2ascii).
I was expecting these outputs to match.

Before submitting an Issue, please review the [Issue Guidelines](https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md#submitting-an-issue).

* Please check whether the bug was already reported or fixed.
* Please provide the following information:
  -  ObsPy version, Python version and Platform (Windows, OSX, Linux ...)
1.1.0, OSX, Python 3.6.4

  -  How did you install ObsPy and Python (pip, anaconda, from source ...)
anaconda

  -  If possible please supply a short, self contained, correct example that
     demonstrates the issue.

```python
st = read('compress_0528_0100_CIAGO__HNZ___2016172.ms')
st.write('compress_0528_0100_CIAGO__HNZ___2016172.txt', format='TSPAIR')
```

  -  If this is a regression (used to work in an earlier version of ObsPy),
     please note when it used to work.

[obspy_submit.tar.gz](https://github.com/obspy/obspy/files/2051436/obspy_submit.tar.gz)

megies:
Can confirm this is a bug.

claudiodsf:
I propose a fix for this in #2179.

claudiodsf:
I think that the solution in #2179 is better code than the existing one. Nevertheless, the `np.linspace()` approach seems to unveil some sort of time formatting bug in `UTCDateTime`, which I synthesize here:

```python
>>> ts1 = 1.466387732989999771e+09
>>> ts2 = 1.466387732999999762e+09
>>> print(ts2-ts1)
0.009999990463256836
>>> time1 = UTCDateTime(ts1)
>>> print(time1)
2016-06-20T01:55:32.990000Z  # microseconds seem rounded here
>>> print(time1.microsecond)
989999
>>> time2 = UTCDateTime(ts2)
>>> print(time2)
2016-06-20T01:55:32.000000Z  # problem here (should be 32.999999 or 33.000000)
>>> print(time2.microsecond)
999999 # no problem here
>>> print(UTCDateTime(ts2).timestamp - ts2)
0.0 # no problem here
```

[...]

PR #2183, opened 2018-06-27
Regression test for #2165

krischer:
Just to make sure


krischer:
Appveyor failures are unrelated.

PR #2188, opened 2018-07-02
WIP: Fix 32bit appveyor failures

krischer:
There have been some 32bit appveyor failures. Just trying to figure out what they are remotely as I don't have access to a windows machine right now.

### DIFF
--- a/obspy/core/tests/test_stream.py
+++ b/obspy/core/tests/test_stream.py
@@ -6,6 +6,7 @@ from future.builtins import *  # NOQA
 import inspect
 import os
 import pickle
+import platform
 import unittest
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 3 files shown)


krischer:
Does anyone understand what is happening here? It only happens on Windows 32bit so its a bit tedious for me to debug.

I realize that comparing floating point numbers is dangerous, but here we are literally comparing

```python
st1 = read(...)
st2 = read(...)

st1.simulate(...)
for tr in st2:
     tr.simulate(...)

assert st1 == st2
```

the `Stream.simulate()` method internally does nothing different so I don't see how this test could fail. Maybe one of the upstream packages like `numpy`, `scipy`, `BLAS`/`LAPACK` has some misbehaving cache here?

@barsch Is there any chance you could debug this locally? It takes forever to do over AppVeyor.

barsch:
@krischer no time today - will look into it Wednesday - which 32bit python version should I play with 2.7, 3.5 or 3.6?

krischer:
Thanks! Python 3.6. The exact environment is written here: https://ci.appveyor.com/project/obspy/obspy/build/1.0.6110-maintenance_1.1.x/job/ib5le487t949gj5j

barsch:
my local environment works without issues - see https://gist.github.com/barsch/1b305aa1b6e4a67d2573f8e9ab0b1aca

will setup an anaconda environment and test again

barsch:
unfortunately same result: https://gist.github.com/barsch/7ccc2b96934287a604691bbda2e21c12

no exception in a new conda setup using the same installation order and packages as listed here: https://ci.appveyor.com/project/obspy/obspy/build/1.0.6110-maintenance_1.1.x/job/ib5le487t949gj5j

how should I proceed further?

krischer:
Hmm - I guess we could just weaken the test a bit and use `np.testing.assert_allclose()` for the data on 32 bit windows. Should be safe enough and we at least get rid of all the annoying appveyor failures.

krischer:
Alright - a slightly silly "fix" but it seems to work.

PR #2189, opened 2018-07-02
More forgiving RESP file parsing

### LINKED ISSUE #2170: Type Error while reading .resp file with read_inventory
Dear all, 

this is my first issue, i'll do my best to give all the information needed.
I try to read a response file in RESP format using obspy.read_inventory, basically like this:

`import obspy 
inv = obspy.read_inventory('/run/media/felixw/workdisc_1/6D6-Trillium-250sps.resp', format='RESP')`

This leads to the following error message: 

File "/home/felixw/anaconda3/lib/python3.6/site-packages/obspy/io/xseed/parser.py", line 1127, in get_response_for_channel
stages[b.stage_sequence_number].append(b)

TypeError: unhashable type: 'list'

I tried converting the .resp file into a dataless seed and read that one - converting works, read_inventory gives the same error. You can find the .resp file here: https://www.dropbox.com/s/644r9lsdqv0larx/6D6-Trillium-250sps.resp?dl=0 (couldn't upload it directly to this issue).

Do you have any idea, what i could do to solve that? That would be a great help to me, thanks a lot!

Some information on my ObsPy and Python: i installed both via anaconda3, using Python 3.6 and ObsPy 1.1 on a Fedora 27 computer.

Best wishes and again, thanks for helping,
Felix

krischer:
Fixes #2170. I would consider the given file to not be fully standards compatible but there is no proper RESP standard in any case so that does not mean much.

The issue was that the RESP parser currently had two conditions to split blockettes when initially reading in the files: A different blockette id or a plus sign in  thecomments. There is now a third condition: A decrease in the field number to 3.

I feel like this is a fairly safe change. Includes a test.


megies:
> The issue was that the RESP parser currently had two conditions to split blockettes when initially reading in the files: A different blockette id or a plus sign in thecomments. There is now a third condition: A decrease in the field number to 3.

Yuck, never dug into RESP so far but it really sounds like it has some design flaws.. ;-)

krischer:
> Is this a valid RESP file? If yes, how do other RESP parser handle this edge case?

Don't know how other parsers do this but I'm pretty sure what we are doing here is not wrong (ignoring the question if its a valid file or not). And I'm also not sure what constitutes a valid RESP file in the first place. I'm not aware of a proper specification. 


> Also, is this limited to blockette 58 ?

I think blockette 58 is pretty much the only blockette where this could happen in practice (in a somewhat well behaved file) - basically as stage gain + total sensitivity following directly one after another.

obspy/io/xseed/parser.py
```diff
@@ -700,12 +702,23 @@ def record_type_from_blocketteid(bid):
                 # g[3]: Everything afterwards.
                 g = m.groups()
                 blockette_number = g[0]
+                field_number = int(g[1])
                 # A new blockette is starting.
-                if blockette_number != last_blockette_id:
+                # Either
+                # * when the blockette number increases or
+                # * when the blockette number suddenly decreases to 3 again.
```
barsch: field number decreases to 3

krischer:
Thanks @barsch for the review! I implented the suggestions and also added a direct test against what evalresp directly would do. IMHO ready to merge!

markcwill:
@krischer @megies @barsch sorry I'm late to this, we've also definitely hit this problem, I'm surprised this is the first time ObsPy has seen it, it's a common occurrence in IRIS Nominal Response Library RESP files. Here's a random one:
https://ds.iris.edu/NRL/sensors/kinemetrics/RESP.XX.NS470..BNZ.Episensor2.DC_200.20VDF.2G

I believe right now we do the same thing: check for field number decrementing. 

I think the ideal is to:
1. Check for stage sequence number change
2. Check for blockette number change
3. Check for field number decrement

We're doing all these now, but not necessarily in that order. I think if you do (1) & (2) in that order, then you may not need (3) (since 58 starts with a stage sequence number which would trigger (1)). We don't rely on comments, since they could be anything. Ideally, you wouldn't need (2) either, but unfortunately, stage number comes after transfer function type in some important blockettes (like 53, 54, etc). Pretty ugly format design IMHO.

Issue #2194, opened 2018-07-04
problem writing segy file: maximum number of traces (too large number in header field)


superzappie:
Hello,
I am encountering the problem where I can't write all my traces to a segy file using obspy. Is there any way around this problem; can I still write segy files with more than 32767 traces?

In another thread I found: 
https://github.com/obspy/obspy/issues/1393
a similar problem arose, then it was about the amount of samples within a trace. 
A minimalistic working example is provided below, as well as the error message. 
As a note, I was able to read in segy files with more than 32767 traces. 
Thanks in advance. 

error: short format requires (-32768) <= number <= 32767

```python
import numpy as np
from obspy import Trace, Stream
import sys
from obspy.io.segy.segy import SEGYTraceHeader, SEGYBinaryFileHeader
from obspy.core import AttribDict

stream = Stream()

for i in range(32768):
    trace=Trace(data=np.array([1,2,3,4],dtype='float32'))
    trace.stats.delta = 0.000125
    if not hasattr(trace.stats,'segy.trace_header'):
        trace.stats.segy={}
    trace.stats.segy.trace_header = SEGYTraceHeader()
    stream.append(trace)

stream.stats = AttribDict()
stream.stats.binary_file_header = SEGYBinaryFileHeader()

stream.write("migrated_infilled.sgy", format="SEGY", data_encoding=1,
             byteorder=sys.byteorder)
```

krischer:
The issue arises from the fact that upon writing SEG-Y data ObsPy writes the "Number of data traces per ensemble" field in the binary file header and sets it to the number of traces. This field is (according to the documentation) mandatory for prestack data. Its a signed 2-byte integer (its also not really clear if some integers in SEG-Y can be unsigned so we currently opt for maximal compatibility) so the largest number it can represent is 2^15-1 = 32767.

The SEG-Y documentation is unfortunately really vague about what exactly that means but that is what we are using right now. If this can be interpreted in a different way I'd be happy to change it and then ObsPy would happily write larger files (it can already read arbitrary sized SEG-Y files).

Let me know what you think.

superzappie:
Thank you for your quick reaction. Indeed, setting:
stream.stats.binary_file_header.number_of_data_traces_per_ensemble=1
solves the problem. Writing the number of traces to this field is not necessary, in my case at least. Might have been nice if the exception somehow pointed me towards this field.

krischer:
> Might have been nice if the exception somehow pointed me towards this field.

[...]

PR #2196, opened 2018-07-05
More verbose error when header packing in SEG-Y fails

krischer:
SEG-Y is fairly restrictive in the value ranges of many of its header values. This PR causes this to raise nicer error messages to the error is easier to diagnose. The example from #2194 now fails with the following error message:

```
  File "/Users/lion/workspace/code/obspy/obspy/io/segy/util.py", line 55, in _pack_attribute_nicer_exception
    raise ValueError(msg % (name, str(x), format, e.args[0]))
ValueError: Failed to pack header value `number_of_data_traces_per_ensemble` (32768) with format `<h` due to: `short format requires (-32767 -1) <= number <= 32767`
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -50,6 +50,8 @@ master:
      (see #2106, #2098, #2095)
  - obspy.io.sac:
    * Fix bug writing inventory with SOH channels to SACPZ (see #2200).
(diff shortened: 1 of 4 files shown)


megies:

megies:
Test fail unrelated, merging! :rocket:

Issue #2198, opened 2018-07-11
_read_rg16 merge and contacts_north raises assertion error


d-chambers:
The `merge` option is not compatible with the `contacts_north` option in `_read_rg16` if three component files are read because the native datatype of the rg16 files is `np.dtype('>f4')` but the data type of the array is changed when the Z component is multiplied by -1 to `np.dtype('float32')` [here](https://github.com/obspy/obspy/blob/a77b38443e8c5ec139adf82168cc026f4e4b47ed/obspy/io/rg16/core.py#L167-L168). `_merge_` then asserts there is only one dtype [here](https://github.com/obspy/obspy/blob/a77b38443e8c5ec139adf82168cc026f4e4b47ed/obspy/io/rg16/util.py#L168), raising an assertion error.

PR #2199, opened 2018-07-11
Fix 2198

d-chambers:
### What does this PR do?
Fixes issue #2198 by multiplying each trace array by 1 after reading it with `_read_rg16` and related functions. This creates a new array with dtype float32 rather than '>f4' thus ensuring the datatypes will always be the same for traces.  

Note: rg16 module is only in master, hence the master branch is selected. 

#2198.

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/obspy/io/rg16/util.py
+++ b/obspy/io/rg16/util.py
@@ -60,7 +60,7 @@ def _read_block(fi, spec, start_bit=0):
     return out
 
 
-def _read(fi, position, length, dtype):
+def _read(fi, position, length, dtype, new_dtype=None):
     """
     Read one or more bytes using provided datatype.
 
@@ -76,6 +76,7 @@ def _read(fi, position, length, dtype):
             >i3 - big endian 3 byte int
             >i. - 4 bit int, left four bits
     :type dtype: str
+    :param new_dtype: Any valid numpy data type.
     :return:
     """
     # if a list is passed as parameters then recurse through each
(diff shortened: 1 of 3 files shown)


obspy/io/rg16/util.py
```diff
@@ -94,7 +94,8 @@ def _read(fi, position, length, dtype):
     if dtype in READ_FUNCS:
         return READ_FUNCS[dtype](fi, length)
     else:
-        data = np.fromstring(fi.read(int(length)), dtype)
+        # multiply 1 to get new array with type np.float32 see #2198
+        data = np.fromstring(fi.read(int(length)), dtype) * 1
```
megies: Uhm.. isn't there a more expressive/canonical way to achieve what you're after here? Maybe `np.fromstring(fi.read(int(length)), dtype).astype(np.float32)`??

d-chambers: You are right, that would be more clear

d-chambers:
it looks like the tests past but something else (maybe related to coverage) is failing the build.

obspy/io/rg16/util.py
```diff
@@ -94,7 +94,8 @@ def _read(fi, position, length, dtype):
     if dtype in READ_FUNCS:
         return READ_FUNCS[dtype](fi, length)
     else:
-        data = np.fromstring(fi.read(int(length)), dtype)
+        # multiply 1 to get new array with type np.float32 see #2198
+        data = np.fromstring(fi.read(int(length)), dtype).astype(np.float32)
```
megies: Ah, that looks much better. :tada: 
Now if you fix that comment above it I'm happy.. ;-)

megies:
> it looks like the tests past but something else (maybe related to coverage) is failing the build.

Hmm.. I don't know, that really looks like an actual problem. The rg16 tests can't even be executed in CI..https://ci.appveyor.com/project/obspy/obspy/build/1.0.6155-master/job/tf07gsrtek7mei5x#L695

d-chambers:
I think so

PR #2200, opened 2018-07-17
skip chans without instrument sensitivity

dsentinel:
When writing SAC pz's to a file from inventory, if there channels that do not have a PolesZerosResponsStage (e.g. SOH), writing will fail raising and exception.

### DIFF
--- a/obspy/io/sac/sacpz.py
+++ b/obspy/io/sac/sacpz.py
@@ -16,6 +16,7 @@ from future.builtins import *  # NOQA
 from future.utils import native_str
 
 import numpy as np
+import warnings
 
 from obspy import UTCDateTime
 from obspy.core import AttribDict
@@ -47,7 +48,17 @@ def _write_sacpz(inventory, file_or_file_object):
             for cha in sta:
                 resp = cha.response
                 sens = resp.instrument_sensitivity
-                paz = resp.get_paz()
+                try:
+                    paz = resp.get_paz()
+                except Exception:
+                    msg = "{}.{}.{}.{} {} has no paz. Skipping.".format(
+                        net.code,
+                        sta.code,
+                        cha.location_code,
+                        cha.code,
+                        cha.start_date)
+                    warnings.warn(msg)
+                    continue
                 input_unit = sens.input_units.upper()
                 if input_unit == "M":
                     pass
@@ -56,8 +67,16 @@ def _write_sacpz(inventory, file_or_file_object):
                 elif input_unit in ["M/S**2", "M/SEC**2"]:
                     paz.zeros.extend([0j, 0j])
                 else:
-                    msg = "Encountered unrecognized input units in response: "
-                    raise NotImplementedError(msg + str(input_unit))
+                    msg = "{}.{}.{}.{} {} ".format(
+                        net.code,
+                        sta.code,
+                        cha.location_code,
+                        cha.code,
+                        cha.start_date)
+                    msg += "has unrecognized input units in "
+                    msg += "response: {}. Skipping".format(input_unit)
+                    warnings.warn(msg)
+                    continue
                 out.append("* " + "*" * 50)
                 out.append("* NETWORK     : %s" % net.code)
                 out.append("* STATION     : %s" % sta.code)
--- a/obspy/io/sac/tests/test_sacpz.py
+++ b/obspy/io/sac/tests/test_sacpz.py
@@ -72,7 +72,8 @@ class SACPZTestCase(unittest.TestCase):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter('always')
             inv.write(f, format='SACPZ')
-            self.assertEqual(len(w), 2)
+        self.assertEqual(len(w), 2)
+
         # Only 2 newlines are written.
         self.assertEqual(2, f.tell())
 
--- a/obspy/io/sac/tests/test_sacpz.py
+++ b/obspy/io/sac/tests/test_sacpz.py
@@ -9,6 +9,7 @@ from future.builtins import *  # NOQA
 import os
 import unittest
 import io
+import warnings
 
 import numpy as np
 
@@ -62,6 +63,18 @@ class SACPZTestCase(unittest.TestCase):
         got = [l for l in got.split("\n") if "CREATED" not in l]
         expected = [l for l in expected.split("\n") if "CREATED" not in l]
         self.assertEqual(got, expected)
+
+    def test_write_sacpz_soh(self):
+        path = os.path.join(self.path, '..', '..', 'stationxml', 'tests',
+                            'data', 'only_soh.xml')
+        inv = read_inventory(path)
+        f = io.StringIO()
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            inv.write(f, format='SACPZ')
+            self.assertEqual(len(w), 2)
+        # Only 2 newlines are written.
+        self.assertEqual(2, f.tell())
 
     def test_attach_paz(self):
         fvelhz = io.StringIO("""ZEROS 3
--- a/obspy/io/sac/tests/test_sacpz.py
+++ b/obspy/io/sac/tests/test_sacpz.py
@@ -72,8 +72,12 @@ class SACPZTestCase(unittest.TestCase):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter('always')
             inv.write(f, format='SACPZ')
+        # Testxml has 2 channels: 1 no paz, 2 unrecognized units.
         self.assertEqual(len(w), 2)
-
+        # Assert warning messages contain correct warnings
+        self.assertTrue(any('has no paz' in str(x.message) for x in w))
+        self.assertTrue(any('has unrecognized input units'
+                            in str(x.message) for x in w))
         # Only 2 newlines are written.
         self.assertEqual(2, f.tell())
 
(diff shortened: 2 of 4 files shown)


dsentinel:
Currently `inv.write(f, format='SACPZ')` with fail with exception.
[Here](https://gist.github.com/dsentinel/f8aeec00f42afc9e32e552ae2a8ebbef) is a gist of the current PR.

megies:
I wouldn't use that file format you're writing to in production workflows, but I guess this change can not hurt..

obspy/io/sac/sacpz.py
```diff
@@ -56,8 +62,10 @@ def _write_sacpz(inventory, file_or_file_object):
                 elif input_unit in ["M/S**2", "M/SEC**2"]:
                     paz.zeros.extend([0j, 0j])
                 else:
-                    msg = "Encountered unrecognized input units in response: "
-                    raise NotImplementedError(msg + str(input_unit))
+                    print("{}.{}.{} ".format(net.code, sta.code, cha.code) +\
+                          "has unrecognized input units in " +\
```
megies: please remove those redundant backslashes (see circle CI)

obspy/io/sac/sacpz.py
```diff
@@ -47,7 +47,13 @@ def _write_sacpz(inventory, file_or_file_object):
             for cha in sta:
                 resp = cha.response
                 sens = resp.instrument_sensitivity
-                paz = resp.get_paz()
+                try:
+                    paz = resp.get_paz()
+                except:
```
megies: Please only catch that particular exception that is being raised when no PAZ is found.

dsentinel: I typically do, but it raises a base Exception.

https://github.com/obspy/obspy/blob/b3bc918ee15ff26ae173db86c997adc90154cbcc/obspy/core/inventory/response.py#L1800

dsentinel:
> I wouldn't use that file format you're writing to in production workflows, but I guess this change can not hurt..

You wouldn't use SAC PZ's as a file format? I wouldn't either! ;)
There are, unfortunately, still many sac users and when given a StationXML; Convert to dataless, convert to sacpz.
This is to help legacy users.

obspy/io/sac/sacpz.py
QuLogic: That's not the base, `BaseException` is. While `Exception` is not great, catching it is still preferable as it won't catch stuff like `KeyboardInterrupt`.

obspy/io/sac/sacpz.py
```diff
@@ -47,7 +47,13 @@ def _write_sacpz(inventory, file_or_file_object):
             for cha in sta:
                 resp = cha.response
                 sens = resp.instrument_sensitivity
-                paz = resp.get_paz()
+                try:
+                    paz = resp.get_paz()
+                except Exception:
+                    print("{}.{}.{} has no paz. Skipping.".format(net.code,
+                                                                  sta.code,
+                                                                  cha.code))
```
krischer: I would also print the epoch times of the channel to uniquely identify it. And instead of printing please raise warnings.

obspy/io/sac/sacpz.py
```diff
@@ -56,8 +62,11 @@ def _write_sacpz(inventory, file_or_file_object):
                 elif input_unit in ["M/S**2", "M/SEC**2"]:
                     paz.zeros.extend([0j, 0j])
                 else:
-                    msg = "Encountered unrecognized input units in response: "
-                    raise NotImplementedError(msg + str(input_unit))
+                    msg = "{}.{}.{} ".format(net.code, sta.code, cha.code)
+                    msg += "has unrecognized input units in "
+                    msg += "response: {}. Skipping".format(input_unit)
+                    print(msg)
+                    continue
```
krischer: Same comment as above.

obspy/io/sac/sacpz.py
megies: Oh overlooked this.. yes definitely a warning instead.

dsentinel:
Can I include a test stationxml, like the one in the test case gist above? A quick look didn't show any in the repo with SOH channels.

megies:
Yes you can, but please limit it to what you really need in the test, it should be as small as possible (just a couple of channels, ...). Especially only include response if really needed for the tests as it is huge, usually.

dsentinel:
I think everything's addressed.
I put a stationxml in obspy's stationxml tests/data. It has 2 channels ACE and QEP, it's 5kb.
Thanks guys!

dsentinel:
Travis CI seems to be failing on 2 linux versions with:
`The command "pip install https://github.com/megies/PyImgur/archive/py3.zip" failed and exited with 1 during .`

Appveyor seems to have an unrelated failure:
`FAIL: test_infinite_loop (obspy.io.mseed.tests.test_mseed_special_issues.MSEEDSpecialIssueTestCase)`

obspy/io/sac/tests/test_sacpz.py
```diff
@@ -69,9 +69,9 @@ def test_write_sacpz_soh(self):
                             'data', 'only_soh.xml')
         inv = read_inventory(path)
         f = io.StringIO()
-        with warnings.catch_warnings(record=True) as w:
-            inv.write(f, format='SACPZ')
-            self.assertEqual(len(w), 2)
+        inv.write(f, format='SACPZ')
+        # Only 2 newlines are written.
+        self.assertEqual(2, f.tell())
```
megies: What is wrong with testing the warning? Hhow you were doing it was fragile though.. check out other parts of the code base that test for warning messages (should ideally be refactored in a utility routine at some point..)

```python
+        # check emitted warning message
+        self.assertTrue(len(w) >= 1)
+        msg = ("'components' is expected to be a list of one or more "
+               "component code combinations (e.g. `components=['Z12']` or "
+               "`components=['Z12', '123']`)")
+        for w_ in w:
+            try:
+                self.assertEqual(str(w_.message), msg)
+            except AssertionError:
+                continue
+            else:
+                break
+        else:
+            self.fail(msg='Expected warning message not shown.')
```

obspy/io/sac/sacpz.py
```diff
@@ -47,7 +48,17 @@ def _write_sacpz(inventory, file_or_file_object):
...
-                paz = resp.get_paz()
+                try:
+                    paz = resp.get_paz()
+                except Exception:
+                    msg = "{}.{}.{}.{} {} has no paz. Skipping.".format(
+                        net.code,
+                        sta.code,
+                        cha.location_code,
+                        cha.code,
+                        cha.start_date)
```
megies: no need to keep everything on a separate line..

obspy/io/sac/sacpz.py
```diff
@@ -56,8 +67,16 @@ def _write_sacpz(inventory, file_or_file_object):
...
                     paz.zeros.extend([0j, 0j])
                 else:
-                    msg = "Encountered unrecognized input units in response: "
-                    raise NotImplementedError(msg + str(input_unit))
+                    msg = "{}.{}.{}.{} {} ".format(
+                        net.code,
+                        sta.code,
+                        cha.location_code,
+                        cha.code,
+                        cha.start_date)
```
megies: same a above..

obspy/io/sac/tests/test_sacpz.py
```diff
@@ -63,6 +63,15 @@ def test_write_sacpz_multiple_channels(self):
         expected = [l for l in expected.split("\n") if "CREATED" not in l]
         self.assertEqual(got, expected)
 
+    def test_write_sacpz_soh(self):
+        path = os.path.join(self.path, '..', '..', 'stationxml', 'tests',
+                            'data', 'only_soh.xml')
+        inv = read_inventory(path)
+        f = io.StringIO()
+        inv.write(f, format='SACPZ')
```
megies: This triggers two (expected) warnings. They need to be properly caught and tested/asserted. (Look for `catch_warnings(...)` in our test files)

dsentinel:
Warnings caught.
Squashed and  force pushed.
No rebase as master hasn't moved.

obspy/io/sac/tests/test_sacpz.py
```diff
@@ -63,6 +64,18 @@ def test_write_sacpz_multiple_channels(self):
...
 
+    def test_write_sacpz_soh(self):
+        path = os.path.join(self.path, '..', '..', 'stationxml', 'tests',
+                            'data', 'only_soh.xml')
+        inv = read_inventory(path)
+        f = io.StringIO()
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            inv.write(f, format='SACPZ')
+            self.assertEqual(len(w), 2)
```
megies: Ideally, this should also check the warning message, you might always get some warnings that are something else entirely.

I'm testing warnings like this nowadays: https://github.com/obspy/obspy/blob/06906f389bdaeb5fa4c4bd0cab9f066082e7da42/obspy/io/shapefile/tests/test_core.py#L179-L192

Ideally we should put this in a helper routine in `obspy/core/util/testing.py`..

```python
def assert_warning(warnings, message_regex, category=UserWarning):
    ...
```

dsentinel: Py3 has `assertWarns` (which I originally implemented) and `assertWarnsRegex`, but not py2.
Although this may be remedied with the backport `import unittest2`, if we want to go that route.

I cringe at copy/pasting something from code to tests.
Catching exactly 2 warnings from the test file that has 2 warning conditions seemed sufficient. 
I can assert they are UserWarnings, or they end in `'. Skipping.'` or a full regex.
Anyway, thanks for caring about the code base. I'll proceed as advised.

dsentinel:
Merge-able?

dsentinel:
Thanks @megies @krischer @QuLogic !!

Issue #2193, opened 2018-07-04
Cartopy local projection for south hemisphere. 


seballgeyer:
Hi, 
There is a misbehavior on the local projection option on the South hemisphere using cartopy:

Test code as follow:

```
import matplotlib.pyplot as plt
import obspy
from obspy.clients.fdsn import Client
auspass = Client("http://auspass.edu.au:8080")
station = auspass.get_stations(network='*',station='*',starttime=obspy.UTCDateTime('1950-01-01'),endtime=obspy.UTCDateTime())
station.plot(projection='local', color_per_network=True)
```

**The fix is the following:**
In obspy/imaging/maps.py (line 618 --  620)
```
        proj_kwargs['central_latitude'] = lat_0
        proj_kwargs['central_longitude'] = lon_0
        proj_kwargs['standard_parallels'] = [lat_0,lat_0] #Line to add to have the same behavior than Basemap
        proj = ccrs.AlbersEqualArea(**proj_kwargs)
```

Thanks

Sebastien

seballgeyer:
Hi

The FDSNWS server will be a long term server. We are just starting to import data into it. We will inform FDSN about it in a near future. 

Cheers, 

Sebastien

megies:
Fix looks good to me on a quick glance..

There's still behavior discrepancy in default of station labels and missing lat/lon ticklabels.. but no idea about that, never touched the cartopy versions of the map code..

## Basemap
![basemap](https://user-images.githubusercontent.com/1842780/44259614-4d03b600-a212-11e8-831c-9dacedfffd26.png)

## Cartopy Before
![cartopy](https://user-images.githubusercontent.com/1842780/44259634-5db42c00-a212-11e8-9440-51439c7d4641.png)

## Cartopy After
![cartopy](https://user-images.githubusercontent.com/1842780/44259610-49702f00-a212-11e8-80d2-2e4234a3705a.png)

seballgeyer:
I might be able to provide the modification for the lat Lon ticklabels and the station labels. 
I will provide it early next week. 

Cheers, 

--
Sébastien Allgeyer
Post-doctoral Fellow
Research School of Earth Science
ANU College of Physical and Mathematical Sciences
The Australian National University
Canberra 0200 Australia
sebastien.allgeyer@anu.edu.au

megies:
> The FDSNWS server will be a long term server. We are just starting to import data into it. We will inform FDSN about it in a near future.

[...]

PR #2205, opened 2018-08-17
add FDSNWS at http://auspass.edu.au:8080/ to URL mappings

megies:
### What does this PR do?

Add url mapping for FDSN WS at http://auspass.edu.au:8080/, see #2193

see #2193

### PR Checklist
- [ ] +TESTS:clients.fdsn
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -21,6 +21,7 @@ master:
      have been deselected due to the default location priorities setting. This
      is a pure usability improvement as it has been confusing users
      (see #2159).
+   * Adding mapping for the AUSPASS service (see #2205, #2193)
  - obspy.io.nordic:
    * Add ability to read and write focal mechanisms and moment tensor
      information. (see #1924)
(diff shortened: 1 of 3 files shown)


Jollyfant:
Technically the FDSNWS specification mandates that the services are available on `:80`. It may be that some institutional firewalls block access to another port. We should at least recommend to change it.

megies:
> Technically the FDSNWS specification mandates that the services are available on `:80`. It may be that some institutional firewalls block access to another port. We should at least recommend to change it.

CC @seballgeyer

seballgeyer:
Hi Tobias,

We had that discussion with IRIS's people last week.  We are investigating
the possibilities. It can be a good idea to stop this pull until we have a
solution on our end.

-- Sebastien

megies:
OK, let's close this PR then, let us know when there are any definitive news, we can add it on top of maintenance branch then..

PR #2210, opened 2018-09-01
Catalog dict conversions

d-chambers:
### What does this PR do?

Adds a module to obspy.core.event named "dict" which contains functions for converting catalogs to and from dictionaries. 

There is some discussion in #1910 on the topic. To recap, this is a first step towards:

1. A viable json (or json-like) format for losslessly reading, and writing catalogs. This could be better IO format than quakeML for some applications. 
2. Enabling other event formats to use an intermediate `dict` for reading/writing. 

### PR Checklist
- [X] All tests still pass.

### How it works

The `catalog_to_dict` function recursively converts obspy objects to dictionaries. In most cases, the dictionary created from an object can be passed to the class' `__init__` method as `kwargs` to create a new and equal instance of the same class. This ensures only public attributes are used in order to better future-proof the serialization scheme. There are a few classes, such as `obspy.UTCDateTime` where a custom serialization function is needed because the the first approach does not work. In `UTCDateTime`'s case the `str` representation is used.

The `dict_to_catalog` function uses a map of attribute names to their appropriate classes to recurse the dictionary and construct the class tree. 

### What it still needs

More tests cases to ensure it works with all valid catalog objects.


obspy/core/tests/test_dict_conversion.py
```diff
@@ -0,0 +1,43 @@
...
+                        unicode_literals)
+from future.builtins import *  # NOQA
+
+import unittest
+
+import obspy
+from obspy.core.event.dict import catalog_to_dict, dict_to_catalog
+from obspy.core.util.testing import create_diverse_catalog
+
+TEST_CATALOGS = [obspy.read_events(), create_diverse_catalog()]
```
calum-chamberlain: I wonder if there is a simple way to get all the catalogs that the different `io` modules use for their tests and test them?

d-chambers: I am not sure if there is a simple way but it wouldn't be too hard to manually specify their test data paths. Might be worth adding something to the core.util.testing module for easily collecting all of them.

megies:
I don't have time to look at this in detail yet.. but any chance this can be compatible with #752?

Only comment I have right now is that naming the file `dict.py` is probably not a good idea. Builtin names are scary and can bite. But obviously that's easy to fix should this get a nod..

d-chambers:
I agree on the name but couldn't come up with something better, suggestions? Of course the three hard problems in computer science are naming variables and off by one errors.

I will take a look at #752 and see what I can do this weekend.

d-chambers:
> naming the file dict.py is probably not a good idea

Change the name to "dictionary" (I know, very creative :smile: )

>  any chance this can be compatible with #752?

Looking at it closer, #752 is doing something quite different than this PR. This PR adds a function to convert a obspy catalog tree to a "json-able" strucutre and back in a loss-less way (for storage/transmission). IMHO the catalog dict and `Catalog` objects should be identical in their structure and attribute names, otherwise the conversion is not "obvious". #752, on the other hand, is about implementing a property that other programs can use to extract spatial information on a subset of objects in the catalog hierarchy. The use cases are really not the same.

megies:
>> naming the file dict.py is probably not a good idea
> 
> Change the name to "dictionary" (I know, very creative smile )

Actually, as an I/O plugin it might be good to move it to `io/ascii/` anyway. It could go into a new file `event.py` there?

> This PR adds a function to convert a obspy catalog tree to a "json-able" strucutre

Oh, so it's not really de/serialization to JSON then, just the preparation for it, I guess? Then the above comment does not really fit, I guess..

calum-chamberlain:
I pulled this branch and added a fairly revolting way to get all the test catalogs into the `obspy.core.utils.testing.py` file. I collects 96 catalogs and they all pass the current test @d-chambers wrote.  I'm loath to push those changes up here just because they are so ugly, I could create a PR onto this branch?

The function is:
```python
def read_test_catalogs(warn=False):
    """
    Read all io test catalogs

    Will try and read everything in the test datasets, regardless of whether
    this can or should be read in.  Errors are ignored, unless warn=True. If
    warn=True then errors are reported as UserWarning.

    :return: list of catalogs
    """
    from obspy import read_events
    catalogs = []
    # This is revolting.
    base = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(
        os.path.abspath(inspect.getfile(inspect.currentframe()))))))
    eps = _get_ordered_entry_points(
        'obspy.plugin.event', 'readFormat', EVENT_PREFERRED_ORDER)
    for name, ep in eps.items():
        test_base = os.path.join(
            base, *ep.module_name.split(".")[:3], "tests", "data")
        valid_files = glob.glob(os.path.join(test_base, "*"))
        for x in valid_files:
            with WarningsCapture():
                try:
                    catalogs.append(read_events(x, name))
                except Exception as e:
                    if warn:
                        UserWarning(e)
                    pass
    return catalogs
```

d-chambers:
Actually that doesn't look all that bad. I am glad the tests pass! A way to collect all tests files would certainly be useful. How many files does it collect and fail to read? Any reason for not using `__file__` rather messing with the `inspect` module? It might be useful to `yield` each catalog to minimize memory usage. Maybe make it general so we could do something similar to collect all waveforms or inventories as well?

I have no problem if you want to push changes directly to this branch.  

I should apologize, the code in this PR is a bit complex (hopefully not complicated) because there is a fair bit of recursion. It is the most elegant solution I could come up with, but I know it can be a bit tricky to think through.

megies:
Putting this in the 1.2.0 milestone so it doesnt get overlooked.. when we eventually get round to doing it.. -_-

calum-chamberlain:
Thanks for those comments @d-chambers!
> How many files does it collect and fail to read?

It collects 130 files and reads 98 of them.  I know some tests have both event and waveform files in their `tests/data` folder (seiscomp for one I think), and/or files that are known to fail (nordic has one of those), so its not surprising that not everything it finds is read. Also, all the other tests we have for those IO formats should take care of reading issues, so I'm keen to just ignore/overlook them here.

> Any reason for not using __file__ rather messing with the inspect module?

I thought that `__file__` wasn't super robust for getting paths, in the past I have been advised to use the `inspect` and `abspath` methods, but in this case (or at least the way I have done it) is really ugly.  If `__file__` is safe then I would rather use that for "prettiness". @megies do you have an opinion on that?

> It might be useful to yield each catalog to minimize memory usage. 

Good suggestion, I will try that.

> Maybe make it general so we could do something similar to collect all waveforms or inventories as well?

Also good suggestion, I will try to do that too.

> I should apologize, the code in this PR is a bit complex (hopefully not complicated) because there is a fair bit of recursion. It is the most elegant solution I could come up with, but I know it can be a bit tricky to think through.

It looks good to me, recursion is the way I would have done it, but it can be hard to follow.

calum-chamberlain:
I generalised and used a `yield`.  I'm not sure how people feel about using the plugin read functions directly or using the `read` and `read_events` functions (which is what I'm doing at the moment)?

megies:
> If `__file__` is safe then I would rather use that for "prettiness". @megies do you have an opinion on that?

Looks like the `__file__` is gone anyways now? In any case, since it was in the testing routines only, no harm from it i'd say. worst thing that could ever happen there are failing tests?

d-chambers:
@calum-chamberlain what do you think is left to do on this? Is it ready for review?

calum-chamberlain:
I think it is probably ready for review - it might be nice to change back to using `__file__` rather than my horrible thing, but otherwise I think it could be okay. Would be great to have this in soon.

d-chambers:
@calum-chamberlain have you pushed all the work you have done on this branch? I am going rebase this soon, and add make a few more tweaks. After that we should see if we can't get this merged and closed.

calum-chamberlain:
Yup, nothing more from me since the last push. Rebase away!

d-chambers:
Ok, hopefully I did that right :man_shrugging:

ThomasLecocq:
the only failing test that looks worrying is for py3.5 on Linux:
```
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/obspy/core/tests/test_dict_conversion.py", line 41, in test_roundtrip_conversion
    self.assertEqual(cat1, cat2)
AssertionError: <obspy.core.event.catalog.Catalog object at 0x7f88c26524e0> != <obspy.core.event.catalog.Catalog object at 0x7f88c23d6cf8>
```
as I think the failure on windows (evalresp) is nor related, is it ?

megies:

d-chambers:
> as I think the failure on windows (evalresp) is nor related, is it ?
Nope, it shouldn't be using that at all.

ThomasLecocq:
@megies only the classic evalresp appveyor issue, otherwise all green

krischer:
I looked at this for a bit and I think we should push this to the next release to not commit to a solution we might not want to maintain in the long run. In general I really like the idea in this PR and I think we should go down that route but I feel like it should be implemented in a more general way.

A few thoughts I had:

* This should be integrated into `obspy.io.json`. The serialization to JSON/dict is actually identical in both so the `catalog_to_dictionary()` function should be use there to not have two implementations which might diverge in the future.

* Instead of dealing with the serialization for all types of objects in a single file/function we should rather define some interfaces. I think a `__json__()` method might be nice - it not a fully standard Python interface but its has been used.

* Almost all the event classes actually inherit from `AttribDict` so a lot of functionality could be collected there.

* The `__init__()` methods of the event objects should be smart enough to work with nested dictionary structures und convert to the proper event objects in the init function, e.g. `Catalog(**nested_dict)` should IMHO work. The main issue here (which this PR also mainly deal with) is how to map keys in the dictionary to Python classes. Maybe we can define some kind of interface for that so it would work for all kinds of objects?

* This is a nice library and we could draw some inspiration from it: https://jsons.readthedocs.io/en/latest/?badge=latest After we release ObsPy 1.2 we'll move to Python 3 so we could consider employing type hints.

d-chambers:
@krischer sounds good. I agree there is room for improvement in both the interface and implementation.

calum-chamberlain:
Where we going to push ahead with this, or did this all end up going into obsplus @d-chambers ?

d-chambers:
It's all in ObsPlus now but we can easily migrate it to ObsPy if we are ok with taking pydantic as a dependency. It would also require some refactoring of the tests as they are written heavily in pytest style. 

This could snowball as I think doing it properly could require refactoring all catalog-related objects to pydantic models.

calum-chamberlain:
Yeah, I would lean towards leaving this in obsplus - do you want to close this then?

d-chambers:
Good idea

PR #2218, opened 2018-09-25
Fix get_gaps overlapping

Brtle:
The [get_gaps](https://github.com/obspy/obspy/blob/master/obspy/core/stream.py#L745) method isn't working as expected if a trace is completely overlapping another trace. The `starttime` of the gaps is the `endtime` of the first trace even if the `endtime` of the second trace is previous.

This little PR just tries to fix that.


Brtle:
> Just add a changelog one-liner please.

Done!

PR #2220, opened 2018-09-25
Bug comparing ComparingObject to an object without __dict__

dsentinel:
### Issue:
Comparing `ComparingObjects` to random objects with `==` / `__eq__` should return `False` not raise and excption.
### Actual case:
If you have 2 `Channels` one with a valid `Equipment` and the other with `None`
`chan1 == chan2` 
raises 
```
  File "/Users/lloyd/code/obspy/obspy/core/util/base.py", line 519, in __ne__
    return not self.__eq__(other)
  File "/Users/lloyd/code/obspy/obspy/core/util/base.py", line 516, in __eq__
    return self.__dict__ == other.__dict__
AttributeError: 'int' object has no attribute '__dict__'
```
The first commit only has the tests, you can pull and exersize.
### Solution:
At first I was worried about this solution breaking comparisons with other types that were compatible, e.g. `CustomComplex` == `complex`, but those kinds of types don't inherit from `ComparingObject`, and those seem to inherit from their compatible types... and the tests passed.

### DIFF
--- a/obspy/core/tests/test_util_base.py
+++ b/obspy/core/tests/test_util_base.py
@@ -4,13 +4,14 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 
 import os
+import copy
 import shutil
 import unittest
 
 from obspy.core.compatibility import mock
 from obspy.core.util.base import (NamedTemporaryFile, get_dependency_version,
                                   download_to_file, sanitize_filename,
-                                  create_empty_data_chunk)
+                                  create_empty_data_chunk, ComparingObject)
 from obspy.core.util.testing import ImageComparison, ImageComparisonException
 
 import numpy as np
@@ -139,6 +140,19 @@ class UtilBaseTestCase(unittest.TestCase):
         self.assertIsInstance(out, np.ma.MaskedArray)
         self.assertEqual(out.dtype, np.float32)
         np.testing.assert_allclose(out.mask, [True, True, True])
+
+    def test_comparing_object_eq(self):
+        co = ComparingObject()
+        # Compare to other types
+        self.assertNotEqual(co, 5)
+        self.assertNotEqual(co, None)
+        self.assertNotEqual(co, object())
+        # Compare same type, different instance, with attributes
+        co.at = 3
+        deep_copy = copy.deepcopy(co)
+        self.assertEqual(co, deep_copy)
+        deep_copy.at = 0
+        self.assertNotEqual(co, deep_copy)
 
 
 def suite():
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -16,6 +16,7 @@
      creating a string representation (see #2119 and #2127).
    * Fix Stream.get_gaps() when a trace is completely overlapping another trace
      (see #2218).
+   * Fig Exception when comparing ComparingObjects (see #2220).
  - obspy.clients.arclink:
    * Raise a warning at import time that the ArcLink protocol will be
      deprecated soon (see #1987).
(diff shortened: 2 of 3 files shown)


dsentinel:
Is there a python3 PR tag or `#TODO: comment` for refactors of code that can be cleaner when not supporting python2. I come across these and feel like I should do something while there in my head.

obspy/core/tests/test_util_base.py
```diff
@@ -4,13 +4,14 @@
...
 import os
+import copy
 import shutil
 import unittest
 
 from obspy.core.compatibility import mock
 from obspy.core.util.base import (NamedTemporaryFile, get_dependency_version,
                                   download_to_file, sanitize_filename,
-                                  create_empty_data_chunk)
+                                  create_empty_data_chunk, ComparingObject)
```
QuLogic: This sorts at the very beginning of the `import` list.

dsentinel: Thanks for the review.
Sorry, I don't understand what this comment means. Are you suggesting a change?

megies:
> Is there a python3 PR tag or `#TODO: comment` for refactors of code that can be cleaner when not supporting python2. I come across these and feel like I should do something while there in my head.

I'd propose to just keep a list of things in a sepcial ticket, see #2228

obspy/core/tests/test_util_base.py
megies: I don't think we ever bothered to enforce alphabetical order *inside* import attribute lists, we just enforce alphabetical order in the module names being imported from. So if this is what @QuLogic meant I think it can stay as is.

megies:
Hmm.. dunno, maybe this should even be considered a bug and go into maintenance instead.. It would be really weird if we were breaking somebody's code by not raising an exception anymore in a place were none is expected to raise, IMHO

obspy/core/tests/test_util_base.py
dsentinel: oh, gotcha, thanks!

dsentinel:
> > Is there a python3 PR tag or `#TODO: comment` for refactors of code that can be cleaner when not supporting python2. I come across these and feel like I should do something while there in my head.
> 
> I'd propose to just keep a list of things in a special ticket, see #2228

Thats good a good idea. Works well for general changes.
I was thinking more along the lines of specific code changes, in the code, or with line numbers
like these lines should look like this after py2 drop.
...
Perhaps just a new branch would be the way to go.

dsentinel:
> Hmm.. dunno, maybe this should even be considered a bug and go into maintenance instead.. It would be really weird if we were breaking somebody's code by not raising an exception anymore in a place were none is expected to raise, IMHO

I definitely consider this a bug.

I pull from master as the maintenance branch seemed to have fallen behind.
I can rebase from master if appropriate. I need this to in a project to publish, so I'd like to see it in 1.1.0 or at least master as soon as can be.

I can't imagine anyone relying on this behavior.

megies:
> I pulled from master as the maintenance branch seemed to have fallen behind.
> I can rebase from master if appropriate. I need this to in a project to publish, so I'd like to see it in 1.1.0 or at least master as soon as can be.

See branching model, changes in maintenance branch are bug fix only and will end up in release version 1.1.1 eventually, master branch changes will end up in release 1.2.0. If it is merged into maintenance branch it can be right away also in master branch, we regularly merge maintenance into master anyway and if there is a need to have it in master we can do it any time.

dsentinel:
rebased and PR'ed against `maintenance_1.1.x`

megies:
Although it might not affect most people, maybe it'd still be good to have a one line changelog message, just for the sake of completeness.

Otherwise good to merge. :+1:

megies:
>  I'm not allowed to merge :/

Hmm.. but it has approving reviews.. dunno, but I'll merge it anyways now. ;-)

dsentinel:
Thanks!

PR #2229, opened 2018-10-05
PPSD: fix/improve time overlap check

megies:
### What does this PR do?

Fixes the time check in PPSD that is done when determining if a given psd slice should be added to the PPSD or if it's representing a time span that already has been processed / overlaps too much with already present data.

Right now sometimes valid data that should go into PPSD are rejected. This e.g. happens when processing gapless data for two consecutive days in separate PPSD instances and storing each day in separate npz files. When loading/adding the two npz data files together in one PPSD instance (e.g. for later plotting), some time slices at the day border were rejected because they overlap, but that overlap is totally fine and actually what the PPSD instance does on processing (unless no overlap was specified during PPSD init).

### PR Checklist
- [ ] All tests still pass.

CC @ThomasLecocq I know you use PPSD so you might want to know about this PR.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -53,6 +53,8 @@ master:
      integer nanosecond POSIX timestamps to avoid any potential floating point
      inaccuracies and since this is also what UTCDateTime is based on nowadays
      (see #2045)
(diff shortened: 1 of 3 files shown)


megies:
Test fails unrelated.

Issue #2222, opened 2018-09-26
read_nlloc_hyp issue when phase hour and minute is '2400' 


ChrisBail:
Hi,
obspy version: >1.1
platform: linux
install: frrom source

issue in `io.nlloc.core.read_nlloc_hyp` line (334)

When reading a nlloc phase file the problem arises when the phase hour and time is '2400', `UTCDateTime.strptime` can't handle it.
Proposed fix:
```
date, hourmin, sec = map(str, line[6:9])
sec=float(sec)
 if hourmin=='2400': 
      hourmin='2359'
      sec=sec+60
t = UTCDateTime.strptime(date + hourmin, "%Y%m%d%H%M") + sec
```
I hope this helps and can be integrated in the future,

Ch.

d-chambers:
Hey @ChrisBail, good find. Would you mind writing a test that captures this behavior? If you are comfortable putting it in obspy [this](https://github.com/obspy/obspy/blob/master/obspy/io/nlloc/tests/test_core.py) would probably be the right place.  If not, you could attach a code sample here that reproduces the issue and I will make sure it gets included in the test suite. Once we have a test that fails then we can work on a fix for it.

It would be great if UTCDateTime could handle this directly without having to put this kind of logic in all the format parsers where this can crop up. I have seen this several times in other formats and I have usually resorted to ad-hoc checks as well.

megies:
I'd consider this a broken time specification, but yeah, I guess we should allow for it. Actually, looking up ISO standard, it even explicitly mentions the 24:00:00 case: https://en.wikipedia.org/wiki/ISO_8601#Times

So, yeah, I think we should just add it as a special allowed case in `UTCDateTime.__init__` directly (even though `datetime.datetime` does not allow this special case). Other opinions?

megies:
> a special allowed case in `UTCDateTime.__init__` directly

we'd also have to patch `strptime`, I guess..

ChrisBail:
Thanks Derrick & Tobias for the quick reply, I would agree with Tobias to put that directly in UTCDateTime instead of doing it in read_nlloc_hyp. I  know for sure that NLLOC and SEISAN (nordic files) use the 24:00 convention.

megies:
Just to make sure.. do NLLOC and Seisan only produce 24:00:00 or does it also happen with additional hours or seconds?

d-chambers:
From line 4 in seisan:
[...]

PR #2232, opened 2018-10-08
Time overflow

d-chambers:
### What does this PR do?
Allows for hour, minute, and second values to exceed the normal limits of 23, 59, 59 when creating a `UTCDateTime` object from key word arguments or using the `replace` method.

This change was discussed in #2222. It will allow several format parsers to favor using this approach when dealing with time values that may exceed their normal limits, rather than adding ad-hoc checks/handling logic in each parser. 

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,9 +7,9 @@ master:
    * UTCDateTime objects now issue depreciation warnings when setting any
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
-   * UTCDateTime objects can now accept hour, minute, and second values
-     greater than their normal limits by setting the strict keyword argument
-     to False (see #2232).
+   * UTCDateTime objects can now accept hour, minute, second, and microsecond
+     values greater than their normal limits by setting the strict keyword
+     argument to False (see #2232).
    * Added replace method to UTCDateTime class (see #2077).
    * Added remove method to Inventory class (see #2088).
    * Added id property to WaveformStreamID (see #2131).
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -8,7 +8,8 @@ master:
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
    * UTCDateTime objects can now accept hour, minute, and second values
-     greater than their normal limits when initialized with kwargs (see #2232)
+     greater than their normal limits by setting the strict keyword argument
+     to False (see #2232).
    * Added replace method to UTCDateTime class (see #2077).
    * Added remove method to Inventory class (see #2088).
    * Added id property to WaveformStreamID (see #2131).
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,8 @@ master:
    * UTCDateTime objects now issue depreciation warnings when setting any
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
+   * UTCDateTime objects can now except hour, minute, and second values
+     greater than their normal limits when initialized with kwargs (see #2232)
    * Added replace method to UTCDateTime class (see #2077).
    * Added remove method to Inventory class (see #2088).
    * Added id property to WaveformStreamID (see #2131).
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,7 +7,7 @@ master:
    * UTCDateTime objects now issue depreciation warnings when setting any
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
-   * UTCDateTime objects can now except hour, minute, and second values
+   * UTCDateTime objects can now accept hour, minute, and second values
      greater than their normal limits when initialized with kwargs (see #2232)
    * Added replace method to UTCDateTime class (see #2077).
    * Added remove method to Inventory class (see #2088).
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1457,6 +1457,15 @@ class UTCDateTimeTestCase(unittest.TestCase):
         utc = base_utc(second=120)
         self.assertEqual(utc, base_utc(minute=2))
 
+    def test_hour_minute_second_overflow_with_replace(self):
+        """
+        The replace method should also support the changes described in #2222.
+        """
+        utc = UTCDateTime('2017-09-18T00:00:00')
+        self.assertEqual(utc.replace(hour=25), utc + 25 * 3600)
+        self.assertEqual(utc.replace(minute=1000), utc + 1000 * 60)
+        self.assertEqual(utc.replace(second=60), utc + 60)
+
 
 def suite():
     return unittest.makeSuite(UTCDateTimeTestCase, 'test')
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1448,26 +1448,34 @@ class UTCDateTimeTestCase(unittest.TestCase):
         kwargs = dict(year=2017, month=9, day=18, hour=0, minute=0, second=0)
         base_utc = partial(UTCDateTime, **kwargs)
         # ensure hour can exceed 23 and is equal to the day ticking forward
-        utc = base_utc(hour=25)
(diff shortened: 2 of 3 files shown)


d-chambers:
+DOCS

CHANGELOG.txt
```diff
@@ -7,6 +7,8 @@ master:
    * UTCDateTime objects now issue depreciation warnings when setting any
      attributes outside of init, or comparing UTCDateTime objects with
      different precisions (see #2077).
+   * UTCDateTime objects can now except hour, minute, and second values
```
QuLogic: s/except/accept/ (though technically the former is valid in this case, I'm not sure it's what you meant)

obspy/core/utcdatetime.py
```diff
@@ -383,8 +388,29 @@ def __init__(self, *args, **kwargs):
...
+                # keep track of second equiv. of overflow
+                extra_seconds += (kwargs['hour'] // 24) * 3600 * 24
+                extra_seconds += (kwargs['minute'] // 60) * 60 * 60
+                extra_seconds += (kwargs['second'] // 60) * 60
+                # reduce value in kwargs to be within normal bounds
+                kwargs['hour'] %= 24
+                kwargs['minute'] %= 60
+                kwargs['second'] %= 60
+            # get the correct timestamp, add extra seconds
+            self._ns = (UTCDateTime(**kwargs) + extra_seconds).ns
```
QuLogic: Do we really need the extra object? Set `dt` here and `extra_seconds = 0` in the alternative case.

CHANGELOG.txt
d-chambers: right, this was a typo, thanks.

obspy/core/utcdatetime.py
d-chambers: 2 things here:

1. With the try/except configuration as it is now adding this new behavior will not induce a performance hit when the normal dt construction works. Since `UTCDateTime` is used extensively I would be reluctant to add overhead to its `__init__` in all cases.

2. We need to avoid changing the `ns` property once it is set, else a depreciation warning will be issued by [this method](https://github.com/obspy/obspy/blob/befecf33ab2a1e2b5ceb4f5b0617cc37e055822b/obspy/core/utcdatetime.py#L1171-L1179. )

d-chambers:
@QuLogic thanks for the review!

megies:
> Allows for hour, minute, and second values to exceed the normal limits of 23, 59, 59 when creating a `UTCDateTime` object from key word arguments or using the `replace` method.

Hmm, @d-chambers I think there might have been a bit of an misunderstanding in #2222. What I wanted to propose there is to leave `UTCDateTime.__init__()` as is (i.e. keep raising exceptions on overflow in hour, minute etc., which is in line with builtin `datetime` behavior) or at most change it so it allows one special case for exactly `24:00:00.0` (which is allowed by [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Times)). I wanted to suggest to only add an internal helper routine to generate `UTCDateTime`s from overflowing kwargs in specific situations in I/O submodules.

Personally, I think we should not promote abusing `UTCDateTime` with input that is not ISO conform. It might obscure bugs in user land codes in my opinion.

But I'm ready to put this to a vote of course.

megies:
.. or another option might be to leave it in `UTCDateTime.__init__()` like you have right now but only switch it on explicitly on request with something like `UTCDateTime.__init__(..., strict=True)` (or `loose=False`, or `sloppy=False`, or ...)



> I won't have time to do this for a few days.

sure no worries

d-chambers:
Ya, I like the idea of a 'strict' keyword. I may still move most the code to another method though to cleanup the init (it is getting quite long)

d-chambers:
Ok @megies, take a look at see if this is more to your liking. I am still trying to leave the overflow behavior in a try/except clause in order to not slow down the `__init__` when the overflow logic is not needed.

obspy/core/utcdatetime.py
```diff
@@ -383,8 +396,36 @@ def __init__(self, *args, **kwargs):
...
+        else:
+            self._from_datetime(dt)
+
+    def _handle_overflow(self, year, month, day, hour=0, minute=0, second=0,
+                         microsecond=0):
+        """
+        Handles setting date if an overflow of usual value limits is detected.
+        """
+        kwargs = locals()
+        kwargs.pop('self')
```
megies: I'm not at all comfortable with using `locals()` here. It can be utterly polluted with other stuff e.g. in debugger sessions etc. Plus there is no need at all for beeing magicky here, we can just be simple, explicit and stable.

Explicit is better than implicit. ;-)

```
kwargs = dict(
    year=year, month=month, day=day, hour=hour, minute=minute,
    second=second, microsecond=microsecond)
```

Or just put all of them variables in the UTCDateTime call at the end?

obspy/core/utcdatetime.py
```diff
@@ -383,8 +396,36 @@ def __init__(self, *args, **kwargs):
...
+        # ensure hour minute second are in kwargs
+        extra_seconds = 0
+        # keep track of second equiv. of overflow
+        extra_seconds += (kwargs['hour'] // 24) * 3600 * 24
+        extra_seconds += (kwargs['minute'] // 60) * 60 * 60
+        extra_seconds += (kwargs['second'] // 60) * 60
```
megies: I don't really like dropping decimal parts silently here. The user should know when they are providing bad/unexpected input.

Since this is basically a drop-in replacement method for `datetime.datetime(*args, **kwargs)`, I would propose to either..

 - be strict with types, i.e. raise `TypeError`s if e.g. `hour` is not an `int` like `datetime` does.. (guess this would be semi strict then)
 - or accept floats but don't silently drop the decimal part like done right now (ultra sloppy mode)

Personally I would go for the first option.

obspy/core/utcdatetime.py
d-chambers: Sure no problem.

obspy/core/utcdatetime.py
d-chambers: Actually the purpose of the int division here is not to drop the decimal parts, but simply to get the second equivalent of overflow (if hour, minute, or second exceed their natural bounds).  Looking at it again, I can see this is  overly complicated, I will simplify it.

Option one is already implemented. This is because the values are given to `datetime.datetime` (line 401) which will raise a `TypeError` if a float for any of those values is passed. Since we are only catching `ValueError` to determine if the overflow logic should kick in, any `TypeError`s will still propagate up the stack.

d-chambers:
Ok,
The Travis failures are related to our image comparison tests. Not sure about the AppVeyor failure.

megies:
Restarted Appveyor, build glitched out it seems.

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1437,6 +1438,48 @@ def test_unsupported_replace_argument_raises(self):
...
+        self.assertIn('hour must be in', str(e.exception))
+
+    def test_hour_minute_second_overflow_with_replace(self):
+        """
+        The replace method should also support the changes described in #2222.
+        """
+        utc = UTCDateTime('2017-09-18T00:00:00')
+        self.assertEqual(utc.replace(hour=25, strict=0), utc + 25 * 3600)
+        self.assertEqual(utc.replace(minute=1000, strict=0), utc + 1000 * 60)
+        self.assertEqual(utc.replace(second=60, strict=0), utc + 60)
```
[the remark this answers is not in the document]
d-chambers: Right you are, I was probably just being lazy or trying to avoid a line wrap.

obspy/core/utcdatetime.py
```diff
@@ -383,8 +396,31 @@ def __init__(self, *args, **kwargs):
...
+        """
+        # Keep track of seconds due to hour, minute, second
+        seconds = 0
+        seconds += hour * 3600
+        seconds += minute * 60
+        seconds += second
+        # Init UTCDateTime based on year, month, day, add seconds
+        utc_base = UTCDateTime(year=year, month=month, day=day)
+        # Add seconds and set nanoseconds on self
+        self._ns = (utc_base + seconds + microsecond / 1000000).ns
```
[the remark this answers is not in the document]
d-chambers: right, added documentation

megies:
Test fails are unrelated, so merging

PR #2234, opened 2018-10-08
Warning capture fix

### LINKED ISSUE #2233: WarningsCapture raises TypeError on warnings.warn
The `WarningsCapture` class in `obspy.core.util.testing` raises a TypeError when used with the `warnings.warn` function. For example:

```python
import warnings

from obspy.core.util.testing import WarningsCapture

with WarningsCapture() as w:
    warnings.warn('something bad is happening')
```
This occurs because the `category` parameter can be none, but is not properly checked for if the `message` parameter is not an instance of `Warning`.

d-chambers:
### What does this PR do?
fixes #2233 where the internal testing utility for capturing warnings reliably across python versions did not work for `warnings.warn`. 

#2233

Note: since this a small change to an internal tool I decided to base of master. Happy to change it if needed.

### DIFF
--- a/obspy/core/util/testing.py
+++ b/obspy/core/util/testing.py
@@ -728,8 +728,7 @@ class WarningsCapture(object):
                                     category=category,
                                     filename="", lineno=0))
 
-    def _warn(self, message, category=None, *args, **kwargs):
-        category = Warning if category is None else category
+    def _warn(self, message, category=Warning, *args, **kwargs):
         if isinstance(message, Warning):
             self.captured_warnings.append(
                 warnings.WarningMessage(
--- a/obspy/core/util/testing.py
+++ b/obspy/core/util/testing.py
@@ -729,6 +729,7 @@ class WarningsCapture(object):
                                     filename="", lineno=0))
 
     def _warn(self, message, category=None, *args, **kwargs):
+        category = Warning if category is None else category
         if isinstance(message, Warning):
             self.captured_warnings.append(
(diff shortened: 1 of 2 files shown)


obspy/core/util/testing.py
```diff
@@ -729,6 +729,7 @@ def _warn_explicit(self, message, category, *args, **kwargs):
                                     filename="", lineno=0))
 
     def _warn(self, message, category=None, *args, **kwargs):
+        category = Warning if category is None else category
```
megies: Any reason to not do `def _warn(self, message, category=Warning, *args, **kwargs):`?

d-chambers: Mainly that the class definition is mutable and it can be better to avoid mutable default args. Probably wouldn't hurt anything here though.

d-chambers: ok, changed it.

d-chambers: Ya I think it will be fine in this case, and it does make the code a bit cleaner.

megies:
I rebased and force-pushed to get rid of the merge commit.

d-chambers:
Thanks @megies,
I forgot to rebase rather than pull to avoid that extra commit. Sorry about that.

krischer:
Thanks for this!

megies:
>  I forgot to rebase rather than pull to avoid that extra commit. Sorry about that.

no worries, quick fix

Issue #2055, opened 2018-01-19
Documentation for trace.detrend


Keowne:
Hi there, just following up from recent discussion at the end of the Pull Request "higher order detrend function #1173". [Am creating a new Issue rather than new Pull request as unfortunately I don't have the time to figure that process out right now and simply create havoc with git at the best of times.] This is for ObsPy 1.1.0 by the way.

So in short, for the documentation regarding the obspy.core.trace.Trace.detrend, "**options" isn't explained in the same way that it is on say, for obspy.core.trace.Trace.filter, and there's no equivalent example usage. So documentation-wise, I suppose incorporating something like the following into obspy.core.trace.Trace.detrend would help:

```
:param options: Necessary keyword arguments for the respective detrend type
            that will be passed on. (e.g. ``order=2``, ``dspline=100`` for ``"spline"``)

.. rubric:: Example

>>> from obspy import read
>>> st = read()
>>> st.detrend('spline',order=2,dspline=100,plot=True)

.. plot::

from obspy import read
st = read()
st.detrend('spline',order=2,dspline=100,plot=True)

```
The reason for using st.detrend as opposed to st[0].detrend or tr.detrend is simply to illustrate that you can apply the process to the whole stream and not just on a trace by trace basis.

Many thanks.

megies:
closed by above PR

PR #2237, opened 2018-10-09
add example and improve stream detrend docs

d-chambers:
### What does this PR do?

Tries to improve the docstring for `obspy.core.stream.Stream.detrend` method by rewording slightly and adding an example.

see #2055

+DOCS

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -2422,6 +2422,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         .. rubric:: Example
 
         Apply an order 3 spline detrend on all traces contained in the stream.
+
         >>> from obspy import read
         >>> st = read()
         >>> st.detrend("spline", order=3, dspline=500)
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -2408,6 +2408,10 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         :meth:`~obspy.core.trace.Trace.detrend` method of
         :class:`~obspy.core.trace.Trace`.
 
+        :param options: collects keyword arguments which are passed to the
+            trace method call. Does not need to be used directly.
+            See the example that follows.
+
          .. note::
 
             This operation is performed in place on the actual data arrays. The
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -2404,9 +2404,25 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         """
         Remove a trend from all traces.
 
-        For details see the corresponding
+        For details on supported methods and parameters see the corresponding
         :meth:`~obspy.core.trace.Trace.detrend` method of
         :class:`~obspy.core.trace.Trace`.
+
+         .. note::
+
+            This operation is performed in place on the actual data arrays. The
+            raw data will no longer be accessible afterwards. To keep your
+            original data, use :meth:`~obspy.core.stream.Stream.copy` to create
+            a copy of your stream object.
+
+        .. rubric:: Example
+
+        Apply an order 3 spline detrend on all traces contained in the stream.
+        >>> from obspy import read
+        >>> st = read()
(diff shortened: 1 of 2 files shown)


megies:
Thanks! Would you mind adding the `:param options:` as proposed by OP in #2055, I think he has a point there.

trichter:
The code block does not render correctly.
Here is a link to the relevant [doc](http://docs.obspy.org/pull-requests/2237/packages/autogen/obspy.core.stream.Stream.detrend.html#obspy.core.stream.Stream.detrend).

d-chambers:
I cherry-picked the relevant commits onto master to get rid of the merge commit. When docs are done building this should be ready to review/merge if the example renders correctly.

megies:
Actually, on a second look, the detail descriptions and example should probably go to the API docs of `Trace.detrend()`, I think. For all methods on Stream that are basically just doing a `for tr in self.traces: ...`, we always have the Stream docs pretty much empty, pointing to the corresponding Trace method, mainly to avoid duplicated maintenance efforts (which are also prone to get broken on changes).

d-chambers:
@megies, sure, I don't mind moving the changes to `trace.detrend`. I think we should keep the in-place warning on `Stream` though. Also, there are other stream methods with fairly extensive docs/examples such as `attach_response`, `normalize`, `std`, `interpolate`, `decimate`, etc. Should we look and pushing most of the detail in the docs for those methods to their corresponding `trace` methods?

krischer:
> @megies, sure, I don't mind moving the changes to `trace.detrend`. I think we should keep the in-place warning on `Stream` though. Also, there are other stream methods with fairly extensive docs/examples such as `attach_response`, `normalize`, `std`, `interpolate`, `decimate`, etc. Should we look and pushing most of the detail in the docs for those methods to their corresponding `trace` methods?

I wanna say yes. I think for a proper solution we should try to find some automated way to deal with this. We could use decorators to mark functions as being in-place and only document things at the `Trace` level and have some sphinx trickery to also render it with the equivalent `Stream` functions.

calum-chamberlain:
[docrep](https://docrep.readthedocs.io/en/latest/) might be useful here - it would be nice to not have to jump from a `Stream` method to a `Trace` method.  I haven't tried docrep, so I don't know if it will work/be useful, and it would add a dependency.

krischer:
> [docrep](https://docrep.readthedocs.io/en/latest/) might be useful here - it would be nice to not have to jump from a `Stream` method to a `Trace` method. I haven't tried docrep, so I don't know if it will work/be useful, and it would add a dependency.

That's fine - I think we can make this a doc building only dependency. We'd just try to import it and if its not available we provide dummy decorators.

This look really nice -  I think I'll use this for a bunch of my things. Thanks for making me aware of it!

d-chambers:
The only subtlety here is that we will want to be careful not to ruin the usefulness of docstrings in the REPL. If we don't make it a hard dependency this would be easy to do.

megies:
> Should we look and pushing most of the detail in the docs for those methods to their corresponding `trace` methods?

That would make it more consistent, yeah. Or look into the docrep thingy right away.
Anything that improves docs is fine by me.

megies:
> If we don't make docrep a hard dependency

As mentioned before, no hard dependency for sure. Just for docs building.

Btw.. how does docrep templates show up in e.g. IPython docstring queries? I guess these will be a bit messy then. Actually, I guess that was what @d-chambers pointed out (no idea what REPL stands for ^^)

d-chambers:
@megies exactly my concern. REPL is Read Eval Print Loop (like ipython). I am done with this unless you think it's worth trying to cleanup all the trace/stream docs in this PR, which I don't mind doing.

megies:
> I am done with this unless you think it's worth trying to cleanup all the trace/stream docs in this PR, which I don't mind doing.

Gonna merge this then. You can do a new PR if you feel like more cleanup, though there might be better spots to work on, I think, esp. what's left in https://github.com/obspy/obspy/milestone/17 and combing through https://github.com/obspy/obspy/milestone/16

PR #2246, opened 2018-10-18
PPSD: fix period_lim with xaxis_frequency=True

### LINKED ISSUE #2245: PPSD.plot: argument xaxis_frequency does not affect argument period_lim
I scanned through Issue results when searching for 'PPSD' and did not see a similar issue, so my apologies if I missed it. 

According to the [docs](https://docs.obspy.org/packages/autogen/obspy.signal.spectral_estimation.PPSD.plot.html), when 'xaxis_frequency' is set to True, 'period_lim' should be expecting a tuple containing the frequency range in Hz as opposed to a tuple containing a period range. I am not finding this to be the case. If what I am finding is true, either the docs should be updated or some code should be changed.

This is not a regression that I am aware of.

Thanks for your time!

ObsPy version: 1.1.0
Python version: Python 3.6.6
Platform: Linux
ObsPy installed from Anaconda

Example:
```python
from obspy.clients.fdsn.client import Client
from obspy import UTCDateTime
from obspy import read
from obspy.signal import PPSD

client = Client("IRIS")

starttime = UTCDateTime('2018-09-18')
endtime = UTCDateTime('2018-09-19')
inventory = client.get_stations(network='PE', station='PAPL',
            starttime=starttime, endtime=endtime, level='response')

tr = read('waveforms/PE.PAPL..HHZ__20180918T000000Z__20180919T000000Z.mseed')
tr = tr.select()[0] 

ppsd = PPSD(tr.stats, metadata=inventory)
ppsd.add(tr)
ppsd.plot(xaxis_frequency=True, period_lim=(1.0, 5.0)
```

megies:
### What does this PR do?

Makes `PPSD.plot(..., period_lim=(...), xaxis_frequency=True)` work as documented in docstring.

Fixes #2245. Since this changes behavior a bit, I decided to base this on master. Doesnt matter too much anyway, since plot ranges can easily be adjusted post-plotting going via matplotlib commands anyway.

### PR Checklist
- [ ] All tests still pass.


megies:

ThomasLecocq:
@megies the images fail to compare in the tests, but I guess this is normal since the documented behaviour was not what was plotted afterwards? e.g. the test requires ``period_lim=(0.01, 50)``  which was not respected in the reference img:
https://i.imgur.com/50BopGl.png

while it is now, thanks to your changes:
https://i.imgur.com/ktvUy8z.png

Should you also update the reference images then ?

the windows' evalresp is not related to this.

megies:
> Should you also update the reference images then ?

Yep, you're right, nice catch!

ThomasLecocq:
@megies still the issue with img generated with mp 2.0.2

megies:
I'll try again with mpl 2.2.2. Maybe I have to up some tolerances if it's still failing..

Btw.. I think we should weed out image tests. Most special case testing should be done by checking the `Axes` objects etc. Like this test is only making sure that changing the x axis label/datalim works, so we shouldn't have a designated image test for it in the first place, I guess.

megies:
OK CI passed, so merging!

Issue #2247, opened 2018-10-19
Add Python 3.7 to CI build matrix


megies:
We should add a Python 3.7 build to CI for at least one platform..

megies:
I think CI should also build 3.7 now for PRs against maintenance branch, I think it looks up the config from master. If not, we should backport #2249.

megies:
- [ ] Also need to add python 3.7 version specifier in setup.py

heavelock:
The newest PRs(ie #2260) are not being built on 3.7 at Travis and the CircleCI is still using old config.  So the #2249 and #2257 should be backported to maintenance branch.

PR #2249, opened 2018-10-19
Adding 3.7 to Travis and Appveyor 

heavelock:


megies:
Well, actually we should have those in obspy main repo anyway, see #2247

heavelock:
Yup, but I wanted first to solve conflicts in my fork and to do the PR when I will solve everything. As for now I see that there is a dependency conflict. I will try to solve it tomorrow.

.travis.yml
```diff
@@ -114,7 +118,13 @@ install:
...
         PYPROJ="pyproj"
       fi
-  - conda create -q -n test-environment
+  - |
+    if [[ "${PYTHON_VERSION}" == '3.7' ]]; then
+      FREETYPE="freetype"
+    else
+      FREETYPE="freetype<2.8"
+    fi
+  - conda create -q -n test-environment numpy scipy matplotlib basemap pyproj
```
megies: `numpy scipy matplotlib basemap pyproj` should already be set in the following lines, no need to add them again on this line

megies:
@heavelock looks good.. mind adding 3.7 on appveyor as well?

heavelock:
Sure! 

Comment about new failures:
1. SDS tests, might be connected with this https://bugs.python.org/issue27364
2. Basemap test are connected with removing `ax._hold` in matplotlib 3.0 https://matplotlib.org/api/api_changes.html#hold-machinery

megies:
> * SDS tests, might be connected with this https://bugs.python.org/issue27364
> 
> * Basemap test are connected with removing `ax._hold` in matplotlib 3.0 https://matplotlib.org/api/api_changes.html#hold-machinery

The matplotlib/basemap guys are already aware of the latter issue, so I guess it'll go away eventually anyways.

PR #2253, opened 2018-10-25
signal: fix AR picker with low amplitude data

### LINKED ISSUE #2252: AR picker fails for low amplitude data
Bug occurs in:
- Current obspy master, Linux, installed from source

This issue was raised on the gitter chat by @Dieselinium_twitter. The AR picker uses floats in the internal c-func, but gives incorrect pick-times when there are low amplitudes. I haven't fully investigated this, but I suspect it is a rounding error and could be fixed by using double-precision for some internal variables in the c-func.

@Dieselinium_twitter found this case after removing instrument response from his data, resulting in low amplitude ground-motion data.

The below minimal example is adapted from the [AR-picker tutorial](https://docs.obspy.org/tutorial/code_snippets/trigger_tutorial.html#ar-picker):

```python
import numpy as np
from obspy.core import read
from obspy.signal.trigger import ar_pick


tr1 = read('https://examples.obspy.org/loc_RJOB20050801145719850.z.gse2')[0]
tr2 = read('https://examples.obspy.org/loc_RJOB20050801145719850.n.gse2')[0]
tr3 = read('https://examples.obspy.org/loc_RJOB20050801145719850.e.gse2')[0]

for tr in [tr1, tr2, tr3]:
    tr.data = tr.data * 1e-6
    tr.data = tr.data.astype(np.float32)
    tr.plot()
df = tr1.stats.sampling_rate
p_pick, s_pick = ar_pick(tr1.data, tr2.data, tr3.data, df,
                         1.0, 20.0, 1.0, 0.1, 4.0, 1.0, 2, 8, 0.1, 0.2)
print(p_pick)
```

This plots each trace, just to make sure that the data can be represented at float32 (which they can).  This outputs a p-pick time of: `-0.10000000149011612`, when it would (without the gain) output `30.6350002289`.

megies:
### What does this PR do?

Fixes #2252 

Precision issues with low amplitude input in AR picker, see #2252.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -73,6 +73,8 @@
      than is used by current ObsPy version (see #2051)
    * The ar_pick() trigger function now raises an error if the three data arrays
      don't have the same length (see #1801, #2148).
+   * fix a precision issue in AR picker in case of low amplitude input (see
+     #2252 and #2253)
  - obspy.taup:
    * Fallback to linear slowness interpolation for very small and
      shallow layers (see #2126, #2129).
--- a/obspy/signal/tests/test_trigger.py
+++ b/obspy/signal/tests/test_trigger.py
@@ -106,6 +106,30 @@ class TriggerTestCase(unittest.TestCase):
         self.assertEqual(err.exception.args[0],
                          "All three data arrays must have the same length.")
 
+    def test_ar_pick_low_amplitude(self):
+        """
+        Test ar_pick with low amplitude data
+        """
+        data = []
+        for channel in ['z', 'n', 'e']:
+            file = os.path.join(self.path,
+                                'loc_RJOB20050801145719850.' + channel)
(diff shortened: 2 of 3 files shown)


calum-chamberlain:
It doesn't look like the failing tests are related. Image comparison fails on travis and appveyor, as well as a couple of other trace ans stream comparison fails on appveyor that don't seem related, and flake8 fails on circleci, but not for the changes made here I think?

megies:
> Looks good to me, and fixes the issue @Dieselinium_twitter had. Thanks for doing this @megies

No worries, simple fix, that. I didnt look too closely at the C code, still think it's strange how it trips over low amplitudes.. if it's a float type, how should the position of the colon matter? I guess I'm missing something here.. whatever, easy fix in Python wrapper

megies:
> It doesn't look like the failing tests are related

Yep, I checked, only unrelated fails. Merging.

PR #2260, opened 2018-11-08
Fix paz poles zeros reading

thefroid:
### What does this PR do?
Fix a bug in the module read_inventory when a SC3ML file is given:
     -Read the poles and zeros even if multiple whitespaces are present between poles and/or zeros 
       in the sc3ml file.

Unable to read the instrumental response of a raspberryshake3d (version5) with the module read_inventory.
### PR Checklist
- [ ] All tests still pass.


Jollyfant:
Thanks! Did you edit one of the test files to include a poles & zeros field with multiple spaces? It's hard to see.

QuLogic:
It is defined [here](https://github.com/obspy/obspy/blob/0ab2d5b4848f4c17d17c950fb7776c7f8a3d9f4a/obspy/io/seiscomp/data/quakeml_types.xsd#L12). All whitespace locations are `*`, which means any number of them are allowed, _including_ 0. I think this means that the code both before and after is still buggy.

There could be no spaces between each element (i.e., `(1,2)(3,4)`) which would never split either before or after this. There could also be spaces within the element (i.e., `( 1 , 2 )`) which would be split too much both before and after.

So this change is good, but incomplete, though that's not the reporter's fault. The original code was not quite correct.

megies:
Nice catch there @QuLogic!

Maybe we need to use some impossible to read regex here? ;-)

```
In [26]: poles = "  (-0.037,0.037)  (-0.037,-0.037) (-6909,     9208)( -6909  ,-9208)  "

In [27]: pattern = r'[ ]*\([ ]*([\-0-9\.]+)[ ]*,[ ]*([\-0-9\.]+)[ ]*\)[ ]*'

In [28]: re.subn(pattern, r'\1 \2 ', poles)
Out[28]: ('-0.037 0.037 -0.037 -0.037 -6909 9208 -6909 -9208 ', 4)
```

..or maybe something less sophisticated but far more readable.. ;-)

```
In [33]: re.sub(r'[(),]', r' ', poles).split()
Out[33]: ['-0.037', '0.037', '-0.037', '-0.037', '-6909', '9208', '-6909', '-9208']

In [35]: np.array(re.sub(r'[(),]', r' ', poles).split(), dtype=np.float64)
Out[35]: 
array([ -3.70000000e-02,   3.70000000e-02,  -3.70000000e-02,
        -3.70000000e-02,  -6.90900000e+03,   9.20800000e+03,
        -6.90900000e+03,  -9.20800000e+03])
```

megies:
In any case, it might be good to do an `assert` afterwards that number of numbers matches opening brackets or something as well..

QuLogic:
If we aren't trying to verify and assume the thing is valid, then a much simpler regex would work: `re.finditer(r'\(\s*([^,\s]+)\s*,\s*([^)\s]+)\s*\)', poles)`
```python
In [1]: poles = "  (-0.037,0.037)  (-0.037,-0.037) (-6909,     9208)( -6909  ,-9208)  "

In [2]: import re

In [3]: re.findall(r'\(\s*([^,\s]+)\s*,\s*([^)\s]+)\s*\)', poles)
Out[3]: 
[('-0.037', '0.037'),
 ('-0.037', '-0.037'),
 ('-6909', '9208'),
 ('-6909', '-9208')]
```

megies:
@thefroid I added some more commits to your PR, think this should be ready to merge then. I also needed to rebase and force-push to get rid of merge conflicts. Hope that's OK for you.

megies:
Fails are unrelated.

megies:
Thanks again, @thefroid!

PR #2264, opened 2018-11-15
build_taup_model() fails for constant-velocity models with more than 2 layers

nackerley:
A layered constant-velocity model can have more first-order discontinuities than a piecewise-linear model. The number of first-order discontinuities in a constant-velocity model is number of layers + 1. Since the minimum number of layers is 1, this should never be less than 2. [n.b. that other issues prevent construction of a 1-layer model, but who cares?] 

Here is where the storage for discontinuities is set up:
https://github.com/obspy/obspy/blob/3ffa3209c637026dc922540f68ad31c9cf69780a/obspy/taup/slowness_model.py#L188-L190
As you can see, the maximum number was the number of layers and the minimum number was 3. This meant that this worked for constant-velocity models with 2 layers, but no more.

I don't understand why numpy.empty() was used; There's no need for speed. Instead I'm proposing numpy.zeros().

The test is a variant on:
https://github.com/obspy/obspy/blob/3ffa3209c637026dc922540f68ad31c9cf69780a/obspy/taup/tests/test_tau.py#L1006-L1024

### PR Checklist
- [x] All tests still pass.


megies:
Looks reasonable to me, not sure if reducing the minimum length from 3 to 2 could harm anybody though? I guess the minimum was three for a reason, but maybe not.. ;-)

nackerley:
I've added tests - I've only checked that the taup test suite has no errors (3 tests skipped).

I think this should be classed as a bug fix - does that mean I need to rebase against a different branch?

nackerley:
> I guess the minimum was three for a reason, but maybe not.

I think it was a mistaken attempt to fix the problem that I've identified - it meant that build_taup_model() started working for models with 2 layers, but no other models. Single-layer models fail for a whole bunch of reasons, starting with, but not limited to, not being able to find the moho. Larger models because the maximum possible number of discontinuities is the number of layer plus one, and this line of code was missing the plus one.

megies:
> I think this should be classed as a bug fix - does that mean I need to rebase against a different branch?

Yep, in that case you should rebase on `maintenance_1.1.x`.

krischer:
I rebased the commits on the latest maintenance branch so we can get this into `1.1.1`. If CI passes this can IMHO be merged.

megies:

krischer:
The 1 failure on appveyor on Python 3.7 is unrelated and we've encountered it before.

PR #2265, opened 2018-11-16
Read fcnt format

thefroid:
### What does this PR do?
Read waveforms and headers of fcnt format. More informations can be extracted from the fcnt file compared to the previous code.

It was initiated because i needed informations which were not extracted by the previous code.
### PR Checklist
- [ ] All tests still pass.


d-chambers:
Hi @thefroid,

Thanks for contributing to ObsPy! It's great to see new contributors, and we look forward to working on ObsPy with you more. 

I certainly see where you are coming from. There is a lot of information contained in the rg16 file format that users might want to extract. The current code only gets enough information to fill out the basics in each traces' stats. However, there are a few issues, which I will list later on, but in general, it is better/easier to take the existing code base and add the features you need rather than replace it with a completely different implementation. This is especially true if you change the signature (the names/number of arguments) of any of the public functions/methods because by doing so you are likely to break existing codes. Also, deleting tests is rarely the right thing to do; each test is probably there for a reason. For example, it may document expected behavior users rely on or ensure a bug does not get reintroduced down the road.

I contributed the original code, and in order to understand some of the specific requirements and discussion of the implementation I suggest you read through the [original PR](https://github.com/obspy/obspy/pull/2070).

A few issues:
1) The signature of `read_rg16` has changed. I know several users (including myself) for which this will cause problems.

2) `read_rg16` should be named `_read_rg16` to be consistent with all the other read functions from the io module. 

3) The automatic mapping from in-line and cross-line to principal components is not desirable. See [this discussion](https://github.com/obspy/obspy/pull/2070#issuecomment-365177223).

4) The efficient `merge` function is important (at least to me), and should not be removed without discussion. 

5) The `three_component` flag is not necessary, and no other parser requires users to specify how many components are contained in the trace before reading the file. 

6) It is probably better to add format-specific information (when `details` is set to `True`) to an attribute dict in the `Stats` object named "rg16". For example, take a look how the sac parser does this for sac-specific info.

7) The read function can no longer accept a file buffer (most parsers should be able to accept both a string specifying path and a file buffer).

8) The file format document included in the docs section is marked as proprietary. I believe (but I am not an IP lawyer) that if we include it in the ObsPy code base it would imply a) that we have the right to distribute it, b) the LGPL license applies to it, and c) the ObsPy copyrights apply to it. I don't think any of these is true. If the document owner provided explicit (written) consent to include it we could probably do it, otherwise I do not feel comfortable including it. We don't want to land ourselves in any legal trouble. 

Moving forward, it would probably be easier to address the issues by modifying the existing code-base (to pull out the information you need) rather than completely replacing it. However, if there is a good reason to replace it, the issues are addressed, and the new code will not break existing codes I am fine with it.

thefroid:
Hi @d-chambers,

Thank you for the code reviewing. I took into account your suggestions and I corrected the points: 1,2,3,4,5 and 7. I am working now on point 6.
 For point 8, I sent an e-mail to a people from FairField to ask if it was possible to put the fcnt documentation in the github repository of ObsPy. He replied to me  : "Yes that is ok. That information is public for users to be able to read our data format". That's why i felt free to share the fcnt documentation. But if it is a problem, I can remove it.
 Concerning point 3, in my mind, the normal process to deploy the nodes is the gold contact toward the north, that's why i automatically mapped in Z,N,E the components. Maybe, It could be automatically mapped in Z, 1, 2 because generally the nodes are deployed fastly on the field, so the orientation error could be more than 5 degrees. It will have the advantage to remove one argument of the function.

ThomasLecocq:
>Concerning point 3, in my mind, the normal process to deploy the nodes is the gold contact toward the north, that's why i automatically mapped in Z,N,E the components. Maybe, It could be automatically mapped in Z, 1, 2 because generally the nodes are deployed fastly on the field, so the orientation error could be more than 5 degrees. It will have the advantage to remove one argument of the function.

hi @thefroid indeed, it's also what is recommended by SISMOB France (they even drew an arrow on top of the nodes :-) )

image

About your fast-merge code, it's very nice and I think it should actually be somehow brought to another level, where it could be useful to any ``Stream`` object, @krischer, @megies: what do you think?

thefroid:
Hi Thomas Lecocq,
The fast-merge code was implemented by Derrick Chambers. I just re-used its function.

d-chambers:
Hey @thefroid,

These changes look promising. Thanks for working on it. I am quite busy atm but I should be able to do a proper review within a week or so. Stay tuned.

megies:
> For point 8, I sent an e-mail to a people from FairField to ask if it was possible to put the fcnt documentation in the github repository of ObsPy. He replied to me : "Yes that is ok. That information is public for users to be able to read our data format". That's why i felt free to share the fcnt documentation. But if it is a problem, I can remove it.

@thefroid It would be good to have the pdf file in the repository. Ideally you should contact FairField again, and be precise, i.e. ask them if we can include the file under LGPL license in obspy. Seems like they have no problem with it anyway.

megies:
> (they even drew an arrow on top of the nodes :-) )

Really funny how a seismic instrument would not sport a "North" arrow when comin from the factory by default..  :sagittarius:

obspy/io/rg16/__init__.py
```diff
@@ -7,18 +7,24 @@
 `Farfield Nodal <fairfieldnodal.com>`_'s
 `Zland <http://fairfieldnodal.com/equipment/zland>`_ product line.
 
+:author:
+    Derrick Chambers
+    Romain Pestourie (Ecole et Observatoire des Sciences de la Terre)
 :copyright:
-    The ObsPy Development Team (devs@obspy.org)
+    the ObsPy Development Team (devs@obspy.org)
```
d-chambers: capitalize "the"

obspy/io/rg16/__init__.py
```diff
@@ -7,18 +7,24 @@
...
 :license:
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 
 .. note::
-    1. Some useful diagrams, provided by Faifield Nodal technical support,
```
d-chambers: These aren't in the manual. I recommend keeping this link in the docs.

obspy/io/rg16/core.py
```diff
@@ -1,233 +1,845 @@
-# -*- coding: utf-8 -*-
-from __future__ import (absolute_import, division, print_function,
```
d-chambers: Unfortunately we need to keep these lines until we drop python 3 support. They should be at the top (after the docstring) of each file.

obspy/io/rg16/core.py
```diff
@@ -1,233 +1,845 @@
...
     :type starttime: optional, obspy.UTCDateTime
-    :param endtime: If None None dont read traces that start after endtime.
+    :param endtime: If not None dont read traces that start after endtime.
     :type endtime: optional, obspy.UTCDateTime
-    :param merge:
-        If True merge contiguous data blocks as they are found. For
-        continuous data files having 100,000+ traces this will create
-        more manageable streams.
+    :param merge: If True merge contiguous data blocks as they are found. For
+     continuous data files having 100,000+ traces this will create
```
d-chambers: multi-line parameter text should be indented an additional level after the first line. See the old code for an example.

obspy/io/rg16/core.py
```diff
@@ -1,233 +1,845 @@
...
-    # get traces and return stream
-    traces = _make_traces(filename, theader_start, gheader, head_only=headonly,
-                          starttime=time1, endtime=time2, merge=merge,
-                          standard_orientation=contacts_north)
+    if starttime is None:
+        starttime = UTCDateTime(1970, 1, 1)
+    if endtime is None:
+        local_time = time.localtime()
+        endtime = UTCDateTime(local_time[0], local_time[1], local_time[2],
+                              local_time[3], local_time[4], local_time[5])
```
d-chambers: It would be easier to just init a far out time eg:
```python
endtime = UTCDateTime('3000-01-01').timestamp
```
one line rather than 3.

d-chambers: infact, combining these to one line might be a good idea:
```python 
starttime = starttime or UTCDateTime(1970, 1, 1)
endtime = endtime or UTCDateTime("3000-01-01")
```

obspy/io/rg16/core.py
```diff
@@ -1,233 +1,845 @@
...
-#     num_traces1 = np.sum([x['num_channels'] for x in channel_dicts])
-#
-#     # try multiplying general_header and num_records. This seems to be right
-#     # when there arent that many treaces in the file
-#     num_traces2 = channel_sets * num_records
-#
-#     return max(num_traces1, num_traces2)
-
-
-if __name__ == '__main__':
```
d-chambers: These lines should be at the end of each file. They run the doctests (when applicable)

d-chambers:
I am looking this over more and there are still some general issues.

1) Make sure to run the linter to catch style violations. You can just install flake8 then run `python obspy/core/tests/test_code_formatting.py` and if the test passes you don't have any issues. The future imports should be at the top of every file.

2) Reading three component data now takes ~80% longer than the previous code. This is a problem because often rg16 files can be very large.

3) `obspy.read` no longer works on rg16 files with or without the format keyword specified. 

4) In testing this code on some large rg16 files from work I am running into some difficult to debug issues. This is bit frustrating because the existing implementation works fine.

There are a few potential paths forward on this one:
1) We keep working through these issues until the new code passes
2) We just modify the old code to pull the additional data out of the headers you need
3) We use the new code as a header parser (e.g. to create the "rg16" attribute dict) then combine it with the old code for when the `details` parameter is set to True. 

Honestly, my preferred path forward on this is 2 or 3. I am having a hard time justifying ripping out the entirety of the old (working/tested) implementation just to add support to read additional data from the headers. However, since I wrote the original implementation I may be a bit biased so perhaps I should "recuse" myself. What do you think @ThomasLecocq or @megies ?

obspy/io/rg16/__init__.py
thefroid: I removed the link because the two first figures seem to me not very obvious to understand.
The third figure is mentioned later /_images/rg16_node_orientation.png.

thefroid:
I corrected the problems raised by Chambers (except point 2 and point 4).

thefroid:
@d-chambers  What are the arguments that you pass to the function "_read_rg16" to conclude that the new code is 80% slower than the old one?
I compared the running time of the new code vs the old code with a big file (14 GB). When the arguments pass to the function "read" are the filename and the format (RG16) the new code is approximately 15% faster than the older. When the arguments starttime and endtime are added, the new code is 40 % faster.
@d-chambers  Could you be more explicit about point 4 ?

ThomasLecocq:
About testing: I will run the code on my FCNT files tonight & let you know...

d-chambers:
I was just testing on the three component test file, which is small. If the new code is performing better on large files that is certainly a good reason to keep it. I will test more on some of our larger files tomorrow and let you know. I would be interested in any profiling info you might capture (@ThomasLecocq). I will also look more into 4 tomorrow, there may have been another issue with the file.

@thefroid, there are a few stylistic things that might be improved, but it is probably easier to show you than try to explain all of them in a code review. Do you mind if I push a few changes to your branch?

thefroid:
@megies I contacted FairField to know if the documentation about rg16 format could be included under LGPL license in obspy. He replied to me that "if the information from the pdf file was just uploaded, it was ok".
 I assume that the pdf file won't be modify, so it seems good.

thefroid:
@d-chambers Feel free to make stylistic changes.

megies:
> He replied to me that "if the information from the pdf file was just uploaded, it was ok".
> I assume that the pdf file won't be modify, so it seems good.

Sure, it's just included as is, for reference/documentation purposes. So that should be OK.

d-chambers:
I did some testing and profiling. You are correct, the new code performs much better on larger files, but worse on small 3 component files. Since the main use-case for this parser is to read many large rg16 files (sometimes 100s or 1000s) and convert them to some other format, I would say the new code has a significant advantage.  

File number | Components | Size (mb) | Old Code Read Time (s) | New Code Read Time (s)
-- | -- | -- | -- | --
1 | 1 | 19030 | 23.3 | 9.01
2 | 3 | 6778 | 9.55 | 4.46
3 | 1 | 0.024 | 4.15 | 4.71
4 | 3 | 0.36 | 2.73 | 4.84

Concerning point 4, I am not able to reproduce it. I will dig into it more but I suspect something was a bit screwy in how I was testing the files, switching between obspy versions, or any number of things.

d-chambers:
Ok @thefroid take a look at those changes and let me know if there is anything you don't agree with. The one thing we still need to address is the failing doctest in `obspy.io.rg16.__init__`, there is no function `_read_initial_headers` so it may be removed, or you may have had something else in mind.

thefroid:
@d-chambers the changes seem to me good. 
In the function _read_rg16 (core.py), when the endtime is not defined by the user, the default date is UTCDateTime("3000-01-01"). It is a detail (as you wrote it's the problem of somebody else ;-) ), but I would prefer to not use a constant, i think writing endtime=UTCDateTime() is a good alternative.
I added the importation of the function _read_initial_headers in the __init__.py, so now the doctest pass. .

megies:
> UTCDateTime("3000-01-01").

In the year 2525... if man is still alive.. :musical_note:

d-chambers:
@thefroid, ok np. I guess there was just something unnerving to me about the default endtime being generated dynamically. Unless there is something wrong with the timestamps in an rg16 file, however, I don't see how it can cause a problem.

thefroid:
Just a question about fields stored in the rg16 object contained in the stream object.
In order to homogenize units, i converted all fields relative to time in seconds (like the field "applied clock correction time shift") and all fields relative to frequency in hertz (like the field "test signal generator frequency 1"). Moreover, some fields are natively multiplied by a constant (for example the receiver position is multiplied by ten). I removed this constant. What is the best practice: let the fields as they are natively in the rg16 files or do what i did ?

d-chambers:
In my opinion you did the most sensible thing. Users will not want to work with decimeters, so if you didn't convert to a standard unit they would have to do it.

d-chambers:
>Is the creation of the key "_format" useful in the function read?

I would say no, neither `obspy.io.mseed.core._read_mseed` or `obspy.io.sac.core._read_sac` add the `_format` key. Remember, the user should not be calling `_read_rg16` directly but rather using the `obspy.read` function, and optionally specifying the format as rg16 and passing its supported keyword arguments.

 >Should i integrate the units in keys contained in the rg16 object

I don't have a strong opinion about this. Introducing the unit names could cause some confusion when numbers are contained in the names. In your example, for instance, "test_signal_generator_amplitude_2" clearly tells me it is the second test signal generator's amplitude (or something along those lines) but "test_signal_generator_amplitude_2_dB" isn't clear from first glance if it is the second test signal generator's amplitude and the units are dB or if it is referring to a test signal amplitude that is being generated at a constant 2 dB (or something along those lines). However, it is up to you. I am ok with whatever you want to do here.

megies:
>  In order to homogenize units, i converted all fields relative to time in seconds (like the field "applied clock correction time shift") and all fields relative to frequency in hertz (like the field "test signal generator frequency 1"). Moreover, some fields are natively multiplied by a constant (for example the receiver position is multiplied by ten). I removed this constant. What is the best practice: let the fields as they are natively in the rg16 files or do what i did ?

We are talking about format specific header fields that end up exclusively in `Trace.stats.rng16` dictionary, right?
I agree with @d-chambers that most people will expect / hope for SI units probably. But on the other hand, we usually stay pretty close to the original headers inside these dictionaries. Another option would be to explicitly add the unit to the converted field name, e.g. make it `Trace.stats.rng16['applied clock correction time shift in seconds']`, that would be most verbose and clear while only sacrificing the perfect match to the original format specific header field name (which could only startle hardcore rng16 users for a moment, I guess).

(And it could mean extra work when writing back to rng16, but I don't think anybody ever will want to add write support)



> Should i integrate the units in keys contained in the rg16 object

I think leaving special properties like these in `dB` it fine.

obspy/io/rg16/core.py
```diff
@@ -1,231 +1,738 @@
 # -*- coding: utf-8 -*-
```
megies: Please remove executable flag from this file again.

obspy/io/rg16/__init__.py
```diff
@@ -29,34 +35,41 @@
...
 .. figure:: /_images/rg16_node_orientation.png
 
-Reading
+Reading the waveforms
 -------
 The rg16 format can be read using two methods:
 
 1. Using the standard :func:`~obspy.core.stream.read` function. Optionally,
-   the format parameter can be specified as "rg16" for a modest speed-up.
+   the format parameter can be specified as "rg16" for a speed-up.
```
megies: Modesty out the window?   (just kidding.. :stuck_out_tongue_winking_eye:)

obspy/io/rg16/__init__.py
```diff
@@ -29,34 +35,41 @@
...
 
 2. Using the :mod:`obspy.io.rg16` specific function
    :func:`obspy.io.rg16.core._read_rg16`.
 
 Noteworthy parameters of  :func:`obspy.io.rg16.core._read_rg16`,
 which can also be passed as kwargs to :func:`~obspy.core.stream.read`:
 
-* merge: If `True` will merge traces belonging to the same channel
+* `starttime` and `endtime`: :class:`~obspy.core.utcdatetime.UTCDateTime`
```
megies: Monospace actually needs double backticks.. I see that a lot of the missed monospacing was already in here before this PR, but would you mind fixing it now, nevertheless, while you're at it?  :-)

obspy/io/rg16/tests/test_util.py
```diff
@@ -1,93 +1,59 @@
+# -*- coding: utf-8 -*-
```
megies: Please remove executable flag from this file again

obspy/io/rg16/util.py
```diff
@@ -1,32 +1,28 @@
 # -*- coding: utf-8 -*-
```
megies: Remove executable flag, please

obspy/io/rg16/util.py
```diff
@@ -1,32 +1,28 @@
...
 import copy
-import struct
+
+from future.builtins import *  # NOQA
+from codecs import encode
 
 import decorator
 import numpy as np
+from future.builtins import list
+from future.utils import native_str as nstr
```
megies: please keep all `future` imports grouped at the top of imports

obspy/io/rg16/util.py
```diff
@@ -39,131 +35,105 @@ def _open_file(func, *args, **kwargs):
...
-            >i. - 4 bit int, left four bits
+    :type length: int or float
+    :param dtype: bcd, binary or IEEE
     :type dtype: str
-    :param new_dtype: Any valid numpy data type.
-    :return:
+    :param left_part: If True, start the reading from the first half part
+        of the byte position. If False, start the reading from the second
+        half part of the byte position.
+    :type left_part: boolean
```
megies: Not sure if `boolean` works (probably does), we usually have `bool`

obspy/io/rg16/util.py
```diff
@@ -39,131 +35,105 @@ def _open_file(func, *args, **kwargs):
...
-        data = np.fromstring(fi.read(int(length)), dtype)
-        if new_dtype is not None:  # cast data to new_dtype (due to #2198)
-            data = data.astype(new_dtype)
+    if dtype == 'bcd':
+        return _read_bcd(fi, length, left_part)
+    if dtype == 'binary':
+        return _read_binary(fi, length, left_part)
+    if dtype == 'IEEE':
+        data = np.frombuffer(fi.read(int(length)), '>f4')
         return data[0] if len(data) == 1 else data
```
megies: this is just cosmetics, but logically this should be `elif`s.

obspy/io/rg16/util.py
```diff
@@ -39,131 +35,105 @@ def _open_file(func, *args, **kwargs):
...
 
 
-@_register_read_func('bcd')
-def _read_bcd(fi, length):
+def _read_bcd(fi, length, left_part):
```
megies: we have some BCD code in `io.reftek` already I think. Just leave as is for now, but ideally we want to avoid duplicated code..

CHANGELOG.txt
```diff
@@ -36,6 +36,8 @@ master:
  - obspy.io.reftek:
    * Implement reading reftek encodings '16' and '32' (uncompressed data,
      16/32bit integers, see #2058 and #2059)
+ - obspy.io.rg16:
+   * implement module to read waveforms and headers from fcnt format.
```
megies: please add ticket number

megies:
I skimmed it and made some nit-picky comments.. aside from that if @d-chambers is happy, I'm happy. Thanks for all the work @thefroid :tada:

obspy/io/rg16/util.py
d-chambers: There was some talk on refactoring all of our parsers to use common utilities functions. It would  be worthwhile but a huge effort best left for another PR.

d-chambers:
I try to always be happy, it's easier to live life that way :smile:

obspy/io/rg16/__init__.py
```diff
@@ -77,7 +90,14 @@
...
+-------
+The initial headers at the beginning of the rg16 file can be read
+separetely using the :mod:`obspy.io.rg16` specific function
+:func:`obspy.io.rg16.core._read_initial_headers`
+
+>>> initial_headers = _read_initial_headers(filename)
+
+
 """
-from __future__ import (absolute_import, division, print_function,
```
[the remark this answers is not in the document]
megies: Yes, @thefroid  these lines have to be in (almost) every .py file. There's also a test for it, but I guess that fail is lost in all the test fail noise we have right now.. :(

obspy/io/rg16/__init__.py
```diff
@@ -4,8 +4,8 @@
 
 Functions to read waveform data from Receiver Gather 1.6-1 format.
 This format is used to store continuous data recorded by
-`Farfield Nodal <fairfieldnodal.com>`_'s
-`Zland <http://fairfieldnodal.com/equipment/zland>`_ product line.
+``Farfield Nodal <fairfieldnodal.com>``_'s
+``Zland <http://fairfieldnodal.com/equipment/zland>``_ product line.
```
megies: I'm sorry if I confused you earlier @thefroid.. these links should stay as before. Please see the autogenerated PR docs here: http://docs.obspy.org/pull-requests/2265/packages/obspy.io.rg16.html?highlight=fairfield

Also see this page for reference: http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

obspy/io/rg16/__init__.py
```diff
@@ -36,32 +36,32 @@
...
 The rg16 format can be read using two methods:
 
-1. Using the standard :func:`~obspy.core.stream.read` function. Optionally,
+1. Using the standard :func:``~obspy.core.stream.read`` function. Optionally,
    the format parameter can be specified as "rg16" for a speed-up.
 
-2. Using the :mod:`obspy.io.rg16` specific function
-   :func:`obspy.io.rg16.core._read_rg16`.
+2. Using the :mod:``obspy.io.rg16`` specific function
+   :func:``obspy.io.rg16.core._read_rg16``.
```
megies: same here @thefroid these are internal sphinx links, see http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-roles  (although we just use `:func:` and not `:py:func:`), so they need exactly one backtick.

obspy/io/rg16/util.py
megies: I agree, should be done in another PR. Mostly mentioned it for future reference and FYI  :+1:

obspy/io/rg16/__init__.py
```diff
@@ -29,34 +35,40 @@
...
 The rg16 format can be read using two methods:
 
 1. Using the standard :func:`~obspy.core.stream.read` function. Optionally,
-   the format parameter can be specified as "rg16" for a modest speed-up.
+   the format parameter can be specified as "rg16" for a speed-up.
 
-2. Using the :mod:`obspy.io.rg16` specific function
-   :func:`obspy.io.rg16.core._read_rg16`.
+2. Using the specific function :func:`obspy.io.rg16.core._read_rg16` in the
+   module :mod:`obspy.io.rg16`.
```
megies: Using the private function should *not* be advertized as a legit standard use case. I know this has been in there before your PR @thefroid but would you be so kind to remove the ad for the private function? You can have a look at e.g. the mseed landing page: http://docs.obspy.org/master/packages/obspy.io.mseed.html?highlight=_read_mseed

The RG16 landing page should clearly advertize using `read()` and should only mention the private method `_read_rg16()` by stating that additional `kwargs` to pass in to `read()` should be looked up there.

CC @d-chambers

megies:
Last minor nit picks:
 - [ ] don't advertize direct use of private function `_read_rg16` on RG16 API landing page
 - [ ] some files are still made executable by this PR, please make them non-executable again like mentioned before (e.g. `obspy/io/rg16/core.py 100644 → 100755 ` see e.g. this page if anything is unclear: https://stackoverflow.com/a/40979016)

obspy/io/rg16/__init__.py
```diff
@@ -89,16 +63,6 @@
...
-
-Reading the initial headers
----------------------------
-The initial headers at the beginning of the rg16 file can be read
-separately using the specific function
-:func:`obspy.io.rg16.core._read_initial_headers` in the module
-:mod:`obspy.io.rg16`
-
->>> from obspy.io.rg16.core import _read_initial_headers
->>> initial_headers = _read_initial_headers(filename)
```
megies: Hmm. @thefroid this should maybe stay in, as this is really functionality that the user can not achieve with `core` high level routines.

megies:
@d-chambers and @thefroid, if I read this correctly..

https://github.com/obspy/obspy/blob/af58138568aaa4acf12cadfb73cf68c27c79e005/obspy/io/rg16/core.py#L63-L67

..then the user could end up missing data that is actually in between user-specified `starttime` and `endtime`, as this check only respects the start time of a given time series. Is the length of the time series known at this time? If so, this should really be changed to include segments that partially overlap given start/endtime and a `stream.trim(...)` added at the very end to cut off any unwanted parts.

obspy/io/rg16/__init__.py
thefroid: In fact, the information contained in the initial headers can be retrieved calling the read function with parameter details set to True. It creates a rg16 object containing information from initial headers and trace headers.

megies: Ah I see, alright then! Is unpacking these "details" very costly? Otherwise it might be worth a thought to unpack those details by default and use that switch to opt out for speed critical scenarios?

thefroid:
@megies Yes you are right, it can miss data. For example when the start of the data block is before the starttime and the end of the data block is after the starttime set by the user, the data block is not read. I corrected this bug in my before last commit. A stream._ltrim and  stream._rtrim is applied in the read function if options starttime and endtime are used, so that's why i didn't do a stream.trim() in the read_rg16 function.

obspy/io/rg16/__init__.py
megies: Ok, then it's fine as it is now. :+1:

megies:
> A stream._ltrim and stream._rtrim is applied in the read function if options starttime and endtime are used, so that's why i didn't do a stream.trim() in the read_rg16 function.

Ah OK, yeah makes sense that we do that in `read` at the end.. didn't remember. ^^
Just saw this minor flaw about selection of preliminary rg16 traces in the rg16 read method. Thanks a bunch for fixing that oversight @thefroid! :tada:

megies:
Looks good, thanks again @thefroid! Merging! :tada:

Issue #1983, opened 2017-11-02
Release 1.1.1


megies:
Just reserved a DOI on zenodo for next release: 10.5281/zenodo.1040770

Things to do, when release candidates and final release are done:
- [x] Run docker tests (@krischer / @megies -- are done routinely now, but double-checking can't hurt)
- [x] source distribution to pypi
- [x] conda packages via conda-forge: conda-forge/obspy-feedstock#6
- [ ] conda packages for Linux 32bit via https://github.com/obspy/obspy/tree/master/misc/installer/anaconda and distributed at https://anaconda.org/obspy/obspy/files (@krischer / @megies)
- [ ] maxiconda bundled installer via recipes in https://github.com/obspy/anaconda-installers but built locally on win+nix and attached to release: https://github.com/obspy/obspy/releases/tag/1.1.0 (@megies and @barsch)
- [ ] wheels for Linux+OSX (@krischer)
- [x] Windows pypi packages (@barsch / appveyor)
- [ ] debian packages (@megies)
- [x] redhat/centos packages (@qulogic)
- [ ] macports packages (@petrrr)
- [x] zenodo: reserve a DOI pre-release (@megies)
- [x] zenodo: upload sdist zipball and finalize release (@megies)
- [x] Update docs once released (@megies)
- [x] News entry in the wiki
- [ ] "What's new?" page in the wiki
- [x] Email to the users and announcements mailing lists
- [x] Tweet
- [ ] update base branches for all issues bumped to 1.1.1 to a new branch `maintenance_1.1.x`
- [ ] check dependency versions, some old distros we used to support might have gone EOL since release of 1.0, so we might be able to bump minimum dependency versions of e.g. numpy, scipy, matplotlib, ...
---
- [ ] Party :tada: (don't tick before all other checkboxes are ticked!)

krischer:
We should probably consider getting `1.1.1` out ASAP (maybe next week?) and just include whatever is in it by then. A ton of things are in the milestone (https://github.com/obspy/obspy/milestone/17) but nothing really crucial and we can go over some of them during the next couple of days.

`1.2.0` will be up and coming fairly soon in any case...

megies:
I think some of the PRs in milestone 1.1.1 are pretty much ready. Maybe you can start by punting others to a new milestone 1.1.2?

megies:
1.1.1rc1 sdist zipball: 
[...]

PR #2269, opened 2018-11-24
Fix test failures on master

d-chambers:
### What does this PR do?

Fixes the test failures on master due to flake8, escape characters in python 3.7, and a few other misc. issues.

The test suite on master was failing due to several small issues. 

### PR Checklist
- [X] All tests still pass.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -113,7 +113,7 @@ install:
   - "activate test"
   # Install default dependencies
   # pip 10.0.1 seems buggy and breaks all our appveyor builds
-  - "conda install -q --yes \"pip<10\" numpy scipy matplotlib lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema pyshp cryptography"
+  - "conda install -q --yes \"pip<10\" numpy scipy matplotlib lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema pyshp cryptography \"proj4=4.9.3\""
   # somehow basemap/pyproj/proj4 can fail when PROJ_LIB is not set correctly
   # for whatever reason, see:
   #  - conda-forge/basemap-feedstock#36
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -669,6 +669,7 @@ class ClientTestCase(unittest.TestCase):
         expected = normalize_version_number(expected)
         self.assertEqual(got, expected, failmsg(got, expected))
 
+    @unittest.expectedFailure
     def test_dataselect_bulk(self):
         """
         Test bulk dataselect requests, POSTing data to server. Also tests
@@ -678,17 +679,18 @@ class ClientTestCase(unittest.TestCase):
         file = os.path.join(self.datapath, "bulk.mseed")
         expected = read(file)
         # test cases for providing lists of lists
+        # Deliberately requesting data that overlap the end-time of a channel.
+        # TA.A25A..BHZ ends at 2011-07-22T14:50:25.5
         bulk = (("TA", "A25A", "", "BHZ",
-                 UTCDateTime("2010-03-25T00:00:00"),
-                 UTCDateTime("2010-03-25T00:00:04")),
+                 UTCDateTime("2011-07-22T14:50:23"),
+                 UTCDateTime("2011-07-22T14:50:29")),
                 ("TA", "A25A", "", "BHE",
                  UTCDateTime("2010-03-25T00:00:00"),
                  UTCDateTime("2010-03-25T00:00:06")),
                 ("IU", "ANMO", "*", "HHZ",
                  UTCDateTime("2010-03-25T00:00:00"),
                  UTCDateTime("2010-03-25T00:00:08")))
-        # As of 03 December 2018, it looks like IRIS returns data shorter than
-        # minimumlength if the start and end time specified force shorter data.
+        # As of 03 December 2018, it looks like IRIS is ignoring minimumlength?
         params = dict(quality="B", longestonly=False, minimumlength=5)
         for client in clients:
             # test output to stream
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -1363,11 +1363,11 @@ class ClientTestCase(unittest.TestCase):
                     endtime=UTCDateTime("2010-02-27T06:40:00.000"))
 
         # test invalid token/token file
-        with self.assertRaisesRegex(
+        with self.assertRaisesRegexp(
                 ValueError,
                 'EIDA token does not seem to be a valid PGP message'):
             client = Client('GFZ', eida_token="spam")
-        with self.assertRaisesRegex(
+        with self.assertRaisesRegexp(
                 ValueError,
                 "Read EIDA token from file '[^']*event_helpstring.txt' but it "
                 "does not seem to contain a valid PGP message."):
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -687,6 +687,8 @@ class ClientTestCase(unittest.TestCase):
                 ("IU", "ANMO", "*", "HHZ",
                  UTCDateTime("2010-03-25T00:00:00"),
                  UTCDateTime("2010-03-25T00:00:08")))
+        # As of 03 December 2018, it looks like IRIS returns data shorter than
+        # minimumlength if the start and end time specified force shorter data.
         params = dict(quality="B", longestonly=False, minimumlength=5)
         for client in clients:
             # test output to stream
@@ -1359,11 +1361,11 @@ class ClientTestCase(unittest.TestCase):
                     endtime=UTCDateTime("2010-02-27T06:40:00.000"))
 
         # test invalid token/token file
-        with self.assertRaisesRegexp(
+        with self.assertRaisesRegex(
                 ValueError,
                 'EIDA token does not seem to be a valid PGP message'):
             client = Client('GFZ', eida_token="spam")
-        with self.assertRaisesRegexp(
+        with self.assertRaisesRegex(
                 ValueError,
                 "Read EIDA token from file '[^']*event_helpstring.txt' but it "
                 "does not seem to contain a valid PGP message."):
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -77,7 +77,8 @@ def normalize_version_number(string):
     Due to Py3k arbitrary dictionary ordering it also sorts word wise the
     input string, independent of commas and newlines.
     """
-    repl = re.sub('v[0-9]+\.[0-9]+\.[0-9]+', "vX.X.X", string).replace(",", "")
+    match = r'v[0-9]+\.[0-9]+\.[0-9]+'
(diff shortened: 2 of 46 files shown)


d-chambers:
Well it works on my machine 😉

calum-chamberlain:
+TESTS:ALL

calum-chamberlain:
I have done a few things in here to make some failing network tests pass - these all seemed to be due to changes in sensor availability, or changes in how the service was delivered. 

I haven't fixed a failing fdsn test `obspy.clients.fdsn.tests.ClientTestCase.test_dataselect_bulk` which checks for a few things, including the use of the `minimumlength` argument.  To me it looks like how IRIS handle that might have changed.  The test specifies a start and end time 4s apart and a `minimumlength` of 5. We are now getting traces less than minimum length. In the FDSN docs it isn't clear whether `starttime` and `endtime` or `minimumlength` should take priority.

For the molepeople plot issues, I did some experimentation and it looks like `proj4=5.2.0` is causing the issue - I tested 4.9.3 and 5.1.0 and found both to work.  I have added a check of the version when importing `obspy.imaging.maps` which follows the other warnings.  I also pinned the version for ci to 4.9.3, this probably isn't ideal.

calum-chamberlain:
I'm not sure what is going on with the appveyor fail relating to the use of `TemporaryWorkingDirectory` - previously I added a check to make sure that the files were closed (they are) but it still complains that something else is accessing the files.  Not sure if we should just ignore it?

d-chambers:
> I also pinned the version for ci to 4.9.3, this probably isn't ideal.

Yes, looks like python 3.7 cant use proj4 4.9.3:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - proj4=4.9.3
  - pyproj
  - python=3.7
Use "conda info <package>" to see the dependencies for each package.
```

Maybe we should just skip the test if the version is greater than 5.1 for most builds and only pin it on one/a few of  them?

calum-chamberlain:
I have marked one of the network tests as an expected fail just to see what other CI fails we have. It fails at the moment because `minimumlength` looks like it is being ignored.  I'm not sure if IRIS changed how they respond to requests or if it is on our end? The following example shows the issue:
```python
from obspy import UTCDateTime

client = Client("IRIS")
st = client.get_waveforms(
    station="A25A", network="TA", channel="BHZ", location="*",
    starttime=UTCDateTime("2011-07-22T14:50:23"),
    endtime=UTCDateTime("2011-07-22T14:50:29"), minimumlength=5)
print(st)
"""
Returns:
1 Trace(s) in Stream:
TA.A25A..BHZ | 2011-07-22T14:50:23.000000Z - 2011-07-22T14:50:25.500000Z | 40.0 Hz, 101 samples
which is less than 5 seconds long.
"""
```

calum-chamberlain:
Current failures: 
1. There are still a few plotting issues which seem to be related to matplotlib version 3. The [PPSD test fail](https://travis-ci.org/obspy/obspy/jobs/463035988#L2843) is just a change in where xticks are plotted for version 3 (this also happens on the appveyor run). 
2. I'm not sure what the [basemap related fails](https://travis-ci.org/obspy/obspy/jobs/463035988#L2707) are on 3.7
3. Some [taup fails](https://travis-ci.org/obspy/obspy/jobs/463035991#L6160) on MacOS Python 3.7 that I haven't debugged.
4. A [simulate fail on appveyor](https://ci.appveyor.com/project/obspy/obspy/builds/20742130/job/alxfomlq6cl4ierq#L963), that I haven't looked into yet.

I'm going to focus on other things for a bit - @d-chambers do you want to do anything more on this?

appveyor.yml
```diff
@@ -113,7 +113,7 @@ install:
   - "activate test"
   # Install default dependencies
   # pip 10.0.1 seems buggy and breaks all our appveyor builds
-  - "conda install -q --yes \"pip<10\" numpy scipy matplotlib lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema pyshp cryptography"
+  - "conda install -q --yes \"pip<10\" numpy scipy matplotlib lxml sqlalchemy mock nose gdal decorator requests basemap jsonschema pyshp cryptography \"proj4=4.9.3\""
```
calum-chamberlain: Could unpin proj4 here, but that would likely mean that plotting of maps wouldn't get testes on windows.

obspy/clients/fdsn/tests/test_client.py
```diff
@@ -668,6 +669,7 @@ def test_str_method(self):
         expected = normalize_version_number(expected)
         self.assertEqual(got, expected, failmsg(got, expected))
 
+    @unittest.expectedFailure
```
calum-chamberlain: Does anyone know if IRIS has changed how they respond to `minimumlength`?

d-chambers:
I should have time to pick this up again after AGU, I will take a look at the remaining issues at that point.

obspy/clients/fdsn/tests/test_client.py
d-chambers: Maybe @chad-iris or @nick-iris know or could find out for us?

megies:
> It fails at the moment because `minimumlength` looks like it is being ignored. I'm not sure if IRIS changed how they respond to requests or if it is on our end? The following example shows the issue:

Looks like a server side issue at IRIS. I'll drop @chad-iris a short mail, he can forward it to the person in charge.

> Could unpin proj4 here, but that would likely mean that plotting of maps wouldn't get tested on windows.

Actually these package version specs in the CI yamls have been annoying me for a long while. I think at some point it would be best to move the packages specs from the yamls to conda pinning files that we copy in the conda test env during CI. That would also make it much easier to avoid getting different package versions when doing more installation commands subsequently. But this should be discussed in a separate issue first, I think.

>     2\. I'm not sure what the [basemap related fails](https://travis-ci.org/obspy/obspy/jobs/463035988#L2707) are on 3.7

That the `..hold` thingy? This is a known issue of some matplotlib versions that broke basemap, I can't find the corresponding issue right now unfortunately..

megies:
@d-chambers @calum-chamberlain if you don't have time to work on the rest of the failures right now, I would still suggest we merge this as soon as possible, whenever appropriate. Best not keep this hanging and get it merged soonish, even if some fails remain.

Feel free to merge when you feel like it!

calum-chamberlain:
The `axis._hold` type issue is [related to this](https://github.com/matplotlib/basemap/issues/435) I think. Added a `skipIf` for matplotlib version 3.0.1 and basemap >= 1.1.0.
I can't reproduce the appveyor `test_simulate` fail, but it looks like it might be similar to #2188. I couldn't reproduce on my 64Bit windows with a clean conda env using the same packages as appveyor.

d-chambers:
> I would still suggest we merge this as soon as possible

Good idea. I wont have time to do anything on this for a week or two so go ahead and merge whenever you are happy @calum-chamberlain. It is nice to see Travis green again.

megies:
> We could just check `allclose` on data on windows as was done for #2188?

[This the only info we have on the fail?](https://ci.appveyor.com/project/obspy/obspy/builds/20776339/job/1589igujyjgducwi#L952)

```
  File "C:\projects\obspy\obspy\core\tests\test_stream.py", line 2207, in test_simulate_seedresp_parser
    self.assertEqual(tr1, tr2)
AssertionError: <obspy.core.trace.Trace object at 0x00000002A2D67080> != <obspy.core.trace.Trace object at 0x00000002A66E6438>
```

If that's the case, we should add an `assertEqual` on the `tr.data` to see how big the difference is, first of all?



> skipping tests for specific versions that either do not work (such as basemap with mpl 3.0.1,

this is the right thing to do, IMHO. we won't see a lot of mpl 3.0.1 anyway once the next minor release is out (that'll fix this hold issue)


> skipping the PPSD plot test for mpl>=3 which moves the xtick locations - I think the new locations are nicer and result in a better use of the figure space

At some point we want to update our baseline images to mpl 3. Maybe we want to wait for mpl 3.0.2, though. We can then have higher tolerances for failing tests with mpl 2.x
I'm okay with merging this right now, just open an issue please as a reminder to switch baseline images to mpl 3 at some point, ideally against the 1.2.0 milestone?

> And ignoring the network test change for `minimumlength`.

Hmm. not sure about this one.. maybe better leave it in as a failing test? I strongly believe it's a server side bug that will be fixed pretty fast by IRIS once it gets the attention of the person in charge (I dropped Chad a mail to forward the issue but he didn't respond yet, probably busy with AGU preps).



> I'm also not super keen on the change I made to [`taup`](https://github.com/obspy/obspy/pull/2269/commits/7eabcca746e89335984e548745a2a5b61b28d3f6#diff-59955951fd157f214a342cd6e6160b84L156) - changing from resizing inplace to generating a new array with `np.resize` as a function might have implications for memory and speed, but I don't know the taup stuff well enough to know if the size of these arrays is going to be an issue

This looks like we need to be careful not to break things.. dunno. I also don't know the taup stuff much, but this looks like it can break references to the same data array?

calum-chamberlain:
Thanks for those comments @megies. 
- I have removed the changes to `obspy.taup.tau_branch`, so I expect those doctests to fail. I think the fails could be related to how the doctests are run - maybe each doctest is trying to work on the same array at the same time? 
- Removed the expected fail flag on the test with the `minimumlength` issue.
- Added some (hopefully useful) extra comparisons for the two `simulate` tests that seemed flakey, hopefully we can see whether there really is an issue or not with that.

chad-iris:
>  It fails at the moment because minimumlength looks like it is being ignored. I'm not sure if IRIS changed how they respond to requests or if it is on our end?

Indeed the option was effectively being ignored.  This is now fixed at IRIS.

**Note**: We are deprecating this option as we have a much better way to do this with our Research Ready Data Sets service, which allows selection of data based on quality metrics including time coverage.  It's supported for now, but not promised to be working in the future.

It's been broken for 6 months and this is the first we've heard about it, which aligns with my suspicion that it's not often used.

calum-chamberlain:
@chad-iris thanks! Personally I haven't used the `minimumlength` constraint, and wouldn't have known it wasn't working if not for the tests.  Nice to hear about the RRDS too, I wasn't aware of that.
Sounds like it might be a good idea for us to not base a test on this then if it won't be supported in the future.

calum-chamberlain:
The simulate fail that remains on windows passes the tests comparing stats and using `allclose` (with default settings, so `rtol=1e-7`) but still fails the [stream and trace comparison tests](https://ci.appveyor.com/project/obspy/obspy/builds/20828146/job/fkbjsw88toc97w7p#L941)

megies:
> Sounds like it might be a good idea for us to not base a test on this then if it won't be supported in the future.

Without looking at the (test) code, we could replace the test by checking that the appropriate URL is built inside the client. But then again, I am confident that the URL building inside the client is tested thoroughly enough, so probably we can just get rid of that parameter when downloading data in tests. On the other hand, since it's fixed on server side, we can also just leave it as is for now.

d-chambers:
I prefer the second option. It should also have an option to ignore `processing` attribute of the `stats` object, as that has been a sticking point in the past as well. Although this would be a nice feature, it probably belongs in a separate PR no?

megies:
OK sounds good, so probably add a utility Stream/Trace assertion function in `core.util.testing` but postpone for a future PR.

megies:
In any case, I'm gonna merge this now to not have it in limbo for too long. Thanks for the initiative on this tedious matter @d-chambers and @calum-chamberlain.

megies:
We might decide to backport this for 1.1.1 to get clean CI builds, probably, see #1983

PR #2275, opened 2018-12-17
Update sc3ml xsl files for amplitude pick conversion

### LINKED ISSUE #2273: obspy.io.seiscomp._read_events() misses some picks
The attached file has 7 picks associated with an event, but this is what I get when I use obspy.io.seiscomp._read_events() to read it:

``` python
from obspy import read_events
print(read_events('test_sc3ml_0.9.xml')[0])
```
```
Event:  1952-12-25T04:28:32.500000Z | +43.850,  -80.970 | 3.0 MN | manual

                    resource_id: ResourceIdentifier(id="smi:ca.seismo/event/evid/19521225.0428002")
                     event_type: 'earthquake'
           event_type_certainty: 'known'
            preferred_origin_id: ResourceIdentifier(id="smi:ca.seismo/origin/solid/19521225.0428002")
         preferred_magnitude_id: ResourceIdentifier(id="smi:ca.seismo/magnitude/solid/magcnt/19521225.0428001/1")
                           ---------
                          picks: 6 Elements
                     amplitudes: 1 Elements
                        origins: 2 Elements
                     magnitudes: 3 Elements
             station_magnitudes: 1 Elements
```
My expectation is that all 7 picks should be loaded.

I have the same problem regardless of whether I use 1.1.1rc1.post0+436.g3ffa3209c6.obspy.master installed using "python setup.py develop" after cloning from github, or version 1.1.0 installed using anaconda.

I have this problem both with sc3ml 0.9 (obspy 1.1.0 and 1.1.1) and 0.10 (obspy 1.1.1): 
[test_sc3ml.xml.zip](https://github.com/obspy/obspy/files/2650661/test_sc3ml.xml.zip).

calum-chamberlain:
### What does this PR do?
Updates the sc3ml xsl conversion files as updated in this [upstream pull request (SeisComP3/seiscomp3#208)](https://github.com/SeisComP3/seiscomp3/pull/208). This allows conversion of picks related to an amplitude (but not an arrival) to be converted to quakeml and therefore read in.

Note, I have based this on master because there have been updates to the sc3ml tests (including adding 0.10 schema) in master.

Fixes #2273 

### PR Checklist
~~- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
~~- [ ] If any network modules should be tested for the PR, add them as a comma separated list
- [ ] All tests still pass.
~~- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .~~


calum-chamberlain:
I originally pushed to the branch `sc3ml_amplitudes`, but realised I mucked up a rebase, so cleaned this one up. I don't have permission to delete that branch though, so if someone else could that would be great!

Brtle:
Thanks for updating these files!

Maybe the new test file is a bit heavy (~35KB)? The current test files are already around 200KB. Perhaps it can be pruned to keep only 1 origin, 1 pick referenced in arrival and 1 pick referenced in amplitude. It should be clearer.

I opened [another PR](https://github.com/SeisComP3/seiscomp3/pull/209) on the seiscomp3 repository to write non-QuakeML nodes in a custom namespace. That should be a nice improvement allowing round-trip conversion without losing nodes but I think I will open another PR for that. There are some changes to do in the sc3ml module before merging these files.

obspy/io/seiscomp/tests/test_event.py
```diff
@@ -88,6 +90,8 @@ def cmp_read_xslt_file(self, sc3ml_file, quakeml_file, validate=True):
                 if validate:
                     self.assertTrue(_validate_quakeml(tf.name))
                 filepath_cmp = os.path.join(self.path, quakeml_file)
+                if debug and not filecmp.cmp(filepath_cmp, tf.name):
+                    shutil.copyfile(tf.name, filepath_cmp + "_failed")
```
Brtle: It's indeed pretty annoying to debug the file comparison. However if this file is written on the disk, it should also be the case for the other tests. Either we write them all or none.

We can perhaps also keep this mechanism with the debug parameter always false. The failed files will be generated only if a developper manually sets it to true. This solution is not perfect since it adds some unreachable code… I don't know what would be best.

calum-chamberlain:
I have pulled out everything except the origin, pick and amplitude from the sc3ml, and the conversion now fails.  It appears that only picks that are referenced by an amplitude that is referenced by a station magnitude are converted now. I have left this as it is for now to demonstrate this, but I'm not sure if this is how you want it @Brtle? I have cases of amplitude picks that are not related to magnitudes that would get lost in this conversion, so I would prefer to see the picks being converted.

Brtle:
I felt free to commit a minimized test file. It's still a valid sc3ml file but indeed the magnitudes and stationmagnitudes must also be kept.

If it's OK for you, I think the PR can be rebased and merged.

calum-chamberlain:
Grand, seems fine to me - if anyone has issues with missing amplitude picks due to them not being associated with magnitudes we can refer them here and if they want to they can update the seiscomp repository.
@megies yup, just copied the files across.
Thanks @Brtle!
I am not authorized to merge this PR - maybe @megies can?

Brtle:
> if anyone has issues with missing amplitude picks due to them not being associated with magnitudes we can refer them here and if they want to they can update the seiscomp repository.

Well, technically a stationmagnitude is needed. Without it, you can't know to which origin the amplitude is related and therefore you can't convert the amplitude correctly. In the sc3ml format, the amplitude is under the EventParameter node but in QuakeML it is in the event node. Or am I missing something?

In the sparse file you did, there is no link between amplitude and origin/event. The only way to convert the amplitude is to assume that the amplitude belongs to the event but you can't be sure. It can be the amplitude of another event.

megies:
Thanks a lot for your work on this guys! And also for getting it into seiscomp3 upstream.

@calum-chamberlain we'll give some more people, you included, push rights for our main branches.. we've been a bit over restrictive with our main branch settings I guess, following some botched PR that got merged into master some while ago which forced a history clean up and force-pushing. I wasn't even really aware that you guys can't merge an approved PR even with current restrictions on main branches..

PR #2279, opened 2019-01-01
Resource id rerefactor

### LINKED ISSUE #2278: Resource IDs can refer to wrong object when former object goes out of scope
Ubuntu 18.04, python 3.7, master branch of ObsPy

Currently `get_referred_object` method of the `ResourceIdentifier` can return the wrong object if the object that was bound to the `ResourceIdentifier` instance gets deleted and a new object is assigned to the same python id. [Here](https://stackoverflow.com/questions/20753364/why-is-the-id-of-a-python-class-not-unique-when-called-quickly) is a good SO post that demonstrates how python object ids are only unique over an object's lifetime, not over the program lifetime.

The following code demonstrates the problem:

```python
import obspy.core.event as ev

# a class to count instance creations
class Counter:
    count = 0
    def __init__(self):
        self.count = self.__class__.count
        self.__class__.count += 1

# simple setup of objects and resource IDs
rid1 = ev.ResourceIdentifier(referred_object=obj1)
rid2 = ev.ResourceIdentifier(referred_object=obj2)

# rid1 and rid2 work as expected, obj1, obj2 have the following ids
print(id(obj1), id(obj2))

# now if we delete obj1 and obj2 then create additional python objects which are
# bound to different resource ids
del obj1
del obj2
obj3, obj4 = Counter(), Counter()
rid3 = ev.ResourceIdentifier(referred_object=obj3)
rid4 = ev.ResourceIdentifier(referred_object=obj4)

#  If the python id of obj1 or obj2 id is re-used we are going to have problems...
print(id(obj3), id(obj4))

# rid3 and rid4 work as expected, of course. However, if obj3 or obj4 has the same python
# id as obj1 or obj2 then rid1 or rid2 will return the wrong object rather than None.
# So, the following will sometimes (~50% of the time) raise an AssertionError because
# obj3 or obj4 is assigned to  a python id that was previously used by obj1 or obj2
# and the object id is the key for the ResourceIdentifier._id_object_map dict
assert rid1.get_referred_object() is None
assert rid1.get_referred_object() is None
```

d-chambers:
### What does this PR do?

This PR is yet another change to the `ResourceIdentifier` class. It does two things:

1) Fixes #2278 (which is a bug only found on the master branch).

2) Adds the ability for users to plug custom logic into `ResourceIdentifier` for returning objects with the `get_referred_object` method not found via the normal means.
 

#2278 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,11 @@ master:
    * Added __eq__ to QuantityError so empty instances equal None (see #2185).
    * Reworked the event scoped resource identifiers for the event classes
      hopefully fixing all edge-cases (see #2091).
+   * Fixed issue where resource_ids could refer to incorrect object if the
+     original object went out of scope and a new object shared the python
+     id of the old object (see #2278).
+   * Added a hook to allow users to customize finding objects for
+     resource_ids which are not found via the normal means (see #2279).
    * Calling Stream.write(...) on an empty stream will now raise an
      ObsPyException consistently across all I/O plugins (see #2201)
  - obspy.clients.fdsn:
@@ -37,7 +42,7 @@ master:
    * Implement reading reftek encodings '16' and '32' (uncompressed data,
      16/32bit integers, see #2058 and #2059)
  - obspy.io.rg16:
-   * implement module to read waveforms and headers from fcnt format (see #2265). 
+   * implement module to read waveforms and headers from fcnt format (see #2265).
  - obspy.io.quakeml:
    * Allow writing invalid ids but raise a warning
      (see #2104, #2090, #2093, #1872).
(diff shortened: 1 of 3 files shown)


d-chambers:
Hey @krischer, any chance you can give this a quick look-over? It is very simple compared to the other resource_id PRs and shouldn't take much time. 

Essentially it just uses a tuple of `(object_id, resource_id)` rather than just the `object_id`  for referencing specific objects, and adds a hook for returning referred_objects from a callable if the normal means fail.

CHANGELOG.txt
```diff
@@ -17,6 +17,11 @@ master:
    * Added __eq__ to QuantityError so empty instances equal None (see #2185).
    * Reworked the event scoped resource identifiers for the event classes
      hopefully fixing all edge-cases (see #2091).
+   * Fixed issue where resource_ids could refer to incorrect object if the
+     original object went out of scope and a new object shared the python
+     id of the old object (see #2278).
```
megies: If I understand correctly, this problem was only introduced on master and it's not in any release version, so I think we don't need a bugfix comment?

d-chambers: Correct, will remove it.

PR #2280, opened 2019-01-09
Fix taup py37 failures

d-chambers:
### What does this PR do?
I am attempting to fix the python 3.7 TauP failure on Travis in the least invasive way possible. 

#2269 fixed a lot of failures on master but we weren't sure how to fix this one.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/taup/tau_branch.py
+++ b/obspy/taup/tau_branch.py
@@ -481,8 +481,8 @@ class TauBranch(object):
     @staticmethod
     def _robust_resize(arr, new_size):
         """
-        Try to resize an array inplace. If an error is raised used numpy
-        resize function then return the array.
+        Try to resize an array inplace. If an error is raised use numpy
+        resize function to create a new array. Return the array.
         """
         try:
             arr.resize(new_size)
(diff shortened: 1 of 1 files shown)


d-chambers:
@calum-chamberlain, what do you think of this? It is a bit of a hack but since no TauP experts are forthcoming this seems like a reasonable compromise. Hopefully all tests turn green after this change.

calum-chamberlain:
This looks good to me. 

If we are uncertain about whether the implementation will impact anyones codes then we could warn that the reference to the original array is lost. It might also be worth linking to this PR so that if an issue does arise then the user can quickly find out why. I don't really like that idea, but its more of a "covering-our-backs" thing.

d-chambers:
Yes, it is probably better to annoy than to surprise.

calum-chamberlain:
The appveyor fail is an unrelated fail as far as I know - trace similarity things as we also discussed in #2269 
Looks good to me!

megies:
Thanks @d-chambers!

PR #2281, opened 2019-01-09
Apollo seismometer format support: obspy.io.alsep

yamamo-to:
**Purpose**
This module is an extension to read seismic data obtained by Apollo missions. The module supports three tape formats: PSE Tape, WTN Tape, and WTH Tape. The module only has the read function. 

**Background**
The Apollo mission was known as the human activity on the moon, but science was also great. A series of experiments was called an Apollo Lunar Surface Experiments Packages (ALSEP). As a part of ALSEP, the seismometers were mounted on the moon, and they measured the lunar seismic events from 1969 to 1977. 

**Data**
The data can be downloaded from the DARTS website:
- http://darts.isas.jaxa.jp/pub/apollo/

**Tutorial**
Let's draw the same figure in figure 8-4 in the following paper:
http://an.rsl.wustl.edu/apollo/data/a15/station8/pse/docs/pse_psr.pdf

```python
from obspy import read, UTCDateTime

st = read('http://darts.isas.jaxa.jp/pub/apollo/pse/p15s/pse.a15.1.2')
st_spz = st.select(id='XA.S15..SPZ')
st_spz.plot(starttime=UTCDateTime('1971-08-02T17:11:22.6'),
            endtime=UTCDateTime('1971-08-02T17:11:35.8'))
```

**Problem**
1. Takes several minutes to download and decode ALSEP tapes
2. Test code uses the real data and all unit tests require more than 20 minutes.

**Reference**
- http://www-udc.ig.utexas.edu/external/yosio/PSE/catsrepts/TechRept118.pdf
- http://darts.isas.jaxa.jp/planet/seismology/apollo/


megies:
Cool. I just tried it and indeed it takes rather long to decode. Are the packets always the same size and all headers etc at the same byte position in the packet? If yes, I think it could be made much faster similar to the `io.reftek` approach, i.e. unpacking in one `numpy.from_buffer` call with a complex dtype ending up with an structured array.

megies:
I assume there will be license problems so we can't include test files in our repo.. (we'd need to be able to include it under LGPL)?!

isas-yamamoto:
> Cool. I just tried it and indeed it takes rather long to decode. Are the packets always the same size and all headers etc at the same byte position in the packet? If yes, I think it could be made much faster similar to the `io.reftek` approach, i.e. unpacking in one `numpy.from_buffer` call with a complex dtype ending up with an structured array.

The tape format is based on 10-bit ALSEP word, and bit-field operation is essential. I will check the `io.reftek` approach with bit-field.

megies:
> The tape format is based on 10-bit ALSEP word, and bit-field operation is essential. I will check the `io.reftek` approach with bit-field.

I mean.. it really depends.. since this data format is exclusively for the old Apollo data I guess, it might not be worth it to spend too much time into optimization though? If somebody needed to read the data loads of times, they could still read it once and convert it to e.g. MiniSEED.

megies:
> Not the license problem, but the filesize. Each test file is about 10 MB. I did not want to include files of this size in the repository. I had to create dummy data for unittest.

dummy data is fine if you can be bothered to mimick the encoding.. otherwise if the data is in fully self-contained packets, you could just trim one original file to the first few packets?

megies:
@isas-yamamoto if you are working on this, you need to eventually rebase on current master branch. This would also get rid of some of the CI test fails.

ThomasLecocq:
@yamamo-to could you rebase this on the latest master branch ?

2018331113:
Hello, i am trying to use your module but I can't use it by just importing "obspy"
I've also checked their docs and there isn't any module available there.
I'm currently facing trouble to read jaxa .2 files.
is there a way to add your module in my project?

ThomasLecocq:
@megies we coud rebase on master & merge ?

megies:
continued in linked PR

megies:
@yamamo-to I'm really sorry this took so long to get merged, given that it was basically ready :sob: 
Anyway, it's in master now for upcoming 1.4.0 release, see #3156

Issue #1825, opened 2017-06-26
Tests fail on Alpine Linux


Brtle:
I wanted to put ObsPy 1.0.3 in an [Alpine Linux docker](https://gist.github.com/Brtle/eb9749f2e345058c365a521319b0aad4) but a unit test fails:

```
======================================================================
ERROR: test_calc_vincenty_inverse (obspy.geodetics.tests.test_util_geodetics.UtilGeodeticsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/obspy/geodetics/base.py", line 155, in calc_vincenty_inverse
    alpha = math.asin(sin_alpha)
ValueError: math domain error

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/obspy/geodetics/tests/test_util_geodetics.py", line 49, in test_calc_vincenty_inverse
    res = calc_vincenty_inverse(0, 0, 0, 17)
  File "/usr/lib/python3.6/site-packages/obspy/geodetics/base.py", line 191, in calc_vincenty_inverse
    raise StopIteration
StopIteration

----------------------------------------------------------------------
Ran 1386 tests in 491.079s

FAILED (errors=1)
```

This is quite weird since this unit test doesn't seem to use any particular library (except `math`). I also tried to pull the latest version of ObsPy without success.

I then find [this numpy bug](https://github.com/numpy/numpy/issues/7128). Do you think the unit test can fail due to a bug in musl's trigonometric functions like in numpy?

krischer:
Hmm - that would be totally possible but I don't know enough about musl to be any help here. Can you maybe add a couple of print statements in the `calc_vincenty_inverse()` function - then we can compare against a non-MUSL libc and potentially narrow down the error.

Brtle:
Hmm, I put some print statements and the [`math.atan2`](https://github.com/obspy/obspy/blob/master/obspy/geodetics/base.py#L145) doesn't give the same result.

On AlpineLinux:

```
>>> import math
>>> math.atan2(0.29237170472273677, 0.9563047559630354)
0.29670597283903605
```

On Ubuntu:

```
>>> import math
>>> math.atan2(0.29237170472273677, 0.9563047559630354)
0.2967059728390361
```

I don't think it's a problem with ObsPy, sorry for opening this issue!

krischer:
[...]

PR #2282, opened 2019-01-10
Fix vincenty inverse for equatorial line

Brtle:
I looked in detail the calculation of vincenty inverse to fix the unit test on Alpine (see #1825). I think that a case is not correctly handled when `cos²α = 0`. On [this website](https://www.movable-type.co.uk/scripts/latlong-vincenty.html), a note describes this case:

> Vincenty observes that eqn. (18) becomes indeterminate over equatorial lines (since cos²α → 0); in this case, set cos2σm to 0 and the result is computed correctly.

The problematic unit test calculates the distance between (0, 0) and (0, 17) which is precisely an equatorial line. Handling the `cos²α = 0` fixes this issue. Another test on the equatorial line was expected to fail (distance between (0, 0) and (0, 13)) and is converging now.

Moreover, this case is also handled in [geopy](https://github.com/geopy/geopy/blob/master/geopy/distance.py#L573) in the same way.

I also did some minor changes:

- Some cos/sin was already calculated, I reused directly the value instead of recalculating the cos/sin.
- `α` was calculated but only `cos²α` was useful, I calculated it directly without using `asin`.
- This code block was inside the loop for no reason:

```python
u2 = pow(math.cos(alpha), 2) * (a * a - b * b) / (b * b)
_a = 1 + (u2 / 16384) * (4096 + u2 * (-768 + u2 *
                                        (320 - 175 * u2)))
_b = (u2 / 1024) * (256 + u2 * (-128 + u2 * (74 - 47 * u2)))
delta_sigma = _b * sin_sigma * \
    (cos2sigma_m + (_b / 4) *
        (cos_sigma * (-1 + 2 * pow(cos2sigma_m, 2)) - (_b / 6) *
            cos2sigma_m * (-3 + 4 * sqr_sin_sigma) *
            (-3 + 4 * pow(cos2sigma_m, 2))))

dist = b * _a * (sigma - delta_sigma)
alpha12 = math.atan2(
    (math.cos(u_2) * math.sin(dlon)),
    (math.cos(u_1) * math.sin(u_2) -
        math.sin(u_1) * math.cos(u_2) * math.cos(dlon)))
alpha21 = math.atan2(
    (math.cos(u_1) * math.sin(dlon)),
    (-math.sin(u_1) * math.cos(u_2) +
        math.cos(u_1) * math.sin(u_2) * math.cos(dlon)))
```

It was useless to calculate these variables at each iteration, I moved the block at the end of the function.

### DIFF
--- a/obspy/geodetics/base.py
+++ b/obspy/geodetics/base.py
@@ -162,8 +162,7 @@ matplotlib/files/matplotlib-toolkits/basemap-0.9.5/
                 cos2sigma_m = cos_sigma - \
                     (2 * math.sin(u_1) * math.sin(u_2) / sqr_cos_alpha)
 
-            c = (f / 16) * sqr_cos_alpha * \
-                (4 + f * (4 - 3 * sqr_cos_alpha))
+            c = (f / 16) * sqr_cos_alpha * (4 + f * (4 - 3 * sqr_cos_alpha))
             last_dlon = dlon
             dlon = omega + (1 - c) * f * sin_alpha * \
                 (sigma + c * sin_sigma *
--- a/obspy/geodetics/base.py
+++ b/obspy/geodetics/base.py
@@ -40,6 +40,13 @@ WGS84_A = 6378137.0
 WGS84_F = 1 / 298.257223563
 
 
+def _isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
+    """
+    Equivalent of the :meth:`math.isclose` method compatible with python 2.7.
+    """
+    return abs(a - b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
+
+
 def calc_vincenty_inverse(lat1, lon1, lat2, lon2, a=WGS84_A, f=WGS84_F):
     """
     Vincenty Inverse Solution of Geodesics on the Ellipsoid.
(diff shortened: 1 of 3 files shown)


obspy/geodetics/base.py
```diff
@@ -140,41 +147,29 @@ def calc_vincenty_inverse(lat1, lon1, lat2, lon2, a=WGS84_A, f=WGS84_F):
...
+            sqr_cos_alpha = 1 - sin_alpha * sin_alpha
+            if _isclose(sqr_cos_alpha, 0):
+                # Equatorial line
+                cos2sigma_m = 0
+            else:
+                cos2sigma_m = cos_sigma - \
+                    (2 * math.sin(u_1) * math.sin(u_2) / sqr_cos_alpha)
+
+            c = (f / 16) * sqr_cos_alpha * \
+                (4 + f * (4 - 3 * sqr_cos_alpha))
```
QuLogic: I think this fits on one line now?

Brtle: Fixed!

obspy/geodetics/base.py
```diff
@@ -40,6 +40,13 @@
 WGS84_F = 1 / 298.257223563
 
 
+def _isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
```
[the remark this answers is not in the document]
Brtle: According to [this comment](https://github.com/obspy/obspy/pull/2265#issuecomment-448998882), `np.isclose` is not available for numpy < 1.7. That's why I didn't use it.

Brtle:
> Looks good, thanks! Can you add a one-liner to the changelog as well?

Oops, I forgot that. Done!

megies:
Thanks @Brtle! :tada:

PR #2284, opened 2019-01-17
Add custom tags and 0.11 SC3ML event support

Brtle:
The 0.11 SC3ML schema is out. Some nodes have been added (DataSegment, DataAttributeExtent, DataExtent and DataAvailability) but they are not in the `EventParameters` node. So the only interesting change for the obspy SC3ML module is the new `passedQC` attribute in `StationMagnitude`.

This is a minor change but every generated files in the unit tests need to be updated due to the version number. So a `change_writing_version` method has been added for unit tests to avoid modifying raw files at each version change. Basically, the files are generated with the latest version of the schema in the unit tests and the version is modified to compare it to the reference file. Tell me what you think about that.

I also added custom tags in this PR. Some changes have been made on the XSLT files which allows round-trip conversion (see https://github.com/SeisComP3/seiscomp3/pull/209). However, some custom tags wasn't correctly converted. I did some changes in the QuakeML module to fix them.

For instance, custom tags without value weren't kept, such as:

```xml
<phaseSetting code="test" lowerPeriod="0.1" upperPeriod="0.2"/>
```

The XSLT conversion also put the namespace of the custom tags directly on the node and not in the header. These tags weren't kept either.

```xml
<scs:passedQC xmlns:scs="http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.11">true</scs:passedQC>
```

Finally, some custom tags are inside `TimeQuantity` node, especially the `pdf` one.

```xml
<time>
  <value>2011-03-11T05:46:24.120000Z</value>
  <pdf>
    <variable>2002-05-30T09:00:00 2003-05-30T09:00:00 2004-05-30T09:00:00</variable>
    <probability>0.5 1.3 2.8</probability>
  </pdf>
</time>
```

It's a bit tricky because the value is stored directly in ObsPy and not in an element tree. So I chose to put the custom tags in the `QuantityError` element. However, the `QuantityError` raises a warning when an `extra` node is added. I muted it but it's pretty ugly:

```python
error.do_not_warn_on.append('extra')
self._extra(el, error)
```

Do you know if there is a better way to do that?


acarapetis:
Are there any plans for movement on this? I'd like to use obspy for more of our workflows but the lack of support for recent seiscomp XML makes it hard.

megies:
This is such a long diff, I might not get to a proper review soon.. if somebody could step up though..

Looks like you update all test files to 0.11? Any chance we can keep testing some older version SC3ML too? If not and its too much work then, well, whatever I guess and just bump test files, dont look back

megies:
@acarapetis i think we'd have to rebase this on current master and then if you could check the branch and play around a bit with old SC3ML files and 0.11 ones and give feedback we might get away with a bit less thorough of a code review

claudiodsf:
Pinging @Brtle on this. That would be very useful for our workflows at IPGP to have SC3ML 0.11 support!

Brtle:
I did a rebase on the current master and the unit tests are running correctly. There have been a lot of new commits over the past 2 years so it would be good to test with real files. I would not be able to test on such files since I don't work in seismology anymore, sorry about that.

> Looks like you update all test files to 0.11? Any chance we can keep testing some older version SC3ML too? If not and its too much work then, well, whatever I guess and just bump test files, dont look back

In the unit tests, the version number of each files is changed and the valid versions of the file are tested. This allows old versions to be tested too.

claudiodsf:
Thanks @Brtle for this extra effort!

Let's ping @megies and @acarapetis to see if someone is interested in testing and reviewing this code.

I'm also available, in case.

filefolder:
FWIW, I have been hacking a way around this for some months without any serious issues (essentially just adding the xslt file and  "0.11" to the list of valid schema haha), so at this point, with SC4 being out over a year now, it may be time to just let this fly and deal with any possible problems rather than drag this out any longer? (would only 0.11 users be affected anyway?)

Regardless, bump, and thanks again to @Brtle for the effort! I can't review the code but I can supply example sc4.5.0 xml where needed.

barsch:
Looks good enough for me -but @krischer is the expert so he should give his ok ;)

ThomasLecocq:
@claudiodsf & others I think we should discuss this soon(ish) - SC3ML support in general and how to make this part of master soon?

claudiodsf:
Hi @ThomasLecocq, I agree!

How do you want to proceed? Should we reserve a one-hour Zoom call for a mini-sprint to merge this?

filefolder:
How's this going? The trouble now is that obspy/seiscomp has become unusable again as they've now been on 0.12 for some months. 

At this point... If it were up to me, I would probably just forego the meetings, click merge, and let god sort em out, but that's probably one of the many reasons it isn't up to me.

damb:
This is kind of OT, however, the current schema version is `0.12`, already (see e.g. https://github.com/SeisComP/common/commit/78f429e62e453c55ea4070b98262aeb0468662c6) . So maybe you use this opportunity to update to the most recent version.

filefolder:
At this point I don't want to get greedy, but yeah I recall the 0.11 to 0.12 edit being trivial, mostly just adding [these xsl files](https://github.com/SeisComP/common/tree/master/libs/xml/0.12) and 0.12 to the accepted versions. Been running well on that dumb hack for quite a while now but there's probably something else somewhere. 

Regardless here's hoping this PR can get submitted before this next release! We're so close!

ThomasLecocq:
@claudiodsf I'l give a shot at starting back from this PR and merging/cherry-picking elements from the others this week.

filefolder:
if it is at all helpful i've already added 0.12 support and also managed to incorporate _most_ of the three PRs for this at the master here: https://github.com/filefolder/obspy

I did not upload their example sxcml files though since they were still at 0.11 i believe. attached are a [0.12 scxml catalog](https://github.com/obspy/obspy/files/7895435/catalog_0.12.scxml.txt) and [0.12 inventory](https://github.com/obspy/obspy/files/7895437/inv_0.12.scxml.txt) file which i have just now created using the master branch linked above.

they write and read fine except when loading the scxml 0.12 inventory i get this error:

```
inv = obspy.read_inventory('inv_0.12.scxml')
/home/seisop/.local/lib/python3.8/site-packages/obspy/io/seiscomp/inventory.py:704: UserWarning: FIR response not in inventory: ResponsePAZ/20220119081455.957826.43, stopping correction before stage 3
  warnings.warn(msg)
```

so there is still some work to be done, but perhaps easier to attend to it afterwards. attaching also the [original fdsnxml](https://github.com/obspy/obspy/files/7895438/au.mcq.xml.txt) version to see what the differences are.

ThomasLecocq:
> if it is at all helpful i've already added 0.12 support and also managed to incorporate _most_ of the three PRs for this at the master here: https://github.com/filefolder/obspy

@filefolder maybe it's better that you create a fresh PR from this branch then ?

PR #2286, opened 2019-01-21
Steam/Trace almost_equal method

d-chambers:
### What does this PR do?
This PR adds the `almost_equal` method to the `obspy.Stream` and `obspy.Trace` class definitions. These methods allow for a less strict comparison of waveform data in two ways:

1) Optionally, only the default parameters of stats objects are used in the comparisons. This is controlled by the `default_stats` parameter. In my experience the `processing` attribute can sometimes differ slightly between two steams that are otherwise identical, in which case it may still be desirable to consider them equal.

2) The numpy arrays of the traces are compared as "close" using the function `numpy.allclose`. This allows the user to specify relative and absolute tolerances for comparison using the `atol` and `rtol` parameters.  

This feature request was mentioned in #2269, and probably in other places before that. It will primarily be useful for obspy's test suite but may find usages in obspy-related projects as well. It will hopefully fix the windows issues the CI is running into.

It would be nice to implement an option that would raise an error if two streams/traces are not equal with a nice message explaining the differences, but that would require a little more energy/time than I currently have so it may be best left for a future PR. 

I am not 100% satisfied with the method name, so if anyone has other ideas please share.

### PR Checklist
- [x] All tests still pass.

+DOCS

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -819,7 +819,7 @@ class Trace(object):
         return out
 
     def almost_equal(self, tr, default_stats=True, rtol=1e-05, atol=1e-08,
-                     equal_nan=False):
+                     equal_nan=True):
         """
         Return True if the trace is approximately equal to another trace.
 
@@ -839,8 +839,11 @@ class Trace(object):
         :return: bool
         """
         # First compare the array values
-        all_close = np.allclose(self.data, tr.data, rtol=rtol, atol=atol,
-                                equal_nan=equal_nan)
+        try:  # Use equal_nan if available
+            all_close = np.allclose(self.data, tr.data, rtol=rtol, atol=atol,
+                                    equal_nan=equal_nan)
+        except TypeError:  # If on older version of numpy
+            all_close = np.allclose(self.data, tr.data, rtol=rtol, atol=atol)
         # Then compare the stats objects
         stats1 = _make_stats_dict(self, default_stats)
         stats2 = _make_stats_dict(tr, default_stats)
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -818,12 +818,12 @@ class Trace(object):
         out.data = data
         return out
 
-    def almost_equal(self, tr, default_stats=True, rtol=1e-05, atol=1e-08,
+    def almost_equal(self, other, default_stats=True, rtol=1e-05, atol=1e-08,
                      equal_nan=True):
         """
         Return True if the trace is approximately equal to another trace.
 
-        :param tr: Another :class:`~obspy.core.trace.Trace` object.
+        :param other: Another :class:`~obspy.core.trace.Trace` object.
         :param default_stats:
             If True only compare the default stats on the traces, such as seed
             identification codes, start/end times, sampling_rates, etc. If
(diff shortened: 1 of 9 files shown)


d-chambers:
> Instead of having it as a method on the Stream/Trace classes I'd rather have it under some obspy.testing namespace

Sure no problem. Let's just make sure we are on the same page before I change anything. 
This is a bit more complicated than numpy because obspy has more than one data class. So should there be a `trace_almost_equal`, and `stream_almost_equal` function, or maybe just an `almost_equal` function and the types should be detected and the inputs dispatched accordingly? Of course if we ever want to extend this to `Catalog` or `Inventory` objects there may need to be different arguments and it may get rather jumbled. 

Also are you ok with the name? Maybe it should be `allclose` like numpy, or would that cause confusion?

obspy/core/trace.py
```diff
@@ -2941,6 +2982,15 @@ def _data_sanity_checks(value):
         raise ValueError(msg)
 
 
+def _make_stats_dict(tr, default_stats):
+    """
+    Return a dict of stats from trace optionally including processing.
```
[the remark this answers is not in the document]
d-chambers: Do you think the `default_stats` param description is insufficient? 

        If True only compare the default stats on the traces, such as seed
        identification codes, start/end times, sampling_rates, etc. If
        False also compare extra stats attributes such as processing and
        format specific information.

jkmacc-LANL:
Couldn’t you add some parameters to `Trace.__eq__` the same way you do to `Stream.__str__` for non-standard behavior?  Expert users could use the dunder method directly.

d-chambers:
> Couldn’t you add some parameters to `Trace.__eq__`

We could, but that would make `Trace.__eq__` quite  a bit uglier and more complex. Also incurs the risk of introducing bugs to a commonly used function rather than just creating a new function. One the flip side, it would be one less function to remember. I don't really have a strong opinion on where this functionality lives but it is needed to (hopefully) fix the last CI failure that crops up on appveyor.

megies:
Didn't look at the details, but the general idea is definitely something very nice to have for us @d-chambers  :rocket: :tada:

obspy/core/util/testing.py
```diff
@@ -1050,5 +1050,167 @@ def setup_context_testcase(test_case, cm):
...
+    # First compare trace stats objects.
+    stats1 = _make_stats_dict(tr1, default_stats)
+    stats2 = _make_stats_dict(tr2, default_stats)
+    if stats1 != stats2:
+        return False
+    # Then  compare the array values
+    try:  # Use equal_nan if available
+        all_close = np.allclose(tr1.data, tr2.data, rtol=rtol,
+                                atol=atol, equal_nan=equal_nan)
+    except TypeError:
```
krischer: We usually handle these issues by using https://github.com/obspy/obspy/blob/master/obspy/core/compatibility.py

An advantage of this is that all compatibility things are collected in one place and we could in theory periodically check and remove no longer needed ones.

obspy/core/util/testing.py
```diff
@@ -1050,5 +1050,167 @@ def setup_context_testcase(test_case, cm):
...
+        # This happens on very old versions of numpy (< 1.7) which do not
+        # have isclose.
+        all_close = _all_close(tr1.data, tr2.data, rtol=rtol,
+                               atol=atol, equal_nan=equal_nan)
+    return all_close
+
+
+def _all_close(ar1, ar2, rtol, atol, equal_nan):
```
krischer: Also move this to https://github.com/obspy/obspy/blob/master/obspy/core/compatibility.py.

d-chambers:
>  How about calling the functions assert_traces_almost_equal() and assert_streams_almost_equal()

I like it. The parity with `numpy.testing.assert_almost_equal` is enough to convince me. It will also, as you say, make it easier to add better error messages in the future.

d-chambers:
hmmm... this is frustrating. Seems like some unrelated failures are starting to crop up.

d-chambers:
rebased to current master

d-chambers:
> Is there anything you still want to add?

I think you are right about creating an obspy.testing module and I would like to add better error messages, but that seems like a 2.0 thing to me ;). Merging in, thanks for the review.

PR #2292, opened 2019-01-24
Pass SAC byteorder flag correctly

### LINKED ISSUE #2285: obspy.read() unaffected by byteorder settings for SAC files
I am under the impression that obspy.read() behavior seems unaffected by setting 'byteorder' settings.
For example, I was not able to load a little-endian SAC trace (created by an home-made Fortran routine of my own). Even if I tried specifying explicitely "byteorder=little" in option.
However, reformatting the same record using a big-endian byteorder using the same routine produced a file that I could load properly using obspy.read() with no option.
Note: My computer is natively ordering bytes into little-endian representation.

My conclusion is that osbpy.read() can only handle big-endian files for SAC format (independently of  byteorder option settings)

Both files are provided in attachement (same record, produced by the same routine, only byteordering differs).

[endian_read_issue_obspy.zip](https://github.com/obspy/obspy/files/2773260/endian_read_issue_obspy.zip)

jkmacc-LANL:
### What does this PR do?

This correctly passes the `byteorder` keyword argument from `obspy.read(..., format='SAC')` through `_internal_read_sac` to `SACTrace.read`.

~~Follows #2285~~ . Closes #2285

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/sac/core.py
+++ b/obspy/io/sac/core.py
@@ -354,7 +354,7 @@ def _read_sac(filename, headonly=False, debug_headers=False, fsize=True,
 
 
 def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
-                       **kwargs):  # @UnusedVariable
+                       byteorder=None, **kwargs):  # @UnusedVariable
     """
     Reads an SAC file and returns an ObsPy Stream object.
 
@@ -377,6 +377,11 @@ def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
     :param fsize: Check if file size is consistent with theoretical size
         from header. Defaults to ``True``.
     :type fsize: bool
+    :param byteorder: If omitted or None, automatic byte-order checking is
+        done, starting with native order. If byteorder is specified and
+        incorrect, a :class:`SacIOError` is raised. Only valid for binary
+        files.
+    :type byteorder: str {'little', 'big'}, optional
     :rtype: :class:`~obspy.core.stream.Stream`
     :return: A ObsPy Stream object.
     """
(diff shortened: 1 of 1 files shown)


obspy/io/sac/core.py
```diff
@@ -377,6 +377,11 @@ def _internal_read_sac(buf, headonly=False, debug_headers=False, fsize=True,
     :param fsize: Check if file size is consistent with theoretical size
         from header. Defaults to ``True``.
     :type fsize: bool
+    :param byteorder: If omitted or None, automatic byte-order checking is
+        done, starting with native order. If byteorder is specified and
+        incorrect, a :class:`SacIOError` is raised. Only valid for binary
+        files.
+    :type byteorder: str {'little', 'big'}, optional
```
megies: Btw, we have `<` and `>` in most (all other?) parts of the codebase.. maybe these should be accepted as aliases here?

On an even wider scope, maybe all across obspy `little`/`<` and `big`/`>` should be accepted interchangeably, but this is probably something for a separate PR and against master then.

N.B.: There is some related stuff in `io/sac/util.py` in `is_same_byteorder` and also some stuff in `io/sac/core.py` where even `0` and `1` are treated as byteorder indicators (also in some mseed code). But like I said, maybe postpone for different PR.

megies:
Could use a minimalistic regression test (e.g. just test that an `obspy.read(sacfile, byteorder=nondefault)` returns a stream object. Otherwise good to go.

megies:
Oh and a changelog one-liner please :-)

obspy/io/sac/core.py
jkmacc-LANL: I'm in support of this, and I'm happy to start an empty PR so it doesn't get lost.

krischer:
> It looks like these failures are unrelated to this PR. Is that correct?

Seems to be.

megies:
added changelog, see c728b62f70b89b32c1c30df52c7d1c4c5a835669

jkmacc-LANL:
That was quick.  Sorry, I had some deadlines this week.

PR #2314, opened 2019-02-14
Allow empty network codes and also don't force source and networks to be set

### LINKED ISSUE #2307: Network: Enable using an empty string as network code?
Currently it's not possible to set an empty string as network code (due to an `if not value:` in the setter), I think this should probably be possible. Similar `if not ...:` checks might also affect other places in `core.inventory` I guess.

```python
from obspy.core.inventory import Network
Network(code='')
```

```
/home/megies/git/obspy/obspy/core/inventory/util.pyc in code(self, value)
     77         if not value:
     78             msg = "A Code is required"
---> 79             raise ValueError(msg)
     80         self._code = str(value).strip()
     81 

ValueError: A Code is required
```

krischer:
Fixes #2307.

Also allows to intialize `Inventory` objects without explicitly setting the source and networks. The source defaults to `ObsPy VERSION` which should be okay. Does not break any existing codes but result in more convenient to use objects.

### DIFF
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -105,8 +105,8 @@ class Inventory(ComparingObject):
 
     In essence just a container for one or more networks.
     """
-    def __init__(self, networks, source, sender=None, created=None,
-                 module=SOFTWARE_MODULE, module_uri=SOFTWARE_URI):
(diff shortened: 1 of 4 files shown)


krischer:
Failures are unrelated and we have to fix this elsewhere.

Issue #2094, opened 2018-03-23
FDSN Client  wadl, catalogs, and contributors query frequency


nick-iris:
When a FDSN Client is instantiated for the first time it makes a request for each web service WADL as well as the event services catalogs and contributors data. This is placing a heavy load on the database since collecting the information for the contributors and catalogs is not an efficient query.

Would it be possible to not request the event catalog and contributors information when a FDSN Client is first created, unless explicitly requested by the user?

**Example:**
Creating a client for the first time makes several wadl requests and a request for the event catalogs and contributors.
```
>>> c = Client("IRIS", debug=True)
Base URL: http://service.iris.edu
Request Headers: {u'User-Agent': u'ObsPy/1.0.3.post0+1835.g80139da8b6.obspy.fdsnws.routing (Darwin-16.7.0-x86_64-i386-64bit, Python 2.7.10)'}
Downloading http://service.iris.edu/fdsnws/dataselect/1/application.wadl with requesting gzip compression
Downloading http://service.iris.edu/fdsnws/event/1/application.wadl with requesting gzip compression
Downloading http://service.iris.edu/fdsnws/station/1/application.wadl with requesting gzip compression
Downloading http://service.iris.edu/fdsnws/event/1/catalogs with requesting gzip compression
Downloading http://service.iris.edu/fdsnws/event/1/contributors with requesting gzip compression
Uncompressing gzipped response for http://service.iris.edu/fdsnws/station/1/application.wadl
Uncompressing gzipped response for http://service.iris.edu/fdsnws/event/1/application.wadl
Uncompressing gzipped response for http://service.iris.edu/fdsnws/dataselect/1/application.wadl
Downloaded http://service.iris.edu/fdsnws/event/1/application.wadl with HTTP code: 200
Downloaded http://service.iris.edu/fdsnws/station/1/application.wadl with HTTP code: 200
Downloaded http://service.iris.edu/fdsnws/dataselect/1/application.wadl with HTTP code: 200
Downloaded http://service.iris.edu/fdsnws/event/1/catalogs with HTTP code: 200
Uncompressing gzipped response for http://service.iris.edu/fdsnws/event/1/contributors
Downloaded http://service.iris.edu/fdsnws/event/1/contributors with HTTP code: 200
Discovered station service
Discovered event service
Discovered dataselect service
Storing discovered services in cache.
```

[...]

PR #2325, opened 2019-02-14
_discover_services keyword in FDSN Client

jkmacc-LANL:
### What does this PR do?

This PR adds a `discover_services` boolean keyword to `obspy.clients.fdsn.Client.__init__`.

`Client._discover_services()` can be an expensive operation for a data
center, but it was run on `Client.__init__` even if it wasn't used.  This
keyword bypasses the service discover/query, and instead assumes the
default services at the data center.  This is helpful for distributed
applications that instantiate many clients in seperate Python processes.

See #2094 for a short discussion.  (Note: an additional PR is necessary to end that issue.)

This PR branches from a recent `master` branch.  

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -30,6 +30,10 @@ master:
    * Added wildcard and url support to read_inventory (see #2326).
    * Modified stream.get_gaps() to deal with overlaps correctly (see #1403)
  - obspy.clients.fdsn:
+   * Add new `_discover_services` boolean flag to the Client, which allows the
+     Client to skip the initial services query at instantiation.  This can
+     reduce the load on service providers, but skips checks against unsupported
+     query parameters.
    * Adding more location codes to the default priority list in the mass
      downloader (see #2155, #2159).
    * The mass downloader now raises a warning if all channels from a station
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -140,7 +140,7 @@ class Client(object):
     def __init__(self, base_url="IRIS", major_versions=None, user=None,
                  password=None, user_agent=DEFAULT_USER_AGENT, debug=False,
                  timeout=120, service_mappings=None, force_redirect=False,
-                 eida_token=None):
+                 eida_token=None, discover_services=True):
         """
         Initializes an FDSN Web Service client.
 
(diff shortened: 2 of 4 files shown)


jkmacc-LANL:
Thanks, Lion.  

* Yeah, I agree.  I'll do that.
* I don't know mocks, really.   I was working on a solution that stuck a fake `.services` dictionary onto the client that contained some reasonable default parameters.  Not my greatest work 💩 

Both suggestions are workarounds for the real solution, however.  What do you think it would take to check for non-default parameters from the user?  I'm willing to try something out if it can avoid doing something would have to be undone.

krischer:
Yea - I think the simplest way for a proper solution would be to just check if only default parameters have been passed to all the functions and then do what you already did and otherwise call the normal `._discover_services()` method. Some more logic is probably required but that should do the trick.

jkmacc-LANL:
Ok, so:

1. Skip `._discover_services()` at `__init__`.
2. In each method, skip `._discover_services` if the passed params are the defaults for that method/service, otherwise call `._discover_services` and do the normal `if "service_name" in self.services` check.

I wonder if I could do the lazy services check @nick-iris requested by adding a keyword like  `._discover_services(service='dataselect')`, thereby splitting up the WADL checks into their respective methods and avoiding unused service queries.

jkmacc-LANL:
> check if only default parameters have been passed to all the functions and then do what you already did and otherwise call the normal ._discover_services() method

I just realized that `._discover_services` would still be called if non-default parameters were specified, which doesn't bypass the potentially expensive service discovery queries if non-default parameters were desired.  The PR is attempting to provide a "fast, unsafe, production" mode for the `Client`.  

I think I'm now leaning back towards the original `discover_services=True` flag that, if `False` simply replaces `self.services` with sensible defaults.  I don't know if [`_create_url_from_parameters`](https://docs.obspy.org/_modules/obspy/clients/fdsn/client.html#Client._create_url_from_parameters) would break, but maybe worth a shot.

krischer:
> I think I'm now leaning back towards the original `discover_services=True` flag that, if `False` simply replaces `self.services` with sensible defaults. I don't know if [`_create_url_from_parameters`](https://docs.obspy.org/_modules/obspy/clients/fdsn/client.html#Client._create_url_from_parameters) would break, but maybe worth a shot.

Fine with me as long we the argument is called `_discover_services=True` and its documented that we don't commit to maintaining the API.

I unfortunately won't have time to look at this properly during the next couple of days.

jkmacc-LANL:
+TESTS:clients.fdsn

megies:
> +TESTS:clients.fdsn

need to update the PR to make this have an effect (i usually amend the last last commit and force push)

jkmacc-LANL:
Could I put it in the commit message, too?

megies:
@jkmacc-LANL currently, no. https://github.com/obspy/obspy_github_api/blob/master/obspy_github_api/obspy_github_api.py#L45-L60

jkmacc-LANL:
Thanks @megies .  

PS.  I don't think I currently have any tasks on this PR.  Please let me know if you intend me to make further changes, otherwise I'm hoping it might be considered for 1.2.0.

megies:
Rebased and force-pushed to get fresh CI

krischer:
Adapted some tests to the latest changes in the master.

PR #2309, opened 2019-02-15
Inventory.plot_response: add option for epoch times in legend

megies:
Add option to include channel epoch start/end date in legend labels of `Inventory.plot_response(.., epoch_dates=False)`. This could make the plot useful for plotting a single NSLC but with multiple epochs (instrumentation changes).


megies:
Should check the inventory landing page if the example plot renders OK.

megies:
> Consider putting it in the gallery.

Hmm yeah, good idea.

megies:
Only need to check gallery when docs buildbot is finished, otherwise ready to merge.

megies:
Plot shows up in gallery just fine, merging.

PR #2337, opened 2019-02-18
Add support for the "today" argument for UTCDateTime

ThomasLecocq:
### What does this PR do?

It adds the possibility to pass "today" to UTCDateTime in order to obtain a UTCDateTime object initialised at the beginning of the current day. The result is similar as:

```python
UTCDateTime(UTCDateTime.date))
```

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -1522,6 +1522,15 @@ class UTCDateTimeTestCase(unittest.TestCase):
             msg = "'%s' does not start with a 4 digit year" % value
             self.assertEqual(msg, e.exception.args[0])
 
+    def test_today(self):
+        """
+        Tests the parsing of 'today' as argument.
+        """
+        now = UTCDateTime.now()
+        t = UTCDateTime(now.year, now.month, now.day)
+        nt = UTCDateTime("today")
+        self.assertEqual(t, nt)
+
 
 def suite():
     return unittest.makeSuite(UTCDateTimeTestCase, 'test')
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -207,6 +207,11 @@ class UTCDateTime(object):
         >>> UTCDateTime(year=1970, month=1, day=1, hour=48, strict=False)
         UTCDateTime(1970, 1, 3, 0, 0)
 
+    (8) Using the magic "today" argument to get the UTCDatetime for the
+        beginning of today:
+
+        >>> today = UTCDateTime("today")
+
     .. rubric:: _`Precision`
 
     The :class:`UTCDateTime` class works with a default precision of ``6``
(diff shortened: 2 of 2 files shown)


megies:
Hmm.. how about `UTCDateTime.today()` which would be inline with `datetime` and less magicky?

ThomasLecocq:
because:

```
In [5]: datetime.datetime.today()
Out[5]: datetime.datetime(2019, 2, 18, 16, 34, 21, 233135)
```
is misleading, according to me...

d-chambers:
Hmmmm... out of curiosity what is the use case for this?

ThomasLecocq:
i.e. for 

- selecting data from a data center for everything since this morning
- building a interator over datetime slices, starting at the beginning of the day

d-chambers:
Ok makes sense. I like the idea @megies mentioned better than allowing yet another input to the standard constructor for two reasons: 

1. UTCDateTime already takes a whole lot (too many) of inputs of different types and meanings resulting in an init method that is very long and complicated.

2. One may also want the start of day of an arbitrary date, not just today. A method would make it easier to generalize this functionality.

Maybe a 'start_of_day' method or property that returns a new UTCDateTime object at the start of the date represented by the data would be a cleaner solution? Then if you want today all you would have to do is 'UTCDateTime().start_of_day', or something like that.

ThomasLecocq:
@d-chambers I fully agree with your comments, and indeed, the "start of the day" for any day is interesting in different cases!

megies:
@d-chambers the `.start_of_day()` seems like a good compromise. :+1:

ThomasLecocq:
OK, seem indeed more interesting to have the ``start_of_day``, more generally useful indeed.

I also created the ``end_of_day`` to be 1ns before midnight, as it could, e.g. be useful and intuitive for 

``` python
c = Client("IRIS")
c.get_waveforms("BE", "MEM", "", "HHZ", UTCDateTime().start_of_day(), UTCDateTime().end_of_day())
```

but this is open for discussion!

megies:
I really don't know about the end-of-day thingy. Feels a bit like introducing floating point glitches deliberately (although we're integer based for time internally) and also feels like redundancy? I like to use start of next day as end time in such situations, feels cleaner for me. If there's really a strong need in any situation it's still possible to just slap a ` - 1` on the time to make sure it's the earlier date..?

ThomasLecocq:
@megies  yep good point :)

d-chambers:
Cool, looks good.

Just a thought, would it ever be useful to have a start of month, year, minute or hour? If so maybe this should be made general like so: 

```python
UTCDateTime().start_of("day")
UTCDateTime().start_of("month") 
...
```
where the main argument would support year, month, day, hour, minute, second.

If day is the most used, however, maybe this isn't needed. It's up to you since this is your PR :wink:  

I do agree that end of day stuff probably isn't needed.

markcwill:
Agree with using a method/property on the time object rather than constructor, and with @d-chambers that a generic solution might be helpful in the future with this as a specific case. Couple notes on prior art:

1. Go `time.Time` has a "Truncate" method that does exactly this and it is awesome, we use it all the time for this type of "time rounding" for buckets, batching, etc: https://golang.org/pkg/time/#Time.Truncate
2. A slightly different approach is in `numpy.datetime64`, which IMHO `UTCDateTime` should inherit/embed as the underlying time store someday? They allow casting of units to get a similar effect: https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html

Looks like you have a `UTCDateTime.start_of_day` sugar property method that does what you want, can always implement for example `UTCDateTime.truncate('day')` later and call that under the hood in a refactor.

PR #2340, opened 2019-02-26
fix #2339

### LINKED ISSUE #2339: Empty EventDescription results with text values of "None"
Running on current obspy master on Ubuntu 16, python 3.7

Empty `EventDescription` objects prevents a catalog from being saved to disk via quakeml and still evaluating to equal.

```python
import io
import obspy.core.event as ev
from obspy import read_events

# create a catalog  with an empty event description
empty_description = ev.EventDescription()
cat1 = read_events()
cat1[0].event_descriptions.append(empty_description)

# serialize the catalog using quakeml and re-read
bio = io.BytesIO()
cat1.write(bio, 'quakeml')
bio.seek(0)
cat2 = read_events(bio)

# the two catalog are not equal
assert cat1 == cat2 

# because one event_description has text of 'None' and the other of None
assert cat1[0].event_descriptions[-1].text == cat2[0].event_descriptions[-1].text
```

d-chambers:
### What does this PR do?

A simple and quick fix to issue #2339.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -52,6 +52,8 @@ master:
      (see #2104, #2090, #2093, #1872).
    * Skip invalid enumeration values during reading but raise a warning.
      (see #2106, #2098, #2095)
+   * Catalogs with empty event description objects can be round-tripped (see
+     #2339, #2340).
  - obspy.io.sac:
    * Fix bug writing inventory with SOH channels to SACPZ (see #2200).
  - obspy.io.segy:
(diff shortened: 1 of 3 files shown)


krischer:
Thanks for this! I added a changelog entry. I guess it is questionable if we want to write the empty event description tag in the first place but I agree that the ability to round-trip to QuakeML and back has a higher priority here.

PR #2344, opened 2019-03-01
Explicitly closing the thread pool for the fdsn routing client

### LINKED ISSUE #2342: fdsn EIDA routed client: problem with spammed subprocesses?
Issue reported on the users mailing list:

```
I am facing a potential issue with the ObsPy FDSN client. I am using the EIDA
routing client to download waveform data and metadata for many stations. The 
way I do it is a simple loop over all stations, using the get_waveforms()
function in the loop. In principle it looks like this:

from obspy.clients.fdsn import RoutingClient as Client
client = Client('eida-routing')
for station in long_list :
    data = client.get_waveforms(...)
    data.write(filename)

The problem is, that each time data is actually downloaded, ObsPy seems to
start several new processes in the background, which are not being closed until
the end of the python script. Thus, for each station in the for-loop new 
additional processes are started but not closed again. Since I have a very long
stationlist and many days that I loop over, basically I end up spawning more
processes than the systen can handle, ultimately resulting in the message:
"Can't start new thread"

This problem only occurs with the fdsn routing client, but e.g. not with the 
ArcLink client. I did not try the non-routed fdsn client. Same behaviour with
ObsPy versions 1.1.0 and 1.1.1 ... 
```

krischer:
Fixes #2342 (I hope).

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -37,6 +37,8 @@ master:
    * make sure that streams fetched via FDSN are properly trimmed to user
      requested times if data center serves additional data around the start/end
      (see #1887, #2298)
+   * fix a problem that could spam subprocesses that were not closed in routed
+     clients (see #2342 and #2344)
  - obspy.io.nordic:
    * Add ability to read and write focal mechanisms and moment tensor
      information. (see #1924)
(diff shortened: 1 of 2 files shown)


megies:
restarting appveyor

megies:
CI didn't run FDSN tests.. but I'll merge this now anyway, looks sane and if we run into problems we can fix them later. We need to run network tests before releasing anyway

PR #2345, opened 2019-03-06
fix bug #1403 (Stream.get_gaps() giving wrong results for some files)

### LINKED ISSUE #1403: Stream.get_gaps() giving wrong results for some files
See details in https://github.com/obspy/obspy/pull/1141#issuecomment-218146707.

megies:
### What does this PR do?

Fix #1403 

See #1403  (Stream.get_gaps() giving wrong results for some files)

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -829,6 +829,24 @@ class Stream(object):
             # skip if is equal to delta (1 / sampling rate)
             if same_sampling_rate and nsamples == 0:
                 continue
+            # check if gap is already covered in trace before:
+            covered = False
+            # you only need to check previous traces because the traces are sorted
+            for prev_i in range(_i):
+                prev_stats = self.traces[prev_i].stats
+                # look if trace is contained in other trace
+                if not (prev_stats['starttime'] < stime < etime < prev_stats['endtime']):
+                    continue
+                # don't look in traces of other measurements
+                elif (stats["network"] == prev_stats["network"] and
+                      stats["location"] == prev_stats["location"] and
+                      stats["station"] == prev_stats["station"] and
+                      not (stats["channel"] == prev_stats["channel"])):
+                    continue
+                else:
+                    covered = True
+            if covered:
+                continue
             gap_list.append([stats['network'], stats['station'],
                              stats['location'], stats['channel'],
                              stime, etime, delta, nsamples])
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -831,20 +831,20 @@ class Stream(object):
                 continue
             # check if gap is already covered in trace before:
             covered = False
-            # you only need to check previous traces because the traces are sorted
-            for prev_i in range(_i):
-                prev_stats = self.traces[prev_i].stats
+            # only need to check previous traces because the traces are sorted
+            for prev_trace in self.traces[:_i]:
+                prev_stats = prev_trace.stats
                 # look if trace is contained in other trace
-                if not (prev_stats['starttime'] < stime < etime < prev_stats['endtime']):
+                prev_start = prev_stats['starttime']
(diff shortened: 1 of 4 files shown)


obspy/core/stream.py
```diff
@@ -829,6 +829,24 @@ def get_gaps(self, min_gap=None, max_gap=None):
             # skip if is equal to delta (1 / sampling rate)
             if same_sampling_rate and nsamples == 0:
                 continue
+            # check if gap is already covered in trace before:
+            covered = False
+            # you only need to check previous traces because the traces are sorted
+            for prev_i in range(_i):
+                prev_stats = self.traces[prev_i].stats
```
megies: i think something like this but be a bit prettier:

```python
for tr in self.traces[:_i]:
    ...
```

Usually we want to avoid iterating over lists using an index number unless it can't be helped

obspy/core/stream.py
```diff
@@ -829,6 +829,24 @@ def get_gaps(self, min_gap=None, max_gap=None):
...
+                if not (prev_stats['starttime'] < stime < etime < prev_stats['endtime']):
+                    continue
+                # don't look in traces of other measurements
+                elif (stats["network"] == prev_stats["network"] and
+                      stats["location"] == prev_stats["location"] and
+                      stats["station"] == prev_stats["station"] and
+                      not (stats["channel"] == prev_stats["channel"])):
+                    continue
+                else:
+                    covered = True
```
megies: If I understand correctly, a `break` after this could avoid looking at more traces when the gap is already confirmed to be a non-gap actually?

obspy/core/stream.py
```diff
@@ -829,6 +829,24 @@ def get_gaps(self, min_gap=None, max_gap=None):
...
+            if covered:
+                continue
             gap_list.append([stats['network'], stats['station'],
                              stats['location'], stats['channel'],
                              stime, etime, delta, nsamples])
```
megies: Just from looking at this I have a feeling this could be prettified a bit with a `for/else` loop. But maybe I'm wrong. The `covered = False` and the final `if covered:` are not on the same indentation level which is confusing me abit.

megies:
@DominikStr thanks for looking into this.. some comments:
 - you can simply open a PR by yourself next time like I did here
 - test case looks good! :rocket: 
 - please fix PEP8 (at the bottom of CircleCI failure log, some long lines and an additional blank line)
 - for PEP8 please check out http://docs.obspy.org/master/coding_style.html?highlight=flake8 which has some instructions/links how to set up pre-commit hooks which help to not make commits locally that will result in PEP8 fails later on
 - your code block looks like it could be "prettified" a bit with a `for:` / `break` / `else:` loop to replace the `covered` flag, but I didn't look at it in detail, so maybe I'm wrong here
 - please add yourself to this list https://github.com/obspy/obspy/blob/master/obspy/CONTRIBUTORS.txt (unless you don't want to appear there ;-) )
 - please add a short changelog entry

Otherwise this looks good to me and I think we can merge this if existing tests and new test are passing.

obspy/core/stream.py
DominikStr: I found no way to prettify it with a `for/else` loop without making it way too complicated.
The `covered = False` and the final `if covered:` are on the same indentation level as far as I can see.

megies: OK then leave it as is :+1:

megies:
@DominikStr there's merge conflicts now, probably just trivial things like changelog. Can you please rebase on current master? Just ping me here if it's unclear what to do.

megies:
I rebased this right now, so we can get rid of finished business in the release project board.. thanks @DominikStr!

I didn't have permission to force push to your branch, so I quickly replaced the PR with a new one #2355

PR #2347, opened 2019-03-06
simple AIC P-picker implementation (Maeda 1985)

mbagagli:
### What does this PR do?
This PR suggest to add a new function in `obspy.signal.trigger.py` module to calculate a first phase-arrival directly from the data without using using the auto regressive approach, by means of using narrower time-windows. The method is explained in Maeda 1985 and adopted in several other works. The article is in Japanes though the abstract is in english).

At the end are just few lines of code, but this is one of  the most used method to recognize phase-arrival and I think a library like obspy should provide an internal function directly.
Suggestion for a faster implementation are welcome.


ThomasLecocq:
I don't know the method but the code looks good. Could you maybe add a paragraph to this page:
https://docs.obspy.org/tutorial/code_snippets/trigger_tutorial.html
?

@jwassermann could maybe comment on the implementation of the AIC?

krischer:
@billy4all This requires a rebase on the latest master and can you please have a look at the comment by @ThomasLecocq 

@jwassermann Any chance for you to have a look?

billy4all:
@krischer @ThomasLecocq  Thanks for the suggestion! I will dig into asap. Sorry for my late reply

ThomasLecocq:
+DOCS

👍 @jwassermann  :-)

mbagagli:
Hello, I still don't understand the failure of the tests. I did a rebase, but no success. Any idea?

misc/docs/source/tutorial/code_snippets/trigger_tutorial.rst
```diff
@@ -396,6 +397,27 @@ For :func:`~obspy.signal.trigger.ar_pick`, input and output are in seconds.
 This gives the output 30.6350002289 and 31.2800006866, meaning that a P pick at
 30.64s and an S pick at 31.28s were identified.
 
+AIC - Akaike Information Criterion
+===========
```
megies: The line of `=` signs should match the header line

misc/docs/source/tutorial/code_snippets/trigger_tutorial.rst
```diff
@@ -396,6 +397,27 @@ For :func:`~obspy.signal.trigger.ar_pick`, input and output are in seconds.
...
+
+For :func:`~obspy.signal.trigger.aic`, the input is a numpy.ndarray and the
+output is in sample.
+
+    >>> from obspy.core import read, UTCDateTime
+    >>> from obspy.signal.trigger import aic
+    >>> trace = read("https://examples.obspy.org/ev0_6.a01.gse2")[0]
+    >>> df = trace.stats.sampling_rate
+    >>> trace_s = trace.slice(UTCDateTime('1970-01-01T01:00:31.6'),
+                              UTCDateTime('1970-01-01T01:00:34.3'))
```
megies: This should be:

```
    ...                        UTCDateTime('1970-01-01T01:00:34.3'))
```
i.e. 3 dots where the Python prompt arrows are in the first line

misc/docs/source/tutorial/code_snippets/trigger_tutorial_aic.py
```diff
@@ -0,0 +1,9 @@
+from obspy.core import read, UTCDateTime
+from obspy.signal.trigger import aic
+trace = read("https://examples.obspy.org/ev0_6.a01.gse2")[0]
+df = trace.stats.sampling_rate
+trace_s = trace.slice(UTCDateTime('1970-01-01T01:00:31.6'),
+                      UTCDateTime('1970-01-01T01:00:34.3'))
+p_idx, aic_f = aic(trace_s.data)
+print(p_idx / df)
+print(UTCDateTime('1970-01-01T01:00:31.6')+(p_idx / df))
```
megies: please have single spaces around the `+` operator to adhere to pep8

obspy/signal/tests/test_trigger.py
```diff
@@ -95,6 +95,17 @@ def test_pk_baer_cf(self):
...
+    def test_aic(self):
+        """
+        Test AIC function aganist implementation for UNESCO short course
+        """
+        filename = os.path.join(self.path, 'manz_waldk.a01.gz')
+        with gzip.open(filename) as f:
+            data = np.loadtxt(f, dtype=np.float32)
+        idx, aic_cf = aic(data)
+        self.assertEqual(idx, 60868)
+        self.assertEqual(len(aic_cf), 119999)
```
megies: maybe also test the values of the cf? compare other tests, e.g. with `np.testing.assert_array_almost_equal`

megies:
There is a weird mismatch of cf length in one of the test builds: https://travis-ci.org/obspy/obspy/jobs/640363478#L1214

obspy/signal/trigger.py
```diff
@@ -428,6 +428,46 @@ def pk_baer(reltrc, samp_int, tdownmax, tupevent, thr1, thr2, preset_len,
...
+            try:
+                var2 = np.log(np.var(td[ii:]))
+            except FloatingPointError:  # if var==0 --> log is -inf
+                var2 = 0.00
+        #
+        val1 = ii*var1
+        val2 = (td.size-ii-1)*var2
+        aic_cf[ii-1] = (val1+val2)
+    # --------------------  Index search
+    idx = sorted(range(len(aic_cf)), key=lambda k: aic_cf[k])[0]
```
megies: This is just fetching the index of the highest value, correct? Then `np.argmax(aic_cf)` should do it.

obspy/signal/trigger.py
```diff
@@ -428,6 +428,46 @@ def pk_baer(reltrc, samp_int, tdownmax, tupevent, thr1, thr2, preset_len,
...
+    directly from the data, without using the AR coefficients (Maeda, 1985)
+
+    Computes P-phase arrival time digital single-component acceleration
+    or broadband velocity record without requiring threshold settings using
+    AKAIKE INFORMATION CRITERION.
+    Returns P-phase arrival time index and the characteristic function.
+    The returned index correspond to the characteristic function's minima.
+
+    :type td: numpy.ndarray
+    :param td: time series as numpy.ndarray float32 data
```
megies: It should not matter what dtype the data comes in, does it?

obspy/signal/trigger.py
```diff
@@ -428,6 +428,46 @@ def pk_baer(reltrc, samp_int, tdownmax, tupevent, thr1, thr2, preset_len,
...
+    Differently from AR-AIC picker, this function calculates AIC function
+    directly from the data, without using the AR coefficients (Maeda, 1985)
+
+    Computes P-phase arrival time digital single-component acceleration
+    or broadband velocity record without requiring threshold settings using
+    AKAIKE INFORMATION CRITERION.
+    Returns P-phase arrival time index and the characteristic function.
+    The returned index correspond to the characteristic function's minima.
+
+    :type td: numpy.ndarray
```
megies:
```
:class:`numpy.ndarray`
```

obspy/signal/trigger.py
```diff
@@ -428,6 +428,46 @@ def pk_baer(reltrc, samp_int, tdownmax, tupevent, thr1, thr2, preset_len,
...
+                var1 = 0.00
+            #
+            try:
+                var2 = np.log(np.var(td[ii:]))
+            except FloatingPointError:  # if var==0 --> log is -inf
+                var2 = 0.00
+        #
+        val1 = ii*var1
+        val2 = (td.size-ii-1)*var2
+        aic_cf[ii-1] = (val1+val2)
```
megies: operators (`+-*` etc) should have single spaces around them, not sure why our CI does not complain

obspy/signal/trigger.py
```diff
@@ -428,6 +428,46 @@ def pk_baer(reltrc, samp_int, tdownmax, tupevent, thr1, thr2, preset_len,
...
+    :param td: time series as numpy.ndarray float32 data
+    :return: (idx, aic_cf) idx sample number of parrival; aic_cf numpy.ndarray
+        containing the values of the characteristic function.
+
+    .. seealso:: [Maeda1985]_
+    """
+    # --------------------  Creation of the characteristic function
+    aic_cf = np.zeros(td.size-1)
+    for ii in range(1, td.size):
+        with np.errstate(divide='raise'):
```
megies: might be better to put this outside of the for-loop, otherwise numpy might be resetting it's settings twice per loop

obspy/signal/trigger.py
mbagagli: it's actually searching the minimum, but yes, using `np.argmin` is much way better. Dunno why I've left that this way, sorry for that

obspy/signal/trigger.py
mbagagli: You are right, it was because of old assertion I had in my project. Now the function checks if the input time-series is an `np.ndarray` class. No checks on dtypes anymore. Docstring updated.

obspy/signal/trigger.py
mbagagli: updated

obspy/signal/trigger.py
mbagagli: updated

obspy/signal/trigger.py
mbagagli: It is indeed better and computational faster, thanks for the comment

misc/docs/source/tutorial/code_snippets/trigger_tutorial_aic.py
mbagagli: done, updated

misc/docs/source/tutorial/code_snippets/trigger_tutorial.rst
mbagagli: updated, I was missing this

misc/docs/source/tutorial/code_snippets/trigger_tutorial.rst
mbagagli: done

mbagagli:
> There is a weird mismatch of cf length in one of the test builds: https://travis-ci.org/obspy/obspy/jobs/640363478#L1214

This was because I typed `len(data - 1)` while it should have been `len(data) - 1`. It was unfortunately a typo. I have anyway updated the test with a shorter trace (faster on testing).
To assert the caractheristic function, though, I had to add a numpy txt matrix on the test file path. I've stored the file with `numpy.save` function, with the option `allow_pickle=False` for better portability.

obspy/signal/tests/test_trigger.py
mbagagli: test on cf added, see comment below

megies:
@mbagagli these test files are too big (close to 1MB in total), can you use some data that is already in the repository? Would the example stream work  (`read()`)? If not we have some other test data files already with events in them

megies:
Also eventually this will have to be rebased cleanly on master of #2577 is merged.. but don't worry about it too much, we can help with that

mbagagli:
> @mbagagli these test files are too big (close to 1MB in total), can you use some data that is already in the repository? Would the example stream work (`read()`)? If not we have some other test data files already with events in them

Yes of course , I was just testing the comparison of AIC Carachteristic Function (CF) numpy.array stored in 2 different ways (txt, npy). If we want to compare also the CF I need to include the least bigger file (npy) that is just 257K.
 I'm already using (reading) a SAC files from `obspy/signal/tests/data` folder. 
Even if lowering the `assert_array_almost_equal` precision from 6 to 4 doesn't help to  2 arrays were used mainly for a test (the numpy array in txt) to see if lowering the decimals (6 to 4) would help on the CF assertions during tests. 
For what it worth though, the returned index is always the same (meaning the lowest value of the CF remains the same) and the functions return as expected.
Is the `assert_array_almost_equal` that seems to create problems in all the builds. 
The generated errors include the python 3.6 interpreter where personally I encountered 0 problem at all. 

Any suggestion on this? The 2 arrays returned by `aic` function  and the stored one are double-precision. I don't know if this could matter in some earlier numpy versions.

mbagagli:
> Also eventually this will have to be rebased cleanly on master of #2577 is merged.. but don't worry about it too much, we can help with that

For this reason (dropping of Python2 support) I was searching the minimum dependecy used in your tests. It results to me that the lowest numpy version used is `numpy 1.8.2` (Python 2.7 architecture). Since `numpy 1.5.0` the recommended function for float-array comparison is [this](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html). I don't think this alone will let the test pass, because I think the problem rely on internal logarithmic calculations. Pushing now

mbagagli:
From the latest tests: 
- (travis): now with the `numpy.assert_allclose` function, all the python3 environment tests are passed (only the 3.5 is broken because of numpy missing during build). Only the  Python2.7 test fails for an actual mismatch
- (appveyor):  the test is failing on 2.7, but seems for some DLL load issue, not related to the PR. The rest are labeled as canceled, don't know why

heavelock:
We are planning to drop support for Py<3.7 so don't worry with the 3.5 failure. The appveyor probably has a fail-fast policy and that's why it cancelled the rest of builds. 
I did a quick PR from your fork against branch of my fork where I was testing Github Actions CI and tests on Windows are passing there, at least on Py3.7.
https://github.com/heavelock/obspy/pull/4/checks?check_run_id=552679753

megies:
> We are planning to drop support for Py<3.7 so don't worry with the 3.5 failure. 


:point_up: @mbagagli check out #2577 no need to worry about any Python < 3.7 or numpy <1.15, see https://github.com/obspy/obspy/pull/2577/commits/bc550632c0b61de4d27972dcdf50d4d3b801b6ca

I will try to get #2577 merged ASAP so that master can be a useful base again for PRs at this point

megies:
>  I'm already using (reading) a SAC files from `obspy/signal/tests/data` folder.

@mbagagli Ok, so maybe we can add some preprocessing to that file to get the size of the CF reduced? (There might be other files somewhere in the tests directories with inherently smaller size you could try instead, didnt search all files)

```python
st = read("/path/to/CRLZ.HHZ.10.NZ.SAC")
st.filter('lowpass', freq=2, corners=4)
st.trim(st[0].stats.starttime + 120)
st.decimate(10, no_filter=True)
st.plot()
# your tests here then
```

AIC

Leave out the trimming if the picker needs the longer background noise signal as a start up. 

Should not affect the test much, but size of the new test data needed will be ~20 times less. Sorry about the bullying but the tarballs are quite blown up already by all the test data, so we try to keep new additions as small as possible.

megies:
@mbagagli this needs to be rebased on master which is Python >=3.7 only now. Ping us if you have problems rebasing, there's quite some potential for merge conflicts with the huge amount of changes due to dropping Py2.

mbagagli:
So, I've unburden the data for testing, rebased, and forced the push. Let me know if there's anything more that you would like to add/remove. Best

d-chambers:
Closed by #2973

PR #2350, opened 2019-03-09
Add high-precision pick time support and long-phase name support

### LINKED ISSUE #2348: reading of seconds on Nordic Type 4 lines
obspy incorrectly reads pick times (specifically seconds) when the seconds are given to millisecond precision (I assume this happens when I use high-accuracy mode, but in any case it is a "standard" output for SEISAN as can be seen in the definition of Nordic files (screenshot attached).

I have attached an example Nordic file: the first and third "events" are in this high-accuracy notation.  I have also attached a script to read this file and print out the pick times so that you can compare for yourself(ves).  I had to add the suffix "txt" so that it could be attached.

The solution is to read in line[22:28] as seconds instead of line[23:28], in obspy.io.nordic.core, I have attached a screenshot of the relevant code from the current git in which the relevant lines are 745 and 749

I am using obspy version 1.1.1, installed using conda (and conda-forge)

nordic_type4line
[correct_10_11.txt](https://github.com/obspy/obspy/files/2940653/correct_10_11.txt)
[nordic_example.py.txt](https://github.com/obspy/obspy/files/2940666/nordic_example.py.txt)
nordic_code_example

calum-chamberlain:
### What does this PR do?
Adds support for high(er)-precision timing in Nordic pick files (closes #2348) and add support for long phase names - from the seisan manual:
> Long phase names: An 8 character phase can be used in column 11-18. There is then not room for polarity information. the weight is then put into column 9. This format is recognised by HYP and MULPLT

Seconds can run in columns 23-28 - I misread the indexing and was reading columns 23-28 rather than 22-28. #2348 pointed out that this is used when seisan is run in "high-accuracy" mode.

Note, pick times are now written out by default in high-accuracy from obspy now. **I need to confirm that seisan is always happy with this**, if not then a flag option for high-accuracy should be implemented.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -158,20 +158,14 @@ def _str_conv(number, rounded=False):
             exponant = int('{0:.2E}'.format(number).split('E+')[-1]) - 1
             divisor = 10 ** exponant
             string = '{0:.1f}'.format(number / divisor) + 'e' + str(exponant)
-    elif rounded == 2 and isinstance(number, (float, int)):
+    elif rounded and isinstance(number, (float, int)):
+        format_str = "{0:." + str(rounded) + "f}"
         if number < 100000:
-            string = '{0:.2f}'.format(number)
+            string = format_str.format(number)
         else:
             exponant = int('{0:.2E}'.format(number).split('E+')[-1]) - 1
             divisor = 10 ** exponant
-            string = '{0:.2f}'.format(number / divisor) + 'e' + str(exponant)
-    elif rounded == 1 and isinstance(number, (float, int)):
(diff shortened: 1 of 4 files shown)


calum-chamberlain:
I think it is going to make more sense to rebase this on master - master has had other nordic related updates, so to make it easier for releasing I may as well make the changes there.

d-chambers:
I think we are done with the 1.1.x releases anyway

calum-chamberlain:
I didn't rebase in the end, too many changes to handle, so I just made the small number of changes here on master and opened a new PR at #2351

PR #2351, opened 2019-03-11
Nordic high accuracy

### LINKED ISSUE #2348: reading of seconds on Nordic Type 4 lines

calum-chamberlain:
### What does this PR do?

Adds support for high(er)-precision timing in Nordic pick files (closes #2348) and add support for long phase names - from the seisan manual:

> Long phase names: An 8 character phase can be used in column 11-18. There is then not room for polarity information. the weight is then put into column 9. This format is recognised by HYP and MULPLT

### Why was it initiated? Any relevant Issues?
Seconds can run in columns 23-28 - I misread the indexing and was reading columns 23-28 rather than 22-28. #2348 pointed out that this is used when seisan is run in "high-accuracy" mode.

Note, pick times are now written out by default in high-accuracy from obspy now. I need to confirm that seisan is always happy with this, if not then a flag option for high-accuracy should be implemented.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -1512,9 +1512,9 @@ def nordpick(event):
             period=_str_conv(peri, rounded=peri_round).rjust(5)[0:5],
             azimuth=_str_conv(azimuth).rjust(6)[0:6],
             velocity=_str_conv(velocity).rjust(5)[0:5],
-            azimuthres= _str_conv(azimuthres).rjust(3)[0:3],
-            timeres= _str_conv(timeres, rounded=2).rjust(5)[0:5],
-            distance= distance.rjust(5)[0:5],
+            azimuthres=_str_conv(azimuthres).rjust(3)[0:3],
+            timeres=_str_conv(timeres, rounded=2).rjust(5)[0:5],
+            distance=distance.rjust(5)[0:5],
             caz=_str_conv(caz).rjust(4)[0:4]))
         # Note that currently finalweight is unsupported, nor is velocity, or
         # angle of incidence.  This is because obspy.event stores slowness in
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -14,13 +14,18 @@ Nordic file format support for ObsPy
     the arrival.pick_id linking the arrival (which contain calculated
     information) with the pick.resource_id (where the pick contains only
     physical measured information).
+
+.. versionchanged:: 1.2.0
+
+    The number of stations used to calculate the origin was previously
+    incorrectly stored in a comment. From version 1.2.0 this is now stored
+    in `origin.quality.used_station_count`
 """
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA @UnusedWildImport
 
 import warnings
-from collections import defaultdict
 import datetime
 import os
 import io
@@ -28,15 +33,16 @@ import io
 from obspy import UTCDateTime, read
 from obspy.geodetics import kilometers2degrees, degrees2kilometers
 from obspy.core.event import (
-    Event, Origin, Magnitude, Comment, Catalog, EventDescription, CreationInfo,
+    Event, Origin, Magnitude, Catalog, EventDescription, CreationInfo,
     OriginQuality, Pick, WaveformStreamID, Arrival, Amplitude,
     FocalMechanism, MomentTensor, NodalPlane, NodalPlanes, QuantityError,
     Tensor, ResourceIdentifier)
+from obspy.io.nordic import NordicParsingError
+from obspy.io.nordic.utils import (
+    _int_conv, _str_conv, _float_conv, _evmagtonor, _nortoevmag,
+    _get_line_tags)
 
 
-MAG_MAPPING = {"ML": "L", "MLv": "L", "mB": "B", "Ms": "s", "MS": "S",
-               "MW": "W", "MbLg": "G", "Mc": "C"}
-INV_MAG_MAPPING = {item: key for key, item in MAG_MAPPING.items()}
 POLARITY_MAPPING = {"": "undecidable", "C": "positive", "D": "negative"}
 INV_POLARITY_MAPPING = {item: key for key, item in POLARITY_MAPPING.items()}
 ONSET_MAPPING = {'I': 'impulsive', 'E': 'emergent'}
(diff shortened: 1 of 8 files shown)


calum-chamberlain:
I checked whether SEISAN was happy reading high-accuracy phase-picks when not run in high-accuracy mode, and it seems to be fine.  Nevertheless a flag has been added for writing to turn off high accuracy mode (which is also tested).

### RE tests:

circleci pep8 tests are not related to files I have changed for this:
```
obspy/core/util/misc.py:746:29: E117 over-indented
obspy/imaging/waveform.py:1481:13: E117 over-indented
```
and it looks like appveyor is failing on response removal things that are unrelated.

I think this is ready for review...

obspy/io/nordic/core.py
```diff
@@ -34,10 +34,16 @@
...
+POLARITY_MAPPING = {"": "undecidable", "C": "positive", "D": "negative"}
+INV_POLARITY_MAPPING = {item: key for key, item in POLARITY_MAPPING.items()}
+ONSET_MAPPING = {'I': 'impulsive', 'E': 'emergent'}
+INV_ONSET_MAPPING = {item: key for key, item in ONSET_MAPPING.items()}
+EVALUATION_MAPPING = {'A': 'automatic', ' ': 'manual'}
+INV_EVALUTATION_MAPPING = {
+    item: key for key, item in EVALUATION_MAPPING.items()}
 # List of currently implemented line-endings, which in Nordic mark what format
 # info in that line will be.
 accepted_tags = ['1', '6', '7', 'E', ' ', 'F', 'M', '3']
```
d-chambers: Mind capitalizing this for consistency? May also be worth making this a tuple just to make sure it doesn't mutate.

obspy/io/nordic/core.py
```diff
@@ -349,21 +344,31 @@ def _read_origin(line):
...
+        new_event.origins[0].time_fixed = True
     new_event.creation_info = CreationInfo(agency_id=line[45:48].strip())
     new_event.origins[0].creation_info = CreationInfo(
         agency_id=line[45:48].strip())
-    ksta = Comment(text='Number of stations=' + line[49:51].strip())
-    new_event.origins[0].comments.append(ksta)
-    if _float_conv(line[51:55]) is not None:
+    used_station_count = line[49:51].strip()
+    if used_station_count != '':
+        new_event.origins[0].comments.append(
```
d-chambers: What is the purpose of this comment if the information is also in the `OriginQuality`?

obspy/io/nordic/core.py
```diff
@@ -723,34 +732,23 @@ def _read_picks(tagged_lines, new_event):
...
-            polarity = "undecidable"
+        polarity = POLARITY_MAPPING.get(polarity, "undecidable")
         # It is valid nordic for the origin to be hour 23 and picks to be hour
         # 00 or 24: this signifies a pick over a day boundary.
-        if int(line[18:20]) == 0 and evtime.hour == 23:
+        pick_hour = int(line[18:20])
+        pick_minute = int(line[20:22])
+        pick_seconds = float(line[22:29])  # 29 should be blank, but sometimes
+        # SEISAN appears to overflow here, see #2348
+        if pick_hour == 0 and evtime.hour == 23:
```
d-chambers: Couldn't you get rid of lines 742 to 748 by just using strict == False in `UTCDateTime`, or was there some reason for not doing that?

obspy/io/nordic/core.py
calum-chamberlain: Good question: it was previously put in as a comment (before this patch) and I left it there just in case anyone was expected it to be there - I would rather remove it though.

obspy/io/nordic/core.py
calum-chamberlain: 742-748 copes with stepping over a day boundary without any change in the day number.  It seems to be valid (I have had files that SEISAN made) that have an origin on one day (say 2010/1/1T23:59:20) and picks on the next day.  Picks in NORDIC do not contain the date, just hour, minutes and decimal seconds, so a pick could be formatted with hour=00 - this if statement should cope with that and know to add a day.
I could probably use `strict=False` to remove the `if pick_hour>=24` statement though, but I figured because I was doing this anyway it could just stay?

obspy/io/nordic/core.py
calum-chamberlain: I think I have actually repeated myself and should remove the non-caps version, thanks for catching it.  And good idea no the tuples :+1:

obspy/io/nordic/core.py
d-chambers: Ah good point. I forgot the pick didn't include day info. No, this is certainly fine the way it is I just didn't understand at first.

obspy/io/nordic/core.py
d-chambers: I don't use seisan anymore but I would be in favor of removing it. Relying on a comment to extract information is a pretty dicey practice anyhow.

calum-chamberlain: Agreed - I think I just didn't see where it should go when I first wrote this.  Keen to remove, I might just put a note in the doc-string as a change for 1.2.0.

calum-chamberlain:
I added a doc-string for the change from comment to the correct place in `origin.quality.used_station_count`. +DOCS

calum-chamberlain:
Docs built fine, the only thing I changed is in the `Changed in version` of [this page](http://docs.obspy.org/pull-requests/2351/packages/autogen/obspy.io.nordic.core.html#module-obspy.io.nordic.core).

Test fails on appveyor are for trace comparison after response removal, the tests that fail are:
- FAIL: test_remove_response (obspy.core.tests.test_stream.StreamTestCase)
- FAIL: test_simulate_seedresp_parser (obspy.core.tests.test_stream.StreamTestCase)

Looks like the version of flake8 installed on circleci has been bumped from 3.6.0 to 3.7.7, which is resulting in a lot of import naming related complaints. Happy to trudge through marking those to be ignored, or changing them today if it helps (in another PR)?

The travis fail looked like a network issue, [I have restarted it](https://travis-ci.org/obspy/obspy/jobs/506018031).

Good to go @d-chambers ?

d-chambers:
Looks good to me, feel free to merge

obspy/io/nordic/core.py
```diff
@@ -714,7 +574,11 @@ def _read_picks(tagged_lines, new_event):
         if len(line) < 80:
             line = line.ljust(80)  # Pick-lines without a tag may be short.
         weight = line[14]
-        if weight == '_':
+        if weight not in ' 012349_':  # Long phase name
```
megies: This might be good to have as a CONSTANT at the top like the other header thingies (e.g. `ALLOWED_WEIGHTS`)

But ignore for now, we don't want another iteration of this PR for reviewing to speed up things.. :-)

megies:
Tests pass and I don't know enough of the nordic implementation details anyway, so I'm happy to follow the previous review, thanks @calum-chamberlain and @d-chambers!

megies:
Oh @calum-chamberlain issue numbers in changelog missing, but I'll add them on master now, so no worries.

Issue #1932, opened 2017-10-14
read_events fails with unicode characters in file path


megies:
```python
cat = read_events('/some/funny/filename_üüöä.xml')
  File "<decorator-gen-42>", line 2, in read_events
  File "/home/megies/git/obspy-master/obspy/core/util/decorator.py", line 334, in _locked_f
    return func(*args, **kwargs)
  File "<decorator-gen-41>", line 2, in read_events
  File "/home/megies/git/obspy-master/obspy/core/util/decorator.py", line 301, in _map_example_filename
    return func(*args, **kwargs)
  File "/home/megies/git/obspy-master/obspy/core/event/catalog.py", line 843, in read_events
    catalog = _read(pathnames[0], format, **kwargs)
  File "<decorator-gen-43>", line 2, in _read
  File "/home/megies/git/obspy-master/obspy/core/util/decorator.py", line 180, in uncompress_file
    elif filename.endswith('.bz2'):
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 36: ordinal not in range(128)
```

QuLogic:
Python 2?

megies:
Yeah, encountered this problem on Python2.

megies:
@DominikStr I think this one is not being worked on right now, maybe you could have a look..

megies:
But maybe this is fixed already by #2326 ?? Dunno.. we need to check

DominikStr:
Following example works for me localy without a problem:  

``` python
from obspy import read_events 
cat = read_events('Földär/töhäküß.xml')
```
So I think it is already fixed. 
Should I write a test case anyway ?

megies:
@DominikStr go ahead and make a PR from your `fix_1932_unicode_characters_in_file_path` branch, so we can see CI test results

PR #2361, opened 2019-03-17
Test Case for Issue #1932

DominikStr:
Adds an example file with umlauts to the test data and a testcase failing if a `UnicodeDecodeError' gets raised.

### What does this PR do?

Contains a test case for issue #1932

See (read_events fails with unicode characters in file path #1932)

### PR Checklist
- [ ] All tests still pass.


megies:
@DominikStr can you please rebase on current master and force-push to this PR? That should get rid of the Circle CI fails.

megies:
Also no need to make a separate issue for the fix, just add the fix here as well. Thanks!

DominikStr:
I force-pushed it but I can't see what else to fix to get it through the circleci test.

megies:
@DominikStr these CircleCI fails are fixed on current master. Can you rebase please and force-push once more? Thanks!

obspy/core/tests/test_event.py
```diff
@@ -223,6 +223,8 @@ def setUp(self):
         self.image_dir = os.path.join(os.path.dirname(__file__), 'images')
         self.iris_xml = os.path.join(path, 'iris_events.xml')
         self.neries_xml = os.path.join(path, 'neries_events.xml')
+        unicode_path = "unicode_char_test_äöüß/unicode_test_catalog_äöüß.xml"
```
krischer: Please also use `os.path.join()` here to make sure it works on all systems.

krischer:
Failures are unrelated - expect the one small issue I have this is IMHO ready to merge.

QuLogic:
Are you sure the test failure is unrelated? It appears to be about "special" characters in file paths, which is what this new file is exercising.

Now, it is unrelated to the _issue_, but I'm not sure it's unrelated to this _PR_.

krischer:
> Are you sure the test failure is unrelated? It appears to be about "special" characters in file paths, which is what this new file is exercising.
> 
> Now, it is unrelated to the _issue_, but I'm not sure it's unrelated to this _PR_.

Yes I think you are right! I did not read the failures carefully enough and the new test file drips up the `obspy.core.util.testing.get_all_py_files()` function.

@DominikStr can you locally test this on Python 2.7 and try to fix it? On Python 3 it will just work as unicode strings are the default.

DominikStr:
I still have trouble setting up my development environment for python 2.7 and obspy. If I manage to get my environment working I will try to fix it.

krischer:
Does `conda create -n obspy_py27 python=2.7 obspy` not work? This should install Python 2.7 with all dependencies of ObsPy. Then just uninstall ObsPy and install the dev version.

DominikStr:
I tried that a few times but I still have my windows problem (`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 28: ordinal not in range(128)`) when I try to use the developer version. The way I did it for 3.x is not working here.

megies:
> I tried that a few times but I still have my windows problem (`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 28: ordinal not in range(128)`) when I try to use the developer version. The way I did it for 3.x is not working here.

At what point is this error popping up? During conda package installations? There are some broken packages out there with bugs in metadata, but just a wild guess..

DominikStr:
The error pops up after installation when I try to import obspy.

PR #2363, opened 2019-03-19
Get rid of the need for deb package build branches

megies:
I need to integrate this commit as a quilt patch, so I don't need a specific deb build branch for the docker farm anymore.

b8a2e529c32fe9394d7364317d27e35cda514179


megies:
Actually maybe I should make the patch less drastic and only remove the `dirty` version suffix parts.. then version number also would not need to be hardcoded.

megies:
Closing, keeping current approach seems easiest to continue as is

PR #2087, opened 2019-03-29
add Inventory.sort()

megies:
It might be useful to have a sorting method on `Inventory`, something like..

```python
inv = read_inventory()
inv.networks = sorted(inv.networks, key=lambda x: x.code)
for net in inv:
    net.stations = sorted(net.stations, key=lambda x: x.code)
    for sta in net:
        sta.channels = sorted(sta.channels, key=lambda x: x.code)
```

Looking at the contents of `inv` before sorting:
```python
for net in inv:
    for sta in net:
        for cha in sta:
            print net.code, sta.code, cha.code
```
```
GR FUR HHZ
GR FUR HHN
GR FUR HHE
GR FUR BHZ
GR FUR BHN
GR FUR BHE
GR FUR LHZ
GR FUR LHN
GR FUR LHE
GR FUR VHZ
GR FUR VHN
GR FUR VHE
GR WET HHZ
GR WET HHN
GR WET HHE
GR WET BHZ
GR WET BHN
GR WET BHE
GR WET LHZ
GR WET LHN
GR WET LHE
BW RJOB EHZ
BW RJOB EHN
BW RJOB EHE
BW RJOB EHZ
BW RJOB EHN
BW RJOB EHE
BW RJOB EHZ
BW RJOB EHN
BW RJOB EHE
```
..and after:
```
BW RJOB EHE
BW RJOB EHN
BW RJOB EHZ
BW RJOB EHE
BW RJOB EHN
BW RJOB EHZ
BW RJOB EHE
BW RJOB EHN
BW RJOB EHZ
GR FUR BHE
GR FUR BHN
GR FUR BHZ
GR FUR HHE
GR FUR HHN
GR FUR HHZ
GR FUR LHE
GR FUR LHN
GR FUR LHZ
GR FUR VHE
GR FUR VHN
GR FUR VHZ
GR WET BHE
GR WET BHN
GR WET BHZ
GR WET HHE
GR WET HHN
GR WET HHZ
GR WET LHE
GR WET LHN
GR WET LHZ
```

We could offer to sort by epoch start or end time or also sampling rate etc..


megies:
> I added you as a collaborator and I will open the pull request as soon as I finish the test case.

No need, I made this issue into a PR now. It will dynamically update whenever you push to your branch again. :-)

krischer:
Instead of basing the test case on two StationXML files it would probably be better to just create some dummy Inventory objects and perform the test on these. Avoids having to commit test files but also is IMHO easier to reason about.

DominikStr:
The new test is a bit long but I tried to keep it as short as possible. I couldn't think of a way to make it more compact.

obspy/core/inventory/inventory.py
```diff
@@ -1117,6 +1117,27 @@ def plot_response(self, min_freq, output="VEL", network="*", station="*",
 
         return fig
 
+    def sort(self):
```
krischer: How about something like this:

```suggestion
    def sort(self, keys=("code", "start_date"), max_level="channel"):
        sort_fct = lambda x: [getattr(x, i) for i in keys]
        self.networks = sorted(self.network, key=sort_fct)
        ...
```

Then it works for all keys and any number of keys (at least those that are present on all object types which could also checked before any in-place sorting takes place). I also added a `max_level` argument so one could for example only sort the networks and not the stations/channels/...

I also think the `.sort()` method should be part of every network/station/... object. If you implement it at the `BaseNode` (https://github.com/obspy/obspy/blob/master/obspy/core/inventory/util.py#L26) this would just work I think.

DominikStr:
I implemented the sort at the base node and changed the inventory sort according to your suggested change. If this is what you have in mind I will add a few lines to deal with wrong input and test cases for the base node implementation.

I'm still not entirely sure what you mean by manually creating the objects. Something like commit b3d7b132f9a3b55b44c787fbbd94424063106cc8 ?

Issue #2356, opened 2019-03-15
MiniSEED output of int64 data


megies:
When creating custom integer data with numpy and trying to write to MiniSEED, we can end up in problems because miniseed has no way to handle `int64` data. In most cases it might be possible to simply downcast the data to `int32` without changing the data before writing it I guess, maybe we should consider doing that @krischer? Or give some better information how to work around this when encountering `int64` data arrays..

```python
import numpy as np
from obspy import Trace, Stream, UTCDateTime


data = [ 
    ("2016-01-07T00:00:00.008300Z", 3010),
    ("2016-01-07T00:00:02.668393Z", 686),
    ("2016-01-07T00:00:09.528393Z", 4086),
    ("2016-01-07T00:00:30.108393Z", 2028),
    ("2016-01-07T00:00:50.388393Z", 6158),
    ("2016-01-07T00:00:57.248393Z", 1370),
    ("2016-01-07T00:01:31.458393Z", 4107),
    ("2016-01-07T00:01:51.968393Z", 2056)]

x = np.arange(30000)
tr = Trace(x)
tr.stats.starttime = UTCDateTime("2016-01-07T00:00:00.008300Z")
tr.stats.sampling_rate = 100 

st = Stream()
for i, (start, numsamp) in enumerate(data):
    tr_ = tr.slice(starttime=UTCDateTime(start))
    tr_.data = tr_.data[:numsamp]
    st.append(tr_)

import pdb; pdb.set_trace()
st.write('/tmp/buggy.mseed', 'MSEED')
```

```
Traceback (most recent call last):
  File "buggy.py", line 26, in <module>
    st.write('/tmp/buggy.mseed')
  File "/home/megies/git/obspy/obspy/core/stream.py", line 1452, in write
    write_format(self, filename, **kwargs)
  File "/home/megies/git/obspy/obspy/io/mseed/core.py", line 791, in _write_mseed
    raise Exception(msg)
Exception: Unsupported data type int64 in Stream[0].data
```

megies:
@LMurrayBergquist you could work on this one. Basically just (in a test case)
 - create a dummy stream with e.g. `np.arange(20, dtype=np.int64)`
 - make sure it can be written to mseed (writing to a `io.BytesIO()` object in the test, search in other test cases how that's done if unclear)
 - in the miniseed writer routine add a check when encountering int64 data for output whether the data is still the same if downcast to int32 and then write as int32 or show a proper error message

[...]

PR #2373, opened 2019-04-04
Mini seed output of int64 data #2356

LMurrayBergquist:
This pull request changes obspy/io/mseed/core.py to allow int64 data to be written to MiniSEED as long as it can be safely downcast to int32. If downcasting the data from int64 to int32 changes the array it is not allowed and results in an error message. A test of writing int64 data to mseed has been added to obspy/io/mseed/tests/test_mseed_reading_and_writing.py. 

This pull request was initiated to fix issue: MiniSEED output of int64 data #2356. 

### PR Checklist
- [ ] All tests still pass.


megies:
I changed base branch for the PR, you inadvertently selected the maintenance branch, but at this point everything goes into `master` (and you properly based your branch on master it seems, so all's good)

obspy/io/mseed/core.py
```diff
@@ -785,6 +785,16 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
                 trace_attr['encoding'] = 1
             elif trace.data.dtype.type == np.dtype(native_str('|S1')).type:
                 trace_attr['encoding'] = 0
+            #int64 data not supported; if possible, downcast to int32, otherwise create error message
+            elif trace.data.dtype.type == np.int64:
```
megies: Looks like the recommended way to compare these is `....type is np.int64` (see https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html first example). Usually it's correct to use `==` but for singletons `is` is preferred usage in Python e.g. `something is None`

obspy/io/mseed/core.py
```diff
@@ -785,6 +785,16 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
...
+                #check if data can be safely downcast to int32
```
megies: PEP8: one space after comment sign (you should consider using something like http://flake8.pycqa.org/en/latest/ or https://github.com/ambv/black to check if your changed code conforms to PEP8)

obspy/io/mseed/core.py
```diff
@@ -785,6 +785,16 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
...
+                trace_data1 = trace.data.copy().astype(np.int32)
```
megies: I think this is copying the data in memory twice, according to numpy docs `trace.data.astype(np.int32, copy=True)` should be fine and only copy once then

obspy/io/mseed/core.py
```diff
@@ -785,6 +785,16 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
...
+                if np.array_equal(trace.data, trace_data1) == True:
```
megies: Just make this test:

`if np.array_equal(trace.data, trace_data1):`

It's shorter and more readable and this is a function only returning True or False. Usually in these cases just relying on this test is fine.

As a side note, in cases where numbers are returned one has to be cautious as `0.0` as a valid number evaluates to `False` same as `None` while and any other valid number evaluates to `True`. This is a common source of bugs when overlooked. But here it's only `False` and `True` to distinguish.

obspy/io/mseed/core.py
megies: Oh, I see you were doing it the same as in other parts of the code below, and I see now why it's done like this. We still support numpy 1.6.1 which has `ndarray.astype(t)` (https://docs.scipy.org/doc/numpy-1.6.0/reference/generated/numpy.ndarray.astype.html?highlight=astype#numpy.ndarray.astype) and no `copy=..` kwarg.

@krischer I think we should avoid copying data twice in the whole mseed writer code, so I think the way to go would be to define a little helper function up top of the file and replace all these double copying operations with it? You agree?

```
def _np_copy_astype(data, dtype):
    try:
        return data.astype(dtype, copy=True)
    except TypeError:
        return data.copy().astype(dtype)
```

obspy/io/mseed/core.py
megies: Ah, nevermind, I just saw it's done like this in the whole file here. Just leave as is then to match the rest of the code.. ;-)

megies:
@LMurrayBergquist I just saw in line 777 there is a check `if not trace_attr['encoding']:`. This looks dangerous to me @krischer as it could evaluate to `False` when a string encoding was set earlier `encoding = 0`. This should be `if trace_attr['encoding'] is None:`

Please change @LMurrayBergquist

obspy/io/mseed/core.py
```diff
@@ -785,6 +785,16 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
...
+                    trace_data.append(trace.data.copy().astype(np.int32))
+                    trace_attr['encoding'] = 11
+                else:
+                    msg = "int64 data only supported if it can be downcast to int32 type data."
```
megies: Please add `...only supported when writing MSEED if ...

megies:
Please add a changelog one-liner as well (also referencing issue number).

Also please add test cases, ideally writing to `BufferIO` objects (you can find examples in the test code e.g. in io/mseed/tests/test_mseed_util.py, it should be `with` blocks like `with io.BytesIO() as buf:`).
 - test that succesfully downcasting and writing to output works and re-read the data to see that it is the same as the input data
 - test that the expected error message is raised when the data can not be downcast (look in test code for examples that use `with self.assertRaises(...)`:

LMurrayBergquist:
> Please add a changelog one-liner as well (also referencing issue number).
> 
> Also please add test cases, ideally writing to `BufferIO` objects (you can find examples in the test code e.g. in io/mseed/tests/test_mseed_util.py, it should be `with` blocks like `with io.BytesIO() as buf:`).
> 
> * test that succesfully downcasting and writing to output works and re-read the data to see that it is the same as the input data
> * test that the expected error message is raised when the data can not be downcast (look in test code for examples that use `with self.assertRaises(...)`:

Ok, I will add the changes you suggest and make these test. But what is the 'changelog', and what should I add to that?

megies:
Check this: https://github.com/obspy/obspy/blob/master/CHANGELOG.txt sitting in the root dir of the git repo

obspy/io/mseed/core.py
```diff
@@ -495,6 +495,12 @@ def allocate_data(samplecount, sampletype):
     del lil  # NOQA
     return Stream(traces=traces)
 
+# Define helper function to copy data, replaces use of: trace.data.copy().astype(dtype)
```
megies: PEP8: 2 blank lines before and after function definitions (see CircleCI log)

Also, this comment should instead be a docstring..

```
def ...():
    """
    .....  <- here
    """
```
```

File: obspy/io/mseed/core.py [hunk 7]
```diff
@@ -495,6 +495,12 @@ def allocate_data(samplecount, sampletype):
...
+def _np_copy_astype(data, dtype):
+    try:
+        return data.astype(dtype, _np_copy_astype=True)
```
megies: this needs to be `..., copy=True)`

obspy/io/mseed/tests/test_mseed_reading_and_writing.py
```diff
@@ -155,6 +155,25 @@ def test_read_and_write_traces(self):
...
+        # create a dummy stream with int64 data 
+        x = np.arange(20, dtype = np.int64)
+        tr = Trace(x)
+        st = Stream()
+        st.append(tr)
+
+        # make sure the data can be written to mseed -> write to io.BytesIO() object in test 
+        file = io.BytesIO()
+        st.write(file, format="mseed")
+        st2 = read(file)
```
megies: Please use `with` construct, that's automatically closing the buffer at the end..

```
with io.BytesIO() as buf:
    st.write(buf, format="mseed")
    st2 = read(buf)
```

Also `file` is a  Python builtin, try to avoid (locally) overwriting builtins at all cost.

megies: Please als otest the failing case as mentioned with `self.assertRaises(...)` but I guess you're still working on this ... :-)

obspy/io/mseed/core.py
LMurrayBergquist: oops, that must have auto-filled in and I didn't notice, thanks!

megies: well, we'd have seen it in failing tests anyway.. that's what they're there for anyway ;-)

megies:
You've based this on an outdated master branch.. that's why all the Circle CI fails etc. Please rebase your branch on a current master and force-push. Get in touch with @DominikStr if you need help.

obspy/io/mseed/tests/test_mseed_reading_and_writing.py
```diff
@@ -167,12 +168,25 @@ def test_downcast_int64_to_int32(self):
...
+        with io.BytesIO() as buf:
+            st.write(buf, format="mseed")
+            st2 = read(buf)
+
+            res = np.array_equal(st,st2)
+            self.assertTrue(res)
+        
+        # Test that error message is indeed raised when data cannot be downcast
+        # Create dummy stream that cannot be properly downcast to int64
+        x_int64 = np.arange(2147483648, 9223372036854775807, 460000000000000000, dtype=np.int64)
```
megies: Ooof that's pretty long numbers to look at. I am assuming this arange will only contain a handful of numbers (it should only use a few so tests stay slim and fast)? But nobody will be able to tell from looking at this, so you should rather hard code it with like 3-4 numbers in it..

```
x_int64 = np.array([2147483648, ..., 9223372036854775807], dtype=np.int64)
```

Also, a good test should cover as many apects / potential inputs as possible. In this case, you should definitely add a test that has large negative numbers, as buggy code might not take that into account properly (I know yours does, but still -- tests should be agnostic of the code). Also, it seems the test data contains only numbers that can't be int32-represented. I think a better test data set would be to have like 4-5 small numbers and then just one large (incompatible) number among them.

Also, come to think of it, the check if data can besafely downcast to int32 could probably be replaced by checking the min/max values of the data (and checking against `np.iinfo(np.int32).min/max`). That might be faster, but we have it super safe I guess, so we'll leave as it for now I think.

LMurrayBergquist: Good point, I've updated the test for negative numbers and a mix of mainly small numbers with one too big one. Also a good point that checking that the min and max are in the right range should be just as good, I suppose it would take about the same space, but maybe it's faster to compute?

QuLogic: These seem like they'd be a bit clearer written as powers of 2/10: `2147483648, 9223372036854775807, 460000000000000000` == `2**31, 2**63 + 1, 460 * 10**15` (though maybe if the exact numbers don't matter too much, the last can be another power of 2 as well.)

megies: > Also a good point that checking that the min and max are in the right range should be just as good, I suppose it would take about the same space, but maybe it's faster to compute?

Just leave as is for now.. it's safe as it's implemented as is right now and that's most important for such an edge case.

megies: > the exact numbers don't matter too much, the last can be another power of 2 as well

Yep maybe clearer to just use powers of 2. And yes the exact numbers don't matter at all.

CHANGELOG.txt
```diff
@@ -28,6 +28,9 @@ master:
      have been deselected due to the default location priorities setting. This
      is a pure usability improvement as it has been confusing users
      (see #2159).
+ - obspy.io.mseed:
+   * Add add ability to write int64 data to mseed if it can safely be downcast
```
megies: Add add ;-)

megies:
Restarting Circle CI, it glitched out somehow..

CHANGELOG.txt
LMurrayBergquist: oops! Thanks!

obspy/io/mseed/core.py
```diff
@@ -785,6 +797,17 @@ def _write_mseed(stream, filename, encoding=None, reclen=None, byteorder=None,
                 trace_attr['encoding'] = 1
             elif trace.data.dtype.type == np.dtype(native_str('|S1')).type:
                 trace_attr['encoding'] = 0
+            # int64 data not supported; if possible downcast to int32, else create error message
+            # After bumping up to numpy 1.9.0 this check can be replaced by numpy.can_cast()
+            elif trace.data.dtype.type == np.int64:
+                # check if data can be safely downcast to int32
+                trace_data1 = _np_copy_astype(trace.data, np.int32)
+                if np.array_equal(trace.data, trace_data1):
```
krischer: How about just checking if the values in the array are in bounds with `np.iinfo(np.int32)` before casting instead of doing the cast + equality check? Feels a bit cleaner and should also be faster.

obspy/io/mseed/tests/test_mseed_reading_and_writing.py
```diff
@@ -155,6 +155,46 @@ def test_read_and_write_traces(self):
...
+        """
+        # create a dummy stream with int64 data 
+        x = np.array([1,2,-3,4], dtype = np.int64)
+        tr = Trace(x)
+        st = Stream()
+        st.append(tr)
+
+        # make sure the data can be written to mseed -> write to io.BytesIO() object in test 
+        with io.BytesIO() as buf:
+            st.write(buf, format="mseed")
```
krischer: Please also add a check that the original data in the traces is still of dtype `int64`.

obspy/io/mseed/core.py
LMurrayBergquist: True, that does look smoother, I'll change to that!

obspy/io/mseed/tests/test_mseed_reading_and_writing.py
LMurrayBergquist: Do you mean after st.write() add a check that st is still of dtype int64?

krischer: exactly

megies:
Still needs a rebase on current master branch (see above comment)

megies:
@LMurrayBergquist I've properly rebased, fixed code formatting and force-pushed to your branch. There were quite a few rebase conflicts due to the messy merging of master branch, looks OK to me, but please double check if the final code state now is what it should be like.

If you think it's OK, @krischer can merge after a final review I guess.

LMurrayBergquist:
> @LMurrayBergquist I've properly rebased, fixed code formatting and force-pushed to your branch. There were quite a few rebase conflicts due to the messy merging of master branch, looks OK to me, but please double check if the final code state now is what it should be like.
> 
> If you think it's OK, @krischer can merge after a final review I guess.

That looks good to me! And the test still passes, so I think it is good now, thanks for doing all that!

megies:
@LMurrayBergquist your last merge commit introduced a sloppy local branch merge again, after I cleanly rebased your PR. I'll force-push back to the previous state.

LMurrayBergquist:
> @LMurrayBergquist your last merge commit introduced a sloppy local branch merge again, after I cleanly rebased your PR. I'll force-push back to the previous state.

Oh dear, sorry about that! When I opened the Github desktop app it said there were some conflicts, so I went through and fixed those, then pushed it back. I guess now that you've cleaned it up I should not push anything, I should just leave it, right?

megies:
Yes, just leave as is. This should be good now and I'll leave the final review and merge to @krischer.

PR #2377, opened 2019-04-11
fix Trace.__eq__ (was super slow in case of >1e6 samples)

megies:
### What does this PR do?

Fix a speed issue when comparing traces/streams with lots of samples in them.

We had an ugly bug in `Trace.__eq__` which caused incredibly long CPU time for `tr1 == tr2` becaus numpy comparison wasn't run on `Trace.data` but rather `Trace` itself. I can't imagine how nobody stumbled over this one before. Maybe numpy made a change so that this only surfaced now??

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,7 @@
 master:
  - obspy.core:
+   * Fixed "==" comparison for Stream and Trace which was very slow in case of
+     traces with >1e6 samples (see #2377)
    * Casting FDSN identifiers to strings upon setting in the stats dictionary
      (see #1997).
    * UTCDateTime objects will now always evaluate equal if their string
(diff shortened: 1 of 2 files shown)


megies:
So embarassing.. after fix the last one takes 0.3 seconds instead of 3 minutes..

```
In [15]: for numsamps in np.logspace(3, 8, 6, dtype=np.int64):
    ...:     tr1 = Trace(data=np.empty(numsamps))
    ...:     tr2 = Trace(data=np.empty(numsamps))
    ...:     t = time.time()
    ...:     tr1 == tr2
    ...:     print('numsamps=%s  %.1f' % (numsamps, time.time() - t))
    ...:     
numsamps=1000  0.0
numsamps=10000  0.0
numsamps=100000  0.2
numsamps=1000000  1.8
numsamps=10000000  18.6
numsamps=100000000  188.0
```

PR #2378, opened 2019-04-11
Add read support for HypoDD pha files

trichter:
### PR Checklist
- [x] All tests still pass.


trichter:
+DOCS

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+def _seed_id_map(inventory=None, id_map=None, id_default='.{}..{}'):
+    if id_map is None:
+        id_map = {}
+    ret = id_map.copy()
+    if inventory is not None:
+        for net in inventory:
+            for sta in net:
+                if len(sta) == 0:
+                    temp = id_map.get(sta.code, id_default)
+                    temp = temp.split('.', 2)[-1]
```
krischer: I was not aware of this extra argument to the `.split()` method. Very cool! :-) Would be a bit more readable if you use `maxsplit=2`.

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+
+def _seed_id_map(inventory=None, id_map=None, id_default='.{}..{}'):
+    if id_map is None:
+        id_map = {}
+    ret = id_map.copy()
+    if inventory is not None:
+        for net in inventory:
+            for sta in net:
+                if len(sta) == 0:
+                    temp = id_map.get(sta.code, id_default)
```
krischer: I'm a bit confused here - if the station is already in the id_map the users have given some mapping explicitly (they might want to for example map the station to a different network code). I'd rather just `continue` here as user specified things should not be overwritten. But I might also just be missing something here.

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+    :param str encoding: encoding used (default: utf-8)
+
+    :rtype: :class:`~obspy.core.event.Catalog`
+    :return: An ObsPy Catalog object.
+    """
+    seed_map = _seed_id_map(inventory, id_map, id_default)
+    with io.open(filename, 'r', encoding=encoding) as f:
+        text = f.read()
+    events = [_block2event(block, seed_map, id_default, ph2comp)
+              for block in text.split('#')[1:]]
```
krischer: Such a fragile format - but if it always originates from the same code I guess this is reliable? Another option would be group the blocks with some regular expression but that might not be worth the effort here.

obspy/io/hypodd/tests/test_pha.py
```diff
@@ -0,0 +1,51 @@
...
+
+from obspy import read_events, read_inventory
+from obspy.io.hypodd import pha
+
+
+class PHATestCase(unittest.TestCase):
+    """
+    Test suite for obspy.io.hypodd.pha
+    """
+
```
krischer: Can you add a test to fully check all the values in the resulting catalog object?

obspy/io/hypodd/tests/test_pha.py
```diff
@@ -0,0 +1,51 @@
...
+    def setUp(self):
+        path = os.path.dirname(__file__)
+        self.fname = os.path.join(path, 'data', 'example.pha')
+
+    def test_is_pha(self):
+        self.assertEqual(pha._is_pha(self.fname), True)
```
krischer: No test for the negative yet. Maybe just run it against all the Python files in the test folder or something like this?

misc/docs/source/packages/index.rst
```diff
@@ -95,6 +95,7 @@ categories.*
     obspy.io.css
     obspy.io.gcf
     obspy.io.gse2
+    obspy.io.hypodd
```
krischer: That's the waveform plugin section. Please add it to the event plugin section.

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+    return ret
+
+
+def _block2event(block, seed_map, id_default, ph2comp):
+    """
+    Read HypoDD event block
+    """
+    lines = block.strip().splitlines()
+    yr, mo, dy, hr, mn, sc, la, lo, dp, mg, eh, ez, rms, id_ = lines[0].split()
+    time = UTCDateTime('{}-{}-{} {}-{}-{}'.format(yr, mo, dy, hr, mn, sc))
```
krischer:
```suggestion
    time = UTCDateTime(int(yr), int(mo), int(dy), int(hr), int(mn), float(sc))
```

Okay - seeing it written out I'm no longer sure this is cleaner.

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+        arrivals.append(arrival)
+    qu = None if rms == '0.0' else OriginQuality(standard_error=float(rms))
+    origin = Origin(arrivals=arrivals,
+                    quality=qu,
+                    latitude=float(la),
+                    longitude=float(lo),
+                    depth=1000 * float(dp),
+                    time=time)
+    magnitude = Magnitude(mag=mg)
+    event = Event(resource_id=id_,
```
krischer:
```suggestion
    event = Event(resource_id="smi:local/event/" + id_,
```

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+        wid = seed_map.get(sta, id_default)
+        _waveform_id = WaveformStreamID(seed_string=wid.format(sta, comp))
+        pick = Pick(waveform_id=_waveform_id, phase_hint=phase,
+                    time=time + float(reltime))
+        arrival = Arrival(phase=phase, pick_id=pick.resource_id,
+                          time_weight=float(weight))
+        picks.append(pick)
+        arrivals.append(arrival)
+    qu = None if rms == '0.0' else OriginQuality(standard_error=float(rms))
+    origin = Origin(arrivals=arrivals,
```
krischer:
```suggestion
    origin = Origin(arrivals=arrivals, resource_id="smi:local/origin/" + id_
```

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+                    quality=qu,
+                    latitude=float(la),
+                    longitude=float(lo),
+                    depth=1000 * float(dp),
+                    time=time)
+    magnitude = Magnitude(mag=mg)
```
krischer:
```suggestion
    magnitude = Magnitude(mag=mg, resource_id="smi:local/magnitude/" + id_)
```

obspy/io/hypodd/pha.py
trichter: I think you are right. I will change this. I wanted to also support inventories with only network and station information, but that does not make much sense.

I think I will also add a warning when the construction of the id_map would not be unique (e.g. BH and HH channels in inventory) and move these functions to some util module, because they are also used by evt plugin. Is it OK if I add a new file obspy/io/util.py?

obspy/io/hypodd/pha.py
trichter: I think its a bit cleaner...

obspy/io/hypodd/pha.py
trichter: I would leave it simple.

obspy/io/hypodd/pha.py
```diff
@@ -0,0 +1,128 @@
...
+    return event
+
+
+def _is_pha(filename):
+    try:
+        with open(filename, 'rb') as f:
+            temp = f.readline()
+    except Exception:
+        return False
+    return temp.startswith(b'#') and len(temp.split()) == 15
```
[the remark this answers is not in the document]
megies: We should also consider putting weak detection routines at the very end of the autodetection chain.

obspy/io/hypodd/pha.py
megies: I use it all the time.. :-)

megies: Btw.. `split()` doesn't even work with named kwargs, I guess it's some Python C API thingy? I could never be bothered to look up the reason..

megies: A side effect is that you can't use `maxsplit` without specifying the separator, even if you want the default one, so you have to say `split(None, 2)` in that case. Weird non-Pythonic glitch somehow.. ;-)

megies:
CircleCI: `obspy/io/hypodd/pha.py:61:27: E999 SyntaxError: invalid syntax`

trichter:
> CircleCI: obspy/io/hypodd/pha.py:61:27: E999 SyntaxError: invalid syntax

Yeah, I already saw this. Committing inside the browser is a bit error prone.

obspy/io/hypodd/pha.py
QuLogic: The non-named kwarg thing is Python 2 only, I think.

trichter:
@krischer thanks for the review. I think it is ready.

obspy/core/util/misc.py
```diff
@@ -748,6 +749,25 @@ def func(obj, parent=None, attr=None):
     return func(obj)
 
 
+def _seed_id_map(
+        inventory=None, user_id_map=None, key='{sta.code}',
+        seed_factory='{net.code}.{{}}.{cha.location_code}.{cha.code:.2}{{}}'):
+    """Return mapping between station code and seed id expressions"""
```
megies: Maybe switch to our normal style..?

```
def xxx():
    """
    ....
    """
```

obspy/io/hypodd/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+The PHA reader hooks into the standard ObsPy event handling
+mechanisms including format autodetection.
+
+>>> from obspy import read_events
+>>> cat = read_events('/path/to/example.pha')
+>>> print(cat)
+2 Event(s) in Catalog:
+2025-05-14T14:35:35.510000Z | +40.225,  +10.450 | 3.5 None
+2025-05-14T15:43:05.280000Z | +40.223,  +10.450 | 1.8 None
```
megies: And here they say you can't predict earthquakes! :sunglasses:

setup.py
```diff
@@ -305,6 +305,7 @@
         'IMS10BULLETIN = obspy.io.iaspei.core',
         'EVT = obspy.io.sh.evt',
         'FOCMEC = obspy.io.focmec.core',
+        'PHA = obspy.io.hypodd.pha'
```
megies: Hmm, maybe `HYPODDPHA` instead? Might be more meaningful but also longer, so dunno..

obspy/io/hypodd/pha.py
megies: Ah OK. I'll probably never tried it on Py3 ;-)

setup.py
trichter: I have no preferences here... just a bit of work

trichter: ... no work at all. I did not change the file names just the format name.

obspy/core/util/__init__.py
```diff
@@ -34,7 +34,8 @@
                                   PROJ4_VERSION)
 from obspy.core.util.misc import (BAND_CODE, CatchOutput, complexify_string,
                                   guess_delta, loadtxt, score_at_percentile,
-                                  to_int_or_zero, SuppressOutput)
+                                  to_int_or_zero, SuppressOutput,
+                                  _seed_id_map)
```
krischer: Any reason to put this here? It could just be directly imported from misc.

trichter: Hmh, it has a leading underscore, so its private, but I thought it might even be of use outside obspy. And even inside obspy its easier to import from core.util instead of core.util.misc. What do you think?

krischer: Our current policy is that everything with a leading underscore is internal only and we are free to change the API of those functions and even remove them so I'd advise against using it outside of ObsPy.

And these "import aliases" just make it harder to find where things originate from and the convenience improvement from not having to write `.misc` is IMHO not worth the trade-off. But that is just my opinion and feel free to choose.

trichter:
OK, I addressed it. However, I am not authorized to merge.

krischer:
> OK, I addressed it. However, I am not authorized to merge.

Yea this is not good. I opened #2384 to discuss it.

setup.py
megies: > I did not change the file names just the format name.

Yep, that's what I meant. :+1:

obspy/core/util/__init__.py
megies: > Our current policy is that everything with a leading underscore is internal only and we are free to change the API of those functions and even remove them so I'd advise against using it outside of ObsPy.

On the other hand, if you do want to use it, you can rename it and remove the leading underscore. But then we need proper tests for it as well (I didnt check if there already are tests for this, maybe there are already).

megies:
:tada: thanks for the PR @trichter

PR #2385, opened 2019-04-16
Improve seg2 loading

### LINKED ISSUE #2022: SEG2 read support for "Data format code 3" (20-bit SEG-D floating point)?
Dear Obspy users/developers,
I am opening this issue because I face an issue when I try to read some Seg2 datafiles from active seismic experiments. Here enclosed are two seg2 files, one is ok (GOOD.DAT) and the other fails (BAD.DAT).
[SEG2_ISSUE.zip](https://github.com/obspy/obspy/files/1536401/SEG2_ISSUE.zip)

Both are generated by a Smartseis (geometrics) acquisition station, but of different age.

I can read the BAD.DAT header without problem, but I face the issue when parsing the trace blockette: it looks like the bit length / data type is badly decoded.

Any help would be very much appreciated (about half of the seismic files I have to work with are of BAD.DAT type..).

Thanks again for this very helpful python tool,
Cheers

Pierre Bottelin
--
ADRGT
Grenoble, France

QuLogic:
### What does this PR do?

This enhances and fixes a couple of things:
* Get format code 3 working (#2022)
* Fix `NOTE` parsing; it's not documented as a key-value pair, and the invalid cleanup code inadvertently removed the newlines.
* Use `unpack_from` a bit to simplify parsing.
* Fix data loading; it _always_ uses native byte order, but should use the order we determined from the file header.

Fixes #2022.

### PR Checklist
- [~] This PR is not directly related to an existing issue (which has no PR yet).
- [N/A] If the PR is making changes to documentation, docs pages can be built automatically.
- [N/A] If any network modules should be tested for the PR, add them as a comma separated list
- [x] All tests still pass.
- [N/A] First time contributors have added your name to `CONTRIBUTORS.txt` .


QuLogic:
The new data file also triggers the warning about `DELAY`, but I don't know how we want to handle that.

obspy/io/seg2/seg2.py
```diff
@@ -262,52 +285,54 @@ def parse_free_form(self, free_form_str, attrib_dict):
...
+        # offset of 0 indicates the end of the strings.
+        offset = 0
+        strings = []
+        while offset + 2 < len(free_form_str):
+            strlen, = unpack_from(self.endian + b'H', free_form_str, offset)
+            if strlen == 0:
+                break
+            strings.append(
+                free_form_str[offset + 2:offset + strlen].rstrip(
+                    self.string_terminator))
```
QuLogic: Not sure if we want to split at first `string_terminator` and drop the rest instead of `rstrip`. This is something an in-place edit might do to avoid re-writing the file.

krischer: I'm not entirely sure what you mean by the in-place edit. Other than that I imagine that `rstrip()` is faster than a split because it only has to create one new string object instead of at least two for the `split()` variant.

obspy/io/seg2/tests/test_seg2.py
```diff
@@ -95,7 +114,31 @@ def test_read_data_format_2(self):
...
+            warnings.simplefilter('always')
+            st = read(basename + ".seg2")
+        self.assertEqual(len(w), 2)
+        self.assertIn('custom', str(w[1]))
+        # read reference ASCII data
+        with gzip.open(basename + '.DAT.gz', 'rb') as f:
+            results = np.loadtxt(f, ndmin=2).T
+        for tr, result in zip(st, results):
+            # convert raw data to unit'd
+            scaled_data = tr.data * tr.stats.calib
```
krischer: How do we in other formats actually deal with the scaling factor? Do we want it applied directly or not?

obspy/io/seg2/seg2.py
QuLogic: SEG2 seems to be a format with intentional 'holes' to allow for in-place editing (e.g., the trace pointer block can be larger than the number of traces, so you can just append a trace and overwrite one pointer).

So what I'm saying here is, if you have something like `Some very long text string\0`, you could replace it by just overwriting the beginning with `A shorter string\0` and end up with `A shorter string\0t string\0`. In that case, only stripping the right give you the wrong thing. Writing this out has mostly convinced me we should do the split instead, but I seem to recall another format where we might have done differently.

obspy/io/seg2/tests/test_seg2.py
QuLogic: I'm not sure; when I plot it, the calibration is definitely applied somewhere. I just copied this line from the other test (but kept out the mm/um unit conversion, since I wasn't sure what units these were). It could be the other test is wrong, though.

Maybe @stefanazz could clarify the exact units and expected magnitude of the `1068.DAT` file he provided.

obspy/io/seg2/seg2.py
krischer: I guess it is a question if the 2-byte offset in front of each string of the terminator takes precedence. I'd personally vote for the offset but I have no further information.

obspy/io/seg2/tests/test_seg2.py
krischer: Yea you are right - I just looked through a couple of other modules and we do not directly apply the calibration factor upon reading so what you are doing here is the correct thing.

Don't know about the unit conversion though.

obspy/io/seg2/seg2.py
QuLogic: I've actually decided to switch it (though not using `split`). It's not exactly about precedence. The offset only tells you where the next string _starts_; it doesn't necessarily tell you where the current string _ends_.

QuLogic:
Add a changelog entry.

megies:
CI fails are unrelated, this has a review and tests, so merging. Thanks @QuLogic and @krischer for reviewing.

PR #2405, opened 2019-05-24
seedlink client: add option to fetch available stations

megies:
### What does this PR do?

Adds a method to seedlink client to be able to fetch a list of network/stations for which waveform streams are provided.

This can help in applications to resolve wildcards in stations/networks, e.g. when using `seedlink-plotter` with `"GR_*:HHZ"` as seedlink string because the realtime version of the seedlink client can not handle wildcards in network/station.

###  PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.seedlink"
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -68,8 +68,12 @@ master:
    * Properly handle fetching poles and zeros in presence of multiple metadata
      files for a given station (see #2411)
  - obspy.clients.seedlink:
-   * add method "get_stations()" to fetch list of network/stations for which
-     streams are available (see #2405)
+   * add method "get_info()" to fetch information on what
+     networks/stations/locations/channels are served by the seedlink server
+     (see #2405)
+   * "get_waveforms()" can now be used with '*' and '?' wildcards in any part
+     of requested SEED ID, i.e. network, station, location and channel (see
+     #2405)
  - obspy.imaging:
    * obspy-scan can now be used with wildcarded SEED IDs when specifying what
      to plot after scanning data (see #2227)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -67,6 +67,9 @@ master:
  - obspy.clients.seishub:
    * Properly handle fetching poles and zeros in presence of multiple metadata
      files for a given station (see #2411)
+ - obspy.clients.seedlink:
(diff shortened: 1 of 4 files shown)


megies:
Maybe should rename the method, because other `get_stations()` methods usually return `Inventory` objects..

d-chambers:
> Maybe should rename the method, because other get_stations() methods usually return Inventory objects

Yes, I think this is important for maintaining a consistent API. 

I am not very familiar with the seedlink client but would it make sense to also return channel level info? or maybe have a `level` keyword argument?

Also, is the cache really merited? I don't imagine users calling this method more than a small number of times (probably once) in the same python process, but maybe there is a use case I am overlooking.

megies:
> Yes, I think this is important for maintaining a consistent API.

Yeah, I totally agree and I only thought about it when I was sending the PR. How about renaming it to `get_contents()` or `get_station_information`?

But actually, I checked again and we can also query down to full SEED ID level, so maybe I change it to return an `Inventory` object and have the `level` keyword.

> Also, is the cache really merited? I don't imagine users calling this method more than a small number of times (probably once) in the same python process, but maybe there is a use case I am overlooking.

I agree that most use cases will only do a single or some few request and that users could do the same caching easily in their programs. But it's a low hanging fruit, not very complex implementation, users tend to be lazy and just call the method again and the station request actually takes pretty long (several seconds), so I think it should stay in for use cases we might not have on the radar.

Jollyfant:
How does this work? Does it use the `CAT` command on seedlink?

megies:
https://github.com/obspy/obspy/pull/2405/files#diff-0eb620aa1b2af06568b44a7b57bc1c04R169

megies:
```
INFO level
    requests an INFO packet containing XML data embedded in a Mini-SEED log record. level should be one of the following: ID, CAPABILITIES, STATIONS, STREAMS, GAPS, CONNECTIONS, ALL. The XML document conforms to the Document Type Definition (DTD) shown in section ???. The amount of info available depends on the conﬁguration of the SeedLink server. 
```
https://www.seiscomp3.org/doc/applications/seedlink.html

megies:
Tests passing should be good to go.

Issue #2408, opened 2019-05-29
response: add helper function to get a flat response


megies:
I think it might be a good idea (and not hard / a lot of work to implement) to have a helper function (or maybe a static method on `Response`) that returns a flat response. Sometimes this is needed with workflows that expect station metadata with a response and when handling data that already was converted to the target units outside of the workflow (e.g. industry data sometimes, converted to m/s during data acquisition).

CC @d-chambers

Opinions?

Here's an example I created at some point, there might be other (better?) options to achieve the same thing.

```xml
<Response>
  <InstrumentSensitivity>
    <Value>1</Value>
    <Frequency>10</Frequency>
    <InputUnits>
      <Name>M/S</Name>
      <Description>Velocity in Meters per Second</Description>
    </InputUnits>
    <OutputUnits>
      <Name>M/S</Name>
      <Description>Velocity in Meters per Second</Description>
    </OutputUnits>
  </InstrumentSensitivity>
  <Stage number="1">
    <PolesZeros>
      <InputUnits>
        <Name>M/S</Name>
        <Description>Velocity in Meters per Second</Description>
      </InputUnits>
      <OutputUnits>
        <Name>M/S</Name>
        <Description>Velocity in Meters per Second</Description>
      </OutputUnits>
      <PzTransferFunctionType>LAPLACE (RADIANS/SECOND)</PzTransferFunctionType>
      <NormalizationFactor>1.0</NormalizationFactor>
      <NormalizationFrequency unit="HERTZ">10.0</NormalizationFrequency>
    </PolesZeros>
    <StageGain>
      <Value>1</Value>
      <Frequency>10</Frequency>
    </StageGain>
  </Stage>
</Response>
```

![Figure_1-2](https://user-images.githubusercontent.com/1842780/58541133-1922db80-81fb-11e9-9746-d403c6aa7152.png)

d-chambers:
@megies,

I think this a good feature to have. Might be better as a classmethod like this:

```python
class Response(object):
    @classmethod
    def get_flat_response(cls, normalization_frequency=1, overall_sensitivity=1, units='M/S'):
        """
        :param units: one of 'VEL', 'ACC', 'DISP', or any arbitrary units string
        """
        response = cls(...)
        return response
```
if we ever anticipate it needing to work with sub-classes of `Response`. 

[...]

PR #2415, opened 2019-06-17
Added a flat response class method.

aringler-usgs:
This pull request is a potential solution to: https://github.com/obspy/obspy/issues/2408

It will likely have a number of changes because of different uses cases.

### What does this PR do?

*Please fill in*

*Please fill in*

### PR Checklist
- [X ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ X] This PR is not directly related to an existing issue (which has no PR yet).
- [ X] All tests still pass.
- [ X] Any new features or fixed regressions are be covered via new tests.
- [ X] Any new or changed features have are fully documented.
- [ X] Significant changes have been added to `CHANGELOG.txt` .

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -961,6 +961,49 @@ class Response(ComparingObject):
 
         return sampling_rates
 
+    @classmethod
+    def get_flat_response(cls, sensitivity=1., frequency=1.,
+                          input_units="M/S", output_units="M/S",
+                          include_pole_zero_stage=False):
+
+        """
+        Makes a flat response object.
+
+        :type sensitivity: float, optional
+        :param sensitivity : The sensitivity of the response, defaults to 1.
+        :type frequency: float, optional
+        :param frequency: The frequency where the sensitivity is stated,
+            defaults to 1 Hz.
+        :type input_units: string, optional
+        :param input_units: The input units going into the response object,
+            default units are "M/S"
+        :type output_units: string, optional
+        :param input_units: The output units going out of the response object,
+            default units are "M/S"
+        :type include_pole_zero_stage: bool, optional
+        :param include_poles_zero_stage: This option allows for a pole/zero
+            stage to be include. If a pole zero stage is included it adds
+            one pole and one zero both with the value of 0 + 1j.
+        """
+        sensitivity = InstrumentSensitivity(sensitivity, frequency,
+                                            input_units, output_units)
+        stage = []
+        if include_pole_zero_stage:
+            pzt = "laplace hz"
+            stg1 = PolesZerosResponseStage(stage_sequence_number=1,
+                                           stage_gain=sensitivity,
+                                           stage_gain_frequency=frequency,
+                                           input_units=input_units,
+                                           output_units=output_units,
+                                           pz_transfer_function_type=pzt,
+                                           normalization_frequency=frequency,
+                                           zeros=[0. + 1j], poles=[0. + 1j])
+            print(stg1)            
+            stage.append(stg1)
+        response = cls(instrument_sensitivity=sensitivity,
+                       response_stages=stage)
+        return response
+
     def recalculate_overall_sensitivity(self, frequency=None):
         """
         Recalculates the overall sensitivity.
--- a/obspy/core/tests/test_response.py
+++ b/obspy/core/tests/test_response.py
@@ -25,7 +25,8 @@ from matplotlib import rcParams
 
 from obspy import UTCDateTime, read_inventory
 from obspy.core.inventory.response import (
-    _pitick2latex, PolesZerosResponseStage, PolynomialResponseStage)
+    _pitick2latex, PolesZerosResponseStage, PolynomialResponseStage,
+    Response)
 from obspy.core.util import MATPLOTLIB_VERSION
 from obspy.core.util.misc import CatchOutput
 from obspy.core.util.obspy_types import ComplexWithUncertainties
@@ -482,6 +483,32 @@ class ResponseTestCase(unittest.TestCase):
             [2.411908e+05 + 2.283852e+04j,
              2.445572e+05 - 2.480459e+03j,
              -2.455459e-01 + 4.888214e-02j], rtol=1e-6)
+
+    def test_get_flat_response_without_poles_and_zeros(self):
+        """
+        Tests the get_flat_response calss method without poles and zeros.
+        """
+
+        resp = Response.get_flat_response()
+        self.assertEqual(resp.instrument_sensitivity.value, 1.)
+        self.assertEqual(resp.instrument_sensitivity.frequency, 1.)
+        self.assertEqual(resp.instrument_sensitivity.input_units, "M/S")
+        self.assertEqual(resp.instrument_sensitivity.output_units, "M/S")
+
+    def test_get_flat_response_with_poles_and_zeros(self):
+        """
+        Tests the get_flat_response calss method with poles and zeros.
+        """
+
+        resp = Response.get_flat_response(include_pole_zero_stage=True)
+        self.assertEqual(resp.instrument_sensitivity.value, 1.)
+        self.assertEqual(resp.instrument_sensitivity.frequency, 1.)
+        self.assertEqual(resp.instrument_sensitivity.input_units, "M/S")
+        self.assertEqual(resp.instrument_sensitivity.output_units, "M/S")
+        self.assertEqual(resp.response_stages[0].poles, [0+1j])
+        self.assertEqual(resp.response_stages[0].zeros, [0+1j])
+        self.assertEqual(resp.response_stages[0].normalization_factor, 1.)
+        self.assertEqual(resp.response_stages[0].normalization_frequency, 1.)
 
     def test_recalculate_overall_sensitivity(self):
         """
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,6 @@
 master:
  - obspy.core:
+   * Added a flat response class method (see #2408)
    * Fixed "==" comparison for Stream and Trace which was very slow in case of
      traces with >1e6 samples (see #2377)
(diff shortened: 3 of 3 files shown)


d-chambers:
Hey @aringler-usgs, Looks like a useful method.

I have a few suggestions:

1). Could you add a few remove response tests to ensure using a flat response doesn't change the stream in any meaningful way? Also, test the responses can be used to change unit types, eg going from m/s to m if this is a valid use case.

2). Under the "include_poles_zero_stage" could you add a bit of text describing why one might want to do this?

aringler-usgs:
@d-chambers thanks for the comments.  For your first suggestion, are you suggesting I add the possibility of different response units?  For example, if the user puts an input as m/s and an output of m, then the flat_response should add an additional zero in the response stage to deal with this case?

krischer:
> @d-chambers thanks for the comments. For your first suggestion, are you suggesting I add the possibility of different response units? For example, if the user puts an input as m/s and an output of m, then the flat_response should add an additional zero in the response stage to deal with this case?

We have to be a bit careful here I guess. `.remove_response()` will always assume that the current data in the stream is in the output units of the last stage. Furthermore it assumes that deconvolving the data with the response without doing anything else will convert it to the input units of the first stage. If the `output` argument of the the `.remove_response()` method differs from the input units of the first stage (I realize the notation becomes a bit confusing here), it will actually differentiate or integrate the response before the convolution.

d-chambers:
@krischer, so I guess this test wouldn't be need then? We should make sure that, in creating this response, it will either honor the input/output units specified or raise an exception and issue a useful error message indicating which units should be used if an invalid combination is selected.

megies:
That's actually a case I hadn't thought of when proposing this feature. If a flat response with e.g. input of velocity units `m/s` is created but later in `remove_response(output='DISP')` is used. This is dangerous terrain, not sure how to handle this potential pitfall. At the time `remove_response()` is called we can't tell anymore that the response was created as a flat dummy response and so are not able to warn/raise for the user.. hmm

aringler-usgs:
The case I use this for is converting an acceleration synthetic to displacement.  So I make a flat response m/s/s to m/s/s and then add two zeros at zero.  This now makes the response with a final stage that has m/s/s and a first stage of m (e.g. in stage 0 you are converting to displacement).

So for this type of case, I would think, you want to make sure your flat response has the units and corrections you want.  Then if you later use remove_response and change the units it should apply additional corrections.  For example, it would remove one of the zeros at zero to go from m/s/s to m/s.

Let me know if I am missing a different pitfall.

obspy/core/inventory/response.py
```diff
@@ -961,6 +961,51 @@ def is_close(a, b):
 
         return sampling_rates
 
+    @classmethod
+    def get_flat_response(cls, sensitivity=1., frequency=1.,
+                          input_units="M/S", output_units="M/S",
+                          include_pole_zero_stage=False):
```
megies: What's the reasoning behind not always having at least one stage? We can't even `Response.plot()` it without any stages..

```
/home/megies/git/obspy/obspy/core/inventory/response.py in _call_eval_resp_for_frequencies(self, frequencies, output, start_stage, end_stage, hide_sensitivity_mismatch_warning)
   1151             msg = ("Can not use evalresp on response with no response "
   1152                    "stages.")
-> 1153             raise ObsPyException(msg)
   1154 
   1155         import obspy.signal.evrespwrapper as ew

ObsPyException: Can not use evalresp on response with no response stages.
```

obspy/core/inventory/response.py
```diff
@@ -961,6 +961,51 @@ def is_close(a, b):
...
+            one pole and one zero both with the value of 0 + 1j.  This allows
+            for the user to apply a deconvolution routine.  E.g. it allows
+            the user to include the tapering, detrendings, and filtering
+            during the response removal.
+        """
+        sensitivity = InstrumentSensitivity(sensitivity, frequency,
+                                            input_units, output_units)
+        stage = []
+        if include_pole_zero_stage:
+            pzt = "laplace hz"
```
megies: I'm not sure if this value works..? In any case, might not hurt to use the canonical values directly..

```
:type pz_transfer_function_type: str
:param pz_transfer_function_type: A string describing the type of transfer
    function. Can be one of:

    * ``LAPLACE (RADIANS/SECOND)``
    * ``LAPLACE (HERTZ)``
    * ``DIGITAL (Z-TRANSFORM)``

    The function tries to match inputs to one of three types if it can.
```

obspy/core/inventory/response.py
```diff
@@ -961,6 +961,51 @@ def is_close(a, b):
...
+            stg1 = PolesZerosResponseStage(stage_sequence_number=1,
+                                           stage_gain=sensitivity,
```
megies: stage gain is just a plain float.

```
import copy
from obspy.core.inventory import Response
st = read()
inv = read_inventory()
inv = inv.select(station='RJOB', time=st[0].stats.starttime)
flat = Response.get_flat_response(include_pole_zero_stage=True)
for cha in inv[0][0]:
    cha.response = copy.deepcopy(flat)
inv.plot_response(0.001)

  File "flat.py", line 10, in <module>
    inv.plot_response(0.001)
  File "/home/megies/git/obspy/obspy/core/inventory/inventory.py", line 1100, in plot_response
    outfile=None)
  File "/home/megies/git/obspy/obspy/core/inventory/channel.py", line 366, in plot
    outfile=outfile)
  File "/home/megies/git/obspy/obspy/core/inventory/response.py", line 1841, in plot
    end_stage=end_stage)
  File "/home/megies/git/obspy/obspy/core/inventory/response.py", line 1703, in get_evalresp_response
    freqs, output=output, start_stage=start_stage, end_stage=end_stage)
  File "/home/megies/git/obspy/obspy/core/inventory/response.py", line 1665, in get_evalresp_response_for_frequencies
    end_stage=end_stage)
  File "/home/megies/git/obspy/obspy/core/inventory/response.py", line 1554, in _call_eval_resp_for_frequencies
    gain_blkt.gain = blockette.stage_gain
TypeError: a float is required
```

megies:
I think we might want to check that `get_evalresp_response()` works in the tests? Most use cases I can think of will want the frequency response output?

obspy/core/inventory/response.py
aringler-usgs: Thanks, I fixed this.  I was writing over my sensitivity unknowingly.

obspy/core/inventory/response.py
aringler-usgs: This should now be fixed.  For Response.plot() it still requires the sampling_rate as we don't have any digital stages in the response that would indicate the sampling_rate.  However, the following works.
Response.plot(0.1, sampling_rate=1)

obspy/core/inventory/response.py
aringler-usgs: I *think* I changed this to what you are suggesting.  Please tell me if I am missing things.

aringler-usgs:
@megies I switched to get_evalresp_response_for_frequencies as test cases.  This seems to produce a better test case situation.  Testing with seismometer_simulate can be a bit awkward because you have a number of other things going on (filters, detrending,...).  Please let me know if you would like additional changes.  Thanks for everyone's help.

obspy/core/inventory/response.py
```diff
@@ -961,6 +961,57 @@ def is_close(a, b):
...
+        :param frequency: The frequency where the sensitivity is stated,
+            defaults to 1 Hz.
+        :type input_units: string, optional
+        :param input_units: The input units going into the response object,
+            default units are "M/S"
+        :type output_units: string, optional
+        :param input_units: The output units going out of the response object,
+            default units are "M/S"
+        :type include_pole_zero_stage: bool, optional
+        :param include_pole_zero_at_zero: This option allows for a pole/zero
```
krischer: I'm a bit confused here. Why exactly is this needed? The output from evalresp should be identical no matter if they are added or not, no?

obspy/core/inventory/response.py
```diff
@@ -961,6 +961,57 @@ def is_close(a, b):
 
         return sampling_rates
 
+    @classmethod
+    def get_flat_response(cls, sensitivity=1., frequency=1.,
```
krischer: Do we want a default value for the sensitivity? I can see arguments for and against it but I just wanted to quickly discuss it.

obspy/core/tests/test_response.py
```diff
@@ -483,6 +484,52 @@ def test_regression_evalresp(self):
              2.445572e+05 - 2.480459e+03j,
              -2.455459e-01 + 4.888214e-02j], rtol=1e-6)
 
+    def test_get_flat_response_without_poles_and_zeros(self):
+        """
+        Tests the get_flat_response calss method without poles and zeros.
+        """
+
+        resp = Response.get_flat_response()
```
krischer: Can you test for some other sensitivity than 1? Testing with 1 sometimes hides some issues as multiplication/division are the same with it.

obspy/core/inventory/response.py
aringler-usgs: I included this mostly so that when I use `seismometer_simulate()` I get the exact same behavior on my synthetic as my data.  For example, the detrending, tapering etc.  This was necessary since it checks if you have poles and zeros.  It appears to still make that check on the dictionary.  I am referring to line 531: `obspy/signal/invsim.py`.

If you think this is no longer necessary, then I can definitely remove it.

obspy/core/inventory/response.py
aringler-usgs: I am happy with either way.  Let me know and I can change it.

obspy/core/tests/test_response.py
aringler-usgs: Good call.  I switched the test cases to the value 5.

aringler-usgs:
Did you need anything else on my end for this pr?

MiguelonGonzalez:
is this still on progress? I think it was a very useful method.

PR #2326, opened 2019-02-15
Generalize Inventory, Catalog and Stream reading - allow wildcards

ThomasLecocq:
### What does this PR do?

This PR resolves the difference between ``read_events`` and ``read_inventory``. Now both call a generic reader function that accepts ``str``, ``file-like-objects``, ``url`` and wildcards for ``str``. The PR also adds the ``__eq__`` and ``extend`` methods to ``Inventory``.

Resolves #2238 , related to #2322 .

### PR Checklist
- [x] Correct base branch selected? `master` for new features,
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] If the PR is making changes to documentation, docs pages can be built automatically.
      Just remove the space in the following string after the + sign: + DOCS
- [ ] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [x] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #2419, opened 2019-06-24
Fix reading from buffer not at start of file and using autodetection

megies:
I am pretty sure we have a bug in our reading routines when
 - auto fileformat detection is used
 - a binary buffer is passed
 - that buffer is not at start of file

To reproduce, add some random bytes into a buffer and fill the rest with a valid waveform files data:

```python
from obspy import read
filename = 'LMOW.BHE.SAC'  # file is in io/sac/tests/data
import io
bio = io.BytesIO()
bio.write(b'123456')
data = open(filename, 'rb').read()
bio.write(data)
bio.seek(6)
st = read(bio)
```

The last line will call `_is_mseed()` first, and then goes over to `_is_sac()` and should eventually read it as a sac file starting at the initial offset position of 6 bytes.

This might be a real edge case, especially since some `is_xxx()` check routines might not be clean about keeping file handles open (did not check all of them..) but still should be fixed eventually.

The fix should be simple, but I found more code that breaks when not at position 0 in buffer (in sac file check) and there might be many more locations with problems of that kind, but still better to fix at least some of them. We probably need a `_generic_is_file_format` routine that consistently handles open file handles etc like was done in #2326 not sure..

```diff
diff --git a/obspy/core/util/base.py b/obspy/core/util/base.py
index 9c81efe17..0a9ffbcea 100644
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -433,7 +434,7 @@ def _read_from_plugin(plugin_type, filename, format=None, **kwargs):
             # check format
             is_format = is_format(filename)
             if position is not None:
-                filename.seek(0, 0)
+                filename.seek(position, 0)
             if is_format:
                 break
         else:
```

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -129,6 +129,19 @@ class WaveformPluginsTestCase(unittest.TestCase):
                             st = read(temp, format=format)
                             self.assertEqual(len(st), 1)
                             self.assertEqual(st[0].stats._format, format)
+                            # BytesIO with an offset (additional data in front
+                            # but file pointer at right position in file), with
+                            # and without autodetection
+                            for autodetect in (format, None):
+                                temp.seek(0)
+                                temp2 = io.BytesIO()
+                                dummy_bytes = b'123456'
+                                temp2.write(dummy_bytes)
+                                temp2.write(temp.read())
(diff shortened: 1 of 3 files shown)


megies:
Rebased + force-pushed

PR #2424, opened 2019-07-16
fix ci builds

d-chambers:
### What does this PR do?

This PR attempts to get the travis CI working again.

### Why was it initiated?  Any relevant Issues?

Travis is failing

### PR Checklist
- [ ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just remove the space in the following string after the + sign: "+ DOCS"
- [ ] If any network modules should be tested for the PR, add them as a comma separated list
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:"
      (you can also add "ALL" to just simply run all tests across all modules)
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are be covered via new tests.
- [ ] Any new or changed features have are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #2426, opened 2019-07-18
Get rid of deprecation warning `from collections import `

heavelock:
I changed all imports `from collections import smth` to be PY2 constant dependent.
That way of importing will de removed in Py3.8 in favour of `from collections.abc import smth`.
Since in the codebase there are already existing imports based on `future.utils.PY2` constant I decided to use that approach. 

I change all occurences of that kind of an import apart of one code snippet in the documentation. which is located here:
`obspy\misc\docs\source\tutorial\code_snippets\quakeml_custom_tags.rst`
I decided to leave that as is because currently that snippet works on all Python versions and adding if would make it more complicated. We could change that when we will be dropping support for PY2.

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -23,7 +23,6 @@ from socket import timeout as socket_timeout
 import textwrap
 import threading
 import warnings
-from collections import OrderedDict
 
 from lxml import etree
 
@@ -40,10 +39,13 @@ if PY2:
     from urllib import urlencode
     import urllib2 as urllib_request
     import Queue as queue  # NOQA
+    from collections import OrderedDict
+
 else:
     from urllib.parse import urlencode
     import urllib.request as urllib_request
     import queue
+    from collections.abc import OrderedDict
 
 
 DEFAULT_SERVICE_VERSIONS = {'dataselect': 1, 'station': 1, 'event': 1}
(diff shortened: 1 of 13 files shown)


heavelock:
Well, CI is kinda out of order for now so it needs to wait until #2424 is merged.

megies:
Looks good @heavelock  and thanks for the review @QuLogic

megies:
@heavelock @QuLogic  what is it with this change, actually? Current CI fails because it's not finding the import it was changed to here?

https://travis-ci.org/obspy/obspy/jobs/569792961#L2544

heavelock:
Oh wow, that's totally my fault. In fact, OrderedDicts and the rest of stuff changed here are not ABC themselves thus they are staying in main `collections` module. 

The real problem that was causing this warning is located elsewhere IMO.
In `obspy\clients\fdsn\mass_downloader\restrictions.py` there is used bare import of `collections` and then `collections.Iterable` instead of going through:
```
from obspy.core import compatibility
compatibility.collections_abc.Iterable` as it should be.
```
This PR needs to be reverted since none of those imports is working. I'm sorry.

megies:
No big deal @heavelock, no harm done really. CI is failing anyway right now most of the time, so nobody even noticed for some weeks

Issue #1741, opened 2017-03-31
Support waveform/stream stacking


seisman:
Waveform stacking is frequently used in seismic data processing. Popular stacking methods are:

- linear stack
- Nth root stack
- phase weighted stack

I will try to implement it if it's possible to be included into obspy, maybe `obspy.Stream` or `obspy.signal`.

calum-chamberlain:
Linear and phase-weighted stacks at implemented in eqcorrscan, would be happy to contribute them here if wanted?

ThomasLecocq:
@calum-chamberlain I'm happy to contribute here too with the/comparing with the pws of msnoise.

krischer:
This (almost done ;-) ) branch (https://github.com/obspy/obspy/pull/1002) also contains these stacks - so work should probably invested into finishing that branch and then generalizing the functionality.

ThomasLecocq:
👍 for generalizing the stacks outside the array stuff... so investing on #1002 to do that is the way to go !

trichter:
Implemented in #2440.

PR #2440, opened 2019-08-22
add Stream.stack method

trichter:
### What does this PR do?

I added a stack method to the Stream object. Any discussion about the implementation is welcome before adding some tests.

See #1741, and maybe #841, #1002

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3154,11 +3154,11 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         1 Trace(s) in Stream:
         BW.RJOB.. | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
         """
+        if group == 'seedid':
+            group = '{network}.{station}.{location}.{channel}'
         if len(type) == 2:
             type, order = type
         groups = collections.defaultdict(list)
-        if group == 'seedid':
-            group = '{network}.{station}.{location}.{channel}'
         for tr in self:
             groups[group.format(**tr.stats)].append(tr)
         stacks = []
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -28,6 +28,7 @@ import numpy as np
 from obspy.core import compatibility
 from obspy.core.trace import Trace
 from obspy.core.utcdatetime import UTCDateTime
+from obspy.core.util.attribdict import AttribDict
 from obspy.core.util.base import (ENTRY_POINTS, _get_function_from_entry_point,
                                   _read_from_plugin, _generic_reader)
 from obspy.core.util.decorator import (map_example_filename,
@@ -3119,11 +3120,13 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             tr.remove_sensitivity(*args, **kwargs)
         return self
 
-    def stack(self, group='all', type='linear', npts_tol=0):
+    def stack(self, group_by='all', stack_type='linear', npts_tol=0,
+              time_tol=0):
         """
         Return stream with traces stacked by the same selected metadata.
 
-        :param str group: Stack waveforms together which have the same metadata
+        :type group_by: str
+        :param group_by: Stack waveforms together which have the same metadata
             given by this parameter. The parameter should name the
             corresponding keys of the stats object,
             e.g. ``'{network}.{station}'`` for stacking all
@@ -3132,14 +3135,19 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             This parameter can take two special values,
             ``'id'`` which stacks the waveforms by SEED id and
             ``'all'`` (default) which stacks together all traces in the stream.
-        :type type: str or tuple
-        :param type: Type of stack, one of the following:
+        :type stack_type: str or tuple
+        :param stack_type: Type of stack, one of the following:
             ``'linear'``: average stack (default),
             ``('pw', order)``: phase weighted stack of given order,
             see [Schimmel1997]_,
             ``('root', order)``: root stack of given order.
-        :param int npts_tol: Tolerate traces with different number of points
+        :type npts_tol: int
+        :param npts_tol: Tolerate traces with different number of points
             with a difference up to this value. Surplus samples are discarded.
+        :type time_tol: float (seconds)
+        :param time_tol: Tolerate difference in startime when setting the
+            new starttime of the stack. If starttimes differs more than this
+            value it will be set to timestamp 0.
 
         :returns: New stream object with stacked traces. The metadata of each
             trace (inlcuding starttime) corresponds to the metadata of the
@@ -3154,18 +3162,34 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         >>> print(stack)  # doctest: +ELLIPSIS
         1 Trace(s) in Stream:
         BW.RJOB.. | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
+
+        .. note::
+
+            This operation is performed in place on the actual data arrays. The
+            raw data will no longer be accessible afterwards. To keep your
+            original data, use :meth:`~obspy.core.stream.Stream.copy` to create
+            a copy of your stream object.
         """
-        if group == 'id':
-            group = '{network}.{station}.{location}.{channel}'
+        if group_by == 'id':
+            group_by = '{network}.{station}.{location}.{channel}'
         groups = collections.defaultdict(list)
         for tr in self:
-            groups[group.format(**tr.stats)].append(tr)
+            groups[group_by.format(**tr.stats)].append(tr)
         stacks = []
         for groupid, traces in groups.items():
             header = {k: v for k, v in traces[0].stats.items()
                       if all(tr.stats.get(k) == v for tr in traces)}
-            header['stack'] = groupid
-            header['stack_count'] = len(traces)
+            header.pop('endtime', None)
+            if 'sampling_rate' not in header:
+                msg = 'Sampling rate of traces to stack is different'
+                raise ValueError(msg)
+            if 'starttime' not in header and time_tol > 0:
+                times = [tr.stats.starttime for tr in traces]
+                if np.ptp(times) <= time_tol:
+                    # use high median as starttime
+                    header['starttime'] = sorted(times)[len(times) // 2]
+            header['stack'] = AttribDict(group=groupid, count=len(traces),
+                                         type=stack_type)
             npts_all = [len(tr) for tr in traces]
             npts_dif = np.ptp(npts_all)
             npts = min(npts_all)
@@ -3174,24 +3198,26 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
                        ' than requested tolerance ({} > {})')
                 raise ValueError(msg.format(npts_dif, npts_tol))
             data = np.array([tr.data[:npts] for tr in traces])
-            if type == 'linear':
+            if stack_type == 'linear':
                 stack = np.mean(data, axis=0)
-            elif type[0] == 'pw':
+            elif stack_type[0] == 'pw':
                 from scipy.signal import hilbert
                 from scipy.fftpack import next_fast_len
                 nfft = next_fast_len(npts)
                 anal_sig = hilbert(data, N=nfft)[:, :npts]
                 norm_anal_sig = anal_sig / np.abs(anal_sig)
-                phase_stack = np.abs(np.mean(norm_anal_sig, axis=0)) ** type[1]
+                phase_stack = np.abs(
+                        np.mean(norm_anal_sig, axis=0)) ** stack_type[1]
                 stack = np.mean(data, axis=0) * phase_stack
-            elif type[0] == 'root':
+            elif stack_type[0] == 'root':
                 r = np.mean(np.sign(data) * np.abs(data)
-                            ** (1 / type[1]), axis=0)
-                stack = np.sign(r) * np.abs(r) ** type[1]
+                            ** (1 / stack_type[1]), axis=0)
+                stack = np.sign(r) * np.abs(r) ** stack_type[1]
             else:
                 raise ValueError('stack type is not valid.')
             stacks.append(traces[0].__class__(data=stack, header=header))
-        return self.__class__(stacks)
+        self.traces = stacks
+        return self
 
     @staticmethod
     def _dummy_stream_from_string(s):
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3135,7 +3135,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         :param type: Type of stack, one of the following:
             `'linear'`: average stack (default),
             `('pw', order)`: phase weighted stack of given order,
-            see Schimmel and Paulssen (1997),
+            see [Schimmel1997]_,
             `('root', order)`: root stack of given order.
         :param npts_tol: Tolerate traces with different number of points
             with a difference up to this value. Surplus samples are discarded.
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3123,7 +3123,15 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
     def stack(self, group_by='all', stack_type='linear', npts_tol=0,
               time_tol=0):
         """
-        Return stream with traces stacked by the same selected metadata.
+        Stack traces by the same selected metadata.
+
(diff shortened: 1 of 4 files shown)


obspy/core/stream.py
```diff
@@ -3118,6 +3119,78 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        :param group: Stack waveforms together which have the same metadata given
+            by this parameter. The parameter should name the corresponding keys
+            of the stats object, e.g. `'{network}.{station}'` for stacking all
+            locations and channels of the stations and returning a stream
+            consisting of one stacked trace for each station. This parameter can
+            take two special values, `'seedid'` which stacks the waveforms by
+            seedid and `'all'` (default) which stacks together
+            all traces in the stream.
+        :param type: Type of stack, one of the following:
+            `'normal'`: normal stack (default),
```
QuLogic: The 'normal' stack is normal, is a bit of a tautology.

ThomasLecocq: 'normal' should be 'linear' or 'average', maybe?

trichter:
Thanks for the comments, +DOCS

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+                from scipy.signal import hilbert
+                from scipy.fftpack import next_fast_len
+                nfft = next_fast_len(npts)
+                anal_sig = hilbert(data, N=nfft)[:, :npts]
+                norm_anal_sig = anal_sig / np.abs(anal_sig)
+                phase_stack = np.abs(np.mean(norm_anal_sig, axis=0)) ** order
+                stack = np.mean(data, axis=0) * phase_stack
+            elif type == 'root':
+                r = np.mean(np.sign(data) * np.abs(data)
+                            ** (1 / order), axis=0)
```
ThomasLecocq: is ``1/order`` safe on all python versions?

trichter: I think its not. I am used to py3, now.

ThomasLecocq: me to and 1.2.0 will be "py3 only", but in case of...

trichter: Probably ;)

trichter:
From my point of view, this PR is ready. Here is a link to the relevant doc: [[l]](http://docs.obspy.org/pull-requests/2440/packages/autogen/obspy.core.stream.Stream.stack.html)

obspy/core/stream.py
trichter: Actually, there is this: `from __future__ import division`

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        """
+        if group == 'seedid':
+            group = '{network}.{station}.{location}.{channel}'
+        if len(type) == 2:
+            type, order = type
+        groups = collections.defaultdict(list)
+        for tr in self:
+            groups[group.format(**tr.stats)].append(tr)
+        stacks = []
+        for groupid, grouptrcs in groups.items():
```
QuLogic: Don't think you need to shorten `grouptrcs`; maybe just `traces` will do?

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+            header = {k: v for k, v in grouptrcs[0].stats.items()
+                      if all(tr.stats.get(k) == v for tr in grouptrcs)}
+            header['stack'] = groupid
+            header['stack_count'] = len(grouptrcs)
+            npts_all = [len(tr) for tr in grouptrcs]
+            npts_dif = max(npts_all) - min(npts_all)
```
QuLogic:
```suggestion
            npts_dif = np.ptp(npts_all)
```

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+            npts = min(npts_all)
+            if npts_dif > npts_tol:
+                msg = ('Difference of number of points of the traces is higher'
+                       ' than requested tolerance ({} > {})')
+                raise ValueError(msg.format(npts_dif, npts_tol))
+            elif npts_dif > 0:
+                grouptrcs = [copy.copy(tr) for tr in grouptrcs]
+                for tr in grouptrcs:
+                    tr.data = tr.data[:npts]
+            data = np.array([tr.data for tr in grouptrcs])
```
QuLogic: You could
```suggestion
            data = np.array([tr.data[:npts] for tr in grouptrcs])
```
and not bother with copying above.

obspy/core/stream.py
d-chambers: nice trick, I didn't know there was a ptp in numpy!

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        Return stream with traces stacked by the same selected metadata.
+
+        :param group: Stack waveforms together which have the same metadata
+            given by this parameter. The parameter should name the
+            corresponding keys of the stats object,
+            e.g. `'{network}.{station}'` for stacking all
+            locations and channels of the stations and returning a stream
+            consisting of one stacked trace for each station.
+            This parameter can take two special values,
+            `'seedid'` which stacks the waveforms by seedid and
```
d-chambers: I would consider chaning this to 'id' to be consistently named with the `Trace.id` property. Also, describe it as "seed id" and not "seedid".

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
             tr.remove_sensitivity(*args, **kwargs)
         return self
 
+    def stack(self, group='all', type='linear', npts_tol=0):
+        """
+        Return stream with traces stacked by the same selected metadata.
+
+        :param group: Stack waveforms together which have the same metadata
```
d-chambers: I recommend also adding `:type group: str` so it is clear this must be a string. Same for the other parameters.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+            `('root', order)`: root stack of given order.
+        :param npts_tol: Tolerate traces with different number of points
+            with a difference up to this value. Surplus samples are discarded.
+
+        :returns: New stream object with stacked traces. The metadata of each
+            trace corresponds to the metadata of the original traces
+            if those are the same. Additionaly, the entries
+            `stack` (result of the format operation on the group parameter) and
+            `stack_count` (number of stacked traces)
+            are written to the stats object.
```
d-chambers: maybe "stats object(s)" as the resulting stream could return more than one trace.

obspy/core/tests/test_stream.py
```diff
@@ -2656,6 +2656,59 @@ def test_write_empty_stream(self):
             self.assertEqual(e.exception.args[0],
                              'Can not write empty stream to file.')
 
+    def test_stack(self):
```
d-chambers: There is a lot going on here, do you mind breaking this up into a few different tests? Or maybe just add a few more  comments about what each subtest is doing if you prefer.

obspy/core/tests/test_stream.py
```diff
@@ -2656,6 +2656,59 @@ def test_write_empty_stream(self):
...
+        """
+        st = read()
+        st2 = st.stack()
+        self.assertEqual(len(st2), 1)
+        self.assertEqual(st2[0].stats.stack, 'all')
+        self.assertEqual(st2[0].stats.stack_count, 3)
+        st += st.copy()
+        st2 = st.stack('seedid')
+        self.assertEqual(len(st2), 3)
+        self.assertEqual({tr.stats.stack for tr in st2}, {tr.id for tr in st})
```
d-chambers: Should this preserve order? If so use a list or tuple rather than set. Same for other set asserts below.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+            `'all'` (default) which stacks together all traces in the stream.
+        :param type: Type of stack, one of the following:
+            `'linear'`: average stack (default),
+            `('pw', order)`: phase weighted stack of given order,
+            see [Schimmel1997]_,
+            `('root', order)`: root stack of given order.
+        :param npts_tol: Tolerate traces with different number of points
+            with a difference up to this value. Surplus samples are discarded.
+
+        :returns: New stream object with stacked traces. The metadata of each
```
d-chambers: I think returning a new `Stream` instance is the best approach, but would it make sense to perform this operation in place in order to be consistent with most other `Stream` methods?

obspy/core/stream.py
```diff
@@ -3118,6 +3119,85 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        >>> stack = st.stack()
+        >>> print(stack)  # doctest: +ELLIPSIS
+        1 Trace(s) in Stream:
+        BW.RJOB.. | 2009-08-24T00:20:03.000000Z - ... | 100.0 Hz, 3000 samples
+        """
+        if group == 'seedid':
+            group = '{network}.{station}.{location}.{channel}'
+        if len(type) == 2:
+            type, order = type
+        groups = collections.defaultdict(list)
```
d-chambers: Nice use of collections!

trichter:
@QuLogic @d-chambers  thanks for the stimulant comments, I will take care of them in a week or so.

obspy/core/tests/test_stream.py
trichter: No, the order is not important. It will only be preserved on python>3.4. Also, the number of traces of the stack and the original stream is different, so list or tuple would not be sufficient. The keys of an OrderedDict would be.

obspy/core/stream.py
trichter: I also think returning a new Stream is the best approach here, because the number of traces changes and a lot of metadata changes, which is not the case for the other, in-place methods.

trichter:
> One major question I have is how this handles temporal alignment. It seems the Stack method does not consider the starttime / endtime of each trace. This is fine, as often one may wish to stack data from different times, but I think this point should be mentioned in the doc string. If the user wants to stack data only from overlapping times it is up to the user to align the traces.

This is done by these lines

```
            """
            :returns: New stream object with stacked traces. The metadata of each
                 trace corresponds to the metadata of the original traces
                 if those are the same.
            """
            ...
            header = {k: v for k, v in grouptrcs[0].stats.items()
                      if all(tr.stats.get(k) == v for tr in grouptrcs)}
```
This means the starttime is preserved if it is the same for all traces in one group. In all other cases the user has to take care of it. I think I will add a test for this and maybe make it more clear in the doc.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,80 @@ def remove_sensitivity(self, *args, **kwargs):
             tr.remove_sensitivity(*args, **kwargs)
         return self
 
+    def stack(self, group='all', type='linear', npts_tol=0):
+        """
+        Return stream with traces stacked by the same selected metadata.
+
+        :param str group: Stack waveforms together which have the same metadata
```
krischer:
```suggestion
        :param group: Stack waveforms together which have the same metadata
```

obspy/core/stream.py
krischer: I see two solutions here:

(1) Actually change the existing stream object in-place. This is what a few other method, e.g. `.merge()` already do and it would fit right in with most other stream methods.
(2) Rename the method to `.create_stack()` which would return a new object. Then its IHMO a bit clearer that this does not modify in-place.

I slightly tend towards the first choice but I'm fine with either.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,80 @@ def remove_sensitivity(self, *args, **kwargs):
             tr.remove_sensitivity(*args, **kwargs)
         return self
 
+    def stack(self, group='all', type='linear', npts_tol=0):
```
krischer: Could you rename `type` to `stack_type` to not clash with the reserved `type` word?

krischer: Also maybe rename `group` to `group_by` which is what most other packages like pandas would probably call an operation like this.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,80 @@ def remove_sensitivity(self, *args, **kwargs):
...
+            group = '{network}.{station}.{location}.{channel}'
+        groups = collections.defaultdict(list)
+        for tr in self:
+            groups[group.format(**tr.stats)].append(tr)
+        stacks = []
+        for groupid, traces in groups.items():
+            header = {k: v for k, v in traces[0].stats.items()
+                      if all(tr.stats.get(k) == v for tr in traces)}
+            header['stack'] = groupid
+            header['stack_count'] = len(traces)
```
krischer: We don't really have a good concept for this type of meta-data for the stack so I also don't really know what's the best way here. But maybe move these two to a new `stack` group in the `stats` dictionary? Really just a suggestion though.

obspy/core/stream.py
```diff
@@ -3118,6 +3119,80 @@ def remove_sensitivity(self, *args, **kwargs):
...
+        """
+        if group == 'id':
+            group = '{network}.{station}.{location}.{channel}'
+        groups = collections.defaultdict(list)
+        for tr in self:
+            groups[group.format(**tr.stats)].append(tr)
+        stacks = []
+        for groupid, traces in groups.items():
+            header = {k: v for k, v in traces[0].stats.items()
+                      if all(tr.stats.get(k) == v for tr in traces)}
```
krischer: I think the function should raise if the sampling rate is not set after this line - otherwise the Trace constructor will set it to 1.0 which is likely wrong and will be confusing for people.

This line IHMO already works really well for the start time - it will either be preserved if stacking different channels/stations that happen to have the same start time or be set to timestamp 0 if stacking multiple time periods. You could consider to allow for a small tolerance for which the start time would still be preserved but that might be overkill.

obspy/core/stream.py
trichter: Just added this, because @d-chambers asked to document the types. For me, it does not matter.

obspy/core/stream.py
trichter: So, you would prefer `stats.stack.id/group` and `stats.stack.count`? I don't know, I think I would prefer slightly the flat variant, but I am also OK with this one.

obspy/core/stream.py
trichter: Yeah, your are right about the sampling_rate.

I was also thinking about the starttime tolerance and thought it would be overkill, but on a second thought this might be pretty handy.

obspy/core/stream.py
trichter: Hmm, one could also add 'stats.stack.type'.

obspy/core/stream.py
krischer: Then you have to use `:type group: str` as you already did for one of the other parameters. I'm not sure if sphinx can use the syntax you originally used but we probably stick to a consistent docstyle syntax.

obspy/core/stream.py
krischer: Yea that might work. I slightly prefer to group all the `stack` meta parameters which I think would look a bit nicer.

obspy/core/stream.py
trichter: I prefer (2) with the name of (1) ;).
But I am also OK with (1) and will change the behavior.

obspy/core/stream.py
trichter: @krischer can you clarify a bit. IMHO it is not necessary to document the basic types, but it also does not hurt. Do you also want me to remove the `int` from `int npts_tol`?

krischer: For the `group` parameter I would document the types as it is not obvious and it can take a string or a tuple. In general I also think its nice to be consistent within a single docstring and if a single type is documented, all should be documented.

But its all a bit moot as we'll hopefully drop Python 2 support very soon and can then just use type hints for documentation so we'll have some work to do in any case.

trichter: Yeah sphinx is OK with this syntax, I use it all the time, because the other form is a bit verbose, but I can of course change it.

trichter:
I implemented @krischer's review and rebased on current master.

trichter:
This is again ready for the next step, be it another review or merging.

krischer:
I think this looks good now. Thanks a lot!

trichter:
It was a pleasure, I think I am going to use this method a lot.

Issue #2228, opened 2018-10-05
Dropping Python 2


megies:
This is just intended as a place to summarize any things to keep in mind once we drop Python 2.7.
Just add bullet points for any things that should be done after dropping Py2K, e.g. making use of Python 3 only features in code.

## High Priority
- [ ] bump major version to 2
- [ ] drop basemap, full switch to cartopy
- [x] remove usages of the `future` package (see #1752 and #1613)
- [ ] make sure docs discourage using pip < 9.0 or setuptools < 24.3
- [x] make sure docs discourage using the setup.py directly (favor using pip)
- [x] raise an informative exception if the user tries to import obspy > 2.0.0 on python 2 as recommended [here](http://python3statement.org/practicalities/#fail-early-at-import-time)
- [x] raise an informative exception if setup.py is run by python 2 as recommended [here](http://python3statement.org/practicalities/#fail-early-in-setuppy)
- [ ] replace usages of urllib with requests (see #1254)
- [ ] add obspy to the [python 3 statement timeline](http://www.python3statement.org/) (see #1764)
- [x] bumpy dependecy versions, e.g. numpy to 1.15, see [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)
- [ ] ...
## Low Priority
- [x] remove usages of the `__future__` imports (perhaps only until [pep 563](https://www.python.org/dev/peps/pep-0563/) is applicable) 
- [ ] add type hints to public functions/methods (or use stub files?)
- [x] maybe can revert part of #2167? Mostly affects Py2 I think
- [ ] use pathlib
- [ ] ...

## Handle some deprecations on 3.7
- [x] `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`

d-chambers:
Would it make sense for a major version bump after dropping 2.7?

krischer:
> Would it make sense for a major version bump after dropping 2.7?

Yes, I think we should do this. Seems to be how all the other big projects handle it as well.

In terms of actually doing it we could follow some form of this strategy: https://blog.jupyter.org/release-of-ipython-6-0-2068cfb2c2a

Probably on in light-version as might not have the manpower to do it at full scale.

QuLogic:
I applied the changes that IPython did into Matplotlib; it is fairly straightforward and should be easy to do for ObsPy.

d-chambers:
[...]

PR #2442, opened 2019-08-28
Importing ABCs through compatibility layer and fix of proj_version check

heavelock:
Okay, this time I hope it is done properly...

I replaced bare imports of Abstract Base Classes to go through compatibility layer. It is necessary for compatibility between 2.7 and 3.8 since 3.8 drops possibility of ABC import directly from `collections` and requires import from `collections.abc`.
I substituted imports only for ABCs so `Mapping`, `Iterable` and `Sequence`. There are only so few changes because in other cases the imports were done already in that way.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -385,7 +385,7 @@ def get_proj_version(raw_string=False):
     _proj = Proj(proj='utm', zone=10, ellps='WGS84')
     if hasattr(_proj, 'proj_version_str'):
         version_string = str(getattr(_proj, 'proj_version_str'))
-    elif hasattr(_proj, 'proj_version')::
+    elif hasattr(_proj, 'proj_version'):
         version_string = str(getattr(_proj, 'proj_version'))
     else:
         from pyproj import proj_version_str
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -383,9 +383,7 @@ def get_proj_version(raw_string=False):
     # of the Proj class that is only set when the projection is made. Make
     # a dummy projection and get the version
     _proj = Proj(proj='utm', zone=10, ellps='WGS84')
-    if hasattr(_proj, 'proj_version_str'):
-        version_string = str(getattr(_proj, 'proj_version_str'))
-    elif hasattr(_proj, 'proj_version'):
+    if hasattr(_proj, 'proj_version'):
         version_string = str(getattr(_proj, 'proj_version'))
     else:
         from pyproj import proj_version_str
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -383,7 +383,10 @@ def get_proj_version(raw_string=False):
     # of the Proj class that is only set when the projection is made. Make
     # a dummy projection and get the version
     _proj = Proj(proj='utm', zone=10, ellps='WGS84')
-    version_string = str(getattr(_proj, 'proj_version_str', 'proj_version'))
+    if hasattr(_proj, 'proj_version_str'):
+        version_string = str(getattr(_proj, 'proj_version_str'))
+    else:
+        version_string = str(getattr(_proj, 'proj_version'))
     if raw_string:
         return version_string
     version_list = [to_int_or_zero(no) for no in version_string.split(".")]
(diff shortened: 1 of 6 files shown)


d-chambers:
Hey @heavelock, 

Thanks for fixing this. These warnings were also starting to annoy me but not quite enough to do a PR myself :wink: 

Can you please change lines 385-388 (ish) in obspy.core.util.base to this:

```python
    _proj = Proj(proj='utm', zone=10, ellps='WGS84')
    version_string = str(getattr(_proj, 'proj_version_str', 'proj_version'))
    if raw_string:
        return version_string
```
or something similar to fix Travis failures on python 3.7? Since this is a compatibility type PR I think it belongs here. I don't think it is that important to address all the 2.7 failures before merge, but we should at least keep make sure 3.7 runs.

heavelock:
Yup, here it comes. On my 3.7 runs, let's see the CI.

megies:
Python 3.8 has b4 out, btw, might be worthwile checking that as well

heavelock:
As far as I can see guys at coda-forge are still working on preparing py3.8 package so we have to wait for them. https://github.com/conda-forge/python-feedstock/pull/271

obspy/core/util/base.py
```diff
@@ -382,7 +382,8 @@ def get_proj_version(raw_string=False):
     # proj4 is a c library, prproj wraps this.  proj_version is an attribute
     # of the Proj class that is only set when the projection is made. Make
     # a dummy projection and get the version
-    version_string = str(Proj(proj='utm', zone=10, ellps='WGS84').proj_version)
+    _proj = Proj(proj='utm', zone=10, ellps='WGS84')
+    version_string = str(getattr(_proj, 'proj_version_str', 'proj_version'))
```
megies: I'm pretty sure this is not doing what you want it to do.

I would just do a `if hasattr(...):` if/else

heavelock: Yup, you are right. The third argument is just a default value.

heavelock:
Hmm, Travis has pyproj 2.3.1 which seems not to have attribute proj_version in the Proj class. it has however `pyproj.proj_version_str`. Here it's being used directly http://pyproj4.github.io/pyproj/stable/_modules/pyproj/_show_versions.html#show_versions
Let's see.

heavelock:
I removed a check for `proj_version_str` attribute because it was pointless - that version of attribute was added  i think somewhere >2.0 so it would not work anyway. In 2.3.1 import directly from pyproj works.

heavelock:
In order to solve those fails at py37 we would need to pin basemap to 1.2.1. In that versionn they copied data to their own repo:
https://github.com/matplotlib/basemap/commit/84ddf4568c05c7a0ed18025b2f65c7d7fa888e85
Should I do it?

d-chambers:
Oops, my bad sorry for misleading you @heavelock.

d-chambers:
> In order to solve those fails at py37 we would need to pin basemap to 1.2.1 

I think this is a sensible thing to do.  Dropping basemap is on the [python 3 road map](https://github.com/obspy/obspy/issues/2228) so it would be only for obspy 1.2 anyway. Do you see any issue @megies?

megies:
> I think this is a sensible thing to do. Dropping basemap is on the [python 3 road map](https://github.com/obspy/obspy/issues/2228) so it would be only for obspy 1.2 anyway. Do you see any issue @megies?

I don't care too much, thats packaging problems of other packages so whatever works for our CI I guess

heavelock:
I also added E/W504 to be ignored in circleci since we talked about that in #2444

heavelock:
Okay, so it turnes out there is a conflict in requirements... Any clues how should I solve it?
```
Package proj4 conflicts for:

basemap=1.2.1 -> proj4[version='>=4.9.3,<6|>=5.2.0,<5.2.1.0a0']

basemap=1.2.1 -> pyproj[version='>=1.9.3'] -> proj4[version='>=5.2,<6']

proj4=4.9.3

pyproj -> proj4[version='6.0.0.*|>=4.9.3,<4.9.4.0a0|>=5.0.1,<5.0.2.0a0|>=5.1.0,<5.1.1.0a0|>=5.2,<6|>=5.2.0,<5.2.1.0a0|>=6.0.0,<6.0.1.0a0|>=6.1.0,<6.1.1.0a0|>=6.1.1,<6.1.2.0a0']
```

megies:
Lookin good, thanks a lot @heavelock :tada:

PR #2444, opened 2019-09-03
Fixing CircleCI

heavelock:
I noticed that CircleCi is not working due to lack of conda on offician conda image. Maybe it's because commands were not using full path to conda?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,13 +14,10 @@ jobs:
       - run:
           name: install dependencies
           command: |
-            /opt/conda/bin/conda create -n test-env
-            /opt/conda/bin/conda activate test-env
             /opt/conda/bin/conda install -y flake8 pep8-naming
 
       - run:
           name: run linting
           command: |
-            source activate test-env
             flake8 --version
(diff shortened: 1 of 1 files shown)


heavelock:
Okay, linting works. In short, I added fullpaths to both conda and flake8 executables and that was it. I also removed whole part of environment creation, I think it is a bit redundant since we are anyway spinning up a new image every.

d-chambers:
Would ignoring W504 for now make sense? There was some talk of using black in the future which would fix all of those types of issues. Over than W504, it looks like there are only a handful of formatting errors to deal with.

heavelock:
I'm also all for black but for now I just wanted to make the CircleCI running again :)

d-chambers:
Yes, and it looks like it runs now, which is a huge improvement. Do you want to merge this now that the linter runs and deal with the linting issues in a different PR?

heavelock:
Yup, I think the linting fixes should be separated out, it might get messy otherwise. And people doing PRs in meantime might fix their own mess.

heavelock:
And to be honest, looking at how many of E504 there are and what that error means, I think it would be good to ignore them. It will be super useless work to fix that since we hope to move to black.

megies:
+1 for ignoring e/w504

megies:
Thanks @heavelock

PR #2454, opened 2019-09-13
make interphinx point to latest python version

trichter:
Intersphinx links in the docs still point to python2.7.
+DOCS

### DIFF
--- a/misc/docs/source/conf.py
+++ b/misc/docs/source/conf.py
@@ -326,7 +326,7 @@ man_pages = [
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
-    'python': ('https://docs.python.org/2.7/', None),
+    'python': ('https://docs.python.org/', None),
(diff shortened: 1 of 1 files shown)


QuLogic:
The build succeeded with warnings, though they seem unrelated. Unfortunately, I don't know where to look to verify this didn't break anything. Anyone know somewhere we might be referencing Python types?

trichter:
[Here](http://docs.obspy.org/pull-requests/2454/packages/autogen/obspy.clients.fdsn.mass_downloader.restrictions.Restrictions.html) are some references and it looks as it is working as expected.

megies:
> The build succeeded with warnings, though they seem unrelated. Unfortunately, I don't know where to look to verify this didn't break anything. Anyone know somewhere we might be referencing Python types?

Yeah, unfortunately nobody ever got around to getting rid of all Sphinx warnings, some of the Sphinx warnings are also really weird, so it's unlikely we will ever get `master` to a clean build

PR #2459, opened 2019-09-17
core: Unify "_is_xxx()" file format checks regarding file handling

megies:
### What does this PR do?

This PR aims at unifying I/O handling during file format checks performed in I/O plugins. It introduces a decorator to be used on `_is_xxx_()` file format checker funtions, that unifies opening binary buffers or rewinding already open files or file-like objects to the initial position after the file format check was performed.

Depending if CI looks OK for the first two format checks already migrated (MSEED, SAC), we should also migrate all other check routines.

#### Summary of most important changes

 - new `file_format_check` decorator that (basically taken from the old `_is_mseed` routine)
   - opens bytes buffer from filenames etc. and closes it at the end
   - passes on open file-like objects (and leaves them open at end)
   - creates `BytesIO`/`StringIO` buffers from bytes/string input
 - some is-format checker routines were behaving differently. most of this was unified and all should be. e.g. some were raising exceptions on certain incompatible input while others just always return False. I think they should probably always return False and never raise an Exception, in any case this part is now exclusively in the decorator, the is-format check routines will now **always** receive a bytes or string buffer to work on
 - some is-format checkers used to behave different with decoding things, added a helper method to wrap a `TextIOWrapper` around the bytes buffer handed over by the new decorator

Currently this logic is duplicated, present to some extent or sometimes completely missing and spread out across 48 `isFormat` routines. This is aiming in the similar direction as #2326 was for reader routines, so there might be room for further unification with that one later on..

### PR Checklist
- [ ] All tests still pass.
- [x] purely internal ~~Significant changes have been added to `CHANGELOG.txt` .~~

### TODOS
 - [x] add a test that goes through all `isFormat` functions registered as plugins, passes in an opened `io.BytesIOBase` object, runs format check, makes sure the object is not closed
 - [x] add a test that runs one file format check with an open binary stream not at start of file and makes sure that after format check the decorator properly rewinds the stream to initial position (compare #2419)
 - [x] migrate all other file format checks if PR gets approved
 - [ ] do we need a check on `_file_size` kwarg collision? i.e. if it is problematic when defined in the decorated function as well
 - [ ] maybe rename from `file_format_check` to something more general like `ensure_open_buffer`
 - [ ] `_open_file` helper moved from rg16 to core/util should maybe be integrated with the `file_format_checker` as well
 - [ ] opening remote sources (http) could be integrated as well
 - [ ] integrate `uncompress` decorator into `file_format_checker` as well?
 - [ ] maybe see if the `_generic_reader` and `file_format_checker` can be combined into one thing
 - [ ] should check again if we can wrap a user-provided string buffer with another encoding by wrapping a second Text wrapper around it again. it did not work earlier but it might have been a problem with what was fixed by e6ff5d276c763ea389d2ee44f90748ac5a746b9f
 - [ ] when a open file-like object not at start of file is passed in, what should be reported as "file size" to the file format checker? Total file size (current behavior) or remaining file size from current position in file (might make more sense).

Also, right now, an opened text stream with automatic decoding will be passed down into the plugins' format check function. It might not be possible to do much about this, any ideas?

### DIFF
--- a/obspy/io/nied/knet.py
+++ b/obspy/io/nied/knet.py
@@ -17,41 +17,11 @@ import numpy as np
 from obspy import UTCDateTime, Stream, Trace
 from obspy.core.trace import Stats
 from obspy.core.util.decorator import file_format_check
+from obspy.core.util.misc import _buffer_proxy
 
 
 class KNETException(Exception):
     pass
-
-
-def _buffer_proxy(filename_or_buf, function, reset_fp=True,
-                  file_mode="rb", *args, **kwargs):
-    """
-    Calls a function with an open file or file-like object as the first
-    argument. If the file originally was a filename, the file will be
-    opened, otherwise it will just be passed to the underlying function.
-
-    :param filename_or_buf: File to pass.
-    :type filename_or_buf: str, open file, or file-like object.
-    :param function: The function to call.
-    :param reset_fp: If True, the file pointer will be set to the initial
-        position after the function has been called.
-    :type reset_fp: bool
-    :param file_mode: Mode to open file in if necessary.
-    """
-    try:
-        position = filename_or_buf.tell()
-        is_buffer = True
-    except AttributeError:
-        is_buffer = False
-
-    if is_buffer is True:
-        ret_val = function(filename_or_buf, *args, **kwargs)
-        if reset_fp:
-            filename_or_buf.seek(position, 0)
-        return ret_val
-    else:
-        with open(filename_or_buf, file_mode) as fh:
-            return function(fh, *args, **kwargs)
 
 
 @file_format_check
--- a/obspy/io/nied/knet.py
+++ b/obspy/io/nied/knet.py
@@ -16,6 +16,7 @@ import numpy as np
 
 from obspy import UTCDateTime, Stream, Trace
 from obspy.core.trace import Stats
+from obspy.core.util.decorator import file_format_check
 
 
 class KNETException(Exception):
@@ -53,16 +54,16 @@ def _buffer_proxy(filename_or_buf, function, reset_fp=True,
             return function(fh, *args, **kwargs)
 
 
-def _is_knet_ascii(filename_or_buf):
+@file_format_check
+def _is_knet_ascii(filename_or_buf, **kwargs):
     """
     Checks if the file is a valid K-NET/KiK-net ASCII file.
 
-    :param filename_or_buf: File to test.
-    :type filename_or_buf: str or file-like object.
+    :type filename_or_buf: :class:`io.BytesIOBase`
+    :param filename_or_buf: Open file or file-like object to be checked
     """
     try:
-        return _buffer_proxy(filename_or_buf, _internal_is_knet_ascii,
-                             reset_fp=True)
+        return _internal_is_knet_ascii(filename_or_buf)
     # Happens for example when passing the data as a string which would be
     # interpreted as a filename.
     except (OSError, UnicodeDecodeError):
--- a/obspy/io/nied/knet.py
+++ b/obspy/io/nied/knet.py
@@ -10,6 +10,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 from future.builtins import *  # NOQA @UnusedWildImport
 
+import locale
 import re
 
 import numpy as np
@@ -17,7 +18,7 @@ import numpy as np
 from obspy import UTCDateTime, Stream, Trace
 from obspy.core.trace import Stats
 from obspy.core.util.decorator import file_format_check
-from obspy.core.util.misc import _buffer_proxy
+from obspy.core.util.misc import _buffer_proxy, _text_buffer_wrapper
 
 
 class KNETException(Exception):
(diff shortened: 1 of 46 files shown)


d-chambers:
@megies, Is this ready for review?

megies:
I think it is @d-chambers, I'll try to summarize changes as best as I can of the top of my head in first post

megies:
@d-chambers actually I'm still gonna make a few tweaks still, I think.

barsch:
@megies I would suggest to rename the parameter ```filename``` to something else, e.g. ```filehandler```, ```fh``` or whatever - filename is just wrong ...

currently it looks like this for all those _is_xxx() functions:

```
:type filename: :class:`io.BytesIOBase`
:param filename: Open file or file-like object to be checked
```

megies:
Yeah @barsch i was already changing some of those, and then decided to go back and leave what was in there before (to not break the call syntax). it's also not consistent across all those functions, we have `filename`, `path_or_filename`, `fh` to name just a few.

It's supposed to be private, and I'm making some changes to internal behavior anyway, so I guess might as well change that as well (I can't imagine anybody using it as a kwarg in code anyway)?

obspy/io/nied/knet.py
```diff
@@ -10,73 +10,61 @@
...
-    :param filename_or_buf: File to test.
-    :type filename_or_buf: str or file-like object.
+    :type filename_or_buf: :class:`io.BytesIOBase`
+    :param filename_or_buf: Open file or file-like object to be checked
+    :type encoding: str
+    :param encoding: Encoding of the file. Given setting is ignored if the
+        input is already a Text stream with a set encoding (default: default
+        system encoding)
+    :rtype: bool
+    :return: ``True`` if KNET ASCII file.
```
d-chambers: Same here about periods.

obspy/io/nied/knet.py
```diff
@@ -10,73 +10,61 @@
...
 
-def _internal_is_knet_ascii(buf):
+def _internal_is_knet_ascii(buf, encoding=None):
     """
     Checks if the file is a valid K-NET/KiK-net ASCII file.
 
     :param buf: File to read.
     :type buf: Open file or open file like object.
+    :type encoding: str
+    :param encoding: Encoding of the file. Given setting is ignored if the
```
d-chambers: period needed

obspy/io/scardec/core.py
```diff
@@ -44,47 +46,18 @@ def _get_resource_id(scardecname, res_type, tag=None):
...
-def _is_scardec(filename_or_buf):
+@file_format_check
+def _is_scardec(filename_or_buf, **kwargs):
     """
     Checks if the file is a SCARDEC file.
 
-    :param filename_or_buf: File to test.
-    :type filename_or_buf: str or file-like object.
+    :type filename_or_buf: :class:`io.BytesIOBase`
+    :param filename_or_buf: Open file or file-like object to be checked
```
d-chambers: period

obspy/io/stationtxt/core.py
```diff
@@ -76,37 +78,24 @@ def utf_8_encoder(unicode_csv_data):
...
-def is_fdsn_station_text_file(path_or_file_object):
+@file_format_check
+def is_fdsn_station_text_file(path_or_file_object, **kwargs):
     """
     Simple function checking if the passed object contains a valid FDSN
     station text file.
 
-    :param path_or_file_object: File name or file like object.
+    :type path_or_file_object: :class:`io.BytesIOBase`
+    :param path_or_file_object: Open file or file-like object to be checked
```
d-chambers: period

obspy/io/stationxml/core.py
```diff
@@ -54,41 +57,29 @@ def _is_stationxml(path_or_file_object):
     test for a StationXML file, many real world files are false negatives as
     they don't adhere to the standard.
 
-    :param path_or_file_object: File name or file like object.
+    :type path_or_file_object: :class:`io.BytesIOBase`
+    :param path_or_file_object: Open file or file-like object to be checked
```
d-chambers: period

obspy/io/wav/core.py
```diff
@@ -39,12 +40,13 @@
...
-def _is_wav(filename):
+@file_format_check
+def _is_wav(filename, **kwargs):
     """
     Checks whether a file is a audio WAV file or not.
 
-    :type filename: str
-    :param filename: Name of the audio WAV file to be checked.
+    :type filename: :class:`io.BytesIOBase`
+    :param filename: Open file or file-like object to be checked
```
d-chambers: period

obspy/io/win/core.py
```diff
@@ -13,41 +13,45 @@
...
-def _is_win(filename, century="20"):  # @UnusedVariable
+@file_format_check
+def _is_win(filename, **kwargs):
     """
     Checks whether a file is WIN or not.
 
-    :type filename: str
-    :param filename: WIN file to be checked.
+    :type filename: :class:`io.BytesIOBase`
+    :param filename: Open file or file-like object to be checked
```
d-chambers: period

krischer:
This is really a very much needed development! A few thoughts:

1. Instead of a decorator I would change the plug-in interfaces, e.g. the `_is_XXX()`, `_read_XXX()`, and `_write_XXX()` function to only accept `io.BufferedIOBase` objects. Then the generic reader/writer routines would make sure to convert whatever comes in to one of these objects.
2. This would be a bigger change and likely break all existing external file format plug-ins. Thus we'd have to move this to 2.0. But this is also much easier with Python 3 only so that might be a good thing.

megies:
OK, so even though I think many aspects would have been improvements for 1.2.0 I can see the beauty of making an even more drastic change (raise on any `read()` from decoded input) and pushing it to 2.0.

megies:
absolutely @heavelock , see also #1794

megies:
This is most likely very outdated by now, and likely superseded by e.g. #2326 and #3346, so closing this one

PR #2463, opened 2019-09-24
.flake8 config file, cleanup of flake8 related stuff, and pre-commit hook for flake8

heavelock:
### What does this PR do?

This PR extracts flake8 config to a single file and cleans up a bit flake8 related stuff. I also add a a pre-commit hook to run flake8 that checks automatically modified *.py files. I think that config file is quite relevant in order to ease up using flake8 consistently. 

The only thing left to do is to modify the documentation to include some information about pre-commit hooks and how to install them. I will add it if you think that pre-commit hooks should be added to obspy.

### PR Checklist
- [ ] All tests still pass.


megies:
Looks like a nice cleanup!

What does this `.pre-commit-config.yaml` actually do? 

We actually have some info on flake8 and pre-commit hooks in the docs.. but I guess most people never get to see that site.. http://docs.obspy.org/master/coding_style.html?highlight=hook

heavelock:
Thanks! I tried to dig into the implementation of flake8 legacy api but unfortunately it looks like there is no way of straightforward parsing the config file which could cleanup quite significantly ` obspy/core/tests/test_code_formatting.py`. 
Oh, I didn't see that mention before! The thing is that without `.pre-commit-config.yaml` as well as `.flake8` devs have to configure themselves all the hooks. In that new way, in order to have flake8 checks being run automatically, devs would need to:
1. `conda install pre-commit`
2. `pre-commit install`
Voila.
In fact, we can add `pre-commit` hook manager to the extra dependencies along with flake8 in order to ease up setup process.

And even without having the pre-commit hooks installed, running flake8 can become with this PR as easy as `flake8` since the config will be read automatically.

Btw, there are quite a few interesting hooks that seem to be useful here such as `mixed-line-ending` which can automatically fix problems with line ending characters or `check-added-large-files` which can reject commits containing too big files (size to be specified).

heavelock:
+DOCS
I added a section how to install pre-commit and the hooks themselves.

d-chambers:
Hey @heavelock, I am currently on Holiday so I haven't had a chance to look closely at this, but using pre-commit seems like a great idea. 

Will flake8 still run as part of the tests in the CI or will it just run before the test suite and prevent the actual tests from running if there is a non-compliant file found?  We will want to make sure we have something blocking non-compliant code in the CI as to not to rely on the hooks being run solely on the developers' machines.

heavelock:
The test running does not change at all, my changes make it simpler to run flake8 checks on your own. For example, previously to run flake8 with the same configuration we are using in CI, the easiest way was to either run a `CodeFormattingTestCase` class or to dig out the command we are using for running flake8 at CircleCI. With this PR merged, in order to run a proper flake8 check you just need to execute flake8. The test case is still the same, I just simplified it in the way I merged by hand all the ignores we are using (pycodestyle and flake8.get_style_guide). 
What is additional plus of that change is that built in linters will be probably also using that config and will be pointing out only errors that should be pointed out.

The pre-commit hooks are not obligatory part of the workflow, you can completely ignore it if you prefer. However, if you choose to use it, files you are trying to commit will be automatically checked for flake8 compliance(of course taking our flake8 configuration into account). If any flake8 erros will occur in the files, commit won't be done and you will be asked to fix the errors. Then you can either fix the mistakes or choose to ignore them and commit anyway with `git commit -n`.

heavelock:
+DOCS

megies:
>     2\. In fact, we can add `pre-commit` hook manager to the extra dependencies along with flake8 in order to ease up setup process.

Yeah I guess we could do that, but I think most people use conda anyway and don't bother with extra-flag and pip install? But yes, it can help as "documentation" I guess

megies:
> And even without having the pre-commit hooks installed, running flake8 can become with this PR as easy as `flake8` since the config will be read automatically.

Yeah, this part is clear and absolutely an improvement

megies:
@heavelock those other commit hooks you mentioned about file size and line endings seem interesting, I think it might not hurt to add them, since the user/dev has to proactively activate them anyway.

It would be really great to add a max file size  to CircleCI btw and fail CircleCI if max file size is exceeded :-)

.circleci/config.yml
```diff
@@ -21,3 +21,6 @@ jobs:
           command: |
             /opt/conda/bin/flake8 --version
             /opt/conda/bin/flake8 obspy
+          name: run file size check
+          command: |
+            git diff --name-only $CIRCLE_COMPARE_URL | xargs -I {} find {} -size +5 | wc -l
```
megies: Looks good.. even better would be to check size of all history in the PR, sometimes large files get added and later removed in another commit, still cluttering repo history size.
I tried this, but it seems it's not listing the size of all objects in the respective commit actually.. maybe just as an inspiration..

`for HASH in $(git log --pretty="format:%H" master..); do git cat-file -s $HASH; done`

.circleci/config.yml
```diff
@@ -21,3 +21,6 @@ jobs:
           command: |
             /opt/conda/bin/flake8 --version
             /opt/conda/bin/flake8 obspy
```
megies: looks like if you don't make it another `- run` section it overwrites the above flake 8 check?

heavelock: Yup, that's right. I will fix that

heavelock:
Damn, there is something weird in the image used by CircleCI. 
On my system ``if [[ $(git diff --name-only master | xargs -I {} find {} -size +500k | wc -l) > 0 ]]; then (exit 1); else (exit 0); fi`` is working perfectly but here the git diff seems to not return anything.

There is a $CIRCLE_COMPARE_URL env that is not working and and someone recommend to use some obscure orb to reconstruct it. that's absurd because the env vars are not persistent between shells so it requires manipulation with files. https://discuss.circleci.com/t/circle-compare-url-is-empty/24549

heavelock:
Okay, new approach. I based on your suggestion `for HASH in $(git log --pretty="format:%H" master..); do git cat-file -s $HASH; done` but instead of doing this loop I piped it into the `git cat-file --batch-check` and then manually checked if filesizes are lower than 500k. What should be the limit set to?

heavelock:
Nope, again it does not fail even if it works perfectly on my system. What the heck, CircleCI?

heavelock:
I ran the image that CCI is using for that check (continuum/anaconda:latest) and my command works as expected. What the heck? Is it possible that it's due to it's one linerish nature?

BTW: I will change the image to miniconda because it's smaller and we do not use anything from full anaconda.

heavelock:
I surrender. I have no idea why this damn thing does not work as it should. The commit is ok, the script is ok, everything runs but CircleCI decides not to accept it.

megies:
Hmm.. the CircleCI SHA1 you're echoing seems fine though? Maybe that `master` identifier in the script is set to a weird value in CircleCI.. they might check out the PR tip as "master", "origin/master" to use the remote should work though (or whatever they use as remote alias)

.circleci/file_size_check.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+EXIT_CODE=0
+for SIZE in $(git log --pretty='format:%H' master.. | git cat-file --batch-check='%(objectsize)')
```
megies: Regardless of the other problems this is not working I think, it seems to reflect the size of the commit (like message size?) but not the contained objects (like binary files commited). My oneliner was just food for thought, it might need digging further into the commit/object structure of git like checking the children of the commit object or something..

Something like this maybe https://askubuntu.com/questions/618186/getting-commit-sizes-in-git

megies:
Merging, thanks @heavelock. Maybe/hopefully we can add a file size check that also checks binary changes in all of PR's commit history at some later point.. :)

Issue #2479, opened 2019-10-10
UTCDateTime, np.datetime64, and np.timedelta64


d-chambers:
It seems `np.datetime64` and `np.timedelta64` are becoming very widely-used. However, `UTCDateTime` currently doesn't interact with these objects as one would expect:

```python
import numpy as np
from obspy import UTCDateTime as UTC

# init numpy time representations
npdt = np.array(['2000-01-01', '2012-01-01T01:10'], dtype='datetime64[ns]')
offset = np.timedelta64(3, 'M')

# init obspy time representation
utc = UTC('2010-01-01')

# gt or lt
utc < npdt  # simply returns False
utc > npdt  # also returns False
npdt < utc  # returns an array of False 
npdt > utc  # return an array of True

# adding/subtracting numpy time delta 
utc + offset  # TypeError raised
offset + utc  # Also raises TypeError
``` 
So, I guess we have four options:

1. We ignore numpy's time representation and continue doing our own thing with `UTCDateTime`.

2. We try to make `UTCDateTime` recognize numpy's time formats and play nicely with them.

3. We *adopt* numpy's time formats as the core of `UTCDateTime` (maybe subclassing  `datetime64`).

4. We *outright replace* `UTCDateTime` with numpy's data structures.

Option 1 doesn't seem wise as we would effectively be alienating ourselves from any future numpy improvements, as well as reinforcing conceptual (ie API) barriers between numpy and obspy.

At first glance Option 2 appears to be the least disruptive, but it may be harder to implement than one might think. For example, if we step through the following code:
```python
npdt < utc
```
numpy's broadcasting machinery ends up calling `UTCDateTime.__gt__` for each element of the datetime64 array with the `other` argument as an `int`. The meaning of this `int` depends on the units specified upon initialization but in this case it means nanoseconds from epoch time. However, `UTCDateTime` interprets real numbers as seconds from epoch time, eg
```python
utc > 10  # True, utc is gt epoch time + 10 seconds
```
So we would either have to  guess at the units of the `other` parameter based on its size (not a good idea) or somehow get the contextual info about the units (not sure if this is possible/feasible).

I am not sure if option 3 would solve the issues mentioned with option 2. We would need to look more into it. 

Option 4 would, of course, be very disruptive and probably break lots, if not most, codes. 

I think @markcwill suggested option 3/4 at one point in time but I don't remember where. Also of note is the conversation on #2072. 

Any thoughts?

markcwill:
[...]

PR #2480, opened 2019-10-11
UTCDateTime support for numpy datetime64

d-chambers:
### What does this PR do?

This PR adds some support for numpy's date time representations. It doesn't solve all the issues but does address most of them without any backwards-incompatible changes.

It also includes some small refactoring of `UTCDateTime` and adds some additional tests for the same.

We will have to bump the min numpy version to 1.7 (released in 2013) before merging this.

### Summary

```python
import numpy as np
from obspy import UTCDateTime

utc = UTCDateTime('2017-09-18T18:00:00')

# convert utc to datetime64 (uses ns as units)
dt64 = utc.datetime64

# can now add np.timedelta64 to utc
utc + np.timedelta64(100, 's')

# comparisons work if UTCDateTime is first
utc == dt64  # True
utc - 1 < dt64  # True
utc + 1 > dt64  # True

# but they wont work at all if datetime64 is first. This is because the numpy ops ends up 
# simply passing an int (with no indication of units) to `UTCDateTime`s dunder methods
# which gets interpreted as a timestamp in seconds. I dont see a way to address this issue.
dt64 == utc  # False 
...
```

Issues mentioned in #2479 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_utcdatetime.py
+++ b/obspy/core/tests/test_utcdatetime.py
@@ -300,6 +300,19 @@ class UTCDateTimeTestCase(unittest.TestCase):
             a + -0.5, UTCDateTime(1969, 12, 31, 23, 59, 59, 500000))
         td = datetime.timedelta(seconds=1)
         self.assertEqual(a + td, UTCDateTime(1970, 1, 1, 0, 0, 1))
+        # Test that ns precision is maintained (time delta has us precision).
+        start = UTCDateTime(ns=1111111111)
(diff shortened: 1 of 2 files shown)


heavelock:
I rebased on master and forced pushed, I hope you don't mind.

d-chambers:
@heavelock not at all. 

I suppose this is almost done. It's not perfect but it at least adds some level of support for numpy datetimes. I will try to schedule some time in the next week or so to finish it.

PR #2482, opened 2019-10-16
Refactor stack

trichter:
I refactored Stream.stack a bit and moved the grouping to ``Stream._group_by()`` and the data stacking to ``obspy.signal.utils.stack()`` so that it can be used with a simple numpy array as well.

Follow-up to #2440.

### DIFF
--- a/obspy/signal/util.py
+++ b/obspy/signal/util.py
@@ -304,9 +304,10 @@ def stack(data, stack_type='linear'):
     :type stack_type: str or tuple
     :param stack_type: Type of stack, one of the following:
         ``'linear'``: average stack (default),
-        ``('pw', order)``: phase weighted stack of given order,
-        see [Schimmel1997]_,
-        ``('root', order)``: root stack of given order.
+        ``('pw', order)``: phase weighted stack of given order
+        (see [Schimmel1997]_, order 0 corresponds to linear stack),
+        ``('root', order)``: root stack of given order
+        (order 1 corresponds to linear stack).
     """
     if stack_type == 'linear':
         stack = np.mean(data, axis=0)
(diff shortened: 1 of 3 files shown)


megies:
I'll leave this to @krischer as I also didnt review the original PR

obspy/signal/util.py
```diff
@@ -297,6 +297,38 @@ def _good_factorization(x):
...
+        ``('pw', order)``: phase weighted stack of given order
+        (see [Schimmel1997]_, order 0 corresponds to linear stack),
+        ``('root', order)``: root stack of given order
+        (order 1 corresponds to linear stack).
+    """
+    if stack_type == 'linear':
+        stack = np.mean(data, axis=0)
+    elif stack_type[0] == 'pw':
+        from scipy.signal import hilbert
+        from scipy.fftpack import next_fast_len
```
megies: `.. versionadded:: 0.18.0` (https://github.com/scipy/scipy/blob/cf4a407189cbc5e214dbf9ddc56f7eb12aae8b81/scipy/fftpack/helper.py#L52)

@trichter 

Needs workarounds for scipy < 0.18 (looks like non-trivial to backport, theres some cython or C involved from what i see). Alternative would be to bump minimum dependency versions but it'll still disrupt functionality on some LTS distros in that case, I guess.

see e.g. https://travis-ci.org/obspy/obspy/jobs/618123342#L2106 from https://github.com/obspy/obspy/pull/2507

trichter: Thanks for the hint. If minimum versions are not bumped, it would be most simple to use nextpow2 for scipy<0.18.

trichter: Done.

megies: Yep sounds good :+1:

PR #2485, opened 2019-10-21
Fix reading rg16 file when start+endtime are inside one data packet

thefroid:
### What does this PR do?
The aim of this PR is to be able to retrieve data when the parameter starttime and endtime are comprised in a data blocket. I also add an unitary test to take into account this case.

I noticed that the function_read_rg16 does not return anything when the user make a request inside a data blocket. For example if my data blocket starts at 16h and end at 17h and if i request data between 16h15 and 16h30, the stream object returned will be empty.

### PR Checklist
- [ ] All tests still pass.


obspy/io/rg16/tests/test_read_rg16.py
```diff
@@ -72,10 +72,12 @@ def test_headonly_option(self):
 
     def test_starttime_endtime_option(self):
         """
-        Test the the options starttime and endtime
+        Test the options starttime and endtime
         """
         t1 = UTCDateTime(2017, 8, 9, 16, 0, 15)
         t2 = UTCDateTime(2017, 8, 9, 16, 0, 45)
+        t3 = UTCDateTime(2017, 8, 9, 16, 0, 1)
+        t4 = UTCDateTime(2017, 8, 9, 16, 0, 22)
```
megies: So these variables basically are sorted like `t3 < t1 < t4 < t2`. That's incredibly confusing, to me at least, please change

obspy/io/rg16/tests/test_read_rg16.py
```diff
@@ -87,6 +89,14 @@ def test_starttime_endtime_option(self):
         for tr, tr1 in zip(st, st1):
             self.assertEqual(tr.stats.starttime, tr1.stats.starttime)
             self.assertEqual(tr.stats.endtime, tr1.stats.endtime)
+        
+        # test when starttime and endtime are comprised in a data blocket.
+        st2 = rc._read_rg16(THREE_CHAN_FCNT, starttime=t3, endtime=t4)
```
megies: If I understand the aim of the PR correctly you want the inside timestamps here, so by your current naming `starttime=t1, endtime=t4`. Did you get confused by your own variable naming? ;-P

megies:
Will need a changelog eventually

obspy/io/rg16/tests/test_read_rg16.py
megies: Also please rename in comment from `blocket` to maybe `packet` or `block`. "Blocket" sounds too much like MiniSEED blockettes.

thefroid: The file three_chans_six_trace.fcnt contains the data of three channels. Each channel owns 2 data block. The first data block starts at 16:00:00.380000 and ends at 16:00:30.378000, the second data block starts at 16:00:30.380000 and ends at 16:01:00.378000. If a user wants to retrieve data between 16:00:45 and 16:00:55, the function _read_rg16 will return an empty stream. That's why i added an if condition in core.py to take into account this case.

obspy/io/rg16/core.py
```diff
@@ -70,6 +70,11 @@ def _read_rg16(filename, headonly=False, starttime=None, endtime=None,
             trace = _make_trace(filename, trace_block_start, headonly,
                                 contacts_north, details)
             traces.append(trace)
+        if trace_starttime <= starttime.timestamp < trace_endtime and\
+           trace_starttime < endtime.timestamp <= trace_endtime:
+            trace = _make_trace(filename, trace_block_start, headonly,
+                                contacts_north, details)
+            traces.append(trace)
```
d-chambers: Maybe it would be more clear to just check if `starttime` or `endtime` exclude the current block, else include it? I am thinking something like this:

```python
    for i in range(0, record_count):
        nbr_bytes_trace_block = _cmp_jump(filename, trace_block_start)
        trace_starttime = _read(filename, trace_block_start + 20 + 2 * 32, 8,
                                'binary') / 1e6
        trace_endtime = trace_starttime + record_length
        # determine if this block is excluded based on start/endtime params
        con1 = starttime.timestamp > trace_endtime
        con2 = endtime.timestamp < trace_starttime
        if con1 or con2:  # if so increment trace_block_start and continue
            trace_block_start += nbr_bytes_trace_block
            continue
        # else collect trace 
        trace = _make_trace(filename, trace_block_start, headonly,
                            contacts_north, details)
        traces.append(trace)
        trace_block_start += nbr_bytes_trace_block
```

thefroid: Thank you for your advices Chambers, i took into account them in the new commit.

obspy/io/rg16/tests/test_read_rg16.py
megies: I understand the problem, I was just curious if you used the starttime and endtime as intended here, but seems OK then. You should maybe add this information about the start/endtimes of the two data packets in each test file as a comment in the test, because the intention/aim of the test is not clear from the test code as is.

obspy/io/rg16/tests/test_read_rg16.py
```diff
@@ -72,10 +72,12 @@ def test_headonly_option(self):
 
     def test_starttime_endtime_option(self):
```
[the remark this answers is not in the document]
megies: Shorter, separate test case also help during debugging test failures.

CHANGELOG.txt
```diff
@@ -106,6 +106,7 @@ master:
      16/32bit integers, see #2058 and #2059)
  - obspy.io.rg16:
    * implement module to read waveforms and headers from fcnt format (see #2265).
+   * Fix reading when the starttime and endtime options are comprised in the same data block (see #2485).
```
megies: please wrap line, it's too long

megies:
@thefroid see CircleCI log for flake 8 fails

megies:
Other test fails are unrelated it seem, see http://tests.obspy.org/?pr=2485 and compare #2486

megies:
CI looks good

PR #2484, opened 2019-10-21
feature Stats.component

trichter:
What do you think about a `component` or `orientation` field in the Stats object? It would get or set the last character of `Stats.channel`.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -209,7 +209,11 @@ class Stats(AttribDict):
             return
         if key == 'component':
             key = 'channel'
-            value = self.channel[:-1] + str(value)
+            value = str(value)
+            if len(value) != 1:
+                msg = 'Component must be set with single character'
+                raise ValueError(msg)
+            value = self.channel[:-1] + value
         # prevent a calibration factor of 0
         if key == 'calib' and value == 0:
             msg = 'Calibration factor set to 0.0!'
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -137,10 +137,10 @@ class Stats(AttribDict):
 
         >>> stats = Stats()
         >>> stats.channel = 'HHZ'
-        >>> stats.component
+        >>> stats.component  # doctest: +SKIP
         'Z'
         >>> stats.component = 'L'
-        >>> stats.channel
+        >>> stats.channel  # doctest: +SKIP
         'HHL'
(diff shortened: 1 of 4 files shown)


trichter:
I added a possible implementation. This variant dynamically gets the component from the channel. It would also be possible to add an extra field in the dictionary, but because the information provided by component is redundant to the information in the channel, I think this implementation is preferential.

obspy/core/trace.py
```diff
@@ -194,6 +207,9 @@ def __setitem__(self, key, value):
                 timediff = float(self.npts - 1) * delta
             self.__dict__['endtime'] = self.starttime + timediff
             return
+        if key == 'component':
+            key = 'channel'
+            value = self.channel[:-1] + str(value)
```
heavelock: I have doubts about that substitution. In the end, the component parameter is quite artificial and this kind magic even if totally reasonable might lead to misunderstandings.

d-chambers: I agree with @heavelock, probably best to make the `component` attribute read-only and force the user reset the whole channel code if he/she really desires to do so.

trichter: I am undecided. Channel code and component are well defined by SEED convention.
I know it is special, because setting the component actually changes the channel. Still, there are other attributes behaving magically.

In the end, making it just read-only is OK with me.

megies: If we have a getter, I think having a setter might be fine as well. However, I'm only feeling uneasy about setting with empty string or multiple chars.. that's changing the length of the channel code and those examples in the tests are pretty weird to me, going from `HHZ` to `H` with two times setting an empty channel code.

I would agree to the setter if we add a `len(..) == 1` check in it.

trichter: OK, I added this check. Do you think it is now reasonable, @heavelock, @d-chambers?

d-chambers: @trichter, yes I suppose that is a decent compromise. I would still rather be more explicit in setting the whole channel but I can see why this could be more convenient.

heavelock: Yup, I think its a reasonable solution.

megies: OK, seems we can all agree that this is a viable solution/compromise. :+1:

megies:
CI looks good, one single unrelated fail, merging

PR #2487, opened 2019-10-23
rg16: most tests were not executed

megies:
### What does this PR do?

Most `io.rg16` tests were not executed by unittest due to some refactoring changes in b984a56fb540958bcf193ed22936ec14a1db9bdd

### Why was it initiated?  Any relevant Issues?

see above

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] If the PR is making changes to documentation, docs pages can be built automatically.
      Just remove the space in the following string after the + sign: "+ DOCS"
- [x] If any network modules should be tested for the PR, add them as a comma separated list
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:"
      (you can also add "ALL" to just simply run all tests across all modules)
- [ ] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Any new or changed features have are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [x] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #2489, opened 2019-10-24
switch test runner to pytest

d-chambers:
# What does this PR do?

1. Refactors ObsPy's test runner to use pytest
2. Adds *some* of the previous `obspy-runtest` command line arguments
~~3. Adds a pytest fixture for collecting test results and submitting them to the ObsPy server~~
4. `obspy-runtests` will now simply change directory to ObsPy's base and run pytest from there
~~5. Adds an `obspy.testing` module. I will move most test utilities here in future commits.~~

This PR is still a work in progress but I thought I would share now to get early feedback.

## Why was it initiated?  Any relevant Issues?

Pytest is more expressive and better maintained than unittest. 

Also, I told @krischer I would work on this. It only took me a year to get to it to this point :smile:

### Advantages

* ObsPy can now use the great features of pytest like fixtures and parameterization
* Test discovery is much easier with pytest (eg see #2487)
* We can delete (and stop maintaining) a lot of test-runner code 
* There is a huge selection of pytest plugins which my be useful to us in the future

### Disadvantages

* The `obspy-runtests` options will change significantly (breaks backwards compatibility)
* Installing pytest will be required for running the test suite (unittest is part of the std lib)

### PR Checklist
- [ ] All tests still pass.


d-chambers:
[Here](http://tests.obspy.org/106118/#skippedtests) is results page sent to tests.obspy.org. It currently is reporting a failure as an Error. I also need to look into how to run the doctests and make the post-test run output more user friendly. There are also some flake8 things to fix.

ThomasLecocq:
@d-chambers could you rebase this on master?

ThomasLecocq:
might need a little force-push :-)

d-chambers:
Ok, I made quite a bit of progress on this.

Here is what it does so far:

1) Completely replaces ObsPy's test runner in favor of using pytest. Although `obspy-runtest` still works and just calls pytest under the hood.

2) Simplifies the test options to just these (from conftest::pytest_addoption):
```python
    parser.addoption('--network', action='store_true', default=False,
                     help='test only network modules', )
    parser.addoption('--all', action='store_true', default=False,
                     help='run both network and non-network tests', )
    parser.addoption('--coverage', action='store_true', default=False,
                     help='Report Obspy Coverage to terminal and generate '
                          'xml report which will be saved as coverage.xml', )
    parser.addoption('--report', action='store_true', default=False,
                     help='Generate and HTML report of test results '
                          'saved as obspy_report.html in obspy directory.', )
    parser.addoption('--keep-images', action='store_true',
                     help='store images created while runing test suite '
                          'in a directory called obspy_test_images.')
```

Of course, any other pytest or pytest-plug in command line args can be used too.

3) Removes all image comparisons in favor of just saving each image in a directory named after the test that created it. 

The entire image directory is then uploaded as an artifact which will expire after one day. There has been talk of disabling the image comparisons before so I just went ahead and did it, but with some more refactoring we could probably switch to pytest-mpl as @QuLogic suggested. It would just be quite a bit of work and it may not be worth it.

4) Switched a significant number of test files from unittest style to pytest style. 

There is something funky going on with the git history but I will look at that tomorrow.

ThomasLecocq:
@d-chambers the PR checks don't check the network tests AFAIK, do they? 

If not, I was thinking of a rule to add to the actions, if a PR contains "+NETWORKTEST" then they could run ?

trichter:
Great work!

I think the obspy-runtests wrapper is important, because as far as I understand one has to specify the directory when using pytest. Obspy can be installed inside conda environments etc.

Is it possible to preserve the ability to upload test reports to tests.obspy.org?

The `-all` flag is not working for me:

```
without flag: 1944 passed, 30 skipped, 213 deselected, 1095 warnings
with --all: 1944 passed, 243 skipped, 1095 warnings
```

On master I have 1973 tests + 16 skipped vs 2183 tests + 19 skipped with `--all` flag (test reports online)

Coverage upload is not yet working in github actions.

barsch:
Any changes needed to [obspy/reporter](https://github.com/obspy/reporter)? Also the transmitted XML should include a indicator that pytest is used instead of the old method. I'm also willing to switch to a JSON API if needed/wanted.

trichter:
Thanks @barsch for your input.

It appears to be relatively easy to get a JunitXML with

```
pytest obspy/io/sh --junitxml ./report.xml
```

which produces the following file:

[junit_report.xml](https://github.com/obspy/obspy/files/7829649/junit_report.txt)


That's quiet different from the layout used at the moment:

[custom_report.xml](https://github.com/obspy/obspy/files/7829653/custom_report.txt)

Information about obspy version and versions of other installed packages needs to be added as well as system information. This should be easy to do.

No clue how to get the information about the slowest tests into the xml. Other infos are also missing, e.g. number of tests run/failed/skiped in the individual modules. Or could this information be derived in the reporter app?

d-chambers:
>  The -all flag is not working for me:

Right, it looks like it is doing something, just not the right thing. I will look into it. 

> Is it possible to preserve the ability to upload test reports to tests.obspy.org?

Not as it is now. My thought was to just have someone with generate the html report and attach it to an issue. I had tried to make the xml work but in the end it was just no feasible.

>  I'm also willing to switch to a JSON API if needed/wanted.

@barsch, If you are willing to do that we can use [pytest-json-report](https://github.com/numirias/pytest-json-report) to easily generate a json to upload. It supports modifying the json output at the test level as well as metadata, so we should be able to add whatever we want to it. 

> If not, I was thinking of a rule to add to the actions, if a PR contains "+NETWORKTEST" then they could run ?

I was thinking of adding a separate job which would only run the network tests on Ubuntu and python 3.8, as running them in all the matrix combinations seems excessive. We could just always run this job or only when "+NETWORKTEST" is used.

trichter:
>  It supports modifying the json output at the test level as well as metadata, so we should be able to add whatever we want to it.

Same can be achieved with xml export via `--junitxml`. Still, a lot of work in the reporter app will be necessary.

trichter:
I played a bit with `pytest-metadata` and got some properties into the xml using the following code in conftest.py:

```py
# not working for me
#pytestmark = pytest.mark.usefixtures('include_metadata_in_junit_xml')

@pytest.fixture(scope="session", autouse=True)
def include_metadata_in_junit_xml(metadata, pytestconfig, record_testsuite_property):
    """Provide test session metadata"""
    metadata_ = pytestconfig._metadata
    for name, value in metadata_.items():
        record_testsuite_property(name, value)

@pytest.mark.optionalhook
def pytest_metadata(metadata):
    metadata['Dependencies'] = {}
    for module in DEPENDENCIES:  # from scripts.runtests
        temp = module.split('.')
        try:
            mod = __import__(module,
                             fromlist=temp[1:])
        except ImportError:
            version_ = '---'
        else:
            try:
                version_ = mod.__version__
            except AttributeError:
                version_ = '???'
        metadata['Dependencies'][module] = version_
    # get system / environment settings
    import platform
    metadata['Platform2'] = {}
    for func in ['system', 'release', 'version', 'machine',
                 'processor', 'python_version', 'python_implementation',
                 'python_compiler', 'architecture']:
        try:
            temp = getattr(platform, func)()
            if isinstance(temp, tuple):
                temp = temp[0]
            metadata['Platform2'][func] = temp
        except Exception:
            metadata['Platform2'][func] = ''
```

Additional metadata can be specified via the command line, i.e. `--metadata pr-url xxx  --metadata ci-url xxx```.

The metadata is written to property tags:
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
    <testsuite name="pytest" errors="0" failures="1" skipped="2" tests="26" time="0.934" timestamp="2022-01-07T18:50:37.781652" hostname="kiste">
        <properties>
            <property name="Python" value="3.8.6" />
            <property name="Platform" value="Linux-4.15.0-166-generic-x86_64-with-glibc2.10" />
            <property name="Packages" value="{'pytest': '6.2.4', 'py': '1.10.0', 'pluggy': '0.13.1'}" />
            <property name="Plugins" value="{'html': '3.1.1', 'cov': '2.10.1', 'json-report': '1.4.1', 'metadata': '1.11.0'}" />
            <property name="JAVA_HOME" value="/usr/lib/jvm/java-8-oracle" />
            <property name="Dependencies" value="{'numpy': '1.19.5', 'scipy': '1.5.3', 'matplotlib': '3.4.2', 'lxml.etree': '4.6.2', 'setuptools': '49.6.0.post20210108', 'sqlalchemy': '1.3.22', 'decorator': '4.4.2', 'requests': '2.25.1', 'flake8': '3.8.4', 'pyimgur': '0.6.0', 'pyproj': '3.1.0', 'pep8-naming': '---', 'm2crypto': '---', 'shapefile': '2.1.2', 'mpl_toolkits.basemap': '---', 'mock': '---', 'pyflakes': '2.2.0', 'geographiclib': '1.50', 'cartopy': '0.19.0.post1'}" />
            <property name="Platform2" value="{'system': 'Linux', 'release': '4.15.0-166-generic', 'version': '#174-Ubuntu SMP Wed Dec 8 19:07:44 UTC 2021', 'machine': 'x86_64', 'processor': 'x86_64', 'python_version': '3.8.6', 'python_implementation': 'CPython', 'python_compiler': 'GCC 9.3.0', 'architecture': '64bit'}" />
        </properties>
        <testcase classname="obspy.io.sh.__init__" name="obspy.io.sh" time="0.290" />
```

d-chambers:
Nice @trichter,

I fixed the --all flag as well as the image/coverage uploads.

barsch:
I don't mind JSON or XML - whatever you guys come up with - but it should provide at least similar information as the old reports so no big frontend changes are needed. Easiest for me would be if you somehow finalize the format and attach me some example reports to a new ticket at https://github.com/obspy/reporter so I can get working on it.

>Other infos are also missing, e.g. number of tests run/failed/skiped in the individual modules. Or could this information be derived in the reporter app?

that can be derived within the reporter app ofc

trichter:
> If not, I was thinking of a rule to add to the actions, if a PR contains "+NETWORKTEST" then they could run ?

May be "+ALLTESTS" will do.

ThomasLecocq:
For the keyword-filter on GitHub actions check this: https://github.com/ROBelgium/MSNoise/blob/master/.github/workflows/build_doc.yml how I did for msnoise, we can adapt rules quite easily

trichter:
That is way easier than via ObsPy's own github API (obshub).

ThomasLecocq:
Quite sure it's doable on PR comments too, I'll investigate tomorrow

trichter:
https://github.com/marketplace/actions/find-comment

d-chambers:
Good find @trichter, it would make our CI so much simpler to just use someone else's action.

d-chambers:
> I don't mind JSON or XML

@barsch I opened the report about supporting json. It looks to me like most of the same information is in the default json report, but I added few things I noticed were missing at the top level. If you don't mind taking a quick look, (@trichter  too if you have time).

trichter:
I can add to this next week. I have some unpushed work.

d-chambers:
Before merging we need to
- [ ] Release 1.3 (since this breaks API it should probably go in 2.0, I guess)
- [ ] Make sure a nice error message is issued if a user tries to run the tests without test requirements installed
- [ ] Make sure all docs are updated about testing
- [ ] Delete old images from source
- [ ] Change the --report flag to generate, zip, and automatically upload the json report summary to obspy report server
- [ ] Remove pytest-html as a test dependency

trichter:
When running `obspy.runtests clients/fdsn` not all FDSN tests are running, correct? Can we do something about it? One option would be to automatically add the `-all` flag in this case. Maybe there is a more elegant solution using pytest magic?

In which way can we run the tutorial doc tests? Previously we had the --tutorial flag. If you do not want to introduce new options, we can also run tutorials within `--all`.

> Before merging we need to Release 1.3 (since this breaks API it should probably go in 2.0, I guess)

I tend to disagree. 99% of users just run obspy-runtests without any parameters. We can also put some work into accepting the other 0.9% of old runtests calls for 1.3.

d-chambers:
> In which way can we run the tutorial doc tests?

I think we will just have a pytest command for this that will run on the CI and which we can list in the development docs. Since the tutorial isn't distributed with the source I don't think we need to include it in `obspy-runtests`. 

> I tend to disagree. 99% of users just run obspy-runtests without any parameters. We can also put some work into accepting the other 0.9% of old runtests calls for 1.3.

Ok, I am good with it either way. However, rather than spending time/effort trying to make the new `obspy-runtests` more compatible with the old one (which is what I think you mean by accepting old runtest calls?) I would rather spend time migrating more test cases to the pytest style.

> I also think we should declare pytest and pytest-json-reporter as hard dependencies. pytest-cov can stay as soft dep imo.

Sounds good to me.

>When running obspy.runtests clients/fdsn not all FDSN tests are running, correct? Can we do something about it? One option would be to automatically add the -all flag in this case. 

We could, but I would rather not do it. If you want to implement it though I don't mind it being there.  If we are re-writing the test input anyway we might also consider supporting both `io/sac` and `io.sac` styles. Again, though, I personally don't think it is worth it.

d-chambers:
A comment from @trichter here: https://github.com/obspy/reporter/issues/22

>We should also report versions of non-required dependencies (i.e. cartopy and co) and report if these are installed. Different sets of installed packages often lead to unexpected failures in the past.

Agree on this, I am just not sure the best way to do it.  The problem is finding all of ObsPy's dependencies in the test report generating code. The setup.py has the following lines which import functions from ObsPy at install time: 

https://github.com/obspy/obspy/blob/dab888c72c3644172ce767a636ed83ac9c6a397f/setup.py#L71-L78

Maybe we move INSTALL_REQUIRES and EXTRAS_REQUIRE into one of those modules, or a new one in a similar location, so we can import the requirements under an obspy namespace and still have the requirements listed in one place?

trichter:
test parsing comments +TEST:--all|clients/fdsn

ThomasLecocq:
Idea : is it possible to detect the tags of a PR? If yes, then if tagged Fdsn, it'd auto test the network stuff?

trichter:
I added a `--no-report` flag and reintroduced the previous default behavior of asking the user if test report should be uploaded. For failing tests this is less work for users, than restarting tests with --report flag.

I used the possiblity to get the data inside python (see pytest-json-report help), but unfortunately this ugly error message shows up:

```
PytestAssertRewriteWarning: Module already imported so cannot be rewritten: pytest_jsonreport
```

I dont know what to do about it. Maybe revert to using a temp file.

I also added comment parsing code to github actions and edited the PR template.
Remove the white space after +:  + TEST:--all to trigger all tests. Other parameters can also be passed (`|` is converted to white space + TEST:io/sh|clients/fdsn|--all
Feel free to change the syntax for the better.

> Idea : is it possible to detect the tags of a PR? If yes, then if tagged Fdsn, it'd auto test the network stuff?

Should be doable.

> > In which way can we run the tutorial doc tests?

> I think we will just have a pytest command for this that will run on the CI and which we can list in the development docs. Since the tutorial isn't distributed with the source I don't think we need to include it in obspy-runtests.

Sounds OK, too. A tutorial flag would have the benefit, that it is easy to use and we can report the results to the test server.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +24,9 @@ https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model
 - [ ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
 - [ ] This PR is not directly related to an existing issue (which has no PR yet).
 - [ ] If the PR is making changes to documentation, docs pages can be built automatically.
-      Just remove the space in the following string after the + sign: "+ DOCS"
-- [ ] If any network modules should be tested for the PR, add them as a comma separated list
-      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:"
-      (you can also add "ALL" to just simply run all tests across all modules)
+      Just remove the space in the following string after the + sign: + DOCS
+- [ ] If all tests inlcuding network modules (e.g. `clients.fdsn`) should be tested for the PR,
```
ThomasLecocq, line 28: inlcuding/including

d-chambers:
> The --network flag is not absolutely necessary, because it is basically obspy-runtests clients --all

Almost, the filesystem clients still run without network tests selected, and selecting only --network should skip them. Probably close enough if you want to remove this flag we can. 

> Idea : is it possible to detect the tags of a PR? If yes, then if tagged Fdsn, it'd auto test the network stuff?

This seems a lot cleaner to me. Rather than having to scan through the comments of an often long PR (eg this one) one can just use the labels to see which tests will run and if the docs should build. 

This SO answer has an example in the comments :https://stackoverflow.com/a/62331521

```
if: contains(github.event.pull_request.labels.*.name, '<label_name>')
```

Maybe we just add a `network` label and do something like this?

```
      - name: run network tests
        if: contains(github.event.pull_request.labels.*.name, 'network')
        shell: bash -l {0}
        run: |
          obspy-runtests --coverage --report --network
```
This also has the advantage of nicely splitting out the normal failures from the network failures on the github CI dashboard. We could do something similar for building docs, then remove all the quirky obspy-github-api business.

d-chambers:
I did a few more things here:

 1) I added an obspy.core.utils.requirements.py module for keeping ObsPy's reqs all in on place accessible via setup.py and from the ObsPy namespace. 

2) I added a nice error message when someone tries to run `obspy-runtests` without the test deps installed:

```
ImportError: 
Not all ObsPy's test requirements are installed. You need to install them before using obspy-runtest. Example with pip: 
        $ pip install pytest pytest-cov pytest-json-report
```
I think this is a better option than adding pytest as a hard dependency since ObsPy runs on some small devices (which may never run the test suite), so less bloat is better. 

3) I cleaned up some more of the old test runner. 

4) I switched the way we control how the CI determines if it should run network test . Now there is an optional job, which is run based on presence a new tag called 'test_network'. If you don't like this behavior we can revert it but it seems much cleaner than comment parsing (and separates the network tests from normal tests on output too!)

I think I updated all the documentation related to testing but I certainly could have missed something there. Barring test failures, I am happy with the PR, anyone want to see anything else added (@trichter, @barsch, @ThomasLecocq ) ? I probably won't have too much more time to work on this as classes start tomorrow.

obspy/scripts/runtests.py
```diff
@@ -3,774 +3,176 @@
...
+(3) Run tests of module :mod:`obspy.io.mseed`::
+        $ obspy-runtests io/mseed
+(4) Run tests of multiple modules, e.g. :mod:`obspy.io.wav` and
     :mod:`obspy.io.sac`::
+        $ obspy-runtests io/wav obspy/io/sac
+(5) Run a specific test case::
+        $ obspy-runtests core/tests/test_stats.py::TestStats::test_init
+(6) Run tests and print a coverage report to screen and save coverage.xml
+    with pytest-cov plugin:
+        $ obspy-runtests --coverage
```
trichter, line 37: --cov or --coverage?

obspy/scripts/runtests.py
```diff
@@ -3,774 +3,176 @@
...
+            print('Your test results have been reported and are available at: '
+                  '{}\nThank you!'.format(report_url))
+        # handle errors
+        else:
+            print(f"Error: Could not sent a test report to {REPORT_URL}.")
+            print(response.reason)
 
 
-def main(argv=None, interactive=True):
+def run_tests(network=False,
```
trichter, line 142: Can this function be removed?

barsch:
btw please be careful about adding all environment variables into the JSON report - I found e.g. ```
JAVA_HOME = "/Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/8.0.312-7/x64/Contents/Home/"``` in https://tests.obspy.org/116094/json/

There could be much more sensitive variables set such as account credentials etc. - I strongly suggest to filter for certain variables and expose only those.

obspy/scripts/runtests.py
d-chambers: Ya, I guess we aren't really supporting running the tests from python anymore; just pytest or obspy-runtests from the command line.

obspy/scripts/runtests.py
d-chambers: Ya, I guess we aren't really supporting running the tests from python anymore; just pytest or obspy-runtests from the command line.

obspy/conftest.py
```diff
@@ -0,0 +1,253 @@
...
+        setattr(config.option, 'markexpr', 'network')
+    elif not network_selected and not all_selected:
+        setattr(config.option, 'markexpr', 'not network')
+
+    # select appropriate options for report
+    config.option.json_report_indent = 2
+
+    # select options for coverage
+    # this is the same as using:
+    # --cov obspy --cov-report term-missing --cov-report='xml' --cov-append
```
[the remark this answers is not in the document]
d-chambers: Sure, I will also add an example in the doc string.

d-chambers:
> btw please be careful about adding all environment variables into the JSON report 

It looks like that was added by pytest-json-report in the metadata section. I will simply filter out anything that looks like an environmental variables as we don't need them.

trichter (approved):
`JAVA_HOME` is still in the JSON.

Otherwise IMO good to go.

d-chambers:
> JAVA_HOME is still in the JSON.

Ok let me fix that quick.

d-chambers:
Also, I am going to make uploading the test images contingent on a new tag (as we did for network tests) so we don't exhaust all of our github storage if we have a flurry of CI runs.

obspy/scripts/runtests.py
```diff
@@ -10,7 +10,7 @@
     --help : Print help message an exit.
     --no-report, --report : Do not ask and automatically (not) submit report
             to ObsPys' report server.
-    --cov : Calculate and display test coverage.
+    --cov obspy: Calculate and display test coverage.
```
trichter, line 13: Actually just using `obspy-runtests --cov` is working fine for me (but does not display missing files).

d-chambers: I think `--cov` will just report the coverage on files that were imported. `--cov obspy` will report all files in ObsPy, even those not imported, as 0% coverage.  You can see the difference if you run 

```bash
obspy-runtest io/sac --cov
```

vs

```bash
obspy-runtests io/sac --cov obspy
```

d-chambers:
There is another small issue, when installed not in development mode the network tests run even without the --network or --all tests. Looking into it now.

d-chambers:
Ok, I am happy with this once the CI is green.

ThomasLecocq:
> Ok, I am happy with this once the CI is green.

it seems the network tests are not so happy: https://tests.obspy.org/?pr=2489 ?

d-chambers:
> it seems the network tests are not so happy: https://tests.obspy.org/?pr=2489 ?

I spent some time digging into this. It seems many of these tests have been neglected for quite some time. It's possible some of the doctest failures might be related to changes in this PR but the failures in the test files are not. 

I made some fixes but there is more work to do (I will need help from someone who is more knowledgeable about FDSN than me). I think we should merge this then give the network tests some love in a different PR.

d-chambers:
Ok, it seems some of the network changes I made caused the EW tests to pass on my machine, but fail on the CI. I added a quick fix so this PR doesn't leave the network tests worse off (from the CI's perspective). 

I also figured a `no_ci` tag would be nice if you don't want to trigger the CI (eg for a work in progress) just to be nice to Microsoft's servers. 

Let's see what the CI does but I think we are ready to merge.

trichter:
I added a link to tests.obspy.org as a new status in the checks tab. Is that OK?

One last comment: I am unsure if we want to use the --network flag or --all flag. Related: the continue on error tag hides all nework fails in the checks tab. Do we really want this?

trichter:
Now previous runs are canceled by the concurrency flag. This is better than checking after starting the runner. I have seen the following warning multiple times.

```
Can't find any online and idle self-hosted or hosted runner in the current repository, account/organization that matches the required labels: 'macos-latest'
Waiting for a self-hosted or a hosted runner to pickup this job...
```

trichter:
I am merging this now. We have to keep in mind to reintroduce the image comparison tests at some point. Also, the tutorial is not tested at the moment.

ThomasLecocq:
image

congratulations @trichter and @d-chambers !!!! Amazing work here !

megies:
@d-chambers @trichter 
Ever seen an error like that in new pytest setup? Probably something we should report upstream to `pytest-json-report`?

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=469723&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0&l=5995

```
signal/tests/test_tf_misfit.py ......                                    [ 95%]
signal/tests/test_trace.py ...                                           [ 95%]
signal/tests/test_trigger.py ........F
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/_pytest/main.py", line 268, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/_pytest/main.py", line 322, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
[......]
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/pytest_jsonreport/plugin.py", line 192, in pytest_json_runtest_stage
INTERNALERROR>     return serialize.make_teststage(
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/pytest_jsonreport/serialize.py", line 58, in make_teststage
INTERNALERROR>     stage['traceback'] = make_traceback(traceback)
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/pytest_jsonreport/serialize.py", line 87, in make_traceback
INTERNALERROR>     return [{
INTERNALERROR>   File "$PREFIX/lib/python3.8/site-packages/pytest_jsonreport/serialize.py", line 88, in <listcomp>
INTERNALERROR>     'path': entry.reprfileloc.path,
INTERNALERROR> AttributeError: 'ReprEntryNative' object has no attribute 'reprfileloc'
```

PR #2492, opened 2019-10-29
Update taper max_percentage

zackspica:
if not set to None, will not respond to other param like max_length.

### What does this PR do?

*Please fill in*

*Please fill in*

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -2002,7 +2002,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
 
     @skip_if_no_data
     @_add_processing_info
-    def taper(self, max_percentage, type='hann', max_length=None,
+    def taper(self, max_percentage=None, type='hann', max_length=None,
               side='both', **kwargs):
         """
         Taper the trace.


megies:
I do not understand the motivation for this change, can you elaborate please?

zackspica:
When you apply the taper with only arg max_length=xx you have the following
error:

<ipython-input-3-69767a402736> in <module>

----> 1 tr.taper(max_length=3)


TypeError: taper() missing 1 required positional argument: 'max_percentage'

So one has to add both arguments (max_percentage=None, max_length=xx) which
I found confusing but I realize there is probably a reason for it...


Ƶ.

megies:
Yes, because `max_percentage` is required. This is partly due to historic reasons, and requiring the percentage option over an absolute length option as mandatory is a matter of taste, I guess. With this proposed change, using `.taper()` with default arguments and no customization would basically do nothing to the data. I don't think that's a better solution compared to forcing input by users.

Also this is a non-problem, just call it with `taper(max_percentage=None, max_length=...)`

PR #2469, opened 2019-10-30
Check Python 3.8 compatibility

megies:
Python 3.8.0rc1 is out, we should check if there are any issues with our master branch before releasing 1.2.0.

Hopefull there will be packages on conda-forge, then we could simply add it to the CI build matrix. See conda-forge/python-feedstock#270

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ matrix:
       env: PYTHON_VERSION=3.6
     - os: linux
       env: PYTHON_VERSION=3.7
+    - os: linux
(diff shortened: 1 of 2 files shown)


megies:
This likely won't run through yet, but at least Python 3.8 packages are out in Anaconda default channel now, so more of a reminder

heavelock:
I restarted travis to see if the conflicts in dependencies got resolved.

heavelock:
Dependencies were successfully solved but the GCF reading does not work.

PR #2499, opened 2019-11-12
Adds type checks and float casting in distaz fixes #2497

### LINKED ISSUE #2497: distaz method of Iris Client returns lxml.objectify.FloatElement
I just noticed that distaz method of Iris client is returning objectify floats instead of pure floats. Only value of `ellipsoidname` is returned as native str. I see that there was similar bug for seishub client which was fixed. In my opinion, those values should be casted to native types. Should I change it?

```python
from obspy.clients.iris import Client
distaz = Client().distaz(0, 0, 1, 1)
type(distaz['ellipsoidname'])
>> str
type(distaz['distance'])
>> lxml.objectify.FloatElement
```

heavelock:
Casts non-native floats to native ones. Adds type tests to that method and the next one.
fixes #2497

### DIFF
--- a/obspy/clients/iris/tests/test_client.py
+++ b/obspy/clients/iris/tests/test_client.py
@@ -166,7 +166,7 @@ class ClientTestCase(unittest.TestCase):
                             filename=tempfile)
             with open(tempfile, 'rt') as fp:
                 self.assertEqual(fp.readline(),
-                                 '1.000000E-05  1.055999E+04  1.792007E+02\n')
+                                 '1.000000E-05 1.055999E+04 1.792007E+02\n')
         # cs as ASCII file
         with NamedTemporaryFile() as tf:
             tempfile = tf.name
@@ -175,7 +175,7 @@ class ClientTestCase(unittest.TestCase):
                             filename=tempfile)
             with open(tempfile, 'rt') as fp:
                 self.assertEqual(fp.readline(),
-                                 '1.000000E-05 -1.055896E+04 1.473054E+02\n')
+                                 '1.000000E-05  -1.055896E+04  1.473054E+02\n')
         # fap & def as ASCII file
         with NamedTemporaryFile() as tf:
             tempfile = tf.name
@@ -184,7 +184,7 @@ class ClientTestCase(unittest.TestCase):
                             filename=tempfile)
             with open(tempfile, 'rt') as fp:
                 self.assertEqual(fp.readline(),
-                                 '1.000000E-05  1.055999E+04  1.792007E+02\n')
+                                 '1.000000E-05 1.055999E+04 1.792007E+02\n')
         # fap & dis as ASCII file
         with NamedTemporaryFile() as tf:
             tempfile = tf.name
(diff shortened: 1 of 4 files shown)


megies:
+TESTS:clients.iris

megies:
Restarted Travis so teh client tests are run.

heavelock:
My tests pass. 
The failing test thing is one of evalresp's. Only on linux, the `fap` file has one space between values instead of expected two. The writing method is from 2011/2013 by Krischer and Barsch, the evalresp fetcher is from 2011 by Barsch. Thus it seems that it's rather something system dependent, not the code itself.

megies:
Lets get some fresh CI, looked like some doctests need fixing or maybe its fixed in master, we'll see

heavelock:
So it looks like IRIS updated on their side entries that the tests suite is checking against because returned values changes. On our side the values were also changed but only in the test data file, not in doctests. But even there there is one value that changed once again. 

Should I fix those comparisons and data stored in doctests and test data file? 

The evalresp error stayed the same as before, either fetched data chenged on writing to ASCII broke.

megies:
> Should I fix data stored in doctests and test data file?

Yeah please adapt them if you don't mind (even if they might be unrelated I guess). You can use ellipsis in doctests `...` (need to also put `# doctest: +ELLIPSIS`) to make them more forgiving wherever it seems fragile.

heavelock:
I changed the expected values. In the end for the sacpz affected was only last digit of lat and lon which might suggest that there is a rounding error somewhere. 
I adjusted also evalresp. I ended up basically adding spaces between values where there was single one and removing where was a double one. Weird.

I will check what Travis will say and see if I should dig deeper.

obspy/clients/iris/tests/test_client.py
```diff
@@ -166,7 +166,7 @@ def test_evalresp(self):
                             filename=tempfile)
             with open(tempfile, 'rt') as fp:
                 self.assertEqual(fp.readline(),
-                                 '1.000000E-05  1.055999E+04  1.792007E+02\n')
+                                 '1.000000E-05 1.055999E+04 1.792007E+02\n')
```
megies: if these keep making problems you could also `.split()` and compare as a list

heavelock: It looks like it's going to pass so I would leave it as it is.

heavelock:
Rebased and force pushed, should ready to merge after CI.

obspy/clients/iris/tests/test_client.py
heavelock: It looks like it should be okay, but yeah, that would be a better approach.

heavelock:
Hmm, after that rebase, at 2.7 type assertion for `str` started to complain. The reason is that the parsed string is returned as a `str` but I'm comparing it to py2-3 compatible string `future.types.newstr.BaseNewStr`. Casting with use of `str()` fixes it. But do you have any idea why it happened now? It's weird.

obspy/clients/iris/tests/test_client.py
megies: I don't care.. if it's passes for now I'm happy

megies:
> Hmm, after that rebase, at 2.7 type assertion for `str` started to complain. The reason is that the parsed string is returned as a `str` but I'm comparing it to py2-3 compatible string `future.types.newstr.BaseNewStr`. Casting with use of `str()` fixes it. But do you have any idea why it happened now? It's weird.

We had stuff like that happen before, but I can't be bothered to look it up right now and after 1.2.0 is released we're getting rid of all that anyway.

megies:
Thanks @heavelock

PR #2505, opened 2019-11-20
Fix gcf reading at Py38

### LINKED ISSUE #2469: Check Python 3.8 compatibility
Python 3.8.0rc1 is out, we should check if there are any issues with our master branch before releasing 1.2.0.

Hopefull there will be packages on conda-forge, then we could simply add it to the CI build matrix. See conda-forge/python-feedstock#270

heavelock:
As #2469 shown, in Py38 gcf reading was failing. It turned out that `char()` do not allow anymore to pass as an argument a list which is the case here. So I changed it to pass only first and only element of the list.

I tested on 3.8 and 3.6 on my machine, let's see what Travis will say.

Based on branch ci_py38, rebased on current master.

Closes #2469

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -84,6 +84,8 @@ master:
      Python2 (see #2413)
    * fix an issue that could make small landmasses not get plotted in basemap
      plots (see #2471, #2477)
+ - obspy.io.gcf:
+   * Fixes Python 3.8 compatibility of GCF reader. (see #2505)
  - obspy.io.mseed:
    * Add ability to write int64 data to mseed if it can safely be downcast
      to int32 data, otherwise raises ObsPyMSEEDError. (see #2356)
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ matrix:
       env: PYTHON_VERSION=3.6
     - os: linux
       env: PYTHON_VERSION=3.7
+    - os: linux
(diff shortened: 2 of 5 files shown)


heavelock:
Works!

QuLogic:
If `decode36` really takes an `int`, then it seems like it's the callers that should be fixed. But even if not the case, the indexing should be done outside the loop.

heavelock:
It's not the `decode36` itself. It's just the `chr`. `data` is a `np.ndarray`, all subsequent operations return an
array with a single element as result.

But yeah, since the docstring says `Converts an integer into a base36 string.` then probably the problem is not here but in the invocation itself. I will fix it later today.

heavelock:
Okay, looks like ready for another revision. @QuLogic

CHANGELOG.txt
```diff
@@ -84,6 +84,8 @@ master:
      Python2 (see #2413)
    * fix an issue that could make small landmasses not get plotted in basemap
      plots (see #2471, #2477)
+ - obspy.io.gsf:
```
megies: typo -> gcf

megies:
I cleaned up the branch and force-pushed @heavelock should be good to merge when CI is done

heavelock:
Perfect! Thanks! I will do follow up adding compatibility with 3.8 to setup.py.

megies:
Thanks @heavelock!

PR #2466, opened 2019-11-26
QuakeML reading: Elements of "http://quakeml.org" native namespace in custom tags section ignored

megies:
When reading QuakeML files that contain custom elements which have a native `http://quakeml.org` namespace, these do not get properly deserialized into `.extra` dictionaries. Writing works, but reading back leaves them out.

```python
from obspy import Catalog, read_events

extra = {'my_tag_1': {'value': u'11111',
                      'namespace': 'http://quakeml.org/xmlns/quakeml/1.2'},
         'my_tag_2': {'value': u'22222',
                      'namespace': 'http://xxx.org/xmlns/quakeml/1.2'}}

cat = Catalog()
cat.extra = extra
cat.write('my_catalog.xml', format='QUAKEML')

cat2 = read_events('my_catalog.xml', format='QUAKEML')
print(cat2.extra)
```

xml output

```xml
<?xml version='1.0' encoding='utf-8'?>
<q:quakeml xmlns:ns0="http://xxx.org/xmlns/quakeml/1.2" xmlns:q="http://quakeml.org/xmlns/quakeml/1.2" xmlns="http://quakeml.org/xmlns/bed/1.2">
  <eventParameters publicID="smi:local/08724e77-3221-4989-ad8e-c6f0fc49d8bf">
    <ns0:my_tag_2>22222</ns0:my_tag_2>
    <q:my_tag_1>11111</q:my_tag_1>
  </eventParameters>
</q:quakeml>                                                                                                                                                                    
```

print output after running the read at end

```
AttribDict({'my_tag_2': AttribDict({'value': '22222', 'namespace': 'http://xxx.org/xmlns/quakeml/1.2'})})
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -128,6 +128,9 @@ master:
      (see #2106, #2098, #2095)
    * Catalogs with empty event description objects can be round-tripped (see
      #2339, #2340).
+   * Correctly handle QuakeML native namespaces (if they are not matching the
+     document root's namespace) as custom namespaces and parse them into
+     `.extra` (see #2466)
  - obspy.io.sac:
    * Fix bug writing inventory with SOH channels to SACPZ (see #2200).
  - obspy.io.segy:
--- a/obspy/io/quakeml/core.py
+++ b/obspy/io/quakeml/core.py
@@ -27,6 +27,7 @@ from future.builtins import *  # NOQA
 import inspect
 import io
 import os
+import re
 import warnings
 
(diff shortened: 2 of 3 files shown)


megies:
@feuchner this should fix the issue you reported, I believe. Would be nice if you could confirm.

megies:
Now also making sure in tests that `extra` is empty when it's supposed to be empty.

megies:
Single test fail is unrelated, merging

PR #2515, opened 2019-12-17
Geographic selection of inventory/network/station

claudiodsf:
### What does this PR do?

Add geographic options to `[Inventory,Network,Station].select()`, following
the same syntax of `Client.get_stations()`.

Also add documentation to "sampling_rate" parameter.

Sometimes datacenter provides preassembled datasets with data and metadata for all the stations in the network (ex.: http://cnt.rm.ingv.it/event/23558121 --> Download). This PR adds possibility of filtering metadata based on geographic parameters, following the same syntax of `Client.get_stations()`.
Not covered by this PR is the possibility of selecting data in stream based on the inventory content (next time ;-)

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -95,7 +95,7 @@ master:
  - obspy.clients.iris:
    * results of distaz method are now returned as native floats (see #2499)
  - obspy.geodetics:
-   * new utility function `inside_geobounds()` to check wether an object is
+   * new utility function `inside_geobounds()` to check whether an object is
      inside a geographic bound (see #2515)
  - obspy.imaging:
    * obspy-scan can now be used with wildcarded SEED IDs when specifying what
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -93,6 +93,9 @@ master:
      #2405)
  - obspy.clients.iris:
    * results of distaz method are now returned as native floats (see #2499)
+ - obspy.geodetics:
+   * new utility function `inside_geobounds()` to check wether an object is
+     inside a geographic bound (see #2515)
  - obspy.imaging:
    * obspy-scan can now be used with wildcarded SEED IDs when specifying what
      to plot after scanning data (see #2227)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -61,6 +61,7 @@ master:
    * Fix a bug that was causing an exception being raised in `Response.plot()`
      because a float was being passed down to numpy.linspace as length of array
      (see #2533)
+   * Geographic select of inventory/network/station (see #2515)
  - obspy.clients.fdsn:
    * Add new `_discover_services` boolean flag to the Client, which allows the
      Client to skip the initial services query at instantiation.  This can
--- a/misc/docs/source/packages/obspy.geodetics.rst
+++ b/misc/docs/source/packages/obspy.geodetics.rst
@@ -10,8 +10,10 @@
        :nosignatures:
 
        ~base.calc_vincenty_inverse
+       ~base.degrees2kilometers
        ~base.gps2dist_azimuth
-       ~base.kilometer2degrees
+       ~base.inside_geobounds
+       ~base.kilometers2degrees
        ~base.locations2degrees
        ~flinnengdahl.FlinnEngdahl
 
@@ -27,4 +29,3 @@
        flinnengdahl
 
     .. comment to end block
-
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -531,7 +531,9 @@ class Inventory(ComparingObject):
 
     def select(self, network=None, station=None, location=None, channel=None,
                time=None, starttime=None, endtime=None, sampling_rate=None,
-               keep_empty=False):
+               keep_empty=False, minlatitude=None, maxlatitude=None,
+               minlongitude=None, maxlongitude=None, latitude=None,
+               longitude=None, minradius=None, maxradius=None):
         r"""
         Return a copy of the inventory filtered on various parameters.
 
@@ -597,10 +599,39 @@ class Inventory(ComparingObject):
             or at given point in time (i.e. channels starting after given time
             will not be shown).
         :type sampling_rate: float
+        :param sampling_rate: Only include channels whose sampling rate
+            matches the given sampling rate, in Hz (within absolute tolerance
+            of 1E-8 Hz and relative tolerance of 1E-5)
         :type keep_empty: bool
         :param keep_empty: If set to `True`, networks/stations that match
             themselves but have no matching child elements (stations/channels)
             will be included in the result.
+        :type minlatitude: float
+        :param minlatitude: Only include stations/channels with a latitude
+            larger than the specified minimum.
+        :type maxlatitude: float
+        :param maxlatitude: Only include stations/channels with a latitude
+            smaller than the specified maximum.
+        :type minlongitude: float
+        :param minlongitude: Only include stations/channels with a longitude
+            larger than the specified minimum.
+        :type maxlongitude: float
+        :param maxlongitude: Only include stations/channels with a longitude
+            smaller than the specified maximum.
+        :type latitude: float
+        :param latitude: Specify the latitude to be used for a radius
+            selection.
+        :type longitude: float
+        :param longitude: Specify the longitude to be used for a radius
+            selection.
+        :type minradius: float
+        :param minradius: Only include stations/channels within the specified
+            minimum number of degrees from the geographic point defined by the
+            latitude and longitude parameters.
+        :type maxradius: float
+        :param maxradius: Only include stations/channels within the specified
+            maximum number of degrees from the geographic point defined by the
+            latitude and longitude parameters.
         """
         networks = []
         for net in self.networks:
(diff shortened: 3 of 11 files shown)


claudiodsf:
Fixed a typo in documentation and force-pushed.

ThomasLecocq:
would it make sense to put the location-based filtering in some generic util module to avoid the large copy/paste of identical codes ?

obspy/core/inventory/inventory.py
```diff
@@ -619,7 +650,12 @@ def select(self, network=None, station=None, location=None, channel=None,
             net_ = net.select(
                 station=station, location=location, channel=channel, time=time,
                 starttime=starttime, endtime=endtime,
-                sampling_rate=sampling_rate, keep_empty=keep_empty)
+                sampling_rate=sampling_rate,
+                minlatitude=minlatitude, maxlatitude=maxlatitude,
+                minlongitude=minlongitude, maxlongitude=maxlongitude,
+                latitude=latitude, longitude=longitude,
+                minradius=minradius, maxradius=maxradius,
+                keep_empty=keep_empty)
```
megies: Please move `keep_empty` to old position in kwargs list. It's very unlikely that people specify everything as args, but it still has potential to break codes.

obspy/core/inventory/inventory.py
```diff
@@ -531,7 +531,9 @@ def get_orientation(self, seed_id, datetime=None):
 
     def select(self, network=None, station=None, location=None, channel=None,
                time=None, starttime=None, endtime=None, sampling_rate=None,
-               keep_empty=False):
+               minlatitude=None, maxlatitude=None, minlongitude=None,
+               maxlongitude=None, latitude=None, longitude=None,
+               minradius=None, maxradius=None, keep_empty=False):
```
megies: Same, keep order of existing kwargs.

obspy/core/inventory/network.py
```diff
@@ -331,7 +332,9 @@ def get_orientation(self, seed_id, datetime=None):
 
     def select(self, station=None, location=None, channel=None, time=None,
                starttime=None, endtime=None, sampling_rate=None,
-               keep_empty=False):
+               minlatitude=None, maxlatitude=None, minlongitude=None,
+               maxlongitude=None, latitude=None, longitude=None,
+               minradius=None, maxradius=None, keep_empty=False):
```
megies: same

obspy/core/inventory/inventory.py
claudiodsf: I thought about that: it seems odd to me to have this `keep_empty` argument right in the middle of the arglist.

But if that's what you prefer, I'll do it 😉

ThomasLecocq:
> > would it make sense to put the location-based filtering in some generic util module to avoid the large copy/paste of identical codes ?
> 
> I thought about it. Where would you put this generic code?

Yeah, not easy indeed, and the way you skip the channel is indeed not so easily "generalizable"...

maybe something like this:

```python

def _filter_object_geographically(object, min_latitude, max_latitude, ...):
    if minlatitude is not None:
        if obj.latitude is None or obj.latitude < minlatitude:
            return False
    if maxlatitude is not None:
        if obj.latitude is None or obj.latitude > maxlatitude:
            return False
    if minlongitude is not None:
        if obj.longitude is None or obj.longitude < minlongitude:
            return False
    if maxlongitude is not None:
        if obj.longitude is None or obj.longitude > maxlongitude:
            return False
    if all([l is not None for l in
           (latitude, longitude, obj.latitude, obj.longitude)]):
        distance = locations2degrees(latitude, longitude,
                                     obj.latitude, obj.longitude)
        if minradius is not None and distance < minradius:
            return False
        if maxradius is not None and distance > maxradius:
            return False
    return True
```
and then in your calls:

```python
for cha in self.channels:
    if _filter_object_geographically(cha, min_latitude, max_latitude, ...):
        channels.append(cha)
```

obspy/core/inventory/inventory.py
megies: yes, please.. potential code breaks trump cosmetics :smiley:

megies:
> would it make sense to put the location-based filtering in some generic util module to avoid the large copy/paste of identical codes ?

avoiding code duplication is always good :+1:

claudiodsf:
> > would it make sense to put the location-based filtering in some generic util module to avoid the large copy/paste of identical codes ?
> 
> avoiding code duplication is always good 👍

Ok, should I put then this common code in `core/inventory/util.py` ?

megies:
> Ok, should I put then this common code in `core/inventory/util.py` ?

I guess it could be useful for events/catalog as well? Then it should go into `core/util/misc.py` or maybe better `geodetics/base.py` (the latter might create a circular import though, didnt check)

ThomasLecocq:
actually, I think it could be in ``geodetics/base.py`` and be a public function, as long as an object exposes the ``latitude`` and ``longitude`` parameters, it could be used.

megies:
Yes, agree. If it causes a circular import we see it in CI

obspy/core/inventory/inventory.py
claudiodsf: Done. Will squash and rebase before merging.

obspy/core/inventory/inventory.py
claudiodsf: Done.

obspy/core/inventory/network.py
claudiodsf: Done

calum-chamberlain:
It might be a good idea to add a test that this works across hemispheres - e.g. in New Zealand we have stations both sides of 180W and have run into issues in the past with simple `minlatitude` and `maxlatitude` not wrapping the search around +/- 180 (GeoNet fixed this issue [here](https://github.com/GeoNet/fdsn/issues/77)).

megies:
We had some issue with wrapping around longitude +-180° at some point, I was pretty sure, but I can't find it now..

ThomasLecocq:
@claudiodsf while you are at it, maybe think about making the selection inclusive or not( > or >= etc.) ?

claudiodsf:
Finally found some time to work on this. Let me know what do you think.
I'll add a `CHANGELOG` entry before merging :wink:

claudiodsf:
> @claudiodsf while you are at it, maybe think about making the selection inclusive or not( > or >= etc.) ?

Not sure whether this is essential: geographic coordinates are never that precise to require strict equality. And that would make the code and the function interface more complex.

claudiodsf:
Not sure which kind of errors the docs builds reports...

claudiodsf:
Also, feel free to propose a better name for `filter_object_geographically()`.
It is more a check than a filter : check that an object is within a certain geographic range.

claudiodsf:
Possible alternative names:

```
object_within_geographic_bounds()
object_inside_geographic_bounds()
object_within_geographic_range()
object_inside_geographic_range()
```

Other ideas?

ThomasLecocq:
``inside_geobounds`` ? :-)

claudiodsf:
> `inside_geobounds` ? :-)

@megies what do you think?

megies:
Works for me

ThomasLecocq:
@claudiodsf please make the name change and I'll pass for another (hopefully last) review of the code.

claudiodsf:
> @claudiodsf please make the name change and I'll pass for another (hopefully last) review of the code.

@ThomasLecocq Done! Let's wait on CI tests, then please review!

CHANGELOG.txt
```diff
@@ -88,6 +89,9 @@ master:
      #2405)
  - obspy.clients.iris:
    * results of distaz method are now returned as native floats (see #2499)
+ - obspy.geodetics:
+   * new utility function `inside_geobounds()` to check wether an object is
```
ThomasLecocq: whether

misc/docs/source/packages/obspy.geodetics.rst
```diff
@@ -10,8 +10,10 @@
        :nosignatures:
 
        ~base.calc_vincenty_inverse
+       ~base.degrees2kilometers
        ~base.gps2dist_azimuth
-       ~base.kilometer2degrees
+       ~base.inside_geobounds
+       ~base.kilometers2degrees
```
ThomasLecocq: @megies why 
https://docs.obspy.org/packages/autogen/obspy.geodetics.base.kilometer2degrees.html 
and 
https://docs.obspy.org/packages/autogen/obspy.geodetics.base.kilometers2degrees.html 
both exist ?

obspy/core/inventory/network.py
```diff
@@ -407,13 +439,23 @@ def select(self, station=None, location=None, channel=None, time=None,
                 if not sta.is_active(time=time, starttime=starttime,
                                      endtime=endtime):
                     continue
+            geo_filters = (minlatitude, maxlatitude,
```
ThomasLecocq: I knoooow I'm a pit picky on that one now, but didn't you say it's not a filter ? ;)

obspy/core/inventory/network.py
```diff
@@ -407,13 +439,23 @@ def select(self, station=None, location=None, channel=None, time=None,
...
+                           minlongitude, maxlongitude,
+                           latitude, longitude,
+                           minradius, maxradius)
+            if not inside_geobounds(sta, *geo_filters):
```
ThomasLecocq: so this is called even if all items in ``geo_filters`` are ``None`` ? It could maybe be skipped?

obspy/core/inventory/station.py
```diff
@@ -394,6 +426,12 @@ def select(self, location=None, channel=None, time=None, starttime=None,
                 if not cha.is_active(time=time, starttime=starttime,
                                      endtime=endtime):
                     continue
+            geo_filters = (minlatitude, maxlatitude,
+                           minlongitude, maxlongitude,
+                           latitude, longitude,
+                           minradius, maxradius)
+            if not inside_geobounds(cha, *geo_filters):
```
ThomasLecocq: same here, skip check if all ``None`` ?

misc/docs/source/packages/obspy.geodetics.rst
megies: No idea must be a typo??

obspy/core/inventory/network.py
megies: just a local variable, so doesnt really matter

megies:
Made some final changes, will merge when CI is green, CC @claudiodsf

claudiodsf:
@megies @ThomasLecocq Thanks for taking care of this: I have been sick and unable to work during the last few days.

PR #2516, opened 2019-12-17
Fix a typo in get_stations() and get_events() docs

claudiodsf:
A very small fix in `fdsn.client.Client.get_stations()` and `get_events()` docs.

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -442,7 +442,7 @@ class Client(object):
         :type latitude: float, optional
         :param latitude: Specify the latitude to be used for a radius search.
         :type longitude: float, optional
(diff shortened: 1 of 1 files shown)


claudiodsf:
+DOCS

ThomasLecocq:
the doc build look OK to me, dunno why the ❌ is there @megies ?

megies:
> the doc build look OK to me, dunno why the x is there @megies ?

For purely idealistic and non-realistic reasons. The intention was that we get a clean docs build without warnings at some point, in order to be able to tell sanity of PR's docs-wise directly from CI results, without having to sift through built docs pages.. but yeah.. that never happened, we have tons of warnings in sphinx build, therefore the :x: in each and every docs buildbot message

megies:
Thanks @claudiodsf

PR #2296, opened 2019-12-18
sc3ml read_inventory performance issue on large files

megies:
**EDIT: This post was made by** @medlin01GA

Obspy version: 1.1.0
Python version: 2.7, 3.x (all platforms)
Platform: all

When loading sc3ml format station inventory using function `obspy.read_inventory` with channel data, file read time grows as O(N*M) where N is the total number of channels and M is the total number sensors, dataloggers, and responses.  Typically M ~ N, so load time is ~O(N^2), but it should be more like O(N).

Reading a large inventory (~500 MB) of global seismic networks in performance profiling tests on high performance workstation type platform takes in the order of 10 hours. The root cause is the fact that sc3ml has an unordered, flat list of instrument data for sensors, dataloggers and responses, and when reading sc3ml format the obspy implementation does a brute force search of such lists for each sensor, datalogger or response in each channel.

Code experimentation in medlin01GA fork shows that by pre-indexing the instrumentation nodes from sc3ml, the complete inventory file read time for ~500 MB file can be brought down to about 4 minutes.

Note: The self-contained example I have requires seiscomp3 to be installed on the test platform. Such an example could be provided on request, but I do not attach the test data file directly since problem is one of scaling to big data, and the file is large.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/seiscomp/inventory.py
+++ b/obspy/io/seiscomp/inventory.py
@@ -465,7 +465,6 @@ def _read_channel(instrumentation_register, cha_element, _ns):
     numerator = _tag2obj(cha_element, _ns("sampleRateNumerator"), int)
     denominator = _tag2obj(cha_element, _ns("sampleRateDenominator"), int)
 
-    # If numerator is zero, set rate to zero irrespective of the denominator.
     # If numerator is non-zero and denominator zero, will raise
     # ZeroDivisionError.
     rate = numerator / denominator if numerator != 0 else 0
--- a/obspy/io/seiscomp/inventory.py
+++ b/obspy/io/seiscomp/inventory.py
@@ -146,16 +146,16 @@ def _read_sc3ml(path_or_file_object):
         "responseFIR": {},
         "responseIIR": {}
     }
-    for response_type in responses.keys():
+    for response_type, all_elements in responses.items():
         for response_element in inv_element.findall(_ns(response_type)):
             public_id = response_element.get("publicID")
             if public_id:
-                if public_id in responses[response_type]:
+                if public_id in all_elements:
                     msg = ("Found multiple matching {} tags with the same "
                            "publicID '{}'.".format(response_type, public_id))
                     raise obspy.ObsPyException(msg)
                 else:
-                    responses[response_type][public_id] = response_element
+                    all_elements[public_id] = response_element
     # Organize all the collection instrument information into a unified
     # intrumentation register.
     instrumentation_register = {
@@ -431,9 +431,8 @@ def _read_channel(instrumentation_register, cha_element, _ns):
         response_id = sensor_element.get("response")
         response_elements = []
 
-        for resp_type in instrumentation_register["responses"].keys():
-            found_response = instrumentation_register["responses"][resp_type]\
-                             .get(response_id)
+        for resp_element in instrumentation_register["responses"].values():
+            found_response = resp_element.get(response_id)
             if found_response is not None:
                 response_elements.append(found_response)
 
@@ -518,9 +517,9 @@ def _read_channel(instrumentation_register, cha_element, _ns):
         if digital_filter_chain is not None:
             response_fir_id = digital_filter_chain.split(" ")
 
-    channel.response = _read_response(instrumentation_register, sensor_element,
-                                      response_element, cha_element,
-                                      data_log_element, _ns,
+    channel.response = _read_response(instrumentation_register['responses'],
+                                      sensor_element, response_element,
+                                      cha_element, data_log_element, _ns,
                                       channel.sample_rate,
                                       response_fir_id, response_paz_id)
 
@@ -553,13 +552,15 @@ def _read_instrument_sensitivity(sen_element, cha_element, _ns):
     return sensitivity
 
 
-def _read_response(instrumentation_register, sen_element, resp_element,
+def _read_response(instrumentation_responses, sen_element, resp_element,
                    cha_element, data_log_element, _ns, samp_rate, fir,
                    analogue):
     """
     reads response from sc3ml format
 
-    :param instrumentation_register: register of instrumentation metadata
+    :param instrumentation_responses: Dictionary of dictionaries of
+        instrumentation response metadata, top level keyed by response type,
+        and subdictionaries keyed by response ID.
     :param _ns: namespace
     """
     response = obspy.core.inventory.response.Response()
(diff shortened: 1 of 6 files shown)


megies:
Done now @medlin01GA, we will have to see if this can make 1.2.0 though, only clean PR's that don't need help from core devs can make it in at this stage (otherwise it'll be anther year to release)

medlin01GA:
I realize this might not make 1.2.0 due to the timing, that's perfectly ok. But I was just wondering if I need to do anything more at the point to get the PR into review? I do not seem to be able to assign reviewers myself.

megies:
> anything more at the point to get the PR into review? I do not seem to be able to assign reviewers myself.

Well, assigning somebody doesn't do much anyway. Mostly you gotta pray that somebody is willing to commit time to do a review. The problem is that hardly any people are and those that are are just swamped in open and half-ready issues.

The whole problem of this PR is/was that is was in such a tangled up state with that other related issue being worked on in parallel and then commits being hijacked from one PR to the other and it then being near impossible to tell whats goin on for people not authoring those PRS.

obspy/io/seiscomp/inventory.py
```diff
@@ -143,7 +143,8 @@ def _ns(tagname):
     responses = {
         "responsePAZ": {},
         "responsePolynomial": {},
-        "responseFIR": {}
+        "responseFIR": {},
+        "responseIIR": {}
```
megies: You could later on do `responses.setdefault(response_type, {})[public_id] = response_element` and not worry about setting up all potential response types ahead of time?

megies:
One minor comment, please add changelog one-liner, otherwise good to go (I will clean up the branch before merging, when its ready otherwise).

obspy/io/seiscomp/inventory.py
medlin01GA: I have considered this suggestion. However I have decided not to action this any further. I would rather that the responses list was not explicit like this, but the previous code also only looked for specific hard-wired response types, and with the current changes at least this list of response types is centralized in one code statement. Using `setdefault` is a good way to simplify population of a dictionary, however making such a change here would result in losing the check for duplication of the public_id, which raises a potentially informative exception.

medlin01GA:
Done.

________________________________

obspy/io/seiscomp/inventory.py
megies: Fine with me, didn't look at all implementation details, was just thinking about what happens when an unknown response type is encountered, but I guess you got all types covered that can be encountered.

megies:
Cleaned up the branch and rebased.

obspy/io/seiscomp/inventory.py
```diff
@@ -111,11 +111,64 @@ def _ns(tagname):
...
+    for response_type in responses.keys():
+        for response_element in inv_element.findall(_ns(response_type)):
+            public_id = response_element.get("publicID")
+            if public_id:
+                if public_id in responses[response_type]:
+                    msg = ("Found multiple matching {} tags with the same "
+                           "publicID '{}'.".format(response_type, public_id))
+                    raise obspy.ObsPyException(msg)
+                else:
+                    responses[response_type][public_id] = response_element
```
QuLogic: Can be shortened a tiny bit (feel free to choose a different name for `all_elements`).
```python3
    for response_type, all_elements in responses.items():
        for response_element in inv_element.findall(_ns(response_type)):
            public_id = response_element.get("publicID")
            if public_id:
                if public_id in all_elements:
                    msg = ("Found multiple matching {} tags with the same "
                           "publicID '{}'.".format(response_type, public_id))
                else:
                    all_elements[public_id] = response_element
```

obspy/io/seiscomp/inventory.py
```diff
@@ -374,9 +431,12 @@ def _read_channel(inventory_root, cha_element, _ns):
         response_id = sensor_element.get("response")
         response_elements = []
 
-        for resp_type in ['responsePAZ', 'responsePolynomial']:
-            search = "{}[@publicID='{}']".format(resp_type, response_id)
-            response_elements += inventory_root.findall(_ns(search))
+        for resp_type in instrumentation_register["responses"].keys():
```
QuLogic: Doesn't seem like you need `resp_type`.
```suggestion
        for resp_elements in instrumentation_register["responses"].values():
```

obspy/io/seiscomp/inventory.py
```diff
@@ -374,9 +431,12 @@ def _read_channel(inventory_root, cha_element, _ns):
...
+            found_response = instrumentation_register["responses"][resp_type]\
```
QuLogic:
```suggestion
            found_response = resp_elements.get(response_id)
```

obspy/io/seiscomp/inventory.py
```diff
@@ -494,12 +553,13 @@ def _read_instrument_sensitivity(sen_element, cha_element, _ns):
     return sensitivity
 
 
-def _read_response(root, sen_element, resp_element, cha_element,
-                   data_log_element, _ns, samp_rate, fir, analogue):
+def _read_response(instrumentation_register, sen_element, resp_element,
```
QuLogic: It doesn't look like you need all of `instrumentation_register` here, so I would pass `instrumentation_register['responses']` to this function. Then below you get shorter lines and less directory indexing.

megies:
@medlin01GA can you please have a look regarding above review comments?

obspy/io/seiscomp/inventory.py
medlin01GA: Changed as suggested.

obspy/io/seiscomp/inventory.py
medlin01GA: Changed as suggested.

obspy/io/seiscomp/inventory.py
medlin01GA: Changed as suggested.

obspy/io/seiscomp/inventory.py
medlin01GA: Changed as suggested.

megies:
Thanks @medlin01GA for working in the review. You might want to play around with rebasing some (https://git-scm.com/book/en/v2/Git-Branching-Rebasing), had to clean up the branch with a manual rebase once again to get rid of these ugly merge commits.. :-)

megies:
CI fails are unrelated, so merging

PR #2527, opened 2019-12-25
stationxml: work around empty string tags getting converted to a string value "None"

### LINKED ISSUE #2519: read_inventory converts empty sub-elements to the 'None' string
When read_inventory() reads a stationxml file, it converts empty sub-elements to the 'None' string.

Looking at obspy/io/stationxml/core.py:

```python
710 def _read_equipment(equip_element, _ns):
713     description = _tag2obj(equip_element, _ns("Description"), str)
...
1448 def _tag2obj(element, tag, convert):
1449     # we use future.builtins.str and are sure we have unicode here
1450     try:
1451         return convert(element.find(tag).text)
```

When an element is present (e.g., sensor) but the subelement is not (e.g., sensor.description),
_tag2obj returns str(None) which is the string 'None'.

So if a stationxml containing:

    <Sensor>
         <Description></Description>
    </Sensor>

is read in and then written back out using inventory.write(), it writes:

    <Sensor>
        <Description>None</Description>
    </Sensor>

i.e., it writes 'None' in place of the empty sub-element.
Is this intentional ?  I know the xsd specifies that a sensor must have a description, but I've seen stationxml files out in the wild that are missing it.

I just don't know if it's better to keep the field empty on disk (=None in memory) or to use the "None" string as placeholder ?

Thanks!

megies:
### What does this PR do?

Tries to fix empty string tags getting converted to a string `"None"` since lxml returns `None` instead of `''` which is what we kind of expected inherently it seems.

Fixes #2519 
CC @mikehagerty

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,8 @@ master:
    * fixed UTCDateTime(..., julday=366) for non-leap years. This was returning
      January 1st of the next year in case of non-leap years being used. Now it
      properly raises an out-of-bounds ValueError (see #2369)
+   * When reading StationXML/SC3ML, make sure to properly read empty string
+     fields as empty strings instead of "None" (see #2519 and #2527)
    * better ISO8601 detection for UTCDateTime objects and UTCDateTime(...,
      iso8601=False) now completely disables ISO8601 handling (see #2447)
    * Added replace method to UTCDateTime class (see #2077).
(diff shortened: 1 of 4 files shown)


megies:
stationxml tests seem to work locally, I think we need to add a regression test still

megies:
@mikehagerty can you please give this a review, since you reported the issue initially?

megies:
Test fails are unrelated and no feedback from original poster, so merging for now

PR #2531, opened 2020-01-12
Select traces in Stream based on an Inventory

claudiodsf:
### What does this PR do?

Select traces in `Stream` based on the content of an `Inventory` object:
trace will be selected if the inventory contains a matching channel
active at the trace start time.

This is a follow-up of #2515.
Sometimes datacenter provides preassembled datasets with data and metadata for all the stations in the network (ex.: http://cnt.rm.ingv.it/event/23558121 --> Download). 
Using #2515, inventory can be filtered based on geographic parameters.
With the present PR, data in stream can be selected based on the inventory content.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1822,7 +1822,9 @@ class Stream(object):
         All other selection criteria that accept strings (network, station,
         location) may also contain Unix style wildcards (``*``, ``?``, ...).
         """
-        if inventory is not None:
+        if inventory is None:
+            traces = self.traces
+        else:
             trace_ids = []
             start_dates = []
             end_dates = []
@@ -1852,7 +1854,7 @@ class Stream(object):
                         traces.append(trace)
                     except ValueError:
                         break
-            return self.__class__(traces=traces)
+        traces_after_inventory_filter = traces
 
         # make given component letter uppercase (if e.g. "z" is given)
         if component is not None and channel is not None:
@@ -1864,7 +1866,7 @@ class Stream(object):
                       "mutually exclusive!"
                 raise ValueError(msg)
         traces = []
-        for trace in self:
+        for trace in traces_after_inventory_filter:
             # skip trace if any given criterion is not matched
             if id and not fnmatch.fnmatch(trace.id.upper(), id.upper()):
                 continue
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1748,10 +1748,16 @@ class Stream(object):
             yield temp
 
     def select(self, network=None, station=None, location=None, channel=None,
-               sampling_rate=None, npts=None, component=None, id=None):
+               sampling_rate=None, npts=None, component=None, id=None,
+               inventory=None):
         """
         Return new Stream object only with these traces that match the given
         stats criteria (e.g. all traces with ``channel="EHZ"``).
+
+        Alternatively, traces can be selected based on the content of an
+        :class:`~obspy.core.inventory.inventory.Inventory` object: trace will
+        be selected if the inventory contains a matching channel active at the
+        trace start time.
 
         .. rubric:: Examples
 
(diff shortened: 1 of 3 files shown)


obspy/core/stream.py
```diff
@@ -1797,6 +1822,38 @@ def select(self, network=None, station=None, location=None, channel=None,
...
+                        if start_date is not None and\
+                                trace.stats.starttime < start_date:
+                            continue
+                        if end_date is not None and\
+                                trace.stats.endtime > end_date:
+                            continue
+                        traces.append(trace)
+                    except ValueError:
+                        break
+            return self.__class__(traces=traces)
```
megies: If I understand correctly any other options specified to `select()` will be ignored if an Inventory is provided?

obspy/core/stream.py
```diff
@@ -1797,6 +1822,38 @@ def select(self, network=None, station=None, location=None, channel=None,
...
+            start_dates = []
+            end_dates = []
+            for net in inventory.networks:
+                for sta in net.stations:
+                    for chan in sta.channels:
+                        id = '.'.join((net.code, sta.code,
+                                       chan.location_code, chan.code))
+                        trace_ids.append(id)
+                        start_dates.append(chan.start_date)
+                        end_dates.append(chan.end_date)
```
megies: So it will only keep Traces that have info in the inventory down to Channel level, it looks like? If thats the case, functionality is pretty similar to the lookup done in `Stream.remove_response()` which eventually calls `Inventory.get_channel_metadata(seed_id, datetime=None)` and it should likely be combined with that one to avoid code duplication

megies: Also `Inventory.get_contents()` has similar functionality, although missing start/end times of channels, but it might be possible to adapt that routine and avoid duplication.

obspy/core/stream.py
claudiodsf: Correct. It seems to me a strange use case to mix the two criteria. What do you think?

megies: I think we should stick with the current approach of chaining all checks with logical AND, i.e. any filter criteria specified by the user should actually also be applied. I don't think it's good practice to have one kwarg completely override others too much, it'd be hell for a user to check which combination of kwargs together actually do what.

krischer: I shared @megies opinion. This would be really confusing to me and one passes multiple criteria I would intuitively expect them all to be applied.

claudiodsf: Ok, I'll modify that.

obspy/core/stream.py
claudiodsf: I tested several approaches and this one, which is similar to `Inventory.get_contents()`, is the fastest.

What should we do to avoid duplication? Is it fine to have an option for `Inventory.get_contents()` which returns also the start and end dates? 

Something like `Inventory.get_contents(channel_dates=False)`, which –if `True`– adds to the returned dictionary two more keys: `'channel_start_dates'` (with a list of start dates as value) and `'channel_end_dates'` (with a list of end dates as value).

What do you think?

d-chambers: >What should we do to avoid duplication?

One thing I have been planning to look into (just haven't had time yet) is add a `get_stations` method to inventory with the same signature as [`fdsn.Client.get_stations`](https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_stations.html). Then all the filtering type methods can simply call `get_stations` with the appropriate keywords under the hood to avoid duplication. I have a functioning implementation [here](https://github.com/niosh-mining/obsplus/blob/master/obsplus/stations/get_stations.py) but it it doesn't yet support all the keywords. Looking closer at it, it doesn't even use pandas so it may be readily adaptable to this case if you want to use/improve it.

The idea of using the `get_x` as an *interface* for selecting data has proved very useful in obsplus. Its nice to write code which only needs to understand the type of data (waveform, station, event) and not the whether the data is a local or remote resource.

megies: In any case, whatever we do, it's gotta be done within the next couple days. If you cant get rid of duplications by then it stays as is for now. Wanna push the RC ASAP. :)

megies: "whatever we do" with "we" as in @claudiodsf  btw ;-P

obspy/core/stream.py
megies: Please try to get it done ASAP so I can bundle the RC thanks! :-)

obspy/core/stream.py
megies: whelp.. its not exposed in a public helper routine.. so gotta get rid of duplications later i guess.. cant hold off RC for it now

obspy/core/stream.py
megies: Did it myself. Tests aren't comprehensive, btw, no fails popping up after my change it seems

claudiodsf:
@megies  Thanks for taking care of this: I have been sick and unable to work during the last few days.

PR #2535, opened 2020-01-21
mseed: fix bug leading to endless loop when reading a fullseed file without data records as MSEED

### LINKED ISSUE #2534: Infinite loop when trying to read full SEED file without any mini SEED records
When trying to read a full SEED file without any mini SEED records, while specifying input format MSEED, the code runs into an infinite loop.

Example file (contained FSEED file): http://webservices.ingv.it/webservices/ingv_ws_data/data/view_waveform_at_ingv.it/2015/08/24/093006_1594696/1594696_data.tgz

Code to reproduce:
```
import obspy
obspy.read('1594696_data/1594696_data.fseed', 'MSEED')
```

Code gets stuck searching for first data record, in the following loop:
https://github.com/obspy/obspy/blob/6d6d20c7c35126b18de5d33af962aa1c51dbfdd7/obspy/io/mseed/util.py#L646

I suppose the problem could be fixed with a check if EOF has been reached. For intended behavior I'm unsure if an exception should be raise or rather if just an empty Stream object should be returned.

ObsPy version: 1.1.0 (installed through pip in a conda virtual env)
Python version: 3.6.5
OS: Ubuntu 18.04

megies:
### What does this PR do?

Removes a bug resulting in an infinite loop when trying to read a full SEED file without any data records.

Fixes #2534 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -99,6 +99,8 @@ master:
  - obspy.io.gcf:
    * Fixes Python 3.8 compatibility of GCF reader. (see #2505)
  - obspy.io.mseed:
+   * Fix a bug resulting in an infinite loop when trying to read a FullSEED
+     file without any data records (see #2534 and #2535)
    * Add ability to write int64 data to mseed if it can safely be downcast
      to int32 data, otherwise raises ObsPyMSEEDError. (see #2356)
    * The recordanalyzer can now detect calibration blockettes 300, 310,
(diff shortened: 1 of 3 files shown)


megies:
Fails are unrelated.. will merge this tomorrow if no objections come up @krischer 
Only question really is whether to raise an Exception or return an empty Trace, but it seems all similar cases inside that subroutine all raise an Exception, so it should OK fine like this.

Issue #2295, opened 2019-01-25
Release 1.2.0


megies:
### Milestone https://github.com/obspy/obspy/milestone/16

Just reserved a DOI on zenodo for next release: `10.5281/zenodo.3674646`

Things to do, when release candidates and final release are done:
- [x] Run docker tests (@krischer / @megies -- ~~are done routinely now, but double-checking can't hurt~~)
- [x] source distribution to pypi
- [x] conda packages via conda-forge: conda-forge/obspy-feedstock#6
- [ ] ~~conda packages for Linux 32bit via https://github.com/obspy/obspy/tree/master/misc/installer/anaconda and distributed at https://anaconda.org/obspy/obspy/files~~ do we still need this?? (@krischer / @megies)
- [ ] maxiconda bundled installer via recipes in https://github.com/obspy/anaconda-installers but built locally on win+nix and attached to release: https://github.com/obspy/obspy/releases/tag/1.1.0 (@megies and @barsch)
- [ ] wheels for Linux+OSX (@krischer)
- [x] Windows pypi packages (@barsch / appveyor)
- [ ] debian packages (@megies)
- [ ] redhat/centos packages (@qulogic)
- [ ] macports packages (@petrrr)
- [x] zenodo: reserve a DOI pre-release (@megies) `10.5281/zenodo.3674646`
- [x] zenodo: upload sdist zipball and finalize release (@megies)
- [x] Update docs once released (@megies)
- [x] News entry in the wiki
- [x] "What's new?" page in the wiki
- [x] Email to the users and announcements mailing lists
- [x] Tweet
- [ ] update base branches for all issues bumped to 1.2.1 to a new branch `maintenance_1.2.x`
- [x] check dependency versions, some old distros we used to support might have gone EOL since release of 1.0, so we might be able to bump minimum dependency versions of e.g. numpy, scipy, matplotlib, ...
---
- [ ] Party :tada: (don't tick before all other checkboxes are ticked!)

megies:
@obspy/developers @obspy/packagers @obspy/conda-forgers
@ThomasLecocq, @krischer and myself will meet for a few days mid February for a little ObsPy sprint. The agenda is kinda obvious, so I hope we'll manage to get at least 1.1.1 out and hopefully also 1.2.0. That means any PRs not in good shape or merging by then will definitely postponed.

megies:
[...]

PR #2540, opened 2020-01-30
do not show taupy warning, shorten traceback for failing network arclink test

trichter:
See comments in #2295.

### DIFF
--- a/obspy/taup/tau_branch.py
+++ b/obspy/taup/tau_branch.py
@@ -489,9 +489,9 @@ class TauBranch(object):
         try:
             arr.resize(new_size)
         except ValueError:
-            # msg = ('Resizing a TauP array inplace failed due to the existence'
-            #        ' of other references to the array, creating a new array. '
-            #        'See Obspy #2280.')
+            # msg = ('Resizing a TauP array inplace failed due to the '
+            #        'existence of other references to the array, creating '
+            #        'a new array. See Obspy #2280.')
             # warnings.warn(msg)
             arr = np.resize(arr, new_size)
         return arr
--- a/obspy/taup/tau_branch.py
+++ b/obspy/taup/tau_branch.py
@@ -8,8 +8,6 @@ from __future__ import (absolute_import, division, print_function,
 from future.builtins import *  # NOQA
 from future.utils import native_str
 
-import warnings
-
 import numpy as np
 
 from .c_wrappers import clibtau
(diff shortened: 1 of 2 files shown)


obspy/taup/tau_branch.py
```diff
@@ -489,9 +487,9 @@ def _robust_resize(arr, new_size):
...
             arr.resize(new_size)
         except ValueError:
-            msg = ('Resizing a TauP array inplace failed due to the existence'
-                   ' of other references to the array, creating a new array. '
-                   'See Obspy #2280.')
-            warnings.warn(msg)
+            # msg = ('Resizing a TauP array inplace failed due to the '
+            #        'existence of other references to the array, creating '
+            #        'a new array. See Obspy #2280.')
+            # warnings.warn(msg)
```
megies: Hmm.. I would assume this warning was put in there for a reason. Maybe leave it in and just catch it where it is raised in tests?

https://github.com/obspy/obspy/commit/223f0b4c7526070e1392dc8e8ca7d123a33e526e

trichter: Yeah, that could be done, but I have the impression it is raised very often under usual circumstances. And I do not feel that there is a big disadvantage using np.resize which satisfies this warning.
@d-chambers What is your opinion on the matter? I think you added the warning.

With silencing I really meant removing. If it should just be silenced: There are a lot of other warnings during the tests, I guess these could then be fixed in a large effort to get rid of all warnings in one rush (after 1.2).

megies: > I guess these could then be fixed in a large effort to get rid of all warnings in one rush (after 1.2).

Yes, after 1.2.0 please. I know there are a lot of warnings, usually just nobody takes them on. We might also want a mechanism to make sure no warnings are introduced by PRs, because we clean them up every now and then but I think many PRs bring in new ones. (e.g. catch all warnings in test runner and raise a test fail if warnings appeared.. might be to rough though, because it might be hard to get of all warnings on all test runs across all platforms etc)

trichter: > Yes, after 1.2.0 please.

I agree.


After rereading #2280, I think you are right and the warning has a reason to be there. It is more of a reminder that there is a minor bug somewhere in taupy. But I think the user should not be bothered with it. I suggest we merge this PR as is and I am going to open a new issue tracking this bug.

trichter:
Continued in referencing PR.

PR #2541, opened 2020-01-30
Taupy resize

trichter:
Continuation of #2280 and #2540. Waiting for CI.


trichter:
Here are the failing reports:
py3.7 http://tests.obspy.org/108445/
py3.8 http://tests.obspy.org/108445/

Interestingly, locally it is not raising any error for me (conda, python 3.7).

trichter:
Will be dealt with in the referenced PR.

PR #2542, opened 2020-01-30
Refactor _robust_resize in TauBranch

trichter:
`_robust_resize` was previously introduced (#2280) because of some numpy resize fails (#2541 for fails).

A warning was introduced, but because of this new addition the resizing now failed in all cases and the warning was therefore seen very often. This PR leaves the warning as is, but rewrites the _resize_robust method, that it only shows the warning and falls back to numpy.resize when it is really necessary.

This PR is branched from #2540 and includes its changes.

### DIFF
--- a/obspy/clients/arclink/tests/test_client.py
+++ b/obspy/clients/arclink/tests/test_client.py
@@ -393,7 +393,7 @@ class ClientTestCase(unittest.TestCase):
         start = UTCDateTime(2008, 1, 1)
         end = start + 1
         result = client.get_networks(start, end)
-        self.assertIn('BW', result.keys())
+        self.assertIn('BW', list(result.keys()))
         self.assertEqual(result['BW']['code'], 'BW')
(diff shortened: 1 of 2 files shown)


trichter:
CI looks good up till now. I am out of office until next week. Feel free to merge, polish or discard. Cheers.

d-chambers:
@trichter, returning the array in the `_robust_resize` was creating the extra references which caused the resizing to fail later on? 

Interesting find and good fix. I updated one thing on the docstring, otherwise looks good to merge.

trichter:
> @trichter, returning the array in the `_robust_resize` was creating the extra references which caused the resizing to fail later on?

I think passing the array to the classmethod in the first place was creating the extra reference.
 
> I updated one thing on the docstring, otherwise looks good to merge.

Thanks.

The test fail is interesting but unrelated.

krischer:
Interesting - I definitely learned something here.

Really nice find and thanks for the fix.

megies:
Yup, good catch.

PR #2550, opened 2020-02-10
io/nordic/core: fix to let read_nordic properly return waveform file name

flixha:
### What does this PR do?
It fixes an issue with the `read_nordic`-function which did not correctly return the waveform filenames when specifying the option `return_wavnames=True`.

The following code should, besides a catalog, return the waveform files for each event:
`from obspy.io.nordic.core import read_nordic
sfile = '/obspy/io/nordic/tests/data/01-0411-15L.S201309'
cat, wavname = read_nordic(sfile, return_wavnames=True)`
The option `return_wavnames=True` correctly worked in obspy 1.1.0. Refactoring of the functions in io/nordic led to a small bug: A loop over a StringIO object that contains the nordic file's contents was supposed to read the file line by line to find the waveform file name. But for the loop to work, the StringIO object needs to be at position 0 (`.seek(0)`).
The issue arose because the tests only checked the `readwavname(..)`-function, but not the `cat, wavnames = read_nordic(sfile, return_wavnames=True`-option.

### PR Checklist
- [] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes (I understand that there's not `maintenance_1.2.x` yet, therefore submitting to master)
- [x] All tests still pass.
~- [ ] Any new or changed features have are fully documented.~
~- [ ] Significant changes have been added to `CHANGELOG.txt` .~

### DIFF
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -812,7 +812,7 @@ def _readwavename(tagged_lines):
     Internal wave-name reader.
 
     :type tagged_lines: list
-    :param tagged_lines: 
+    :param tagged_lines:
         List of tuples of (strings, line-number) of the waveform lines.
 
     :return: list of wave-file names
--- a/obspy/io/nordic/core.py
+++ b/obspy/io/nordic/core.py
@@ -429,14 +429,13 @@ def _extract_event(event_str, catalog, wav_names, return_wavnames=False):
     for event_line in event_str:
         tmp_sfile.write(event_line)
     tagged_lines = _get_line_tags(f=tmp_sfile)
-    tmp_sfile.seek(0)
     new_event = _readheader(head_lines=tagged_lines['1'])
     new_event = _read_uncertainty(tagged_lines, new_event)
     new_event = _read_highaccuracy(tagged_lines, new_event)
     new_event = _read_focal_mechanisms(tagged_lines, new_event)
     new_event = _read_moment_tensors(tagged_lines, new_event)
     if return_wavnames:
-        wav_names.append(_readwavename(f=tmp_sfile))
+        wav_names.append(_readwavename(tagged_lines=tagged_lines['6']))
     new_event = _read_picks(tagged_lines=tagged_lines, new_event=new_event)
     catalog += new_event
     return catalog, wav_names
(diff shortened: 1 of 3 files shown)


obspy/io/nordic/tests/test_nordic.py
```diff
@@ -430,6 +430,9 @@ def test_read_wavename(self):
         testing_path = os.path.join(self.testing_path, '01-0411-15L.S201309')
         wavefiles = readwavename(testing_path)
         self.assertEqual(len(wavefiles), 1)
+        # Check that read_nordic reads wavname when return_wavnames=True
+        cat, wavefiles = read_nordic(testing_path, return_wavnames=True)
+        self.assertEqual(len(wavefiles[0]), 1)
```
megies: Any reason to not test the content of `wavefiles` fully?

flixha: The only reason is that the function `readwavename` which calls `_readwavename` is generally tested well. The only test that was lacking is the call from `read_nordic` to `_readwavename`. 
The 3 lines above are the simplemost test to avoid the bug that led to this PR, by doing the same test as is done on `readwavname` (lines 430-432). It seemed unnecessary to redo the whole test that happens for `readwavname` when only the way that the function is called changes. 
But you can see that I haven't contributed much yet, so I am very happy to extend the tests further if you think it's better to do that.

megies: I mean.. what is in this `wavefiles` variable..? It sounds like it should be a filename or something? But it's only length 1. What I was saying is, why only test `len(wavefiles[0])` and not what is actually stored in there?

calum-chamberlain: Agree - it may as well check what the returned filename actually is, although from memory the filename returned might be install-path dependent, but that shouldn't be too hard to test.

calum-chamberlain: Oh, and @megies, `wavefiles` is a `list` of `str`. Multiple wavefiles can be stored in the nordic s-file, so a list is returned by default.

calum-chamberlain:
Thanks for this catch @flixha.  Great to have more people looking at the nordic things.

How do you feel about making `_readwavename` take `tagged_lines` like most of the other internal functions, then the seek is unnecessary and the internal functions are more self-consistent? It looks like `_readwavename` is already looking for the `6` tag, so it is effectively duplicating the work that `_tag_lines` is doing.

obspy/io/nordic/tests/test_nordic.py
flixha: It's no problem to check the actual content of the list of strings `wavefiles` in this case, as it is the filename without a path. 
I was thinking that it would probably not be necessary to duplicate the tests that are done by calling `_readwavname` through `readwavname`, but I have now fixed this very comparison.

megies: If you think its good then it maybe is. If it was already tested somewhere then I don't understand how a bug fix was even necessary :stuck_out_tongue_winking_eye:

flixha: Well, the only thing that was not tested is the alternative way to call the function `_readwavename`. Its output itself was correct if it was supplied with the right input ;-)

flixha:
You're right that it would make the call to `_readwavename` from `_extract_event` cleaner. I don't see any disadvantage with doing it, so I've added the few changes that it required to this PR. All tests pass and I did the slight change to the docuemtnation of the function's input. Please let me know if you can think of any problems that can happen because of the change.

obspy/io/nordic/core.py
```diff
@@ -800,23 +799,27 @@ def readwavename(sfile, encoding='latin-1'):
...
 
-def _readwavename(f):
+def _readwavename(tagged_lines):
     """
     Internal wave-name reader.
 
-    :type f: file
-    :param f: File open in read-mode
+    :type tagged_lines: list
+    :param tagged_lines:
```
megies: Trailing whitespace here makes flake8 fail

flixha: sorry, should have checked flake8. Fixed now.

megies: No big deal and thanks.

PR #2555, opened 2020-02-12
adding RASPISHAKE to FDSN mapping

iannesbitt:
### What does this PR do?

This adds the base of the Raspberry Shake FDSN tree (http://fdsnws.raspberryshakedata.com) to the URL mapping list under `"RSHAKE"` so that it can be accessed via

```
client = Client("RSHAKE")
```

This is being initiated to add the [RASPISHAKE datacenter](ds.iris.edu/files/stats/datacenters/) (DOI: [10.7914/SN/AM](https://doi.org/10.7914/SN/AM)) to the list of URL mapping endpoints.

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.fdsn"
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -80,6 +80,7 @@
      clients (see #2342 and #2344)
    * make it possible to use signed EIDA tokens and also skip token validation
      completely (see #2297)
+   * adding a mapping for RSHAKE
  - obspy.clients.neic:
    * properly use specified timeout value (see #2450)
  - obspy.clients.seishub:
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -80,7 +80,7 @@
      clients (see #2342 and #2344)
    * make it possible to use signed EIDA tokens and also skip token validation
      completely (see #2297)
-   * adding a mapping for RSHAKE
+   * adding a mapping for RASPISHAKE
  - obspy.clients.neic:
    * properly use specified timeout value (see #2450)
  - obspy.clients.seishub:
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -67,6 +67,7 @@ Michelini, Alberto
 Miller, Nathaniel C.
 Morgenstern, Bernhard
 Murray-Bergquist, Louisa
+Nesbitt, Ian M.
 Nof, Ran Novitsky
 Panning, Mark P.
 Rapagnani, Giovanni
--- a/obspy/clients/fdsn/__init__.py
+++ b/obspy/clients/fdsn/__init__.py
@@ -33,7 +33,7 @@ available providers are:
 
 >>> from obspy.clients.fdsn.header import URL_MAPPINGS
 >>> for key in sorted(URL_MAPPINGS.keys()):
-...     print("{0:<7} {1}".format(key,  URL_MAPPINGS[key]))
+...     print("{0:<11} {1}".format(key,  URL_MAPPINGS[key]))
 BGR         http://eida.bgr.de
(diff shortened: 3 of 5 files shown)


iannesbitt:
It looks like (relevant) tests have passed, so updating PR checkbox above.

krischer:
One thing to consider is if we should exclude this special data center from the default data centers in the mass downloader (https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.mass_downloader.html).

It could still be included if users choose to do so but I think this should be an explicit choice given the non-standard nature of this particular data center. It might also hit it with a lot of potentially unwanted requests.

megies:
> It could still be included if users choose to do so but I think this should be an explicit choice given the non-standard nature of this particular data center. It might also hit it with a lot of potentially unwanted requests.

:+1: 

Also, I'd vote for `RASPISHAKE`
 - `RSHAKE` is not an (semi)official abbreviation as far as I know
 - it's way clearer
 - only 3 chars longer
 - it's the official Twitter handle
 - it's the name in the FDSN datacenter list

iannesbitt:
> Also, I'd vote for `RASPISHAKE`

I had considered this but decided against it because it was longer than any others, and required a reformatting of the tab alignment on the docs page. I agree that it fits better with the datacenter's other identifiers.

iannesbitt:
> One thing to consider is if we should exclude this special data center from the default data centers in the mass downloader

@krischer disabled by default would be best I think.

iannesbitt:
@krischer am I correct in assuming that the exclusion for RASPISHAKE is done in the init function of the MassDownloader class, i.e. [https://docs.obspy.org/_modules/obspy/clients/fdsn/mass_downloader/mass_downloader.html#MassDownloader.\_\_init\_\_](https://docs.obspy.org/_modules/obspy/clients/fdsn/mass_downloader/mass_downloader.html#MassDownloader.__init__)?

megies:
Looks like it. Also please add a note in the docstring for `providers` that raspishake is omitted by default and has to be specified/added explicitly.
I guess one would have to do e.g. something like the following to get raspishake in there

```python
from obspy.clients.fdsn import URL_MAPPINGS
MassDownloader(providers=URL_MAPPINGS)
```

iannesbitt:
@megies ok, consider these changes:

```python
class MassDownloader(object):
    """
    Class facilitating data acquisition across all FDSN web service
    implementations.

    :param providers: List of FDSN client names or service URLS. Will use
        all FDSN implementations known to ObsPy except RASPISHAKE if
        set to None. The order in the list also determines their priority, if data
        is available at more then one provider it will always be downloaded
        from the provider that comes first in the list. To include RASPISHAKE,
        you must set this parameter to
        :class:`~obspy.clients.fdsn.header.URL_MAPPINGS` explicitly.
    :param debug: Debug flag passed to the underlying FDSN web service clients.
    :type providers: list of str or :class:`~obspy.clients.fdsn.client.Client`
        instances
    """
    def __init__(self, providers=None, debug=False):
# ...
        if providers is None:
            providers = dict(URL_MAPPINGS.items())
            del providers['RASPISHAKE']
# ...
```

megies:
This needs to be adapted: https://travis-ci.org/obspy/obspy/jobs/650081061#L1438

Also some trailing whitespace.

megies:
@iannesbitt just for future reference, you should work on branches in your fork, your master is now out of sync with the main repo master branch.

iannesbitt:
> @iannesbitt just for future reference, you should work on branches in your fork, your master is now out of sync with the main repo master branch.

Ok. Noted. Thanks for the quick responses.

Issue #2564, opened 2020-02-21
adjust to setuptools removing "Feature"


megies:
setuptools is removing `Feature` which we use in `setup.py`.

e.g. https://ci.appveyor.com/project/obspy/obspy/builds/30961775/job/b6ah67ssmapinays?fullLog=true#L506
```
setup.py:586: DistDeprecationWarning: Features are deprecated and will be removed in a future version. See https://github.com/pypa/setuptools/issues/65.
  require_features=('system-evalresp', 'system-libmseed')
setup.py:574: DistDeprecationWarning: Features are deprecated and will be removed in a future version. See https://github.com/pypa/setuptools/issues/65.
  setuptools.Feature.__init__(self, *args, **kwargs)
C:\Miniconda-x64\envs\test\lib\site-packages\numpy\distutils\numpy_distribution.py:16: DistDeprecationWarning: Features are deprecated and will be removed in a future version. See https://github.com/pypa/setuptools/issues/65.
  Distribution.__init__(self, attrs)
```

ThomasLecocq:
@megies done in 1.2.1 and 2.0.0 is rebased, right ?

megies:
Well the quick fix is in 1.2.1, but we might want to reimplement this mechanism again. Not sure if anybody actually uses that mechanism though??

megies:
@QuLogic I know you are packaging libmseed separate from obspy for fedora.. so you're really the only person I could think of that might be using this mechanism.. no idea really, you have an opinion on whether we need to reimplement this? I'm tempted to just close this, so far I have no indication that anybody is using this feature whatsoever

QuLogic:
We could patch it downstream, I guess. Though I currently have the annoying contention that libmseed 3 is out, but ObsPy doesn't yet support it. If we want to keep this in, we could do like Matplotlib, which reads the setting from a `setup.cfg`.

megies:
Assuming this is obsolete if nobody needed this in the meantime, so closing

PR #2577, opened 2020-03-09
Sunset py2

megies:
### What does this PR do?

WIP Removes all Python 2 compatibility from the code base

### Why was it initiated?  Any relevant Issues?

Next major release after 1.2.0 will be Python 3 only.

**No PRs are to be merged into master before this one**

Supersedes #1613 

Covers most important action items in #2228, leaving open for now to track optional things to do

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just remove the space in the following string after the + sign: "+ DOCS"
- [ ] If any network modules should be tested for the PR, add them as a comma separated list
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:"
      (you can also add "ALL" to just simply run all tests across all modules)
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are be covered via new tests.
- [ ] Any new or changed features have are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .

PR #2578, opened 2020-03-09
Fix for Feature class removal from setuptools v46

gely:
The setuptools.Feature class was removed from setuptools v46.0.0.  For now, detect that case and disable the --with-system-libs option. A better solution would find an alternative to setuptools.Feature.

Sorry, I haven't followed the PR checklist. This is just a quick fix you may want to use or not.

EDIT: I see there is already an issue: https://github.com/obspy/obspy/issues/2564

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -565,7 +565,7 @@ def export_symbols(*path):
 
 # adds --with-system-libs command-line option if possible
 def add_features():
-    if 'setuptools' not in sys.modules:
+    if 'setuptools' not in sys.modules or not hasattr(setuptools, 'Feature'):
         return {}
 
     class ExternalLibFeature(setuptools.Feature):


megies:
Thank you for the quick fix, that will be useful to get valid CI results until we get a real fix for this (if anybody even uses this feature).

megies:
I've cherry-picked this into #2577 which is the big cleanup PR right now. Thanks!

krischer:
@megies How open are you to do a quick `1.2.1` release? This issue here prevents installing ObsPy in an up-to-date clean Python installation as of yesterday without downgrading `setuptools` first.

`1.2.1` would be `1.2.0` and this one commit here.

megies:
Yeah I guess we have to, couldn't they release the new setuptools 2 days earlier? How unlucky can you get ^^

Screenshot from 2020-03-10 17-13-33
Screenshot from 2020-03-10 17-13-25

ThomasLecocq:
should we reopen this then ?

megies:
@gely thanks again and we'll add you to Contributors list if you want to :-)

gely:
@megies No obligation for half a line of code, but sure, I would be happy to be listed if you like. Thanks for obspy--a great tool!

megies:
It's alright, we like to acknowledge people that proactively help out, especially if it's a PR that we can just merge as is. :tada:

PR #2580, opened 2020-03-10
Add a kwarg for plotting section plots within axes objects

shicks-seismo:
### What does this PR do?

Allows axes objects to be passed to `plot(type="section")`. This can be used for example to make multiple subplots containing section plots. I have added a kwarg called `return_ax` to plot(). This is set to False as default. 

Currently `plot(type="section")` only allows figure objects to be passed.

Example: 
```
from obspy import read
st = read()
print(st)
st[0].stats.distance = 100
st[1].stats.distance = 200
st[2].stats.distance = 300
# This is an example of passing an axis object to fig
ax = plt.subplot(111)
st.plot(type="section", fig=ax, return_ax=True, orientation="horizontal")
ax.axvline(3)
ax.axhline(0.15)
plt.show()
# This is an example showing that passing a figure object still works
fig = plt.figure()
st.plot(type="section", fig=fig, orientation="horizontal")
plt.show()
```

### PR Checklist
- [ x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ x] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.


megies:
Hmm.. not sure, I think I would leave `fig` keyword as is and rather add another optional `axes` keyword, taking a list of `Axes` objects (in case different types of plots might use more than one `Axes`, or if not just use `ax` keyword.

shicks-seismo:
Ok, fair enough. I've now replaced with the `sect_ax` kwarg which is the name of the axes object. This defaults to None.

obspy/imaging/waveform.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, **kwargs):
         self.sect_offset_min = kwargs.get('offset_min', None)
         self.sect_offset_max = kwargs.get('offset_max', None)
         self.sect_dist_degree = kwargs.get('dist_degree', False)
-        self.return_ax = kwargs.get('return_ax', False)
+        self.sect_ax = kwargs.get('sect_ax', None)
```
megies: Waveform Plotting object is a monster from the dark, early days of ObsPy, so I guess one more option to the gazillion in there makes no difference. The new option should be documented though in `Stream.plot()` docstring. Also `sect_ax` will be a single `Axis` object looks like? In that case I would also do this maybe:

```python
if self.sect_ax is not None:
    self.fig_obj = self.sect_ax.figure
    msg = "Option 'sect_ax' overrides option 'fig'."
    warnings.warn(msg)
```

Issue #2574, opened 2020-03-05
Add calc_normalization_factor to PolesZeros


WayneCrawford:
It seems to me that the `PolesZerosResponseStage` class should have a `calc_normalization_factor()` method.  The `__init__` function allows us to specify a normalization_factor, but I think this value could be automatically calculated from the poles, zeros, and the normalization frequency.  I must be missing some subtlety, otherwise it would already have been implemented, but anyway here's the code I use to calculate the normalization factor (except for z-transforms).  If validated, it could be used by users to calculate/verify their normalization factor, and by obspy to verify that a provided normalization factor is reasonable:

Regards
@WayneCrawford

`

    def calc_normalization_factor(self):
        """
        Calculate the normalization factor for given poles-zeros

        The norm factor A0 is calculated such that
                           sequence_product_over_n(s - zero_n)
                A0 * abs(------------------------------------------) === 1
                           sequence_product_over_m(s - pole_m)
        for s_f=i*2pi*f if the transfer function is in radians
                i*f     if the transfer funtion is in Hertz
        """
        if not self.normalization_frequency:
            return None

        A0 = 1.0 + (1j * 0.0)
        if self.transfer_function_type == "LAPLACE (HERTZ)":
            s = 1j * self.normalization_frequency
        elif self.transfer_function_type == "LAPLACE (RADIANS/SECOND)":
            s = 1j * 2 * m.pi * self.normalization_frequency
        else:
            print("Don't know how to calculate normalization factor "
                  "for z-transform poles and zeros!")
            return False
        for p in self.poles:
            A0 *= (s - p)
        for z in self.zeros:
            A0 /= (s - z)

        A0 = abs(A0)
        return A0

`

megies:
@aringler-usgs what's your take on this? It's pretty late for 1.2.0 but it seems like a small enough thing to consider maybe adding it to 1.2.0 still. If so this will need tests ASP, though.

Comments on the above patch
 - avoid abbreviation in method name -> `calculate_normalization_factor`
 - the `else` branch should be a `NotImplementedError` instead of the `print`
 - variable names lower case so -> `a0`

aringler-usgs:
@megies this seems very valuable, but as you say, it might be a bit late for 1.2.0.  I could write some test cases, but likely not until late next week.  Which might delay 1.2.0.

I do see value in this addition.

megies:
Ok, I think this will ahve to come in after 1.2.0 then

Issue #2379, opened 2019-04-11
Evalresp 5.0.0


krischer:
There is a new evalresp out there and it seems to have a proper API this time around. We should investigate using it.

https://github.com/iris-edu/evalresp/releases

I'm not sure if #1718 or the new evalresp is the better choice but either is a fairly large but necessary undertaking.

megies:
Ugh. Loads a work.

aringler-usgs:
Does someone have a feel for how much work this is?  Based on the above comments it sounds like it isn't simple.

I think the following: iris-edu/stationxml-seed-converter/issues/42 could quickly introduce a lot of broken code for people.  I know it is starting to cause us problems.

If you go with #1718 I could help write some test cases.

chad-iris:
At IRIS we are discussing potential mitigation to the issue @aringler-usgs linked to.  But that is only the IRIS converter and I would expect different (proper) unit names and symbols to become more common in the StationXML world generally.  We will keep evalresp up to date as fast as we can during a transition from SEED units to SI units, but, obviously, that requires using new releases.

krischer:
An initial implementation of this or of #1718 would not be that much work I think. Testing and making sure it does not break anything would though - one would need to make sure it does the right thing, independent whether or not the response information originates from a SEED, a RESP, or a StationXML file. 

Ideally ObsPy's parsers would make sure that any necessary transformation happens but I don't see how that can be guaranteed so the only feasible way to test this is to run end-to-end (from file -> response) tests and compare against some reference solution for a significant fraction of all files out there.

We have been doing this for the current implementation here (https://github.com/obspy/sandbox, stationxml_test and the the_great_response_test) so this could be repeated.

In terms of #1718 vs upgrading to evalresp 5 I'd personally slightly prefer #1718 not because I wrote it but because it already almost works with quite little code and because the signal processing functions in numpy/scipy are battle tested by a much broader community. Also I think another response computation code aside from evalresp would be beneficial for the community. While the SEED standard is very extensive it is complicated and there is not a straightforward way of telling if a response has been computed correctly aside from comparing against evalresp which at one point of course becomes a bit of a chicken-egg problem.

PR #2592, opened 2020-03-26
Evalresp update

amkearns-usgs:
### What does this PR do?

This is based on the proof of concept in pull request #1718 to remove evalresp as a necessary dependency from obspy. While the methods to interact with evalresp still exist in the code, we are hoping that these new methods will be as good or better means of getting response information from metadata and possibly allow for deprecation of evalresp interactions in the future.

We have also added a method and test case for calculating the stage_gain for pole/zero stages based on code shared in an open issue (see below).

Unfortunately we are not able to get full matching with the provided test cases; the initial response-generating code was not complete and did not include methods for FIR or gain stages, for example. While we have added additional code we also find that there are precision issues that affect our results relative to the code from which our results have been derived.

The PZ stage gain calculation is derived from issue #2574. The removal of evalresp as a dependency was discussed in issue #2379 and (as noted above) pull request #1718. 

### PR Checklist
- [*] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [*] This PR is not directly related to an existing issue (which has no PR yet).
- [-] If the PR is making changes to documentation, docs pages can be built automatically.
- [-] If any network modules should be tested for the PR, add them as a comma separated list
- [ ] All tests still pass.
- [*] Any new features or fixed regressions are be covered via new tests.
- [*] Any new or changed features have are fully documented.
- [*] Significant changes have been added to `CHANGELOG.txt` .
- [-] First time contributors have added your name to `CONTRIBUTORS.txt` .


krischer:
I'm personally quite happy to see  some work on this 👏 

Unfortunately some things got messed up in the git history of this branch. Can you rebase this on the latest master or fix the problem in some other way?

megies:
@krischer it looks like this is based on this branch in #2577 which makes a lot of sense and there only seem to be two empty merge commits at the start which should go away with a rebase, so only a very minor rebase might be needed and probably no conflicts to be expected.

We need to get #2577 finalized and merged ASAP to clear up the confusion a bit..

amkearns-usgs:
Yes , that's correct. I don't believe anything we do relies on anything tied to python 2 so we figured we would get this future proofed.

Right now our biggest concern is that the test cases in the PR this code is derived from use a stricter relative tolerance than our results are capable of providing in some cases. Individual stages can get under 1E-5 relative but not the responses as a whole necessarily.

aringler-usgs:
I think @amkearns-usgs just rebased it to remove the sunset py2 issues.  What else would you like from us?

megies:
@aringler-usgs it might take a couple weeks til i can look at this, just a heads up

amkearns-usgs:
I've gone ahead and merged in this pull request: https://github.com/amkearns-usgs/obspy/pull/4

My thought is that while most of the scipy code is in C and probably well-optimized (possibly better than evalresp), it's probably a good idea to add an interpolation step for efficiency's sake; response curves are rather predictably smooth in my experience.

amkearns-usgs:
Working from the response test code (https://github.com/obspy/sandbox/tree/master/the_great_response_test_scipy) one of the responses is failing specifically because the phase is off by 180 degrees (pi rad). 1.57 is expected but -1.57 is gotten.
obspy afi resp error

This is with the AFI.xml response specifically, as far as I can tell.

krischer:
> I've gone ahead and merged in this pull request: [amkearns-usgs#4](https://github.com/amkearns-usgs/obspy/pull/4)
> 
> My thought is that while most of the scipy code is in C and probably well-optimized (possibly better than evalresp), it's probably a good idea to add an interpolation step for efficiency's sake; response curves are rather predictably smooth in my experience.

Very much yes from my side! Interesting that the interpolation happens on amplitude and phase and not directly on the complex numbers but I guess this makes sense because the curve is smooth in these two.

Two possible improvements/discussion point:

1. I think instead of having a boolean flag `fast` it would be useful to specify the maximum number of frequencies to calculate it at. This would allow tuning it to the problem at hand. The current 10k is likely a sane default.
2. I don't really have a good feeling for this but maybe computing the original response points using `np.logspace()` might be advantageous?

Thanks a lot for this!

CC @flixha

amkearns-usgs:
> 2\. I don't really have a good feeling for this but maybe computing the original response points using `np.logspace()` might be advantageous?

This is probably a good idea; I'm used to working with response plots in logarithmic spaces (though usually when I generate them directly for analysis I normally use a linear frequency space for simplicity's sake). An overly sparse set of frequencies would be worse for the low-frequency end than the high-end.

krischer:
> Working from the response test code (https://github.com/obspy/sandbox/tree/master/the_great_response_test_scipy) one of the responses is failing specifically because the phase is off by 180 degrees (pi rad). 1.57 is expected but -1.57 is gotten.

That particular response is from a temperature sensor.

A guess of mine: This particular FIR filter is symmetrical in its coefficients and evalresp has this `check_sym()` function that internally converts the FIR representation to something else if it detects this symmetry. In theory I guess this should result in the same end result but evalresp has a bunch of oddities so I don't really know.

flixha:
@krischer  @amkearns-usgs  I implemented what you commented here regarding the interpolation in this PR: https://github.com/amkearns-usgs/obspy/pull/5 :

>  1. I think instead of having a boolean flag `fast` it would be useful to specify the maximum number of frequencies to calculate it at. This would allow tuning it to the problem at hand. The current 10k is likely a sane default.

I changed the `fast=True` to a value `n_frequencies_limit_for_interp=10000` now. If you can think of a better, and self-explanatory name of that variable, please let me know ;-)


>  2. I don't really have a good feeling for this but maybe computing the original response points using `np.logspace()` might be advantageous?

I changed it to logspace. I guess in my head I was mostly concerned how well logspace would be able to represent the high-amplitude response, e.g., due to FIR filter around the Nyquistfrequency (see, e.g., [the phase-response here](http://service.iris.edu/irisws/evalresp/1/query?net=IU&sta=ANMO&loc=10&cha=HHZ&time=2019-06-07T20:00:01&width=800&height=600&output=plot&annotate=true). But I think as you both noted the low frequencies may be quite inaccurately represented with linspace. And with a default of n=10000, logspace should be good enough also at the high frequencies.

> Very much yes from my side! Interesting that the interpolation happens on amplitude and phase and not directly on the complex numbers but I guess this makes sense because the curve is smooth in these two.

This also replicates how interpolation was done in evalresp. But it's a good point to check, as the interpolation becomes the computationally most expensive call once interpolation is chosen. In `InterpolatedUnivariateSpline(resp_frequencies, amp, k=3)(frequencies)` I changed to `k=2` to best replicate the original interpolation chosen in evalresp, which was CubicSpline interpolation.

obspy/core/inventory/response.py
```diff
@@ -177,7 +177,13 @@ def __str__(self):
     def _repr_pretty_(self, p, cycle):
         p.text(str(self))
 
-    def get_response(self, frequencies):
+    def get_response(self, frequencies, fast=True):
```
megies, line 181: might be cleaner to just use `**kwargs`

obspy/core/inventory/response.py
```diff
@@ -598,7 +640,17 @@ def get_response(self, frequencies):
...
-        final_resp = np.empty_like(resp)
+
+        # If "fast", then interpolate the amplitude and phase onto the
+        # originally requested frequencies.
+        if len(frequencies) > 10000 and fast:
+            amp = scipy.interpolate.InterpolatedUnivariateSpline(
+                    resp_frequencies, amp, k=2)(frequencies)
+            phase = scipy.interpolate.InterpolatedUnivariateSpline(
+                    resp_frequencies, phase, k=2)(frequencies)
+            final_resp = np.zeros_like(frequencies) + 0j
```
megies, line 664: I think this is duplicated, better define a helper function for the interpolation part

```python
def _interpolate(resp_frequencies, amp, phase, frequencies):
    ....
```

flixha: In https://github.com/amkearns-usgs/obspy/pull/5 I have added a helper function  `_check_response_interpolation` to do the check whether interpolation is necessary, and to provide the log-spaced-values. I didn't include the interpolation-call itself into the helper function, because it's only twice used on both phase and amplitude, and once on only the amplitude - so I think in the current version saves us the most lines of code and is easiest to read. But let me know if you prefer it differently.

megies: I only saw know that you do it separately, so I would still refactor it, duplicated code is always good to avoid (function naming is just a proposol, other name is fine)

```python
def _interpolate(resp_frequencies, quantity, frequencies):
    ...

amp = _interpolate(resp_frequencies, amp, frequencies)
phase = _interpolate(resp_frequencies, phase, frequencies)
```

Also, I see linspace used there.. so linear spacing? Should be logspace instead?

flixha: Ok, no problem, added the interpolate-helper function as well.
Regarding the linspace, thanks for catching that - it shouldn't have been there and must have gotten back in when solving some merge conflicts..

I did some more tests for the linspace vs logspace-issue. As expected, for `FIRResponseStage` the logspace introduces a bit more of a difference, while for `PolesZerosResponseStage`, using linspace introduces some more - but they are within very reasonable limits.
Even though I did quite some tests, it's probably still good to also test the interpolation on a larger set of responses / seismograms. Sometimes I got extremely small relative differences after correcting with calculated vs. interpolated responses (e.g., 10^-13), and sometimes I got still small, but bigger differences (relative error: ~10^-5). May be related to values very close to 0 (--> waterlevel etc.).
It appears to me that

megies: Without having looked into it myself, if both linear and log-spacing isn't ideal.. there are other spacing strategies out there as well. I don't have a good resource, but I think these matplotlib axes scale graphs give a good idea (e.g. something like logit, which looks like log spacing but with dense points at both ends): https://matplotlib.org/3.1.0/gallery/scales/scales.html

aringler-usgs:
@krischer @amkearns-usgs I got through about 6600 response files.  These appear to pass if I restrict to seismic data (VH*, LH*, BH*, HH*, EH*).  

I can't get the VK1 channel to run in evalresp 5.0.0.  I think it might have a strange call in the wrapped code.  The issue is coming from stage 3.

I will run through the rest of the responses for seismic data and then work on the various weather responses.  My guess is that these are somewhat of "edge cases."

amkearns-usgs:
Hi, it's been a while since I've commented on this?

Over the past couple months @aringler-usgs and I have been going back to this and trying to see what we can do with respect to matching up evalresp calculations vs. our new methods. Right now we see with most of the responses we've compared against that we get very good matchup between the amplitudes in obspy and evalresp. However, there are odd jumps in phase often near the end of the evalresp curve that we're not certain the justification for, or if they're even correct. They appear to be tied to smoothing/scaling parameters (which includes an unwrap operation).

I'm far more of a programmer than I am a seismologist so I don't know enough to trust my own intuition on this, but those evalresp phase curves don't actually look right to me. Note that I've been trying to do comparisons (inc. source analysis) with evalresp 3.3.3 specifically since to my understanding that's the one that obspy has been using; I believe that the code was refactored heavily in the later releases to the point that the processing steps look very different (or at least are in different parts of the codebase, making direct comparisons between routines' affects on the results a little difficult).

@aringler-usgs would you say that's a reasonable summary of our current status, or is there anything else I've forgotten to add.

obspy/core/inventory/response.py
```diff
@@ -577,7 +789,7 @@ def __init__(self, stage_sequence_number, stage_gain,
                  decimation_offset=None, decimation_delay=None,
                  decimation_correction=None):
         self._symmetry = symmetry
-        self.coefficients = coefficients or []
+        self._coefficients = coefficients or []
```
flixha, line 792: The code snippet below demonstrates a small bug in the current implementation here:
```python
from obspy.core.inventory import read_inventory
inv = read_inventory(
    'http://service.iris.edu/irisws/resp/1/query?net=GE&sta=KBS&loc=10&cha=BHZ'
    '&starttime=1999-01-01T00:00:00&endtime=2000-01-01T00:00:00')
inv.write('KBS.stationxml', format='STATIONXML')
```

This will fail with an `AttributeError` because the values `FIRResponseStage._coefficients` are not initiated as `FloatWithUncertainties`, but just as `float`. I don't fully understand what are the advantages of the implementation with `coefficients` and `_coefficients` as properties, so maybe you can comment on how the setters here work / are supposed to work @amkearns-usgs?
```python
# should be FloatWithUncertainties, but is just float:
inv.networks[0].stations[0].channels[0].response.response_stages[4].coefficients
```
It may be possible that a similar issue could also affect other Stages and their setters if the setters write directly to an attribute with `_attr` when the setter is defined for `attr`, but I have only encountered the problem right here for now.

flixha (commented):
I just noted a small bug related to the initiation of coefficients for the FIRResponseStage (and possibly other stages). I marked it in a review below where I think it's coming from.

Unfortunately, I can't comment directly on your recent question about the differences in response computation..

ThomasLecocq:
@amkearns-usgs @aringler-usgs we're giving a new push to make a new release of ObsPy, so I wanted to check with you if you feel like we could move forward here? If yes, please rebase & ping back :-)

amkearns-usgs:
> @amkearns-usgs @aringler-usgs we're giving a new push to make a new release of ObsPy, so I wanted to check with you if you feel like we could move forward here? If yes, please rebase & ping back :-)

Sure. Just to ensure I don't start breaking the commit history, I should be rebasing onto the master branch, correct?

ThomasLecocq:
Yes, rebase on master and force-push if possible, so your commit history is clean. There will probably be merge conflicts with the changelog at least, but we'll solve that

filefolder:
hi all, can we start again with this one? may need another rebase since 1.3 as the error logs are missing now. 

happy to help test / etc

megies:
I don't have a good idea of how the status is here. Sounds like it's almost ready, but with things like this the devil is in the detail and.. response calculations are like one of the most important parts of obspy, so anything touching it has to go through the most rigorous testing. And trying to figure out why response information XYZ is behaving slightly differently takes a lot of time if you look at enough responses.
When implementing the current status of response calculations we had a monster test case, basically downloading gigabytes of instrument data from IRIS and comparing results of standalone evalresp with our implementation re-using evalresp. I can't find that test setup anymore, @krischer you still know where that is (I think it was either in a separate repository or in some branch somewhere, but I can't find it).

liamtoney:
Hi all, I stumbled across this PR and also issue #2574 when I was looking into contributing an $A_0$ normalization factor calculation method to the `PolesZerosResponseStage`. My code looks very similar to that issue's, unsurprisingly.

Anyhow, I think such functionality would be useful separate from the much larger task this PR is [nobly] working towards. Would the ObsPy devs be open to a PR solely focused on addressing #2574? I've got the code and some basic tests written up for rad/s and hertz units (many thanks to [this article](https://docs.fdsn.org/projects/stationxml/en/latest/response.html) for guidance!).

Commenting here since this thread seems relatively more active. Ping @aringler-usgs!

liamtoney:
> a PR solely focused on addressing #2574

Two notes now that I've browsed the code for this PR:
- It looks like the tests I've written for this method are similar to those in this PR. 
- The addition of automatic calculation of the normalization factor if it doesn't exist (having it default to `None` rather than `1.0`) seems worthwhile to keep in the proposed new specific PR.

aringler-usgs:
@liamtoney since this PR has kind of stalled, maybe what you are suggesting is a good first direction.  If I recall, evalresp does a few somewhat strange things at the high-frequency end of the response.  It was unclear to us what the "correct" way to resolve these discrepancies are.  Making various changes, without knowing the purpose, simply to match evalresp didn't seem ideal.  However, you could also change a fairly fundamental function.

I agree the smaller [](https://github.com/obspy/obspy/issues/2574)#2574 would be a useful addition.

PR #2598, opened 2020-04-07
Fix azimuth read type in nlloc read

shicks-seismo:
ObsPy V1.2.0, OSX

When reading in a NonLinLoc .hyp event file, the value assigned to `arrival.azimuth` is the "RAz" parameter (Col. 23), which is according to the NLL manual:

> Ray take-off azimuth at maximum likelihood hypocenter or expectation in degrees CW from North

not the epicentral distance "SDist" :

> Maximum likelihood or expectation hypocenter to station epicentral azimuth in degrees CW from North

which is in the preceding column (Col. 22) of the hyp file.
When using the `arrival.azimuth` object to plot traces against azimuth, I was getting very different results from those obtained with `gps2dist_azimuth`

So I have made this PR to read the epicentral azimuth, "SDist" instead.

### DIFF
--- a/obspy/io/nlloc/core.py
+++ b/obspy/io/nlloc/core.py
@@ -319,7 +319,7 @@ def _read_single_hypocenter(lines, coordinate_converter, original_picks):
         phase = str(line[4])
         arrival.phase = phase
         arrival.distance = kilometer2degrees(float(line[21]))
-        arrival.azimuth = float(line[23])
+        arrival.azimuth = float(line[22])
         arrival.takeoff_angle = float(line[24])
         arrival.time_residual = float(line[16])
         arrival.time_weight = float(line[17])


megies:
Hmm.. indeed QuakeML states

```
Arrival > azimuth: Azimuth of station as seen from the epicenter.
```

Personally for me it feels like losing information if we just put the plain geometrical azimuth in there, like you said it can always be easily computed from station location. The take-off azimuth can not be reconstructed afterwards though.

Other opinions?

shicks-seismo:
True although I don't get why the takeoff azimuth given by nlloc is equal to SAz +/-180degrees? Is it because RAz is used for plotting the points (e.g. polarities) on a stereonet projection?

megies:
Oh so its about a 180 degree flip? I thought it was about minor differences between ray take off and great circle azimuth..

megies:
RAz and SDist should be pointing in same direction from quoted above NLLoc docs above, no?

shicks-seismo:
Yep, it's a 180 degree flip - which I realised after posting the initial PR - see example screenshot below.
It was also my understanding that they should be pointing in the same direction, but maybe it's just for stereonet plotting convenience?

Screenshot 2020-04-07 at 10 52 20

In any case, we should probably go ahead with the mod

shicks-seismo:
Cc. @alomax who might be able to answer the question on the 180 deg flip

alomax:
> Cc. @alomax who might be able to answer the question on the 180 deg flip

SAzim is the azimuth from the hypocenter to the station
RAz is the ray take-off azimuth from the hypocenter back towards the station
With a 1D model, RAz is same as SAzim, with a 3D model there may be a small difference

This is what I see in some NLL output files I just checked.  What is the NLL control configuration for the output where you see a 180deg flip?

shicks-seismo:
@alomax - thanks. Hmmm, the plot thickens.
Here is my control file: [nlloc.in.txt](https://github.com/obspy/obspy/files/4443569/nlloc.in.txt)

I just looked at the NLL6.00samples example cases that you provide - they look okay - SAzim is close to RAz. Maybe something to do with the coordinate projection???

alomax:
Can you provide uk.nor so I can take a look at running here?

ThanksQ

> @alomax - thanks. Hmmm, the plot thickens.
> Here is my control file: [nlloc.in.txt](https://github.com/obspy/obspy/files/4443569/nlloc.in.txt)
> 
> I just looked at the NLL6.00samples example cases that you provide - they look okay - SAzim is close to RAz. Maybe something to do with the coordinate projection???

shicks-seismo:
Here you go
[uk.nor.txt](https://github.com/obspy/obspy/files/4443896/uk.nor.txt)

alomax:
> Here you go
> [uk.nor.txt](https://github.com/obspy/obspy/files/4443896/uk.nor.txt)

That gives agreement between SAzim and RAz.
I am using NLLoc:v7.00.08(20Aug2019)
image

shicks-seismo:
Hmmmm … I just double-checked using the pre-compiled Mac OSX binaries (http://alomax.free.fr/nlloc/soft7.00/tar/NLL7.00_bin_x86_64-apple-darwin14.tgz) and I am still having the same issue.

alomax:
Very odd! I ran with the the same pre-compiled binary (downloaded from the link) and still get the same azimuths for SAzim and RAz.  

Are you doing further processing after running NLLoc, or do you see the difference directly in vinti.20170804.144349.grid0.loc.hyp?

shicks-seismo:
No, there is no further processing that I am doing ... the difference is directly in the raw .hyp file.

shicks-seismo:
Right, so I think I've found the problem ... it looks I was using either Vel2Grid and/or Grid2Time with an older version than NLLoc, this seemed to result in SAzim != RAz. Since this was more related to NLL than ObsPy, and assuming other people have a sensible NLL installation, I'll close this PR. Thanks for the support and sorry for the silly error on my side.

alomax:
Thanks for the update. I took a quick look and did not find an explanation for this behavior in the NLL code or change notes.

PR #2601, opened 2020-04-15
Fix deepcopy edge cases

### LINKED ISSUE #2600: Trace.copy: Using deepcopy causes derived Stats keys setting parent keys with computed values (leading to wrong values in edge cases)
Hi everybody,

First, I hope all of you are happy and healthy.

Second, I've got a puzzling issue when working with highly-sampled data:

-  ObsPy version: 1.2.1 (anaconda)
-  Python version: 3.7
-  Arch Linux

Working with sampling rates >= 100000. Hz produces unexpected behavior.

```
tr = Trace(data=np.ones(100000), header={'sampling_rate': 100000.})
print(tr.stats.sampling_rate)
100000.0
print(tr.copy().stats.sampling_rate)
99999.99999999999
```
I think this is a result of floating point ["representation error"](https://docs.python.org/3/tutorial/floatingpoint.html), whenever the Trace AttribDict needs to be re-evaluated, [specifically where `delta` is involved](https://github.com/obspy/obspy/blob/88687a146a7c3ca8f35c608db2243cf5fed6813c/obspy/core/trace.py#L177-L182).

Not sure how to go about addressing this though...

trichter:
Fixes #2600 and the same issue for pickled objects

deepcopy uses the `__deepcopy__` method or if not present `__getstate__`, `__setstate__` methods.
I added custom `__getstate__/__setstate__` methods in Stats and removed the `__deepcopy__` method in Attribdict. 

Update: I also remove `__getstate__/__setstate__ ` in AttribDict. I think they are not necessary.

### DIFF
--- a/obspy/core/util/attribdict.py
+++ b/obspy/core/util/attribdict.py
@@ -106,15 +106,6 @@ class AttribDict(compatibility.collections_abc.MutableMapping):
     def __delitem__(self, name):
         del self.__dict__[name]
 
-    def __getstate__(self):
-        return self.__dict__
-
-    def __setstate__(self, adict):
-        # set default values
-        self.__dict__.update(self.defaults)
-        # update with pickle dictionary
-        self.update(adict)
-
     def __getattr__(self, name, default=None):
         """
         Py3k hasattr() expects an AttributeError no KeyError to be
--- a/obspy/core/util/attribdict.py
+++ b/obspy/core/util/attribdict.py
@@ -131,11 +131,6 @@ class AttribDict(compatibility.collections_abc.MutableMapping):
     def copy(self):
         return copy.deepcopy(self)
 
-    def __deepcopy__(self, *args, **kwargs):  # @UnusedVariable
-        ad = self.__class__()
-        ad.update(copy.deepcopy(self.__dict__))
-        return ad
-
     def update(self, adict={}):
         for (key, value) in adict.items():
             if key in self.readonly:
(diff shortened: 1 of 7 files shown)


obspy/core/util/attribdict.py
```diff
@@ -131,11 +131,6 @@ def __getattr__(self, name, default=None):
     def copy(self):
         return copy.deepcopy(self)
 
-    def __deepcopy__(self, *args, **kwargs):  # @UnusedVariable
```
trichter, line 134: Is this `self.__dict__.update(self.defaults)` in `__setstate__` necessary? It's already in `__init__`. Are these `__get/setstate__` methods necessary?

megies: No idea, to be honest.. but I guess if tests keep passing we might be OK to remove them? dunno

megies:
Looks good overall, although I never used these `__getstate__` mechanics, so no idea if there's any other implications that we're not seeing/testing right now..

Changelog please @trichter :-)

megies:
Would be good to have other opinions, like I said, never used `__get/setstate__` before

trichter:
> never used __get/setstate__ before

Neither did I. The doc of [deepcopy](https://docs.python.org/3/library/copy.html#copy.deepcopy) refers to pickle and this practice.

megies:
Alright, tests pass and no other comments/reviews, so..  :man_shrugging:

PR #2611, opened 2020-05-06
EIDA Routing Client creating mass requests when asking for non-existing data

megies:
Looks like we have a bug in EIDA Routing Client, see EIDA/userfeedback#56.

  -  ObsPy version 1.2.1


megies:
+TESTS:clients.fdsn

megies:
This should be backported to `maintenance_1.2.x`, I think, as it might hit a significant amount of people.

PR #2614, opened 2020-05-14
remove_response with ResponseList type response

megies:
In some cases when working with industry, a Frequency-Amplitude-Phase table is the only information on Response for a given sensor/datalogger combination.
These calibration sheets usually only cover a relatively narrow set of frequencies, e.g. ~20 discrete frequencies, in this case from 0.5 to 80 Hertz.

Currently, ObsPy will error out in basically all scenarios when trying to `remove_response()` with such a `ResponseList` only instrument response, due to this piece of code: https://github.com/obspy/obspy/blob/1440229af113d30933ca808a7a2b1daaa1eb6af6/obspy/core/inventory/response.py#L1375-L1382

When commenting out that check, evalresp actually computes an output and it seems sensible, when using appropriate means to stabilize the deconvolution (in this case: `pre_filt=0.2, 0.5, 50, 80`).

I would propose to convert that Exception into a warning, but would like to hear comments / other opinions, especially from people that are more familiar with evalresp internals.

@krischer, maybe @aringler-usgs ?

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,10 @@ Changes:
  - obspy.core:
    * Fix wrong values in Stats object after deepcopy or pickle of Stats object
      for edge cases (see #2601)
+   * when calling evalresp with a ResponseList type response that does not
+     cover the full frequency range, show a meaningful warning instead of
+     raising an exception and continue nevertheless with extrapolated values
+     (see #2614)
  - obspy.clients.fdsn:
    * EIDA routing client: fix an issue that leaded to a request of *all* EIDA
      data when requesting an invalid, out-of-epochs time window for a valid
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1374,12 +1374,15 @@ class Response(ComparingObject):
 
                 if min_f < min_f_avail or max_f > max_f_avail:
                     msg = (
-                        "Cannot calculate the response as it contains a "
+                        "The response contains a "
                         "response list stage with frequencies only from "
                         "%.4f - %.4f Hz. You are requesting a response from "
-                        "%.4f - %.4f Hz.")
-                    raise ValueError(msg % (min_f_avail, max_f_avail, min_f,
-                                            max_f))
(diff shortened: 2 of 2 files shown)


krischer:
Yea I guess the spline interpolation makes sense here. One could think about setting `ext="const"` and it will no longer return the spline values (which will go crazy outside the domain) but just the boundary values. Would definitely help keep it stable. The downside is that the interpolated response will no longer be continuous in the first derivative which is maybe kind of dangerous. But probably does not matter if we throw a warning and assume a strict pre-filter is always applied.

megies:
> The downside is that the interpolated response will no longer be continuous in the first derivative which is maybe kind of dangerous

Not sure how it could make a difference, since it's only discrete sampling points in frequency anyway?

> One could think about setting `ext="const"` and it will no longer return the spline values (which will go crazy outside the domain) but just the boundary values

That sounds like a good option. Keeping it constant should in normal cases (of the response falling off at the ends) lead to less overamplification, since we are applying the inverse actually. For the phase we still might want to keep the spline (or some kind of linear-ish) extrapolation though..?

krischer:
> Not sure how it could make a difference, since it's only discrete sampling points in frequency anyway?

That's a good point.

> That sounds like a good option. Keeping it constant should in normal cases (of the response falling off at the ends) lead to less overamplification, since we are applying the inverse actually. For the phase we still might want to keep the spline (or some kind of linear-ish) extrapolation though..?

I guess one could do a manual linear extrapolation at the edges using the two first and two last samples? Still a bit wild but better than nothing.

And none of these changes would change existing code because existing code could never reach it :-)

megies:
> And none of these changes would change existing code because existing code could never reach it :-)

good point :+1:

megies:
Part of me wants this in 1.2.2, but on the other hand it is altering behavior to some extent (although it might not actually hurt anybody, if this case even is encountered by anybody so far..).

Other opinions.. @obspy/maintainers ? see #2655

trichter:
Dealt with in #2988.
I think I would have preferred the const option mentioned in here, but now its already released and with the warning it should be OK.

PR #2617, opened 2020-05-20
clients.neic: Make NEIC client socket blocking

jmfee-usgs:
### What does this PR do?

This removes the call `socket.setblocking(0)` in the NEIC client.

Excessive CPU usage has been observed on systems using this client.

A `setblocking(0)` call is commented out in the seedlink client, but is otherwise not referenced within obspy for any other socket handling.  Using a blocking socket waits for data to be available instead of endlessly looping.

The python documentation states:
> You have (of course) a number of choices. You can check return code and error codes and generally drive yourself crazy. If you don’t believe me, try it sometime. Your app will grow large, buggy and **suck CPU**.
> https://docs.python.org/2/howto/sockets.html#non-blocking-sockets

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.neic"
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,5 +1,9 @@
+maintenance_1.2.x
+=================
+
+Changes:
  - obspy.clients.neic:
-   * Remove neic client socket setblocking(0) call
+   * Make client socket blocking (see #2617)
 
 1.2.1 (doi: 10.5281/zenodo.3706479)
 ===================================
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,6 @@
+ - obspy.clients.neic:
+   * Remove neic client socket setblocking(0) call
+
 1.2.1 (doi: 10.5281/zenodo.3706479)
 ===================================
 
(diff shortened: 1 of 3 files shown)


jmfee-usgs:
@ketchum-usgs : fyi

megies:
I'm not personally using the NEIC client, so I'm totally ready to just merge this without much looking into.

However, this should go either into `maintenance_1.2.x` or `master`. I'll change the base branch of the PR, git should be able to figure it out, since only more (unrelated) commits will pop up in between.

Also.. we'll want a one-liner in the Changelog

megies:
Ok, so this actually needs tweaking, since we added a `with` context around the part of the code this line sits in @jmfee-usgs 

You can either rebase locally and force push after, or just reset your branch to maintenance_1.2.x and then redo the commit. Let me know if anything unclear @jmfee-usgs

jmfee-usgs:
I'll update this, originally had it against master and then switched to the maintenance branch referenced in CONTRIBUTING.md:
https://github.com/obspy/obspy/blame/master/CONTRIBUTING.md#L32

jmfee-usgs:
@megies : I think is should be ready, let me know if this needs any other changes

megies:
+TESTS:clients.neic

will restart Travis so that the neic tests are run on it

megies:
NEIC tests passed, adjusted changelog and will merge now, thanks @jmfee-usgs

PR #2618, opened 2020-05-26
signal: first try at PAZ fitting from FAP info

megies:
### What does this PR do?

Adds functionality to derive Poles-and-Zeros response information from Frequency-Amplitude-Phase type discrete response info (e.g. from calibration at discrete frequencies)

Currently response removal with FAP type response does not work and even with changes in #2614 it might be desirable in some or most cases to convert FAP response to PAZ instead of directly using a FAP response.

### PR Checklist
- [ ] All tests still pass.

CC @aringler-usgs @krischer


megies:
Here's an example use case:

```python
from math import cos, sin, radians
import numpy as np
from obspy.signal.invsim import optimize_paz, _paz_to_freq_resp

frequencies = np.array(
    [0.5, 0.8, 1, 1.25, 1.6, 2, 3.15, 5, 10, 16, 40, 63, 80])
magnitudes = [95.86, 155.48, 182.27, 202.98, 214.98, 217.61, 224.64, 234.46,
              238.78, 239.48, 236.67, 226.75, 216.80]
phases = [101.6, 75.7, 63.0, 50.8, 38.8, 30.5, 20.3, 12.0, 2.6, -2.8, -17.3,
          -29.7, -38.2]
response = np.array([r * (cos(radians(phi)) + 1.0j * sin(radians(phi)))
                     for r, phi in zip(magnitudes, phases)])

misfits = []
num_paz = list(range(30))
for num in num_paz:
    poles, zeros, scale_fac, inverted_response, misfit = optimize_paz(
        frequencies=frequencies, response=response, numpoles=num,
        numzeros=num, num_tries=2)
    misfits.append(misfit)
    print("========================================")
    print(f"Inverted values ({num:d} poles/zeros):")
    print("")
    print("Poles:", poles)
    print("Zeros:", zeros)
    print(f"Scale_fac: {scale_fac:.3g}")
    print(f"Misfit: {misfit:.3g}")

    frequencies_dense = np.logspace(
        np.log10(frequencies[0]) - 1, np.log10(frequencies[-1]) + 1, 200)
    response_dense = _paz_to_freq_resp(
        freqs=frequencies_dense, zeros=zeros, poles=poles,
        scale_fac=scale_fac)

    fig, (ax1, ax2) = plt.subplots(nrows=2, sharex=True)
    ax1.set_title(
        f'Number of Poles/zeros: {num:d} -- Misfit: {misfit:.3g}')
    ax1.semilogx(frequencies, np.abs(response), lw=5, color="blue",
                 label="ground truth response")
    ax1.semilogx(frequencies_dense, np.abs(response_dense), color="red",
                 lw=2, label="response from inverted PAZ")
    ax1.semilogx(frequencies, np.abs(inverted_response), color="red",
                 ls='', marker='o', label="response from inverted PAZ")
    ax2.semilogx(frequencies, np.angle(response), lw=5, color="blue")
    ax2.semilogx(frequencies_dense, np.angle(response_dense), color="red",
                 lw=2)
    ax2.semilogx(frequencies, np.angle(inverted_response), color="red",
                 ls='', marker='o')
    ax1.set_ylabel('Amplitude')
    ax2.set_ylabel('Phase')
    ax2.set_xlabel('Frequency [Hz]')
    ax1.legend(loc=4)
    ax1.grid()
    ax2.grid()
    fig.tight_layout()
    fig.savefig(f'/tmp/optimize_paz_{num:02d}.png')
    plt.close(fig)
fig, ax = plt.subplots()
ax.plot(num_paz, misfits)
ax.set_xlabel('Number of Poles/Zeros')
ax.set_ylabel('Misfit (euclidean distance on complex response)')
fig.tight_layout()
fig.savefig(f'/tmp/optimize_paz_misfits.png')
plt.close(fig)
```

optimize_paz_10
[optimize_paz.pdf](https://github.com/obspy/obspy/files/4681882/optimize_paz.pdf)

Plotting functionaly would definitely have to go into `invsim.py` too, as the misfit measure on the discrete frequencies can not catch problems with overshooting reverberations in between discrete frequency sampling points (see pdf).

Probably in the optimization that needs accounting too.. so maybe add a penalty in the misfit calculation for very curvy results in between sampling points..

So probably still quite a bit to improve, but could be very worthwhile to have something like this in obspy.

dsentinel:
This is fantastic!
We've been doing the same, as we are getting units which only supply FAP's.
I'll give this a run with our responses

aringler-usgs:
@dsentinel would you be willing to share your FAP?  Also, do you know the combination of digitizer and sensor your FAP is from?  Thanks!

dsentinel:
@aringler-usgs, You may already be familiar with our work at PASSCAL as @akram-pic is the one working on this, and I think she's been using your Levenberg-Marquardt algorithm.

I'm happy to share and don't see a problem, but I should check first to ensure I'm not breaking policy or distributing proprietary or evaluation only type information. We have to be respectful of the manufacturers that share information with us.

tychoaussie:
I have been struggling with this subject for a year now, and have generated a pole-zero estimator from traditional calibration parameters for the legacy Soviet analog stations. I am currently matching up some .fap files to poles and zeros; So I will see if I can lend a hand.

obspy/signal/invsim.py
```diff
@@ -763,6 +764,76 @@ def estimate_wood_anderson_amplitude_using_response(response, amplitude,
...
+    return poles, zeros, scale_fac
+
+
+def _response_misfit(reference, other):
+    """
+    Calculate a misfit measure from two complex response vectors
+    """
+    # just the simple euclidean distance for now. tweaking this might improve
+    # the fitting maybe..
+    misfit = scipy.spatial.distance.euclidean(reference, other)
```
krischer, line 791: If this converges it is probably hard to create a better misfit functional - if you'll find a few examples where it does not converge one has to think a bit harder. I think that in particular the phase might sometimes be a bit hard to fit.

obspy/signal/invsim.py
```diff
@@ -763,6 +764,76 @@ def estimate_wood_anderson_amplitude_using_response(response, amplitude,
...
+    misfit = np.inf
+    for result in results:
+        _poles, _zeros, _scale_fac = _unpack_paz(
+            result.x, numpoles=numpoles, numzeros=numzeros)
+
+        _inverted_response = _paz_to_freq_resp(
+            freqs=frequencies, zeros=_zeros, poles=_poles,
+            scale_fac=_scale_fac)
+
+        _misfit = _response_misfit(response, _inverted_response)
```
krischer, line 826: This does not need to be recomputed here. The items in `results` should have it available at the `.fun` attribute. You can also query if it did converge and if none of the results converged, throw an error.

megies:
> We've been doing the same

@dsentinel would be good to see the code, no need to use the "real" values if it might be problematic, you can try it with these or just use some forged ones or obfuscated ones.

@dsentinel @aringler-usgs @tychoaussie To compare different approaches it would be ideal to meld them into one interface structure, like different misfit functions that can be swapped out etc. I'm totally open to adjust the current function/interface layout of this PR to make it happen.

> I don't see the loop over the number of poles/zeros you mentioned in the chat.

@krischer for now the loop over number of poles/zeros lives in the driver Python script quoted above. That part could totally also be integrated into the PR, as some even higher level on top of the current code.
The fitting dosn't work as good for each attempt with a different number of poles/zeros, the above picture is basically the best one. The pdf linked in the comment has all tries in it.

obspy/signal/invsim.py
megies: Well like mentioned above (or at least somewhere) I see a tendency of very curvy response results, that matches on the sampling points but doesn't make much sense, while still having a very low misfit. So I think the euclidean might be a good start but I am pretty sure it needs a penalty if getting to curvy in between sampling points.. maybe add a second misfit criterion that compares to a simple linear interpolation of the original sampling points but with a much denser sampling.

krischer: I think the spurious oscillations could be reduced by just limiting the maximum number of poles and zeros based on the number of initial response curve values. 20 poles and zeros are 80 degrees of freedom which is too much for the given data without some form of regularization. A smoother curve can also be enforced via regularization if so desired.

krischer:
> @krischer for now the loop over number of poles/zeros lives in the driver Python script quoted above. That part could totally also be integrated into the PR, as some even higher level on top of the current code.
> The fitting dosn't work as good for each attempt with a different number of poles/zeros, the above picture is basically the best one. The pdf linked in the comment has all tries in it.

Yea I should really learn to read the initial descriptions carefully 🙈

obspy/signal/invsim.py
```diff
@@ -763,6 +764,76 @@ def estimate_wood_anderson_amplitude_using_response(response, amplitude,
...
+            freqs=frequencies, zeros=zeros, poles=poles, scale_fac=scale_fac)
+        misfit = _response_misfit(response, new_response)
+        return misfit
+
+    results = []
+    for i in range(num_tries):
+        out = scipy.optimize.minimize(
+            fun=minimize,
+            method="BFGS",
+            x0=np.random.random(numpoles * 2 + numzeros * 2 + 1),
```
aringler-usgs, line 813: The starting point is fairly critical.  When I run this on an STS-2 if change the guess to 
`guess = [-0.037, -0.037, -0.037, +0.037, 0, 0, 0, 0]`

where `x0=guess`
COR_good_guess

However, if I leave in the initial guess as random, then I get the following:
`guess = np.random.random(numpoles * 2 + numzeros * 2 + 1)`
COR_random

This makes me think you want to allow for an initial guess.

aringler-usgs:
@megies when I run your example I get cases where the poles have positive real parts.  So the system becomes unstable.  

E.g. the following from your code:
```
Inverted values (4 poles/zeros):

Poles: [ 443.06583520 +18.55689168j   -5.06934957  +1.39415247j
  -58.06117249-149.49157574j -295.75378102+326.76766451j]
Zeros: [   1.08211258  +1.30802766j -655.00899410+347.09262671j
 -193.23494616-357.19525228j  306.55213295 -35.09948853j]
Scale_fac: 84.4
Misfit: 26.6
```
I don't have a good feel for how that could mess up things in other locations.  E.g., for most applications is this okay or could this become problematic if you use these for deconvolution?

We tend to force the sign of the poles.  However, we are starting with an initial model and then perturbing that.  Do you think adding an optional `initial_guess` would help?

It seems like there are two use cases: one where someone has some initial information and wants to pin down the response with a high degree of certainty and a case where you have no information other than the FAP.  

I will try to make some changes to deal with the first case next week.

aringler-usgs:
@tychoaussie I don't know if you have seen this, but I was messing around with a similar problem.  The solution I came up with was to fit the calibration pulses in the time domain.  If you have calibration pulses, then this might be an option.
[srl-2019201.1.pdf](https://github.com/obspy/obspy/files/4697116/srl-2019201.1.pdf)

I apologize for this being a little bit off-topic.

obspy/signal/invsim.py
megies: Yeah, initial guess is where a lot of smart brains could go into, right now its far from ideal with just random numbers

megies:
> We tend to force the sign of the poles. However, we are starting with an initial model and then perturbing that. Do you think adding an optional `initial_guess` would help?

sure thing, providing an initial guess definitely should be an option

tychoaussie:
Hey all,
I will try to get my updated code up tomorrow, but for now, why not take a
look at my program, PZCalc? It has a more sophisticated pole/zero estimator
built into it, that fits four poles/four zeros to a solution of a 1980's
vintage analog seismometer/galvonometer system. We are using it to generate
poles & zeros for digitized analog seismograms. I am having some trouble
with phase right now, and struggling with it to some extent: I seem to be
out by a factor of pi/2 or maybe pi; but I think the estimator works pretty
well. THere's some sample data in there as well, and PZCalc matches the
classical Soviet calibration curves within one percent.

I am working now, on creating a direct calculation of poles/zeros to
replace the estimator for cases where we have Ts,Tg,Ds,Dg, S^2, and
sensitivity. However, the estimator should work on a .fap response file
just as well as the Soviet calibration, since they both generate curves of
amplitude and phase. In fact, I have done that very thing with a .fap file
from Lamont Doherty, from their digitized PNE project.

I'll update my github with the latest code tomorrow, when I can access my
workstation. You are welcome to view it and see if it's a useful solution.

-Dan
https://github.com/tychoaussie/PZcalc

krischer:
> We tend to force the sign of the poles. However, we are starting with an initial model and then perturbing that. Do you think adding an optional `initial_guess` would help?

There is a simple way to add this to the current proposed implementation: A few optimization methods (e.g. `L-BFGS-B`) in scipy allow specifying bounds for all parameters. That would solve that and it would probably only converge marginally slower than the current full BFGS method.

Regarding the initial guess one could have a few known curves for each number of poles and zeros, scale them to the correct amplitude, and use them as initial guess. Most response curves are somewhat similar in the end and these could make it much easier to find sensible gradients for the optimization. But I am just guessing as I have never tried this.

tychoaussie:
LDEO_vs_1973_calibration
Is this what you are trying to do? 
Here is a graph of a .fap file, taken from the Lamont Doughtery digitization project, of a short period SKM instrument that is connected to a galvanometer. (pictured in blue). The green line is from an actual station calibration booklet for station MNAS, circa 1973. The red line is the graph of the calculated poles/zeros that are a best-estimate of the curve. It generally hits about one percent when estimating both amplitude and phase.

PZCALC , which can probably be modified to other, more generic graphs over wider frequencies, tries to fit complex amplitude first, using the BFGS estimator. Then, it iterates through 128 estimations until it finds the best-fit curve. Once the misfit falls below about 5, it stops the estimation, and saves the best-pole-zero solution. Once it is complete, (after running through three channels), it generates a dataless SEED file for use with digitized analog records from that particular historical station. 

Is this helpful?

aringler-usgs:
@krischer good call on the constrained inversion.  I added this in as an option.  Your suggestion about an initial curve set of curves is good.  I think most folks are basically going to fit poles and zeros to either a broadband seismometer with a Butterworth filter type corner or a geophone with a similar shape.  Both of these are going to be complex conjugate pairs.  

I will add some additional test cases in the coming days.

@tychoaussie your example could definitely be used as a test case. For a general obspy routine I would think we would want to avoid things like assuming there is a galvo in the system.

megies:
Thanks for linking your PZcalc project. How would i call it given the F-A-P list from the example used here? https://github.com/obspy/obspy/pull/2618#issuecomment-633974011

tychoaussie:
Obspy_experimental1

`import numpy as np
import scipy.signal
import scipy.optimize
import sys

def misfit(a,b):    # Bring in two lists, compare them, 
                    # and present a scalar representing a ratio of how well they fit.
                    # a and b should be the same length.
    d = []
    n = []
    for i in range(0,len(a)):
        n.append((np.abs(a[i]-b[i]))) # Check the difference between the two values
        d.append((np.abs(a[i]+b[i])))
    misfit = np.sum(n)/np.sum(d)
    return(misfit)

    
                    # PAZ subroutine that uses the same frequencies as derived from original curve
def pazto_freq_resp(freqs, zeros, poles, scale_fac): 
    b, a = scipy.signal.ltisys.zpk2tf(zeros, poles, scale_fac)
    if not isinstance(a, np.ndarray) and a == 1.0:
        a = [1.0]
    return scipy.signal.freqs(b, a, freqs * 2 * np.pi)[1] 
            # list element 0 is frequencies
            # list element 1 is the complex amplitudes


                    # import list of complex numbers and return the angle between 90 and 270 degrees
def phasecalc(testresponse):  
    testphase = []
    for i in range(0,len(testresponse)): # ,t in testresponse.enumerate():
        tp = np.arctan2(testresponse[i].imag , testresponse[i].real) * 180. / np.pi
        if (testresponse[i].imag > 0) & (testresponse[i].real < 0):
            tp = tp-360.
        testphase.append(tp) # adjust phase to better match what is seen in the real world calibrations
    return(testphase)  

                    # Convert phase lag from seconds per period into degrees. 
def phase2degree(phase,period):
    phasedeg = []
    for i in range(0,len(phase)):
        phasedeg.append(float(phase[i])/float(period[i])*360.) 
    return(phasedeg)


                    # Convert phase lag from degrees into seconds/period
def degree2phase(phasedeg,period):
    phasesec = []
    for i in range(0,len(phasedeg)):
        phasesec.append((float(phasedeg[i])/360.0)*float(period[i]))
    return(phasesec)

    # Readcal returns a list of channel calibration parameters from a calibration file.
    # Depending on if a dataless seed is to be generated or not.

def minimize(_var,frequencies,response):    # Uses data found in frequencies, and in response. 
    p1r, p1i, p3r, p4r,p5r,z1r,z2r,z3r, scale_fac = _var
    new_resp = pazto_freq_resp(
        freqs=frequencies,

        zeros=np.array([0.0 + 0.0 * 1j,
                        0.0 + 0.0 * 1j,
                        z1r + 0.0 * 1j,
                        z2r + 0.0 * 1j,
                        z3r + 0.0 * 1j], dtype=np.complex128),                        

        poles=np.array([p1r + p1i * 1j,
                        p1r - p1i * 1j,
                        p3r + 0.0 * 1j,
                        p4r + 0.0 * 1j,
                        p5r + 0.0 * 1j], dtype=np.complex128),
        scale_fac=scale_fac)
    return ((np.abs(new_resp) - np.abs(response)) ** 2).sum()
    
    
def findpaz(channel,frequencies,magnitudes,phases): 

    Component = channel[0]
    Caldate   = channel[1]
    Period = []
    for freq in frequencies:
        Period.append(1/freq)
    
    Phasedeg = phases
    response = np.array(magnitudes, dtype=np.float32)

    evaluation = 1.0E+09 # For evaluating how close the solution is to the original curve
    np.seterr(divide='ignore')
    for z in range(0,128): # iterate 128 times to find the solution that best describes the phase response.
        initial_x=[]
        X0=np.random.random(9)
        #                                Using the minimize function, find the poles & zeros solution that best describes
        #                                the instrument response as found in responses, on frequencies breakpoint "frequencies"
        out = scipy.optimize.minimize(
            fun=minimize,
            args = (frequencies,response), # An important detail that cost me three weeks to discover.  
            method="BFGS",
            x0=X0,
            options={"eps": 1e-6, "maxiter": 1e8}) # defines the step size & max iterations for the random tests
        x = out.x
        new_poles = np.array([-abs(x[0]) + abs(x[1]) * 1j,
                              -abs(x[0]) - abs(x[1]) * 1j,
                              -abs(x[2]) + 0.0 * 1j,
                              -abs(x[3]) + 0.0 * 1j,
                              -abs(x[4]) + 0.0 * 1j], 
                              dtype=np.complex128)    
     
        new_zeros = np.array([ 0.0 + 0.0 * 1j,
                               0.0 + 0.0 * 1j,
                              -abs(x[5]) + 0.0 * 1j,
                              -abs(x[6]) + 0.0 * 1j,
                              -abs(x[7]) + 0.0 * 1j],
                              dtype=np.complex128)
        new_scale_fac = x[8]
        #              Create the response curve that results from this theoretical new poles and zeroes solution
        inverted_response = pazto_freq_resp(freqs=frequencies, zeros=new_zeros, poles=new_poles,scale_fac=new_scale_fac)    
        inphase = phasecalc(inverted_response)                        # phase from inverted response, listed in degrees
        curvefit = np.sqrt(((np.array(Phasedeg) - np.array(inphase))**2).mean()) # This is the rmse function for misfit
        if (curvefit) < evaluation:
            final_iteration = z
            best_poles=new_poles
            best_zeros=new_zeros
            best_scale_fac=new_scale_fac
            print(f'\nIteration # {z}: Phase misfit reduced to {curvefit:0.3f}')
            evaluation = curvefit
            if evaluation < 12.0:    # Evaluation is a measure of how well the poles and zeros fit the original response & phase.
                break               # Less than 12 is Good enough. End the loop early to speed up the process.
        else:
            sys.stdout.write('.')
            sys.stdout.flush()
    print('\n')
    inverted_resp = pazto_freq_resp(freqs=frequencies, zeros=best_zeros, poles=best_poles,scale_fac=best_scale_fac)
    inverted_phase = phasecalc(inverted_resp)

    # PAZ gain constant is a convolution of the AO normalization factor and the instrument magnification.
    # Create code that properly breaks apart the AO normalization and sensitivity such that the plot equals 1.0 at the peak
    # frequency, and report both AO normalization, peak frequency, and the amplification for use in a proper PZ file. 
    AO_index = np.argmax(np.abs(inverted_resp))
    Sense = np.abs(inverted_resp[AO_index])
    Sensefreq = 1./Period[AO_index] # in Hz.
    AO_norm = best_scale_fac/Sense
    # Also must include frequency where sensitivity is highest
 
    paz =      []
    paz.append(best_poles)
    paz.append(best_zeros)
    paz.append(best_scale_fac)
    paz.append(AO_norm)
    paz.append(Sense)
    paz.append(Sensefreq)
    paz.append(evaluation)
    paz.append(Component)
    paz.append(Caldate)
    plotchan = []
    plotchan.append(Component)
    plotchan.append(Caldate)
    plotchan.append(Period)
    plotchan.append(magnitudes)                   #(Gain)
    plotchan.append(phases)                       #(Phase)
    plotchan.append(inverted_resp)
    plotchan.append(inverted_phase)
    # Save the poles & zeros to a file, and print the final result to a plot and save it as well.
    return(plotchan,paz)


def respplot(plotchan):

    # plotchan contains a list for each channel to be plotted where each item consists of:
    # plotchan[][0] = Component: The channel name.
    # plotchan[][1] = Calibration date for the channel
    # plotchan[][2] = Period is the list of periods upon which all calculations are based
    # plotchan[][3] = Gain contains the list of gains for all three axes from the published parameters
    # plotchan[][4] = Phase contains the list of phase delays for all three axes as a function of seconds delay from published parameters
    # plotchan[][5] = inverted_resp is the list of gains from the poles and zeros
    # plotchan[][6] = inverted_phase is the list of phase values in degrees. from poles and zeros


    title = "Frequency response and Phase response vs time \n Calculated from published parameters" \
            + "for cal date of "+plotchan[0][1] # use the first channel's date.
    colorwheel = ['blue','orange','green','red','cyan','magenta', \
                  'blue','orange','green','red','cyan','magenta', \
                  'blue','orange','green','red','cyan','magenta', \
                  'blue','orange','green','red','cyan','magenta'  ]


                # Determine the min/max x axis scale to fit the frequency band.
    maxfreq=0
    minfreq=1.0E+6 # version 1 had conflict with variables named 'max' and 'min'
    for p in plotchan:
        if (np.amax(p[2]) > maxfreq):
            maxfreq = 1/np.amin(p[2]) # maximum frequency based on the period in p
        if (np.amin(p[2]) < minfreq):
            minfreq = 1/np.amax(p[2]) # minimum frequency

    print(f'Minimum frequency = {minfreq} Hz and maximum frequency = {maxfreq}')
    plt.figure()
    plt.subplot(211)
    plt.axis([0.01,100,10,500]) #[minfreq*0.5,maxfreq*2,10,500]) # scale the plot between 0.1 and 100K magnification
#    plt.axis([0.01,10,0.1,100000])            # plot the amplitude curves for each of the loaded channels and their associated PAZ estimation
    for i in range (0,len(plotchan)):
        Freq = [1./ float(plotchan[i][2][j]) for j in range(len(plotchan[i][2]))]        
        plt.loglog(Freq,plotchan[i][3],color='blue',lw=5)
        plt.loglog(Freq,np.abs(plotchan[i][5]),color='red',lw=1)
                # plot the phase curves and their associated PAZ estimation

                # plot the axes
    plt.xlabel('Frequency [Hz]')
    plt.ylabel('Amplitude')
#      Don't annotate the record with channel names if there's more than six channels. Its too busy.
    XY = [[[0.02,100],[0.02,60.0],[0.02,35],[0.02,600],[0.02,350.0],[0.02,200]], \
          [[0.015,0.4],[0.22,0.4],[3.0,0.4],[0.015,0.4],[0.22,0.4],[3.0,0.4]]]
    if len(plotchan)<7:
        for i in range(0,len(plotchan)):
            plt.annotate(plotchan[i][0] + " amp",         xy=XY[0][i],xytext=XY[0][i], color = colorwheel[i])
    plt.grid(True, which="both")

    plt.subplot(212)
    plt.axis([0.01,100,-120,120]) #[minfreq*0.5,maxfreq*2,-120,120]) # scale the plot between 0.1 and 100K magnification
    for i in range(0,len(plotchan)):
        Freq = [1./ float(plotchan[i][2][j]) for j in range(len(plotchan[i][2]))]
#        channelphasedeg = phase2degree(plotchan[i][4],plotchan[i][2])  # phase, period
        plt.semilogx(Freq,plotchan[i][4],color='blue',lw=3) # frequency vs phase in seconds?
        plt.semilogx(Freq,plotchan[i][6],color='red',lw=1)
    plt.title = "Phase Response of SKM"
    plt.xlabel('Frequency [Hz]')
    plt.ylabel('Phase(degrees)')
    plt.text(maxfreq*2.6, 0.1, 'Phase (degrees)', fontsize=11,
                   rotation=90.0, rotation_mode='anchor')
    plt.suptitle(title)

    XY = [[[0.02,-270],[0.02,-200],[0.02,0],[0.02,45],[0.02,90],[0.02,125]], \
          [[0.015,-100.0],[0.22,-100],[3.0,-100],[0.015,-100],[0.22,-100],[3.0,-100]]]    
    if len(plotchan) < 7:
        for i in range(0,len(plotchan)):
            plt.annotate(plotchan[i][0]+" phase",xy=XY[1][i],xytext=XY[1][i], color = colorwheel[i])
    plt.annotate("Inverted parameters from calculated poles & Zeros",xy=(minfreq*0.2,-100),xytext=(minfreq*0.2,-100),color='red')
    plt.suptitle(title) 

#    plt.savefig(outfil+".png")
#    exportcsv(plotchan,outfil)  # Export the parameters as a csv file for additional analysis
    plt.show()

%matplotlib qt
Plotchan = []
Paz = []
frequencies = np.array(
    [0.5, 0.8, 1, 1.25, 1.6, 2, 3.15, 5, 10, 16, 40, 63, 80])
magnitudes = [95.86, 155.48, 182.27, 202.98, 214.98, 217.61, 224.64, 234.46,
              238.78, 239.48, 236.67, 226.75, 216.80]
phases = [101.6, 75.7, 63.0, 50.8, 38.8, 30.5, 20.3, 12.0, 2.6, -2.8, -17.3,
          -29.7, -38.2]
channel = ['HHZ','20200602']

plotchan,paz = findpaz(channel,frequencies,magnitudes,phases)
Plotchan.append(plotchan)
Paz.append(paz)
print("Pole zero estimator complete")

Channel = []
Caldate = []
Real_pole = []      # Poles.real
Imaginary_pole = [] # Poles.imaginary
Real_zero = []      # Zeros.real
Imaginary_zero = [] # Zeros.imaginary
AO_norm = []        # Normalization factor
Norm_freq = []      # Normalization frequency at which point max sensitivity is reached & normalized gain = 1 
Vo = []             # Max sensitivity (peak amplification factor between ground motion and its deflection on paper)

for pole in paz[0]:
    print (f'Poles:{pole.real:0.4f} {pole.imag:0.4f}j ')
for zero in paz[1]:
    print (f'Zeros:{zero.real:0.4f} {zero.imag:0.4f}j ')    
print(f'AO Correction factor: {paz[3]:0.4f}')
print(f'Vmax = {paz[4]:0.3f} @ {paz[5]} Hz')      
respplot(Plotchan)
`
I hope this isn't too much code, but it's a five pole/zero estimator that seems to work nicely with short-period electromechanical instruments with a coupled galvonometer. I adapted it to your example to see what would happen.

tychoaussie:
I am really sorry with how sloppy that last message looks. I can't seem to figure out how to insert properly code.

aringler-usgs:
@tychoaussie you can use single quotes for inline code.  Like \`print\` produces: `print`.  For multiple lines of code you can use
 \`\`\`print
print \`\`\` 

To produce:
```
print
print
```

This might help: [Mastering Markdown](https://guides.github.com/features/mastering-markdown/)

megies:
@tychoaussie thanks, seems pretty similar. The main difference I see is that you have quite strong constraints on the poles/zeros, which might be a good way to go in most cases, I guess and that you focus on fitting the absolute amplitude value of the response (I think?). The misfit calculations you do are a bit confusing to me right now, I tried swapping all misfit measures out for a simple `euclidean(.., ..)` on the complex values and it seems pretty similar, with better fit in the phase and a bit worse fit on the amplitudes (trading off more between the two). But I guess this might also be due to the strong constraints on the poles/zeros not being able to accommodate the "bump" on the left side of the amplitude response.

Also, I think you can get rid of all the custom phase unwrapping. I just swapped that out for `np.angle(complex_response)` and it seems to work without issues.

So I guess the main difference really is just the constraints on the poles/zeros..?

Figure_1

megies:
> I am really sorry with how sloppy that last message looks. I can't seem to figure out how to insert properly code.

no worries, and check this for folding long details in comments: 
 - https://gist.github.com/joyrexus/16041f2426450e73f5df9391f7f7ae5f
 - https://gist.githubusercontent.com/joyrexus/16041f2426450e73f5df9391f7f7ae5f/raw/f774f242feff6bae4a5be7d6c71aa5df2e3fcb0e/README.md

obspy/signal/invsim.py
megies: Yeah, I agree, I just wanted to leave it fully unconstrained for now. Maybe it would be easiest to just calculate the misfit not only on sample points but maybe on oversampled points with just linear interpolation (but giving more weight to the actual sampling points)

krischer: Regularizing can have the same effect and would be less hacky I think. Adding some kind of term to the misfit that promotes smoothness of the solution might work.

PR #2619, opened 2020-06-03
Adds pyproject.toml (PEP518) to make pip install more robust

heavelock:
This little file is a containing a list of required packages to be installed before obspy. Current version of pip (probably since v19) are able to resolve the order and install multiple packages at once.

Currently, if you are in a fresh environment and want to install numpy and obspy, you have to run pip twice. `python -m pip install numpy && python -m pip install obspy` 
It's because: 
```
$ python -m pip install numpy obspy

...

      File "/tmp/pip-req-build-pk9jzee2/setup.py", line 37, in <module>
        raise ImportError(msg)
    ImportError: No module named numpy. Please install numpy first, it is needed before installing ObsPy.
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

``` 

With this file included, pip will first install first numpy and only then will try to build wheels for obspy.

You can try it yourself:

```
$ pip install numpy git+git://github.com/heavelock/obspy@41bb2ae
```

I'm merging it into master but it could be also backported to 1.2.x branch with changed versions.

https://github.com/pypa/pip/issues/5761

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,11 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = [
+    'numpy>=1.15.0',
+    'scipy>=1.0.0',
+    'matplotlib>=3.2.0',
+    'lxml',
+    'setuptools',
+    'sqlalchemy',
+    'decorator',
+    'requests']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,6 @@
 [build-system]
 # Minimum requirements for the build system to execute.
 requires = [
-    'numpy>=1.15.0',
-    'scipy>=1.0.0',
-    'matplotlib>=3.2.0',
-    'lxml',
-    'setuptools',
-    'sqlalchemy',
-    'decorator',
-    'requests']
+    'numpy>=1.6.1',
+    'setuptools'
+    ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
 # Minimum requirements for the build system to execute.
+# see PEP518: https://www.python.org/dev/peps/pep-0518/
 requires = [
     'numpy>=1.6.1',
     'setuptools'


megies:
So, if this is only specifying the pre-install requirements, it should only include numpy and setupttols, I believe. Anything other than that should not be needed and it will just eventually get overlooked to be updated. The numpy version could also be way more lax I believe, as it could get updated to the needed version in the second installation step?

heavelock:
Hmmm, Initially I though that it was supposed to be a kind of future replacement of a setup.py, but now I think you are right, it's only about the build dependencies. 

I just tested it with newest commit that includes your proposition (numpy relaxed to >=1.6.1 as required by 1.2.1) and pip happily installs `numpy-1.18.5`

heavelock:
Should I backport it to 1.2.x? Is there any maintenance release planned?

megies:
I think we will have a bugfix release on 1.2 since there were a few significant fixes. It will probably be only a pypi release and maybe conda packages but we'll see (we might only get conda packages for Python 3, they probably switched off Python2 building already)

heavelock:
So it's even better since it doesn't affect conda :) 
I will do a PR targetting 1.2.x. Since the numpy's version is according to minimum there, I will leave everything as here.

trichter:
I now get this warning when installing with pip:

```
pip install --no-deps -e .
Obtaining file:///home/eule/dev/obspy
  Installing build dependencies ... done
  WARNING: Missing build requirements in pyproject.toml for file:///home/eule/dev/obspy.
  WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Installing collected packages: obspy
  Running setup.py develop for obspy
Successfully installed obspy
```

megies:
>   WARNING: Missing build requirements in pyproject.toml for file:///home/eule/dev/obspy.

that warning should be reported upstream at pip, assuming you *do* have numpy and setuptools installed

>   WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.

thats also a real weird warning, don't even understand what it could hint at in your case, in any case both are for upstream i think

trichter:
This pip call did not show the warnings, it created a wheel.

```
pip install --no-deps git+git://github.com/heavelock/obspy.git@56ba838
Collecting git+git://github.com/heavelock/obspy.git@56ba838
  Cloning git://github.com/heavelock/obspy.git (to revision 56ba838) to /tmp/pip-req-build-uxugq_w1
  Running command git clone -q git://github.com/heavelock/obspy.git /tmp/pip-req-build-uxugq_w1
  WARNING: Did not find branch or tag '56ba838', assuming revision or ref.
  Running command git checkout -q 56ba838
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: obspy
  Building wheel for obspy (PEP 517) ... done
  Created wheel for obspy: filename=obspy-1.0.3.post0+3883.g56ba838849-cp37-cp37m-linux_x86_64.whl size=21366224 sha256=1ba039f287a336a576564f911dd31940a3a153c26f4aa33784cf3ca3b2d758b9
  Stored in directory: /tmp/pip-ephem-wheel-cache-0o9mbxpv/wheels/81/a4/58/f7a671ffaff2f409e991bf299abd4652dc8f37745655fee973
Successfully built obspy
Installing collected packages: obspy
Successfully installed obspy-1.0.3.post0+3883.g56ba838849
```

PR #2636, opened 2020-06-18
SDS: Continue get_all_stations() when invalid channel codes are encountered

FMassin:
### What does this PR do?

Adding `TypeError` in `get_all_stations` allows to have weird channels name in filenames like `1967/XX/WEIRD/D-6.D` without crashing ...

Because I have a channel code called `D-6` and it crashes `obspy-sds-report`.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,9 @@ Changes:
      station (see #2611)
    * update RASPISHAKE URL mapping to use https
    * fix a bug of not handling HTTPException in mass_downloader (see #2606)
+ - obspy.clients.filesystem:
+   * sds: continue get_all_stations() even if encountering an invalid channel
+     code (see #2636)
  - obspy.clients.neic:
    * Make client socket blocking (see #2617)
  - obspy.io.nordic:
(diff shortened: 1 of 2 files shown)


megies:
Since this is failing in the (network, station) lookup routine another option would be to make the lookup in there only try to match those two, then you could lookup the station without bothering about the "invalid" channel code.

If sticking to this approach I'll add a small change to be a little more specific in catching the exception

FMassin:
> Since this is failing in the (network, station) lookup routine another option would be to make the lookup in there only try to match those two, then you could lookup the station without bothering about the "invalid" channel code.

That sounds like a good design idea, but then the risk is to consider stations that actually don't have any valid seedids... please do as you want, maybe I should have open an issue instead of trying to fix the code myself. 

> If sticking to this approach I'll add a small change to be a little more specific in catching the exception

I wouldn't feel ashamed if you decide to follow your own way. This should be called a "Pull suggestion" instead of a request :)

megies:
@FMassin nah, all good, I'm fine with it is right now.

megies:
I guess this could actually go into maintenance branch, actually

FMassin:
Anything I should do?

megies:
I did it real quick

megies:
I'd say ready to merge when CI looks ok

Issue #2640, opened 2020-06-23
Travis not building


heavelock:
I think it is a problem connected with changes introduced by me in #2619. In my project CI was building with that setup up to Friday, but crashed today. I fixed it with adding to pyproject.toml required by PEP517. I will do a PR to add this line here.

Quoting @megies:

Not sure whats wrong with Travis but looks like a clash between setuptools and numpy, so hopefully will go away eventually and we can just wait it out.

```
    File "/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pip-build-env-dq61kdlp/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 142, in run_setup

      exec(compile(code, __file__, 'exec'), locals())

    File "setup.py", line 47, in <module>

      from numpy.distutils.core import DistutilsSetupError, setup

  ImportError: cannot import name 'DistutilsSetupError' from 'numpy.distutils.core' (/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pip-build-env-dq61kdlp/overlay/lib/python3.7/site-packages/numpy/distutils/core.py)
```

_Originally posted by @megies in https://github.com/obspy/obspy/pull/2626#issuecomment-648019872_

heavelock:
Okay, It's not my fault. Numpy released version 1.19 on 2020.06.20 on PyPi. As for now, neither documentation site nor the home site is updated and doesn't show that fact.

PR #2641, opened 2020-06-23
Adds build backend to pyproject.toml

heavelock:
This line introduces `build backend` to pyproject.toml in order to fulfil the requirements of PEP517. Should fix building on Travis.

Ref to: #2640

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,4 @@ requires = [
     'numpy>=1.6.1',
     'setuptools'
     ]
+build-backend = "setuptools.build_meta"


heavelock:
@QuLogic I saw that you were adding `pyproject.toml` to Cartopy, any thoughts on that issue here?

megies:
Seems like this is getting rid about a warning but the error seems like incompatible setuptools <---> numpy versions.

P.S.: Fun fact of the day, this is your second PR with spelling it "pyptoject.toml" (the first time I was actually wondering what kind of file that would even be :rofl: )

heavelock:
> P.S.: Fun fact of the day, this is your second PR with spelling it "pyptoject.toml" (the first time I was actually wondering what kind of file that would even be 🤣 )

Recently I decided to relearn how I type on keyboard so I will say that for sure it's due to that XD

PR #2647, opened 2020-06-24
Backport numpy distutils fix

krischer:
This is a backport of https://github.com/obspy/obspy/pull/2643 to the maintenance branch.

Installation of ObsPy with the latest numpy is currently not possible on platform without wheels. I think this warrants a new minor ObsPy release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,8 +43,9 @@ import os
 import sys
 import platform
 from distutils.util import change_root
+from distutils.errors import DistutilsSetupError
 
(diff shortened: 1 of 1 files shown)


megies:
I tend to agree, we might have to do a release ASAP, aka today. We already get bombarded with installation issue reports about this

@krischer

krischer:
I'd agree - I think we can just tag 1.2.2 immediately and then push to pypi - the other packages can come later but then the issue is resolved for most people.

PR #2656, opened 2020-06-25
FDSN: set timeout according to default or user setting in service discovery

megies:
### What does this PR do?

Uses the FDSNWS Clients' timeout value for service discovery too.

Service discovery was on a different timeout than every other http request issued by the client. Discovering services is vital and should not be on a short timeout. 10s is relatively bad response time for a server, but nevertheless we should allow higher timeout here.
I stumbled over this while trying to debug some flakiness in service discovery and some other requests in EIDA routing client use.

### PR Checklist
- [ ] All tests still pass.

CC @krischer

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,7 @@ Changes:
      station (see #2611)
    * update RASPISHAKE URL mapping to use https
    * fix a bug of not handling HTTPException in mass_downloader (see #2606)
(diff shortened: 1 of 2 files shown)


megies:
+TESTS:clients.fdsn

megies:
Looks OK now. http://tests.obspy.org/?pr=2656

PR #2665, opened 2020-07-01
Removal of matplotlib<3 compatibility checks

heavelock:
I noticed a few TODOs to remove some compatibility checks when minimum matplotlib version will be higher than 2.0.  Since we have minimum of 3.2 now, I went ahead and removed them. I also searched the codebase for all checks agains `MATPLOTLIB_VERSION` and I removed all that were checking for versions <3.0. There are left a few of them that are checking for `MATPLOTLIB_VERSION == [3, 0, 1]` but since we still support 3 as a major version, I didn't touch them.

### DIFF
--- a/obspy/core/util/testing.py
+++ b/obspy/core/util/testing.py
@@ -331,10 +331,9 @@ class ImageComparison(NamedTemporaryFile):
         # which is after https://github.com/matplotlib/matplotlib/issues/7905
         # has been fixed.
         #
-        # Thus test images should accurate for matplotlib >= 2.0.1 anf
+        # Thus test images should accurate for matplotlib >= 2.0.1.
         if adjust_tolerance:
-
-            # One last pass depending on the freetype version.
+            # Adjusts tolerance depending on the freetype version.
             # XXX: Should eventually be handled differently!
             try:
                 from matplotlib import ft2font
--- a/obspy/core/util/testing.py
+++ b/obspy/core/util/testing.py
@@ -321,13 +321,8 @@ class ImageComparison(NamedTemporaryFile):
         self.plt_close_all_exit = plt_close_all_exit
         self.no_uploads = no_uploads
 
-        if (MATPLOTLIB_VERSION < [1, 4, 0] or
-                (MATPLOTLIB_VERSION[:2] == [1, 4] and style is None)):
-            # No good style support.
-            self.style = None
-        else:
-            import matplotlib.style as mstyle
-            self.style = mstyle.context(style or 'classic')
+        import matplotlib.style as mstyle
+        self.style = mstyle.context(style or 'classic')
 
         # Adjust the tolerance based on the matplotlib version. This works
         # well enough and otherwise testing is just a pain.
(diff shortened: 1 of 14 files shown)


heavelock:
Yeah, I think the plan is to drop the basemap completely but for now I just figured out to keep it.

heavelock:
O, I caused few new errors on Flake8, I will fix them in a minute.

heavelock:
Okay, flake8 fixed almost completely.

However, there are two errors that pop out:

```
obspy/io/xseed/blockette/blockette048.py:57:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
obspy/io/xseed/blockette/blockette058.py:90:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
```

Well the errors are there, however, last time it was touched was in 2016 by @krischer.  What to do with them?

heavelock:
Oooookay. Dragons live there in that code. Basically, there are only few blockette objects that are tested. 

Blockette010, Blockette051, Blockette053, Blockette054
Blockette041, Blockette050, Blockette054, Blockette060

So those ones are not tested at all it seems.

heavelock:
And the error came from Flake8 including support to pyflakes 2.2.x.  https://github.com/PyCQA/flake8/commit/76eeccad89ac83d7049ec7ebb10776609f7a103d

obspy/core/util/testing.py
```diff
@@ -337,30 +332,7 @@ def __init__(self, image_path, image_name, reltol=1,
         # has been fixed.
         #
         # Thus test images should accurate for matplotlib >= 2.0.1 anf
```
QuLogic, line 334:
```suggestion
        # Thus test images should accurate for matplotlib >= 2.0.1.
```

obspy/imaging/tests/test_ppsd.py
```diff
@@ -43,11 +43,7 @@ def test_ppsd_plot_frequency(self):
         """
         Test plot of ppsd example data, normal (non-cumulative) style.
         """
-        # mpl < 2.2 has slightly offset ticks/ticklabels, so needs a higher
-        # tolerance (see e.g. http://tests.obspy.org/102260)
         reltol = 2
```
QuLogic, line 46: Inline this in the `ImageComparison` call.

obspy/signal/tests/test_spectral_estimation.py
```diff
@@ -478,11 +477,7 @@ def callback(t_array):
             np.testing.assert_array_equal(selection_got, expected_selection)
 
         # test one particular selection as an image test
-        # mpl < 2.2 has slightly offset ticks/ticklabels, so needs a higher
-        # tolerance (see e.g. http://tests.obspy.org/102260)
         reltol = 1.5
```
QuLogic, line 480: Inline this in the `ImageComparison` call.

megies:
> However, there are two errors that pop out:
> 
> ```
> obspy/io/xseed/blockette/blockette048.py:57:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
> obspy/io/xseed/blockette/blockette058.py:90:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
> ```

I've mentioned those before in #2631, they are unrelated but should get fixed. I was hoping somebody that actually worked on the blockette stuff more could have a look, because I would have to look up details for these blockettes first and would probably just go by some examples I find.

barsch:
> 
> > However, there are two errors that pop out:
> > ```
> > obspy/io/xseed/blockette/blockette048.py:57:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
> > obspy/io/xseed/blockette/blockette058.py:90:21: F507 '...' % ... has 4 placeholder(s) but 3 substitution(s)
> > ```
> 
> I've mentioned those before in #2631, they are unrelated but should get fixed. I was hoping somebody that actually worked on the blockette stuff more could have a look, because I would have to look up details for these blockettes first and would probably just go by some examples I find.

I'll take a look

heavelock:
@QuLogic done, take a look please.

QuLogic:
> The default `reltol` is 1, so you should just remove any explicit usage of it when it's also 1.

Don't forget this part.

heavelock:
Fixed. @QuLogic

heavelock:
Rebased on current master and force pushed.

megies:
Hmm.. I guess we might want to consider updating all test baseline images on master too? Since we bumped matplotlib massively and I think many of those were done on older matplotlib versions.. I guess we want to use matplotlib 3.2.0 as our new min dependency version for all baseline images from now on?

heavelock:
Well, you are right about that. Since now we don't have that much of a wiggle with versions, we could just update everything. Also, I think the basemap is not compatible with matplotlib 3 so we should also do that long overdue migration to cartopy.

But I think it should belong to a different PR.

megies:
Sounds good, lets merge this first of all then.

PR #2668, opened 2020-07-02
should fix flake8 issues for io.xseed

barsch:
unfortunatly I can't find a single RESP example containing blockette 48 or 59 **with** any calibration data - therefore no test case atm ...

see https://github.com/obspy/obspy/pull/2665#issuecomment-652869882 and https://github.com/obspy/obspy/pull/2631#issue-432383084

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,9 @@ master
 Changes:
  - obspy.clients.filesystem:
    * add get_waveforms_bulk() method to SDS client (see #2616, #2626)
+ - obspy.io.xseed:
+   * fix a bug reading SEED blockettes 48 and 58 which was likely never
+     encountered (see #2668)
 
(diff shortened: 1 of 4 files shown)


megies:
Looks like this is actually a real change and not only cosmetics? If yes, we'd want a one-liner in changelog to be able to keep track

heavelock:
Why not convert it to a .format(). Fstring would work also but it might cloud up too much the readability of the line.

barsch:
@megies it is an undetected bug because never tested within ObsPy - as I said I need real example files in order to test it

@heavelock there are various things I would like to change - e.g. a variable ```string``` is bothering me really - anyway, the RESP part wasn't create by me so I didn't want to mess with it to much

aringler-usgs:
@megies sorry for the slow response.  These changes look fine.  I doubt anyone has ever used these fields.  I couldn't find them in our data holdings.

Issue #2653, opened 2020-06-25
Fine grained exceptions for FDSN client


yetinam:
I'm using the Obspy FDSN client for a fairly huge download and found myself parsing error messages from `FDSNException` objects, which is a bit ugly. I was wondering, if obspy could introduce more fine grained exceptions for different FDSN errors, e.g. `FDSNAuthentificationException`, `FDSNServerException`, ... This was previously done for the `FDSNNoDataException` (#1656).

If there is interest in this, I could write a pull request.

megies:
I don't see a reason why we would not want more fine grained exceptions, so as far as I am concerned, feel free to make a PR.

yetinam:
Hey @megies ,
could you please convert this issue into a PR? I implemented a draft for the exceptions at ec22245ad410b89613f04010396dabca36738f93.

megies:
@yetinam i could, but that would make the first post in this ticket appear as if it was made by me.
I never understood why github did not add the issue->PR conversion properly, they have it hanging in this semi-official deprecated state for years. Let me know if I should go ahead

yetinam:
Okay, than I'll just open a PR and reference this issue.

PR #2670, opened 2020-07-03
Introduce fine-grained FDSN client exceptions

yetinam:
### What does this PR do?

This PR introduces fine-grained exceptions for the FDSN client to allow easier exception handling. As all newly introduced exceptions subclass `FDSNException` this change should be compatible with any existing code.

Feature was proposed in #2653

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -34,11 +34,16 @@ from obspy.clients.fdsn.header import (DEFAULT_USER_AGENT, URL_MAPPINGS,
                                        FDSNNoDataException,
                                        FDSNRequestTooLargeException,
                                        FDSNBadRequestException,
-                                       FDSNAuthenticationException,
+                                       FDSNNoAuthenticationServiceException,
                                        FDSNTimeoutException,
                                        FDSNNoServiceException,
-                                       FDSNServerException,
+                                       FDSNInternalServerException,
                                        FDSNTooManyRequestsException,
+                                       FDSNServiceUnavailableException,
+                                       FDSNUnauthorizedException,
+                                       FDSNForbiddenException,
+                                       FDSNDoubleAuthenticationException,
+                                       FDSNInvalidRequestException,
                                        DEFAULT_SERVICES)
 from obspy.core.inventory import Response
 from obspy.geodetics import locations2degrees
@@ -116,7 +121,7 @@ class ClientTestCase(unittest.TestCase):
         msg = ("Empty 'bulk' parameter potentially leading to a FDSN request "
                "of all available data")
         for bad_input in [[], '', None]:
-            with self.assertRaises(FDSNBadRequestException) as e:
+            with self.assertRaises(FDSNInvalidRequestException) as e:
                 get_bulk_string(bulk=bad_input, arguments={})
             self.assertEqual(e.exception.args[0], msg)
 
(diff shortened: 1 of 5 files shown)


yetinam:
+TESTS:clients.fdsn

yetinam:
There are currently no tests for `FDSNServerException` (triggered if the server returns http code 500) and `FDSNTooManyRequestsException`, as I was not able to consistently trigger them from a web service. I'd appreciate any suggestions on how to test this.

krischer:
Instead of relying on webservices to return a certain status (which might change in the future) it would be easier and more stable to mock out the response to trigger all the code paths in the tests. A simplistic way of doing this would be to just mock out the `download_url()` function: https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/client.py#L1751

Please have a look at the mock library here in case you are not familiar with it: https://docs.python.org/3/library/unittest.mock.html

yetinam:
Thanks for the pointer. I'll modify and extend the tests accordingly.

yetinam:
I changed the tests so that they uses mocks now and extended them. The test `test_eida_token_resolution` currently fails, because the EIDA token expired.

megies:
circle ci fails are unrelated and fixed on master

megies:
@krischer you wanna give it the final nod, if you're happy with how the tests are set up?

obspy/clients/fdsn/tests/test_client.py
```diff
@@ -1448,15 +1456,96 @@ def test_get_waveforms_empty_seed_codes(self):
...
+                        return_value=False):
+            self.assertRaises(FDSNAuthenticationException, Client,
+                              eida_token="TEST")
+
+        self.assertRaises(FDSNAuthenticationException, Client, "IRIS",
+                          eida_token="TEST", user="TEST")
+
+        download_url_mock.return_value = (400, io.BytesIO(b"token is expired"))
+        self.assertRaises(FDSNAuthenticationException,
+                          self.client.get_stations)
```
krischer, line 1495: Even though this is indeed an authentication related problem the return code is still 400 and I think it would be a bit cleaner if every return code is mapped to a unique exception.

megies:
> It might be nice to have the HTTP code as part of the exception classes

requests names it `status_code`, might be worth a thought following that attribute naming maybe?

krischer:
> requests names it `status_code`, might be worth a thought following that attribute naming maybe?

Good point!

yetinam:
Thanks for the comments. I modified the exceptions to make sure there is a 1:1 mapping between exceptions and HTTP status codes. I also added the `status_code` to the FDSNException.

I did not create a custom exception for code 414, as it currently is a `NotImplementedException`. As this suggests that it might be implemented sometime, I'd suggest keeping this exception here, but I'm fine with changing it as well.

krischer:
Thanks a for this! Looks good to me know and can be merged from my side if CI passes.

You'll have to rebase on the latest `master` branch to get rid of the conflict.

yetinam:
I rebased the branch.

megies:
not sure why travis didn't run.. would be good to see Travis, otherwise FDSN tests are not run

megies:
@yetinam can you do something like `git commit --amend --no-edit` and force push, to try to trigger Travis again? I don't think there is a way to manually trigger a Travis build for PRs from branches on forks.

yetinam:
That seems to have done the job.

megies:
Tests are in, looks like thats only the old fails we had before, but maybe you can look through them real quick if its what you expect from the PR:
http://tests.obspy.org/111494/

yetinam:
I agree, looks like the failing tests are unrelated

megies:
Thanks @yetinam!

PR #2671, opened 2020-07-03
Brings back consistency in ignored errors in .flake8 and test-suite internal config

heavelock:
So we have two places where we keep config of flake8. One is `.flake8`, second is `obspy/core/tests/test_code_formatting.py` with list of  `FLAKE8_IGNORE_CODES`. 

The .flake8 was recently modified, without modifying the second place. 

flake8 still did not get new API that would allow for using a config file with the api calls. https://flake8.pycqa.org/en/latest/user/python-api.html

### DIFF
--- a/obspy/core/tests/test_code_formatting.py
+++ b/obspy/core/tests/test_code_formatting.py
@@ -36,6 +36,7 @@ FLAKE8_IGNORE_CODES = [
     'E704',
     'W503',
     'W504',
+    'E741',
 ]
 FLAKE8_EXCLUDE_FILES = [
     "*/__init__.py",


d-chambers:
Just a thought but why not just get rid of the flake 8 tests once we migrate to gh actions? That way we only need to configure flake8 in one place, the linting happens exactly once (rather than for each os/python version) and the errors are actually readable.

heavelock:
That mechanism of running linting is included in the test suite. So the question is not if we want to run it inside of CI in this or another way but rather if we want to have users run it inside of the suite or not. Probably not since it's rather targeted for developers and it's quicker to run it as a pre-commit hook or just from cli.

d-chambers:
I agree. When the current flake8 tests fail it's not exactly easy to figure out why, especially for new contributors. Running it from the command line or a precommit hook is much more straight forward.

megies:
> Just a thought but why not just get rid of the flake 8 tests once we migrate to gh actions?

Personally, I like having a flake8 git commit hook, that already stops me committing flake8 violations in the first place. But I havent looked in the mentioned duplication a lot

PR #2677, opened 2020-07-08
enables hiding sensitivity mismatch warning in response.get_evalresp_response

rizac:
### What does this PR do?

This branch simply adds the possibility to hide  sensitivity mismatch warnings in response.get_evalresp_response, leaving the default of False for legacy code using the method

In the Response object public methods a user can not ignore sensitivity mismatch warning, printing to stderr and particularly tedious because the warnings seem to be issued from external C-libraries and thus they are hard (albeit not impossible) to capture from within Python. The fix is simple: the parameter hide_sensitivity_warnings is already implemented in the private class methods: This PR simply adds the optional parameter in response.get_evalresp_response, passing then the value to the nested private methods

Note: I did not implement any new test because it would involve testing external libraries which should not be ObsPy goal (on a terminal it works though)

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1627,11 +1627,11 @@ class Response(ComparingObject):
         :rtype: :class:`numpy.ndarray`
         :returns: frequency response at requested frequencies
         """
+        hsmw = hide_sensitivity_mismatch_warning  # PEP8
         output, chan = self._call_eval_resp_for_frequencies(
             frequencies, output=output, start_stage=start_stage,
             end_stage=end_stage,
-            hide_sensitivity_mismatch_warning=
-            hide_sensitivity_mismatch_warning)
+            hide_sensitivity_mismatch_warning=hsmw)
         return output
 
     def get_evalresp_response(self, t_samp, nfft, output="VEL",
@@ -1676,10 +1676,10 @@ class Response(ComparingObject):
         except Exception:
             freqs = np.linspace(0, fy, int(nfft // 2) + 1).astype(np.float64)
 
+        hsmw = hide_sensitivity_mismatch_warning  # PEP8
         response = self.get_evalresp_response_for_frequencies(
             freqs, output=output, start_stage=start_stage, end_stage=end_stage,
-            hide_sensitivity_mismatch_warning=
-            hide_sensitivity_mismatch_warning)
+            hide_sensitivity_mismatch_warning=hsmw)
         return response, freqs
 
     def __str__(self):
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1598,7 +1598,8 @@ class Response(ComparingObject):
         return output, chan
 
     def get_evalresp_response_for_frequencies(
-            self, frequencies, output="VEL", start_stage=None, end_stage=None):
+            self, frequencies, output="VEL", start_stage=None, end_stage=None,
+            hide_sensitivity_mismatch_warning=False):
         """
         Returns frequency response for given frequencies using evalresp.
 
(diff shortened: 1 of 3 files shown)


megies:
I'm a bit torn about this one. I understand that people might want to hide those warnings, especially when processing large amounts of data, with metadata from a wide variety of sources/authors.

However, I tend to think this should be handled in user code using e.g. `with warnings.catch_warnings(...)` context.

I'd even maybe prefer some mechanism maybe in `obspy.util` to globally deactivate specific warnings before running obpsy code.

Other opinions?

krischer:
> I'm a bit torn about this one. I understand that people might want to hide those warnings, especially when processing large amounts of data, with metadata from a wide variety of sources/authors.
> 
> However, I tend to think this should be handled in user code using e.g. `with warnings.catch_warnings(...)` context.
> 
> I'd even maybe prefer some mechanism maybe in `obspy.util` to globally deactivate specific warnings before running obpsy code.
> 
> Other opinions?

In general I very much share this opinion - I think we can just always defer people to use Python's `warnings` module - they can filter and hide to their hearts' content.

This particular case here is different though because it must be passed on to the underlying C `evalresp` routine which either prints it or not so catching it on the python side is really hard. We added this option at one point to be able to recompute the total sensitivity without that warning but never exposed it elsewhere.

megies:
Oh ok, I missed that point, skimming the PR.. fair enough then

obspy/core/inventory/response.py
```diff
@@ -1620,16 +1621,22 @@ def get_evalresp_response_for_frequencies(
         :type end_stage: int, optional
         :param end_stage: Stage sequence number of last stage that will be
             used (disregarding all later stages).
+        :type hide_sensitivity_mismatch_warning: bool
+        :param hide_sensitivity_mismatch_warning: Hide the evalresp warning
+            that computed and reported sensitivities don't match.
```
megies, line 1626:
```suggestion
            that computed and reported sensitivities do not match.
```

obspy/core/inventory/response.py
```diff
@@ -1654,6 +1661,9 @@ def get_evalresp_response(self, t_samp, nfft, output="VEL",
...
```
megies, line 1666:
```suggestion
            that computed and reported sensitivities do not match.
```

megies:
I guess you can directly pass this down through `**kwargs` all the way from `Stream.remove_response()`? If yes maybe add a short comment in that docstring

obspy/core/inventory/response.py
rizac: Note that I copied the comment from legacy code so the "don't" is still around

obspy/core/inventory/response.py
rizac: Just a side note: I copied the comment from legacy code so the "don't" is still around

rizac:
> I guess you can directly pass this down through `**kwargs` all the way from `Stream.remove_response()`? If yes maybe add a short comment in that docstring

I personally prefer to be explicit (which incidentally is also one of the - probably overused - Python dogmas ), given that I do not see major drawbacks (except that yes, maybe the parameter name is too long). so I would let the code as it is, if you do not mind

rizac:
Just a side note (to reply once to all the comments in the thread). First of all, we use ObsPy as library to process large amount of data, so we absolutely need to control the output in our software. A terminal (or log file) polluted with, say, 10,000 printouts of the same line of text is obviously more than a nuisance.

Now, with that in mind, the solution proposed above to deactivate globally specific (when not all) warnings is quite catchy, and it would be a great solution to implement.

But this does not solve all cases: as you realised, the problem here is to control **external** libraries output. And that's not easy. The fix we implement for instance [taken from here](https://stackoverflow.com/questions/5081657/how-do-i-prevent-a-c-shared-library-to-print-on-stdout-in-python/14797594#14797594) is quite convoluted. I recall for instance also having some similar problems in other circumstances (reading miniSEEd maybe?) but I did not find any easy flag in that case, as in this one. So if there is anything I can do to help adding flags also other parts of code with these kind of issues (maybe with someone addressing me to the relevant bits of code), I would like to help.

krischer:
There is this external package here which makes this a lot easier: https://github.com/minrk/wurlitzer

But in general I agree - this should be dealt with but sometimes is just really a lot of work for little perceived benefit.

`obspy.io.mseed` uses `libmseed`'s printing callback functions to convert all log output, warnings, and exceptions to their Python equivalents so they can be caught with the normal means - I think this is by far the cleanest solution but not all external libraries we use in ObsPy have that option.

I'm fine with having this particular option here. Evalresp's days in ObsPy are hopefully also numbered once this here is merged: https://github.com/obspy/obspy/pull/2592

megies:
> > I guess you can directly pass this down through `**kwargs` all the way from `Stream.remove_response()`? If yes maybe add a short comment in that docstring
> 
> I personally prefer to be explicit (which incidentally is also one of the - probably overused - Python dogmas ), given that I do not see major drawbacks (except that yes, maybe the parameter name is too long). so I would let the code as it is, if you do not mind

I was on about how this gets passed down from `Trace.remove_response()` nad yes it seems via kwargs, which is fine for me. Maybe it would be good to mention it very briefly in the `Trace.remove_response()` docstring that it talking about how details get passed down to evalresp methods.

megies:
@rizac we have `CatchOutput` in `core/util/misc.py`, not sure about it's current status though, but it should be able to catch most stuff I think.

megies:
Merging now, but as I said @rizac still feel free to add yourself to CONTRIBUTORS file if you want.

rizac:
> Merging now, but as I said @rizac still feel free to add yourself to CONTRIBUTORS file if you want.

I am already there ;-) 
Thanks

Issue #2688, opened 2020-07-30
gh actions: add workflows on master (on pushes)


megies:
Following up on these comments in #2591: https://github.com/obspy/obspy/pull/2591#issuecomment-666370888

---

`validate` workflow is targeting PRs exclusively. @d-chambers could you please make two more workflows that run on pushes to master?
 - one with full build matrix, only doing non-network tests
 - one with only one build, but including network tests

Reason to have two is kinda some of the discussion in pyOpenSci/software-review#16
 - we want network tests run regularly in CI, to have them in our [test reports](tests.obspy.org)
 - people are confused by the red badge in README

That way we could use the more stable test result excluding volatile network tests in the README but still have network tests run in CI.

d-chambers:
Sure, I should have time today or tomorrow to work on this

megies:
likely obsolete.. dunno

PR #2689, opened 2020-07-30
Gh actions master

d-chambers:
### What does this PR do?

Adds workflows for running the default test suite on the supported OS/python version and the network test suite on Linux and one python version. Also updated the build badge. 

Requested in #2688.  

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/workflows/master_default_tests.yml
+++ b/.github/workflows/master_default_tests.yml
@@ -8,7 +8,7 @@ on:
 
 # Set globally scoped environmental variables.
 env:
-  OBSPY_GITHUB_BRANCH: gh-actions
+  OBSPY_GITHUB_BRANCH: release_0.10.0
 
 jobs:
   run_tests:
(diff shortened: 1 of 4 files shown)


d-chambers:
Hmmm... failures are related to basemap's use of the matplotlib cook book.

.github/workflows/master_default_tests.yml
```diff
@@ -0,0 +1,63 @@
...
+# It simply runs all of ObsPy's default tests on all support OS/python versions
+name: master_default_tests
+on:
+  push:
+    branches:
+      - master
+
+# Set globally scoped environmental variables.
+env:
+  OBSPY_GITHUB_BRANCH: gh-actions
```
megies, line 11: will this need to be changed later?

megies:
Theose basemap fails we cant do much about, anyway not related to this PR. There's some other fails so that are troublesome, but again, not related.

Go ahead and merge whenever you think this is ready, we'll only see it in effect after merging anyway.

.github/workflows/master_default_tests.yml
d-chambers: You mean if I make a proper version branch for obspy_github_api? Yes that would be good to do :smile:

d-chambers:
Ok. One thing that bothers me is the amount of repetition in these actions. Supposedly the GH action folks are working on something to help this but there isn't a good way to do it yet. I will refactor later when there is a clean way to write more precise GH actions.

PR #2694, opened 2020-08-03
Make deep and shallow copies in Inventory add consistent

### LINKED ISSUE #2675: Inventory __add__ mixes deep and shallow copies
Adding two inventories using the `__add__` operator leads to an inventory consting of a deep copy of the first inventory, but a shallow copy of the second inventory. For example `inv1 + inv2` returns a deep copy of `inv1`, but a shallow copy of `inv2`. This also implies that runtime of inventory addition can change substantially by swapping the summands, if the two inventories differ strongly in size. The function in question is implemented here: https://github.com/obspy/obspy/blob/95b3adbe138bbb9460d3d5dc779660df35daadbb/obspy/core/inventory/inventory.py#L168-L184

I'd suggest either using deep or shallow copies of both inventories. Personally I'd prefer shallow copies for performance reasons and because they leave the option of making a deep copy to the user. I could write a PR, if this should be changed.

yetinam:
### What does this PR do?

Makes use of deep and shallow copies for inventory addition consistent, by consequently using shallow copies. In addition, `Inventory.extend` now copies Inventory metadata when given an Inventory as argument. Makes minor fixes to documentation.

Fixes #2675, including the suggestions from @trichter.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,10 @@
+maintenance_1.2.x
+=================
+
+Changes:
+ - obspy.core:
+   * Inventory addition now consistently uses shallow copies (#2675, #2694)
+
 1.2.2 (doi: 10.5281/zenodo.3921997)
 ===================================
 
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -304,11 +304,11 @@ class Inventory(ComparingObject):
 
     def extend(self, network_list):
         """
-        Extends the current Catalog object with a list of Network objects.
+        Extends the current Inventory object with another Inventory or a list of Network objects.
         """
         if isinstance(network_list, list):
             for _i in network_list:
-                # Make sure each item in the list is a event.
+                # Make sure each item in the list is a Network.
                 if not isinstance(_i, Network):
                     msg = 'Extend only accepts a list of Network objects.'
                     raise TypeError(msg)
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -315,6 +315,7 @@ class Inventory(ComparingObject):
             self.networks.extend(network_list)
         elif isinstance(network_list, Inventory):
             self.networks.extend(network_list.networks)
+            self.__copy_inventory_metadata(network_list)
         else:
             msg = 'Extend only supports a list of Network objects as argument.'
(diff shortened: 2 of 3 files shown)


trichter:
Thanks for the fix and the PR! I think it can be merged.

krischer:
I agree. Simple but comprehensive fix. It will have an impact on performance but I'm not sure how to otherwise deal with this and we are more concerned with correctness here.

trichter:
> It will have an impact on performance but I'm not sure how to otherwise deal with this and we are more concerned with correctness here.

Hmh, on the other hand side, most often a deepcopy is not wanted. And consequently, one had to do the same for catalog, which does not perform a deepcopy for these operations. On a second thought I am for changing the behavior to not create a deepcopy at all. What is your opinion? This behavior could also be documented in `__(i)add__` which shows up in the documentation.
@krischer @megies

yetinam:
I agree with @trichter and personally prefer shallow copies as well, as I have a project with large inventories including responses, where addition takes several minutes. In addition deep copies can easily be made by the user, if required. I guess we can put it into the documentation of `__(i)add__` but I suspect few people will read the documentation of these functions.

megies:
I don't know, I think shallow copies are dangerous here, they can trip up users badly, I know of somebody running into this issue personally just last week and taking a long while figuring out what was going on.

If users see `"copy"` they usually expect a deepcopy I would say.

megies:
Huh, nevermind, I just had a *proper* look.. this isn't about copying like I thought somehow, this is just about the `+` operator..

`ìnv = inv1 + inv2`

It should just return a shallow copy of `inv1` with it's networks extended by `inv2.networks`, I think. So everything shallow (like e.g. `stream = stream1 + stream2` only mashes together the traces)

yetinam:
I changed the PR to consistently use shallow copies now.

megies:
We'll need a changelog one liner @yetinam and I'd like to hear out @krischer and @trichter again, since this is touching our core. From my point of view it's looking good. Not sure if all the syntax you use in the tests works on 2.7, since this is targeting maintenance branch. To get CI results the new way, it might be worth a thought to target master branch and backport from there, maybe, but I don't care either way.

yetinam:
Thanks for the feedback. I just added the changelog statement.

PR #2695, opened 2020-08-07
Sc3ml response

jdassink:
### What does this PR do?

Modification so that _read_sc3ml continues parsing the XML file instead of exiting, when errors in the response part of the SC3 XML are encountered. The response information is left out of the Inventory.

I often come across these issues when parsing SC3 XML files that have incomplete PAZ information, for example when working with SC3XML files from the International Data Centre (CTBTO). 

Example error:

```
  File "obspy-test-sc3ml_response.py", line 2, in <module>
    inv=_read_sc3ml('IM.I31KZ_SC3_formatted.xml')
  File "/Users/assink/infrasound/development/obspy_dev/obspy-master/obspy/io/seiscomp/inventory.py", line 178, in _read_sc3ml
    net_element, _ns))
  File "/Users/assink/infrasound/development/obspy_dev/obspy-master/obspy/io/seiscomp/inventory.py", line 233, in _read_network
    sta_element, _ns))
  File "/Users/assink/infrasound/development/obspy_dev/obspy-master/obspy/io/seiscomp/inventory.py", line 297, in _read_station
    channel, _ns))
  File "/Users/assink/infrasound/development/obspy_dev/obspy-master/obspy/io/seiscomp/inventory.py", line 449, in _read_channel
obspy.core.util.obspy_types.ObsPyException: Could not find response tag with public ID 'ResponseFAP#20200807061628.711794.14'.
```

Exiting prematurely (what happens now) is somewhat drastic as the import to FDSN StationXML is still very useful. The correct Response information can always be added to FDSN StationXML file.

### PR Checklist
- [x] All tests still pass.


obspy-test-sc3ml_response.py
```diff
@@ -0,0 +1,3 @@
+from obspy.io.seiscomp.inventory import _read_sc3ml
+inv=_read_sc3ml('IM.I31KZ_SC3_formatted.xml')
+print(inv)
```
megies, line 3: This needs to be refactored into a proper unittest method in `obspy/io/seiscomp/tests/test_inventory.py`

IM.I31KZ_SC3_formatted.xml
```diff
@@ -0,0 +1,792 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<seiscomp xmlns="http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.8" version="0.8">
```
megies, line 2: This file needs to be stripped down significantly. It should only contain 2-3 stations with 1-3 channels each, so that the size is reduced as much as possible.

obspy/io/seiscomp/inventory.py
```diff
@@ -446,12 +457,23 @@ def _read_channel(instrumentation_register, cha_element, _ns):
         if len(response_elements) == 0:
             msg = ("Could not find response tag with public ID "
                    "'{}'.".format(response_id))
-            raise obspy.ObsPyException(msg)
+            #raise obspy.ObsPyException(msg)
+            print ('{} - Response ERROR:'.format(seed_id))
+            print (msg)
+            print (' -> Omitting response information from Inventory')
+            print ('')
```
megies, line 464: This needs to be changed into a `warning` being issued rather than printing to stdout.

IM.I31KZ_SC3_formatted.xml
jdassink: OK. File can be found in `tests/data/IM.I31KZ.sc3ml`

jdassink: OK.

obspy-test-sc3ml_response.py
jdassink: A test is now added to `obspy/io/seiscomp/tests/test_inventory.py`:

```
    def test_omit_response(self):
        """
        Test SC3ML inventory with response information that leads to ObspyException
        Read up to channel level and omit response information
        """
        inv = read_inventory(os.path.join(self.data_dir,
                                          "IM.I31KZ.sc3ml"))
        self.assertEqual(inv[0].code, "IM")
        self.assertEqual(inv[0][0].code, "I31H1")
        for cha in inv[0][0].channels:
            self.assertTrue(cha.code in ["BDF", "LWD", "LWS", "LKO", "LDA"])

```

jdassink: By the way, I noticed that the  test  `def test_stage_empty_poles_and_zeros(self):` gave errors, but it also seems to give errors in the obspy/master branch...

obspy/io/seiscomp/inventory.py
jdassink: Fixed it.

jdassink:
> I can see the usefulness of this PR @jdassink 👍, but it will need some changes, see comments.
> 
> It will also need a changelog entry and we'll need to clean up the branch at some point, to get rid of that merge commit. A simple `git rebase master` on the branch should do it, but we can also do that at the very end if you prefer.

Thanks! I did not enter a changelog entry because it was a minor change, but could do it. What do you mean by `git rebase master`? It yields `fatal: invalid upstream 'master'` when I run it in my branch (this is the first time I contribute to something on GitHub... be gentle ^__^

obspy/io/seiscomp/tests/test_inventory.py
```diff
@@ -306,6 +306,17 @@ def test_stage_empty_poles_and_zeros(self):
...
+        """
+        Test SC3ML inventory with response information that leads to ObspyException
+        Read up to channel level and omit response information
+        """
+        inv = read_inventory(os.path.join(self.data_dir,
+                                          "IM.I31KZ.sc3ml"))
+        self.assertEqual(inv[0].code, "IM")
+        self.assertEqual(inv[0][0].code, "I31H1")
+        for cha in inv[0][0].channels:
+            self.assertTrue(cha.code in ["BDF", "LWD", "LWS", "LKO", "LDA"])
```
megies, line 319: This is not making sure all those channels are actually present.

```
expected = ["BDF", "LWD", "LWS", "LKO", "LDA"]
got = [cha.code for cha in inv[0][0].channels]
self.assertEqual(expected, got)
```

Ideally you also want to test that `.response` is `None` for the invalid ones and that there is a response for the valid ones.

obspy/io/seiscomp/tests/test_inventory.py
```diff
@@ -306,6 +306,17 @@ def test_stage_empty_poles_and_zeros(self):
         self.assertEqual(zeros, [])
         self.assertEqual(poles, [])
 
+    def test_omit_response(self):
+        """
+        Test SC3ML inventory with response information that leads to ObspyException
+        Read up to channel level and omit response information
+        """
+        inv = read_inventory(os.path.join(self.data_dir,
+                                          "IM.I31KZ.sc3ml"))
```
megies, line 315: looks like this can fit on the previous line

obspy/io/seiscomp/inventory.py
```diff
@@ -457,20 +457,18 @@ def _read_channel(instrumentation_register, cha_element, _ns):
...
                    "public ID '{}'.".format(response_id))
-            #raise obspy.ObsPyException(msg)
-            print ('{} - Response ERROR:'.format(seed_id))
-            print (msg)
-            print (' -> Omitting response information from Inventory')
-            print ('')
+            warnings.warn(msg)
+            msg = ("Omitting response information from Inventory for "
+                    "'{}'.".format(seed_id))
+            warnings.warn(msg)
```
megies, line 471: We want to have those two messages combined into a single warning with two sentences/lines, as it is also logically only one event to message about.

obspy/io/seiscomp/inventory.py
```diff
@@ -457,20 +457,18 @@ def _read_channel(instrumentation_register, cha_element, _ns):
...
                    "'{}'.".format(response_id))
-            #raise obspy.ObsPyException(msg)
-            print ('{} - Response ERROR:'.format(seed_id))
-            print (msg)
-            print (' -> Omitting response information from Inventory')
-            print ('')
+            warnings.warn(msg)
+            msg = ("Omitting response information from Inventory for "
+                    "'{}'.".format(seed_id))
+            warnings.warn(msg)
```
megies, line 463: see below, combine please

megies:
> What do you mean by `git rebase master`? It yields `fatal: invalid upstream 'master'` when I run it in my branch (this is the first time I contribute to something on GitHub... be gentle ^__^)

My bad, it should have been `git rebase origin/master` (depending on what your fork is named as, can check with `git remote -vv`). If you do that you'll have to `git push -f` to rewrite history of the remote branch on your fork.

megies:
> I did not enter a changelog entry because it was a minor change, but could do it

Yes, please do so. It is a minor change but it *does* change how users would handle such cases in their code, it isn't very likely that somebody was catching these exceptions and handling them in a custom way, but we always want to take the possibility into account.
We prefer our changelog to be comprehensive, so users really can be sure that if there's nothing in the changelog, then nothing changes for them in specific parts of the code.

obspy/io/seiscomp/tests/test_inventory.py
jdassink: Added these lines as well as a check that the response stages are empty.

obspy/io/seiscomp/inventory.py
jdassink: Fixed it.

obspy/io/seiscomp/inventory.py
```diff
@@ -444,14 +455,23 @@ def _read_channel(instrumentation_register, cha_element, _ns):
                 response_elements.append(found_response)
 
         if len(response_elements) == 0:
-            msg = ("Could not find response tag with public ID "
-                   "'{}'.".format(response_id))
-            raise obspy.ObsPyException(msg)
+            msg = ("Found multiple matching response tags with the same "
+                   "public ID '{response}'. "
```
megies, line 459: You got hit by copy/pasta there ;-)

megies:
Two things left:
 - clean up flake8 (see `lint_code` CI action)
 - test the warning message being shown, see e.g. [here](https://github.com/obspy/obspy/blob/79558d2169e1f63b9f73d60d48f383d497c6ffb0/obspy/io/xseed/tests/test_core.py#L567-L574) using our helper context manager for that purpose

obspy/io/seiscomp/inventory.py
jdassink: Ah yes, the first warning message ``"Could not find response tag with public ID '{response}'"`` has been restored now for when ``len(response_elements) == 0``

jdassink:
> Two things left:
> 
> * clean up flake8 (see `lint_code` CI action)
> * test the warning message being shown, see e.g. [here](https://github.com/obspy/obspy/blob/79558d2169e1f63b9f73d60d48f383d497c6ffb0/obspy/io/xseed/tests/test_core.py#L567-L574) using our helper context manager for that purpose

Man, that flake8 is precise. Even more precise than the version installed on my computer. Anyway, that seems fixed.

What exactly do you mean with the second bullet point? Should I adjust the `test_inventory.py`?

obspy/io/seiscomp/inventory.py
```diff
@@ -458,16 +458,16 @@ def _read_channel(instrumentation_register, cha_element, _ns):
...
             warnings.warn(msg)
             response_element = None
         elif len(response_elements) > 1:
             msg = ("Found multiple matching response tags with the same "
                    "public ID '{response}'. "
                    "Omitting response stage information from Inventory for "
-                   "channel '{channel}'.".format(response=response_id,channel=seed_id)
-                   )
+                   "channel '{channel}'.".format(response=response_id, 
+                                                    channel=seed_id))
```
megies: remove 4 spaces to fix indentation, see [PEP8](https://www.python.org/dev/peps/pep-0008). same above

megies: nevermind, somehow i looked at old PR code

megies:
> What exactly do you mean with the second bullet point? Should I adjust the `test_inventory.py`?

Ideally, you should add a test in `test_inventory.py` which does a `read_inventory` call that is emitting the warning message you have in this PR and makes sure this warning is shown. We have a context manager to make this easy, see the link in above comment.

jdassink:
> > What exactly do you mean with the second bullet point? Should I adjust the `test_inventory.py`?
> 
> Ideally, you should add a test in `test_inventory.py` which does a `read_inventory` call that is emitting the warning message you have in this PR and makes sure this warning is shown. We have a context manager to make this easy, see the link in above comment.

Hey, still a bit puzzled with this. The edit in the code lead to suppression

jdassink:
Oops, I think I "closed" this

jdassink:
> > What exactly do you mean with the second bullet point? Should I adjust the `test_inventory.py`?
> 
> Ideally, you should add a test in `test_inventory.py` which does a `read_inventory` call that is emitting the warning message you have in this PR and makes sure this warning is shown. We have a context manager to make this easy, see the link in above comment.

I am afraid I still don't get it. I got rid of this warning by editing the code...

I tried adding a test using your example that shows the warning that I report on in the PR:

```
    def test_warning_when_response_is_erroneous(self):
        filename = os.path.join(self.data_dir, "IM.I31KZ.sc3ml")
        # Fail if responses are explicitly not skipped.
        # Otherwise continue, but raise a warning.
        msg = ("obspy.core.util.obspy_types.ObsPyException: "
               "Could not find response tag with public ID "
               "'ResponseFAP#20200807061628.711794.14'.")
        # This triggers a number of warnings as the file is constructed and
        # misses all kinds of information.
        with CatchAndAssertWarnings(expected=[(Warning, msg)]):
            read_inventory(filename)

```

but this test returns:

```
FAIL: test_warning_when_response_is_erroneous (__main__.SC3MLTestCase)
  File "test_inventory.py", line 321, in test_warning_when_response_is_erroneous
    read_inventory(filename)
  File "/Users/assink/opt/miniconda3/envs/py37/lib/python3.7/site-packages/obspy/core/util/base.py", line 751, in __exit__
    self._assert_warnings()
  File "/Users/assink/opt/miniconda3/envs/py37/lib/python3.7/site-packages/obspy/core/util/base.py", line 786, in _assert_warnings
    raise AssertionError(msg % (category.__name__, regex))
AssertionError: Expected warning not raised: (Warning, obspy.core.util.obspy_types.ObsPyException: Could not find response tag with public ID 'ResponseFAP#20200807061628.711794.14'.)
```

Could you help?

megies:
@jdassink the thing is, your test should hit the condition that a response is not available and then emit that warning. The context manager catches all warnings inside the block and at the end of the block checks if the expected warning message/type is among all encountered warnings. So either your warning doesn't get emitted as expected or maybe the string you are lokking for doesn't match exactly.

If you use pytest (can be installed via conda), you could to `pytest -pdb test_inventory.py` and then check what warnings were caught in your context block. Or  you could simply run the code in your test case in an interactive shell and check if the warning gets shown

megies:
To check it you could also make the expected message match more via regex.. just to check

```
        msg = ".*Could not find response tag.*"
```

claudiodsf:
Hi @jdassink!

Sorry for coming back to this PR after such a long time. 

If your are still interested, could you please rebase the PR against the current `master` branch and force-push?
We will then see if the tests pass and then come back to the question raised by @megies.

Thanks!

ThomasLecocq:
I believe this has been closed by the #2962 , adding the "should be closed" tag :-)

PR #2696, opened 2020-08-10
Remove extra executable bits and shebangs

QuLogic:
### What does this PR do?

These files have no code to run, so remove the executable bit and shebang.

Some files do have a doctest boilerplate, but no doctests to run, so remove the boilerplate too. I left the shebang and boilerplate in files with actual doctests, though actually, I think we should remove the shebang in those files too since they were never marked executable.

### PR Checklist
- [] If the PR is making changes to documentation, docs pages can be built automatically.
- [] If any network modules should be tested for the PR, add them as a comma separated list
- [x] All tests still pass.
- [] Any new features or fixed regressions are be covered via new tests.
- [] Any new or changed features have are fully documented.
- [??] Significant changes have been added to `CHANGELOG.txt` .
- [] First time contributors have added your name to `CONTRIBUTORS.txt` .


megies:
I fully agree with the executable bit / shebang stuff. With the superfluous  doctest parts, I'm a bit torn. Obviously they should in principle be removed, but I think it bears the danger of people copy/pasting to create new submodules and then failing to add the doctest part again (if there are doctests anyway). It's easy to get overlooked in PR reviews

QuLogic:
TBH, I'd rather remove all the doctest boilerplate altogether. The files aren't executable, so you have to run `python path/to/file.py` anyway, then you might as well run `python -m doctest path/to/file.py`, and for installed files, you could use `pytest` (bit unfortunate to have to use a different command there though.)

megies:
Actually yeah, I kinda agree, we might as well get rid of all these

ThomasLecocq (approved):

megies:
sorry for not merging earlier

Issue #2685, opened 2020-07-27
About the function inv.plot_response


AmbroseTeo:
I use the IRIS_Webinar files to learn how to use ObsPy in my MAC, but there is a problem that confused me a lot.

> inv.plot_response(min_freq=0.001,station="637A")

Then error occurred:

> No handles with labels found to put in legend.
> ---------------------------------------------------------------------------
> ValueError                                Traceback (most recent call last)
>  in 
>       1 import numpy as np
> ----> 2 inv.plot_response(min_freq=0.001,station="637A")
> 
> /usr/local/lib/python3.7/site-packages/obspy/core/inventory/inventory.py in plot_response(self, min_freq, output, network,  station, location, channel, time, starttime, endtime, axes, unwrap_phase, plot_degrees, show, outfile)
>     884         if not axes:
>     885             from obspy.core.inventory.response import _adjust_bode_plot_figure
> --> 886             _adjust_bode_plot_figure(fig, plot_degrees, show=False)
>     887         if outfile:
>     888             fig.savefig(outfile)
> 
> /usr/local/lib/python3.7/site-packages/obspy/core/inventory/response.py in _adjust_bode_plot_figure(fig, plot_degrees, grid,   show)
>    1966         minmax2 = ax2.yaxis.get_data_interval()
>    1967         yticks2 = np.arange(minmax2[0] - minmax2[0] % (pi / 2),
> -> 1968                             minmax2[1] - minmax2[1] % (pi / 2) + pi, pi / 2)
>    1969         ax2.set_yticks(yticks2)
>    1970         ax2.set_yticklabels([_pitick2latex(x) for x in yticks2])
> 
> ValueError: arange: cannot compute length

What should I do?
Thanks!

AmbroseTeo:
>No handles with labels found to put in legend.
>---------------------------------------------------------------------------
>ValueError                                Traceback (most recent call last)


and


>ValueError: arange: cannot compute length

megies:
You'll have to provide a [reproducable example](http://sscce.org/), or it will cost way more time for people to debug.

megies:
OK, nevermind, it's easy to reproduce, the problem occurs when the specified filters do not match any channel in the Inventory.

```python
from obspy import read_inventory
inv = read_inventory()
inv.plot_response(min_freq=0.001,station="ABCDE")
```

megies:
Should be an easy fix, I guess we'll just want to raise an exception with a meaningful message in that case.

[...]

PR #2701, opened 2020-08-19
Fix plot_response method to fail gracefully

gbbofh:
### What does this PR do?

The `plot_response` methods of the `Inventory` and `Network` classes, as well as the `plot` method of the `Station` class all cause `_adjust_bode_plot_figure()` to raise a `ValueError` exception when no matching channels are found. This pull request fixes that problem by checking whether or not the `select` method was able to return a non-empty list, and instead emits a warning to the user while displaying an empty plot.

This pull request was initiated to fix a bug reported by [Issue #2685](https://github.com/obspy/obspy/issues/2685), to mitigate an undesirable exception without properly notifying the user of what actually occurred.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -3,6 +3,7 @@ Allgeyer, Sebastien
 Ammon, Charles J.
 Antunes, Emanuel
 Arnarsson, Ólafur St.
+Bates, Garrett
 Bagagli, Matteo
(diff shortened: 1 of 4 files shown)


dsentinel:
I think this could use a test.
and
Update the doc string.

gbbofh:
I have updated the docstring summaries for the method and added test cases to the network and inventory test suites.

Issue #2686, opened 2020-07-29
xseed Parser: Misleading error message when trying to read non existent filename path


Alec-Mileman:
I've noticed whilst using the Parser function with .dlseed file type that the error 'First byte of data must be in [0-9<]' occurs when an incorrect directory name is entered. This can make it hard to identify whether your file type is corrupt, or that you've entered a wrong directory name

I'd like to make the developers aware, so they can include an error line that identifies an incorrect directory to the user.

heavelock:
What do you mean by incorrect directory name? Inexistent one? 
Could you provide a small, easily [reproducible example ](http://sscce.org/)

Alec-Mileman:
'''
from obspy.io.xseed import Parser
import obspy as ob

mseedpath = '/home/code/my_mseedfile.mseed'
dlseedpath = '/home/code/my_dlseedfile.dlseed'

st = ob.read(mseedpath)
parser = Parser(dlseedpath)
print(st)
'''

If you have a dlseed file in the above directory and change the file name to '123.dlseed', which doesn't exist, you'll get the error stated above, rather than an error indicating that the file directory does not exist.

megies:
I can confirm this, we (had?) similar misleading errors when trying lo load a non existent path with lxml (e.g. to read StationXML).

This should ideally be fixed by centralizing file opening, I think we should be able to use `_generic_reader()` function for this from `obspy.core.util.base`

megies:
See #2062 for a similar fix and we should reuse the same error message and exceptions as done there

gbbofh:
I've started looking into resolving this issue.

Issue #1978, opened 2017-11-01
Old git shows unexpected diff (CRLF vs LF)


trichter:
Unexpected output of `git diff` in a fresh clone of obspy due to #1964 (Ubuntu 16.04):

```
(workhorse) eule@kiste:~/dev/tmp$ git clone https://github.com/obspy/obspy.git
Cloning into 'obspy'...
remote: Counting objects: 103834, done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 103834 (delta 0), reused 2 (delta 0), pack-reused 103830
Receiving objects: 100% (103834/103834), 104.52 MiB | 5.61 MiB/s, done.
Resolving deltas: 100% (72854/72854), done.
Checking connectivity... done.
(workhorse) eule@kiste:~/dev/tmp$ cd obspy
(workhorse) eule@kiste:~/dev/tmp/obspy$ git diff
warning: CRLF will be replaced by LF in obspy/core/data/example.npz.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/data/example_py2.pickle.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/data/example_py3.pickle.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/data/ffbx_unrotated_gaps.mseed.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/basemap_combined_stations-events.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-basemap1.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-basemap2.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-basemap3.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-basemap_long-wrap.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-cartopy1.png.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in obspy/core/tests/images/catalog-cartopy2.png.
The file will have its original line endings in your working directory.
[...]

PR #2703, opened 2020-08-24
Fix misleading error in XSEED parser

gbbofh:
### What does this PR do?

The XSEED parser will currently raise a misleading exception when a local filename is given, if the file does not exist. This error occurs because `Parser.read()` will attempt to parse strings as RESP, DLSEED or XSEED data, if a local file does not exist, or if the string does not match a URI scheme, or does not match the start of a RESP file.

This pull request will fix the erroneous exception while attempting to maintain backwards compatibility by:

1) Checking if the `data` parameter to the `Parser` constructor is of type `str` or `native_str`, and if so, checks if the first byte is valid for a DLSEED, XSEED or RESP file using `re.match()`. If the match succeeds, `data` is encoded as a `bytes` object, and passed to `Parser.read()` directly.

2) Forwarding all other types of data which are not `None` to `obspy.core.util.base._generic_reader()`, with `Parser.read()` acting as the callback function.

This pull request was initiated in response to []Issue #2686(https://github.com/obspy/obspy/issues/2686).

### PR Checklist
- [ ] All tests still pass.


heavelock:
How about adding a test case? And also, those two first commits look odd, could you rebase on current master and force push?

gbbofh:
I'm having an issue attempting to rebase that seems to be caused by [Issue #1978](https://github.com/obspy/obspy/issues/1978). Once I have it sorted out I will attempt to rebase again and do a force push.

Other than that, I will add a test case here shortly as well.

megies:
>     1. Checking if the `data` parameter to the `Parser` constructor is of type `str` or `native_str`

`future` and thus `native_str` is removed in master for eventual 2.0 release. Two options:
 - keep PR as is and merge into `maintenance_1.2.x` instead
 - base PR on `master` and no need for Py2/3 compatibility then

Since this is a rather minor problem, I would vote to do the latter.

PR #2706, opened 2020-08-28
gh-actions: slightly verbose output for flake8

megies:
### What does this PR do?

Makes flake8 check print some slightly verbose information when linting code in CI / gh-actions.

Without that `-v` flag, the actual flake8 run shows no output at all if no linter doesn't find any problems, which can be confusing when looking at CI results because then "Successfully installed flake8" is the last thing that is shown in the CI log:

Screenshot from 2020-08-28 11-08-31

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           pip install wheel
           pip install flake8
-          python -m flake8 obspy
+          python -m flake8 obspy -v
 
   # Runs the tests on combinations of the supported python/os matrix.
   test_code:


d-chambers:
I am not opposed to this but  now it may be too verbose. For example, I could see the following confusing a new contributor (especially the "found a total of x violations 0 reported" bit):

image

Alternatively, we could just test flake8's return code and echo "No flake8 violations found" if it is 0. We could also move the python installation to a prior step so the flake8 window only gives relevant information on code linting.

QuLogic:
Maybe we can use reviewdog, like I added to Matplotlib? https://github.com/matplotlib/matplotlib/pull/17143

megies:
Automated comments inline from flake8 check? Nice, I'm all for it

d-chambers:
I haven't used reviewdog but from the documentation it looks good. I am ok with either cleaning up the linting job a bit or just using reviewdog.

QuLogic:
So this is replaced by #2717, I think.

megies:
Ok, sweet, thanks again @QuLogic!

PR #2722, opened 2020-10-06
Add Python 3.9 to test matrices

heavelock:
Python 3.9 is out so I added it to the test matrices for both PR and master workflows.

Changelog of 3.9:
https://www.python.org/downloads/release/python-390/


trichter:
You could also add the corresponding classifier in `setup.py`.

heavelock:
Done. Additionally, I noticed that the minimum version in the classifiers is 3.7 but the minimum version that was checked in the actual setup code was set to 3.6. Thus I upgraded it to 3.7 to be unified. Should it stay like that or was it on purpose?

trichter:
I think it should stay like that. Now, we only need to wait for the dependencies to appear on conda.

ThomasLecocq:
will be happy when this is approved :-) 
Ref: https://travis-ci.org/github/ROBelgium/MSNoise/jobs/738899087

setup.py
```diff
@@ -54,7 +54,7 @@
 
 
 # The minimum python version which can be used to run ObsPy
-MIN_PYTHON_VERSION = (3, 6)
+MIN_PYTHON_VERSION = (3, 7)
```
ThomasLecocq, line 57: just not sure about this one @megies @krischer , but for the rest the PR is clean & easy to review :-)

ThomasLecocq (commented):
just not sure about the `MIN_PYTHON_VERSION` thingy, for the rest should be OK

setup.py
megies: It's fine, but thanks for looking out :-)

megies:
Retriggering CI to see if all dependencies are in place

trichter:
@heavelock I think we should remove `basemap` and `future` from the test environment in `.github/test_conda_env.yml`.

Basemap support will be removed in the next major obspy version and it is also python3 only.

megies:
Fail to setup conda env on Python 3.9

heavelock:
The conda env still cannot be set up. I removed the basemap and future as @trichter suggested to see if it will help. 
I did not add the cartopy though, maybe I should have.

ThomasLecocq:
anything I can do here? Would be good to have the 3.9 working...

heavelock:
Well, there are conflicts in the dependencies, including some of them not having officially released python 3.9, for example jsonschema or wheel. 

Looking at that test environment I also found that:

`docopt` does not seem to be necessary. What is more, it was not updated since 2014 and latest supported python is 3.3.
```
$ grep -rnw . -e "docopt"                                                                                                                                                                                                                                                                                               
./.github/test_conda_env.yml:16:  - docopt
````

`nose` looks to be not necessary too. It's only installed everywhere but only place in the code that is ever mentioned says it is no longer necessary :P See `/obspy/core/util/testing.py:159`

I will go ahead and remove them.

**Comment by trichter** [comment]
> There are few errors, but nothing tragic.
> https://tests.obspy.org/112352/

Most (all?) of these are already fixed in master.

**Comment by megies** [comment]
Retriggering CI, to see if it builds by now.

But actually, I think this should eventually be rebased on master due to #2750

**Comment by heavelock** [comment]
Rebased, let's see now.

**Comment by d-chambers** [comment]
Hey @heavelock,

Thanks for tackling this! 3.9 support/testing is certainly important.

When I am debugging CI I tend to use my fork of ObsPy and a pull request which targets my fork so that not all 100+ watchers of ObsPy get notified each time I push a commit. It certainly isn't a problem for me to simply delete these emails but I wouldn't want anyone else to get annoyed and unwatch the ObsPy or have to manually unsubscribe from this PR thread. After the CI is green on my fork then I push the changes to the official ObsPy repo or open a PR against it.

**Comment by heavelock** [comment]
Oh dammit, I did not think that by default all the watchers are pinged on that. Anyway, I will move to testing that stuff on my own fork. Sorry for the spam...

**Comment by heavelock** [comment]
Changes in this PR:
- Changed test build system. No more conda and it's inability of properly solving dependencies. Tests are now using pip to download and install all the dependencies 
- I extracted step for parsing results of coverage in order for action to properly fail. The success/failure is determined based on exit code of last command so if the parsing of coverage was always last, it was always a success.
- I turned off fail-fast strategy for the testing matrix. Since we have failing image comparison tests on all the elements of matrix, sometimes it was impossible to get any information out of tests - before the relevant test failed, it was killed due to other tests finishing before.
- I removed doctest from #2786 and added python version explanation

I think it is ready for a proper review.

**Comment by d-chambers** [comment]
> No more conda and it's inability of properly solving dependencies.

Is this just kicking the can down the road? I mean we are going to try to make a conda build of the 3.9 version of ObsPy 1.x series right?

File: .github/workflows/pull_request.yml [hunk 2]
```diff
@@ -29,7 +29,6 @@ jobs:
           python-version: '3.x'
 
       - name: create ci config
-        shell: bash -l {0}
         run: |
           pip install git+https://github.com/obspy/obspy_github_api@${OBSPY_GITHUB_BRANCH}
```
d-chambers, line 33: Want to change this to `python -m pip` for consistency?

.github/workflows/pull_request.yml
```diff
@@ -107,19 +111,31 @@ jobs:
...
 
       - name: run test suite
         shell: bash -l {0}
         run: |
+          export PATH="$pythonLocation/bin/:$PATH"
           export MODULELIST=`obshub read-config-value module_list --path=${CONFIG_PATH}`
           export MODULELISTSPACES=`obshub read-config-value module_list_spaces --path=${CONFIG_PATH}`
           coverage run --rcfile=.coveragerc --source=${MODULELIST} -m obspy.scripts.runtests --no-flake8 -n gh-actions -r --ci-url="${CI_URL}" --pr-url="${PR_URL}" $MODULELISTSPACES
+      
+      - name: parse coverage results
```
d-chambers, line 131: Duplicate job?

obspy/core/utcdatetime.py
```diff
@@ -1417,17 +1417,13 @@ def isoweekday(self):
...
 
         :rtype: tuple of ints
-        :return: Returns a tuple containing ISO year, ISO week number and ISO
-            weekday.
-
-        .. rubric:: Example
-
-        >>> dt = UTCDateTime(2008, 10, 1, 12, 30, 35, 45020)
-        >>> dt.isocalendar()
-        (2008, 40, 3)
```
d-chambers, line 1430: Let's add this back but use the solution in #2786

obspy/core/utcdatetime.py
```diff
@@ -1417,17 +1417,13 @@ def isoweekday(self):
 
     def isocalendar(self):
         """
-        Returns a tuple containing (ISO year, ISO week number, ISO weekday).
+        Returns a named tuple containing (ISO year, ISO week number,
+        ISO weekday).
+        In Python version =< 3.8 it returns a standard tuple.
```
d-chambers, line 1422: We should be able to also revert this.

d-chambers (approved):
Looks good to me. Just a few minor changes. I am a little worried about the idea of abandoning conda as we might start to get more platform dependent issues creeping up using only pip but we can address those sorts of issues when they arise.

.github/workflows/pull_request.yml
heavelock: Nope, previously it was a just as a command following the testrunner command. I had issues that some failing tests were not shown as failed jobs so I extracted this parsing of coverage to a separate step

heavelock:
> Is this just kicking the can down the road? I mean we are going to try to make a conda build of the 3.9 version of ObsPy 1.x series right?

Hmm, in my experience pip got much better with solving dependencies and a lot of packages started routinely compiling wheels for different platforms so it shouldn't be such an issue anymore. Also, my recent experiments on this branch as well as on #2783 show that, at least currently, installing and building obspy on all major platforms within CI is not that problematic.

I tried to backport this testing workflow to maintenance. Unfortunately, it was not working because of the github api package. Now as I'm thinking, I will try to backport the workflow without the api integretion for simplicity and we will see If it will bring up some problems.

barsch:
@heavelock is this ready to merge? if so, can you rebase one more time - thanks

.github/workflows/pull_request.yml
heavelock: Done

heavelock:
Hey! Sorry for such a delay on this PR. I rebased on top of current master, did some minor changes (updating versions of actions used, unifying naming of steps, unifying pip calls) and force pushed.

As you can see, the tests are failing. This is due to the image comparison tests. For all of the test conditions, there are 46 failing tests and 92 instances of ImageComparisonException during the grep through logs - it means that only image comparison tests are failing. To be honest, I have no idea why this happens since on master, that sets up dependencies with conda, everything is fine. Maybe differences in how matplotlib is compiled for conda and for pip?

I tested also if conda finally came to it's senses and if it can finally resolve the dependencies. The answer is no. See results of the test here: https://github.com/heavelock/obspy/runs/2864189711

I think this is ready to merge but we should address somehow the issue of image comparison tests.

@barsch @d-chambers

barsch (approved):

ThomasLecocq:
hey all,

catching up with some needed PRs now, so could you update on the status if this one ? good to go ? 

@heavelock @d-chambers  ?

d-chambers:
I am not seeing a way to re-trigger the CI because this is older. Any chance you can push another change of some sort @heavelock ? I am wondering if we will see the same '_NoValueType' errors the master's CI is producing.

heavelock:
@d-chambers I pushed empty commit, let's see.

@ThomasLecocq As far as I remember those changes should be mergeable. As far as I remember I did not modify the master tests but those could be done in a separate PR.

heavelock:
@d-chambers The problem with macos is still here... :<

heavelock:
I've tested on two macos versions: macos-11 and macos-10.15. Tests fail in the same way on both of them. They are both x86-64 so it is not some random change to arm architecture. 

On positive note, Py3.10 seems to be working in the same way as previous versions - only image comparison tests fail.

trichter:
> What would you think about just disabling the image comparison tests?

I support disabling image comparison tests.

heavelock:
+1 from me. What do you think @ThomasLecocq ?

ThomasLecocq:
>We would still want to make sure the plotting code runs, just not check the image diffs.

I agree on the short term, but we should still try to make sure on the long term that we get the diffs back in...

.github/workflows/master_default_tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
```
ThomasLecocq, line 19: @heavelock why not '3.10' here (note the single quotes, I had issues earlier today with 3.10 being interpreted as 3.1 (best is to have them all single-quoted)

ThomasLecocq:
@heavelock please check my comment above, and PRs #2904 #2908 #2910 if they are relevant here? Then merge!

ThomasLecocq:
PRs #2904 #2908 #2910 are merged and all tests are 🟢 now. Maybe easier to remake a PR for adding 3.9 and 3.10 to the test matrix

Issue #2690, opened 2020-07-31
Test fails in UTCDateTime.matplotlib_date


megies:
Current master, see https://tests.obspy.org/111684/#3 and #2689 

e.g.
```
File "/home/runner/work/obspy/obspy/obspy/core/trace.py", line 2536, in obspy.core.trace.Trace.times
Failed example:
tr.times("matplotlib")
Expected:
array([ 733643.01392361, 733643.01392373, 733643.01392384, ...,
733643.01427049, 733643.0142706 , 733643.01427072])
Got:
array([ 14480.01392361, 14480.01392373, 14480.01392384, ...,
14480.01427049, 14480.0142706 , 14480.01427072])
```

QuLogic:
Probably the epoch change https://matplotlib.org/users/whats_new.html#dates-use-a-modern-epoch

megies:
Ah, thanks for the link, so basically everything's fine and we just have to use `set_epoch()` in our test suite.

megies:
Actually with all the different ways our tests can be executed, it might be a bit tricky to swap to the epoch that test results expect in all cases for doctests (unittests should be trivial, can just use `set_poch()` inside the test method)..
 - via `$ obspy-runtests`: can simply use `set_epoch()` inside `scripts/runtests.py`
 - via `$ pytest ...`: could maybe check for [`PYTEST_CURRENT_TEST`](https://docs.pytest.org/en/4.6.1/example/simple.html#pytest-current-test-environment-variable) inside `UTCDateTime.matplotlib_date()`  (although seems a bit convoluted)
 - via `$ python some_of_our_files.py` -- maybe no chance

kdalbey:
I encountered this problem USING obspy, there was existing code that use obspy that worked on linux and mac but to get it to compile/link on windows10 using anaconda I had to update a few libraries including matplotlib and obspy.    Anyway I encountered this problem and got stuck on it for several days, the code that called obspy was always doing a matplotlib.dates.date2num(<UTCDateTime>.datetime) to get a number comparable to res[:,0] coming out of array_analysis.array_processing() 

but a small change to this the array_analysis.array_processing() would transparently fix this issue for your users in a backwards compatible way (requires matplotlib >=2.1.2, which is when matplotlib.dates.date2num started taking numpy.datetime64 which was introduced in numpy 1.7 so numpy >=1.7 also required)

if you 
import matplotlib.dates as mdates
at the top of array_analysis.py and then make the following change circa line 1022

[...]

PR #2723, opened 2020-10-07
fix mpl epoch issue in array analysis

trichter:
### What does this PR do?

Fix https://github.com/obspy/obspy/issues/2690#issuecomment-686549268

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/signal/array_analysis.py
+++ b/obspy/signal/array_analysis.py
@@ -20,6 +20,7 @@ Functions for Array Analysis
 import math
 import warnings
 
+from matplotlib.dates import datestr2num
 import numpy as np
 from scipy.integrate import cumtrapz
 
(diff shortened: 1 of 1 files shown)


heavelock:
The same could be achieved by adding `matplotlib.dates.EPOCH_OFFSET` instead of `datestr2num`. As unlikely as change of epoch, this would tie us better to their system, I think.

trichter:
> The same could be achieved by adding matplotlib.dates.EPOCH_OFFSET instead of datestr2num. As unlikely as change of epoch, this would tie us better to their system, I think.

I checked it right now, it appears to be cleaner, *but* `EPOCH_OFFSET` always returns the offset of the old mpl epoch.

> And btw, both EPOCH_OFFSET and datestr2num('1970-01-01') return 719163.0 while in our code there is 719163.

That's the same number. If you update your matplotlib installation, EPOCH_OFFSET is still 719163.0 while datestr2num('1970-01-01') is 0.0.

heavelock:
Aaaaah, okay. Now I understand the issue. 

So I think the PR is good to go.

heavelock (approved):

megies:
Do we need a changelog entry for this? Looks like it might be good to have one

trichter:
OK, I can add a changelog entry.

PR #2728, opened 2020-10-15
add changelog for #2723

trichter:
I wanted to push directly to master, but I was not allowed. So, here is the changelog entry.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -23,6 +23,10 @@ Changes:
      encountered (see #2668)
  - obspy.clients.fdsn:
    * introduce fine-grained FDSN client exceptions (see #2653)
+ - obspy.signal.array_analysis
+   * fixed an issue in array_processing function returning wrong times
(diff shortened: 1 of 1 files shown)


megies:
That's weird.. I just checked and you are in @obspy/maintainers and that team should be able to make normal fast-forward pushes to master (no force pushes), according to the branch protection rules.

megies (approved):
feel free to merge this one or try to push commit directly on master again (should work in principle)

trichter:
Now pushing the master branch was working, but I suspect it only worked because you approved the commit here.
Previously an error about one needed approving review was displayed.

megies:
> Now pushing the master branch was working, but I suspect it only worked because you approved the commit here.

Hmm.. i still think you should be able to push fast-forwardable commits to master directly according to the settings. Well no big deal anyway

trichter:
I'll try again, the next time.

PR #2736, opened 2020-10-25
Open CSS binary waveform files compressed with gzip.

zacherle:
### What does this PR do?

io/css/core.py:
If the _read_css function does not find the waveform file referenced by
the wfdisc table, it will try to open a compressed .gz suffix file instead.
io/css/tests/test_core.py:
Add test_css_2_read_via_module method.
io/css/tests/data/:
New test data files 201101311155_2.be.w.gz, 201101311155_2.le.w.gz,
test_css_2.wfdisc.

This modification allows you to work with archived seismic records in CSS
format, where the binary waveform files are gzipped with the suffix .gz,
while the dfile pointer in the wfdisc table shows the names of the original
uncompressed files.
This method of storing continuous data is supported by the GEOTOOL software
(CTBTO).
It is useful for calculating PPSD statistics over a long period of time
directly from archived data without having to first extract all the records.

### PR Checklist
- [ ] All tests still pass.


heavelock (commented):
I'm not familiar with CSS file format so I have a question about that change. Is Is it kind of a multi-file format where a central file contains links to sub-files containing the data itself?

If yes, I think it is necessary to differentiate names of variables containing the `metafile`  filename (line 119/120)  and those containing data itself (line 127 and further ones)

megies:
> Is Is it kind of a multi-file format where a central file contains links to sub-files containing the data itself?

Yes, the main file basically only holds ASCII lines with trace infos and filenames where to find the data

https://github.com/obspy/obspy/blob/master/obspy/io/css/tests/data/test_css.wfdisc

megies:
This seems a bit hacky (by the mentioned tools storing the compressed data) but since you say this is used by some standard tools I guess we should add this.

Some comments:
 - [ ] please avoid code duplication, wrap the four common lines in the `with` blocks into a tiny (private, i.e. underscore prefixed) helper function
 - [ ] now if the files stated in the wfdisc table are not found it will show an error message for the gzip file. this might be slightly confusing to users. It might be better to reraise the error message from the un-suffixed filename (see below for suggestion)
 - [ ] changelog one liner
 - feel free to add yourself to `CONTRIBUTORS.txt`

zacherle:
> I'm not familiar with CSS file format so I have a question about that change. Is Is it kind of a multi-file format where a central file contains links to sub-files containing the data itself?
> 
> If yes, I think it is necessary to differentiate names of variables containing the `metafile` filename (line 119/120) and those containing data itself (line 127 and further ones)

Yes you're right. I was guided by the effort to only add functionality and not further improve the existing code,
but the fact is that the variable "filename" is inappropriately redefined during the _read_css function.
It would be more appropriate to name the variables as in the official CSS 3.0 specification.
eg here https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/2002/023055.pdf
I'm sending a commit, where I've added the wfdisc_dfile and dfile variables for clarity.
However, the mentioned programming impurity is also in the _read_nnsa_kb_core function, which I did not modify.

zacherle:
@megies  Thank you for reviewing the code.

> This seems a bit hacky (by the mentioned tools storing the compressed data) but since you say this is used by some standard tools I guess we should add this.

The described method of data storage is really quite hacking,
but it can be found where Antelope (BRTT) or Continuous Data Protocol (CTBTO) is used.
Additional compression of waveform files with gzip is quite effective in the case of low-noise short-period stations.
Assuming that analytical software can work directly with compressed data.
 
> Some comments:
> 
>     * [x]  please avoid code duplication, wrap the four common lines in the `with` blocks into a tiny (private, i.e. underscore prefixed) helper function
> 
>     * [x]  now if the files stated in the wfdisc table are not found it will show an error message for the gzip file. this might be slightly confusing to users. It might be better to reraise the error message from the un-suffixed filename (see below for suggestion)
>
>     * [x]  changelog one liner

I tried to modify the code according to your comments
and tried to be more pythonic here, if an old fortranist can be.

>     * feel free to add yourself to `CONTRIBUTORS.txt`
Next time.

obspy/io/css/core.py
```diff
@@ -134,11 +146,20 @@ def _read_css(filename, **kwargs):
...
             fmt = read_fmt
-        with open(filename, "rb") as fh:
-            fh.seek(offset)
-            data = fh.read(read_fmt.itemsize * npts)
-            data = from_buffer(data, dtype=read_fmt)
-            data = np.require(data, dtype=fmt)
+
+        try:
+            with open(dfilename, "rb") as fh:
+                data = _read_segment()
```
heavelock, line 152: This piece of code made me realize that I do not fully know how does the scope of variables work in Python. 

Anyway, I feel about not passing a arguments to a method directly and using variables from outside of its scope a bit unclear. But to be honest, I am not sure what is the obspy's policy on that.

obspy/io/css/ChangeLog.txt
```diff
@@ -0,0 +1,14 @@
+
+package obspy.io.css
+====================
+
+2020-10-28 pz
+Open CSS binary waveform files compressed with gzip.
```
heavelock, line 6: In terms of the changelog, I think @megies meant rather a one line in the `CHANGELOG.txt` located in main directory of repo :)

zacherle: Thanks for advice. Done.

obspy/io/css/core.py
zacherle: I rewrote it in a simple way to commit 0ab2244.

zacherle:
Can you advise? Which method of dev branch synchronization is more appropriate:

1) git rebase upstream / master and then git push -f
or
2) git merge upstream / master and then git push

The first method gives a clearer result but rewrites the history. That's why I chose the second method (commit f334ff8).

obspy/io/css/core.py
heavelock: I meant here to explicitly pass the fh as argument to read_segment but your solution also works fine :)

heavelock:
Usually in order to synchronise developmental branches with master we are rebasing but since you already did the merge, I would not fiddle with git anymore. 

I think it's good to go.

heavelock (approved):

obspy/io/css/core.py
```diff
@@ -134,11 +135,25 @@ def _read_css(filename, **kwargs):
...
-            data = from_buffer(data, dtype=read_fmt)
-            data = np.require(data, dtype=fmt)
+
+        try:
+            # assumed that the waveform file is not compressed
+            fh = open(dfilename, "rb")
+        except FileNotFoundError as e:
+            # If does not find the waveform file referenced in the wfdisc,
+            # it will try to open a compressed .gz suffix file instead.
+            try:
```
jkmacc-LANL, line 145: I haven't seen in the reference documentation for KB Core or [CSS 3.0](https://lanl-seismoacoustics.github.io/pisces/data/Anderson1990.pdf?raw=true) where adding ` .gz` to the `dfile` entry means that the waveform is gzip compressed, though it may be a convention.  This is normally done with the `datatype` entry in the `wfdisc` row, which determines how to interpret the bytes in the `dfile`, and gzip doesn't appear in the documentation for `datatype`.

zacherle: It is an operating practice that does not contradict official standards. Seismic records are continuously stored in 'css 3.0' format, waveform files uncompressed in the selected numeric representation (datatype 's4'). After a few days, the waveform files are gzip-compressed with a suffix '.gz' in a compression ratio of approximately 3:1. The 'fwdisc' table files remain intact, the 'dfile' link is not modified in any way. It is very practical if the software can work with such archived data directly and will take care of unzipping at the level of individual segments. This is supported by the CTBTO software - gbase library cssio, geotool. So I guess we are not the only ones who enjoy this method.

megies:
> Usually in order to synchronise developmental branches with master we are rebasing but since you already did the merge, I would not fiddle with git anymore.

Yes, we'll want to rebase (and force push) this before merging. Just a simple `git rebase -i upstream/master` should get rid of the merge commit.

CHANGELOG.txt
```diff
@@ -7,6 +7,8 @@ Changes:
      instrument response (see #2677)
  - obspy.clients.filesystem:
    * add get_waveforms_bulk() method to SDS client (see #2616, #2626)
+ - obspy.io.css:
+   * open CSS waveforms even if gzip-compressed
```
megies, line 11: Please add PR number

obspy/io/css/core.py
```diff
@@ -134,11 +136,26 @@ def _read_css(filename, **kwargs):
...
+            fh = open(dfilename, "rb")
+        except FileNotFoundError:
+            # If does not find the waveform file referenced in the wfdisc,
+            # it will try to open a compressed .gz suffix file instead.
+            try:
+                fh = gzip.open(dfilename + '.gz', "rb")
+            except FileNotFoundError:
+                msg = "Missing waveform file: {}".format(dfilename)
+                warnings.warn(msg, UserWarning)
+                continue
```
megies, line 151: It looks like previously an exception was raised when an expected file was not found. Now  you are showing a warning and continuing on. I think it would be better to stick to raising an exception, as this should be considered a "broken waveform file" in that case.

```python
try:
    ...
except FileNotFoundError as e:
    try:
        ...
    except FileNotFoundError:
        raise e
```

Simply reraising the first one should be fine. If you want to go the extra mile, you could modify the exception message and add that also no `*.gz` file was found either.

megies (changes requested):
I think we should keep raising an exception like before, otherwise looking good

obspy/io/css/core.py
zacherle: I am returning to the previous solution, where the function `_read_css` could raise `FileNotFoundError` exception.
By handling the error exception as warning, I tried to give the `_read_css` function a certain tolerance for data inconsistencies.
However, I understand that the core code policy is more strict.
Now, if the user does not want to disable data from the whole wfdisc file (= a whole day), he must first ensure the consistency of the data.

CHANGELOG.txt
zacherle: Done.

zacherle:
> Yes, we'll want to rebase (and force push) this before merging. Just a simple `git rebase -i upstream/master` should get rid of the merge commit.

I assume that I have to wait with rebasing until you decide to merge to be synchronized to the current state of main branch.
And what about squashing my multiple commits into one during interactive rebasing?
Is it possible here?

obspy/io/css/tests/test_core.py
```diff
@@ -118,6 +127,32 @@ def test_nnsa_kb_core_read_via_module(self):
...
+        Warns that waveform files are missing.
+        """
+        # 1
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always", UserWarning)
+            _read_css(self.filename_css_3)
+            # Two waveform files are missing
+            assert len(w) == 2
+            assert "missing_dtype_c0" in str(w[0].message)
+            assert "missing11155" in str(w[1].message)
```
megies: There's always a (slight) chance that this test might break later on in case an unexpected warning pops up (e.g. a simple deprection warning in numpy or some other dependency). We have a helper context wrapper with sophisticated testing of warnings, e.g. used here:

https://github.com/obspy/obspy/blob/e3b19478c2a63ab714b80c0f2efcff9e8b6b9cbb/obspy/io/stationxml/tests/test_stationxml.py#L283-L289

megies:
> I assume that I have to wait with rebasing until you decide to merge to be synchronized to the current state of main branch.
> And what about squashing my multiple commits into one during interactive rebasing?
> Is it possible here?

I only have one minor comment about testing the exception, other than that I think this should be good to go. :+1: 

Yes, please also rebase this on current master branch and force push. Feel totally free to squash commits together in the process if you like. For sure the "flake8" commits can/should be squashed together. For other commits, we are fine keeping individual commits if they show some semantic decision making of the process. That said, yes, I think this overall is a small enough change so that squashing everything together might be a good idea.

obspy/io/css/tests/test_core.py
zacherle: Thanks for pointing to the CatchAndAssertWarnings class and recommending good test code.
However, I do not have the opportunity to use it now, because I have returned to the original stricter solution, where the missing data segment is not tolerated by the warning message, but causes an error message and terminates the procedure. While this can interrupt many hours of calculation, it is the user's responsibility to keep the data in a consistent state.
It's not up to me to change the tolerance level of data errors in the core code.

megies (approved):

megies:
Thanks @zacherle :rocket:

zacherle:
Thank you @megies and @heavelock for help with my first PR. It was instructive and very beneficial for me.

PR #2739, opened 2020-10-28
FDSN: add short URL mapping for IRISPH5

### LINKED ISSUE #2737: Add IRISPH5 to the list of known FDSN web service implementations
IRISPH5 web services (https://service.iris.edu/ph5ws/) should be added to the list of known FDSN web service implementations. To access these services, one must currently supply the url, as well as the `service_mappings` argument when initializing the FDSN client.

```
irisph5_client = Client('http://service.iris.edu/ph5ws/', service_mappings={
     'dataselect': 'http://service.iris.edu/ph5ws/dataselect/1',
     'station': 'http://service.iris.edu/ph5ws/station/1'}
)
```

Obspy should be able to automatically discover this service by adding it to `obspy.clients.fdsn.header.URL_MAPPINGS` so that the client could be initialized by running

```irisph5_client = Client('IRISPH5')```

This would also make it nicer to use MassDownloader since it defaults to using the list of providers given in `URL_MAPPINGS`. To use IRISPH5 in the MassDownloader in combination with the other FDSN clients, one must first initalize the IRISPH5 client using the method above, and then create the MassDownloader using 

```
from obspy.clients.fdsn.mass_downloader import MassDownloader
from obspy.clients.fdsn.header import URL_MAPPINGS
mdl = MassDownloader(providers=[irisph5_client] + list(URL_MAPPINGS))
```

### PR DESCRIPTION
**Pull request by megies**
### What does this PR do?

Adds support for IRISPH5 short URL mapping in FDSNWS Client. To make it work it would need some changes to internals unfortunately, since this FDSNWS is not following the FDSN specifications in that it does not have a `/fdsnws` subpath in between the base URL and the service endpoints.

Two possibilities to alleviate it:
 - change internals as done in this PR. implies that users should be able to navigate service versions using `major_versions` on Client init.
 - other option: set the full service URLS in `headers.py` and leave internals as is. less changes in code, but then users would not be able to use the `"IRISPH5"` URL mapping and use an (eventual) higher service major version. (not done in this PR)

This isn't really beautiful, so I'd like to hear some opinions on this.

Tests pass locally: http://tests.obspy.org/112201/

+TESTS:clients.fdsn

closes #2737 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/mass_downloader.py
+++ b/obspy/clients/fdsn/mass_downloader/mass_downloader.py
@@ -55,11 +55,13 @@ class MassDownloader(object):
     implementations.
 
     :param providers: List of FDSN client names or service URLS. Will use
-        all FDSN implementations known to ObsPy except RASPISHAKE if set to
-        None. The order in the list also determines their priority, if data
-        is available at more then one provider it will always be downloaded
-        from the provider that comes first in the list. To include RASPISHAKE,
-        you must set this parameter to
+        all FDSN implementations known to ObsPy except RASPISHAKE (generally
+        worse quality data) and IRISPH5 (active source / nodal experiments that
+        might match a very large amount of data occasionally) if set to None.
+        The order in the list also determines their priority, if data is
+        available at more then one provider it will always be downloaded from
+        the provider that comes first in the list. To include RASPISHAKE and
+        IRISPH5, you must set this parameter to
         `obspy.clients.fdsn.header.URL_MAPPINGS` explicitly.
     :param debug: Debug flag passed to the underlying FDSN web service clients.
     :type providers: list of str or :class:`~obspy.clients.fdsn.client.Client`
@@ -95,6 +97,11 @@ class MassDownloader(object):
                 del providers["ORFEUS"]
             else:
                 has_orfeus = False
+
+            # leave out IRISPH5 which is for nodal experiments and might match
+            # insanely large datasets, depending on restrictions
+            if "IRISPH5" in providers:
+                del providers["IRISPH5"]
 
             _p = sorted(providers)
             if has_orfeus:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by krischer — CHANGES_REQUESTED** [review]
This looks fine to me and I think this works as expected.

One thing I think should be changed is to not have this data center as one of the default data centers in the mass downloader. It is fine if it can be optionally selected but this should be a conscious user choice as it could potentially download a truly massive amount of largely large-N or active source data which is not what most users of ObsPy would expect.

**Comment by megies** [comment]
> It is fine if it can be optionally selected but this should be a conscious user choice as it could potentially download a truly massive amount of largely large-N or active source data which is not what most users of ObsPy would expect.

Yes, I agree. @jrekoske-usgs, you did explicitly mention using it for `MassDownloader`, but I think it should already help a lot having the short URL to go with in the mass downloader? Having it in there as a default could really lead to awkward scenarios for users.. we can mention it prominently in the docs for `MassDownloader`, I guess.

> This looks fine to me and I think this works as expected.

Yeah, I guess the code is already kinda convoluted anyway, so just the minor bit more complexity doesn't hurt too much

File: obspy/clients/fdsn/mass_downloader/mass_downloader.py [hunk 1]
```diff
@@ -55,11 +55,13 @@ class MassDownloader(object):
...
-        is available at more then one provider it will always be downloaded
-        from the provider that comes first in the list. To include RASPISHAKE,
-        you must set this parameter to
+        all FDSN implementations known to ObsPy except RASPISHAKE (generally
+        worse quality data) and IRISPH5 (active source / nodal experiments that
+        might match a very large amount of data occasionally) if set to None.
+        The order in the list also determines their priority, if data is
+        available at more then one provider it will always be downloaded from
+        the provider that comes first in the list. To include RASPISHAKE and
+        IRISPH5, you must set this parameter to
```
krischer, line 64: I forgot that RASPISHAKE is also already excluded - but definitely a sane default to exclude the raspian and irisph5 one!

krischer (approved):

jrekoske-usgs:
@megies This looks great, I agree that it makes sense to leave IRISPH5 out of the MassDownloader by default. Thanks for getting to this so quickly!

PR #2743, opened 2020-11-03
pathlib.Path support

d-chambers:
### What does this PR do?

This PR adds support for `pathlib.Path` for `read_events`, `read_inventory`, and `read` functions. 

It is implemented by simply converting `pathlib.Path` objects to str instances in `obspy.core.util.base._generic_reader`. 

It is onerous to have to convert `Path` objects to strings before reading inputs. This PR will just make it a bit more ergonomic. 

Support for `pathlib.Path` objects is discussed in #1794. However, that issue also talks about support for additional input types such as `ParseResult`, which I have not addressed here.

### PR Checklist
- [x] All tests still pass.


ThomasLecocq:
🎉 excellent! let me know if you need help !

d-chambers:
@ThomasLecocq,

Sure, feel free to review. It was surprisingly simple so maybe I am missing something but it seems to work fine.

ThomasLecocq:
> Sure, feel free to review. It was surprisingly simple so maybe I am missing something but it seems to work fine.

Well it depends if you want to use pathlib everywhere a path is used in the code :) there are quite a few, certainly in all the tests etc...

d-chambers:
> Well it depends if you want to use pathlib everywhere a path is used in the code :) there are quite a few, certainly in all the tests etc...

Although this would clean up the code quite a lot, its probably better left to be included in a larger refactoring effort. My main goal is to just ensure ObsPy's public functions which accept strings representing paths can also accept `pathlib.Path` objects. I only covered/tested `read`, `read_events`, and `read_inventory` so far but I probably need to track down others.

obspy/core/tests/test_stream.py
```diff
@@ -1948,6 +1948,16 @@ def test_read_url_via_network(self):
...
 
+    def test_read_path(self):
+        """
+        Test for reading a pathlib object.
+        """
+        base_path = Path(__file__).parent / 'data'
+        data_path = base_path / 'IU_ULN_00_LH1_2015-07-18T02.mseed'
+        assert data_path.exists()
+        st = read(data_path)
+        assert isinstance(st, Stream)
```
megies, line 1959: historically we used to go with unittest's `self.assertTrue()` but personally I don't mind the `assert`s, they're shorter and less convoluted and I don't think they should have any significant downsides as opposed to the unittest style

obspy/core/util/base.py
```diff
@@ -646,6 +647,9 @@ def download_to_file(url, filename_or_buffer, chunk_size=1024):
 
 def _generic_reader(pathname_or_url=None, callback_func=None,
                     **kwargs):
+    # convert pathlib.Path objects to str for compatibility.
+    if isinstance(pathname_or_url, Path):
+        pathname_or_url = str(pathname_or_url)
```
megies, line 652: Do people actually use `PurePath`? I don't really know why they have that parent class to `Path` but since you are converting to string and then file opening is done outside of pathlib, we might just make this check go for `PurePath`, I guess?

megies (approved):
One comment on the option to maybe use `PurePath`, otherwise looks good to go.

I think it would be a good idea to spell out the three user-faced functions that are getting pathlib-enabled by this.

Also, I think we want to mention this in the docs for these three functions, e.g. https://docs.obspy.org/master/packages/autogen/obspy.core.stream.read.html?highlight=pathname_or_url

megies:
such a small change only needed for this thanks to previous work by @ThomasLecocq in #2326 :clap:

obspy/core/tests/test_stream.py
d-chambers: @megies, your right. I have just become so accustomed to pytest I forgot to do that here, I will fix it.

obspy/core/util/base.py
d-chambers: You are right, `PurePath` should be used here. If someone wanted to simulate path handling not native to the machine they are working on they have to reach for something besides `Path` (eg `WindowsPath` can be used on linux). However, in order for that to work we would need to wrap any instance of `PurePath` in a normal `Path` to make sure the string conversion will be in the right form.

d-chambers:
Alright @megies, I addressed your comments and performed a few small cleanups which were preventing the `pre-commit run -all` command from passing. Should be good to merge depending on CI.

obspy/core/util/base.py
megies: My understanding from reading the docs is, `PurePath` is parent to `Path` implementing everything besides actual file opening. It's parent to all other Path flavors, so I would expect this is exactly what it's there for.

obspy/core/util/base.py
```diff
@@ -646,6 +647,9 @@ def download_to_file(url, filename_or_buffer, chunk_size=1024):
 
 def _generic_reader(pathname_or_url=None, callback_func=None,
                     **kwargs):
+    # convert pathlib.Path objects to str for compatibility.
+    if isinstance(pathname_or_url, PurePath):
+        pathname_or_url = str(Path(pathname_or_url))
```
megies, line 652: No need to cast to `Path`, `__str__` is implemented by `PurePath` as parent anyway

d-chambers: The reason I included this is a bit esoteric, but here is the weird edge case I had in mind:
```python
import pathlib
# assume we are using linux
windows_path = pathlib.PureWindowsPath('base') / 'filename.mseed'
print(str(windows_path))  # prints 'path\\filename.mseed' which is not a valid path for linux, hence we need the extra Path call
```
I don't think this will really happen much, if at all, but the code as it is written will handle it.

d-chambers:
rebased and force pushed

obspy/core/util/base.py
megies: I still don't get it.. for me using..

```
print(str(pathlib.Path(windows_path)))
```

..in there doesn't change anything on the output

d-chambers: Ok this is probably too far of an edge case anyway (why would someone use a PureWindowsPath on linux with ObSpy to read a file?) so I am going to just remove the extra `Path` constructor.

megies:
Thanks @d-chambers :rocket:

megies:
I just saw this @d-chambers: https://docs.python.org/3/library/os.html#os.PathLike
Not sure if that's probably even better, but I guess it might not make too much of a difference.

d-chambers:
Wow, I didn't know `PathLike` existed, I guess it pays to read the docs :stuck_out_tongue_winking_eye:. I guess `PathLike` would be better than `PurePath` but I think either one should work fine for how we are using them in this PR.

PR #2752, opened 2020-11-19
cleanup: Some updates after dependency minimum versions were bumped

megies:
### What does this PR do?

Remove some workarounds for legacy dependency versions, which should be obsolete on master now.

Cleanup..

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/utcdatetime.py
+++ b/obspy/core/utcdatetime.py
@@ -13,6 +13,7 @@ import calendar
 import math
 import operator
 import re
+import sys
 import time
 import warnings
 
@@ -38,6 +39,9 @@ _ISO8601_REGEX = re.compile(r"""
     )?
     $
     """, re.VERBOSE)
+# Regular expression used in the init function of the UTCDateTime objects which
+# is called a lot. Thus pre-compile it.
+_YEAR0REGEX = re.compile(r"^(\d{1,3}[-/,])(.*)$")
 
 TIMESTAMP0 = datetime.datetime(1970, 1, 1, 0, 0)
 
@@ -337,6 +341,12 @@ class UTCDateTime(object):
                     value = value.decode()
                 # got a string instance
                 value = value.strip()
+
+                # Raising in the case where the leading string is less than 4
+                # chars; linked to #2167
+                if re.match(_YEAR0REGEX, value):
+                    raise ValueError(
+                        "'%s' does not start with a 4 digit year" % value)
 
                 # check for ISO8601 date string
                 if iso8601 is True or (iso8601 is None and
@@ -1261,6 +1271,8 @@ class UTCDateTime(object):
         See methods :meth:`~datetime.datetime.strftime()` and
         :meth:`~datetime.datetime.strptime()` for more information.
         """
+        if sys.platform.startswith("linux"):
+            format = format.replace("%Y", "%04Y")
         return self.datetime.strftime(format)
 
     @staticmethod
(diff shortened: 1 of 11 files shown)


heavelock (approved):
I think it's good to go if the tests are passing

heavelock:
So, I brought back the catching of exception when trying to load an old ppsd file without `allow_pickle=True` because test `obspy.signal.tests.test_spectral_estimation.PsdTestCase.test_can_add_npz_without_pickle` was checking for a nice error message. This brought me to a similar situation in `obspy.signal.spectral_estimation.PPSD.load_npz` which also is affected by not passing the `allow_pickle=True`. There is also a test checking for that `obspy.signal.tests.test_spectral_estimation.PsdTestCase.test_can_read_npz_without_pickle` but it seems like it does not work as intended. It seems that the `np.load` does not raise an exception when loading the test file that was passed to it. So it means that the file itself does not require it. 

In order to fix that problem, I extracted a method to create a file that requires pickle from other test method and used it here with almost the same testing procedure.

obspy/signal/tests/test_spectral_estimation.py
```diff
@@ -932,32 +931,32 @@ def test_can_read_npz_without_pickle(self):
             PPSD.load_npz(self.example_ppsd_npz)
```
heavelock, line 931: @d-chambers Could you take a look on that code here? I have some doubts that it really does what it says.

heavelock (changes requested):

obspy/signal/tests/test_spectral_estimation.py
heavelock: I think this part of the test is checking if this error is raised:
obraz

(sorry for screenshot, gh does not allow for commenting code that is out of the scope of a diff...)

I am not sure that this message says the truth. At that point, the data is already loaded so passing a `allow_pickle=True` does not change anything.

ThomasLecocq: so you'd want the msg to say "cannot convert back to list"? (or it could actually silently do it ?)

ThomasLecocq:
@QuLogic could you take a look at this one ? It's somehow related to your setup/requirement improvements

QuLogic:
Some is, some isn't, but it should be good now.

megies:
rebased and force pushed

megies:
Went through all changes and it looks good to me

obspy/core/utcdatetime.py
```diff
@@ -1282,52 +1271,10 @@ def strftime(self, format):
...
-                    self.__class__._strftime_four_digit_year_works = (
-                        len(check) == 4)
-                    # now delegate again
-                    return self.strftime(format)
-            # just a safety net to not break UTCDateTime.strftime if some
-            # unforeseen problems with the above approach appear
-            except Exception:
-                pass
-
+        if sys.platform.startswith("linux"):
```
trichter, line 1274: Only needed for linux?

trichter (approved):
Looks also good to me. Please merge when its finished.

obspy/core/utcdatetime.py
megies: To my understanding, yes. This is a glibc implementation thing, `"%0xY"` is only implemented on Linux glibc it seems and Linux might otherwise not put leading zeros, while Mac/Windows seem to put leading zeros nevertheless and rather error out on using `"%0xY"`

megies: I'll add those comments pointing to the Python bug reports back in

PR #2758, opened 2020-11-27
Faster loadtime

### LINKED ISSUE #2748: Making ObsPy load faster
This has been already discussed in the past, but for another project, I was checking solutions to check import-times.
From [this ](https://stackoverflow.com/questions/16373510/improving-speed-of-python-module-import) I found out about [tuna](https://github.com/nschloe/tuna) which allows to graphically see the import stacks/dependencies.

Running it on obspy : ``python -X importtime -c "import obspy" 2> obspy.log`` then, ``tuna obspy.log`` :

image
shows we have two major culpirts: ``UTCDateTime``, and more surpisingly, ``obspy.geodetics`` which seems to have its load time strongly affected by loading a whole ``scipy.stats`` stack.

This comes from a single method required by ``obspy.geodetics.base.mean_longitude`` which calls ``scipy.stats.circmean``. Moving the import inside the function, and re-running ``tuna``:
image
We win more than 1 second in load time !

Continuing with ``UTCDateTime``, looks like ``obspy.core.util.deprecation_helpers`` is responsible alone for 1.1 second (>33%) of the remaining load time:
image

I don't really understand the reason of that big stack, looks even a bit "circular" at some point... Anyway, removing the import of the ``from obspy.core.util.deprecation_helpers import ObsPyDeprecationWarning`` from  ``obspy.core.utcdatetime``  (I know, that would break the code if the Deprecation needs to be raised, but wanted to test):
image

we win > 0.6s or so... 

Remaining "heavy" (long) imports are ``Attribdict``, then I guess some of the ``pkg_resources`` scanning the entry points etc...

What do you think? The first change (importing scipy.stats only if needed) shouldn't harm. 

ps: running the importtime on "from obspy import read" gives +- the same timings.

megies:
### What does this PR do?

Speed up import of obspy and most used routines (read, etc.)

Fixes #2748 

### PR Checklist
- [ ] All tests still pass.


megies:
Maybe you could have a look @ThomasLecocq. The only thing you mentioned that i left untouched for now is the import of `runtests` in `core`, but I moved the requests import in there into a function. Maybe thats the biggest import in there anyway?

ThomasLecocq:
will test & check this weekend! Thanks!

ThomasLecocq:
looks good, 2.73s instead of more than 4s originally ! 
image

maybe remaining:

## pyproj (in obspy.core.util.base) 

actually it seems "import pyproj" is only present in:
+ io.nlloc.test_util.py 
and "from pyproj ..." in:
+ util.base (L371 & L382) --> only because we want to have "PROJ4_VERSION" available


## doctest (in obspy.core.util.base) 

only places where it is imported outside ``if __name__ == "__main__"``
+ io.gse2.libgse2.py 
+ io.gse2.paz.pi 
+ runtests.py

## matplotlib (in obspy.imaging.cm) 

but in mannnny places too, so I guess this shouldn't be removed


tests are OK on my machine:  https://tests.obspy.org/112612/

ThomasLecocq (changes requested):
could just try to remove the last doctest (and discuss pyproj?)

megies:
- we could hide PROJ check inside functions, I guess and remove it from core/util.py if you think it's worth the 0.2s speedup (the occurence in the test case file shouldn't matter anyway since that should not be loaded on "import obspy" or similar?!)
 - i'm hiding doctest inside functions, no reason not to as a builtin it's always there anyway
 - i also hid that `imaging.cm` import, I believe it should disappear from imports now
 
 @ThomasLecocq

ThomasLecocq:
there are still matplotlib.cm at different locations ... :'( this is annoying....

the import time is OK (around 2 seconds) , and without a major refactor/usage of `reentry` or else to minimise `pkg_resources` heavy scanning time lost, it's difficult to go much further down.

About `pyproj`, if not annoying, I'd hide it too, it's not (almost never) necessary for running stuff in obspy, so really not needed.

image

megies:
@ThomasLecocq tried to get rid of `matplotlib.cm` again

ThomasLecocq:
ok, new pass on this:
`` python -X importtime -c "import obspy" 2> obspy.log `` 

master: 5.175 s
this branch: 1.512 s

the remaining elements taking time are:
- pkg_resources
- numpy

it's quite difficult to shrink those more.

Question for @obspy/maintainers do we still need any of the proj4 version checks etc?  Weren't those only meant to be friendly to basemap, which is no longer there anyway (at least for the inventory/event plot, I completely removed the support for it); Here I changed the way to get the proj version (is bad, I know, but since pyproj uses PROJ v8 now, not sure it's really needed to check for 5.2.0 ?

before:
image

after:
image

ThomasLecocq:
I have NO idea why now some trig test fail for one platform only :'( https://tests.obspy.org/117456/

ThomasLecocq:
@QuLogic I could use your wizard skills here :-)

obspy/core/util/base.py
```diff
@@ -406,7 +404,7 @@ def get_proj_version(raw_string=False):
 SCIPY_VERSION = get_dependency_version('scipy')
 MATPLOTLIB_VERSION = get_dependency_version('matplotlib')
 BASEMAP_VERSION = get_dependency_version('basemap')
-PROJ4_VERSION = get_proj_version()
+PROJ4_VERSION = get_dependency_version('pyproj')
```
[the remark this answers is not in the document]
ThomasLecocq: actually removed it in the #2961 branch

obspy/core/util/base.py
```diff
@@ -159,6 +156,7 @@ def create_empty_data_chunk(delta, dtype, fill_value=None):
                  mask = ...,
                  ...)
     """
+    import numpy as np
```
trichter, line 159: I don't have a clue about the new failing test. Could you move this numpy import again to the global scope and see if it helps? Numpy is imported at other places in the global scope, so it should not degrade the load time.

ThomasLecocq:
@trichter sorry am full full this end of the week, if you have time, feel free to push to this branch, no worries.

obspy/core/util/base.py
ThomasLecocq: good idea, indeed, numpy is imported elsewhere anyway

ThomasLecocq:
this is wild, the numpy error is probably linked to a loading order of scipy vs numpy, but ... wow...

ThomasLecocq:
current timing: ~1.5s for "from obspy import read"
image

ThomasLecocq:
I really don't get it... reverting the 2 scipy changes doesn't solve the issue, placing pkg_resources at the top of runtests either...

ThomasLecocq:
so, adding "import requests" in `obspy/__init__.py` solves the numpy issue... WTF....

trichter:
> so, adding "import requests" in obspy/__init__.py solves the numpy issue... WTF....

That's crazy... I have no clue how this could happen.

d-chambers:
I was playing with this a bit on my private branch and I found if I re-installed libffi the tests would pass, but just putting libffi in the env file didn't work. It may have to be after ObsPy is installed, but I am still trying to figure it out.   

```bash
install -c conda-forge libffi=3.3.*
```

d-chambers:
Adding this step to the github default_test job, after the obspy install step, fixes the issue:

```bash
        # reinstall libffi to avoid python 3,8 failures in pk_baer, see (#2758)
      - name: fix libffi
        shell: bash -l {0}
        if: runner.os == 'Linux' && contains(matrix.python-version, '3.8')
        run: conda install libffi=3.3.*
```

Its probably better to just add the step to the workflow than to keep the request import in `obspy.__init__` since the issue affects a very limited subset of the supported OS/python version and the fix is simple if any actual users run into it. Its up to you though @ThomasLecocq.

Good to merge after that?

ThomasLecocq (commented):
is it also needed on the master tests?

d-chambers:
> is it also needed on the master tests?

Only if we want them to pass 😉

d-chambers:
Ready to merge or is there something left you want to do @ThomasLecocq ?

ThomasLecocq:
good to go for me ! thanks a lot for the help/debug!

PR #2759, opened 2020-11-27
Add "level" option to read_inventory / _read_stationxml

megies:
### What does this PR do?

Make it possible to read reduced level of detail from StationXML files.

Reading response is by far the most time (and potentially memory) consuming part and in many cases it is not needed, e.g. when only station coordinates or available channels are of interest.

Example speedups with a 20 MB StationXML file with 1000 channels with responses: not parsing responses reduces read time by 80%. Only reading station level cuts time down by 90%.
```
$ python -m timeit -n 3 -r 5 -s 'from obspy import read_inventory' 'read_inventory("inventory_jane.xml", "STATIONXML", level="response")'
3 loops, best of 5: 3.71 sec per loop

$ python -m timeit -n 3 -r 5 -s 'from obspy import read_inventory' 'read_inventory("inventory_jane.xml", "STATIONXML", level="channel")'
3 loops, best of 5: 745 msec per loop

$ python -m timeit -n 3 -r 5 -s 'from obspy import read_inventory' 'read_inventory("inventory_jane.xml", "STATIONXML", level="station")'
3 loops, best of 5: 335 msec per loop

$ python -m timeit -n 3 -r 5 -s 'from obspy import read_inventory' 'read_inventory("inventory_jane.xml", "STATIONXML", level="network")'
3 loops, best of 5: 244 msec per loop
```

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -38,7 +38,8 @@ def _create_example_inventory():
 
 
 @map_example_filename("path_or_file_object")
-def read_inventory(path_or_file_object=None, format=None, *args, **kwargs):
+def read_inventory(path_or_file_object=None, format=None, level='response',
+                   *args, **kwargs):
     """
     Function to read inventory files.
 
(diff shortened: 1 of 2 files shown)


ThomasLecocq:
Does this apply to any other inventory formats ?

megies:
@ThomasLecocq it's similar to `headonly` for waveforms. This PR will add it as a user-facing high level option on `read_inventory`, but it's up to each individual plugin whether to implement it or not. It's most relevant for StationXML right now, as it's a huge difference there, and other plugins might not even support details down to response level, so it's not really needed everywhere. Of course it might be interesting for Dataless/FullSEED too, but I think most people should be on StationXML by now for most workflows.

ThomasLecocq (approved):

PR #2766, opened 2020-12-09
initial commit on waveform timemarks featuring

mbagagli:
### What does this PR do?

Yet another feature on `Stream.plot()` method. This PR introduces a new boolean parameter called `plot_time_marks`. This parameter  will allow the user to plot `timemarks` on plto types `normal`, `reference` and `section`. This PR introduce a new parameter for StThese time marks are customizable like any other matplotlib line object. 

These timemarks are defined in an additional attribute in `Trace.stats` object. The attribute is a container of tuples/lists of 2 elements each: a time, and a matplotlib dictionary plot.

```
tr.stats.timemarks=[
            (UTCDateTime("2011-11-24T21:11:07.0"),
            {'color': 'b', 'markeredgewidth': 3, 'markersize': 10 ...}),
            ...
]
```

It started due to a personal need to fast-plot custom TimeMarks over Stream section's plot. 
Then I thought the community could also benefit from it, and I've implemented also on `normal` and `relative` type-plot as well. 
We all know that users must create their own custom scripts for ad-hoc figures, but this additional parameter could give a quick vision of different time-features (ie. picks, swarns etc..). 
No relevant issues. At the moment it is not implemented in the `dayplot` type but I could of course add it if we prefer so.

### PR Checklist
- [ x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ x] This PR is not directly related to an existing issue (which has no PR yet).
- [ x] If the PR is making changes to documentation, docs pages can be built automatically.
- [ ] All tests still pass.
- [x ] Any new or changed features have are fully documented.

### DIFF
--- a/misc/docs/source/tutorial/code_snippets/waveform_plotting_tutorial.rst
+++ b/misc/docs/source/tutorial/code_snippets/waveform_plotting_tutorial.rst
@@ -114,13 +114,25 @@ by setting parameter ``type`` to ``'section'``:
 
     >>> stream.plot(type='section')
 
-To plot a record section the ObsPy header ``trace.stats.distance`` (Offset) must be 
-defined in meters. Or a geographical location ``trace.stats.coordinates.latitude`` & 
-``trace.stats.coordinates.longitude`` must be defined if the section is plotted in 
-great circle distances (``dist_degree=True``) along with parameter ``ev_coord``. 
+To plot a record section the ObsPy header ``trace.stats.distance`` (Offset) must be
+defined in meters. Or a geographical location ``trace.stats.coordinates.latitude`` &
+``trace.stats.coordinates.longitude`` must be defined if the section is plotted in
+great circle distances (``dist_degree=True``) along with parameter ``ev_coord``.
 For further information please see :meth:`~obspy.core.stream.Stream.plot`
 
 .. plot:: tutorial/code_snippets/waveform_plotting_tutorial_6.py
+
+----------------------------------------------------
+Plotting a Stream or a Record Section with TimeMarks
+----------------------------------------------------
+
+A :class:`~obspy.core.stream.Stream` object plot with type equal to section,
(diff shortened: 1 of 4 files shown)


heavelock:
I have mixed feelings about using as input a field in `tr.stats`. It's not vere explicit and quite much separates data from plotting. 
I was recently implementing some per trace processing and I went in direction of using as argument a dict with seedid as key which I think is a bit more clear for the user. From the perspective of plotting routines it's probably a bit more complicated though.

What are your thoughts?

mbagagli:
@heavenlock your points are right but I think it could be the classic 2-faces of the same medal. 
One could also think to pass a list of time-objects directly at invocation. But while being easy to do for a single `Trace` object, it would became cumbersome in case of big streams (as you said we would need a seedid linked to each single trace to keep timemarks separated).

In my opinion it would be easier (with all the updated documentation) to associate `timemarks` directly on the trace of interests rather than using a separate dict with seedid.
I think that originally the idea of such a `plot` method was to quickly document the _state_ of a `Stream` directly with different methods.

I've also came up with the idea of using a list of modified `origin.Pick` class instances, but rather than adding attributes to that, I preferred to use standard dictionaries. This choice came also from the fact that `trace.stats` is holding metadata, and some official tweaks has been done before (i.e. `trace.attach_response()` that creates the `trace.stats.response` attribute.

Maybe we could think of a better structured dictionary for this field? Or maybe creating a really small class `Timemark`? Timemarks may indeed refer to totally un-correlated information respect to the waveforms and should not be treated strictly as phase-picks. Eventually, this attribute could be used in at some stage in other class-methods as well?

heavelock:
I see your point here. I think the best solution for this problem would be really going for either little helper class as you are saying or some other defined data structure. This would help with validation of the inputs from users and probably helped a lot with reusability of that structure. 

However, this I think need a broader discussion.

PR #2775, opened 2021-01-06
fix a small bug of writing CMTSOLUTION file

lynlyf:
### What does this PR do?

*Please fill in*

*Please fill in*

### PR Checklist
- [ ] All tests still pass.
- [*] First time contributors have added your name to `CONTRIBUTORS.txt` .

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,1 +1,1 @@
-obspy/CONTRIBUTORS.txt
+Lei LI
--- a/obspy/io/cmtsolution/core.py
+++ b/obspy/io/cmtsolution/core.py
@@ -419,12 +419,12 @@ def _internal_write_single_cmtsolution(buf, event, **kwargs):
             half_duration = mt.source_time_function.duration / 2.0
         else:
             warnings.warn("Source time function has no duration. The half "
-                          "duration will be set to 1.0.")
-            half_duration = 1.0
+                          "duration will be set to 0.")
+            half_duration = 0.0
     else:
         warnings.warn("Moment tensor has no source time function. Half "
-                      "duration will be set to 1.0.")
-        half_duration = 1.0
+                      "duration will be set to 0.")
+        half_duration = 0.0
 
     # Now attempt to retrieve the event name. Otherwise just get a random one.
     event_name = None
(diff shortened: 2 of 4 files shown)


ThomasLecocq:
hi, thanks for the PR. Please:

- provide context of why it is needed
- provide an explanation about the choices of setting 0 instead of 1 for the STF
- correct the contributors file, you didn't put your name at the right location (the one you modified links to the right file)

lynlyf:
Hi Thomas,

I'm a bit confused. I apologize in advance if it's due to my mis-operations to pull all the changes (I always work locally with github and am not so familiar with the inter-repositories PRs). 

I only requested a PR long time ago to fix the bug of writing the magnitudes in the CMTSOLUTION file. The other changes were only made in my own repository. I changed the missing half duration values from 1.0 to 0.0 to distinguish missing values from existing values (over 100 events in the CMT catalog already have their half duration equal to 1.0). I'm not sure if other obspy users need such a utility.

ThomasLecocq:
Hi @lynlyf , this explains it indeed, once open, a PR tracks the original branch and auto-updates! 

Your original fixes seem needed:
- the lowercase mb check
- the half time 1.0 or 0.0

The former makes no difference for the users, but the latter could be changing default behaviour. As such, I would say the choice of 1.0 was wrong from the start, except if it's documented somewhere in the CMT format that it has to be so? I believe it'd be better to change it, even to 999.999 or else, especially if many CMT events have a real true value == 1.

lynlyf:
> Hi @lynlyf , this explains it indeed, once open, a PR tracks the original branch and auto-updates!
> 
> Your original fixes seem needed:
> 
> * the lowercase mb check
> * the half time 1.0 or 0.0
> 
> The former makes no difference for the users, but the latter could be changing default behaviour. As such, I would say the choice of 1.0 was wrong from the start, except if it's documented somewhere in the CMT format that it has to be so? I believe it'd be better to change it, even to 999.999 or else, especially if many CMT events have a real true value == 1.

Hi Thomas,

CMTSOLUTION file can be used as source param file in some simulation codes (like SPECFEM_GLOBE). It's a common practice to set the half duration to 0 and one can convolve a source wavelet in the postprocessing.

That's why I favored to fill the missing values with 0 instead of a flag like 999.999 (SPECFEM will convolve the wavefield with a very long period STF during the simulation.)

lynlyf:
Regarding the letter cases, `_i.magnitude_type.lower() == "mb"` is more robust. I think I made the change because I met exceptions with the original codes (sometimes `.magnitude_type` being `MB` or `mb`)

PR #2782, opened 2021-02-08
RG16 parser to account for bad alias frequency dtype

d-chambers:
### What does this PR do?

This PR allows the rg16 parser to account for a small bug which can occur in the file format where the anti alias filter frequency (in the channel description block) is written as an int32 rather than BSD.

I had a colleague send me a large rg16 file which ObsPy was not able to read due to the issue described above.

### PR Checklist
- [x] All tests still pass.


d-chambers:
@thefroid, do you mind taking a quick look at this? It should be very straight-forward to review.

d-chambers:
Another issue that came up with this strange rg16 file is that the channel set descriptors contain identical `RU_channel_number` values, despite the channels describing different components. This caused the number of records to be underestimated by 3X. I have modified `_cmp_nbr_records` to first look for the `RU_channel_number` value, if a 0 is found then the `channel_set_number` will be used to determine how many records exist.

d-chambers:
I just rebased this branch.

Does anyone have a few minutes to look at it? The changes are very minor and well tested. If not, I will go ahead and merge in a few days so this small but important fix doesn't get forgotten.

obspy/io/rg16/util.py
```diff
@@ -50,9 +50,11 @@ def _read(fi, position, length, dtype, left_part=True):
         return _read_bcd(fi, length, left_part)
     elif dtype == 'binary':
         return _read_binary(fi, length, left_part)
-    elif dtype == 'IEEE':
-        data = np.frombuffer(fi.read(int(length)), '>f4')
-        return data[0] if len(data) == 1 else data
+    if dtype == 'IEEE':
+        dtype = '>f4'
+    # dtype should be understood by numpy, dispatch to frombuffer
+    data = np.frombuffer(fi.read(int(length)), dtype)
```
ThomasLecocq, line 56: shouldn't this return be in the if previously, and raise an exception if dtype is not known?

ThomasLecocq:
seem OK to me, just check my comment about dtype ...

obspy/io/rg16/util.py
d-chambers: Thanks for pointing this out, the docs for this function are a little out of date. The function is actually correct because it allows ints and other numpy dtypes to also be used. I updated the docs string to be more clear.

ThomasLecocq:
ok for me if all test pass

d-chambers:
All tests green and with @ThomasLecocq review I will go ahead and merge.

PR #2783, opened 2021-02-12
Action building wheels 

heavelock:
Hi,
I thought about adding one more workflow for GH Actions. This one is for building wheels for different platforms and python versions. In principal it would help with doing releases and take some release burden of @megies back.
Currently, it is building wheels for 3 major platforms for py 37 and 38 on default architectures.

I did not add yet the option to upload to Pypi because I think it requires a bit of discussion.

Looking at official examples of cibuildwheel https://github.com/joerick/cibuildwheel/blob/master/examples/github-deploy.yml
it is quite easy to make the upload work only for tags or gh releases. The question is, which one should I set up.

The second question I have is if this action should be building wheels for PRs, just for master, or only for releases. I would say that master only plus upload to pypi on release/tag. Currently it is running on all PRs so we can see the results. It also uploads artifacts so they are available for some time to download (https://github.com/actions/upload-artifact#where-does-the-upload-go)

What do you think?


heavelock:
In fact, I just noticed that we have already a workflow `release_published.yml`. However, my version is also building binary wheels for matrix of systems.

So my PR should be rather targeting modification of `release_published.yml`. For now I will leave it without change until review - currently this workflow is building on each push and the target worfklow only when there is a GH release done. 

@d-chambers Would you mind taking a look?

d-chambers:
I just glanced at it. it looks good to me. I would suggest just running this on release tags and including the PyPi upload. Is there is a reason to run the wheel building on each push/PR? You should be able to merge the two actions in question together quite easily. We would need to get @megies or someone else with the pypi credentials to setup a token and upload it to ObsPy's github secrets it if hasn't already been done.

I have not spent the time to understand wheels very well but it looks like you already have the platform specific parts working. I guess you would just need to add all the wheels to the source directory before uploading to PyPI?

heavelock:
No, definitely there is no reason for this workflow to be ran every push, it should be only on tags/releases. The only reason that is like this right now is just a verification that it works as expected.

And yes, that was the end goal to supplement your workflow with wheel building. 

I don't think there is a need even to add wheels to source - it''s just an artifact so it will be sufficient to upload upload them locally and upload them to PyPi directly.

heavelock:
Summary of updates:
- This workflow will be triggered only by pushes to master as well as release done on Github.
- Wheels and sdist will be built for both triggers and exposed as action artifact. Artifacts are available for 90 days, then they are removed automatically.
- Wheels are built for matrix of systems and python versions. There could be also added some non-standard architectures but I just kept the defaults.
- Wheels are tested every time they are built. Test results are uploaded to our test report system.
- Wheels and sdist will be uploaded to pypi only on release. 
- Currently the workflow is failing due to the failing tests on wheels (image comparison tests). 
- Since the wheel build is not triggered on PR but is on master, it will fail on master.

Some CI runs from my fork:
1. Standard run to test if wheels are built properly, without running tests so the run succeeds - https://github.com/heavelock/obspy/actions/runs/586980253
2. Run without tests but with upload to the test pypi. The upload itself failed because the version in wheel in metadata was not okay. https://github.com/heavelock/obspy/actions/runs/586980924
3. Run with testing of all the wheels and upload to PyPi only in case of `release` event: https://github.com/heavelock/obspy/actions/runs/588574125

barsch:
@heavelock @d-chambers good to merge?

d-chambers:
The PR itself looks good to me. I trust @heavelock has tested it on his own branch. The only thing I can see we still need to do is generate a pypi token and add it to our github secrets named "PYPI_TOKEN".

barsch:
@heavelock wrote me on gitter:

_"... I would prefer not to have it merged yet. I started backporting it to the maintenance branch to check if it can be used out there too ... I will not have time in April to do anything with that ... In May though, I will come back to that ..."_

so guess it will take a bit more time - but thanks for your input @d-chambers

trichter:
Superseded by referenced PR.

heavelock:
Thanks for finishing that up @trichter !

PR #2806, opened 2021-03-09
Default units "DEF" option for sensor response

claudiodsf:
### What does this PR do?

Adds a `"DEF"` option (default units) to `Inventory.plot_response()` and `Trace.remove_response()` (plus other helper functions).

Useful if the units for a particular type of sensor (e.g., a pressure sensor) cannot be converted to displacement, velocity
or acceleration.

This option is available in the underlying library `evalresp` (see `-u def` in evalresp manpage). This PR simply exposes it via
the ObsPy API.

The current options for `output` in `Trace.remove_response()` (`"DISP"`, `"VEL"` or `"ACC"`) are confusing for people working on non-ground-motion sensors (e.g., a pressure sensors). Correctly removing the response for these kind of sensors requires using the `"VEL"` option, which is misleading.

An example of user confused by the option is [here](https://gitter.im/obspy/obspy?at=60461bc444f5a454a4613a01).

In this PR, I expose the `"DEF"` option which is already available in `evalresp` (see `-u def` in [manual](https://ds.iris.edu/ds/nodes/dmc/software/downloads/evalresp/5-0-0/manual/))

### PR Checklist
- [ ] All tests still pass.


barsch:
> 
> 
> Not sure if `1T_MONN_00_EDH.xml` is in the right path (currently in `obspy/core/tests/data/`).

shouldn't that test be part of ```io.stationxml``` ?

ThomasLecocq (changes requested):
agree with @barsch the XML could go to the io/stationxml/data location

ThomasLecocq:
@claudiodsf could you move the XML file & rebase, then it's good to go

ThomasLecocq:
the macOS test will fail, but I'm on it, so it'll require one last rebase if I succeed :-)

ThomasLecocq:
@claudiodsf looks like the test will pass smoothly... Just re-reading it all now, and wonder if it wouldn't be better to be more explicit for the ``output`` and call it ``DEFAULT`` instead of ``DEF`` ?

claudiodsf:
> @claudiodsf looks like the test will pass smoothly... Just re-reading it all now, and wonder if it wouldn't be better to be more explicit for the `output` and call it `DEFAULT` instead of `DEF` ?

I don't know: other options are abbreviated (`"DISP"`, `"VEL"`, `"ACC"`), mimicking --I guess-- `evalresp` options (`dis`, `vel`, `acc`, `def`).

ThomasLecocq:
> > @claudiodsf looks like the test will pass smoothly... Just re-reading it all now, and wonder if it wouldn't be better to be more explicit for the `output` and call it `DEFAULT` instead of `DEF` ?
> 
> I don't know: other options are abbreviated (`"DISP"`, `"VEL"`, `"ACC"`), mimicking --I guess-- `evalresp` options (`dis`, `vel`, `acc`, `def`).

ok for me :-)

ThomasLecocq:
@claudiodsf sorry, again: please add a line in the changelog about this!

ThomasLecocq (approved):

ThomasLecocq:
thanks! merged!

PR #2808, opened 2021-03-10
Fix Seedlink basic_client get_info failing when location and channel not in Seedlink reply

brenmous:
### What does this PR do?

In `seedlink.basic_client.get_info`, sets `loc` and `cha` as empty strings if they are not included in the Seedlink reply.

Current behaviour is to set `loc` and `cha` in `station_cache` as `None` if they aren't in the Seedlink reply. This causes the function to fail, as `station_cache` is then searched using `fname`, with `loc` and `cha` provided as `name` argument. A value of `None` for `name` will cause `fname` to raise, as it calls `os.path.normpath(name)`, which expects a string. Empty strings get around this issue. This does mean missing codes will be returned as '' and it's up to the user to understand this means Seedlink didn't reply with the code, and that the stream may not be available.

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.seedlink"
- [ ] All tests still pass.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -68,6 +68,7 @@ Meschede, Matthias
 Michelini, Alberto
 Miller, Nathaniel C.
 Morgenstern, Bernhard
+Moushall, Brenainn
 Murray-Bergquist, Louisa
 Nesbitt, Ian M.
 Nof, Ran Novitsky
--- a/obspy/clients/seedlink/basic_client.py
+++ b/obspy/clients/seedlink/basic_client.py
@@ -308,7 +308,7 @@ class Client(object):
                 # seems the seedlink server replies with no subtags for the
                 # channels
                 if not subtags:
-                    station_cache.add(item + (None, None))
+                    station_cache.add(item + ('', ''))
             else:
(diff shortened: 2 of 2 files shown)


ThomasLecocq:
@brenmous this looks good. could you 
- rebase this on current master
- what do you think about adding a warning.warn in the ``if not subtags:`` condition?

megies:
Hmm.. we could alternatively change this part of the code to let items with `None` values get through (or not) instead:

```python
264             info = [(net, sta, loc, cha) for net, sta, loc, cha in                             
265                     self._station_cache if                                                     
266                     fnmatch.fnmatch(net, network or '*') and                                   
267                     fnmatch.fnmatch(sta, station or '*') and                                   
268                     fnmatch.fnmatch(loc, location or '*') and                                  
269                     fnmatch.fnmatch(cha, channel or '*')]                                      
270             return sorted(info)                                                                
```

What do you think @brenmous? That way we could still allow `None` values to indicate something for that network+station code is advertized but there is no data currently.

megies:
Somehow this didn't get listed in the `1.4.1` milestone by github like it should have and got left out..

PR #2749, opened 2020-11-16
Fixed bug in nordic _str_conv and set default phase weight to ' '

WayneCrawford:
Change "if not number" to "if number is None"

### What does this PR do?

Fixes a bug in the function _str_conv(), which converts floats or ints to a string.  The original version
returned ' ' instead of '0' if the input value was 0.

### Why was it initiated?  Any relevant Issues?

Nordic files created from catalogs were having weighting set to "0" (best quality) if the catalog had
no weighting information.  The bug could affect other translations as well.

Example original output file:
```
 2019  519  6 9 45.3 L                           12 0.0                        1
 Action:ARG 20-11-16 12:55 OP:OBSP STATUS:               ID:20190519060945     I
 2019-05-19-0609-17M.MAYOB_051                                                 6
 STAT SP IPHASW D HRMM SECON CODA AMPLIT PERI AZIMU VELO AIN AR TRES W  DIS CAZ7
 MOFA E3  Pg  3A   6 949.001                                                    
 MOFA E2  Sg  2A   6 951.153                                                    
 IF5A E2  Pg   A   6 950.929                                                    
 MONA E3  Pg  1A   6 948.834                                                    
 MONA E2  Sg  3A   6 951.522                                                    
 MOSA E3  Pg  3A   6 948.846                                                    
 MOSA E2  Sg  3A   6 951.438                                                    
 MOFA E2  IAML A   6 953.057       367.4 0.10                                   
 IF5A E2  IAML A   6 953.057       496.3 0.10                                   
 MONA E2  IAML A   6 953.050       367.4 0.10                                   
 MOSA E2  IAML A   6 953.054       367.4 0.10                                   
```

Output file with all corrections, good pick at station IF5A has correct weight=0
```
 2019  519  6 9 45.3 L                           12 0.0                        1
 Action:ARG 20-11-16 13:01 OP:OBSP STATUS:               ID:20190519060945     I
 2019-05-19-0609-17M.MAYOB_051                                                 6
 STAT SP IPHASW D HRMM SECON CODA AMPLIT PERI AZIMU VELO AIN AR TRES W  DIS CAZ7
 MOFA E3  Pg  3A   6 949.001                                                    
 MOFA E2  Sg  2A   6 951.153                                                    
 IF5A E2  Pg  0A   6 950.929                                                    
 MONA E3  Pg  1A   6 948.834                                                    
 MONA E2  Sg  3A   6 951.522                                                    
 MOSA E3  Pg  3A   6 948.846                                                    
[...]

PR #2251, opened 2018-10-24
resolve full SEED ID of picks when reading NLLOC HYP files and nordic files, refactor the functionality for sh evt and hypodd pha files

megies:
### What does this PR do?

Fixes #1587 

### Why was it initiated?  Any relevant Issues?

See #1587 

### PR Checklist
- [ ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just remove the space in the following string after the + sign: "+ DOCS"
- [ ] If any network modules should be tested for the PR, add them as a comma separated list
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:"
      (you can also add "ALL" to just simply run all tests across all modules)
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are be covered via new tests.
- [ ] Any new or changed features have are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .

Might need some additional polishing/checking but mostly ready. CC @bvicic

Issue #2007, opened 2017-11-17
seishub: authentication check not working


megies:
The authentication check `Client.test_auth()` seems broken and always returns `True`. Probably not worth fixing given that seishub kind of is replaced by [jane](http://github.com/krischer/jane).

CC @trichter

barsch:
lets just deprecate obspy.seishub and obspy.db ...

krischer:
Hehe - @megies would not like that ;-)

https://github.com/obspy/obspy/pull/1947

But we should IMHO really consider deprecating them long-term.

megies:
I'm not intending to work on this issue, it's just for information..

trichter:
@megies Thx, I used to use seishub for querying local cluttered waveform files, but not anymore.

trichter:
> lets just deprecate obspy.seishub and obspy.db ...

I don't know the current status, but with sqlalchemy 1.4.0 we have new test fails in the db module.

See #2720 and https://tests.obspy.org/113566/

barsch:
fixed in #2815

PR #2814, opened 2021-03-17
I/O support for New Nordic format

flixha:
### What does this PR do?

This PR implements read and write support for the [New Nordic format (see bottom of linked page](https://seis.geus.net/software/seisan/node239.html)). The New Nordic format currently only affects "type 4 lines", that means picks. Compared to the previous version, it can now in addition store the instrument, network and location codes, agency and author for each pick, and it always saves seconds with `%5.3f` now.  As the format has the same fixed line width, the order of other parameters (e.g., pick time, amplitude, residuals, etc. is changed compared to the previous version).

Active PRs related to Nordic format I/O: #2749, #2251

Initiated to enable direct I/O of Nordic files in the New format. 

**For reading New Nordic format, the following points are not yet fully supported:**

- ~~read/write magnitude for each amplitude pick (e.g., for each trace with amplitude pick). I believe I now have an idea that I would need to add these magnitudes as `obspy.core.event.magnitude.StationMagnitude`, which then links to the `obspy.core.event.magnitude.Amplitude` that are correctly read already. But I'll need to check that; this wasn't possible in the Old Nordic format.~~(addressed now, see list below)
- as for the old format, when writing to Nordic format then amplitude picks are written without `arrival.distance` and `arrival.azimuth` (azimuth at source), because amplitude measurements are not associated with an arrival in the `obspy.core.event`, right? 

**Other points:**
- I have done some tests myself and added tests to the code, but I have not yet tested this exhaustively. It may be useful to compare the conversion from Nordic --> QuakeML with obspy to the results from [`nor2qml`](http://seis.geus.net/software/seisan/node147.html) that is distributed with Seisan does it (which is the standalone version of the [online nor2qml tool](http://nnsn.geo.uib.no/nor2qml/)).
- To write to the New Nordic format, the user now needs to invoke `obspy.core.io.nordic.core.write_select` with `version='NEW'`, while `version='OLD'` is currently the default. Should this stay like that, or should we include a depreciation warning so that `version='NEW'` can become the default in some future version?
- In the New format, there are now three fields (`PAR1`, `PAR2`, `RES`) that can contain different parameters depending on what type of observation (e.g., phase pick, amplitude pick, coda pick) the line contains. That makes it a bit more tricky to check what type of observation a line contains than previously. Here I tried to implement this logic in `io.nordic.core.check_nordic_format_version` as similar as possible as is done in Seisan (see `subroutine check_if_phase_line_format` in `LIB/indata.for` in Seisan source code)
- I have now implemented I/O of magnitude residuals associated with each amplitude measurement as `obspy.core.event.magnitude.StationMagnitude`. As the New Nordic format really stores the magnitude residuals for each amplitude measurement and not the magnitude computed for that trace, reading in Nordic file now stores the residual as `obspy.core.event.magnitude.StationMagnitude.mag_errors.uncertainty`. This way, it's possible to read & write magnitude residuals from/to Nordic format. But I guess `mag_errors.uncertainty` is not exactly the same as a magnitude residual, so if you think this may cause compatibility problems, please let me know.

**Changes that affect reading/writing of the Old Nordic format:**
- The PR contains a small change for parsing hours / minutes / seconds both in the New and the Old format: In Seisan, empty hours / minutes / seconds are acceptable, and they correspond to Zero. This PR allows empty hours / minutes / seconds to be parsed as Zeros.
- When writing, the operator ID can now be shorter than 4 characters (this is as in Seisan), while it previously had to be exactly 4 characters
- When reading an Old Nordic file, the network code is now `"  "` instead of `"NA"`, because [ `"NA"` is a network code that is actually in use](http://ds.iris.edu/mda/NA/). Previously it stood for "not available". Please comment if you don't like this change.

### PR Checklist
- [] If the PR is making changes to documentation, docs pages can be built automatically.
- [] ~~If any network modules should be tested for the PR, add them as a comma separated list
- [X] All tests still pass.

This may be of interest to @calum-chamberlain , @WayneCrawford , @d-chambers


flixha:
I know added the I/O of magnitude residuals for each amplitude measurement and described it the main post. 
From what I can see the tests fail somewhere in `sqlalchemy` when testing the obpsy-clients. I don't really see how my commits could have led to that.. any ideas?

barsch:
> 
> 
> I know added the I/O of magnitude residuals for each amplitude measurement and described it the main post.
> From what I can see the tests fail somewhere in `sqlalchemy` when testing the obpsy-clients. I don't really see how my commits could have led to that.. any ideas?

those errors are unrelated - see #2007 and #2815

flixha:
I fixed a formatting bug and some checks when reading Old Nordic format and writing to New Nordic format that were related to 

- `creation_info.agency_id` and `creation_info.author`
- how empty network code is parsed

The full description of these is in the main list in the top comment.

obspy/io/nordic/core.py
```diff
@@ -898,13 +1162,17 @@ def blanksfile(wavefile, evtype, userid, overwrite=False, evtime=None):
         f.write(' ' + write_wavfile + '6'.rjust(79 - len(write_wavfile)) +
                 '\n')
         # Write final line of s-file
-        f.write(" STAT SP IPHASW D HRMM SECON CODA AMPLIT PERI AZIMU" +
-                " VELO AIN AR TRES W  DIS CAZ7\n")
+        if version == 'OLD':
```
calum-chamberlain, line 1165: It might be worth defining these outside these functions (here and `write_select`) to avoid possible typos?

calum-chamberlain (approved):
I had some really minor comments that you might want to consider. Otherwise this looks good to me.

calum-chamberlain:
> as for the old format, when writing to Nordic format then amplitude picks are written without arrival.distance and arrival.azimuth (azimuth at source), because amplitude measurements are not associated with an arrival in the obspy.core.event, right?
Right - `Amplitude`s aren't linked to `Arrival`s directly, but might be linked via a `Pick`, but it isn't common AFAIK.

> I have done some tests myself and added tests to the code, but I have not yet tested this exhaustively. It may be useful to compare the conversion from Nordic --> QuakeML with obspy to the results from nor2qml that is distributed with Seisan does it (which is the standalone version of the online nor2qml tool).

This is a great suggestion - I was trying to make a collection of Nordic files that broke my original code, that might have been useful, but I think all I have now are the test files here. Is there a similar `qml2nor` that we could compare round-trips for?

> To write to the New Nordic format, the user now needs to invoke obspy.core.io.nordic.core.write_select with version='NEW', while version='OLD' is currently the default. Should this stay like that, or should we include a depreciation warning so that version='NEW' can become the default in some future version?

I think this depends on the status of SEISAN updates - I probably wouldn't change to `"NEW"` until the version of seisan that supports the new nordic format is in general use. Happy to be over-ruled though.


Thanks heaps for this @flixha - it is great to have more eyes on this code.

obspy/io/nordic/core.py
```diff
@@ -660,6 +661,118 @@ def _read_picks(tagged_lines, new_event):
...
+        # If they are int, then it is probably new nordic format.
+        try:
+            if str(int(old_format_secs.replace(' ', '').replace('A', '')))\
+                    == old_format_secs:
+                nordic_format = 'NEW'
+        except ValueError:
+            nordic_format = 'OLD'
+            pass
+
+        if nordic_format == 'NEW' or nordic_format == 'UKN':
```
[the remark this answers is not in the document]
flixha: Yes they are! This particular code is largely copied from the function `check_if_phase_line_format` in Seisan. I'm not sure I would have been able to make up the logic like this as easily myself, so I thought it's safest to use the logic that's been tested some in Seisan already. Did a tiny change now to make it more readable though.

obspy/io/nordic/core.py
flixha: good point; also to avoid the duplication. done!

obspy/io/nordic/core.py
```diff
@@ -710,7 +823,7 @@ def _read_picks(tagged_lines, new_event):
         # Create a new obspy.event.Pick class for this pick
         _waveform_id = WaveformStreamID(station_code=line[1:6].strip(),
                                         channel_code=line[6:8].strip(),
-                                        network_code='NA')
+                                        network_code='  ')
```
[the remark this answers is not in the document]
flixha: Yes I think None is better; which means one can leave out the line completely.

obspy/io/nordic/core.py
```diff
@@ -660,6 +661,118 @@ def _read_picks(tagged_lines, new_event):
...
+                        line[10:12] == 'IV' or
+                        line[10:12] == 'AM' or
+                        line[10:12] == 'AP' or
+                        (line[10:11] == ' ' and
+                         line[9:10] == 'I' or line[9:10] == 'E')):
+                    is_phase = True
+                    nordic_format = 'OLD'
+                    return nordic_format, is_phase
+    # If neither New nor Old format, return "UKN"
+    return nordic_format, is_phase
```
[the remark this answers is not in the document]
flixha: There could be a pick with no phase information (not even "E" or "I"). Then the function correctly returns `is_phase=False`, but it can still sense the new vs old Nordic version based on the time in the line. Not that that's a common occurrence, but with files from old systems I don't want to rule out ever encountering a strange edge case.

But you're right that the `is_phase`-command is not currently made use of anywhere else...

flixha:
Thanks a lot for the review! I did a few minor changes according to the comments above, and commented what changed.

- I agree that the default call to the function should stay with `format='OLD'` for a while.
- Yes there is also a qml2nor, distributed with the [Seisan Alpha versions](ftp://ftp.geo.uib.no/pub/seismo/SOFTWARE/SEISAN/ALPHA/linux64-v11.2.tar.gz); so we can eventually check

obspy/io/nordic/core.py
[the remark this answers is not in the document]
calum-chamberlain: Ah yup, sorry, I missed the bit in lines 714-718 - I thought `nordic_format` was only ever getting reset at line 767 and 746. :+1:

calum-chamberlain:
@flixha if you are happy then I'm happy for this to go as is I think. @megies are you okay if we merge this into master?

flixha:
> @flixha if you are happy then I'm happy for this to go as is I think. @megies are you okay if we merge this into master?

Hi Calum,
I am suggesting another set of changes:

- read in type-3 lines (comments in Nordic files) and save content to `obspy.core.event.comments` as `obspy.core.event.Comment`
- read in waveform filenames into comments with a prefix "Waveform-filename: "
- On writing with a Nordic file, `write_comment`will recognize these lines starting with "Waveform-filename: ", remove the prefix, and write it as a type-6 (waveform line). Reading and writing via supplying `wavefiles` is still supported, but these changes allow the user to read & write Nordic format without having to save & load waveform-filenames from an extra variable saved alongside the catalog.
- Right now, the function continues to write the DUMMY-string by default, but I'm wondering to stop it doing that eventually.

obspy/io/nordic/core.py
```diff
@@ -637,6 +644,34 @@ def _read_moment_tensors(tagged_lines, event):
...
+    :type tagged_lines: dict
+    :returns: updated event
+    :rtype: :class:`~obspy.core.event.event.Event`
+    """
+    if '3' not in tagged_lines.keys():
+        return event
+    # Get comment lines
+    com_lines = tagged_lines['3']
+    # can contain SPEC which is read in as spectral information - 
+    # should such lines not be read as comments?
```
calum-chamberlain, line 661: Probably not? But I don't know where else they could go. There might be somewhere in QuakeML for it?

calum-chamberlain:
Those sound like good suggestions and I agree stopping the "DUMMY" wavefile would be good - can Nordic files have no type 6 line?

obspy/io/nordic/core.py
flixha: I was considering to stop th code from reading those type-3 lines that contain structured information (like SPEC, XNEAR). But then I thought that's only useful once we have fully implemented a different I/O way to handle this information. For XNEAR, nothing is implemented yet, while for SPEC, I think only reading is supported yet?

That's why I thought it may be good to keep reading and writing all type-3 lines right now, because that allows all information to be transferred from Nordic -> Obspy -> Nordic.

flixha:
Ok, thanks for your input!
Yes, Nordic files work fine without type-6 lines in Seisan. The only lines that Seisan strictly asks for are hypocenter and ID-line.

flixha:
More updates:
- fix bug where lat/lon-errors were read as lon/lat
- fix bug where origin-error was written from event RMS (standard_error) rather than time_error
- add support for I/O of apparent velocity / horizontal slowness
- add support for writing of multiple origins
- add event-type mapping between Nordic and Obspy/Quakeml (do not fully match)
- read pick-weight as pick.extra.nordic_pick_weight (was arrival.time_weight) and read finalweight into arrival.time_weight (or backazmiuth_weight, respectively) instead.
- deactivated sorting of magnitudes by their quantity for now (this leads to the undesired effect that the order of agencies constantly changes instead by default)
- added a `strict`-parameter to the similarity-tests for the unittests. `strict=True` is intended for a direct comparison of obspy-objects for event read from and written to Nordic files, to catch regressions that the tests didn't check. `strict=False` is for comparing Quakeml vs.Nordic, where the mapping is not complete.

Let me know if you disagree with any of the changes above @calum-chamberlain or have other comments!

flixha:
> all this looks good. Are there elements from the discussion above / explanations that should go in the documentation of the format ? otherwise it looks good to go

Thanks! Based on your suggestion, I have added a note to the docs regarding limitations of the functions when converting Nordic --> Obspy --> Nordic. I have also indicated in the docs a few changes in how the functions treats pick-weights, empty network codes, and that it does no longer sort magnitudes by size. All other bug fixes and new functionality are in the Changelog.

Other small changes make the functions more robust against None-values or non-existant fields (in stationXML / obspy.event), but I think those changes do not need to be explicitly mentioned in the docs or Changelog. So I am happy for this to be merged!

calum-chamberlain:
Hey @flixha - can you fix the merge conflict above, then I am keen to have a quick look over the code and merge if you are happy that you have fixed all of the issues I made (thanks!!!) and got this to where you want to be?

obspy/io/nordic/core.py
```diff
@@ -668,27 +878,25 @@ def _read_picks(tagged_lines, new_event):
...
         # or undecidable.
         # It is valid nordic for the origin to be hour 23 and picks to be hour
         # 00 or 24: this signifies a pick over a day boundary.
-        pick_hour = int(line[18:20])
-        pick_minute = int(line[20:22])
-        pick_seconds = float(line[22:29])  # 29 should be blank, but sometimes
-        # SEISAN appears to overflow here, see #2348
+        pick_hour = int(line[18:20].strip() or 0)
+        pick_minute = int(line[20:22].strip() or 0)
+        pick_seconds = float(line[22:29].strip() or 0)  # 29 should be blank,
```
calum-chamberlain, line 898:
```suggestion
        pick_seconds = float(line[22:29].strip() or 0.0)  # 29 should be blank,
```

calum-chamberlain (approved):
Looks great, sorry this took so long :( The failing tests don't look like they have anything to do with this code. I'm just going to run the test-suite locally on my Windows machine to check, and if they pass I will merge if you are okay with that @flixha ?

calum-chamberlain:
I have a handfull of unrelated test fails on Windows Python 3.8 (response comparison and data download). I am happy to merge with your final sign off @flixha

flixha:
@calum-chamberlain Great, thanks a lot for checking and I understand you were busy (and hopefully having fun) in the field! 
If you can approve this to be merged, that would be perfect - ready to go for me!

calum-chamberlain:
Done, great work @flixha - thank you, I'm sure lots of people will appreciate your efforts here!

PR #2866, opened 2021-08-12
adding apply_calib=False by default for Kinemetrics' EVT format

ThomasLecocq:
this is a replacement for https://github.com/obspy/obspy/pull/2585 ; I didn't manage to clean/rebase properly, so here we are.

still have to add warning that the default behaviour is now to return the RAW data and no longer the corrected ones


ThomasLecocq:
@grapagrani this changes the calls (raw is not apply_calib) and behaviour is flipped....

ThomasLecocq:
test failures are unrelated

PR #2878, opened 2021-09-13
FDSN client: Don't reject valid hostnames/TLDs

acarapetis:
### What does this PR do?

Updates the URL validation regex in `obspy.clients.fdsn` to accept some valid hostnames and TLDs that were previously rejected.

PR #1162 introduced a validation regex to the FDSN client which rejects TLDs of length greater than 6. According to RFC 1034, TLDs up to length 63 are valid.

The regex also allowed only alphabetic characters in TLDs, which excludes the usage of Internationalized Domain Names (which start with xn-- and may include digits).

Neither of these concerns are theoretical - see the list of registered TLDs here, which include many long examples and many internationalized examples:

http://data.iana.org/TLD/tlds-alpha-by-domain.txt

The example that prompted me to submit this patch is a private one - we use the private DNS zone ".private" in our systems, which the FDSN client was rejecting for being too long.

Finally, the regex for unqualified hostnames was not allowing hyphens.

This PR adds some test cases to illustrate the issues above, and alters the URL regex to validate them. I also made a small change to the validation test to give better feedback on failures.

Based on the comments in #1162, I understand that the decision to disallow "generic TLDs" was an intentional one, but I think this is misguided - FDSN is a HTTP-based protocol, so not allowing connection to valid HTTP URLs seems wrong. *Maybe* a warning could be justified, but currently obspy outright refuses to connect to the domain names we use, so we have to work around it by resolving IP addresses manually.

Honestly I think all that's really necessary is `^https?://` - this would catch the original mistake of typos in abbreviations without stepping on any toes - but I stuck to the current model for the sake of minimizing change.

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS: clients.fdsn"
- [x] All tests still pass.


claudiodsf (approved):
Fine for me, provided that tests pass.

Now that CI is fixed (thanks @ThomasLecocq!), @acarapetis, could you please rebase and force push to trigger the tests? Thanks!

acarapetis:
@claudiodsf It looks like the CI is still broken on the maintenance_1.2.x branch - did you want me to rebase+retarget this at `master`?

claudiodsf:
> @claudiodsf It looks like the CI is still broken on the maintenance_1.2.x branch - did you want me to rebase+retarget this at `master`?

Yes, please 😉

ThomasLecocq:
What do you think about adding some comment about "valid" hostname/TLD in  https://github.com/obspy/obspy/blob/254fcf6f37bab857d10fafcc441b50624438a26c/obspy/clients/fdsn/client.py#L160-L162
?

claudiodsf:
> What do you think about adding some comment about "valid" hostname/TLD in

Not sure that it is necessary.

Thanks to this PR, all valid hostnames/TLD should be now covered and the check should fail only for rare and clearly wrong cases.

ThomasLecocq (approved):

ThomasLecocq:
thanks 🟢

PR #2886, opened 2021-10-20
~10x speedup for selecting traces from Stream by non-wildcarded trace ID

flixha:
### What does this PR do?
Speed up the simple selection of traces in a stream by their fixed trace-ID (i.e., using `st.select(id=...)`, with no wildcards or single NSLC-codes involved). Test on a 512-trace stream shows a >10x speedup.

`stream.select`is the most concise and convenient function to select specific traces from a stream, but if that function is called on larger streams (~> 100 traces) it can be surprisingly slow. The reason appears to be that it uses several invocations of `fnmatch.fnmatch` for each trace and for all string comparisons even if there are no wildcards. For this type of usage, it's easy to just revert to standard string-comparison in python. Some codes use the `select` so often that this PR makes a noteable difference for total execution time (e.g., ~35 % speedup for `eqcorrscan.utils.catalog_to_dd.write_correlations`).

Demonstration of the speedup for `st.select()` (test on master-branch, this PR speeds up formulation A to 0.45 ms per loop):
```python
import obspy
st = obspy.read()
for j in range(1,10):
    st = st + st

# A: 4.77 ms per loop:
%timeit st.select(id='BW.RJOB..EHZ')
# B: 2.9 ms per loop:
%timeit obspy.Stream([tr for tr in st  if tr.id=='BW.RJOB..EHZ'])
# C: 0.36 ms per loop:
%timeit obspy.Stream([tr for tr in st  if tr.stats.network=='BW' and tr.stats.station == 'RJOB' and tr.stats.location == '' and tr.stats.channel=='EHZ'])
```

### PR Checklist
- [X] All tests still pass. (Ubuntu tests here on github succeed)


d-chambers:
Any idea what's going on with the macos fails?

flixha:
> Any idea what's going on with the macos fails?

I am really unsure what those mean, I'm hoping someone may be able to help. It doesn't seem like those were directly related to any small changes that I did with this PR. I tested this PR on my own Mac (MacOS 10.13.7 / Darwin 17.7.0) for both python versions, and both succeeded:
https://tests.obspy.org/114682/ (python3.8)
https://tests.obspy.org/114684/ (python3.7)

The tests runs that fail produce lots of error messages similar to `TypeError: int() argument must be a string, a bytes-like object or a number, not '_NoValueType'` (or with `float`) in calls to `max()` in numpy. Looking for any differences between the failed runs and my runs, I can see a tiny difference for numpy (GH-Action-runs with numpy1.21.2 failed, my own with numpy1.21.3 worked), and of course there's the difference for the actions running on macOS 10.15.6 beta 2: https://tests.obspy.org/114657/

I can at least find one successful recent test on macOS 11.5 beta2 (Darwin 20.6.0): https://tests.obspy.org/114619/

I just triggered a new Action run to see if there's any change, let's see.

I don't have access to a system with macOS 10.15.6 beta 2 to test (why beta?!), and I don't think I have any access to the Obspy-actions to use these for debugging? Maybe someone has other suggestions how to debug on a specific system?

d-chambers:
@flixha I made a "hard fork" of ObsPy on my personal github account (hard so the CIs are not connected) where I test the CI. You can do this by cloning ObsPy, then adding another remote pointing to your fork and manually pushing. Then you can test CI on your own and it wont notify all the ObsPy watchers each time you push to your own fork. 

The fails are clearly not related to your PR though.

d-chambers:
This looks like a good contribution. One question:

How much does this slow down `select` when wildcards are used?

flixha:
@d-chambers thanks for the point on the forks. I guess I kind of have set up a hard fork, but haven't really used its full potential for the actions, so it's great to get a tip in the right direction. I'll try that out!

Good question regarding using the wildcards, here's the numbers:
Previously: `%timeit st.select(id='B?.RJO?.*.EHZ')` --> 4.0 ms per loop
With this PR:`%timeit st.select(id='B?.RJO?.*.EHZ')` --> 4.2 ms per loop (i.e., 5 % longer runtime)

d-chambers:
Ok, I think it is worth 5% slower wildcard queries for the 10x speed up without them. Anyone else have an opinion?

d-chambers:
Ok, I guess no one else cares. Can you add a quick changelog entry then I think we are good to merge.

flixha:
> Ok, I guess no one else cares. Can you add a quick changelog entry then I think we are good to merge.

Thanks for inviting to get more feedback and reviewing the PR! I added a line to the Changelog now, so then it should be ready to go.

d-chambers (approved):

ThomasLecocq:
@d-chambers scrolling through the PRs and this is indeed unrelated to MacOS sh**t testing , and I don't see why macos or win would behave differently here (nothing related to paths or stdlib) - good to go for me

claudiodsf:
Continous Integration works again, thanks to @ThomasLecocq! 🥳 

@flixha: could you please rebase and force push, to trigger all the tests?

flixha:
@claudiodsf  claudiodsf 
> Continous Integration works again, thanks to ThomasLecocq! 
> 
> @flixha: could you please rebase and force push, to trigger all the tests?

I rebased and triggered the tests to run again.

ThomasLecocq:
@claudiodsf why is the changed files list soooo long ? not forced-rebased ?

flixha:
@ThomasLecocq @claudiodsf
> @claudiodsf why is the changed files list soooo long ? not forced-rebased ?

Yes that was the mistake. I fixed it now.

ThomasLecocq:
ok, 🟢 light from the CI, merging.

For the future, was wondering if there wouldn't be a more clever way of using `re` for matching trace ids than sequentially over each parameter. Aaaanyway, for later.

Issue #2912, opened 2022-01-01
Incompatibility with numpy 1.22


yetinam:
Obspy (1.2.2, installed via pip) seems to be incompatible with numpy 1.22 (which is automatically installed through pip when installing obspy). When importing obspy, an AttributeError occurs. I pasted the full error log below. The problem does not occur when using numpy 1.21. The problem occured on Ubuntu, MacOS and Windows in python 3.8 and 3.9.

The issue occurs here: https://github.com/obspy/obspy/blob/c4faef6b2df110e844c28ca2aa9515888c019cb8/obspy/core/util/misc.py#L205-L215
From what I can tell, this code block checks whether the `numpy.loadtxt` function accepts a parameter `ndmin`. However, this parameter has been introduced in numpy 1.6.0, while obspy requires at numpy>=1.15.0. Therefore, I'd suggest to simply remove the function, which should also resolve the version incompatibility issue.

```
In [1]: import obspy
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-42a9ad3b2ac1> in <module>
----> 1 import obspy

~/anaconda3/envs/tmp_version_tests/lib/python3.9/site-packages/obspy/__init__.py in <module>
     37 
     38 # don't change order
---> 39 from obspy.core.utcdatetime import UTCDateTime  # NOQA
     40 from obspy.core.util import _get_version_string
     41 __version__ = _get_version_string(abbrev=10)

~/anaconda3/envs/tmp_version_tests/lib/python3.9/site-packages/obspy/core/__init__.py in <module>
    122 
    123 # don't change order
--> 124 from obspy.core.utcdatetime import UTCDateTime  # NOQA
    125 from obspy.core.util.attribdict import AttribDict  # NOQA
    126 from obspy.core.trace import Stats, Trace  # NOQA

~/anaconda3/envs/tmp_version_tests/lib/python3.9/site-packages/obspy/core/utcdatetime.py in <module>
     25 import numpy as np
     26 from obspy.core.compatibility import py3_round
---> 27 from obspy.core.util.deprecation_helpers import ObsPyDeprecationWarning
     28 
     29 # Regular expression used in the init function of the UTCDateTime objects which

~/anaconda3/envs/tmp_version_tests/lib/python3.9/site-packages/obspy/core/util/__init__.py in <module>
     25 # import order matters - NamedTemporaryFile must be one of the first!
     26 from obspy.core.util.attribdict import AttribDict
[...]

PR #2913, opened 2022-01-02
Remove wrapper for numpy loadtxt

yetinam:
### What does this PR do?

Removes obsolete wrapper around numpy loadtxt and changes all usages from the wrapped version to the original. This wrapper was causing import errors when using numpy 1.22. See #2912 for details.

As the issue caused an import error and all usages of loadtxt should anyhow have test coverage already, I did not add any new tests.

Some tests are failing, at least in my local installation, but from what I can tell this is unrelated.

#2912 

### PR Checklist
      (e.g. `clients.fdsn,clients.arclink`) after the colon in the following magic string: "+TESTS:clients.iris"
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/iris/client.py
+++ b/obspy/clients/iris/client.py
@@ -17,6 +17,7 @@ import io
 import platform
 import sys
 from lxml import objectify
+import numpy as np
 
 if sys.version_info.major == 2:
     from urllib import urlencode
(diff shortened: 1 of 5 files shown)


QuLogic (approved):

ThomasLecocq:
@QuLogic @d-chambers this is a strong case for drafting a 1.2.3 version, isn't it?

QuLogic:
Probably, but I'm not really up to date on the releasing procedure.

ThomasLecocq:
> Probably, but I'm not really up to date on the releasing procedure.

me neither, but this is an important one here, so we should try to draft it

yetinam:
I just realized that I did not include an entry in the change log for the fix. Not sure if and how this should be amended.

ThomasLecocq:
> I just realized that I did not include an entry in the change log for the fix. Not sure if and how this should be amended.

create a small PR for adding the references to #2912 #2913

PR #2919, opened 2022-01-04
tentative removal of numpy from runtests.py -- only macos actions

ThomasLecocq:
wondering if numpy's import in runtests is responsible for the failed tests on MacOS

NOTE: this will FAIL many tests (doctests print standard is no longer fixed to legacy numpy 1.13)


ThomasLecocq:
fails miserably https://github.com/obspy/obspy/runs/4702793970?check_suite_focus=true

PR #2885, opened 2021-10-14
Fixes crash when resampling very short traces

yetinam:
### What does this PR do?

When working with a few very fragmented traces I stumbled across a corner case in the resampling logic that caused a crash. When resampling very short traces to a lower sampling rate, obspy inferred an output length of zero samples, leading to a ZeroDivisionError. This PR changes the behavior to always keep at least one sample when resampling a non-empty trace. Also adds a unit test for the bug.

### Why was it initiated?  Any relevant Issues?

see above

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] All tests still pass.
- [x] Any new features or fixed regressions are be covered via new tests.
- [x] Significant changes have been added to `CHANGELOG.txt` .

PR #2926, opened 2022-01-05
Fixes crash when resampling very short traces (duplicate of #2885)

yetinam:
This PR is a duplicate of #2885, but targeting `master` instead of `maintenance_1.2.x`. It was necessary to create as the source branch of a PR can not be changed (while the target can) and rebasing the branch onto `master` seemed considerably more error prone than cherry picking the commit onto a fresh `master` branch.

### What does this PR do?

When working with a few very fragmented traces I stumbled across a corner case in the resampling logic that caused a crash. When resampling very short traces to a lower sampling rate, obspy inferred an output length of zero samples, leading to a ZeroDivisionError. This PR changes the behavior to always keep at least one sample when resampling a non-empty trace. Also adds a unit test for the bug.

see above

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -88,6 +88,7 @@ Changes:
    * Removed obsolete wrapper around numpy.loadtxt causing import error with
      numpy 1.22 (see #2912, #2913)
    * Fix iso8601 regex for issue #2868 to cope with day 360 properly.
+   * Fix crash when resampling very short traces (see #2885)
  - obspy.clients.fdsn:
    * add URL mapping for IRISPH5 (see #2739)
    * update RESIF URL mapping to use https
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -88,7 +88,7 @@ Changes:
    * Removed obsolete wrapper around numpy.loadtxt causing import error with
      numpy 1.22 (see #2912, #2913)
    * Fix iso8601 regex for issue #2868 to cope with day 360 properly.
-   * Fix crash when resampling very short traces (see #2885)
+   * Fix crash when resampling very short traces (see #2926)
  - obspy.clients.fdsn:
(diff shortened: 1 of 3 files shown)


ThomasLecocq:
is this a desired behaviour ? I mean, if resampling a 2 sample, 100 Hz trace to 30 Hz, I'd imagine it should return an empty trace, or a np.nan, masked trace ?

yetinam:
I admit, it's a corner case. I'd personally argue for keeping always at least one sample. When resampling a 2 sample trace from 100 Hz to 30 Hz, I think the output should be the best estimate for a 30 Hz trace we can get from the original, which would be a single sample. This still contains more information than an empty trace, because it contains the bias of the trace. In addition, it's what an instrument with 30 Hz sampling rate would have recorded during the recording of the original trace (assuming it would have taken a sample at all in the time frame...).

In practical applications, I hope this shouldn't occur too often, but when treating larger datasets, this corner case will pop up. That's also how I stumbled across it. Therefore, I'd also advocate to go with an option that will cause less trouble for downstream applications. I think this is probably the one sample option rather than the empty trace, because may operations (even simple things like np.mean) will fail on the empty trace, but work fine on the single sample.

Nonetheless, I'd also be fine with changing the PR to any of the options you suggested if that turns out to be the preferred option. My main goal is to avoid the crash.

claudiodsf:
Maybe we can keep 1 sample but issue a `warnings.warn()` ?

yetinam:
Sounds good to me and seems like there is agreement on the suggestion. I'll update the PR tomorrow.

yetinam:
I added the warning and pushed the respective commit.

claudiodsf (approved):
Perfect! Good to go! :shipit:

claudiodsf:
Thanks again!

PR #2936, opened 2022-01-06
Rewriting the CI recipes to make use of mamba + cache

ThomasLecocq:
# What does this PR do?
- It implements a new way to create the environments for CI testing on github-actions.
- The recipes have been modified to install mamba & to save/reuse daily snapshots of the environments.
- There is also a new job that checks for existing "running" or "queued" jobs in _the same PR_ (!!) and cancels them, to save CI energy for the latest push, and discard outdated content.

Using cached environments saves at least 2 minutes on Ubuntu, could save up to 10 min on macOS & Windows for the environment solving step. This comes at a price: every day, the first CI run will need to cache the environment (tar-gz it and save it remotely, which is almost instantaneous on *nux, takes 2min on windows).

### DIFF
--- a/.github/workflows/master_default_tests.yml
+++ b/.github/workflows/master_default_tests.yml
@@ -13,21 +13,51 @@ env:
 jobs:
   run_tests:
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.7', '3.8', '3.9']
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        python-version: [ '3.7', '3.8', '3.9' ]
+        include:
+          - os: ubuntu-latest
+            label: linux-64
+            prefix: /usr/share/miniconda3/envs/test
+
+          - os: macos-latest
+            label: osx-64
+            prefix: /Users/runner/miniconda3/envs/test
+
+          - os: windows-latest
+            label: win-64
+            prefix: C:\Miniconda3\envs\test
 
     steps:
       - uses: actions/checkout@v2
 
-      - name: Setup conda
-        uses: conda-incubator/setup-miniconda@v2
+      - name: Get current date
+        id: date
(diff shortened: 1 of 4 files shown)


ThomasLecocq:
So, that was the first run: 
image
will now trigger a new build with a changelog entry

ThomasLecocq:
made a mistake previously, now the cache is properly created:
image

ThomasLecocq:
After caching:
image
It's not amaaaazing... 

**but:**

Ubuntu tests start with a lead time of about 85 seconds, compared to 197 seconds before.
If I re-enable the "fail tests on first matrix element failure", it'd mean the ubuntu results should already let users know something is wrong with their PR within less than 4 to 6 minutes!

ThomasLecocq:
I added a job canceller in case we push a new commit to the same PR, to avoid loosing test-minutes on outdated content.

ThomasLecocq:
question to @obspy/maintainers should the other matrix elements be cancelled when 1 fails (e.g.: ubuntu 3.8 fails, should the macOS and windows (usually slower) be cancelled, as it's "useless to run them", or should they be allowed to run, to see if the error is only on ubuntu 3.8 ?

d-chambers:
Probably let them run, why not?

claudiodsf:
I'm for letting other jobs run: it's important to know if failure is platform-related.

claudiodsf (approved):
Fine for me if all the tests are green.

I would suggest to squash all the commits before merging.

ThomasLecocq:
rebased & ready to squash-merge when 🟢

ThomasLecocq:
merged 🟢 , let's see what master test complain about :-)

claudiodsf:
But I would have preferred to squash commits before 🙄

ThomasLecocq:
why ? it's squashed on master

ThomasLecocq:
🟢 all tests work on master, me happy :-)

claudiodsf:
> why ? it's squashed on master (I always do "Squash & Merge" when merging)

Ah ok, didn't know that 🙄

PR #2937, opened 2022-01-06
Improve _is_pickle check (>25 second test on windows)

ThomasLecocq:
### What does this PR do?

Improves the `_is_pickle` method by checking for a known pattern (b``obspy.core.stream``) in the lead part of the read file. This way, we avoid to pass the content to `pickle.load`, which for no clear reason takes more than 25 seconds on Windows (only). 
Example: https://tests.obspy.org/115617/ -->
>33.182s | test_is_format (obspy.core.tests.test_waveform_plugins.WaveformPluginsTestCase)

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3568,7 +3568,7 @@ def _is_pickle(filename):  # @UnusedVariable
     if isinstance(filename, str):
         try:
             with open(filename, 'rb') as fp:
-                if "obspy.core.stream" in fp.read(100):
+                if b"obspy.core.stream" in fp.read(100):
                     fp.seek(0)
                     st = pickle.load(fp)
                     return True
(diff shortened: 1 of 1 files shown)


ThomasLecocq:
DO NOT MERGE YET, am playing with slow tests :-)

ThomasLecocq:
if good for you, good for me :-) I'll dive deeper in the slow tests later, but that was an easy one :-)

d-chambers (approved):
This looks like a great idea to just avoid the full pickle load! Do you know if it works on all pickle versions?

ThomasLecocq:
As far as I remember, the code tells you that you can only read pickles written by the very same version. And I guess that's the reason the pickle tests are called with a live-made tmp file, nothing saved in the tests/data folder.

PR #2944, opened 2022-01-13
tune action workflows

trichter:
Add outcome of network tests to checks


d-chambers:
Hey @trichter, good idea. We need it to indicate there was a problem but not fail the build. I was also wondering if we might want to create a second conditional job for network tests and only run one os/py version or maybe just a few rather than running them on the whole matrix. What do you think?

trichter:
I added the status of network tests now only for ubuntu-latest. I was not able to solve an error with curl on windows. In any case, network tests are running on all platforms. I also changed master default tests in the same way and removed the network tests workflow.

trichter:
OK, i switched to a new job for network tests with `continue-on-failure` on job level as already suggested by @d-chambers. Adding the statuses manually was too much of a hassle. I think the current state of the PR should result in the desired behavior.

d-chambers (approved):
Looks good to me. The only question I have is about the `if: always()`; this just ensures the step will run regardless of previous steps failing correct?

d-chambers:
One other inconsequential suggestion while we are making changes: Maybe we change the name of the job from "pull_request_tests" to just "PR_tests" or "PR" to reduce the amount of text in the CI summary window: 

image

After that feel free to merge.

trichter:
> The only question I have is about the if: always(); this just ensures the step will run regardless of previous steps failing correct?

Yes, but I had to tune this expression further more to ensure that it only runs if the tests got executed.

PR #2949, opened 2022-01-17
Fix NumPy deprecations

QuLogic:
### What does this PR do?

Fixes many warnings from NumPy when running tests.

NumPy is now warning about aliases to the builtin types, on Python 3.7+, which is our minimum. 

It is also warning that `np.matrix` is probably not what you want to use. While it is sort of intentional in this case, we can swap to `np.array` easily since most of the code uses `np.dot` and not multiplication.

### PR Checklist
- [x] All tests still pass.


ThomasLecocq:
maybe a changelog entry for this one ?

obspy/imaging/maps.py
```diff
@@ -468,7 +468,7 @@ def linspace2(val1, val2, n):
     # scatter plot is removing valid x/y points with invalid color value,
     # so we plot those points separately.
     try:
-        nan_points = np.isnan(np.array(color, dtype=np.float))
+        nan_points = np.isnan(np.array(color, dtype=float))
```
d-chambers, line 471: Would it be better to specify a `np.float64` rather than native float type?

obspy/imaging/scripts/mopad.py
```diff
@@ -772,13 +776,19 @@ def _find_faultplanes(self):
         """
         # reference Double Couple (in NED basis)
         # it has strike, dip, slip-rake = 0,0,0
-        refDC = np.matrix([[0., 0., -1.], [0., 0., 0.], [-1., 0., 0.]],
-                          dtype=np.float)
+        refDC = np.array([
+            [0, 0, -1],
+            [0, 0, 0],
+            [-1, 0, 0],
+        ], dtype=float)
```
d-chambers, line 783: Same comment here, it may be a bit more robust to use a specific np float type here.

obspy/imaging/scripts/mopad.py
```diff
@@ -772,13 +776,19 @@ def _find_faultplanes(self):
...
         refDC_evals, refDC_evecs = np.linalg.eigh(refDC)
 
         # matrix which is turning from one fault plane to the other
-        flip_dc = np.matrix([[0., 0., -1.], [0., -1., 0.], [-1., 0., 0.]],
-                            dtype=np.float)
+        flip_dc = np.array([
+            [0, 0, -1],
+            [0, -1, 0],
+            [-1, 0, 0],
+        ], dtype=float)
```
d-chambers, line 791: and here

d-chambers (changes requested):
Looks good.  The only change I recommend, which might help us avoid some future OS-specific bugs, is to use explicit numpy dtypes for float, int, and complex (np.float64, np.int64, np.complex128 ).

obspy/imaging/maps.py
QuLogic: Maybe? But these were aliases, so this just fixes the warning without making a choice on the matter.

ThomasLecocq: @QuLogic what you mean is:
currently; on a 32bits system, np.float was just an alias for native float32 ; and on a 64bits system, np.float was float64 ? right ?

ThomasLecocq:
just wondering also, are there still any 32bits system around those days ?

obspy/imaging/maps.py
QuLogic: No, what I meant is that `np.float` is an alias for `float`, and @d-chambers is asking about what you mentioned. 

From testing it out though, using `float` is always 64-bit, even on a 32-bit build (tried on Linux, don't know about Windows).

QuLogic: It _does_ have an effect on `int`, which is 32-bit on 32-bit and 64-bit on 64-bit, but that's the same as `np.int`, because they're the same thing.

d-chambers: Yes, I have run into `int` issues on windows before and had to specify np.int64. Good point though, you aren't changing any behavior though, so its probably fine. I will merge.

d-chambers (approved):

PR #2954, opened 2022-01-17
deploy docs to server

trichter:
WIP

### DIFF
--- a/misc/scripts/docs_deploybot/deploy_docs.py
+++ b/misc/scripts/docs_deploybot/deploy_docs.py
@@ -1,0 +1,165 @@
+# Copyright 2022, GNU LGPL, Obspy developers
+from datetime import datetime, timezone
+from io import BytesIO
+import logging
+import os
+import shutil
+import tarfile
+from zipfile import ZipFile
+
+import requests
+
+
+class DeployError(Exception):
+    pass
+
+
+logging.basicConfig(level=logging.INFO, format='%(message)s')
+log = logging.getLogger('deploy_docs')
+TOKEN = os.getenv('GITHUB_TOKEN', '')
+RKW = {'headers': {'accept': 'application/vnd.github.v3+json',
+                   'authorization': f'token {TOKEN}'}}
+API = 'https://api.github.com/repos/obspy/obspy/'
+BASE = '/home/obspy/htdocs/docs/'
+PATH = {'pr': BASE + 'pr/{pr}/',
+        'master': BASE + 'master/',
+        'stable': BASE + 'stable/',
+        'archive': BASE + 'archive/{v}/'}
+URL = 'https://docs.obspy.org/pr/{pr}'
+
+
+def get_runs(event=None, time=0):
+    data = {'status': 'success', 'per_page': 100}
+    r = requests.get(API + 'actions/runs', params=data, **RKW)
+    r.raise_for_status()
+    runs = r.json()['workflow_runs']
+    log.info(f'Requested {len(runs)} runs')
+    for run in runs:
+        t = datetime.fromisoformat(run['updated_at'][:-1])
+        run['time'] = t.replace(tzinfo=timezone.utc).timestamp()
+    runs = [run for run in runs if
+            run['time'] > time and
+            run['name'] == 'docs' and
+            run['conclusion'] == 'success' and
+            (event is None or run['event'] == event)]
+    runs = sorted(runs, key=lambda run: -run['time'])
+    log.info(f'Selected {len(runs)} runs')
+    return runs
+
+
+def deploy_artifact(run, other_path=None, overwrite=False):
+    if run['event'] == 'pull_request':
+        if len(run['pull_requests']) != 1:
+            raise DeployError(f"found {len(run['pull_requests'])} PRs")
+        pr = run['pull_requests'][0]['number']
+        path = (other_path or PATH['pr']).format(pr=pr)
+        overwrite = False
+    elif run['event'] == 'push':
+        pr = None
+        path = (other_path or PATH['master'])
+    elif run['event'] == 'release':
+        pr = None
+        version = run['github_ref']
+        path = (other_path or PATH['archive']).fomat(v=version)
+    else:
+        raise DeployError(f"unexpected event {run['event']}")
+    msg = (f"Check run {run['id']} triggered by {run['event']}, "
+           f"PR: {pr}, {run['conclusion']}")
+    log.info(msg)
+    if (os.path.exists(path) and
+            run['time'] < os.path.getmtime(path) + 1 and
+            not overwrite):
+        return
+    assert run['conclusion'] == 'success'
+    log.info(f'Deploy to {path}')
+    r = requests.get(run['artifacts_url'], **RKW)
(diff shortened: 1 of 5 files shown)


ThomasLecocq:
just a note for later: we'll have to be careful with the ``no_ci`` tag and limit it to cases where the PR is **only** about documentation pages (e.g. changing the Contributing page etc), and be careful that there are no changes to the codebase which _could_ fail with regular tests.

trichter:
> just a note for later: we'll have to be careful with the no_ci tag and limit it to cases where the PR is only about documentation pages (e.g. changing the Contributing page etc), and be careful that there are no changes to the codebase which could fail with regular tests.

True. I rather see it as a flag which can be used when work is still in progress. It should be removed at the end if any changes to the code base were introduced.

> And I would still run the lint_code part even with no_ci.

Good idea. Will do.

d-chambers:
> True. I rather see it as a flag which can be used when work is still in progress. It should be removed at the end if any changes to the code base were introduced.

This is what I had in mind as well when creating the tag. If you are pushing a flurry of commits (https://xkcd.com/1296/) it seems nice to give the CI servers a break. Ideally, I would like to see the tests run before merging anything though.

trichter:
This PR is ready for review and merging. Docs are deployed to docs.obspy.org. See the link in the checks tab for the doc for this PR. I also tested deploying to master directly from this branch (see docs.obspy.org/master). It appears to work as expected. Hopefully the release event can be tested soon, too...

d-chambers:
Hey @trichter, at first glance it looks good to me but I am going to defer to either @ThomasLecocq  or @barsch for a proper review.

The docs build bot stuff runs server side no?

trichter:
> The docs build bot stuff runs server side no?

Yes I started the `deploybot.py` script with nohup on the server.

trichter:
I think it is better to post the status only when the bot finished its job. Will push and update.

misc/scripts/docs_deploybot/remove_old_pr_docs.py
```diff
@@ -0,0 +1,25 @@
...
+from glob import glob
+import logging
+import os
+import shutil
+
+
+logging.basicConfig(level=logging.INFO, format='%(message)s')
+log = logging.getLogger('remove_docs')
+GLOB = '/home/obspy/htdocs/docs/pr/*'
+GLOB = '/home/eule/tmp/odoc/pr/*'
```
ThomasLecocq, line 11: this should be removed?

misc/scripts/docs_deploybot/deploy_docs.py
```diff
@@ -0,0 +1,158 @@
...
+TOKEN = os.getenv('GITHUB_TOKEN', '')
+RKW = {'headers': {'accept': 'application/vnd.github.v3+json',
+                   'authorization': f'token {TOKEN}'}}
+API = 'https://api.github.com/repos/obspy/obspy/'
+BASE = '/home/obspy/htdocs/docs/'
+PATH = {'pr': BASE + 'pr/{pr}/',
+        'master': BASE + 'master/',
+        'stable': BASE + 'stable/',
+        'archive': BASE + 'archive/{v}/'}
+URL = 'https://docs.obspy.org/pr/{pr}'
```
ThomasLecocq, line 24: is it normal that the URL is not built from the PATH? (or only for pr) ?

ThomasLecocq (approved):

ThomasLecocq:
is it possible to make the "deploy link" use target="_blank" ? :-)

misc/scripts/docs_deploybot/deploy_docs.py
trichter: Yes that is normal, it can be configured. The advantage is that the stable branch sits in an extra dictionary and not in root.

trichter:
> is it possible to make the "deploy link" use target="_blank" ? :-)

It is possible, but do we really want this? All other links in this tab do not use target="_blank".

ThomasLecocq:
> > is it possible to make the "deploy link" use target="_blank" ? :-)
> 
> It is possible, but do we really want this? All other links in this tab do not use target="_blank".

I wish they did 🤣

ThomasLecocq:
Once the last leftover from your rebase is gone, it's good to go for me ! Amazing job on this @trichter , the doc-build-bot is back !!! ❤️ 


Todo next (when this is merged, I'd say, what do you think?): add the warnings.log with the built PR deployed, and write some clever parser to make the output readable for the authors, starting by cleaning the non-package warnings:

```python
#!/opt/miniconda3/envs/py38obs/bin/python
import sys

lines = open(sys.argv[-1], 'r').read().splitlines()

for line in lines:
    if line.count("docstring of textwrap") or line.count("docstring of collections") or line.count("packages/obspy.io.nied.knet' that doesn't have a title"):
        continue
    print(line)

```

trichter:
I think everything should be working now. Lets see if everything turns green.

trichter:
> Todo next (when this is merged, I'd say, what do you think?): add the warnings.log with the built PR deployed, and write some clever parser to make the output readable for the authors, starting by cleaning the non-package warnings

Nice to have, but we can see all doc build warnings in github actions, cant we?

ThomasLecocq:
> > Todo next (when this is merged, I'd say, what do you think?): add the warnings.log with the built PR deployed, and write some clever parser to make the output readable for the authors, starting by cleaning the non-package warnings
> 
> Nice to have, but we can see all doc build warnings in github actions, cant we?

No, it's really difficult to separate them from the logs, and I'd love to avoid reaching 3250 warnings again in the future... :-)

trichter:
> No, it's really difficult to separate them from the logs, and I'd love to avoid reaching 3250 warnings again in the future... :-)

I see, but that is easy to have. Write all warnings to a file wit `-w`. You posted the command yourself somewhere. After that filter and print the warnings in a new step.

```
cat warnings.txt | grep obspy > obspy_warnings.txt
cat obspy_warnings.txt
echo `wc -l obspy_warnings.txt` warnings
```

Also, I think most warnings should also occur when running the tests and are reported to tests.obspy.org.

ThomasLecocq:
> > No, it's really difficult to separate them from the logs, and I'd love to avoid reaching 3250 warnings again in the future... :-)
> 
> I see, but that is easy to have. Write all warnings to a file with `-w`. You posted the command yourself somewhere. After that filter and print the warnings in a new step.
> 
> ```
> cat warnings.txt | grep obspy > obspy_warnings.txt
> cat obspy_warnings.txt
> echo `wc -l obspy_warnings.txt` warnings
> ```
> 
> Also, I think most warnings should also occur when running the tests and are reported to tests.obspy.org.

The warning.log contains a lot of lines that don't have "obspy" in them (tracebacks from failed plots etc). So the full file is OK for uploading to the server, it'll allow parsing it later. The warning log contains also sphinx-related errors (missing index, wrong :type: etc), that are not tested using regultar tests, so it's good to have when someone pushes new code/doc.

Side note: I've started testing ``darglint`` as a very agressive linter for docstrings, it's quite good actually :-) I'll give it a pass on all the code base over the next weeks.

Issue #2953, opened 2022-01-17
failed tests missing from tests.obspy.org


ThomasLecocq:
https://github.com/obspy/obspy/actions/runs/1709170752 has failed tests, but on https://tests.obspy.org/?pr=2662 they don't appear

trichter:
Strange, the test report is uploaded https://tests.obspy.org/117184/

But the tests list is completely empty in the reported JSON.

@d-chambers Do you have a clue what is going on?

Maybe a bug in pytest-json-report?

d-chambers:
> Maybe a bug in pytest-json-report?

No,

It looks like pytest didn't run anything because there was an error when collecting the tests: 

```python
/usr/share/miniconda3/envs/test3.9/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
core/tests/test_network.py:25: in <module>
    from obspy.imaging.maps import HAS_BASEMAP
E   ImportError: cannot import name 'HAS_BASEMAP' from 'obspy.imaging.maps' (/home/runner/work/obspy/obspy/obspy/imaging/maps.py)
```

If one of the test files is not importable the test collection will fail. Curious the report was still generated and uploaded though.

ThomasLecocq:
I knew I'd find a way to break your system :-)

d-chambers:
I never doubted you for a moment 😝 

I don't think we really need to make any changes per se. An empty test report means there is an error in importing one of the test files.

Or, I guess pytest might emit a different return code if the collection fails and we can raise a big error message to avoid uploading and empty report, but it probably ins't much informative than what it is already doing.

QuLogic:
> Or, I guess pytest might emit a different return code if the collection fails and we can raise a big error message to avoid uploading and empty report, but it probably ins't much informative than what it is already doing.

It might return a 5? https://docs.pytest.org/en/latest/reference/exit-codes.html

trichter:
> https://docs.pytest.org/en/latest/reference/exit-codes.html

Interesting, but according to the uploaded json, the exit code had been 2.
We could check for exit code > 1 and not upload any test report in this case.

d-chambers:
closed by #2955

PR #2955, opened 2022-01-18
raise error if pytest ret code > 1

d-chambers:
### What does this PR do?
Raises an Error if pytest fails to run when using `obspy-rutests`, thereby avoiding submitting an empty test report. 

See #2953

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/scripts/runtests.py
+++ b/obspy/scripts/runtests.py
@@ -45,7 +45,6 @@ line argument is also accepted.
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
-import os
 import re
 import sys
 from pathlib import Path
@@ -54,6 +53,7 @@ import pkg_resources
 import requests
 
 import obspy
+from obspy.core.util.misc import change_directory
 from obspy.core.util.requirements import PYTEST_REQUIRES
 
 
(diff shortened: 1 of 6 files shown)


obspy/scripts/runtests.py
```diff
@@ -109,10 +109,19 @@ def main():
     os.chdir(base_obspy_path)
     try:
         status = pytest.main(plugins=[plugin])
+    except Exception as e:
```
trichter, line 112: Catching the exception and reraising it. I think this has no effect.

obspy/scripts/runtests.py
```diff
@@ -109,10 +109,19 @@ def main():
...
+    else:
+        if int(status) > 1:
+            msg = (
+                'A failure occurred while running ObsPys test suite. Please '
+                'check the output carefully and try again.'
+                   )
+            raise ValueError(msg)
+        upload_json_report(report=report, data=plugin.report)
+        sys.exit(status)
     finally:
```
trichter, line 123: I'd move the finally block up again and the new code out of the try block.

obspy/scripts/runtests.py
```diff
@@ -109,10 +109,19 @@ def main():
...
         status = pytest.main(plugins=[plugin])
+    except Exception as e:
+        raise e
+    else:
+        if int(status) > 1:
+            msg = (
+                'A failure occurred while running ObsPys test suite. Please '
+                'check the output carefully and try again.'
+                   )
+            raise ValueError(msg)
```
trichter, line 120: Maybe just `sys.exit(msg)` and also reporting the pytest return code in msg?

obspy/scripts/runtests.py
d-chambers: It was just required to use an `else` clause

trichter: OK, I see, but here you are basically catching an exception inside pytest. I have not encountered such a case.

d-chambers: Ya, it was required syntactically to have an else clause. I will just refactor though so we don't need either.

d-chambers:
Ok, I cleaned this up a bit by implementing a CD context manager. Also decided not to emit a custom message, pytest does a good job telling users what is wrong so we will just return its exit code.

d-chambers:
pre-commit was complaining about some of the work done in previous PRs in docs so I cleaned that up too. 

@ThomasLecocq Do we need the cartopy import at the top of docs/conf.py?

trichter:
> Do we need the cartopy import at the top of docs/conf.py?

Maybe this can be checked when dealing with #2956.

trichter (approved):

PR #2965, opened 2022-01-23
Fix statements in macros

QuLogic:
Doing `#define foo(x) (y=x);` will break code like:
```
if (...)
    foo(0);
else
    foo(1);
```
because semicolon in the macro will be the end of the statement in the `if`, and the semicolon outside the macro will be an empty statement, meaning the `else` is no longer attached to the `if`.

Similarly, placing multiple statements into a macro will cause braceless `if` to break as only the first statement will be within the condition. This occurs a few times with `swapflag` in `parseutils.c`.

The incorrectly enclosed `MS_SWAPBTIME` is (correctly) warning on latest gcc as it is in a braceless `if`.

See also https://github.com/iris-edu/libmseed/pull/73

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/mseed/src/libmseed/libmseed.h
+++ b/obspy/io/mseed/src/libmseed/libmseed.h
@@ -589,22 +589,22 @@ typedef struct Selections_s {
  * pack byte orders */
 extern flag packheaderbyteorder;
 extern flag packdatabyteorder;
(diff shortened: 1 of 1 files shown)


QuLogic:
Replaced this with #2966.

Issue #2956, opened 2022-01-18
docs not showing correct version


trichter:
@ThomasLecocq 

The doc version you uploaded at msnoise.org is showing the correct version number, but this appears not to be the case for the doc configuration in master. E.g. https://github.com/obspy/obspy/actions/runs/1714061599

![obs_missing_version](https://user-images.githubusercontent.com/2046150/150008808-175ac387-f40f-4130-a3a6-c1fbec9dc1f7.png)

ThomasLecocq:
It's because `master` has no "version"

https://github.com/obspy/obspy/blob/8c838cc6a56cecea65542bd5236bf4337f88eea8/obspy/core/util/version.py#L153-L155

trichter:
Maybe it's because git is not installed in the docs container. Could we try to switch to the conda/mamba installation process? It's not so much overhead compared to the docs build time.

ThomasLecocq:
agree !

trichter:
Alos, test reports do not show the correct version string. It started failing around the time when we switch from travis to github.

```
#112264 | 1 | 0.0.0.dev+0.g45206c13c1 | #2740 | 1943 | 49 | gh-actions | 3.8.6 | Linux (64bit) | a year ago
#112263 | 1 | 0.0.0.dev+0.g6bb7c1bf35 | #2740 | 1943 | 49 | gh-actions | 3.7.8 | Linux (64bit) | a year ago
#112262 | 30 | 1.2.2.post0+32.g2f990f92c2.obspy.m… | - | 2188 | 58 | travis-ci | 3.8.6 | Darwin (64bit) | a year ago
#112261 | 29 | 1.2.2.post0+32.g2f990f92c2.obspy.m… | - | 2188 | 58 | travis-ci | 3.7.8 | Darwin (64bit) | a year ago
```

Locally, its working for me.

```
In [2]: obspy.__version__
Out[2]: '1.2.2.post0+538.g7039b83550'
```

trichter:
I think we can fix the first (docs) issue by installing git.

The second is probably an incarnation of https://github.com/obspy/obspy/commit/b437f312883bc1e4f8b0b2697c06e3604ee45f0c#commitcomment-2278698 mentioned in #482.

We probably just need to set `fetch-depth: 0` in the checkout actions.

d-chambers:
closed by #2969

PR #2969, opened 2022-01-25
get full version back

trichter:
Fix for #2956 

Try to get the correct obspy version for ci tests and the documentation.

### DIFF
--- a/.github/docs_conda_env.yml
+++ b/.github/docs_conda_env.yml
@@ -15,10 +15,10 @@ dependencies:
   - sphinx>=4.2.0
   - m2r2>=0.2.7 # m2r is not maintained anymore, this fork is  
   - sphinx_rtd_theme
-  - cython # required by cartopy
   - cartopy>=0.20
+  - pip
   - pip:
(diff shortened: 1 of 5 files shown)


.github/docs_conda_env.yml
```diff
@@ -0,0 +1,27 @@
...
+  # docs
+  - sphinx>=4.2.0
+  - sphinx_rtd_theme
+  - cartopy>=0.20
+  - pip
+  - pip:
+    - pybtex
+    - m2r2 >= 0.3.2  # m2r is not maintained anymore, this fork is, need for latest version which is not yet included in conda-forge
+  #- numpy # required by cartopy
+  #- matplotlib
```
d-chambers, line 23: Do we still need these comments?

d-chambers (approved):

d-chambers:
Looks good, it seems it is working now.

.github/docs_conda_env.yml
trichter: I removed the latter two in another PR.

PR #2970, opened 2022-01-25
add tests with installed cartopy etc and check conda env files used in tests

trichter:
I added 3 test runs - one for each OS - to the strategy matrix with cartopy and other soft dependencies installed. I also cleaned the env files a bit. The 4 different files - one for each python version are not needed anymore. Caching introduced in #2936 was not working properly. I think I fixed it. The "Update environment" step is now skipped when the cached files are used. I also tested the changes for the push event in the next-to-last commit. Finally, I added pytest and pytest-json-report version numbers to the test reports.


d-chambers (approved):
Looks good to me

ThomasLecocq (approved):
impressive how this CI test suite has changed over the last months :-) Great!

Issue #2971, opened 2022-01-26
Add warning when writing non-compatible stream metadata to MSEED


jawooll:
When writing stream objects to MSEED, the _network, station, location, channel_ code, each string can be up to [11 bytes long](https://github.com/obspy/obspy/blob/master/obspy/io/mseed/headers.py#L426-L446). With fields > 11 bytes raising an appropriate ValueError. 

Then reading the same stream back in, these fields are then cut to the first 2, 5, 2, and 3 characters for the _network, station, location, channel_ respectively. Whilst this point has been raised previously (#2588), should an appropriate warning should be logged to the user from obspy when they try to write metadata fields > required length?

As a bit of background, this is something we are encountering when attempting to write out probabilistic CFs as stream objects, with the label information encoded in the channel string [see here](https://github.com/seisbench/seisbench/issues/54).

Issue #2588, opened 2020-03-18
Strange behavior of '-' in string fields of stream header


jbtruffault:
Hi ! 

The header of my Stream object is not correctly stored when I write it to a Miniseed file. To be more specific, the fields "networks", "station", "location" and "channel" are changed when I try to introduce a '-'. 

```python
    stats = {'network': "ABC-XXX",
             'station': "XYZ-XXX",
             'location': 'DEF-XXX',
             'channel': "CP-XXX",
             'npts': 123,
             'sampling_rate': 1,
             'mseed': {'dataquality': 'D'},
             'starttime': UTCDateTime(0),
             'endtime': UTCDateTime(123)}

    data = np.array(np.random.rand(123))

    st = Stream([Trace(data=data, header=stats)])

    print(st[0].stats)

    st.write("E:\Cooperant\Documents\JB_2020\Projets/2020 - Conversion Fonyx to Mseed/poc.mseed", format="MSEED")

    print("----------------------------------------")
    st = None

    st = obspy.read("E:\Cooperant\Documents\JB_2020\Projets/2020 - Conversion Fonyx to Mseed/poc.mseed")

    print(st[0].stats)
```

The result of the previous code is here:
```
network: ABC-XXX
         station: XYZ-XXX
        location: DEF-XXX
         channel: CP-XXX
       starttime: 1970-01-01T00:00:00.000000Z
         endtime: 1970-01-01T00:02:02.000000Z
   sampling_rate: 1.0
           delta: 1.0
            npts: 123
           calib: 1.0
           mseed: AttribDict({'dataquality': 'D'})
----------------------------------------
         network: AB
         station: XYZ-X
        location: DE
         channel: CP-
       starttime: 1970-01-01T00:00:00.000000Z
         endtime: 1970-01-01T00:02:02.000000Z
   sampling_rate: 1.0
           delta: 1.0
            npts: 123
           calib: 1.0
         _format: MSEED
           mseed: AttribDict({'dataquality': 'D', 'number_of_records': 1, 'encoding': 'FLOAT64', 'byteorder': '>', 'record_length': 4096, 'filesize': 4096})
```

megies:
MiniSEED only stores a fixed amount of characters:
 - 2 for network
 - 5 for station
 - 2 for location
 - 3 for channel

Everything longer than that will just be cut off when writing to miniSEED.

jbtruffault:
I just realized that ... thank you !

PR #2972, opened 2022-01-27
Add warning for metadata truncation when writing to mseed

jawooll:
### What does this PR do?

Changes default behaviour when writing to mseed files to now log a warning if the provided metadata would be truncated during writing.

See issues/questions (#2971, #2588).

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -113,6 +113,7 @@ Wassermann, Joachim
 van Wijk, Kasper
 Williams, Mark C.
 Winkelman, Andrew
+Woollam, Jack
 Zaccarelli, Riccardo
 Zad, Seyed Kasra Hosseini
 Zhu, Lijun
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -997,6 +997,15 @@ class MST(object):
         self.mst.contents.numsamples = trace.stats.npts
         self.mst.contents.sampletype = sampletype.encode('ascii', 'strict')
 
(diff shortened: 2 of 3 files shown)


yetinam:
Hey Jack! If I remember correctly, this truncation issue affects a few formats that you listed here https://github.com/seisbench/seisbench/issues/54 . Even though mseed will be the most prominent one, I was wondering whether it would be reasonable to add warnings for all formats in one go.

jawooll:
Hey Jannes, Good point! This requires a more involved PR though. Let's first see if the obspy team are open to the idea of logging such warnings when writing streams, and I could then look into extending this if required.

d-chambers:
> Let's first see if the obspy team are open to the idea of logging such warnings

Totally open to it, please run with it.

jawooll:
Ok great thanks, I'll extend the PR to log warning on all appropriate formats.

megies:
This looks pretty good to me, just a few things/thoughts..
- please add a changelog one-liner
- please actually test that the appropriate warning is shown, currently it only checks if any warning appears which could be anything (we have a [helper context manager making this easy](https://github.com/obspy/obspy/search?q=catchandassertwarnings))
- currently this is in a really low-level spot, which is OK, I guess, but if you plan on extending it to other file formats, I'd just move the check to inside `Stream.write()`. That is the recommended high-level routine and I'd argue people going the discouraged low-level way are kind of on their own with detail issues like this. Having it in a high level place like that would make it easier to extend for more formats in the future, by just adding to some dictionary storing limitations per format code

PR #2978, opened 2022-02-07
filter warnings in spectral estimation test

trichter:
Fix for #2977.

Let's see if it is working.

PR #2980, opened 2022-02-08
fix more network tests

trichter:
I fixed some more of the network tests. I skipped this one clients/iris doctest which I did not get to pass. The same thing is tested in the regular test suite.

This can be merged as is. Someone else needs to figure out those earthworm stuff. I have no clue about it.


d-chambers:
Looks like the non-network tests are failing due to new warnings cropping up. Perhaps not a big deal, should be easy to fix.

d-chambers:
Looks like you are addressing the failures in #2978

obspy/clients/iris/tests/test_client.py
```diff
@@ -234,14 +234,15 @@ def test_resp(self):
         t2 = UTCDateTime("2008-001T00:00:00")
         result = client.resp("IU", "ANMO", "00", "BHZ", t1, t2)
         self.assertIn(b'B050F03     Station:     ANMO', result)
+        # Exception: No response data available
```
d-chambers, line 237: Maybe choose a different response so that we can still test this part of the code?

trichter:
@d-chambers Feel free to finish this one :)

trichter:
I implemented one more test in clients.iris module as suggested. I think this can be merged.

d-chambers (approved):

PR #2981, opened 2022-02-09
deal with warnings

trichter:
I tried to get a grip on the warnings.

Added two more runs with `-W error` flag to remind us of all the warnings. To ignore some warnings also in this case and to not fail when collecting tests due to arclink deprecation warnings I had to adapt conftest.
Moved most options to `pytest.ini` file.
(Again) restructured the workflow ymls a bit.
Tried to not import obspy in the runtests script to get rid of this warning (did not work):
```
taup/tests/test_taup_plotting.py ...............                         [ 99%]
/usr/share/miniconda3/envs/test/lib/python3.10/site-packages/coverage/inorout.py:535: CoverageWarning: Module obspy was previously imported, but not measured (module-not-measured)
  self.warn(msg, slug="module-not-measured")
taup/tests/test_velocity_model.py .                                      [100%]
```

In the end number of emitted warnings went down from ~560 to below 40 for default tests on linux.  Later we can deal with the rest.


trichter:
I think I eliminated all warnings in the default tests. So please review fast, before new warnings pop up :stuck_out_tongue_winking_eye:

obspy/core/event/resourceid.py
```diff
@@ -198,37 +198,9 @@ class ResourceIdentifier(object):
...
-    >>> res_id  # doctest: +ELLIPSIS
-    ResourceIdentifier(id="smi:obspy.org/pick/...")
-    >>> # If the given ID is already a valid QuakeML
-    >>> # ResourceIdentifier, nothing will happen.
-    >>> res_id = ResourceIdentifier('smi:test.org/subdir/id')
-    >>> res_id
-    ResourceIdentifier(id="smi:test.org/subdir/id")
-    >>> res_id.convert_id_to_quakeml_uri()
-    >>> res_id
-    ResourceIdentifier(id="smi:test.org/subdir/id")
```
d-chambers, line 231: What's the reason for deleting all these examples from the docstring? There is (potentially) some useful information here.

obspy/core/tests/test_stats.py
```diff
@@ -281,7 +282,7 @@ def test_casted_stats_nscl_writes_to_mseed(self):
         stats_items = set(Stats())
         new_stats = Stats()
         new_stats.__dict__.update({x: st[0].stats[x] for x in stats_items})
-        with warnings.catch_warnings(record=True):
+        with pytest.warns(UserWarning):
```
d-chambers, line 285: Any match matching here needed to ensure the correct warnings are being raised?

obspy/imaging/scripts/mopad.py
```diff
@@ -3554,7 +3554,7 @@ def _check_curve_in_curve(self):
         # correct for ONE special case: double couple with its
         # eigensystem = NED basis system:
         testarray = [1., 0, 0, 0, 1, 0, 0, 0, 1]
-        if np.prod(self.MT._rotation_matrix.A1 == testarray) and \
+        if np.prod(np.ravel(self.MT._rotation_matrix) == testarray) and \
```
d-chambers, line 3557: can this be replaced by `np.allclose(self.Mt._rotation_matrix, test_array) and self.MT._eigenvalues[1] == 0` ? I find the use of `prod` to mean `all` a bit opaque.

obspy/io/kinemetrics/evt_base.py
```diff
@@ -131,7 +131,7 @@ def _time(self, blocktime, param, val, offset):
             frame_milli = 0
         frame_time += 315532800  # diff between 1970/1/1 and 1980/1/1
         time = UTCDateTime(frame_time) + frame_milli / 1000.0
-        return UTCDateTime(ns=time.ns, precision=3)
```
d-chambers, line 134: Why remove the specified precision? If the data format only stores miliseconds wouldn't we want this reflected in the ObsPy objects?

obspy/io/segy/tests/test_core.py
```diff
@@ -350,7 +351,8 @@ def test_writing_using_core(self):
                 with NamedTemporaryFile() as tf2:
                     out_file2 = tf2.name
                     # Write twice and catch header warnings
-                    with warnings.catch_warnings(record=True):
+                    with warnings.catch_warnings():
```
d-chambers, line 354: Can this be replaced with `pytest.warns` or `obspy.core.utils.testing.WarningCapture`?

obspy/io/segy/tests/test_segy.py
```diff
@@ -310,7 +310,8 @@ def test_read_and_write_textual_file_header(self):
                 self.assertEqual(segy.textual_header_encoding, header_enc)
             # The header writes to a file like object.
             new_header = io.BytesIO()
-            with warnings.catch_warnings(record=True):
+            with warnings.catch_warnings():
```
d-chambers, line 313: Same here

obspy/io/segy/tests/test_segy.py
```diff
@@ -358,7 +359,8 @@ def test_read_and_write_segy(self, headonly=False):
             segy_file = _read_segy(file, headonly=headonly)
             with NamedTemporaryFile() as tf:
                 out_file = tf.name
-                with warnings.catch_warnings(record=True):
+                with warnings.catch_warnings():
```
d-chambers, line 362: And here

obspy/scripts/runtests.py
```diff
@@ -45,24 +45,39 @@
...
+        os.chdir(origin)
+
+
 def _ensure_tests_requirements_installed():
     """
     Ensure all the tests requirements are installed or raise exception.
 
     This function is intended to help less experienced users run the tests.
     """
+    import pkg_resources
```
d-chambers, line 80: Did you move this import to speed obspy's overall import speed? I would think obspy.scripts.runtest should be imported unless the test suite is called.

obspy/scripts/runtests.py
```diff
@@ -45,24 +45,39 @@
...
-import obspy
-from obspy.core.util.misc import change_directory
-
 
 # URL to upload json report
 REPORT_URL = "tests.obspy.org"
 
 
+@contextlib.contextmanager
+def change_directory(path):
```
d-chambers, line 59: Why do you prefer this function here? It may be generally useful to other functions/test suites then we will have to move it back.

obspy/scripts/runtests.py
```diff
@@ -81,27 +96,28 @@ def main():
...
         sys.exit(0)
-    elif all(['--json-report-file' not in arg for arg in sys.argv]):
-        sys.argv.append('--json-report-file=none')
-    # Use default traceback for nicer report display.
-    sys.argv.append("--tb=native")
+    try:
+        import pytest
+        from pytest_jsonreport.plugin import JSONReport
+    except ImportError:
+        _ensure_tests_requirements_installed()
```
d-chambers, line 106: This kinda works for now, but it effectively short-circuits checks for any other test requirements we may want to use in the future. Are you doing this to speed up test execution?

obspy/conftest.py
```diff
@@ -103,12 +106,14 @@ def pytest_addoption(parser):
 def pytest_collection_modifyitems(config, items):
     """ Preprocessor for collected tests. """
     network_module_names = set(NETWORK_MODULES)
-
     for item in items:
+        # explicitely add filter warnings to markers so that they have a higher
```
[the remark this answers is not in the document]
d-chambers: Ok, I think I see what is going on. This is for specific markers which specify in the code certain warnings and filter behavior correct?

d-chambers (changes requested):
Great effort! It will be nice to see the warnings reduced (but I don't know how long 0 can last :wink:) 

My suggestion is to add some comments in the workflow files to help with maintainability as they are getting a bit complicated. 

I also had some questions regarding specific changes as outlined in the review.

obspy/conftest.py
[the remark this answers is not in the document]
trichter: I think it is sensible to filter out some specific expected warnings when consequently checking for warnings even when using `-W error`. They could be defined here in the first place, but I found it clearer to define them in `pytest.ini` and add them here with higher priority than command line flags again.

Another option to reach the same goal is to append `-W ignore:...` flags in runtests, but I found it much clearer this way.

obspy/conftest.py
```diff
@@ -145,15 +144,8 @@ def pytest_configure(config):
...
-    # but the existence of that file messed with the config.py hooks for
-    # some reason.
-    config.addinivalue_line(
-        "markers", "network: Test requires network resources (internet)."
-    )
-    config.addinivalue_line(
-        "markers", "image: Test produces a matplotlib image."
-    )
+    # do not collect tests in arclink module
+    pytest.importorskip('obspy.clients.arclink')
```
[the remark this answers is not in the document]
trichter: The comment is wrong. It should be `# ignore possible import error due to DepreciationWarning when using -W error flag`. arclink is the only module which raises a Warning upon import. The other modules only warn when a function is called. Adding the other deprecated modules here, would have no effect, because this command only skips the imports when an error is encountered.

obspy/core/event/resourceid.py
trichter: The reason is that `convert_id_to_quakeml_uri()` is deprecated and raises this warning:


Converts the current ID to a valid QuakeML URI. This method is deprecated, use :meth:`get_quakeml_id` instead. Only an invalid QuakeML ResourceIdentifier string it will be converted to a valid one. Otherwise nothing will happen but after calling this method the user can be sure that the ID is a valid QuakeML URI. The resulting ID will be of the form smi:authority_id/prefix/resource_id :type authority_id: str, optional :param authority_id: The base url of the resulting string. Defaults to ``"local"``.

obspy/core/tests/test_stats.py
trichter: Could be done, but not really necessary, because only a small amount of code is tested. The goal of the code before was to simply ignore the warnings (not test for), but it was not sufficient, because filterwarnings('ignore') was missing.

obspy/imaging/scripts/mopad.py
trichter: Unsure about this one. Maybe `np.all` or `all` is equivalent. I was just translating `matrix.A1` to `ravel(array)`.

obspy/io/kinemetrics/evt_base.py
trichter: Yep, you are correct. On the other side, a UTCDateTime object with non-default precision might become useless if you try to compare it to other UTCDateTime objects:
```
ObsPyDeprecationWarning: Comparing UTCDateTime objects of different precision is not defined will raise an Exception in a future version of obspy
```

I am OK with changing it back and filter the warning in the tests instead. But, this is the only place in ObsPy in which a different precision than the default is used.

obspy/io/segy/tests/test_core.py
trichter: This is an interesting thing. `pytest.warns()` is advertised in pytest documentation, but raises an error for my local pytest version (6.3) (missing expected warning category). `pytest.warns(None)` is working for pytest 6.3, but is deprecated since pytest 7.0. `pytest.warns(UserWarning)` and `pytest.warns(SpecialSEGYWarning)` (dont remeber the correct warning name) are not catching the warning, although SpecialSEGYWarning is a child of UserWarning. I think its a bug in pytest. Same below. Let's just leave it as it is now.

trichter:
I'd appreciate it. One idea I had, when reading your comments: To make the different flags easier to understand, we could just use a single flag ala `flag: warnings`, `flag: mindeps`.

d-chambers:
Hey @trichter I added a few comments, used the `WarningCapture` to cleanup some warning handling, and caught a catropy bug. I think things are mostly good to go but I am still not sure about the changes to the `_ensure_tests_requirements_installed()`.

trichter:
@d-chambers Thanks for your pushes.

After spending so much time on github action definitions, I invested some more time to merge the two different test workflows (pr and master) and the two different jobs (default and network tests) into one job and workflow. On the pro side, we have to maintain only a single place with conda setup and obspy installation, e.t.c. On the contra side, the matrix definition got a bit more complicated, but I tried to add comments for clarification.

I also changed the matrix, so that installing most dependencies including cartopy is now the default and installing only minimal dependencies (mindep) is tested for the lowest python version only.

> but I am still not sure about the changes to the _ensure_tests_requirements_installed().

I tried running tests with only pytest and pytest-json-report and it is working fine, e.g. https://tests.obspy.org/119544/. Some tests are skipped, similar as if any of the other soft dependencies was not installed. IMHO we should not force exit from runtests in this case. So in the end I am for removing the `_ensure_tests_requirements_installed()` again and regard the test requirements defined in setup.py as recommendations. I would even remove `pytest-cov` from there, because it is only needed for development and ci.

trichter:
For the case that we keep `_ensure_tests_requirements_installed()`, we need to make it more robust. I installed the obspy tag from this branch from test.pypi.org and encountered the following error:

```
(seis) eule@sauron:/RAID6/public$ obspy-runtests
  File "/RAID6/public/anaconda/envs/seis/bin/obspy-runtests", line 8, in <module>
    sys.exit(main())
  File "/RAID6/public/anaconda/envs/seis/lib/python3.10/site-packages/obspy/scripts/runtests.py", line 84, in main
    _ensure_tests_requirements_installed()
  File "/RAID6/public/anaconda/envs/seis/lib/python3.10/site-packages/obspy/scripts/runtests.py", line 72, in _ensure_tests_requirements_installed
    pkg_resources.get_distribution(package_req).version
  File "/RAID6/public/anaconda/envs/seis/lib/python3.10/site-packages/pkg_resources/__init__.py", line 477, in get_distribution
    dist = get_provider(dist)
  File "/RAID6/public/anaconda/envs/seis/lib/python3.10/site-packages/pkg_resources/__init__.py", line 353, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
```

setup.py
```diff
@@ -91,20 +91,24 @@
...
-    'arclink': ['cryptography'],
+    'geo': ['geographiclib'],
+    'imaging': ['cartopy'],
     'io.shapefile': ['pyshp'],
 }
+EXTRAS_REQUIRES['all'] = [dep for depl in EXTRAS_REQUIRES.values()
+                          for dep in depl]
+# arclink decryption also works with: pycrypto, m2crypto, pycryptodome
+# We do not include it in 'all', because the module is deprecated.
+EXTRAS_REQUIRES['arclink'] = ['cryptography']
```
megies, line 110: If arclink is causing any trouble or work, could also just delete it for good, it's been deprecated for years and my guess is no servers are even up to talk to, main servers have been shut down years ago. If people *really* need it, they can always just install an old obspy version to use it.

obspy/conftest.py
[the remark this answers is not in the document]
d-chambers: Lets remember to remove this  in #2994

trichter:
> The one thing I don't like is that running pytest from obspy's base directory (same level as setup.py) doesn't automatically determine rootdir correctly (at least on my machine) so the pytest.ini doesn't get used.

Does simply creating a symlink to obspy/pytest.ini work?

PR #2998, opened 2022-02-28
Fix remaining network test fails in clients.earthworm

trichter:
Try to fix the remaining fails in network tests (earthworm). It's branched from #2981 and should be merged after this PR.

~~Some warnings about sockets in clients.fdsn module remain. One warning only occurs when initializing `Client('IRISPH5')`.~~

### DIFF
--- a/obspy/clients/earthworm/__init__.py
+++ b/obspy/clients/earthworm/__init__.py
@@ -24,8 +24,8 @@ Basic Usage
       'AKV',
       '--',
       'BHE',
-      2021-12-30T14:36:15.761000Z,
-      2022-02-28T14:35:57.199000Z)]
+      UTCDateTime(2021, 10, 30, 12, 2, 27, 473000),
+      UTCDateTime(2021, 12, 29, 12, 2, 16, 899000)]
     >>> t = response[0][4]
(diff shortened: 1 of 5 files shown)


obspy/clients/earthworm/__init__.py
```diff
@@ -18,26 +18,26 @@
...
     >>> print(response)  # doctest: +SKIP
     [('AV',
-      'ACH',
+      'AKV',
       '--',
       'BHE',
-      UTCDateTime(2020, 4, 30, 12, 2, 27, 473000),
-      UTCDateTime(2020, 6, 29, 12, 2, 16, 899000)]
+      2021-12-30T14:36:15.761000Z,
+      2022-02-28T14:35:57.199000Z)]
```
megies, line 28: this is skipped so it doesnt matter too much anyway, but if I do that `print()` I *do* get `UTCDateTime(...)` and not some plain string

megies (approved):
Looks good to me

obspy/clients/earthworm/__init__.py
trichter: Mee, too, in python and ipython. Maybe I used it inside spyder?

PR #2988, opened 2022-03-04
Response removal prevented unneccessarily for response list type response in some cases

megies:
In the case of Responses containing (or well, sometimes solely built of) a Response list type stage (Frequency-Amplitude-Phase tuples), we are forcefully raising an exception if an extrapolation would have to occur, i.e. the given response list stage does not span the full range of frequencies needed for response removal.

https://github.com/obspy/obspy/blob/38d86b1b61a9875350b9fde11fb8c9aa3410184c/obspy/core/inventory/response.py#L1387-L1394

```python
File .../lib/python3.9/site-packages/obspy/core/inventory/response.py:1385, in Response._call_eval_resp_for_frequencies(self, frequencies, output, start_stage, end_stage, hide_sensitivity_mismatch_warning)
   1379 if min_f < min_f_avail or max_f > max_f_avail:
   1380     msg = (
   1381         "Cannot calculate the response as it contains a "
   1382         "response list stage with frequencies only from "
   1383         "%.4f - %.4f Hz. You are requesting a response from "
   1384         "%.4f - %.4f Hz.")
-> 1385     raise ValueError(msg % (min_f_avail, max_f_avail, min_f,
   1386                             max_f))
   1388 amp = scipy.interpolate.InterpolatedUnivariateSpline(
   1389     f, amp, k=3)(frequencies)
   1390 phase = scipy.interpolate.InterpolatedUnivariateSpline(
   1391     f, phase, k=3)(frequencies)

ValueError: Cannot calculate the response as it contains a response list stage with frequencies only from 0.1000 - 80.4000 Hz. You are requesting a response from 0.0038 - 125.0000 Hz.
```

When this was implemented this was probably chosen just as the easiest and safest route possible, however it prevents response removal in some very valid scenarios. Even if the response list stage does not cover all frequencies technically needed, it may well cover the frequency range of interest (a lot of times a bandpass is applied later on anyway) and in any case it is very easy to ensure there are no problems by using `pre_filt`  (or `water_level`) to suppress out-of-range frequencies.

I think we should change this exception into a warning. If we want, we could even make the warning conditional and check in the high level routines (probably just need it in `Trace.remove_response`) if `water_level`/`pre_filt` seem sufficient to prevent problems.

This is a (simplified) example response to illustrate the issue, green box with red borders is the frequency range defined in the response list stage:

resp_list

When interested in e.g. 1-20 Hz, there really isn't a problem why the response shouldn't be usable, a simple `pre_filt=(0.1, 0.3, 50, 80)` can make sure there are no problems in the deconvolution.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -39,6 +39,12 @@ Changes:
      numpy 1.22 (see #2912, #2913)
    * fix iso8601 regex for issue #2868 to cope with day 360 properly
    * fix crash when resampling very short traces (see #2926)
+   * response list stages: do not raise an exception if response calculation
+     involves extrapolation outside of frequency range defined by the response
+     list stage but rather only show a warning. raising an exception is the
+     safe route but it also prevents valid calculations as it is up to the user
+     to make sure that signal spectrum is properly suppressed in those
+     frequency ranges outside of the valid response information (see #2988)
  - obspy.clients.fdsn:
    * introduce fine-grained FDSN client exceptions (see #2653, #2964)
    * support for "eventtype" parameter in get_events(), as specified in version
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -1386,12 +1386,16 @@ class Response(ComparingObject):
 
                 if min_f < min_f_avail or max_f > max_f_avail:
                     msg = (
-                        "Cannot calculate the response as it contains a "
+                        "The response contains a "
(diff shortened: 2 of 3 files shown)


megies:
Changed raising an exception into showing a warning.
I'd like to have this in 1.3.0, for absolutely selfish reasons ;-)

trichter (approved):

PR #3009, opened 2022-03-08
SDS client, optionally speed up latency check routine

megies:
### What does this PR do?

Make it possible to skip one check in the latency checker, that can be very slow on slow file systems and thus can be counter productive

Default is the same as before, so no change for existing codes

Latency check in the SDS client can be very slow on slow file systems (in this case NFS) and it turns out that some glob check was responsible.

### PR Checklist
- [ ] All tests still pass.


megies:
This is a small change and should be unproblematic and not hold up the release.

PR #3010, opened 2022-03-09
update conda env files for ci tests

trichter:
Now that the feedstock for pytest-json-report was updated it can be installed from conda-forge for the ci tests.


megies:
triggered a CI re-run, since now the new package is actually updated in the anaconda package list

megies (approved):

PR #3011, opened 2022-03-09
update of gcf reader to latest revision of format

paitor:
### What does this PR do?

Update of gcf reader to latest revision of format as described in [SWA-RFC-GCFR, version F - December 2021](https://www.guralp.com/documents/SWA-RFC-GCFR). Move from pure python code to C-base to speed up decoding

### Why was it initiated?  Any relevant Issues?

At present the gcf reader in obspy is not up to date (e.g. see issue #2291) and references to official format documentation is missing (issue #3003). The choice to move to C-base is primarily to speed up reading gcf data but also since I hace an updated C-library that is up to date and tested.

### PR Checklist
(to be completed once update is completed)
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] If the PR is making changes to documentation, docs pages can be built automatically.
      Just add the "build_docs" tag to this PR.
- [ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the "test_network" tag to this PR.
- [x] All tests still pass.
- [x] Any new features or fixed regressions are covered via new tests.
- [x] Any new or changed features are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [x] First time contributors have added your name to `CONTRIBUTORS.txt` .
- [ ] If the changes affect any plotting functions you have checked that the plots
      from all the CI builds look correct. Add the "upload_plots" tag so that plotting 
      outputs are attached as artifacts. 
- [ ] Add the "ready for review" tag when you are ready for the PR to be reviewed.

Issue #3003, opened 2022-03-02
GCF mentioning inofficial superseded format documents


megies:
Güralp staff contacted us that [official GCF file format specifications are available online](https://www.guralp.com/documents/SWA-RFC-GCFR) and that the document mentioned in our io.gcf front page (`io/gcf/__init__.py`) should not be used as official source of information.

Just a heads up for any of the devs using GCF, in case they are curious and want to have a look..

CC @rannof

ThomasLecocq:
I'm happy if Guralp wants to PR & provide support to make the code better

paitor:
If not I have a long time plan to update the obspy reader to the latest revision of the format, think I've mentioned this before and apologies that I haven't found time for it  yet. Sorry to say an obstacle for me here is to set up a proper dev environment that will not interfere with my production environment as I've never done this before.

A quick question here, how is it. Can C-extension code be added to obspy or do you require pure python code (think the mseed reader is C-extension, no)?

If C-extension can be added I already have a home grown c-code for reading/writing gcf data records as well as a python extension module. This has been extensively tested against our archive of 15 years of gcf data from (69 stations) albeit I have to admit that testing of read/write of data sampled at > 2000 Hz suffers from lack of access to test files (hence testing here have been done on files written and read by my code). Think updating obspy should then be less than a day (including adding writing capability) + the time it takes to set up a proper testing environment. Should I have to translate to pure python the time will grown by at least pi-squared. In any case I'd probably would need to ask some dumd question while setting up a dev environment.

paitor:
Should perhaps add that a C-extension would significantly speed up reading of gcf files (about a factor of 10 from my experience using obspy vs. my c-extension python module + wrapping into a stream object)

megies:
I don't think there's a reason why not to do the reading in C. But I might not be aware of potential problems, just cause not having much experience doing that in obspy. Sounds like a complete rewrite of the obspy module to me, which is fine too, we just should try to be as much backwards compatible as possible (I guess mostly in terms of how header fields are stored in the Trace objects).

Regarding dev environment.. are you on Linux or Mac? Should be trivial then, if Windows, Thomas can help.

paitor:
Good, do you know whom of you I could consult if I run into trouble updating with c-code (i.e. a person that preciously added C-extension modules to the package)?

A few more or less related questions:
Would it be of interest to add the gcf writer, or is only the gcf reader of interest?

[...]

Issue #2291, opened 2019-01-22
New GCF format options and digitizer types


mmcgo:
in libgcf.py, lines 93/94
```
    if sysid >> 31 & 0b1 > 0:
        sysid = (sysid << 6) >> 6
```
See [GCF format](http://www.guralp.com/apps/ok?doc=GCF_format) section 3.1.3 - there is another option to steal more bits from the SysID field. This is, in fact, used for Minimus and Affinity (see section 3.2).

Please also note that we have updated the above document to add support for some new sample rates that the Minimus can generate (see section 4.1.1).
Section 4.1.2 also updates the fractional timestamps, and allocates the 4th bit of “compression code” byte, as mentioned in #2290 , for use with 5000sps.

megies:
Please feel free to provide patches or Pull Requests to extend functionality, see https://github.com/obspy/obspy/blob/master/CONTRIBUTING.md

paitor:
Hi

Has anyone had a look at this yet? Else I've just updated my C-code to the latest (2018) revision of the gcf format so could have a go at it (seems like a good opportunity to finally make a reinvestment).

megies:
I believe this might be obsolete with the recent changes in GCF module

PR #3055, opened 2022-04-26
update of gcf reader to latest revision of format and gcf write support

paitor:
### What does this PR do?

Update of gcf reader to latest revision of format as described in [SWA-RFC-GCFR, version F - December 2021](https://www.guralp.com/documents/SWA-RFC-GCFR). Move from pure python code to C-base to speed up decoding. Add write functionality for GCF format

At present the gcf reader in obspy is not up to date (e.g. see issue #2291) and references to official format documentation is missing (issue #3003). The choice to move to C-base is primarily to speed up reading gcf data but also since I have an updated C-library that is up to date and tested.
supersedes #3011 

### PR Checklist
(to be completed once update is completed)
- [x] All tests still pass.


obspy/core/tests/test_waveform_plugins.py
```diff
@@ -53,13 +53,16 @@ def test_read_and_write(self):
...
         formats = _get_default_eps('obspy.plugin.waveform', 'writeFormat')
         for format in formats:
+            start = UTCDateTime(2009, 1, 13, 12, 1, 2, 999000)
             # XXX: skip SEGY and SU formats for now as they need some special
             # headers.
             if format in ['SEGY', 'SU', 'SEG2']:
                 continue
+            elif format in ['GCF']:
+               # XXX: GCF format does not support fractional start time for sampling rates <= 250 Hz
+               start = UTCDateTime(2009, 1, 13, 12, 1, 3)
```
megies, line 65: Any particular reason why GCF can not be tested with split seconds? We use a wonky number for a reason :)

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -256,13 +265,17 @@ def test_read_thread_safe(self):
...
+            start = UTCDateTime(2009, 1, 13, 12, 1, 2, 999000) 
             # XXX: skip SEGY and SU formats for now as they need some special
-            # headers.
-            if format in ['SEGY', 'SU', 'SEG2']:
+            # headers. Also skip GCF as format does not permitt fractional
+            # start time for sampling rates < 250
+            if format in ['SEGY', 'SU', 'SEG2','GCF']:
                 continue
+            elif format in ['GCF']:
+               # XXX: GCF format does not support fractional for sampling rates <= 250 Hz
```
megies, line 277: Oh ok, weird. Maybe we should just set a sampling rate in the whole test case then, maybe 300 Hz if that works

obspy/io/gcf/__init__.py
```diff
@@ -18,11 +19,83 @@
...
+:meth:`~obspy.core.stream.Stream.write` method of a
+:class:`~obspy.core.stream.Stream` or
+:class:`~obspy.core.trace.Trace` object.
+
+>>> st.write('GCF-filename.gcf', format='GCF') #doctest: +SKIP
+
+
+Several key word arguments are available to set required GCF specific header information and
+allowed missalignment of supported starttime of first data sample:
+``streamID``, ``systemID``, ``isLeap``, ``gain``, ``ttl``, ``digi``, ``sysType``, ``missalign``
```
megies, line 63: these kwargs will have to adhere to our naming scheme (snake case), so e.g. `streamID` will have to become `stream_id` or similar

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+   return is_gcf
 
-def _read_gcf(filename, headonly=False, **kwargs):  # @UnusedVariable
-    """
-    Reads a GCF file and returns a Stream object.
 
-    only GCF files containing data records are supported.
-
-    .. warning::
+def _read_gcf(filename, networkcode='', stationcode='', locationcode='', bandcode="H", instrumentcode="H", channel_prefix=None,
```
megies, line 186: in most instances we just have `network` where it's intuitive that a network code is meant and not like a Network object, so i'd vote for just `network`, `station`, etc (`bandcode` and `instrumentcode` are fine, it's just about those NSLC / SEED ID items)

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+   :param stationcode: station code to use, if not specified unitID in gcf header will be used (first 4
+   :   characters in streamID)
+   :type locationcode: str, optional
+   :param locationcode: location code to use
+   :type bandcode: str, optional
+   :param bandcode: 1-character band code to use, ignored if channel_prefix is input
+   :type instrumentcode: str, optional
+   :param instrumentcode: 1-character instrument code to use, ignored if channel_prefix is input
+   :type channel_prefix: str, optional
+   :param channel_prefix: 2-character channel prefix to use, if not input channel_prefix = bandcode+instrumentcode
```
megies, line 212: I would strongly recommend to not offer clashing kwargs. Do the data ever contain band and/or instrument code? If they might do in some cases and don't in others I think we should get rid of `channel_prefix` to allow partial overrides. If band and instrument code are never recoverable from the data, i would get rid of band+instrument code kwargs and just keep `channel_prefix`

obspy/io/gcf/core.py
megies: Also, you could keep `headonly` at second position. Not that important, since private routines can have breaking changes without notice (and its unlikely that people would use `headonly` as a positional argument rather than a kwarg), but kind of like a low hanging fruit to basically stay compatible with old call.

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+   :param blockmerge: if True merge blocks if aligned
+   :type headonly: bool, optional
+   :param headonly: if True only read block headers
+   :type cleanoverlap: bool, optional
+   :param cleanoverlap: ignored if blockmerge is False, if True, remove overlapping identical data prior to merging blocks, 
+       if False overlapping blocks will not be merged even if data in overlap is identical
+   :type errorret: bool, optional
+   :param errorret: if True block and header issues will be set in trace.stats.gcf.stat for each :class:`~obspy.core.trace.Trace`
+       object in returned :class:`~obspy.core.stream.Stream` object, else function will raise raise an IOError with an appropriate
+       error message
```
megies, line 223: in some readers we have `strict` for things like that I believe.. maybe we could use that wording? or maybe rename to `raise_on_error`?

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
-                break
-    st = Stream(traces=traces)
-    if headonly:
-        st = merge_gcf_stream(st)
-    else:
-        st.merge(-1)
-    return st
+   """
+   if not os.path.exists(filename):
+      raise IOError("file %s could not be located (erroneous path?)" % (filename))
```
megies, line 291: Since we are Python 3 only, for new additions f-string notation would be nice, but thats just cosmetics, so feel free to ignore =)

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+               elif obj.seg[i].err == 5:
+                  err_msg = "start time of first data sample in data block %d is negative" % (i+1)
+               elif obj.seg[i].err == 9:
+                  err_msg = "failure to decode header values of data block %d" % (i+1)
+               elif obj.seg[i].err == 10:
+                  err_msg = "failure to decode data block %d (last data != RIC)" % (i+1)
+               elif obj.seg[i].err == 11:
+                  err_msg = "1'st first difference != 0 in data block %d" % (i+1)
+               elif obj.seg[i].err == 21:
+                  err_msg = "failure to decode data block %d (last data != RIC, 1'st first difference != 0)" % (i+1)
```
megies, line 357: These should be grouped together at the top of the file in some mapping variable with error codes

```
# at top of file
ERROR_MESSAGES = {
    21: "failure to decode data block %d (last data != RIC, 1'st first difference != 0)",
    ...
    }

# inside code
err_msg = ERROR_MESSAGES.get(
    obj.seg[i].err,
    "unknown error code (%d) set for data block %d" % (obj.seg[i].err, i+1))
```

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+      ret = gcf_io.read_gcf(b_filename, obj, mode)
+      if ret == -1:
+         if os.path.isfile(filename):
+            raise IOError("cannot open file %s in read mode (missing read permissions for user?)" % (filename))
+         else:
+            raise IOError("%s is not a file" % (filename))
+      elif ret == 1:
+         raise IOError("file %s is not a GCF data file" % (filename))
+      elif ret:
+         raise IOError("failed to read file %s (unknown error code %d returned)" % (filename,ret))
```
megies, line 331: like above, would be nice to have f-string notation, but just cosmetics ;-)

```python
          raise IOError(f"failed to read file {filename} (unknown error code "
                        f"{ret:d} returned)")
```

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+         ###  proper exception messages
+
+         ### Check data type
+         if not isinstance(trace.data[0],np.int32):
+            data = []
+            for d in trace.data:
+                if d < -2147483648:
+                    d = -2147483648
+                elif d > 2147483647:
+                    d = 2147483647
```
megies, line 544: would be good to set magic numbers at the top as constants and reuse here, I think, similar to the C header file. also can be cleaned up a bit

something like.. (naming just a suggestion)
```
SAMPLE_MAX = 2147483647
SAMPLE_MIN = -2147483648

[...]

trace.data[trace.data < SAMPLE_MIN] = SAMPLE_MIN
...
```

or even better, use `np.ceil` and `np.floor`

obspy/io/gcf/core.py
```diff
@@ -33,68 +142,584 @@ def merge_gcf_stream(st):
...
+                data.append(d)
+            trace.data = np.array(data, dtype=np.int32)
+         trace.data = np.ascontiguousarray(trace.data)
+
+         ### Check sampling rate
+         sps_denom = 1
+         if not compatible_sps(trace.stats.sampling_rate):
+            sps = ".3f" % (trace.stats.sampling_rate) if trace.stats.sampling_rate < 1 else "%d" % (trace.stats.sampling_rate)
+            raise TypeError("trace sampling rate, %s, in trace %d not supported in GCF format" % (sps,i+1))
```
megies, line 553: ValueError probably fits a tiny bit better

obspy/io/gcf/__init__.py
```diff
@@ -18,11 +19,83 @@
...
+have precedenc. For default values if neither set by ``stats.gcf`` nor by use of appropriate 
+keyword arguments refer to :meth:`~obspy.io.gcf.core._read_gcf`
+
+..Note:: 
+The GCF format is only guaranteed to support 32-bit signed integer values. While data with values
+out of range may be properly stored in the GCF format (if first and last data sample can be represented
+as a 32-bit signed integer as well as all first difference values of the dat avector) the current 
+implementaion only permitts input data to be representable as a 32-bit signed integer. If input waveforms
+cannot be representable as 32-bit signed integers they will silently be clipped at -2,147,483,648 and 
+2,147,483,647
```
megies, line 77: It would actually be trivial to add a warning message at that part where you crop to max/min.

```
data = np.ceil(tr.data, ...)
data = np.floor(data, ...)
data = np.require(data, dtype=np.int32, requirements=['C_CONTIGUOUS'])
if not np.all(data == tr.data):
    msg = '...'
    warnings.warn(msg)
tr.data = data
```

obspy/core/tests/test_waveform_plugins.py
paitor: That's a format restriction, we probably have to ask Guralp why this is so, but if I'm to elaborate a bit: The gcf data files consists of 1024 byte blocks, each having a header of 16-byte header leaving 1008 byte for the data. The lowest compression (using signed 32-bit first difference) can then at maximum store 250 data points. So basically for any sampling rate < 250 Hz you can fit an integer number of whole seconds of data into each block. To keep it neat and clean I guess at the time designing the format specification they simply decided that all time-series should sample at integer seconds and for sampling rates < 250 Hz each block should hold whole seconds data segments. Note that even for sampling rates > 250 Hz the start-time is restricted. e.g. 500 Hz data may only start at whole or half seconds (i.e. 0.00s or 0.50s)

obspy/core/tests/test_waveform_plugins.py
paitor: Sorry, 300 Hz is not supported by GCF format (You can find the supported sampling rates as the keys in dict `_SPS_MAP` in `core.py`, but, even so the start time is restricted to specific values (basically at `n/fractional_time_denominator` where `fractional_time_denominator` can be found as the values in  dict `_SPS_MAP` in `core.py`)

obspy/io/gcf/__init__.py
paitor: Sure, will fix it

obspy/io/gcf/core.py
paitor: Sure will fix

obspy/core/tests/test_waveform_plugins.py
megies: Ah ok, fine then!

obspy/core/tests/test_waveform_plugins.py
megies: :+1:

obspy/io/gcf/core.py
megies: see https://github.com/obspy/obspy/pull/3055#discussion_r859630213

obspy/io/gcf/core.py
paitor: Well personally I'd like to automatically set the channel_code according to SEED specification but that would require knowing the instrument natural period which is not contained in the GCF header. If you prefer to stick with the `channel_prefix` I can fix this. Just wondering if its not better to use the standard names to eventually be able to phase out (deprecate) the non-standard `channel_prefix` argument?

obspy/io/gcf/core.py
paitor: Sorry I do not really follow, can you elaborate?

obspy/io/gcf/core.py
paitor: Well, f-strings appeared from python 3.6 (or are they backported to all 3.x versions)? In any case unless support for %-formatting are expected to be deprecated in future versions of python I'd prefer to keep them around :-þ

obspy/io/gcf/core.py
paitor: Sure can do. I think for most users this will be of little interest (and the codes and messages are not part of the standard but have grown from my own work trying to categorized the various issues I found in the 20+ years of GCF-archives we have at SNSN).

obspy/io/gcf/core.py
paitor: Same reply as above then ;þ  (but I'll think about if some use of f-strings may simplify cutting up strings to conform to Flake8).

obspy/io/gcf/core.py
paitor: Will do (one of the two)

obspy/io/gcf/core.py
paitor: Agree, will fix

obspy/io/gcf/__init__.py
paitor: Sure, will do

paitor:
So I've managed to get the code to compile and install on windows but it seems all tests fails meaning that something is broken on windows. I do not have easy access to a windows computer nor experience (must be over ten years now since I last used windows except occasional googling adventures to help the elders), hence need some guidance on how to progress from here. Are there perhaps some online virtual windows servers one could be use or something else?

paitor:
So I resorted to use github to try to debug the win build, question though. Is there a way to temporarily turn of some of the checks to speed things up (should for now be enough to build and run one of the win test suites)?

obspy/core/tests/test_waveform_plugins.py
```diff
@@ -256,13 +269,18 @@ def test_read_thread_safe(self):
...
         for format in formats:
+            start = UTCDateTime(2009, 1, 13, 12, 1, 2, 999000)
             # XXX: skip SEGY and SU formats for now as they need some special
-            # headers.
-            if format in ['SEGY', 'SU', 'SEG2']:
+            # headers. Also skip GCF as format does not permitt fractional
+            # start time for sampling rates < 250
+            if format in ['SEGY', 'SU', 'SEG2', 'GCF']:
                 continue
+            elif format in ['GCF']:
```
megies, line 280: This `elif` will not be reached, since GCF is added above in the block that just does a `continue`.

These checks for `GCF` (also a few more cases above) would better be rewritten as `elif format == 'GCF':`, I think, but thats just cosmetics

obspy/io/gcf/__init__.py
```diff
@@ -1,13 +1,14 @@
...
-obspy.io.gcf - Guralp Compressed Format read support for ObsPy
-==============================================================
-This module provides read support for GCF waveform data and header info.
-Most methods are based on info from Guralp site
-http://geophysics.eas.gatech.edu/GTEQ/Scream4.4/GCF_Specification.htm
+obspy.io.gcf - Guralp Compressed Format, GCF, read and write support for ObsPy
+==============================================================================
+This module provides read and write support for `GCF<https://www.guralp.com/apps/ok?doc=GCF_Intro>` 
+waveform data and header info according to GCF Reference `SWA-RFC-GCFR Issue F, December 2021
+<https://www.guralp.com/apps/ok?doc=GCF_format>`.
```
megies, line 7: These links need to be reformatted, see https://docs.obspy.org/pr/gcf-c-source/packages/obspy.io.gcf.html#module-obspy.io.gcf

First link might need a space after "GCF", not sure. In any case, both links need to be postfixed with an underscore, see https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

obspy/io/gcf/__init__.py
```diff
@@ -18,11 +19,83 @@
 The format will be determined automatically.
 
 >>> from obspy import read
->>> st = read("/path/to/20160603_1955n.gcf")
+>>> st = read("/path/to/20160603_1955n.gcf")  #doctest: +ELLIPSIS
```
megies, line 22: This "#doctest: +ELLIPSIS" doesn't do anything. "Ellipsis" modifier is for allowing "..." as placeholder when testing printed text output, but this command does not have any text output.

paitor:
Thanks, I'll have a look at it. Will probably have to put this aside for two weeks now though (upcoming meeting, NGWM, and some work to do prior to that so I have something to present...)

obspy/io/gcf/core.py
megies: You could consider adding a kwarg to pass in an `Inventory` to provide that information.. if you really wanna open that box. We could check how often band code and instrument code get used across the code base, but I think it might only be very few occurrences, so it might be more worthwhile to have one single kwarg instead of two for the job. But either way is fine for me. I just dont want three clashing kwargs =)

paitor: Mmmm, passing an `Inventory` seems like using an elephant gun to shot mosquito's. A quick scan of the master branch yields no further use of `instrumentcode` nor `bandcode`, but then again neither if `channel_prefix` so presumably then, in order not break backwards compatibility with any code actually using this I guess it's the best choice is to stick with `channel_prefix`, will fix it after NGWM next week.

obspy/core/tests/test_waveform_plugins.py
paitor: Yup, adjusted, the reason for `format in [...]` is mainly to pave the way for other formats with similar requirements in the future (e.g. the bc format I plan to add once I've figured out how to finish the current project)

obspy/io/gcf/__init__.py
paitor: Ok, fixed (pretty much at copy-paste-hammer level when it comes to the sphinx syntax, sorry)

obspy/io/gcf/__init__.py
paitor: Fixed (I think... Not sure how to build docs locally so will see after next push)

megies:
Just ping me whenever you are done @paitor

paitor:
@megies Sure will do, at least I've narrowed down the issue to that the windows build by some reason is only reading the first block in the gcf-files so presumably the return value from the windows C function _read differs from a fully posix compliant read function causing the code to think that it has read to the end of the file. At least this is my current working hypothesis. Sorry about using git hub to compile and test run the code but at present I do not have access to a windows computer to use

paitor:
The failing build for ubuntu-latest, py3.8 is a bit of surprise though, have not seen this before.

megies:
> The failing build for ubuntu-latest, py3.8 is a bit of surprise though, have not seen this before.

Don't worry about it, it's unrelated, conda failing to install some package. Hoping it's gonna go away by itself for now.

megies:
> Sorry about using git hub to compile and test run the code

Not a problem!

paitor:
@megies Finally got hold of a windows can and fixed the remaining issues. The remaining failing tests on windows are related to io.segy (Extended textual headers are not yet supported) so not related. Thus think this is ready for review now, thanks for the patience.

paitor:
Failing tests seems unrelated.

paitor:
@megies Guess you are quite busy but bumping this again as it would be nice to finish it of before the release of obspy 1.4. 

The remaining failing test seems unrelated (related to the segy format: `Extended textual headers are not yet supported. Please contact the developers.`)

megies:
@paitor exact commands depend on what the remotes are called in your local git clone.
Assuming the main repo is called "origin", you want to do:

```
git fetch origin
git checkout gcf-c-source
git rebase origin/master
git push -f
```

The last one depends a bit on your git config too, basically the command you push your branch with an additional `--force` to overwrite history (which can always be reverted when having the git hash of the former branch tip commit, which github states in a comment on every force push in a PR).

You might run into merge conflicts though, which you would have to solve manually and then continue the rebase. If anything goes wrong you can abort with `git rebase --abort` and I can do it instead.

paitor:
@megies Another newbie question here, how do I figure out the name of the main repo?

When I originally cloned a fork of obspy I ran

    $ git clone https://github.com/paitor/obspy.git
    Username: 
    Password: 
    $ cd obspy
    $ git checkout master
    Already on 'master'
    Your branch is up-to-date with 'origin/master'.

So i guess the name is `origin`?

I'll get to it tonight but may take me a few days to debug if the failing test persists on windows as I could not install obspy from source (some issues getting scipy properly installed) but only installed a striped down version containing the gcf reader. 

I suspect that the cause for the failing tests is the segy reader by some reason identifying the gcf file written in the test case as a segy file so potentially (if I can verify my suspicion) I will may need some assistance from someone knowledgable about the segy format (who to talk to?).

megies:
> I suspect that the cause for the failing tests is the segy reader by some reason identifying the gcf file written in the test case as a segy file so potentially

Yes, that's the only thing that makes sense to me too. But I saw you didnt add new GCF test data files, so it didn't make sense to me, when there are no changes to `_is_segy` or GCF test data files. But if you write a GCF temporary file in a test and run `read()` on it without the filetype specified, then indeed it could be `_is_segy` hitting that file in the file format detection process. In that case `_is_segy` would need changing to not give a false positive on your GCF file.

paitor:
Nopp, didn't add any new test files as I figured with the added write capability I could just write the files. Then again if there is are matched bugs in writing/reading these may be masked out...

paitor:
@megies Tracked down the failing test to trigger when the test function wrote a gcf file using a big endian data vector. Not entirely sure I understand why this occurs though as it seems to only trigger if the gcf writer is compiled using the MSVC compiler, compiling using MinGW on windows do not trigger, nor writing a gcf file using a small endian data vector so presumably it is related to differences in memory alignment (btw. why the test using a big endian data vector on a small endian OS, sure it can be done but is there a reasonable user scenario for this?). 

Now the segy header is a fairly large beast, hence the _is_segy will read in bits from the data sector when reading a gcf file. This coupled to the data vector used when writing the file basically being a straight line with a constant first difference of one (the gcf format stores first differences in BIG-endian using either 8, 16, or 32 bits/sample). Upon converting the read in bits for the case of BIG-endian on  small endian machine and code compiled using MSVC all converted supposed segy header parameters become 1 which is in permitted range. Ergo my workaround for now is mainly to mask this out by adjusting the data vector used in test_read_and_write to hold random number (theoretically this does not exclude the issue to trigger again as the probability of such a vector to not result in valid numbers in the valid ranges of the decoded supposed segy header variables is small but not zero).

At the end of the day I think that (if possible) the _is_segy function could be improved (e.g. including a check that the size of the file matches the expected size given the number of samples and bits used for each sample), however not being familiar with the specifications of the format I'm quite hesitant to put my fingers in that can.

BTW. I note that in the test_read_and_write method only header values are compared and not the actual data vector. Is this deliberate (guess the use of big-endian on small-endian architectures makes this a bit complicated) or perhaps something to add to the test?

megies:
> Sorry, I must be missing something here, here's what I did:

No worries. The pitfall here is that you only have your fork as a remote. The master branch on your fork is not automatically updated to match the main repo master branch, it's just a snapshot from the time when you made your fork. You could in principle keep the master branch on your fork up to date, but it isn't needed really. What I recommend in general is just have your local master branch not track your fork's master branch (which it does right now) but have it track the main repo one.
So you want to do something like:

```
git remote add upstream https://github.com/obspy/obspy.git
git fetch upstream
```

To get info on the main repo.

Optionally to have your local master branch track the main repo master and update it:

```
git checkout master
git branch --set-upstream-to=upstream/master
git pull 
```

And then like above

```
git rebase upstream/master
```

Also if you do lots of pulling and pushing, I recommend setting up ssh keys to not have to enter credentials all the time: https://docs.github.com/en/authentication/connecting-to-github-with-ssh

paitor:
Great, thanks for the input. Probably should spend some more time reading up on git and github before I go about adding a bc reader/writer (as well as start using it as a repository for my own stuff)

paitor:
@megies 

Think I'm done as far as I can. I note that my post while pushing the last edits seem to have vgone with the dodo bird, so in summary:

Short story, the _is-segy() checks needs to be improved by someone more knowledgable (on the seg-y) than me, for now I've updated the test suite to not trigger but potentially non seg-y files may be recognised as seg-y files with current implementation. 

Long version The issue (as far as I understand it) is four-fold, (i)the large header associated with the seg-y file, (i) the fairly weak test to if the file may be a seg-y file and (iii) the test file used containing a straight line with slope 1, (iv) writing big-endian data on a small-endian architecture + using MSVC compiler
(i) results in the _is_segy() reading in the data block in the gcf file 
(ii) means only checking a few selected values for which most (all?) 1 is a valid value
(iii) gcf format stores first difference
(iv) I have not analysed the files at bit level nor looked into how executables produced by MSVC align memory, but as the issue only triggers if the executable is compiled using MSVC and the testcase use pushes big-endian data to file on a small endian architecture (testcase do not fail on linux, mac, nor on windows if MINGW is used - BTW whats the user case for this scenario?).

Knowing the above I changed the data in the test file used to be random data instead of a straight line and, viola, the test pass on all systems. This is of'course a weak work-around and presumably a better woprk around would be to update the _is_segy() function. On suggestion (if possible), estimate the file size based on header size, number of data samples and bits used for each sample (not familiar with the seg-y format so can't tell if this would be doable without parsing the full file, e.g. the bc-format which I plan to add reader/writer for once the gcf update is done use an adaptive bit-compression algorithm which requires de-compresing the full data vector and I guess one would like to avoid this in a quick file-format check).

paitor:
@megies Is there something I've missed here as this has been hanging for almost two months now?

ThomasLecocq:
"summer", probably

paitor:
@ThomasLecocq  Jaaah, could be, but summer is long gone now (already had frost near Uppsala) so would be nice to finish this, no?

megies:
Yea, trying to get to this, sorry

paitor:
Great, thanks

obspy/io/gcf/core.py
```diff
@@ -41,60 +175,673 @@ def _is_gcf(filename):
...
     :type headonly: bool, optional
-    :param headonly: If True read only head of GCF file.
+    :param headonly: if True only read block headers
+    :type network: str, optional
+    :param network: network code to use
+    :type station: str, optional
+    :param station: station code to use, if not specified unitID in gcf header
+       will be used (first 4 characters in stream_id)
+    :type locationcode: str, optional
+    :param locationcode: location code to use
```
megies, line 245: I'll rename this `location` to be more in line with rest of our code base

megies: Also, I kinda want to get rid of having a default channel_prefix for users. They should be responsible for using a sensible channel prefix themselves, otherwise people can end up with a misleading one set by our default

obspy/io/gcf/core.py
megies: done

megies:
There will be a review of build docs needed

Screenshot from 2022-10-18 18-07-12

obspy/io/gcf/core.py
```diff
@@ -41,60 +175,673 @@ def _is_gcf(filename):
...
+    if not os.path.exists(filename):
+        raise IOError("file %s could not be located (erroneous path?)"
+                      % (filename))
+
+    stream = None
+    # Load shared library
+    try:
+        gcf_io = _load_cdll("gcf")
+    except Exception as e:
+        raise IOError(str(e))
```
megies, line 353: I'm not sure I like overriding this exception.. what's the point? It might mask the real problem with a misleading error type

obspy/io/gcf/core.py
```diff
@@ -41,60 +175,673 @@ def _is_gcf(filename):
...
+    else:
```
megies, line 354: why have the rest of the reading in this super long "else" block? could just dedent all the rest or am i missing something?

obspy/io/gcf/core.py
paitor: Totally agree, however the existing gcf reader in obspy do use `HH` as default for `channel_prefix` ergo the failing test that are all due removing this. I kept it in order to not mess up code using obspy and relying on this.

obspy/io/gcf/core.py
paitor: Yup, but no longer a concern due to removing the `try` clausule

megies:
@paitor i'm sorry about force-pushing the rebase without asking, in any case those commits you added afterwards are messed up, I'm guessing you "git pull"ed in the changes, when they have a completely different history than your local branch was left with?

paitor:
@megies No need for you to be sorry, if anyone should be sorry its me for not knowing git good enough. 

Yes, I naively thought a simple pull would do but ended up with a bunch of conflicts and trying to resolve these probably messed up things. Apologies for this, how to best resolve (must be possible to roll back? and how to best do so with my local version?)

megies:
If you don't mind, I'll work on this a bit more but then it might be best if you save your local changes (if you have any) with e.g. "git stash" but then reset the branch to the online version

paitor:
No minding at all, just assumed that you didn't want to do the changes your'self as you probably have enough on your hands anyway. Feel free to change whatever and let me know if/when you want my fingers in there  again.

paitor:
No local changes (well except my futile attempts to resolve the mess I pulled...)

obspy/io/gcf/core.py
megies: fair enough, I did not check that, I guess we can leave the default then :+1:

obspy/io/gcf/__init__.py
megies: fixed. also now not showing a clipping warning but a correct warning in case of no clipping but losing decimal info when downcasting floats

obspy/io/gcf/core.py
```diff
@@ -41,60 +184,650 @@ def _is_gcf(filename):
...
+            mode = 2
+        elif not cleanoverlap:
+            mode = 1
+        else:
+            mode = 0
+
+    # read file
+    obj = _GcfFile()
+    b_filename = filename.encode('utf-8')
+    ret = gcf_io.read_gcf(b_filename, obj, mode)
```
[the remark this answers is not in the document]
paitor: Well, one could make a deep copy of `obj` directly after the read call and then free the memory. Currently the free call occurs a soon as all info in `obj` have been re-packed into a stream object.

I guess things can break but at present I can't see what (e.g. attributes used in denominators are all guaranteed to not be 0 and coming from C type issues should be present)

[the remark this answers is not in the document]
megies: Nevermind, looks like that doesnt work

megies:
@paitor I'm done with going over the code, feel free to look at my changes, they are all grouped with clean commit messages (since otherwise it is near impossible to judge changes when indentation gets changed). Let me know if anything should not be done like I changed it.

What will need thorough inspection is the documentation pages:
https://docs.obspy.org/pr/gcf-c-source/packages/obspy.io.gcf.html

Because I noticed some long "rubric" paragraphs did not show up in the rendered docs, maybe rubric is not accepted by sphinx in function docstrings.. no idea, anyway please check those rendered docs

I'd happily encourage people that actually use GCF to give this a look, otherwise, I'm OK with merging this, I guess

paitor:
Had a quick look and all looks good to me. Also in favour of encouraging gcf users to cast an eye if you know of them.

paitor:
And, thanks for the effort put in sofar.

megies (approved):
Would be good to have GCF users look at this, otherwise will be merged soon

megies:
No other feedback, so merging. Sorry this took so long @paitor, it's just the larger the scope of a change the longer it usually takes, and I'm tending to be quite cautious when merging things that affect parts of the code I don't use myself.. and well nobody actually using GCF offering feedback.. :man_shrugging:

paitor:
Great thanks. Quick question due to my inexperience with github, is it recommended that I now delete this branch and/or this fork (I see a message here that the former can "safely" be done, whilst the later can "also" be done, but do not really know if this has implications, e.g. cleaner structure of the project)

megies:
Deleting the branch basically just means the branch will not show up in the dropdown list of branches anymore, it's just deleting the branch name label stuck to the last commit on the branch. No real information is lost by "deleting the branch", usually you want to do it to avoid the number of branches to explode, like in obspy we would probably have a list of 3000 branches if we didn't delete em after merging PRs.

You can also delete the fork if you dont need or want it anymore, it has no implications on the main repo, all history of this PR will be kept. Although there probably isn't much reason to delete the fork.

Issue #2682, opened 2020-07-14
TauP issue for some depths using custom models


aceciliad:
I want to get the travel times for some simple custom models of Mars. But for some of them i get the error:

_Traceback (most recent call last):
  File "/home/admin/.local/lib/python3.6/site-packages/obspy/taup/tau_model.py", line 228, in load_from_depth_cache
    value = self._depth_cache.pop(depth)
KeyError: 60.0_

And then the message: 

_"TauBranch depths not compatible with slowness sampling."
obspy.taup.helper_classes.TauModelError: TauBranch depths not compatible with slowness sampling._

It works fine with Taup but not with obspy.taup, and this error only appears when considering sources between 60-80km depth.

One of the models is this one (i also tried the same model but with solid outer-core and still getting same error): 

```
     0.000   2.8622   1.6541   1.7004
     1.000   2.8622   1.6541   1.7004
     1.000   5.3851   3.1126   2.3155
    10.000   5.3851   3.1126   2.3155
    10.000   5.3847   3.1126   2.4316
    39.174   5.3847   3.1126   2.4316
    39.174   5.3850   3.1126   2.4330
    80.000   5.3850   3.1126   2.4330
mantle
    80.000   8.4775   4.7507   3.7300
   200.000   8.4775   4.7507   3.7300
   200.000   8.9174   4.9508   3.9233
   400.000   8.9174   4.9508   3.9233
   400.000   8.9116   4.8169   3.9625
   900.000   8.9116   4.8169   3.9625
   900.000  10.6193   5.6744   4.5620
  1550.000  10.6193   5.6744   4.5620
outer-core
  1550.000   5.8950   0.0000   8.9500
  3360.000   5.8950   0.0000   8.9500
inner-core
  3360.000  10.6193   5.6744   4.5620
  3389.500  10.6193   5.6744   4.5620
```

An example code is:

````
from obspy.taup import taup_create, TauPyModel
taup_create.build_taup_model('model_1.nd', output_folder='output')
model_Tau = TauPyModel('output/model_1.npz')
arr_p   = model_Tau.get_travel_times(source_depth_in_km=60.0, distance_in_degree=30.0, phase_list=['P'])
````

Using Obspy 1.2.2 and Python 3.6.9

sstaehler:
any news on this? Or would a taup developer have an idea on how to debug this?

megies:
@sstaehler if you can reproduce this with the original Java TauP, you might be able to have @crotwell have a look

[...]

Issue #1938, opened 2017-10-16
taup issue for sources at depth


lisesismo:
I want to put sources at (large) depths. It worked with the `TauP` package (or `ttimes`) but I get an error message with `obspy.taup` when it gets too deep.

Example:

```python
>>> from obspy.taup import TauPyModel
>>> model = TauPyModel(model="prem")
>>> arrivals = model.get_travel_times(source_depth_in_km=1940,distance_in_degree=32)

Error (long version below):
TauModelError: TauBranch depths not compatible with slowness sampling.

---------------------------------------------------------------------------
TauModelError                             Traceback (most recent call last)
<ipython-input-3-afcb0cbbbd02> in <module>()
----> 1 arrivals = model.get_travel_times(source_depth_in_km=1940,distance_in_degree=32)

/Users/liseretailleau/anaconda/lib/python2.7/site-packages/obspy/taup/tau.pyc in get_travel_times(self, source_depth_in_km, distance_in_degree, phase_list, receiver_depth_in_km)
    638         tt = TauPTime(self.model, phase_list, source_depth_in_km,
    639                       distance_in_degree, receiver_depth_in_km)
--> 640         tt.run()
    641         return Arrivals(sorted(tt.arrivals, key=lambda x: x.time),
    642                         model=self.model)

/Users/liseretailleau/anaconda/lib/python2.7/site-packages/obspy/taup/taup_time.pyc in run(self)
     36         arrival times will be in self.arrivals.
     37         """
---> 38         self.depth_correct(self.source_depth, self.receiver_depth)
     39         self.calculate(self.degrees)
     40 

/Users/liseretailleau/anaconda/lib/python2.7/site-packages/obspy/taup/taup_time.pyc in depth_correct(self, depth, receiver_depth)
     48         if self.depth_corrected_model is None or \
     49                 self.depth_corrected_model.source_depth != depth:
---> 50             self.depth_corrected_model = self.model.depth_correct(depth)
     51             self.arrivals = []
     52         if receiver_depth != depth:

/Users/liseretailleau/anaconda/lib/python2.7/site-packages/obspy/taup/tau_model.pyc in depth_correct(self, depth)
    219             raise TauModelError("Can't depth correct to a source deeper than "
    220                                 "the radius of the planet.")
--> 221         return self.load_from_depth_cache(depth)
    222 
    223     def load_from_depth_cache(self, depth):

[...]

PR #3062, opened 2022-05-04
Fix for "TauBranch depths not compatible with slowness sampling" errors

johnrudge:
### What does this PR do?

In some circumstances taup fails with a "TauBranch depths not compatible with slowness sampling." error. A simple example which triggers this is:

```
model = TauPyModel("ak135")
arrivals = model.get_ray_paths(2000.0, 60.0, ["P"])
```

I believe the bug lies in the `_fix_other_layers` routine in taup/slowness_model.py. That routine splits a layer, but the indexing assumes only ever a single layer gets split. In some circumstances more than one layer needs to be split, but later layers get inserted incorrectly because the size of `out` has grown and the subsequent indexing needs to be adjusted. This pull request fixes the indexing so the above example now works correctly. 

A related issue has been reported in #2682. This may also fix that issue, but I don't have the custom model used to test if that is the case. I think this is also the same issue as #1938. 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -10,6 +10,8 @@ Changes:
  - obspy.taup:
    * Fix cycling through colors in ray path plots, now also fixes cartesian
      plot version (see #2470, #2478, #3041)
+   * Fix slowness layer splits to avoid "TauBranch depths not compatible with
+     slowness sampling." errors. (see #3062, #2682, #1938)
 
 
 1.3.0 (doi: 10.5281/zenodo.6327346)
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -88,6 +88,7 @@ Rapagnani, Giovanni
 Reyes, Celso
 Ringler, Adam
 Rothenhäusler, Nicolas
+Rudge, John F.
 Russo, Emiliano
 Sales de Andrade, Elliott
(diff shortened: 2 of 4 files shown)


megies:
Nice catch! :+1: 
This should go into `maintenance_1.3.x`, actually. You could do `git rebase maintenance_1.3.x` on your branch, or if that makes problems simply delete the branch or hard reset it on maintenance and cherry-pick your commit.

Let me know if you need help. If all fails you can also just make a new branch and new PR I guess.

johnrudge:
Sorry about the wrong branch. It's now changed to maintenance_1.3.x

megies:
> Sorry about the wrong branch. It's now changed to maintenance_1.3.x

No worries!

Can you add a changelog one-liner for this please? Also please add the above example as a regression test. Just those three lines and maybe adding `assert int(arrivals[0].time) == 480` or similar should do, I think.

megies (approved):
- [x] changelog
- [x] simple regression test

Otherwise good to go =)

megies:
You could also add yourself to contributors on this branch if you want, I think you're only in the master branch contributors file =)

johnrudge:
Ok, I think I've done what you've asked @megies !

johnrudge:
Thanks for a speedy merge!

Issue #2816, opened 2021-03-22
taup arrival times are wrong for custom velocity model


martijnende:
(I've seen numerous closed/open issues that report bugs with taup, but I did not immediately recognise my problem as being a duplicate of previously opened issues. Please forgive my possible duplicate report)

# Problem description

I've been playing around with taup models for solving hydro-acoustic problems. To this end I created a custom velocity model (`tvel` format) using IASPEI91 as a starting model. For some reason I get various error messages when removing redundant layers, but that's a problem for another day.

`acoustic_model.tvel`:

```
Modified IASPEI91 model
Anything below 2.75 km is bogus
     0.000    1.5062    0.001     1.0
     2.750    1.5493    0.001     1.0
     2.750    5.8000    3.3600    2.7200
    20.000    5.8000    3.3600    2.7200
    20.000    6.5000    3.7500    2.9200
    35.000    6.5000    3.7500    2.9200
    35.000    8.0400    4.4700    3.3198
```

When I run the following code, I get some results, but the arrival times (even the direct ones) are completely wrong:

```
import os
import obspy
from obspy.taup import TauPyModel
from obspy.taup.velocity_model import VelocityModel
from obspy.taup.taup_create import build_taup_model

filename = "acoustic_model.tvel"

vmodel = VelocityModel.read_tvel_file(filename)
assert vmodel.validate()

taup_model = build_taup_model(filename, output_folder=os.getcwd())

model = TauPyModel(model=f"{filename[:-5]}.npz")

arrivals = model.get_ray_paths(
    source_depth_in_km=2.,
    distance_in_degree=5,
)

print(arrivals[:2])

fig = plt.figure(figsize=(9, 5), constrained_layout=True)
ax = arrivals[:2].plot_rays(plot_type="cartesian", fig=fig)
```

Output:
```
2 arrivals
	p phase arrival at 2.349 seconds
	P phase arrival at 2.696 seconds
```

![taup](https://user-images.githubusercontent.com/6959003/112023689-c3d45e80-8b33-11eb-9241-8e26f50fce22.png)

While I find it curious that the diffracted arrival does not exactly hit the receiver, I am more worried about the direct phase which supposedly arrives after 2.3 seconds. The acoustic speed in water is about 1.5 km/s, so to bridge a 5 degree gap would take about 6 minutes (which I verified with an Eikonal simulation).

# Other information

ObsPy version: `1.2.2`
Installation method: `conda`
Operating system: `Linux Mint 19`

[...]

PR #3064, opened 2022-05-10
Calculate more accurate ray paths in taup

johnrudge:
### What does this PR do?

The ray paths calculated by obspy.taup are often rather inaccurate, in that they miss the receiver by a small, but significant, margin. A simple example of this is:

```
import numpy as np

model = TauPyModel("prem")
arrivals = model.get_ray_paths(source_depth_in_km = 0.0, distance_in_degree = 120.0, phase_list = ['PKIKP'])
arrival = arrivals[0]
diff = (arrival.purist_distance*np.pi/180.0 - arrival.path["dist"][-1])
print(diff*180/np.pi, "degrees")

diff_time = (arrival.time - arrival.path["time"][-1])
print(diff_time, "seconds")
```
which outputs
```
-0.026268729081271754 degrees
-0.05079122889264909 seconds
```
i.e. the path misses the receiver by 0.03 degrees, and has a discrepancy in arrival time of 0.05 seconds. While not terrible, users might expect better accuracy than this, depending their application.

The key bit of logic that controls how accurately the ray paths hit the receiver is the `refine_arrival` method of `SeismicPhase` in `obspy/taup/seismic_phase.py`. This uses a shooting method to determine the ray parameter that brings the arrival closest to the station. A hand-coded false postion (regula falsi) root finding scheme is used to find the ray parameter. This hand-coded scheme terminates either after a set number of iterations (default 5), or when the arrival is within REFINE_DIST_RADIAN_TOL (default is 0.0049 degrees). In the example above the algorithm has terminated because it has done 5 iterations, not because it has reached the desired distance tolerance.

To improve the accuracy, this pull request replaces the hand-coded regula falsi scheme with one of scipy's built-in root-finding routines (`brentq`). This routine gives much faster convergence to the root. The pull request also changes the default tolerance and iteration limit to give much more accurate ray paths by default.

This pull request does have a notable effect on runtime. It roughly doubles the amount of time needed to calculate a ray path, as more calls to `shoot_ray` are made. However, I think this extra runtime is worth the extra accuracy. The current implementation is certainly not doing enough iterations to hit the receiver with the desired accuracy.

A consequence of this pull request is divergence with the Java taup toolkit, which uses an identical regula falsi scheme with the same defaults. This affects the tests in test_tau.py which often compare directly with the output from the java version. With the pull request, subtle differences in `ray_param_sec_degree`, `takeoff_angle`, `incident_angle` in the arrival occur, as well as subtle differences in the ray paths. But these are expected because of the more accurate computation of the ray parameter. To ensure the tests pass this pull request increases the desired test tolerance on comparisons which will change because of the new implementation. It also currently skips the `test_paths_for_crustal_phases` path comparison test.

See #1189 for the original discussion of the shoot_ray implementation. 
For the `test_paths_for_crustal_phases` see #1392
#2816 is a current issue that shows an example plot where an arrival does not hit the receiver with sufficent accuracy.

### PR Checklist
- [X] All tests still pass.
- [X] Add the "ready for review" tag when you are ready for the PR to be reviewed.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,8 @@ Changes:
  - obspy.taup:
    * add option "indicate_wave_type" to distinguish S waves in ray paths
      plot by using wiggly lines for shear waves (see #3047)
+   * improved accuracy of ray paths by change of root-finding algorithm
+     in SeismicPhase.refine_arrival
 
 maintenance_1.3.x
 =================
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,7 +12,7 @@ Changes:
    * add option "indicate_wave_type" to distinguish S waves in ray paths
      plot by using wiggly lines for shear waves (see #3047)
    * improved accuracy of ray paths by change of root-finding algorithm
-     in SeismicPhase.refine_arrival
+     in SeismicPhase.refine_arrival (see #3064)
 
 maintenance_1.3.x
 =================
--- a/obspy/taup/__init__.py
+++ b/obspy/taup/__init__.py
@@ -106,7 +106,7 @@ object which can be queried for various attributes.
 
 >>> arr = arrivals[0]
 >>> arr.ray_param, arr.time, arr.incident_angle  # doctest: +ELLIPSIS
-(453.7535..., 485.2100..., 24.3988...)
+(453.7536..., 485.2100..., 24.3988...)
 
 Ray Paths
 ^^^^^^^^^
--- a/obspy/taup/__init__.py
+++ b/obspy/taup/__init__.py
@@ -64,7 +64,7 @@ and model. By default it returns arrivals for a number of phases.
     PcP phase arrival at 674.865 seconds
     PP phase arrival at 794.992 seconds
     PKiKP phase arrival at 1034.098 seconds
-    pPKiKP phase arrival at 1050.529 seconds
+    pPKiKP phase arrival at 1050.528 seconds
     sPKiKP phase arrival at 1056.721 seconds
     S phase arrival at 1176.948 seconds
     pS phase arrival at 1195.508 seconds
--- a/obspy/taup/seismic_phase.py
+++ b/obspy/taup/seismic_phase.py
@@ -1275,30 +1275,34 @@ class SeismicPhase(object):
         Use a shooting method to improve ray path.
         """
 
+        left = Arrival(self, degrees, self.time[ray_index],
+                       self.dist[ray_index], self.ray_param[ray_index],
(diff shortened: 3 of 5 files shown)


johnrudge:
Ok, I've fixed the minor doctest issues. All over to you @megies to srutinise!

megies (changes requested):
In principle it's a good idea to stay as close to the Java original version as possible, but this change is not a lot of code changes and it's confined to one method. And we definitely want to prioritize for a bit more accuracy even if it's a bit slower, we've seen how people got confused/annoyed with too large discrepancies in those other issues you linked.

So I'm all for it. :+1: 

Two things tho:

1) This adds about a minute on total (non-network) test runtime, doubling the taup submodule test runtime (which already is on the brink):
 - [recent master test](https://tests.obspy.org/123542/)
 - [this PR](https://tests.obspy.org/124013/)

Can we please have the tests run with less accuracy to speed it up? You should be able to set those two constants to lower values at the top of test_tau...*.py or maybe make it some pytest fixture or something (not super familiar with how pytest does things like that best, @trichter?).
Tests should just make sure that the results are reproducible and stable, they don't have to be on point in terms of accuracy of those paths. =)
You can leave 1 or a couple tests with full accuracy though, to have that case covered too.

Screenshot from 2022-05-12 12-55-15

2) We should maybe consider some way to point out to users how to speed up these calculations while sacrificing accuracy. We probably don't want to go all out on this, but maybe we could point out to users in the docs in the respective places that these constants could be changed to tweak it? Dunno.. maybe not necessary

megies:
> One could fiddle with `REFINE_DIST_RADIAN_TOL` in the tests to bring the peformance closer to current master, but I don't think it's worth it.

Well.. that's a hen/egg problem. Are people not caring about test speed because the whole test suite takes ages anyway or does the whole test suite take ages because people were not caring about test speed. :zany_face: 

I know it' rough to spend additional time on speeding some part of the tests up with the whole thing taking so long, but on the other hand, it just gets outta hand more and more if we slack on it. :grimacing: 

> so going around 25% slower

I think we could live with that increase if you don't wanna mess with tweaking tests more

> I think it could be potentially confusing having the tests run with a non-default value of `REFINE_DIST_RADIAN_TOL`

I don't think it would be confusing. A lot of tests just run on some random data, to not have to read in actual real data.
We would just want to make sure to not reduce those constants for doctests (that might produce pics even). But for unittests it really doesn't matter if the results are less accurate, as long as there is like 1-2 tests with the new regular better tolerance.

megies (approved):

johnrudge:
I don't want to mess with the tests any further. I'm afraid I'll leave that to you!

PR #3070, opened 2022-05-22
Add support for models without a core or inner core in taup

johnrudge:
### What does this PR do?

This pull request builds on PR #2390, adding support in obspy.taup for models with missing discontinuities, such as no core or no inner core. The pull request contains the code from #2390, along with a port of the changes made to the java code to handle these cases https://github.com/crotwell/TauP/commit/b712c614f9a589b3beb6524096e03230a460a914
PR #2390 failed because of a bug in the `TauBranch.difference` method. The method assumed that if `index_p == -1` then one would also always have `index_s == -1`, but there are cases where that fails (and in particular it fails for their test example). This PR fixes the logic for `TauBranch.difference`.
Credit for this PR should go to @not522, @trichter, and @crotwell. I'm just putting the pieces together.

### Why was it initiated?  Any relevant Issues?

This fixes the issues with Mars models #1200 and #1423, and likely also fixes the issues with the solar model #2086 and the custom model in #2192. 

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just add the "build_docs" tag to this PR.
      Docs will be served at [docs.obspy.org/pr/{branch_name}](https://docs.obspy.org/pr/) (do not use master branch).
      Please post a link to the relevant piece of documentation.
- [ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the "test_network" tag to this PR.
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are covered via new tests.
- [ ] Any new or changed features are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .
- [ ] If the changes affect any plotting functions you have checked that the plots
      from all the CI builds look correct. Add the "upload_plots" tag so that plotting 
      outputs are attached as artifacts. 
- [x] Add the "ready for review" tag when you are ready for the PR to be reviewed.

PR #3080, opened 2022-05-25
Taup fixes for low velocity zones

johnrudge:
### What does this PR do?

This PR ports two bug fixes from recent updates to the java TauP toolkit. The first of these is:
https://github.com/crotwell/TauP/commit/062e05169c4e6acc5ece47b46f4f534d63d8b687
which deals with an issue where phases cannot turn due a low velocity layer. This comes with a new test involving a fruity planetary model!

The second is a one-liner:
https://github.com/crotwell/TauP/commit/f74aadf578a962c100048ede339ef70c88f098d8
which follows on from #3070, and stops diffracted phases occuring in models without a core.

Follows on from #3070, in improving support for more general planetary models.

### PR Checklist
- [ ] All tests still pass.


johnrudge:
I'm converting this to a draft as it may be affected by issue https://github.com/crotwell/TauP/issues/18 in the java version. I will come back to this PR when that issue is resolved.

megies:
Ok, just let us know when its time to check back on this

obspy/taup/seismic_phase.py
```diff
@@ -832,6 +832,31 @@ def add_to_branch(self, tau_model, start_branch, end_branch, is_p_wave,
...
+                if (tau_model.s_mod.depth_in_high_slowness(
+                     tau_model.get_tau_branch(b_num, is_p_wave).top_depth,
+                     self.min_ray_param,
+                     is_p_wave) and
+                    (b_num + 1 >= len(tau_model.tau_branches[0])
+                     or self.min_ray_param <=
+                     tau_model.get_tau_branch(b_num+1,
+                                              is_p_wave
+                                              ).max_ray_param
+                     )):
```
megies, line 850: even if these conditions are single use, I think it would be better to have them singled out for readability.. even just the names for the condition would help a lot when coming back to this later  (naming just as an example as educated guesses)

```
while ...:
    slow_layer = tau_model...
    last_branch = b_num + 1 >= len(...
    ray_param_out_of_bounds = self.min_ray_param <= ...
    if (slow_layer and last_branch) or ray_param_out_of_bounds:
        ...
```

megies (approved):
One little cosmetic recommendation, otherwise looks good as always @johnrudge :rocket:

obspy/taup/seismic_phase.py
johnrudge: The code style here was inherited from the java version. But I totally agree it would read better with the logic broken down into smaller steps, so I've done that. Once the tests have run take another look.

megies:
About these if/else condition bool cosmetics, I totally understand the being in line with the java implementation and also I know that it is done like that throughout our whole port. So I totally agree if you don't particularly feel like doing this differently. It's just a pity overall, since if anything in taup comes up, people that aren't into the details (like me) can literally do nothing about reported bugs, since every single line and condition is like a book with seven seals. With some more intuitive verbosity to conditions it might be easier accessible to more people, but probably also just wishful thinking.. :rofl: 
In any case go about everything in our taup fork as you see fit. :wink:

PR #3088, opened 2022-06-07
bugfix 3067

### LINKED ISSUE #3067: Inventory map plot: Missing legend and artifacts from legend setup
### Avoid duplicates

### Bug Summary

With 9659cf3e4499fe15fa71bd0bd3a9600aca7f180e, while fixing the docs build apparently, two small bugs were introduced.

 a) Legend is removed from the plot, right after adding it
 b) Helper artists (bogus scatter plots), used to set up the legend are not removed anymore, ending up in markers lingering at 0°N 0°E

```diff
@@ -1010,20 +1013,20 @@ class Inventory(ComparingObject):
 
         if legend is not None and color_per_network:
             ax = fig.axes[0]
-            count = len(ax.collections)
             for code, color in sorted(color_per_network.items()):
                 ax.scatter([0], [0], size, color, label=code, marker=marker)
             # workaround for older matplotlib versions
             try:
-                ax.legend(loc=legend, fancybox=True, scatterpoints=1,
-                          fontsize="medium", markerscale=0.8,
-                          handletextpad=0.1)
+                leg = ax.legend(loc=legend, fancybox=True, scatterpoints=1,
+                                fontsize="medium", markerscale=0.8,
+                                handletextpad=0.1)
+                leg.remove()
             except TypeError:
                 leg_ = ax.legend(loc=legend, fancybox=True, scatterpoints=1,
                                  markerscale=0.8, handletextpad=0.1)
-                leg_.prop.set_size("medium")
+                leg_.remove()
             # remove collections again solely created for legend handles
-            ax.collections = ax.collections[:count]
+            # ax.collections = ax.collections[:count]
 
         if outfile:
             fig.savefig(outfile)
```
 a) is trivial, just get rid of these remove calls again
 b) currently calls like `.pop()` and `.remove()` work on `ax.collections`, but there's a [note that with matplotlib 3.7 this will get changed to being immutable](https://github.com/matplotlib/matplotlib/blob/c6c7ec1978c22ae2c704555a873d0ec6e1e2eaa8/lib/matplotlib/axes/_base.py#L1327-L1336), so this needs to be done differently and correctly right now

---

In addition, this part also shows a warning:

```
*c* argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with *x* & *y*.  Please use the *color* keyword-argument or provide a 2D array with a single row if you intend to specify the same RGB or RGBA value for all points.
```

This simply needs to be changed to `color=color`:

```
ax.scatter([0], [0], size, color, label=code, marker=marker)
```

### Code to Reproduce

```python
from obspy import read_inventory

inv = read_inventory()
inv.plot(color_per_network=True, projection='global')
```

### Error Traceback

_No response_

### ObsPy Version?

master

### Operating System?

_No response_

### Python Version?

_No response_

### Installation Method?

developer installation / from source / git checkout

christadler:
### What does this PR do?
 
Fixes  #3067 

Legend is no longer removed. The unnecessary marker at 0degN, 0degE will disappear.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,8 @@ Changes:
    * Trace.resample: Changed default for `window` to "hanning" following a name
      change in scipy, "hann" is not recognized anymore in newest scipy (see
      #3117)
+   * Fix missing legend and plot artifacts in Inventory map plots at
+     intersection of equator and prime meridian (see #3067)
  - obspy.clients.fdsn:
    * add URL mapping 'EIDA' for http://eida-federator.ethz.ch (see #3050)
  - obspy.clients.nrl:
--- a/obspy/core/inventory/inventory.py
+++ b/obspy/core/inventory/inventory.py
@@ -1016,14 +1016,9 @@ class Inventory(ComparingObject):
             for code, color in sorted(color_per_network.items()):
                 helpers.append(ax.scatter([0], [0], size,
                                color=color, label=code, marker=marker))
-            # workaround for older matplotlib versions
-            try:
-                ax.legend(loc=legend, fancybox=True, scatterpoints=1,
-                          fontsize="medium", markerscale=0.8,
-                          handletextpad=0.1)
-            except TypeError:
-                ax.legend(loc=legend, fancybox=True, scatterpoints=1,
-                          markerscale=0.8, handletextpad=0.1)
+            ax.legend(loc=legend, fancybox=True, scatterpoints=1,
+                      fontsize="medium", markerscale=0.8,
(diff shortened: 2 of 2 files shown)


trichter:
Maybe you could also use the approach mentioned in https://github.com/obspy/obspy/issues/3067#issuecomment-1127668349 and replace `ax.scatter` with `Line2D` plus using `handles=helpers` in `ax.legend`? In this way no additional data is plotted on the axis itself.

PR #3092, opened 2022-06-14
Rapid travel time plotting in taup

johnrudge:
### What does this PR do?

This pull request modifies `plot_travel_times` in `obspy.taup` to allow for rapid plotting. It does this by plotting the precalculated times in the `TauModel` rather than recalculating arrivals at specific epicentral distances. 

The old behaviour is:
```
from obspy.taup import plot_travel_times
plot_travel_times(source_depth=10, phase_list=["P", "S", "PP"], npoints=50) 
```
old

The new behaviour is:
```
plot_travel_times(source_depth=10, phase_list=["P", "S", "PP"], npoints=None) 
```
new

where I have made `npoints=None` the new default, rather than `npoints=50`.

### Why was it initiated?  Any relevant Issues?

I don't think this is related to any existing issues. Recomputing all the travel times at a series of fixed epicentral distances was painfully slow, especially with a lot of phases. I've updated the relevant figures that use this function in the docs. The old behaviour is still available by setting the `npoints` argument to an integer value.

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [x] This PR is not directly related to an existing issue (which has no PR yet).
- [x] If the PR is making changes to documentation, docs pages can be built automatically.
      Just add the "build_docs" tag to this PR.
      Docs will be served at [docs.obspy.org/pr/{branch_name}](https://docs.obspy.org/pr/) (do not use master branch).
      Please post a link to the relevant piece of documentation.
- [ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the "test_network" tag to this PR.
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are covered via new tests.
- [ ] Any new or changed features are fully documented.
- [x] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .
- [ ] If the changes affect any plotting functions you have checked that the plots
      from all the CI builds look correct. Add the "upload_plots" tag so that plotting 
      outputs are attached as artifacts. 
- [x] Add the "ready for review" tag when you are ready for the PR to be reviewed.

Issue #1639, opened 2017-01-10
malloc error in taup get_travel_times for certain Mars nd format models


mpanning:
obspy version: 1.0.2
Python version: 2.7.12
Platform: OS X 10.12.2 using macports python and obspy ports

For certain distances, the get_travel_times method fails causing the script to crash with no Python error with the following error message:

> Python(66056,0x7fffebf493c0) malloc: *** error for object 0x7fe21140ea88: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort

However, after adding a print statement into my script within the try-except construction that is calling get_travel_times, the code fails with a Segmentation Fault.  I assume this is a problem in some non-python portion of the taup package, but I have not yet attempted to debug further.

I'm attaching a python zip file with a python script which generates the failure (plot_travel_times.py), as well as 2 model files (DWThot.nd and EH45Tcold.nd).  The first one fails unpredictably at distances between 64 and 67 degrees, while the second model proceeds to completion. Both models actually have a completely fluid core, and include a dummy layer at the end of the file with inner-core named discontinuity at the planet center. 

I'm also including the npz files generated, but these can be moved to another directory, and the script will use taup.taup_create.build_taup_model to make a new npz file.

The script expects an argument with the input model file, and a second optional argument with an output file name (which defaults to tt.png).

[obspy_taup_error.zip](https://github.com/obspy/obspy/files/697198/obspy_taup_error.zip)

krischer:
Both of these issues are already fixed in the `master` or `maintenance_1.0.x` branches. I confirm that both errors also happen on my machine with `1.0.2`.

Still - this does not mean that the travel times are correct. The hot one looks weird. Why should it have an S-wave shadow zone at those epicentral distances?

It might be that the phase naming scheme just breaks down if some assumptions about Earth are no longer fulfilled? Do you know how the Java TauP handles these models?

**DWThot:** ![dwthot](https://cloud.githubusercontent.com/assets/800487/21829970/e469fa62-d799-11e6-957f-a8634cbccc4d.png)

**EH45Tcold:** ![eh45tcold](https://cloud.githubusercontent.com/assets/800487/21829971/e46ae35a-d799-11e6-9af1-7478225075b3.png)

krischer:
Reference travel times for example through ray tracing would be very useful to have!

martinvandriel:
[...]

PR #3103, opened 2022-06-27
Avoid plotting across shadow zones in plot_travel_times

johnrudge:
### What does this PR do?

This is a small tweak to the rapid travel time plotting in #3092. That function plotted each phase as one continous line, which is an issue if there is a shadow zone. This PR breaks the curve at all shadow zones.

An example, using `DWThot.npz` from issue #1639
```
from obspy.taup import TauPyModel, plot_travel_times
model = TauPyModel("DWThot.npz")
plot_travel_times(0.0, model=model, phase_list=["P", "S"])
```
Old behaviour:
old
New behaviour:
new
With this PR the shadow zone for S is plotted correctly.

This also fixes a small issue with `Arrivals.plot_rays()`. When `indicate_wave_type=True` and `legend=True` there could be duplicate legend labels. This PR removes the duplicates.

Builds on #3092.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/taup/seismic_phase.py
+++ b/obspy/taup/seismic_phase.py
@@ -1655,6 +1655,30 @@ class SeismicPhase(object):
     def get_earliest_arrival(cls, rel_phases, degrees):
         raise NotImplementedError("baaa")
 
+    def _shadow_zone_splits(self):
+        """
+        A list of slices that split  on any shadow zones.
+        """
+        if self.head_or_diffract_seq:
+            return [slice(0, None)]
+        else:
+            return _repeated_value_splits(self.ray_param)
+
+
+def _repeated_value_splits(x):
+    """
+    A list of slices that split a 1D numpy array where values repeat.
+    """
+    split_indices = []
+    idxs = np.where(x[1:] - x[:-1] == 0)[0] + 1
+    left_index = 0
+    for i in idxs:
+        split_indices.append(slice(left_index, i))
+        left_index = i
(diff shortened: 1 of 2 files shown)


obspy/taup/seismic_phase.py
```diff
@@ -1655,6 +1655,32 @@ def calc_incident_angle(self, ray_param):
...
+            # shadow zones are where ray parameter repeats
+            return _repeated_value_splits(self.ray_param)
+
+
+def _repeated_value_splits(x):
+    """
+    A list of slices that split a 1D numpy array where values repeat.
+    """
+    split_indices = []
+    idxs = np.where(x[1:] - x[:-1] == 0)[0] + 1
```
megies, line 1675: looks like this might be equivalent to

```python
idxs = np.diff(x) == 0
```

which might be a bit easier to read. but it's just cosmetics, so doesn't bother me too much.

megies (approved):
Looks good to me :rocket:

obspy/taup/seismic_phase.py
megies: ah nevermind, you'd need to wrap around `np.argwhere` or something so lets just leave as is :smile:

megies: ah but `x[1:] - x[:-1]` is equivalent to `np.diff(x)`, let me quickly do that and merge then

PR #3107, opened 2022-07-08
Added exception for unpacking error of AH file reading

s-schneider:
### What does this PR do?

Fixes an issue with `_unpack_string` method for AH file format. I encountered that already some time ago but only fixed it locally

No relevant issues yet.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/ah/core.py
+++ b/obspy/io/ah/core.py
@@ -116,7 +116,7 @@ def _unpack_string(data):
     string = data.unpack_string()
     try:
         string = string.split(b'\x00', 1)
-    except Exception:
+    except UnicodeDecodeError:
         string = string.split()
     string = string[0].strip().decode("utf-8")
     return string
--- a/obspy/io/ah/core.py
+++ b/obspy/io/ah/core.py
@@ -114,8 +114,8 @@ def _get_ah_version(filename):
 
 def _unpack_string(data):
     try:
-        string = data.unpack_string().split(b'\x00',
-                                            1)[0].strip().decode("utf-8")
+        string = data.unpack_string().split(
+            b'\x00', 1)[0].strip().decode("utf-8")
     except Exception:
         string = data.unpack_string().split()[0].strip().decode("utf-8")
     return string
--- a/obspy/io/ah/core.py
+++ b/obspy/io/ah/core.py
@@ -15,6 +15,8 @@ a number of values followed by the time series data.
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
+import warnings
+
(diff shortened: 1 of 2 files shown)


megies:
@s-schneider exception handling should be as specific as possible to avoid catch-all situations, so I made some changes. I'd still like the `except` clause to be more specific.. what kind of exception gets raised? Ideally a short test case with a small test file if needed would be good to have

megies:
Base branch will hate to be changed to `maintenance_1.4.x` once 1.4.0 is released.

s-schneider:
Sorry for the exceptional late reply. I agree with the exception handling. I need to look into the files again and check which ones raised the error, and which error it was in specific.

s-schneider:
I'm not sure what is going on with the tests, that seems not related to my PR?

d-chambers:
Hey @s-schneider, something is messed up with the CI. I will look into sometime this week.

obspy/io/ah/core.py
```diff
@@ -113,7 +113,13 @@ def _get_ah_version(filename):
 
 
 def _unpack_string(data):
-    return data.unpack_string().split(b'\x00', 1)[0].strip().decode("utf-8")
+    string = data.unpack_string()
+    try:
+        string = string.split(b'\x00', 1)
```
megies, line 118: I don't see how this would ever be able to trigger a `UnicodeDecodeError`.. this is on bytes? what was the original problem?

Would be good to add a small test case I think..

megies:
@s-schneider can you please post an example of what the exact problem is you are trying to solve, because I'm confused at the "fix" which seems to be to put code in a `try/except UnicodeDecodeError` that I can't see how it ever would be able to raise that exception.. Tryin to get 1.4.0 out soon and might bump this otherwise

s-schneider:
I will add that soon, have a bit of a tight schedule work wise right now. I don't think it needs to be in 1.4, it is not an urgent fix

s-schneider:
[112998B.AH.zip](https://github.com/obspy/obspy/files/10001719/112998B.AH.zip)
This is an example file

megies:
Yeah, the changes in here don't fix reading that file, looks like `event.comment` field has some data in it that do not make sense in terms of UTF or ASCII. Most sense probably makes showing a warning and continuing to decode with replace errors option.

megies:
@s-schneider with this tiny fix your file can be read. also rebased on current master.

Screenshot from 2022-11-15 13-11-08

megies:
Merging this for now, feel free to handle it differently in another PR.

s-schneider:
Thanks, the warning seems like a better solutiion

PR #3112, opened 2022-07-19
docs: Fix a few typos

timgates42:
There are small typos in:
- obspy/io/mseed/src/libmseed/genutils.c
- obspy/signal/src/evalresp/spline.c

Fixes:
- Should read `denominator` rather than `denomiator`.
- Should read `diagonal` rather than `diagaonal`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


megies:
Thanks, but both of these should rather be reported upstream.

PR #3117, opened 2022-08-01
Changed from hanning to hann following scipy's new convention.

### LINKED ISSUE #3116: Scipy v1.9.0 renames scipy.signal.hanning to scipy.signal.hann
### Problem

Automated tests that use scipy v1.9.0 fail because `scipy.signal.hanning` has been removed in favor of `scipy.signal.hann` as noted in the [changelog](https://docs.scipy.org/doc/scipy/release.1.9.0.html). It seems like this has been in the works since [v1.1.0](https://docs.scipy.org/doc/scipy/release.1.1.0.html?highlight=hann).

### Proposed solution

Move to embracing the `scipy` convention by using `hann` instead of `hanning` in the code base. 

Alternatively, just make this change when calling `scipy.signal.get_window`. Note that `numpy` hasn't made this change and still uses `numpy.hanning` see this [discussion](https://github.com/numpy/numpy/issues/17303).

arkottke:
### What does this PR do?

Fixes #3116 by changing *hanning* to *hann* following the scipy statnd.

See #3116. Scipy v1.9.0 removed hanning.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,9 @@ maintenance_1.3.x
 Changes:
  - obspy.core:
    * Fix exclusion of magnitude 0 events in Catalog.filter (see #3105)
+   * Trace.resample: Changed default for `window` to "hanning" following a name
+     change in scipy, "hann" is not recognized anymore in newest scipy (see
+     #3117)
  - obspy.clients.fdsn:
    * add URL mapping 'EIDA' for http://eida-federator.ethz.ch (see #3050)
  - obspy.io.wav:
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -49,6 +49,7 @@ Isken, Marius
 Kearns, Aaron
 Ketchum, David
 Khan, Lewis
+Kottke, Albert
 Koymans, Mathijs
 Kremers, Simon
 Kress, Victor
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -2291,7 +2291,7 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             tr.trigger(type, **options)
         return self
 
-    def resample(self, sampling_rate, window='hanning', no_filter=True,
+    def resample(self, sampling_rate, window='hann', no_filter=True,
                  strict_length=False):
         """
         Resample data in all traces of stream using Fourier method.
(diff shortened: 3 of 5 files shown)


megies:
Replacing it in the call syntax for resample as default for `window` is a no-brainer and that doesn't break anything. But I agree, maybe it's not necessary to rename it in the Lancosz part? To not break things. I don't think it's necessary that it's the same name in both parts.

arkottke:
@megies yeah, I thought that was a bit drastic of a change, but I was hoping for consistency and then I discovered the numpy hasn't changed this as well.

We might want to add an alias of *hanning* to *hann* in `resample(...)` and note the is name is depreciated.

megies:
> We might want to add an alias of _hanning_ to _hann_ in `resample(...)` and note the is name is depreciated.

Not sure if that's even needed. It was the default, so I imagine most people weren't explicitely specifying it. Furthermore, I'd imagine scipy has been showing a deprecation warning.

megies:
In the lancosz part, it doesn't even has to do with numpy, as far as I see it's strictly our own naming in there and I think there's no reason to change that part.

arkottke:
My brain just likes consistency so referring to the window as *hann* in one part of the code and *hanning* in another felt wrong, which is why I made the modifications that I did. I rolled back the changes to the interpolation routines to keep using *hanning*.

megies:
I totally agree, it would be good to have the same naming consistently. But in this case, I'd say stability and not breaking existing codes is more important
Thanks for rolling that part back  =)

megies:
OK, so this is pretty much ready to merge.. but, I actually think we should have this change in the maintenance branch for upcoming 1.3.1. It does not break anything and will make sure that code that is using resample without specifying a different window will keep working.

So.. I would argue this should be rebased on maintenance_1.3.x. Let me know if you need help with that. I can do that too if needed.

arkottke:
It would be great if you can take care of rebasing on maintenance_1.3.x as
I won't be able to get to it for about 10 days.

megies:
> It would be great if you can take care of rebasing on maintenance_1.3.x as I won't be able to get to it for about 10 days.

Done. I'll also force push your master branch to resemble our main master after merging this. Ideally you want to make dedicated branches for PRs =)

megies (approved):

megies:
Actually github is pretty smart with their permissions, it wouldn't let me force push to your branch after merging this PR, so you'll have to reset your master branch yourself eventually (or just ignore your fork's master branch like I usually do, it's not really needed for anything ;)

CHANGELOG.txt
```diff
@@ -4,6 +4,9 @@ maintenance_1.3.x
 Changes:
  - obspy.core:
    * Fix exclusion of magnitude 0 events in Catalog.filter (see #3105)
+   * Trace.resample: Changed default for `window` to "hanning" following a name
```
liamtoney, line 7: "hanning" and "hann" are reversed here.

liamtoney (commented):
FYI, I noticed a small mixup in the changelog.

CHANGELOG.txt
```diff
@@ -4,6 +4,9 @@ maintenance_1.3.x
...
+     change in scipy, "hann" is not recognized anymore in newest scipy (see
```
liamtoney, line 8: And here.

megies: Fixed, thanks!

megies:
Fixed, thanks @liamtoney !

Issue #2608, opened 2020-05-02
Data length from Stream.trim differ from other trim/slice methods


GentleMin:
I am using obspy 1.2.1 on windows. I have encountered a rather weird situation using `stream.trim` with `stream.slice`, `trace.trim` and `trace.slice`. Not sure if it is intended, but the 1st methods seems to be behaving slightly different from the others. Basically, in the following instance, `stream.trim` yields a 1-point shorter array than the others, even if `nearest_sample` is always  used.
This is the original stream.
```
ipdb> print(st_b)
3 Trace(s) in Stream:
BD.BD09..HLE | 2017-05-24T09:10:14.284121Z - 2017-05-24T12:10:14.274121Z | 100.0 Hz, 1080000 samples
BD.BD09..HLN | 2017-05-24T09:10:14.284121Z - 2017-05-24T12:10:14.274121Z | 100.0 Hz, 1080000 samples
BD.BD09..HLZ | 2017-05-24T09:10:14.284121Z - 2017-05-24T12:10:14.274121Z | 100.0 Hz, 1080000 samples
```

Now `stream.slice` is used. The result has 6801 sample points.
```
ipdb> st_b_2 = st_b.slice(starttime=reftime + float(move[0]), endtime=reftime + float(move[1]), nearest_sample=True)

ipdb> print(st_b_2)
3 Trace(s) in Stream:
BD.BD09..HLE | 2017-05-24T09:16:52.124121Z - 2017-05-24T09:18:00.124121Z | 100.0 Hz, 6801 samples
BD.BD09..HLN | 2017-05-24T09:16:52.124121Z - 2017-05-24T09:18:00.124121Z | 100.0 Hz, 6801 samples
BD.BD09..HLZ | 2017-05-24T09:16:52.124121Z - 2017-05-24T09:18:00.124121Z | 100.0 Hz, 6801 samples
```

Next I tried `trace.trim` on the result. There is no difference, and the length remains 6801.
```
ipdb> st_b_2[0].trim(starttime=reftime + float(move[0]), endtime=reftime + float(move[1]), nearest_sample=True)

ipdb> print(st_b_2)
3 Trace(s) in Stream:
BD.BD09..HLE | 2017-05-24T09:16:52.124121Z - 2017-05-24T09:18:00.124121Z | 100.0 Hz, 6801 samples
[...]

PR #3123, opened 2022-08-10
Fix Stream.slice/trim return traces with different npts

trichter:
### What does this PR do?

Stream.trim/slice return traces with the same number of data points now, even for edge cases.
Added keep_empty_traces parameter to Stream.trim to stay consistent with Stream.slice.

See issue #2608

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -1515,6 +1515,8 @@ class Stream(object):
         BW.RJOB..EHN | 2009-08-24T00:20:20.000000Z ... | 100.0 Hz, 501 samples
         BW.RJOB..EHE | 2009-08-24T00:20:20.000000Z ... | 100.0 Hz, 501 samples
         """
+        if not self:
+            return self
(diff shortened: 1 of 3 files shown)


obspy/core/stream.py
```diff
@@ -1530,13 +1533,14 @@ def trim(self, starttime=None, endtime=None, pad=False,
                     endtime = tr.stats.endtime + delta * tr.stats.delta
             except TypeError:
                 msg = ('starttime and endtime must be UTCDateTime objects '
-                       'or None for this call to Stream.trim()')
+                       'or None for this call to Stream.slice()')
```
megies, line 1536: isn't this supposed to state `trim`? seems to be inside a trim call, no?

megies (approved):

obspy/core/stream.py
trichter: You are right. I changed the wrong error message.

Issue #2623, opened 2020-06-05
Stream.rotate() with multiple-station Streams not implemented?


liamtoney:
I ran into an issue today trying to rotate a Stream containing multiple stations from `'NE->RT'`. Each Trace has `stats.back_azimuth` defined; of course for different stations the values are different. I kept on getting weird results until I realized that the code for `Stream.rotate()` uses the `stats.back_azimuth` of the first Trace in the Stream. This clearly does not work for mixed-station Streams.

I guess this issue is either a feature request (for mixed-station Stream rotation) or a documentation suggestion (for more clarity that `st[0].stats.back_azimuth` is used **and overwrites the value for all other Traces**), since I had to dig a bit to realize that `Stream.rotate()` only works on single-station Streams.

My workaround for this is
```python
import numpy as np

for station in np.unique([tr.stats.station for tr in st]):
    st.select(station=station).rotate('NE->RT')
``` 
which is not much of a pain. Thanks all, I hope this isn't a dupe (didn't see it in my search).

**Setup details:**

* Python version: 3.8.3
* ObsPy version: 1.2.1 installed via conda
* macOS Mojave

trichter:
Hi Liam, thanks for reporting this issue. This behaviour could definitely be improved. It's not only a problem for data of different stations and one event, but also for a single station and multiple events. Also, rotate will have problems if - for some reason - the traces in the stream are not "correctly" ordered. Related to #1918.

liamtoney:
Thanks for your response, Tom.

> Also, rotate will have problems if - for some reason - the traces in the stream are not "correctly" ordered.

What do you mean by not correctly ordered?

trichter:
> What do you mean by not correctly ordered?

Considering a stream of traces `ST1..Z, ST2..N, ST1..E, ST2..Z, ST1..N, ST2..E` rotate will mix data of different stations.

liamtoney:
Yikes! Though I guess that is an unlikely scenario.

megies:
This is indeed really not ideal.. without looking at all the detials, I think we should change it to something like this:

```python
class Stream():
    def rotate(self, *args, **kwargs):
        st_new = Stream()
        seed_ids = set([tr.id[:-1] + '?' for tr in st])
        for seed_id in seed_ids:
            net, sta, loc, cha = seed_id.split('.')
            st_new += self.select(net, sta, loc, cha)._rotate(*args, **kwargs)
        self.traces = st_new.traces
        return self
[...]

PR #3124, opened 2022-08-10
fix Stream.rotate with multiple-station or multiple-event streams

trichter:
### What does this PR do?

Fix Stream.rotate() with multiple-station Streams

See #2623

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -10,6 +10,7 @@ Changes:
    * Fix missing legend and plot artifacts in Inventory map plots at
      intersection of equator and prime meridian (see #3067)
    * Fix a bug in recaculation of overall instrument sensitivity (see #3099)
+   * Fix Stream.rotate with multiple-station or multiple-event Streams (see #2623)
  - obspy.clients.fdsn:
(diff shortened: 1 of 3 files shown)


megies:
rebased and force pushed

megies:
I'm not sure about this one @trichter.. it looks like more magic introduced by this PR, I would've just gone with a recursive scheme which is not changing the existing logic. That should work without problems, although I didn't write rigorous tests for it yet.

trichter:
> it looks like more magic introduced by this PR,

I think I disagree, because this PR introduces minimal changes. It just looks up the baz and inci in the stats object of the first trace of the rotating substream and not in the first trace of the whole stream. As before it assumes the traces are already correctly sorted.

It also fixes a test case which I think was created exactly for this issue, but did not fail. (Now it would fail if only the test case was changed.)

Your PR is also nice, and IMHO includes much more magic than this one :) It fixes a different issue https://github.com/obspy/obspy/issues/2623#issuecomment-639872511. This issue here is fixed on the way, but not completely. I think the test case adapted in this PR will fail. It simulates the case in which two events are registered at a single station.

megies:
superseded by #3155

PR #3142, opened 2022-09-16
try to fix CI problem

megies:
### What does this PR do?

Changes obspy installation in CI from editable to a regular pip installation.

Also remove the `[tests]` modifier, which does nothing in combination with `--no-deps` anyway and we rely on an environment file for all (testing) dependencies anyway.

Some test currently fails because files are being looked up in the install location, which is only holding a link to the source tree though, due to being a pip editable installation.

 - https://github.com/obspy/obspy/actions/runs/3066925112/jobs/4952650308#step:10:268
 - https://tests.obspy.org/127117/#1
 - https://github.com/obspy/obspy/blob/9d10a93d9e862b46b4ce761530ddbb55b414a283/obspy/core/tests/test_waveform_plugins.py#L211-L220

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -178,7 +178,7 @@ jobs:
       - name: install obspy
         shell: bash -l {0}
         run: |
-          python -m pip install --no-deps -e .[tests]
+          python -m pip install --no-deps .
 
       # try to remove when min dep versions are upped
       - name: remove cartopy for mindepversion MacOS


megies:
@trichter not sure if there was a reason for doing an editable install in CI, but it started to cause issues and it looks like a regular install does just fine..

trichter:
I think for no specific reason.

trichter:
What I do not understand is, why this error appeared all of a sudden in the CI setup. I myself use this setup for my obspy installation. Will report when/if it breaks.

megies:
I use editable installs all the time too, locally. But CI can have some special setups makin it weird, looks like auto stripping of test file directries maybe? Who knows..

megies:
I actually can reproduce the problem locally now.. happens when using a pip editable install and then going via `obspy-runtests`.

trichter:
For my local installation it is still working. Maybe a change in the entry points stuff of setuptools. Should we revert this PR?

megies:
> For my local installation it is still working. Maybe a change in the entry points stuff of setuptools. Should we revert this PR?

This was fixed by #3142. On a fresh installation with all the newest packages and installing obspy editable after reverting that commit, it fails. As you say, seems to be a change in setuptools, reporting the shimmy directory instead of the git repository location as install base dir.

Anyway, I dont care, if you want to revert this go ahead.

megies:
to reproduce:
```
conda create -n trash obspy
conda activate obspy
conda remove --force --offline obspy
pip install -e /path/to/obspy
# manually revert this commit
obspy-runtests
```

megies:
Oh boy.. too many swapping around tabs and issues.. @trichter yes this change in CI script can be changed back to editable install if you want

trichter:
It's OK. I don't care enough.

PR #3161, opened 2022-10-02
Modified energy ratio trigger methods

dizcza:
### What does this PR do?

* Added `energy_ratio` and `modified_energy_ratio` trigger methods described in [Han2009](https://www.crewes.org/Documents/ResearchReports/2009/CRR200930.pdf)
* Improved `aic_simple` tutorial: visualized the results
* Added an explanation of STA/LTA to the trigger tutorial
* Changed the behaviour of `trigger_onset` from strict `>` to `>=` threshold comparison. This allows the users to pass the minimum and maximum values of the evaluated characteristic function as the arguments.
* Added `plot_trace` to the trigger module that plots the computed characteristic function without the trigger onset and offset.

### PR Checklist
- [x] All tests still pass.

----

I don't think that these methods should have unit test coverage as long as they are embedded in the tutorial - add tests for all other trigger methods and then it'd be fair to ask the users to follow the unit testing guideline.


misc/docs/source/tutorial/code_snippets/trigger_tutorial.rst
```diff
@@ -163,6 +173,30 @@ Delayed Sta Lta
...
+The picture below is taken from [Han2009]_.
+
+.. figure:: /_images/MER.png
+
+    FIG. 2: Definition of the MER method. Preceding and following energy
+    collection windows with equal lengths are located at a test point. The
+    window lengths are two to three times the dominant periods of the seismic
+    arrival.
+
+.. plot:: tutorial/code_snippets/trigger_tutorial_modified_energy_ratio.py
```
megies, line 197: I added `build_docs` label so we can have a look if this page renders alright

misc/docs/source/tutorial/code_snippets/trigger_tutorial_aic_simple.py
```diff
@@ -1,16 +1,25 @@
 """
-Credits: mbagagli
+Credits: mbagagli, Danylo Ulianych
 """
+import matplotlib.pyplot as plt
 
 from obspy.core import read, UTCDateTime
 from obspy.signal.trigger import aic_simple
-
+from obspy.signal.trigger import plot_trace
```
megies, line 8: you can import both on one line =)

```python
from obspy.signal.trigger import aic_simple, plot_trace
```

megies (changes requested):
This is mostly looking alright, we need to check out the docs build though when CI is done with that (https://docs.obspy.org/pr/).

What I can see is missing is tests, there needs to be a test for energy ration and modified ER routine.

Also, you might want to add these routines to the [trigger method plugin registry](https://github.com/obspy/obspy/blob/c1a9c0f9883b89f142423f1e9062b7e76a9a47ac/setup.py#L530) and accordingly add to docstring of `Stream/Trace.trigger()`

megies:
Oh and this will have to be rebased on master, mostly just because of the usual merge conflict in changelog, I believe. Let me know if you need help with that part.

dizcza:
> What I can see is missing is tests, there needs to be a test for energy ration and modified ER routine.

I looked at how you perform tests on other functions in the module and found nothing save for running the doc examples. So I did the same. Write at least a few tests to cover some of the functionality there to let other devs know how they should test new functions. What you're asking is to create a new file to start writing tests while all other functions are not tested at all. Or am I missing something and you do have tests and I did overlook them?

d-chambers:
Hey @dizcza,

The trigger tests are located in obspy/signal/tests/test_trigger.py. I would recommend, however, that you create a new test case for each of your triggers so everything isn't lumped into one giant test case as it is now. So you might have: `EnergyRatioTestCase` and `ModifiedEnergyRatioTestCase` classes that subclass `unittest.TestCase`. You can then see how other triggers are tested in the giant `TriggerTestCase` and follow the same patterns. Eventually, we will migrate the testing style for this file from the old unittest to pytest but you don't need to worry about that right now. 

Generally, you will find a tests directory for each high-level ObsPy module.

d-chambers:
No worries, ObsPy is a big project. It is easy to miss things. 

The tests should be straight-forward to write based on the other examples in that file but don't hesitate to ask if you have other questions.

megies:
@d-chambers is there any drawback to just putting test functions in the file without any class wrapped around it? pytest should pick up either, right?

d-chambers:
> @d-chambers is there any drawback to just putting test functions in the file without any class wrapped around it? pytest should pick up either, right?

Yup, just personal preference. It helps keep things organized to group tests by the functionality they apply to  with classes. "Namespaces are a honking good idea."

d-chambers (approved):
I didn't go over the method in detail but the code/tests looks good.

d-chambers:
A few minor things about the docs I missed. In [the new doc build](https://docs.obspy.org/pr/mer/tutorial/code_snippets/trigger_tutorial.html#energy-ratio), for both the energy and modified energy ratios the code to produce the figures in those sections is not shown. The code blocks are shown for the other trigger sections. Can you add the code to produce the figures to the trigger_tutorial.rst file? 

It should look like this:

image

rather than the missing code block like this: 

image

Also, add the new methods to the autosummary under the available methods section like this:

```rst
.. autosummary::
    :toctree: ../../packages/autogen

    ~obspy.signal.trigger.recursive_sta_lta
    ~obspy.signal.trigger.carl_sta_trig
    ~obspy.signal.trigger.classic_sta_lta
    ~obspy.signal.trigger.delayed_sta_lta
    ~obspy.signal.trigger.z_detect
    ~obspy.signal.trigger.pk_baer
    ~obspy.signal.trigger.ar_pick
    ~obspy.signal.trigger.energy_ratio
    ~obspy.signal.trigger.modified_energy_ratio
```
So they show up in the table.

dizcza:
Done both things.

megies:
> Yup, just personal preference. It helps keep things organized to group tests by the functionality they apply to with classes. "Namespaces are a honking good idea."

Ah yeah, but no need for new test classes to inherit from `unittest.TestCase` can just be a plain class, I guess? Ideally we want that thing gone and not have more of it :wink:

obspy/signal/trigger.py
```diff
@@ -303,12 +303,16 @@ def energy_ratio(a, nsta):
 
     .. seealso:: [Han2009]_
     """
+    if nsta > len(a) // 2 or nsta == 0:
+        # Half forward, half backward -> empty medium
+        # If nsta is zero, the sum is undefined -> return zero
+        return np.zeros(len(a), dtype=np.float64)
```
megies, line 311: Shouldn't `nsta=0` be considered nonsensical input and rather raise a `ValueError` maybe? Any point to allowing `nsta` to be zero? @dizcza

megies: Actually, I fell the same about `nsta` being longer than half the signal, I think that also should raise a `ValueError`, which would make it much easier for people to wrap a `try/except ValueError` around their code in case of weird signal input (like gappy stream with very short data parts), although I guess this could be treated on a higher level, checking e.g. traces in a stream before running the trigger on them.

dizcza: > Any point to allowing nsta to be zero?

No.

Perhaps, indeed raising an error would be better.

megies: Some of these routines are from the very early days, so there can be code that is close to 14 years old, so not everything what's in there is 100% up to par. Where looking at new code much more closely nowadays, for the most part. :smile:

Issue #2882, opened 2021-09-22
Memory leak through creation of "hidden" lists of strings in long-running application


qdbp:
Obspy version: 1.2.2

I have a long-running application that processes real-time seismic data, and has a memory leak which I've traced to the way ObsPy record history through `_add_processing_info`.

I have a single Trace to which I append new data, perform some statistics, then trim it down. However, every time I trim it down, `_add_processing_info` appends a string to the statistics:

```'ObsPy 1.2.2: trim(endtime=None::fill_value=None::nearest_sample=True::pad=False::starttime=UTCDateTime(2013, 11, 15, 4, 24, 4, 4500))'```

This causes memory to leak overtime in a very non-obvious way, since this is not part of the obspy public API (for instance, "processing" returns no hits on https://docs.obspy.org/packages/autogen/obspy.core.trace.Stats.html). It took me quite a while to hunt this issue down, and even more effort to attribute it `_add_processing_info` since for some reason `gc` does not pick up back-references for these lists correctly.

I believe a warning if this list grows past a certain amount should be issued, or a flag be documented as to how to turn this functionality off.

**Comment by qdbp** [comment]
I see this is done as part of `_add_processing_info` -- I want to make sure I'm not making a mistake somewhere with this before reopening it as an ObsPy problem.

**Comment by qdbp** [comment]
I've reopened this issue with an updated description.

**Comment by ThomasLecocq** [comment]
You can 

```python
for tr in st:
    tr.stats.processing = []
```
to prevent the processing info to be stored

qdbp:
Yes, I ended up doing that. However, I still think that this is undocumented behavior that can cause hard-to-discover leaking memory merits at least documentation. Personally, I consider the fact that this processing recording is on by default to be a strange design choice.

ThomasLecocq:
Thank you for your comment. Please suggest a Pull Request for change in the documentation and/or to the logic.

megies:
I agree, we should probably set some limit on how many entries processing can have at maximum and then either..
 - let it grow larger still but at least the warning is shown
 - stop appending to processing info
 - make processing roll over in FIFO fashion

megies:
fixed in above PR

PR #3186, opened 2022-10-18
limit length of list which records processing info

trichter:
### What does this PR do?

Limit length of list, which records processing info, to 100 entries. Also emits a warning when reaching this limit.

Fix for #2882

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2721,3 +2721,14 @@ class TestTrace:
 
         assert tr.stats.sampling_rate == 30
         assert tr.data.shape[0] == 1
+
+    def test_long_processing_list(self):
+        """
+        issue 2882
+        """
+        tr = read()[0]
+        with pytest.warns(UserWarning, match='.*maximal length') as record:
+            for _ in range(110):
+                tr.trim(0.01)
+        assert len(tr.stats.processing) == 100
+        assert len(record) == 1
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -2727,8 +2727,12 @@ class TestTrace:
         issue 2882
         """
         tr = read()[0]
-        with pytest.warns(UserWarning, match='.*maximal length') as record:
-            for _ in range(110):
+        for n in (100, 5, 10):
+            if n != 100:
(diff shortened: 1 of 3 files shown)


obspy/core/trace.py
```diff
@@ -2298,7 +2298,13 @@ def _internal_add_processing_info(self, info):
         trace's :class:`~obspy.core.trace.Stats` object.
         """
         proc = self.stats.setdefault('processing', [])
-        proc.append(info)
+        if len(proc) == 99:
```
megies, line 2301: I'd propose to add this as a class attribute on `Trace`, e.g. `_max_processing_info`. That way people are in full control of this and can set it to other values.

obspy/core/tests/test_trace.py
```diff
@@ -2721,3 +2721,14 @@ def test_resample_short_traces(self):
...
         assert tr.data.shape[0] == 1
+
+    def test_long_processing_list(self):
+        """
+        issue 2882
+        """
+        tr = read()[0]
+        with pytest.warns(UserWarning, match='.*maximal length') as record:
+            for _ in range(110):
+                tr.trim(0.01)
```
megies, line 2732: you could set `tr.stats.processing = [''] * 99` and only do the one call emitting the wanring in the context manager

trichter: Good idea, both implemented.

obspy/core/trace.py
```diff
@@ -2298,7 +2299,12 @@ def _internal_add_processing_info(self, info):
         trace's :class:`~obspy.core.trace.Stats` object.
         """
         proc = self.stats.setdefault('processing', [])
-        proc.append(info)
+        if len(proc) == self._max_processing_info-1:
```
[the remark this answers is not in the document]
trichter: Then the warning would be raised for every operation on trace after the limit is reached. Now it is only raised once when the limit is reached.

I think people tampering with this stuff do not need any warning at all.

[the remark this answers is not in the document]
trichter: I'd prefer to leave it as it is now.

megies (approved):

Issue #2794, opened 2021-03-01
A bug for MassDownloader()


core-man:
**Codes**

I used the following code to download seismic waveforms at station GE.CSS from GFZ data center.

```Python
from obspy.core import UTCDateTime
from obspy.clients.fdsn.mass_downloader import CircularDomain, Restrictions, MassDownloader

latitude, longitude = 33.901, 35.519
origin = UTCDateTime("2020-08-04T15:08:18.000Z")
starttime, endtime = origin - 60, origin + 6 * 60

domain = CircularDomain(latitude=latitude, longitude=longitude, minradius=0.0, maxradius=5.0)
restrictions = Restrictions(
    starttime=starttime,
    endtime=endtime,
    network="GE",
    station="CSS",
    reject_channels_with_gaps=False,
    minimum_length=0.0,
    sanitize=False,
    minimum_interstation_distance_in_m=0.0,
    channel_priorities=('HH[ZNE12]', 'HL[ZNE12]', 'BH[ZNE12]', 'BL[ZNE12]'),
)

mdl = MassDownloader(providers=["GFZ"])
mdl.download(domain, restrictions, mseed_storage="mseed", stationxml_storage="stations")
```

No data is downloaded:
```bash
[2021-03-01 20:16:40,602] - obspy.clients.fdsn.mass_downloader - INFO: Initializing FDSN client(s) for GFZ.
[2021-03-01 20:16:40,608] - obspy.clients.fdsn.mass_downloader - INFO: Successfully initialized 1 client(s): GFZ.
[2021-03-01 20:16:40,611] - obspy.clients.fdsn.mass_downloader - INFO: Total acquired or preexisting stations: 0
[2021-03-01 20:16:40,612] - obspy.clients.fdsn.mass_downloader - INFO: Client 'GFZ' - Requesting unreliable availability.
[2021-03-01 20:16:41,071] - obspy.clients.fdsn.mass_downloader - INFO: Client 'GFZ' - Successfully requested availability (0.46 seconds)
[2021-03-01 20:16:41,079] - obspy.clients.fdsn.mass_downloader - INFO: Client 'GFZ' - Found 1 stations (3 channels).
[2021-03-01 20:16:41,080] - obspy.clients.fdsn.mass_downloader - INFO: Client 'GFZ' - Will attempt to download data from 1 stations.
[...]

PR #3188, opened 2022-10-19
Fix massdownloader channel priority missing data

megies:
### What does this PR do?

fdsn massdownloader: make channel_priority robust to advertized but missing data 
 
currently (at least when only "weak" availability data is available), any channels that are in principle available according to metadata overrule all other channels that come later in the channel_priority listing, e.g. if the the first item in channel_priority successfully matches a channel, all other channels are ignored, even if the former selected channel actually yields "No data available at server" while some other channels actually do have data but are later in the channel_priority (see #2794)  
 
Currently the only way to fix this is to first try and download *all* channels' data that match any of the given channel_priority wildards, and then at the very end it is evaluated if some higher priority data was downloaded and lower priority data get deleted again.  
 
This certainly is not ideal, since it might blow up the amount of data downloaded and subsequently discarded, but it is likely the lesser evil compared to losing whole stations from the final download result  

*Please fill in (link relevant issues with "see #123456", mark issues that get resolved with "fixes #12345")*

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/download_helpers.py
+++ b/obspy/clients/fdsn/mass_downloader/download_helpers.py
@@ -49,6 +49,15 @@ class _SlotsEqualityComparisionObject(object):
                     for _i in self.__slots__])
 
 
+def _sta_int_key(station, interval):
+    """
+    Returns a tuple that can be used as a key in dictionaries to identify a
+    unique combination of Station+Interval
+    """
+    return (station.network, station.station, interval.start.ns,
+            interval.end.ns)
+
+
 class Station(_SlotsEqualityComparisionObject):
     """
     Object representing a seismic station within the download helper classes.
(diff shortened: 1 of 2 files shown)


megies:
CC @core-man

megies:
An alternative approach could be to have `Massdownloader.download` call itself recursively, iterating through the priorities and skipping station+interval combinations if downloaded data is present in the output. Not sure if I get to trying this out though.

megies:
This isn't an ideal solution yet, I'm tempted to just show a warning about this behavior when `channel_priorities` is used and postpone this to a later release..

megies:
Showing a warning actually makes no sense, since channel priorities has a channel list as a default, so the warning would show pretty much all the time.

Properly fixing this would take a major refactoring, I fear, so I'll postpone this to a later release, since this PR likely causes a serious amount of additional downloading of data that is later deleted again.

To properly fix this the downloading part should be refactored I think so that the hierarchy isnt "Station > Channel > Interval" but rather "Station > Interval > Channel" and then at the interval level one could keep track of whether data has been downloaded while iterating through the channel priorities and eventually stopping when data was found, rather than the original approach of filtering out channels based on station metadata and channel priorities before starting any downloading.

As stated above that means a major rewrite, which I don't have time for right now. Anybody willing to tackle this, feel free to take a shot at it.

Issue #3193, opened 2022-10-27
Adding publicID of seiscomp inventories to station object


s-schneider:
### Problem

When reading seicomp inventory xml files, the unique station identifier is not saved as an attribute of the station object.
It would be handy to have that id saved, to keep track of stations in the inventory that changed the value of `station.code`

### Proposed solution

Either add a `publicID` attribute to BaseNode and further add something like this to `_read_station` in `obspy.io.seiscomp.inventory`

"""
publicID = sta_element.get("publicID")
station = obspy.core.inventory.Station(code=sta_element.get("code"),
                                           latitude=latitude,
                                           longitude=longitude,
                                           elevation=elevation,
                                           publicID = publicID)
"""
and `stream.select(publicID =MY_ID)` for easy selection.

Or save the publicID in another attribute of station? maybe `source_id`?

PR #3194, opened 2022-10-27
Added publicID to station

s-schneider:
### What does this PR do?
Adds public_id to station objects, if station is read from inventory files from seiscomp.

see #3193 
https://github.com/obspy/obspy/issues/3193

### PR Checklist
- [ x If the PR is making changes to documentation, docs pages can be built automatically.
- [ ] All tests still pass.
- [ x If the changes affect any plotting functions you have checked that the plots


d-chambers:
Alright @s-schneider, the CI is working again so the failures you see here are probably related to your changes. Take a look and see if you can figure out how to fix them.

megies:
Hmm.. our inventory type classes are kept really close to the StationXML schema. Anything that diverges from it could potentially be trouble in the future, when we might have to adjust to changes in StationXML, or just be confusing to people because there are things that are not in StationXML or are missing when writing to StationXML. So, I'm not sure how I feel about this one.. It's quite a harmless looking change obviously, but it does have the above implications..

megies:
Would it be an option to keep this SCML specific piece of information in the [`extra` part of the station objects](https://docs.obspy.org/master/tutorial/code_snippets/stationxml_custom_tags.html) maybe?

s-schneider:
Hm if we can add that field to the `select` method of inventories and networks, that would be a possible solution

megies:
> Hm if we can add that field to the `select` method of inventories and networks, that would be a possible solution

I think I would prefer that, and I don't see a big problem having it in `select()` if the docstring for it makes clear that it is strictly a seiscomp thing and wont do anything in most cases.

s-schneider:
Should be no problem, I’ll change that and push an update

s-schneider:
[EB_response_NEW_stationXML.zip](https://github.com/obspy/obspy/files/10067189/EB_response_NEW_stationXML.zip)
I actually just experienced something weird while trying to fix the tests. If I take that example file, read it and write it again, it it adds a description to the response stage input units

```
  <Stage number="1">
    <PolesZeros name="EBR.2002.091.HE" resourceId="ResponsePAZ#20121207153142.208772.15386">
      <InputUnits>
        <Name>M/S</Name>
        <Description>None</Description>
      </InputUnits>
```
vs
```
  <Stage number="1">
    <PolesZeros name="EBR.2002.091.HE" resourceId="ResponsePAZ#20121207153142.208772.15386">
      <InputUnits>
        <Name>M/S</Name>
      </InputUnits>
```

The code I used:
```
from obspy import read_inventory

file = "EB_response_NEW_stationXML"
newfile = "EB_response_NEW2_stationXML"
stationxml_inventory = read_inventory(file, format="STATIONXML")
nsmap = {
    'scml': 'seiscompml', 
    'resp': 'http://geofon.gfz-potsdam.de/ns/seiscomp3-schema/0.7'
}
stationxml_inventory.write(newfile, format="STATIONXML", nsmap=nsmap)
```

megies:
@s-schneider I've rebased this branch on current master and force pushed, to avoid merge conflicts. Can you continue work on this version? Or let me know if that's a problem for you.

obspy/io/seiscomp/inventory.py
```diff
@@ -267,10 +269,20 @@ def _read_station(instrumentation_register, sta_element, _ns):
...
     elevation = _read_floattype(sta_element, _ns("elevation"), Distance,
                                 unit=True)
+    public_id = sta_element.get("publicID")
+    ns = "seiscompml"
+    extra_tag = AttribDict({})
+    extra_tag.public_id = AttribDict({
+            'value': public_id,
+            'namespace': ns
+        }
+    )
```
megies, line 279: There are a few things that should be changed here I think.

#### use namespaces correctly

Namespace should be set to the actual namespace value of the attribute `publicID` that is being parsed and not quite arbitrarily `"seiscmopml"`. There is an easy way that might be doing the right thing..

```
namespace = sta_element.nsmap[None]
```

... but this is working on assumptions how namespaces are set in sc3ml and might break later on.
So to correctly determine the namespace of this attribute, you should check the attribute for a namespace first and if it doesn't have one it should be OK to use the default namespace of it's element

```python
match = re.match(r'{([^}]*)}.*', sta_element.tag)
if match:
    namespace = match.group(1)
else:
    namespace = sta_element.nsmap[None]
```

I haven't messed around with details of lxml and xml namespaces in a while, but it seems that the default namespace is set on the station element, even though the whole document is operating on the implicit default namespace declared on the root element, so it should be OK to assume we can use that.

Actually, as a last thought.. when accessing `public_id = sta_element.get("publicID")` that is already assuming that `publicID` has the default namespace of the station item, I believe, so maybe just using station default namespace is fine..

#### naming of extra item

I would probably keep the name `publicID` as the name of the item in extra. Even though that means internally you would look at `station.extra.publicID` which isn't in line with our Python object naming scheme.. this a) makes it clear that this is not coming from Python land and b) if writing an inventory read from sc3ml as stationxml, your custom attribute would kind of properly show up in the xml file as something like..

```xml
<FDSNStationXML xmlns="http://www.fdsn.org/xml/station/1" schemaVersion="1.2">
  <Source>scxml import</Source>
  <Sender>ObsPy Inventory</Sender>
  <Module/>
  <ModuleURI/>
  <Created>2022-12-01T12:14:43.872676Z</Created>
  <Network code="EB" startDate="1980-01-01T00:00:00.000000Z" restrictedStatus="open">
    <Description>SINGLE STATION</Description>
    <Station xmlns:ns0="seiscompml" code="EBR" startDate=..... ns0:public_id="Station/EB/EBR/2002-04-01T00:00:00.0000Z">
      <Latitude unit="DEGREES">40.820599</Latitude>
      <Longitude unit="DEGREES">0.4933</Longitude>
...
```

#### should set type as attribute

Then it would come out as an attribute on StationXML output as opposed to an element.

obspy/io/seiscomp/tests/test_inventory.py
```diff
@@ -75,13 +75,19 @@ def test_compare_xml(self):
         to stationXML format and compare line by line
         """
         sc3ml_bytes = io.BytesIO()
-        self.sc3ml_inventory.write(sc3ml_bytes, "STATIONXML")
+        self.sc3ml_inventory.write(
+            sc3ml_bytes, "STATIONXML",
+            nsmap={'my_ns': 'seiscompml'}
```
megies, line 80: Should use a proper sc3ml namespace URI

megies (changes requested):

obspy/core/inventory/network.py
```diff
@@ -465,6 +468,19 @@ def select(self, station=None, location=None, channel=None, time=None,
                 if not fnmatch.fnmatch(sta.code.upper(),
                                        station.upper()):
                     continue
+
+            if public_id is not None:
+                if hasattr(sta, 'extra'):
+                    if hasattr(sta.extra, 'publicID'):
+                        if not fnmatch.fnmatch(
+                            sta.extra['publicID'].value.upper(),
+                            public_id.upper()
```
megies, line 477: I'm not sure if `publicID` in sc3ml is case sensitive, but personally, I'd probably leave the `upper()` away. I know it's done like this in some other comparisons, and I'm not sure if I agree with that but it is what it is and thats more coming from the SEED perspective where everything was supposed to be all uppercase characters only anyway.

megies:
I pushed some of the adjustments, if you wanna work on the rest?

megies:
CC @filefolder since you worked on sc3ml stuff recently

filefolder:
> CC @filefolder since you worked on sc3ml stuff recently

sure, but still out digging holes for another week

megies:
trying to wrap this up for the release @s-schneider @filefolder, maybe you can have a look, I think there was some last things to do

There were some merge conflicts, so need to double-check, most notably 0.13 read support was added in the meantime, so that needs checking if that changes anything

ThomasLecocq:
@filefolder your expertise is needed here to check this PR (or to bump it to the "Next Release" milestone)

filefolder:
bumped to 1.5.1 as i get the sense this will require a bit of a teardown/rebuild after the 0.14 PR goes through

Issue #3199, opened 2022-11-03
CSV/CSZ event reader breaks ObsPy


trichter:
### Avoid duplicates

- [X] I searched existing issues

### Bug Summary

I managed to break ObsPy by installing a custom event reader plugin.

It can be installed with 

```
pip install git+https://github.com/trichter/obspycsv.git
```

Running `obspy-runtests` results in several fails related to BytesIO objects: https://tests.obspy.org/129345/

### Code to Reproduce

```python
# first test fail
import io
from obspy import UTCDateTime, read_events
from obspy.core.event import Event, Origin
origin = Origin(time=UTCDateTime('2016-01-01'))
event1 = Event(origins=[origin])
bio = io.BytesIO()
event1.write(bio, 'quakeml')  # its working with a real file
event2 = read_events(bio)  # with or without specifying quakeml format!
assert len(event2) == 1
```

### ObsPy Version?

current master

### Operating System?

Lubuntu 18.04

### Python Version?

3.8.6

### Installation Method?

conda

trichter:
Here is the traceback  when changing to `event2 = read_events(bio, 'QUAKEML')`

```
AttributeError                            Traceback (most recent call last)
~/dev/obspy/obspy/io/quakeml/core.py in _xml_doc_from_anything(source)
     76             try:
---> 77                 xml_doc = etree.fromstring(source.encode())
     78             except Exception:

AttributeError: '_io.BytesIO' object has no attribute 'encode'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-5-732540130d2f> in <module>
----> 1 event2 = read_events(bio, 'QUAKEML')

<decorator-gen-154> in read_events(pathname_or_url, format, **kwargs)

~/dev/obspy/obspy/core/util/decorator.py in _map_example_filename(func, *args, **kwargs)
    289                         except IOError:
    290                             pass
--> 291         return func(*args, **kwargs)
    292     return _map_example_filename
    293 

~/dev/obspy/obspy/core/event/catalog.py in read_events(pathname_or_url, format, **kwargs)
    808         return _create_example_catalog()
    809     else:
--> 810         return _generic_reader(pathname_or_url, _read, format=format, **kwargs)
    811 
    812 

~/dev/obspy/obspy/core/util/base.py in _generic_reader(pathname_or_url, callback_func, **kwargs)
    623             # first try reading directly
    624             print('try reading bytes')
[...]

Issue #3192, opened 2022-10-25
Pop check_compression from argument list for readers?


trichter:
I wrote a small io plugin for ObsPy events based on zipped files, see https://github.com/trichter/obspycsv. Because ObsPy automatically unpacks zip files, I had some difficulties to get it working.
I found the check_compression argument in the uncompress_file decorator with which its working fine. I think, however, that it should be popped from the argument list [here](https://github.com/obspy/obspy/blob/master/obspy/core/util/decorator.py#L139). Otherwise:

```py
In [1]: ev = read_events()
In [2]: ev.write('test.xml', 'QUAKEML')
In [3]: read_events('test.xml', check_compression=False)
TypeError: _read_quakeml() got an unexpected keyword argument 'check_compression'
```

Ideally, a plugin could define on its own if the compression check should be skipped, e.g. by setting an additional entry point. I see, however, that this feature needs quite some refactoring of the reader code.

trichter:
> I think, however, that it should be popped from the argument list [here](https://github.com/obspy/obspy/blob/master/obspy/core/util/decorator.py#L139)

This is fixed now.

> Ideally, a plugin could define on its own if the compression check should be skipped, e.g. by setting an additional entry point. I see, however, that this feature needs quite some refactoring of the reader code.

This is not fixed, but I do not want to tackle it on my own. The `check_compression` work-around is good enough for me at the moment.

Therefore, I close the ticket for now.

trichter:
> > Ideally, a plugin could define on its own if the compression check should be skipped, e.g. by setting an additional entry point. I see, however, that this feature needs quite some refactoring of the reader code.
> 
> This is not fixed, but I do not want to tackle it on my own. The `check_compression` work-around is good enough for me at the moment.

I have a new simple idea how to fix my problem. @megies In [decorator.py](https://github.com/obspy/obspy/blob/master/obspy/core/util/decorator.py#L167-L168) can we check for a comment in the zip file to prevent the unpacking?

```
            with zipfile.ZipFile(filename) as zip:
                if b'obspyarchive' in zip.comment.lower():
                    obj_list = None
                else:
                    obj_list = [zip.read(name) for name in zip.namelist()]
```

megies:
I'm not sure I get what the problem or above solution actually are..

It should just be possible to use `check_compression=False`, if that's not working for some reason (like `_read_quakeml()` protesting on unrecognized kwargs?) then we should just fix that I think?

trichter:
[...]

PR #3204, opened 2022-11-03
Fix reading some bytesio objects in tests

trichter:
### What does this PR do?

Fix for #3199

Another change introduced in the PR is to pop the ckeck_compression flag in the reader routines, see #3192

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/tests/test_event.py
+++ b/obspy/core/tests/test_event.py
@@ -487,6 +487,8 @@ class TestCatalog:
         event1 = Event(origins=[origin])
         bio = io.BytesIO()
         event1.write(bio, 'quakeml')
+        # rewind bytes stream
+        bio.seek(0)
         # read from disk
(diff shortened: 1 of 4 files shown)


megies:
OK, so basically this is a non issue with just sloppy test code that should always have had a `seek(0)` there and that for some weird reason used to work, but that custom plugin passing `_is_obspycsv` check on an empty file prevented that weird magic safety net to come in effect.

That's a really funny string of events, happy that there doesn't seem to be an actual big problem :rofl:

megies (approved):
Not sure what that get -> pop change is about, but I'm assuming this is to not leak that kwarg to underlying reader plugins :+1:

trichter:
> Not sure what that get -> pop change is about, but I'm assuming this is to not leak that kwarg to underlying reader plugins

Yes, exactly!

PR #3212, opened 2022-11-11
get_stations_bulk params / turn off get_stations autodiscovery

### LINKED ISSUE #3140: fdsn.client: add query parameters to get_station_bulk()

filefolder:
this is the same PR as #3173 which I was unable to sync with master without starting over (?)

it currently has the get_stations_bulk parameters defined as below, which are jumbled but non-breaking, as discussed  

```
    def get_stations_bulk(self, bulk, level=None, includerestricted=None,
                          includeavailability=None, filename=None,
                          minlatitude=None, maxlatitude=None,
                          minlongitude=None, maxlongitude=None, latitude=None,
                          longitude=None, minradius=None, maxradius=None,
                          updatedafter=None, matchtimeseries=None, format=None,
                          **kwargs):
```

forgive the delays!

fixes #3140 
supersedes #3173

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,8 @@ Changes:
    * limit length of list which records processing info (see #2882)
  - obspy.clients.fdsn:
    * update URL for NCEDC to https (see #3203)
+   * add all valid parameters to get_stations_bulk (see #3140)
+   * do not use autodiscovery of file format in get_stations (see #3138)
  - obspy.clients.seishub:
    * submodule removed completely, since it is outdated and not even test
      servers have been running for years (see #2994)
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -1062,12 +1062,12 @@ class Client(object):
             self._attach_dataselect_url_to_stream(st)
             return st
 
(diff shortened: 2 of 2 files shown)


megies (approved):
thanks for handling this :rocket:

PR #3216, opened 2022-11-15
updating seiscomp xsl/xsd files 

filefolder:
just updating the files currently included in obspy with current (as of yesterday) editions at seiscomp/common

this one https://github.com/SeisComP/common/pull/74 in particular may speed up SC > QML conversions quite significantly but the benefits may be outside the realm of most typical obspy use cases. all credit to @acarapetis

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -75,9 +75,8 @@ Changes:
      from an old SAC header, if one exists. (see #2760)
    * fix writing SACPZ when a channel's sensor filed is unset (see #3187)
  - obspy.io.seiscomp:
-   * update seiscomp 0.11 and 0.12 schema to potentially speed up sc3ml to qml
-     event catalog conversions (see https://github.com/SeisComP/common/pull/74)
-   * add Z suffix to xz:dataTime elements in 0.11 for SC < 5 compatiblity 
+   * update SeisComP 0.11 and 0.12 schema to potentially speed up sc3ml to qml
+     event catalog conversions (see #3216 and SeisComP/common#74)
  - obspy.signal:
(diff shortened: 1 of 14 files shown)


megies:
If it's not too much hassle, could you try to avoid merging master branch into your feature branch, but instead do "git pull; git rebase origin/master" instead (changes history of your branch so needs a "--force" for pushing the result)? That way we dont get these ugly "merge branch master" commits (it should also get rid of them if you do it now or later).
I don't think that for you it makes much difference, you have to resolve a merge conflict either way (assuming it was just a conflict in the changelog).

If you're not comfortable with that, that's fine though and you can just do as before.

filefolder:
ahh! yes sorry I will certainly start doing that. as you can tell i barely know how to use git so I mostly do it via web.

megies:
> ahh! yes sorry I will certainly start doing that. as you can tell i barely know how to use git so I mostly do it via web.

All good, you can also keep doing as you do right now, no big deal really, just cosmetics.

megies:
rebased, force pushed and changed base branch to `release_1.4.0`

filefolder:
fwiw i think it's all set to go but not too important for 1.4. happy to let it stew until the next minor update

megies:
> fwiw i think it's all set to go but not too important for 1.4. happy to let it stew until the next minor update

Nah, seems fine to me, will merge for 1.4.0. If there really are problems, can be handled in minor releases.

megies:
@filefolder just noticed you weren't in developers team, sent you an invite, that way you can also push branches to the main repo, which makes it easier if somebody wants to add changes to your PRs.

PR #3218, opened 2022-11-15
response: do not report unknown units as type "displacement" to evalresp

### LINKED ISSUE #2945: Default instrument set to DISP if not known but VEL is simulated
Hi - it's not directly an issue but an inconvenient default setting. We are working more and more with rotational motion sensors which physical unit is rad/s. As we ask the manufacturers of the instrument to send stationXML or NRL lookups together with the equipment it would be best to include the rad/s and rad physical units also into obspy. This means, however, that the list of intputs and outputs in the obspy remove_response routine needs to be extended to rotation rate and rotation (if we want to be completely precise). 
A current "flaw" in obspy remove_response default settings is (at least for me) that if the instrument type is not known, it will assume its is a DISP sensor and simulates then as default a VEL sensor. I think  this is not a good strategy even when the first step IS mentioned in the docs. I prefer to have both defaults in the same physical space, which would result in less surprises - at least that's what I have done in my local copy. Cheers Jo

megies:
### What does this PR do?

Do not report unknown / unrecognized units / units not representable in evalresp to evalresp as type "displacement", as this leads to e.g. input units of radian/second on a rotational sensor with using the unchanged default of `output="VEL"` to have a differentiation added on top of the response stated in metadata, which makes no sense.
If unknown units come into play, just report to evalresp as "undefined" which seems to consistently just use the response from metadata as is, essentially ignoring "output" parameter.

Seems unproblematic and I think it makes total sense. No existing tests had to be adjusted, since I'm assuming they all are in the realm of DIS/VEL/ACC.

Fixes #2945 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,11 @@ Changes:
    * Minimum supported Python version is now 3.8 (see #3081, #3087)
    * Minimum supported dependency versions are now numpy 1.20, scipy 1.7 and
      matplotlib 3.3 (see #3081, #3087)
+   * response: avoid reporting all unkown units (e.g. "RAD/S") to evalresp as
+     displacement ("DIS") which with default value for "output" ("VEL") leads
+     to evalresp adding a differentiation during response calculation. Instead
+     report undefined unit, which makes evalresp use response as is and
+     essentially ignoring "output" parameter (see #2945)
    * Improved expanded channel information in string representation of Station,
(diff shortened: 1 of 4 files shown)


trichter (approved):
The changes seem sensible. Though, I am not that familiar with evalresp.

megies:
CC @jwassermann will be in 1.4.0

PR #3220, opened 2022-11-16
skip_on_gaps=True & subpplot spectrogram option

FMassin:
### What does this PR do?

- PPSD init: Changes the default skipping segments with gaps from False to True, so it does not accumulate ~-3000 db values, and plots looks correct (and not [completely clipped as in default spectrogram with gaps](http://mercalli.ethz.ch/~fmassin/VolcSAT/seismo-viz/2022/11/15/GI.STG10.00.HHZ.spec.png))
-  PPSD spectrogram: allows plotting in a given axes, with extra option [for subplot adjustments](http://mercalli.ethz.ch/~fmassin/VolcSAT/seismo-viz/2022/11/16.spec.png)).

The default  gap interpolation would produce  ~-3000 db values, affecting plots and interpretation.

### PR Checklist
- [X ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [X] While the PR is still work-in-progress, the `no_ci` label can be added to skip CI builds
- [ ] All tests still pass.
- [X] Any new features or fixed regressions are covered via new tests.
- [X] Any new or changed features are fully documented.
- [X] If the changes affect any plotting functions you have checked that the plots
- [X] Add the `ready for review` tag when you are ready for the PR to be reviewed.


FMassin:
Hello

Thanks for taking care of this. I am not a PQLX user, but I have a lot of data with gaps (some for good reason, some not), the idea of interpolating gaps with zeros (the current default behaviour of obspy PPSD) never seems like a good idea to my honest opinion. I would at least add a very visible warning in obspy documentation and on the plot. 

> Amplitude vs. Frequency, binned histogram plot, where the gappy PSD slices show up as erratic lines and can still be interpreted

Well right now the code silently interprets no data as amplitude zero and sets the related PSD values to ~-3000db. There is no warning telling the user how to interpret the resulting plot. 

> Even your own example has time on x-Axis
Here's an example:
http://mercalli.ethz.ch/~fmassin/VolcSAT/seismo-viz/2022/11/16/GI.STG9.00.HHZ.spec.png

> Do we really need this "flip" option that adds so much complexity to the code?
This way, it's easier to compare with an helicorder, e.g.:
http://mercalli.ethz.ch/~fmassin/VolcSAT/seismo-viz/2022/11/16/GI.STG9.00.HHZ.png

megies:
> the idea of interpolating gaps with zeros (the current default behaviour of obspy PPSD) never seems like a good idea to my honest opinion. I would at least add a very visible warning in obspy documentation and on the plot.

It is not a good idea in terms of meaningful signal processing, fully agree. It is meaningful in a different way, though, as it visually displays problematic features in the data, see BSSA, Vol. 94, No. 4, pp. 1517–1527 "Ambient Noise Levels in the Continental United States" by Daniel E. McNamara and Raymond P. Buland.



> right now the code silently interprets no data as amplitude zero and sets the related PSD values to ~-3000db. There is no warning telling the user how to interpret the resulting plot.

Hmm need to look at that, not sure how that would happen since when adding data to PPSD it goes through the traces and cuts out slices, but I dont think it should do that when there is no data at all. It should just fill up slices when there is some data to start with.

> This way, it's easier to compare with an helicorder [e.g. corresponding to the above](http://mercalli.ethz.ch/~fmassin/VolcSAT/seismo-viz/2022/11/16/GI.STG9.00.HHZ.png).

Hmm yea, I guess we can make do with the added complexity..

FMassin:
> not sure how that would happen

... [merging with 0 filling](https://github.com/obspy/obspy/blob/29f6127e38963e857111a8f9773ad0d14fdf77af/obspy/signal/spectral_estimation.py#L872) is done with method [0 by default](https://github.com/obspy/obspy/blob/29f6127e38963e857111a8f9773ad0d14fdf77af/obspy/signal/spectral_estimation.py#L872), which [means no interpolation, just filling](https://docs.obspy.org/packages/autogen/obspy.core.trace.Trace.html#obspy.core.trace.Trace.__add__) with 0.

megies:
> ... [merging with 0 filling](https://github.com/obspy/obspy/blob/29f6127e38963e857111a8f9773ad0d14fdf77af/obspy/signal/spectral_estimation.py#L872) is done with method [0 by default](https://github.com/obspy/obspy/blob/29f6127e38963e857111a8f9773ad0d14fdf77af/obspy/signal/spectral_estimation.py#L872), which [means no interpolation, just filling](https://docs.obspy.org/packages/autogen/obspy.core.trace.Trace.html#obspy.core.trace.Trace.__add__) with 0.

Ah, I see, that's really long gaps, longer than one individual PSD slice.

FMassin:
Hey, happy new year dear ObsPy dev team :)

I really like to help with this and to see some improvements there. Please let me know if there's anything I could do or reconsider to make this compliant with your expectations.

Cheers

Fred

ThomasLecocq:
@FMassin discovering this PR :-) I love the Axes stuff, and agree that changing the default is not ideal. So I'd suggest to revert that change, but to add a warning as information to the users, when large gaps are zero-filled , that they will receive an output with large negative values, and they can avoid it by setting ``skip_on_gaps=True``

FMassin:
Thanks @ThomasLecocq !  

Would you like this @megies ?

ThomasLecocq:
@FMassin you can go ahead !

Thus:

- [x] rebase on current master 
- [ ] make sure the default remains the same
- [ ] add a warning message if -3000 dB (or np.all(array==0), e.g.) is found to the users that they can avoid that using the new flag
- [ ] add a test case with a gappy file: check both cases (which should have 1 or 0 warnings)
- [ ] add a section in the documentation with your examples above, with and without (e.g. based on the same test file?)
- [ ] add lines to the CHANGELOG
- [ ] get some cool drinks & check the CI pass ✅ 

:-)

tom

PS: and check my PR if you have time #3589 :-)

ThomasLecocq:
I did the update of your branch already :-)

Issue #3148, opened 2022-09-21
Importing Obspy in ipython disables colored output on Windows CMD


samuelchodur:
### Avoid duplicates

- [X] I searched existing issues

### Bug Summary

After importing obspy, my IPython terminal does not print colored output. Tried this in several terminals on Windows (Git Bash and Anaconda Powershell Prompt). I can import other packages and continue to see colored output. Tested with 1.2.2 as well and had same problem. 

![image](https://user-images.githubusercontent.com/752993/191554499-d6a6768a-a264-449f-9a9b-79a8c2ad8460.png)

### Code to Reproduce

```python
ipython
print(dne) # variable dne doesn't exist and prints error in color
import obspy
print(dne) # variable dne doesn't exists and does not print error in color
```

### ObsPy Version?

1.3.0, 1.2.2

### Operating System?

Windows

### Python Version?

3.10

### Installation Method?

conda

filefolder:
working as expected here on linux / python 3.9.5 / ipython 8.30 and osx python 3.9.9 / ipython 7.22

megies:
Can't reproduce. Maybe try updating IPython. Please reopen if needed.

samuelchodur:
> Can't reproduce. Maybe try updating IPython. Please reopen if needed.

I am using IPython version 8.5.0. Are you trying to reproduce the problem on Windows? I don't have issues on Linux.

megies:
Oh this is Windows.. whelp, can't say anything about Windows really.

samuelchodur:
> Oh this is Windows.. whelp, can't say anything about Windows really.

could this get set back to bug-unconfirmed then until someone else comes along with same issue as I am having?

megies:
@samuelchodur what kind of prompt is that then exactly? Ipython started in an Anaconda shell or however that is called? If so what Anaconda version?

@ThomasLecocq ever had any issues like that?

filefolder:
I just tried this on an windows laptop and I **can** reproduce with python 3.10.7, Ipython 8.5.0, obspy 1.3.0. This is when calling ipython from the dos terminal. When running from the IDLE shell, or via the non-ipython terminal, it works normally.

In ipython I even get the same "1;31m" errors just by attempting to import a non-existing obspy module, e.g. `from obspy.geodetics.base import blahblah` so it seems like the whole environment is being contaminated just by browsing. No idea what that's about or what `1;31m` refers to.

megies:
[...]

PR #3229, opened 2022-11-23
Remove some code only necessary for python<3.8

trichter:
### What does this PR do?

Cleanup, see title.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -518,7 +518,7 @@ class Trace(object):
             st += self.copy()
         return st
 
-    def __div__(self, num):
+    def __truediv__(self, num):
         """
         Split Trace into new Stream containing num Traces of the same size.
 
@@ -561,9 +561,6 @@ class Trace(object):
             tstart = tend + self.stats.delta
             tend = tstart + (self.stats.delta * packet_length)
         return st
-
-    # Py3k: '/' does not map to __truediv__ anymore in Python 3
-    __truediv__ = __div__
 
     def __mod__(self, num):
         """
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -562,7 +562,7 @@ class Trace(object):
             tend = tstart + (self.stats.delta * packet_length)
         return st
 
-    # Py3k: '/' does not map to __div__ anymore in Python 3
+    # Py3k: '/' does not map to __truediv__ anymore in Python 3
     __truediv__ = __div__
 
     def __mod__(self, num):
--- a/obspy/core/util/attribdict.py
+++ b/obspy/core/util/attribdict.py
@@ -8,7 +8,7 @@ AttribDict class for ObsPy.
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
-import collections
+import collections.abc
 import copy
 import warnings
 
--- a/obspy/core/util/attribdict.py
+++ b/obspy/core/util/attribdict.py
@@ -8,12 +8,11 @@ AttribDict class for ObsPy.
     GNU Lesser General Public License, Version 3
     (https://www.gnu.org/copyleft/lesser.html)
 """
+import collections
 import copy
 import warnings
 
 import numpy as np
-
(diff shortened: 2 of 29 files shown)


obspy/core/util/attribdict.py
```diff
@@ -123,7 +122,7 @@ def __setitem__(self, key, value):
             value = self._cast_type(key, value)
 
         mapping_instance = isinstance(value,
-                                      compatibility.collections_abc.Mapping)
+                                      collections.abc.Mapping)
```
QuLogic, line 125: Should fit on one line, I think?

obspy/core/trace.py
```diff
@@ -562,7 +562,7 @@ def __div__(self, num):
             tend = tstart + (self.stats.delta * packet_length)
         return st
 
-    # Py3k: '/' does not map to __div__ anymore in Python 3
+    # Py3k: '/' does not map to __truediv__ anymore in Python 3
     __truediv__ = __div__
```
QuLogic, line 566: I don't think this is correct; you probably want to rename the method to `__truediv__` and remove `__div__`.

obspy/core/util/testing.py
```diff
@@ -26,10 +26,9 @@ def compare_xml_strings(doc1, doc2):
     """
     Simple helper function to compare two XML strings.
 
-    :type doc1: str
-    :type doc2: str
+    :type doc1: bytes or str
+    :type doc2: bytes or str
```
QuLogic, line 30: Still kind of a compatibility hack; we should make it one type only and remove everything in the `try`.

obspy/imaging/spectrogram.py
```diff
@@ -165,10 +165,7 @@ def spectrogram(data, samp_rate, per_lap=0.9, wlen=None, log=False,
     halfbin_time = (time[1] - time[0]) / 2.0
     halfbin_freq = (freq[1] - freq[0]) / 2.0
 
-    # argument None is not allowed for kwargs on matplotlib python 3.3
```
QuLogic, line 168: Is Python the problem here or Matplotlib?

obspy/core/util/testing.py
trichter: Yes, it was quick and dirty; the callers use the function in both ways.

obspy/core/trace.py
trichter: Not sure. I didn't even know that this functionality existed.
In the tests it is used in the following way

```
        tr = Trace(data=np.arange(1000))
        st = tr / 5
        assert len(st) == 5
```
but I agree that using only `tr // 5` would make sense.

obspy/imaging/spectrogram.py
trichter: I assumed matplotlib was the problem.

obspy/core/util/testing.py
trichter: Turns out, it some calls the first string was utf-8 and the second binary.

obspy/core/util/misc.py
```diff
@@ -290,55 +290,6 @@ class Output():
...
-            f.encoding,
-            f.errors,
-            f.newlines,
-            f.line_buffering)
-
-    sys.stdin = _reopen_stdio(sys.stdin, 'rb')
-    sys.stdout = _reopen_stdio(sys.stdout, 'wb')
-    sys.stderr = _reopen_stdio(sys.stderr, 'wb')
-
-
```
megies, line 341: @trichter I believe this is still needed, IIRC this is for Win + Py>=3.6
I can't test, because no windows but I had people try without that and I believe it crashes windows, see https://github.com/obspy/obspy/issues/3148#issuecomment-1274254892

obspy/core/trace.py
QuLogic: That one is `__floordiv__`; there is no `__div__` any more.

trichter: Thanks, now I understand.

megies (approved):

PR #3230, opened 2022-12-02
nlloc hyp reading: set "rejected" when nonlinloc reports bad location run and store info in comments

### LINKED ISSUE #2986: NLLoc rejected locations & phases added in nlloc.core
NonLinLoc outputs hyp files for each location, they can be read into an obpsy catalog object ([source code](https://docs.obspy.org/_modules/obspy/io/nlloc/core.html#_read_single_hypocenter)). 
**Currently obspy reads in rejected locations and rejected arrivals in the same way as those that are not rejected.**

obspy 1.2.2, python 3.7.3 - install from conda-forge in an anaconda environment on Ubuntu 18.04.4
using nonlinloc 7.0
Can't find any mention of this in other issues raised previously

# Origins
I have seen two types of locations in hyp files (there may be others). Currently both are processed in the same way:
- a successful location. The first line of each hyp file for a successful location reads something like:
> NLLOC "./locs/test_loc/test.20220101.000000.gri0" "LOCATED" "Location completed."
- a location which was completed but was then rejected for being on the boundary of the search region. The first line of each hyp file for a rejected locations reads something like:
> NLLOC "./locs/test_loc/test.20220101.000000.gri0" "REJECTED" "WARNING: max prob location on grid boundary 10, rejecting location."

I have attached two example hyp files (one for a successful location [test_located.txt](https://github.com/obspy/obspy/files/8084723/test_located.txt) and one for a rejected location [test_rejected.txt](https://github.com/obspy/obspy/files/8084727/test_rejected.txt)).

I don't think the fact that NLLoc rejects the location is currently marked in the created origin in any way - it probably should be? Maybe they shouldn't be included at all.

# Arrivals
The hyp file lists the phases related to the location: both those which are used, and those which are rejected are listed. Similarly to the above, currently obspy reads in all of these as arrivals for each origin. 

The fact that NLLoc rejected the arrival is not specifically marked, but rejected arrivals seem to have:
- azimuth 359.0
- distance 0.0
- takeoff_angle -1.0
- time residual 0.0
- time_weight 0.0

and you can identify the origins with the problem:
`if len(origin.arrivals) == origin.quality.associated_phase_count and len(origin.arrivals) != origin.quality.used_phase_count:`

An example of a hyp file with rejected arrivals can be found in [test_rejected.txt](https://github.com/obspy/obspy/files/8084727/test_rejected.txt) (note: you do also get rejected phases in hyp files for locations which are not rejected).

megies:
### What does this PR do?

Some additions to reading nonlinloc hyp files:
 - set `origin.evaluation_status` to `"rejected"` when nonlinloc reports the location run as "ABORTED", "IGNORED" or "REJECTED"
 - put the "NLLOC" info line (that has the above info in it) as a comment on both the `event` and `origin` objects

The last part of #2986 (rejected arrivals) is not handled any further since QuakeML has no appropriate field like "evaluation status" on arrivals and since these Arrivals are set to weight `0.0` by nonlinloc (in addition to bogus values in most arrival info fields), so there is a way to find out for the user if an Arrival was rejected or not

Fixes #2986 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,10 @@ Changes:
  - obspy.core:
    * response: allow recalculating overall sensitivity even if unit type is not
      one of VEL/ACC/DISP (see #3235)
+ - obspy.io.nlloc:
+   * set origin evaluation status to "rejected" if nonlinloc reports the
+     location run as "ABORTED", "IGNORED" or "REJECTED" (see #3230)
+   * store "NLLOC" info header line in event and origin comments (see #3230)
 
 maintenance_1.4.x
 =================
--- a/obspy/io/nlloc/__init__.py
+++ b/obspy/io/nlloc/__init__.py
@@ -35,7 +35,7 @@ Event:	2006-07-15T17:21:20.195670Z | +51.658,   +7.737
                                    version='NLLoc:v6.02.07')
  preferred_origin_id: ResourceIdentifier(id="smi:local/...")
                 ---------
-            comments: 1 Elements
+            comments: 2 Elements
                picks: 5 Elements
              origins: 1 Elements
 
@@ -54,7 +54,7 @@ Origin
                                    creation_time=UTCDateTime(2013, 6, 21, ...),
                                    version='NLLoc:v6.02.07')
                 ---------
-            comments: 1 Elements
+            comments: 2 Elements
             arrivals: 5 Elements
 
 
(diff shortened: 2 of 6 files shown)


filefolder (approved):
all looks good to me.

one minor question about this part for my own education

```
    if any(nlloc_location_status == status
           for status in ('ABORTED', 'IGNORED', 'REJECTED')):
```
I assume this is the "correct" way to do this.. is it because it is faster than something dumber like

`if nlloc_location_status in ('ABORTED', 'IGNORED', 'REJECTED'):` ?

megies:
> I assume this is the "correct" way to do this.. is it because it is faster than something dumber like

Nah, you're right :rofl: that's a bit stupid. I think I had something like `x in status_line for x in (...)` there first, which would've been more generous in terms of changing whitespaces and quotation signs in that line nonlinloc writes, but then decided to just trust that line to stay the same and really extract that word on the expected position.

megies:
Test fails unrelated (some deprecation warnings from shapely)

PR #3235, opened 2022-12-07
response: allow recalculating overall sensitivity with non translational units

### LINKED ISSUE #2990: add pressure (PA) to response unit_map
minor idea, but I do run into this sometimes working with hydrophone responses, particularly when trying to recalculate_overall_sensitivity. the label isn't relevant to the code and just seems to act as a safety. 

any harm in simply adding "PA" to line ~993 of core/inventory/response.py? 

e.g.

```
        unit_map = {
            "DISP": ["M"],
            "VEL": ["M/S", "M/SEC"],
            "ACC": ["M/S**2", "M/(S**2)", "M/SEC**2", "M/(SEC**2)",
                    "M/S/S", "PA"]}
```

any other instances I'm forgetting? unfortunately my master branch is still tied up in another PR so can't make this myself.

megies:
### What does this PR do?

Let evalresp recalculate the overall sensitivity even if the input unit is not of type VEL/ACC/DISP, e.g. for hydrophone with Pascal as unit. We just tell evalresp that target unit is "DEF" ("default"), i.e. just calculate response as is.

Closes #2990 

### PR Checklist
- [x] All tests still pass.


megies:
Also see #2945 for a similar topic/issue

obspy/core/inventory/response.py
petrrr, line 1: @megies: I just stepped into this issue while working with infrasound data (also "Pa") and am reviewing the fix.

megies: @petrrr good point. Probably we just didn't think of that and just overdid it here for no reason when we implemented this.

I just ran the tests with just always passing `"DEF"` to evalresp and everything seems fine. I agree we could just remove that block of unit checking and just use `"DEF""`. On the other hand, there's probably no real reason to act, it shouldn't change anything, now after this fix that's what happens for all "unusual" units anyway. Also one thing is that currently we only ever run into "M/S" or since this PR also "PA" in the tests, so we should add a test with a displacement and acceleration case first of all.

PR #3243, opened 2022-12-16
Energy ratio adjustments

megies:
### What does this PR do?

Minor adjustments in new energy ratio triggers, raise `ValueError` on nonsensical input values instead of returning zero-filled arrays.

CC @dizcza

minor adjustement to #3161
see https://github.com/obspy/obspy/pull/3161#pullrequestreview-1220710744

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/signal/tests/test_trigger.py
+++ b/obspy/signal/tests/test_trigger.py
@@ -4,11 +4,13 @@ The obspy.signal.trigger test suite.
 """
 import gzip
 import os
+import re
 import unittest
 import warnings
 from ctypes import ArgumentError
 
 import numpy as np
+import pytest
 from numpy.testing import assert_array_almost_equal, assert_array_equal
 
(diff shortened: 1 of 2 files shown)


dizcza:
LGTM, save for I'd use `assertRaises` without any expected message as (1) the message could change in the future and (2) you duplicate code by doing that.

Also, consider switching to either pytest or unittest. You have a mix of both and I don't see a reason for that.

d-chambers:
> Also, consider switching to either pytest or unittest. You have a mix of both and I don't see a reason for that.

The reason the mix exists is because we are transitioning from unittest to pytest. Agreed this is a good time to transition this test file to pytest if @megies is up for it.

megies:
> The reason the mix exists is because we are transitioning from unittest to pytest. Agreed this is a good time to transition this test file to pytest if @megies is up for it.

I might do a separate PR for this

PR #3247, opened 2022-12-21
Ignore DeprecationWarnings and PendingDeprecationWarnings triggered by other modules

trichter:
### What does this PR do?

See title. It adds two filters to pytest.ini

Deprecation warnings triggered in external modules have to be fixed upstream. E.g. this deprecation in shapely has to be fixed in cartopy and it should not be shown in our test results:
https://tests.obspy.org/132205/#1

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/pytest.ini
+++ b/obspy/pytest.ini
@@ -20,5 +20,8 @@ filterwarnings =
     ignore:Lines of type I have not been implemented yet
 # https://github.com/matplotlib/matplotlib/issues/21723
     ignore:Auto-removal of grids
-# see issue 3164, can be rmoved when NRL online tests get removed
+# see issue 3164, can be removed when NRL online tests get removed
     ignore:(?s).*Direct access to online NRL
+# ignore DeprecationWarnings and PendingDeprecationWarnings triggered by other modules
+    ignore::DeprecationWarning:(?!obspy).*
+    ignore::PendingDeprecationWarning:(?!obspy).*


trichter:
@megies

`1 << 7 == 128` and `np.int8(128)=-128`

This is probably not the desired result?

```
File "[/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/obspy/io/mseed/tests/test_mseed_util.py](https://github.com/obspy/obspy/blob/32cb3df757/obspy/io/mseed/tests/test_mseed_util.py#L309)", line 309, in test_get_flags
_create_mseed_file(
File "[/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/obspy/io/mseed/tests/test_mseed_util.py](https://github.com/obspy/obspy/blob/32cb3df757/obspy/io/mseed/tests/test_mseed_util.py#L75)", line 75, in _create_mseed_file
arr[:flags[key][name]] = 1 << _i
~~~^^^^^^^^^^^^^^^^^^^
DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays. The conversion of 128 to int8 will fail in the future.
For the old behavior, usually:
np.array(value).astype(dtype)`
will give the desired result (the cast overflows).
```

d-chambers:
The warning filters look fine, I didn't realize you could use them to ignore everything except a specified module, neat trick. 

I am not following the error from the bit shift magic going on in the miniseed module though. Is that related to the new warning filters?

trichter:
> I am not following the error from the bit shift magic going on in the miniseed module though. Is that related to the new warning filters?

I do not understand what exactly is tested here, but some values of 128 were silently converted to -128 by np.int8. This is deprecated in NumPy's newest version 1.24.0. It is also visible as [warning](https://tests.obspy.org/132169/) on master, but the traceback is only visible in this branch (via `-W error`), because the other warning which is filtered here is raised earlier.

megies:
> @megies
> 
> `1 << 7 == 128` and `np.int8(128)=-128`
> 
> This is probably not the desired result?

I have no idea what is going on in there, but that definitely doesn't feel right. Ideally somebody with better mseed internals knowledge could help sort it out, so we don't have to dig it up..?
In any case, if you change the dtype of that array to `uint8` so that `128` can be stored in there the test passes as well, which is kinda weird but makes me think maybe the dtype is just bad and maybe mseed is looking at `abd()` of these values..? No idea, really.

megies (approved):

PR #3258, opened 2023-01-20
pin sqlalchemy<2

megies:
### What does this PR do?

Pin SQLalchemy to version <2 and suppress deprecation warnings in tests

Non backwards compatible changes coming in upcoming version 2.0

- https://docs.sqlalchemy.org/en/20/errors.html#error-b8d9
- https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

CC @chad-earthscope 

Screenshot from 2023-01-20 10-13-22

### PR Checklist
- [ ] ~~While the PR is still work-in-progress, the `no_ci` label can be added to skip CI builds~~
- [ ] ~~If the PR is making changes to documentation, docs pages can be built automatically.
      Please post a link to the relevant piece of documentation.~~
- [ ] ~~If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the `test_network` tag to this PR.~~
- [ ] All tests still pass.
- [ ] ~~Any new features or fixed regressions are covered via new tests.~~
- [ ] ~~Any new or changed features are fully documented.~~
- [ ] ~~Significant changes have been added to `CHANGELOG.txt` .~~
- [ ] ~~First time contributors have added your name to `CONTRIBUTORS.txt` .~~
- [ ] ~~If the changes affect any plotting functions you have checked that the plots
      outputs are attached as artifacts.~~
- [ ] ~~New modules, add the module to `CODEOWNERS` with your github handle~~

### DIFF
--- a/obspy/clients/filesystem/db.py
+++ b/obspy/clients/filesystem/db.py
@@ -15,6 +15,18 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.schema import PrimaryKeyConstraint
 
 
+# we pinned our dependency to sqlalchemy <2.0 for now, so these warnings are
+# irrelevant. eventually code should be migrated to work with sqlalchemy >=2.0
+# as well
+try:
+    from sqlalchemy.exc import MovedIn20Warning
+except ImportError:
+    pass
+else:
+    import warnings
+    warnings.filterwarnings('ignore', category=MovedIn20Warning)
+
+
 Base = declarative_base()
 
 # Every declarative class should only be instantiated once. Thus we just use a
--- a/obspy/pytest.ini
+++ b/obspy/pytest.ini
@@ -22,6 +22,14 @@ filterwarnings =
     ignore:Auto-removal of grids
 # see issue 3164, can be removed when NRL online tests get removed
     ignore:(?s).*Direct access to online NRL
+# sqlalchemy showing deprecation warnings, seems like big API change with version 2
+# fixed sqlalchemy<2 for now in setup.py
+# also suppressing these warnings directly in clients.filesystem.db because
+# they get raised before test execution during test discovery
+# see https://docs.sqlalchemy.org/en/20/errors.html#error-b8d9
+# see https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
+    ignore:Deprecated API features .* are not compatible with SQLAlchemy 2.0
+    ignore:function is now available as sqlalchemy.orm.declarative_base
 # ignore DeprecationWarnings and PendingDeprecationWarnings triggered by other modules
     ignore::DeprecationWarning:(?!obspy).*
     ignore::PendingDeprecationWarning:(?!obspy).*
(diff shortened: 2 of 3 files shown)


trichter (approved):
In the long run we should migrate our code. Maybe open a new issue?
Also:
> The most prominent architectural features and API changes that are considered to be “SQLAlchemy 2.0” were in fact released as fully available within the 1.4 series

megies:
> In the long run we should migrate our code. Maybe open a new issue?

Well yeah, somebody would have to step up though, I am not familiar with that code.

megies:
@trichter I'm confused, might be doing something wrong. The added filters in pytest.ini catch the warnings but with `-W err` like in the test runner, they will still show an error

megies:
Even with `ignore:(?s).*` or `ignore:.*` if I add `-W err` the test run fails with that warning raised as an error

megies:
Aren't we doing it like this already for some other warnings? But if I understand docs correctly these filters don't apply with `-W err`?

Screenshot from 2023-01-20 13-24-38

trichter:
> Aren't we doing it like this already for some other warnings?
Exactly
>  But if I understand docs correctly these filters don't apply with -W err?
No, this warning is about the python standard filters.

I'll check out locally.

trichter:
I checked, you are right.

> The added filters in pytest.ini catch the warnings but with -W err like in the test runner, they will still show an error

It is working as expected for the other warnings.

trichter:
I pushed one unrelated commit to this branch.

trichter:
Now I remember... I have seen this before

> ERROR collecting clients/filesystem/db.py

The error is raised in the collection stage before the test suite is started.
If you replace `filterwarnings =` in `pytest.ini` with
````
filterwarnings =
error
```

and start the runtests normally it is working as expected.
I see two solutions:
1. Move `Base = declarative_base()` into the two functions or
2. Do not use `-W error`, but a different custom flag which writes a temporary ini file and load this with `-c`

megies:
> The error is raised in the collection stage before the test suite is started.

Ah... makes sense. Nice catch.

> I see two solutions:
> 
>     1. Move `Base = declarative_base()` into the two functions or
> 
>     2. Do not use `-W error`, but a different custom flag which writes a temporary ini file and load this with `-c`

Two more options:

 3. we could put a warning filter into `db.py` directly and catch it there manually for this one special case, dont see any negatives to this
 4. we could make use of that ENV variable you can set allegedly to suppress the warning


```
<stdin>:1: MovedIn20Warning: Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
```

megies:
I think goin with `3.` should be fine

trichter:
> I think goin with 3. should be fine

:+1: 

I guess the filter in pytest.ini is not needed anymore, but it does not hurt either.

PR #3269, opened 2023-02-25
sqlalchemy 2+ compatibility

filefolder:
as far as I can tell this should do it.. only two changes in `select` calls in tsindex.py as well as a change in the `declarative_base` import. tests work for me for both 1.4 and 2.0.4

feel free to edit anything peripheral I left out

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,8 @@ Changes:
    * response: allow recalculating overall sensitivity even if unit type is not
      one of VEL/ACC/DISP (see #3235)
    * trace: 5x quicker retrieval of seed-id with trace.id / trace.get_id()
+ - obspy.clients.filesystem:
+   * update syntax for SQLAlchemy 2.0 compatibility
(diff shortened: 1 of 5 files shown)


d-chambers (approved):
Hey @filefolder, Looks good to me. The `case` usage improves readability over the ternary operators IMHO. I am not very familiar with the `tsindex` module but I don't think you are changing any behavior and since the test suite is still passing I think we can merge it. I will wait a day or two though just in case anyone else wants to weigh in.

d-chambers:
Ok, looks like we are good to go. We can always do a follow-up PR if something needs to be changed.

megies:
@filefolder do we still need the pinning for `sqlalchemy>=1.4` in setup.py? If not we should remove it in #3570  CC @ThomasLecocq

filefolder:
probably not but it won’t hurt anything. will have to investigate any
concessions for v1 that still exist

megies:
Ah, its not really important I think.. Probably it's "whatever" =)

PR #3285, opened 2023-03-24
read/write support for csv, eventtxt and csz

### LINKED ISSUE #1467: Read/write support for eventtxt format
We should add read/write support for the txt format for events as defined by FDSNWS specifications.

trichter:
### What does this PR do?

Id adds read and write support for csv, eventtxt and csz. csz is a custom format and basically a collection of csv files bundled in a zip file (similar to numpys npz format) for io of a catalog with picks.

Fixes  #1467

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/csv/core.py
+++ b/obspy/io/csv/core.py
@@ -1,4 +1,3 @@
-## Copyright 2022 Tom Eulenfeld, MIT license
 """
 CSV and CSZ read/write support for ObsPy earthquake catalogs
 
@@ -50,10 +49,10 @@ FIELDS = {
     'eventtxt': (
         '{id},{time!s:.25},{lat:.6f},{lon:.6f},{dep:.3f},{author},,{contrib},,'
         '{magtype},{mag:.2f},{magauthor},{region}').split(',')
-    }
+}
 PFIELDS = {
     'basic': '{seedid} {phase} {time:.5f} {weight:.3f}'
-    }
+}
 CSZ_COMMENT = f'CSZ format v{__version__} obspy_no_uncompress'.encode('utf-8')
 # for load_csv
 DTYPE = {
@@ -64,7 +63,7 @@ DTYPE = {
     'mag': float,
     'magtype': 'U10',
     'id': 'U50'
-    }
+}
 
 # catalog and contribid are not used
 EVENTTXT_NAMES = (  # for reading
@@ -79,14 +78,14 @@ EVENTTXT_HEADER = (  # for writing
 def _is_csv(fname, **kwargs):
     try:
         return _read_csv(fname, format_check=True)
-    except:
+    except Exception:
         return False
 
 
 def _is_eventtxt(fname, **kwargs):
     try:
         return _read_eventtxt(fname, format_check=True)
-    except:
+    except Exception:
         return False
 
 
@@ -97,7 +96,7 @@ def _is_csz(fname, **kwargs):
             assert (zipf.comment.startswith(b'CSZ') and
                     zipf.comment.endswith(b'obspy_no_uncompress'))
         return True
-    except:
+    except Exception:
         return False
 
 
@@ -160,7 +159,7 @@ def _read_csv(fname, skipheader=0, default=None, names=None,
         catalog = read_events('external.csv', 'CSV', skipheader=1, names=names)
     """
     if default is None:
-        default= DEFAULT
+        default = DEFAULT
     events = []
     with _open(fname) as f:
         for _ in range(skipheader):
@@ -195,7 +194,7 @@ def _read_csv(fname, skipheader=0, default=None, names=None,
                 longitude=row['lon'],
                 depth=dep,
                 creation_info=info
-                )
+            )
             try:
                 # add zero to eliminate negative zeros in magnitudes
                 mag = float(row['mag']) + 0
@@ -227,7 +226,7 @@ def _read_csv(fname, skipheader=0, default=None, names=None,
                 origins=[origin],
                 resource_id=id_,
                 event_descriptions=descs
-                )
+            )
             events.append(event)
             if format_check:
                 return True
@@ -306,7 +305,7 @@ def _write_csv(events, fname, fields='basic', depth_in_km=True, delimiter=',',
             d = {'time': origin.time,
                  'lat': origin.latitude,
                  'lon': origin.longitude,
-                 'dep' if depth_in_km else 'depm' : dep,
+                 'dep' if depth_in_km else 'depm': dep,
                  'mag': mag,
                  'magtype': magtype,
                  'id': evid,
@@ -340,9 +339,9 @@ def load_csv(fname, skipheader=0, only=None, names=None,
             names = f.readline().strip().split(',')
         names = _names_sequence(names)
         dtype = [(n, DTYPE[n]) for n in names if n in DTYPE and
-                        (only is None or n in only)]
+                 (only is None or n in only)]
         usecols = [i for i, n in enumerate(names) if n in DTYPE and
-                        (only is None or n in only)]
+                   (only is None or n in only)]
         kw.setdefault('usecols', usecols)
         kw.setdefault('dtype', dtype)
         return np.genfromtxt(f, delimiter=delimiter, **kw)
@@ -410,7 +409,7 @@ def _write_picks(event, fname, fields_picks='basic', delimiter=','):
     weights = {str(arrival.pick_id): arrival.time_weight
                for arrival in origin.arrivals if arrival.time_weight}
     phases = {str(arrival.pick_id): arrival.phase
-               for arrival in origin.arrivals if arrival.phase}
+              for arrival in origin.arrivals if arrival.phase}
     with _open(fname, 'w') as f:
         f.write(delimiter.join(fieldnames) + '\n')
         for pick in event.picks:
--- a/obspy/io/csv/core.py
+++ b/obspy/io/csv/core.py
@@ -33,11 +33,8 @@ import zipfile
 
 
 import numpy as np
-from obspy import UTCDateTime as UTC
+from obspy import UTCDateTime as UTC, __version__
 from obspy.core import event as evmod
-
-
-__version__ = '0.8.1-dev'
 
 # for reading
 DEFAULT = {'magtype': ''}
@@ -53,7 +50,7 @@ FIELDS = {
 PFIELDS = {
     'basic': '{seedid} {phase} {time:.5f} {weight:.3f}'
 }
-CSZ_COMMENT = f'CSZ format v{__version__} obspy_no_uncompress'.encode('utf-8')
+CSZ_COMMENT = f'CSZ format ObsPy v{__version__} obspy_no_uncompress'.encode('utf-8')
 # for load_csv
 DTYPE = {
     'time': 'datetime64[ms]',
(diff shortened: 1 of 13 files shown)


trichter:
[Doc link](https://docs.obspy.org/pr/csv/packages/obspy.io.csv.html#module-obspy.io.csv). Some functions are mising its doc string.

trichter:
Do we need an `__init__.py` file in the tests directory for the tests to be discovered by pytest? Why?

megies:
> Do we have another 'test_csv.py' anywhere?

Doesn't look like it.

megies (dismissed):
Still needs changelog entry, otherwise looks good to me :rocket:

obspy/io/csv/tests/test_csv.py
```diff
@@ -0,0 +1,215 @@
...
+def test_csv_incomplete_catalog():
+    events = read_events()
+    del events[0].magnitudes[0].magnitude_type
+    events[1].magnitudes = []
+    events[1].preferred_magnitude_id = None
+    events[1].origins[0].depth = None
+    events[2].origins = []
+    events[2].preferred_origin_id = None
+    with NamedTemporaryFile(suffix='.csv') as ft:
+        with pytest.warns(Warning):
```
d-chambers, line 53: Might be good to add a match argument here to make sure we are getting the right warning.

obspy/io/csv/tests/test_csv.py
```diff
@@ -0,0 +1,215 @@
...
+                assert len(ev2.origins[0].arrivals) == \
+                    len(ev1.origins[0].arrivals)
+                assert len(ev2.picks) == \
+                    len(ev1.picks)
+        _test_write_read(events)
+        _test_write_read(events, compression=False)
+        try:
+            import zlib  # noqa: F401
+        except ImportError:
+            pass
```
d-chambers, line 111: consider `suppress` from context lib here to make it more concise and clear.

obspy/io/csv/tests/test_csv.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            _test_write_read(events, compression=True, compresslevel=6)
+        # test with missing origin and waveformid
+        events[1].origins = []
+        events[0].picks[0].waveform_id = None
+        with pytest.warns(Warning):
```
d-chambers, line 117: again, a match would be good here

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
+"""
+CSV and CSZ read/write support for ObsPy earthquake catalogs
+
+You can use the following field names to read an external CSV file
+time or year, mon, day, hour, minu, sec
+lat, lon, dep, mag, magtype, id
+(see also global FIELDS variable and help of read_csv)
+
+Note: This plugin can be easily extended to write and read more event
```
d-chambers, line 9: "plugin" -> "module"

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+from warnings import warn
+import zipfile
+
+
+import numpy as np
+from obspy import UTCDateTime as UTC
+from obspy.core import event as evmod
+
+
+__version__ = '0.8.1-dev'
```
d-chambers, line 40: Probably doesn't need its own version now that it is versioned with ObsPy?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+}
+
+# catalog and contribid are not used
+EVENTTXT_NAMES = (  # for reading
+    'id time lat lon dep author catalog contrib contribid '
+    'magtype mag magauthor region')
+EVENTTXT_HEADER = (  # for writing
+    '#EventID | Time | Latitude | Longitude | Depth/km | Author | Catalog | '
+    'Contributor | ContributorID | MagType | Magnitude | MagAuthor | '
+    'EventLocationName')
```
d-chambers, line 75: Maybe headers and text should just be the same? Shorter names does save some typing but now there are two things to maintain. Also, are the names fixed by the format you are supporting? Otherwise the headers might be better off being identical to the corresponding attribute names in ObsPy's catalog/event objects.

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+
+
+def _is_csv(fname, **kwargs):
+    try:
+        return _read_csv(fname, format_check=True)
+    except Exception:
```
d-chambers, line 81: Maybe difficult here, but is there anyway to be more specific with the exception type?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+        return False
+
+
+def _is_csz(fname, **kwargs):
+    try:
+        assert zipfile.is_zipfile(fname)
+        with zipfile.ZipFile(fname) as zipf:
+            assert (zipf.comment.startswith(b'CSZ') and
+                    zipf.comment.endswith(b'obspy_no_uncompress'))
```
d-chambers, line 97: If python is run in optimized mode (-O flag) this wont work because asserts are ignored.

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+
+
+@contextmanager
+def _open(filein, *args, **kwargs):
+    """Accept both files or file names"""
+    if isinstance(filein, str):  # filename
+        with open(filein, *args, **kwargs) as f:
+            yield f
+    else:  # file-like object
+        yield filein
```
d-chambers, line 118: What about a `pathlib.Path` object?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+        meter (False)
+    :param default: dictionary with default values, at the moment only
+         magtype is supported,
+         i.e. to set magtypes use `default={'magtype': 'Ml'}`
+    :param names: determined automatically from header line of file,
+        otherwise can be specified as string, sequence or dict
+    :param check_compression: not used by read_csv
+    :param format_check: Read only the first event
+    :param **kargs: all other kwargs are passed to csv.DictReader,
+        important additional arguments are fieldnames, dialect, delimiter, etc
```
d-chambers, line 151: Check these params, they don't match up with the function signature (no fname, in params, no depth_in_km in signature)

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+
+    You can read csv files created by this module or external csv files.
+    Example reading an external csv file:
+
+        from obspy import read_events
+        names = ('year mon day hour minu sec _ '
+                 'lat lon dep _ _ mag _ _ _ _ _ _ id')
+        catalog = read_events('external.csv', 'CSV', skipheader=1, names=names)
```
d-chambers, line 159: Should be a doctest with an examples header?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+                     if region else [])
+            event = evmod.Event(
+                magnitudes=magnitudes,
+                origins=[origin],
+                resource_id=id_,
+                event_descriptions=descs
+            )
+            events.append(event)
+            if format_check:
+                return True
```
d-chambers, line 232: Certainly optional, but this is a bit long. Maybe can be factored into separate helper functions for reading each part?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+                warn(f'No origin found -> do not write event {evid}')
+                continue
+            try:
+                author = origin.creation_info.author
+            except AttributeError:
+                author = ''
+            try:
+                contrib = origin.creation_info.agency_id
+            except AttributeError:
+                contrib = ''
```
d-chambers, line 279: These cant be re-phrased as `getattr` with default values because both creation_info and creation_info's sub attribute might be missing right?

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+    """
+    Convert ObsPy catalog to numpy array
+
+    All kwargs are passed to `load_csv`, e.g. use `only=('lat', 'lon', 'mag')`
+    to get an array with lat, lon, mag parameters.
+    """
+    with io.StringIO() as f:
+        _write_csv(events, f)
+        f.seek(0)
+        return load_csv(f, **kw)
```
d-chambers, line 360: This is generally useful, but I wonder if it is faster than just parsing the data from the event objects directly? Also, should it live in this module? If we decide to make a generic version of it that doesn't end up writing to csv first its current location doesn't make a lot of sense. 

In light of that, you might consider making it a private function for now.

obspy/io/csv/core.py
```diff
@@ -0,0 +1,484 @@
...
+    weights = {str(arrival.pick_id): arrival.time_weight
+               for arrival in origin.arrivals if arrival.time_weight}
+    phases = {str(arrival.pick_id): arrival.phase
+              for arrival in origin.arrivals if arrival.phase}
+    with _open(fname, 'w') as f:
+        f.write(delimiter.join(fieldnames) + '\n')
+        for pick in event.picks:
+            pick_id = str(pick.resource_id)
+            try:
+                seedid = pick.waveform_id.id
```
d-chambers, line 418: Could also be None or ""? In that cause should the warning also be emitted?

d-chambers (commented):
Good addition, I just left a few suggestions.

megies:
Thanks for the thorough review @d-chambers, definitely some very valid points there :smiley:

obspy/io/csv/tests/test_csv.py
trichter: I added a match argument. Unmatched warnings are silenced though, see https://github.com/pytest-dev/pytest/issues/9288.

obspy/io/csv/tests/test_csv.py
trichter: I did not change anything here for now

obspy/io/csv/core.py
trichter: Sure, a leftover, thanks.

obspy/io/csv/core.py
trichter: Let's postpone that

obspy/io/csv/core.py
trichter: Exactly

obspy/io/csv/core.py
trichter: Unfortunately, I already have some CSV files out there with these short headers. Therefore, I would prefer to keep those short headers, if that is OK.

obspy/io/csv/core.py
trichter: Yes, I agree to turn it to a private function. Actually it should be the other way round, this function implemented in core and as stream method without everything converted to strings and write_csv and cohorts building upon that.
Though, I think the conversion to and from strings here will not increase the running time significantly. Was such a nice low hanging fruit. Related: #1516

obspy/io/csv/core.py
trichter: I guess not. We do not want to have any exceptions in the format checkers.

obspy/io/csv/core.py
trichter: Good catch

megies (approved):
I collapsed everything that was addressed and made linter happy. Looks like the rest is optional stuff that can be left as is?

d-chambers:
👍 looks good to me.

Issue #3228, opened 2022-11-22
Release 1.5.0


megies:
This is a very much shortened release check list, after focusing on wheel/conda/pypi only releases.

#### Before release

- [x] Discuss planned release date and update minimum Python and dependency versions accordingly on `master` branch, compare [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) and [scipy supported versions](https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy) (see also [Python release schedule](https://devguide.python.org/versions/)) see #3570 
- [x] [zenodo](https://zenodo.org/): reserve a DOI pre-release (DOI: **10.5281/zenodo.19005357**) **or** change zenodo release workflow
- [x] "What's new?" page in the wiki --> https://github.com/obspy/obspy/wiki/What's-New-in-ObsPy-1.5
- [x] test wheel building by adding test_wheels label to a PR, alternatively test wheel building by tagging a release candidate (source distribution and wheels for release candidates get uploaded to [test.pypi](https://test.pypi.org/project/obspy) if all wheel builds and their test suites succeed)
- [x] make sure Python 3.13 gets added to CI wheel building, should be fixed by #3570 

#### To make the release

- [x] set the [version number and DOI in CHANGELOG.txt](https://github.com/obspy/obspy/blob/master/CHANGELOG.txt), set a git tag, [publish tag as a release](https://github.com/obspy/obspy/tags)
- [x] source distribution to [pypi](https://pypi.org/project/obspy/) ([done automatically by github actions](https://github.com/obspy/obspy/pull/2943))
- [x] wheels for all platforms on [pypi](https://pypi.org/project/obspy/) ([done automatically by github actions](https://github.com/obspy/obspy/pull/2943))
- [x] [conda packages](https://anaconda.org/conda-forge/obspy/files) built via [conda-forge feedstock](https://github.com/conda-forge/obspy-feedstock/pulls)
- [x] [zenodo](https://zenodo.org/): upload sdist zipball and finalize release

#### After release

- [x] Update [docs](https://docs.obspy.org/) once released (should be done automatically, built in gh-actions and [deployed on our server by a bot](https://github.com/obspy/obspy/tree/master/misc/scripts/docs_deploybot)), optionally tar the docs and attach to github release
- [x] Ensure new [docset](https://docsets.obspy.org) is in place (worked automatically through CI ~~currently no docset building set up in release CI~~)
[...]

PR #3087, opened 2022-06-07
raise min versions of numpy, scipy, mpl in setup.py

trichter:
### What does this PR do?

Finalize #3016 and #3081. See https://github.com/obspy/obspy/pull/3081#issuecomment-1148537557

### PR Checklist
- [ ] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] This PR is not directly related to an existing issue (which has no PR yet).
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just add the "build_docs" tag to this PR.
      Docs will be served at [docs.obspy.org/pr/{branch_name}](https://docs.obspy.org/pr/) (do not use master branch).
      Please post a link to the relevant piece of documentation.
- [ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the "test_network" tag to this PR.
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are covered via new tests.
- [ ] Any new or changed features are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .
- [ ] If the changes affect any plotting functions you have checked that the plots
      from all the CI builds look correct. Add the "upload_plots" tag so that plotting 
      outputs are attached as artifacts. 
- [ ] Add the "ready for review" tag when you are ready for the PR to be reviewed.

PR #3287, opened 2023-03-30
Fixes for Seisan + Nordic files: parsing of int16 waveform files, edge cases in event files, compact event files

flixha:
### What does this PR do?

 - obspy.io.nordic:
   * Parse ID-lines (type I) into event.extra['nordic_event_id]
   * Support reading of compact Nordic files (only containing header lines)
   * Write amplitude with highest possible precision
   * Fix writing / reading of high-frequency IAMLHF amplitudes
   * Fix parsing of BAZ when they precede the associated phase pick in Nordic file
   * Fix bug where check for nordic format version crashed on long phase names
 - obspy.io.seisan:
   * Fix parsing of files written in int16 (previously read as int32 which resulted in waveforms that looked very similar to the actual waveforms, but had wrong number of samples and wrong amplitudes)

This PR takes care of some additional, rather rare edge cases for reading of Seisan waveform files and Nordic event parameter files. There were not related open issues that this PR resolves.

I realize this PR now combines a few more commits than I originally anticipated, so sorry for that. Let me know if you'd rather have it split up.

### PR Checklist
- ~[ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Please post a link to the relevant piece of documentation.~
- ~[ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the `test_network` tag to this PR.~
- [ ] All tests still pass.
- ~[] Any new or changed features are fully documented.~
- ~[ ] First time contributors have added your name to `CONTRIBUTORS.txt` .~
- ~[ ] If the changes affect any plotting functions you have checked that the plots
      outputs are attached as artifacts. ~
- ~[ ] New modules, add the module to `CODEOWNERS` with your github handle~


obspy/io/nordic/core.py
```diff
@@ -798,6 +814,35 @@ def _read_comments(tagged_lines, event):
...
+    if 'I' not in tagged_lines.keys():
+        return event
+    id_lines = tagged_lines['I']
+    if len(id_lines) > 1:
+        warnings.warn('Nordic file has more than one ID line, will use first' +
+                      'ID line only.')
+    for id_line, tag in id_lines:
+        event_id = id_line.split('ID:')[-1].split(' ')[0].strip('dSLRD')
+        break
+    event.extra = {
```
megies, line 838: Right now this might be the only instance setting anything on `extra` but that might change.. might be safer to do something like 

```python
extra = event.setdefault('extra', {})
extra['nordic_event_id'] = ...
```

obspy/io/nordic/core.py
```diff
@@ -824,7 +869,10 @@ def _read_picks(tagged_lines, new_event, nordic_format='UKN', **kwargs):
                     tagged_lines[tag], key=lambda tup: tup[1])])
         except KeyError:
             pass
-    header = sorted(tagged_lines['7'], key=lambda tup: tup[1])[0][0]
+    try:
+        header = sorted(tagged_lines['7'], key=lambda tup: tup[1])[0][0]
+    except IndexError:  # Compact catalog file without picks?!
+        pass
```
megies, line 875: This might be OK but indexing is used 4 times on that line, so it might or nmight not be too "catch all" of an exception handling? Up to you though, I don't know the details here too much.

megies:
I don't know this format at all, so I only have a few comments after briefly looking at this.

Two things to consider though:
 - I feel like this is a bit of a mix of things that could go into maintenance branch and some other things that feel like changing too much for that and should rather go into master. It's fine for me to have all this together in this PR but maybe we should rather merge into master then (can be easily changed up top, only changelog would need moving after a quick rebase on master)
 - Current status is we might still support Python 3.8 for the next set of bugfix+feature release (see discussion in #3228 and #3087) and then drop it right after, so maybe it might be good to work around that new `removeprefix()` convenience method

flixha:
Thanks a lot for the review!
I will get to fix your suggestions, may just need a few days until I get on it.
I understand and agree that this may be better suited for merging into master for the next release - after all it fixes relatively rare edge cases while the new functionality would have better gone into a new release anyways.

ThomasLecocq:
@flixha I bumped on your PR when helping Kris on #3584 , wondering if you'd have time to rebase & update this PR ?

ThomasLecocq:
@flixha I merged current maintenance branch into this one, and corrected the first comment from @megies in the latest commit. Could you check the remaining one (regarding the nasty indexing+lambda func).

I did not change the `removeprefix` things, as I'm hoping we'll drop 3.8 (😇 and 3.9) for the release :-)

Don't worry for the branch (maintenance), maintenance & master will soon be merged anyway.

ThomasLecocq:
rebased & changed the startswith thingy to 3.8 pass, doesn't really harm. small bugfix in the reading of the extra element.

ThomasLecocq:
only missed test is related to a warning that is not caputed, it seems. Otherwise all good.

PR #3306, opened 2023-05-24
preserve FDSN server response when error

### LINKED ISSUE #3305: preserve URL on error for FDSN web service requests
### Problem

Often I receive something like this from the FDSN Client:
```
FDSNBadRequestException: Bad request. If you think your request was valid please contact the developers.
HTTP Status code: 400
Detailed response of server:

```
which lets me know my request was bad. But because the server provides no details and the original request URL is not part of the FDSNBadRequestException, it is pretty hard to debug. If the URL were retained and printed as part of the exception, I would at least be able to paste it into a web browser and try various changes to determine why it is failing, then update my script. 

### Proposed solution

Might be better ways to do this, but I think passing optional url parameter to `raise_on_error()` in clients/fdsn/client.py and also adding an url field to FDSNException would be easy way to preserve the request url.

This probably would not fix cases where POST was used, but would at least help for the more common GET style requests.

crotwell:
### What does this PR do?

Currently the response from server on error was either read directly into the debug msg or tossed, so never can be part of exception. Better to preserve the message, so can be passed on and be included in the final exception retured to the user.

fixes #3305 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -1799,12 +1799,7 @@ def raise_on_error(code, data):
     """
     # get detailed server response message
     if code != 200:
-        try:
-            server_info = data.read()
-        except Exception:
-            server_info = None
-        else:
-            server_info = server_info.decode('ASCII', errors='ignore')
+        server_info = data.decode('ASCII', errors='ignore')
         if server_info:
             server_info = "\n".join(
                 line for line in server_info.splitlines() if line)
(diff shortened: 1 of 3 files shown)


obspy/clients/fdsn/client.py
```diff
@@ -1890,14 +1885,12 @@ def download_url(url, opener, timeout=10, headers={}, debug=False,
         url_obj = opener.open(request, timeout=timeout, data=data)
     # Catch HTTP errors.
     except urllib_request.HTTPError as e:
+        error_data = e.read()
```
megies, line 1888: I'm confused about this one.. I don't see that `HTTPError` even has a `.read()` method.. :thinking: looks like `urllib` is adding some magic stuff if a filepointer is given to `HTTPError.__init__()` and then those additionally populated methods are left out in the docs..  but hey it works :smile:

megies (approved):
Looks great, thanks a bunch! :rocket: 

Was wondering about the missing server response every now and then but sadly never enough to dig into it  :blush:

PR #3333, opened 2023-07-25
replace pkg_resources with importlib.metadata from stdlib

megies:
### What does this PR do?

Replaces `pkg_resources` which was used internally to handle entry points / plugin structure with `importlib.metadata` API. The `EntryPoint` objects from `importlib` (and other objects like `Distribution` all have very minor differences to the old `pkg_resources` objects but most of these routines are private and not intended for public use. Hopefully tests catch all use cases.

pkg_resources recently started showing this message:

```
DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/test_mindepversion_conda_env.yml
+++ b/.github/test_mindepversion_conda_env.yml
@@ -1,7 +1,6 @@
 name: test
 channels:
   - conda-forge
-  - defaults
 dependencies:
   # obspy
   - numpy=1.21
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,6 +166,7 @@ jobs:
       - name: Update environment
         run: conda env update -n test -f ${{ env.env_file }}
         if: steps.cache.outputs.cache-hit != 'true'
+        timeout-minutes: 5
 
       - name: print package info
         shell: bash -l {0}


ThomasLecocq:
i wonder if this is going to make a difference in load-time of obspy :-)

megies:
> i wonder if this is going to make a difference in load-time of obspy :-)

No idea, just wanted to act on that `DeprecationWarning` from `pkg_resources` :wink: 

Rebased after merging current maintenance branch into master

megies:
Looks like `importlib.metadata` in Python 3.8 and 3.9 still has some reduced API for entry points. If we keep following numpy we'd drop Python 3.8 anyway for 1.5.0, but that would still leave Python 3.9 until next April or so.
So if we wanna go through with this now we'd have to put some compatibility layer in there that emulates the missing filtering logic. So not sure if it's worth pushing this now, maybe just leave this open and push it to 1.6.0 or some other release coming next year.

megies:
This is starting to get ugly with the workarounds needed for Py 3.8 and 3.9. Maybe best make this switch after dropping Py 3.9 and until then just add an ignore rule for that deprecation warning on Py 3.12

obspy/conftest.py
```diff
@@ -237,17 +238,26 @@ def pytest_json_modifyreport(json_report):
...
     version_info = {'obspy': obspy.__version__}
-    for req in distribution.requires():
-        name = req.name
-        version = pkg_resources.get_distribution(name).version
+    for req in importlib.metadata.requires("obspy"):
+        # can not find proper documentation for how exactly these requirement
+        # strings are defined, but it looks like a semicolon is used as a
+        # separator for additional definitions like "extra" in optional
+        # dependencies and old pkg_resources ignored those, so skip them with
+        # importlib too
```
QuLogic, line 247: I think it follows [PEP508](https://peps.python.org/pep-0508/); you could [parse it with `packaging`](https://packaging.pypa.io/en/stable/requirements.html).

obspy/core/event/catalog.py
```diff
@@ -497,8 +497,8 @@ def write(self, filename, format, **kwargs):
             format_ep = EVENT_ENTRY_POINTS_WRITE[format]
             # search writeFormat method for given entry point
             write_format = buffered_load_entry_point(
-                format_ep.dist.key, 'obspy.plugin.event.%s' % (format_ep.name),
-                'writeFormat')
+                format_ep.dist.name,
+                'obspy.plugin.event.%s' % (format_ep.name), 'writeFormat')
```
QuLogic, line 501: Are we able to change all of these to f-strings? There are quite a few of these, and they're all a bit different, such as this one having superfluous parentheses.

megies:
Regarding the getting select entry points, see https://github.com/python/cpython/issues/107220#issuecomment-2263800387 we can do something like this:

```python
importlib.metadata.distribution('pip').entry_points.select(name='pip', group='console_scripts')
```

ThomasLecocq:
> Regarding the getting select entry points, see [python/cpython#107220 (comment)](https://github.com/python/cpython/issues/107220#issuecomment-2263800387) we can do something like this:
> 
> ```python
> importlib.metadata.distribution('pip').entry_points.select(name='pip', group='console_scripts')
> ```

I see a helper function coming for that one :-)

ThomasLecocq:
@megies @QuLogic I would like to work on this in the coming days, what's the best strategy? merge changes from master & build on that ?

ThomasLecocq:
I updated the branch to be // with master, so we can test :-)

ThomasLecocq:
Sooooo ok, ok, I have to switch to local testing... Of course on windows it's a nightmare to test 3.8 and 3.9, I'll switch to the server for the dev :)

ThomasLecocq:
4 to go !  https://tests.obspy.org/146985/

ThomasLecocq:
2 ! https://tests.obspy.org/147018/

ThomasLecocq:
ONLY ONE ! https://tests.obspy.org/147069/

Honestly doesn't matter so much, it's just the cache of the entrypoint - if it works, it works...

Have to check if the network tests pass too !

ThomasLecocq:
So, happy me :-) 

Failing tests:

- related to pkg_resources vs importlib : ONE - and it's just the caching of the entrypoint
- not related : the network tests that should be solved by #3628

ThomasLecocq:
I added a version check for the failing test. The reason it failed is linked to the number of calls to the mock' entry point object - the entry point system works, all tests pass for all IO etc.

megies:
Removal of `pkg_resources` might happen end of this November with new setuptools release, so need this for 1.5

ThomasLecocq:
> Removal of `pkg_resources` might happen end of this November with new setuptools release, so need this for 1.5

i believe this PR is ready :-)

ThomasLecocq:
OK, remaining tests are not related to here... it's a NRL issue...

Afraid, but going to click the button :-)

ThomasLecocq (approved):

PR #3345, opened 2023-08-29
seiscomp schema 0.13 upgrade

filefolder:
I think these are mostly speed enhancements so should hopefully not break anything (see https://github.com/SeisComP/common/pull/88)

Despite the branch name it is for 0.6 to the current 0.12. Will be testing pretty thoroughly over the weeks.

update: bit off more than I can chew on this one, all of the event test data needs to be restructured. 

### PR Checklist
- [x] All tests still pass.


filefolder:
dusting this off for a moment (sorry for falling off there), hope to rebase and complete this soon

1) Seiscomp schema 0.13 is now out (alongside version 6), will incorporate

2) The previous speed sc3ml speed update(s) change the actual structure of the converted xml files, so in order for the obspy tests to pass I will have to replace nearly all the example files. I am a bit wary about this. I am wondering if perhaps instead of testing that the files are converted files identical, we only test that the elements of the loaded catalog object are equivalent? What would be the best way to do this?

megies:
> 2\. What would be the best way to do this?

Hmm.. yeah not sure. I haven't looked at the changes but I'm guessing we can keep it to be able to read the old one as well still right? So maybe we could leave the current tests and just add a few ones for the new version on top maybe?

filefolder:
Couldn't figure out a way to avoid changing the test material so just went for it. A pain but hopefully won't have to do it again for long time.

Will have to do some more tests with 0.13 compatibility but the changes are minor so don't expect any issues. Will post again when it seems all clear.

Kept the files for versions .9 and lower as-is, last relevant in 2017.

megies:
Ah, yeah, it didn't click last time.. we only write the latest SCML version so the files with expected test output for writing SCML needed adaption and no way around it.

megies (approved):
PRs touching these multi-version reader things are always kinda hard to review with the many changes involved. But it looks good overall, tests still pass and there's only minimal and sound changes to the test code. So, should be good, thanks! :rocket:

Issue #3320, opened 2023-07-06
FDSN client string decoding glitch 


filefolder:
### Avoid duplicates

- [X] I searched existing issues

### Bug Summary

Seems like the current version of fdsn/client.py not handling (some?) server response properly. In particular, when attempting to access restricted data with an incorrect username or password.

Doesn't like trying to decode a string-- not sure what version of python this started happening with, is this possibly a very very very old bug? Obspy is strictly Python 3 now yes?

### Code to Reproduce

```python
from obspy.clients.fdsn import Client
ap = Client('AUSPASS',user='nope',password='password')
st = ap.get_waveforms(station='AKL10',channel='BHZ',location='*',network='X5',starttime=obspy.UTCDateTime(2023,5,20,1,52),endtime=obspy.UTCDateTime(2023,5,20,2,2))
```

### Error Traceback

```Python traceback
File ~/.local/lib/python3.11/site-packages/obspy/clients/fdsn/client.py:872, in Client.get_waveforms(self, network, station, location, channel, starttime, endtime, quality, minimumlength, longestonly, filename, attach_response, **kwargs)
    867 url = self._create_url_from_parameters(
    868     "dataselect", DEFAULT_PARAMETERS['dataselect'], kwargs)
    870 # Gzip not worth it for MiniSEED and most likely disabled for this
    871 # route in any case.
--> 872 data_stream = self._download(url, use_gzip=False)
    873 data_stream.seek(0, 0)
    874 if filename:

File ~/.local/lib/python3.11/site-packages/obspy/clients/fdsn/client.py:1486, in Client._download(self, url, return_string, data, use_gzip, content_type)
   1481     headers['Content-Type'] = content_type
   1482 code, data = download_url(
   1483     url, opener=self._url_opener, headers=headers,
   1484     debug=self.debug, return_string=return_string, data=data,
   1485     timeout=self.timeout, use_gzip=use_gzip)
-> 1486 raise_on_error(code, data)
   1487 return data

File ~/.local/lib/python3.11/site-packages/obspy/clients/fdsn/client.py:1807, in raise_on_error(code, data)
   1805     server_info = None
   1806 else:
-> 1807     server_info = server_info.decode('ASCII', errors='ignore')
   1808 if server_info:
   1809     server_info = "\n".join(
   1810         line for line in server_info.splitlines() if line)

AttributeError: 'str' object has no attribute 'decode'
```

### ObsPy Version?

1.4.0

### Operating System?

Debian 12

[...]

PR #3348, opened 2023-09-07
fdsn: try to fix regression introduced by #3306, see #3320

megies:
### What does this PR do?

Aims to fix some internal handling of what info we get back from server in case of errors in FDSN requests.
No changelog needed since this regression is not yet in a released version.

 - Fixes a regression introduced by #3306 and..
 - should solve an issue reported in https://github.com/obspy/obspy/issues/3320#issuecomment-1706809150 and..
 - in principal should also fix the original potential bug reported in #3320 (even though I can not reproduce the original report and thus can not be sure)

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -1799,7 +1799,21 @@ def raise_on_error(code, data):
     """
     # get detailed server response message
     if code != 200:
-        server_info = data.decode('ASCII', errors='ignore')
+        # let's try to resolve all the different types that `data` can sadly
+        # have..
+        # first, if it's `BytesIO` (or should it for whatever reason be
+        # `StringIO` which it shouldn't..) then break it down by reading it
+        try:
+            server_info = data.read()
+        # if there is no `read()` method it should be `bytes`
+        except AttributeError:
+            server_info = data
+        # now decode the bytes (or if for whatever weird reason we ended up
+        # with a string, then do nothing more)
+        try:
+            server_info = server_info.decode('ASCII', errors='ignore')
+        except AttributeError:
+            pass
         if server_info:
             server_info = "\n".join(
                 line for line in server_info.splitlines() if line)
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -1899,7 +1899,10 @@ def download_url(url, opener, timeout=10, headers={}, debug=False,
         url_obj = opener.open(request, timeout=timeout, data=data)
     # Catch HTTP errors.
     except urllib_request.HTTPError as e:
-        error_data = e.read()
+        try:
+            error_data = e.msg
+        except AttributeError:
+            error_data = e.read()
         if debug is True:
             msg = "HTTP error %i, reason %s, while downloading '%s': %s" % \
                   (e.code, str(e.reason), url, error_data)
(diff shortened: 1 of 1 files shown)


crotwell:
This is just a guess, but if the server incorrectly sets the Content-Type HTTP header to a binary type, like `application/vnd.fdsn.mseed` before the error occurs, then python may think that the response is not actually text, and hence returns bytes.

It would be really useful to get the actual URL that obspy is generating to test against the server using curl to see exactly what is being returned.

Otherwise, it seems like being defensive about parsing the returned error message is wise.

filefolder:
can confirm this fixes both the 401 bad auth error and the 204 no data error.

if one wanted to try a wget test, the actual link can often be manually created via the direct fdsnws implementation (e.g. server.com/fdsnws) at most FDSN centers. for GFZ it is here: https://geofon.gfz-potsdam.de/fdsnws/station/1/builder

here's an example for a non-existing network: https://geofon.gfz-potsdam.de/fdsnws/station/1/query?network=HH&format=text&nodata=404

filefolder (approved):
confirmed works. for the second part i might have done this

```
if not isinstance(server_info,str):
    server_info = server_info.decode('ASCII', errors='ignore')
```

jsaul (approved):
Fixes the issue for me.

megies:
> for the second part i might have done this
> 
> ```
> if not isinstance(server_info,str):
>     server_info = server_info.decode('ASCII', errors='ignore')
> ```

I agree that might be cleaner.. I think all the fragile ugliness in the Py2+3 transition phase when it came to `isinstance(..., (str, new_str, ...))` has made me avoid that construct for checking for `str` once and for all :sweat_smile:

megies:
So after looking at #3320 again on this branch, it looked like the original "string decode" problem was fixed here but we still ran into another problem running the example code from #3320 original post:

```python
File ~/git/obspy/obspy/clients/fdsn/client.py:1482, in Client._download(self, url, return_string, data, use_gzip, content_type)
   1480 if content_type:
   1481     headers['Content-Type'] = content_type
-> 1482 code, data = download_url(
   1483     url, opener=self._url_opener, headers=headers,
   1484     debug=self.debug, return_string=return_string, data=data,
   1485     timeout=self.timeout, use_gzip=use_gzip)
   1486 raise_on_error(code, data)
   1487 return data

File ~/git/obspy/obspy/clients/fdsn/client.py:1902, in download_url(url, opener, timeout, headers, debug, return_string, data, use_gzip)
   1900 # Catch HTTP errors.
   1901 except urllib_request.HTTPError as e:
-> 1902     error_data = e.read()
   1903     if debug is True:
   1904         msg = "HTTP error %i, reason %s, while downloading '%s': %s" % \
   1905               (e.code, str(e.reason), url, error_data)

File ~/miniconda3/envs/master/lib/python3.10/tempfile.py:477, in _TemporaryFileWrapper.__getattr__(self, name)
    473 def __getattr__(self, name):
    474     # Attribute lookups are delegated to the underlying file
    475     # and cached for non-numeric results
    476     # (i.e. methods are cached, closed and friends are not)
--> 477     file = self.__dict__['file']
    478     a = getattr(file, name)
    479     if hasattr(a, '__call__'):

KeyError: 'file'
```

some kind of weird behavior of the `HTTPError` acting like a file-like object (built on tempfile?!) but then failing to read anything because probably no extra information in this case. Hope the above commit fixes it.

megies:
Tried even harder to get more details into error message after the above example from our forum didn't show proper details. Here's an example that is improved with last commit:

```python
from obspy import UTCDateTime

client = Client("EMSC")

starttime = UTCDateTime("2023-09-08T00:00:00")
endtime = starttime + 86400

coords = (-10000.3, 29.7, -6.6, 32.2)
cat = client.get_events(starttime=starttime, endtime=endtime, minlatitude=-100000.123)
```

megies:
Should be good to go

PR #3361, opened 2023-10-26
Stream dayplot: Fix some issues with x tick positioning and labeling

megies:
### What does this PR do?

This PR fixes two partially interacting problems with how Stream "dayplots" (helicorder plots) are ticked and labeled on the x-Axis.

#### 1. Tick Positioning

When "number_of_ticks" is not set, in many cases with sensible choice of "interval" parameter, bad automatic choice of number of x ticks were made. E.g. with "interval=10", 10 x ticks were used (should be 11), so that all ticks fall on decimal numbers.

#### 2. Tick Labeling

The labeling of x-ticks was forced to `"%i" % x`, i.e. forcibly cutting of any decimals. In principle it is desirable to show "3" instead of "3.0" for the ticks but combined with the former problem leading to many cases of oddly placed ticks this was leading to plots having wrong labeling in many cases (e.g. labels 0, 1, 2, 3, ..., 12, 13, 15 with ticks not actually placed at these integers numbers). 

##### 1. Automatic routine selecting number of ticks giving picking bad number of ticks

Notice how tick labels jump from "8" to "10" on last tick label. Ticks are actually *not* at the integer number positions that are suggested by the labels, but rather oddly placed due to only 10 ticks for a range 0-10 (should be 11 ticks).

```python
from obspy import UTCDateTime
from obspy.imaging.tests.test_waveform import TestWaveformPlot
start = UTCDateTime(0)
st = TestWaveformPlot()._create_stream(start, start + 3 * 3600, 100)

st.plot(type='dayplot', interval=10)
```

**Before:**
Figure_1

**After:** 11 ticks are chosen and now ticks actually are at the integer positions
Figure_1a

##### 2. User chooses weird number of ticks and labeling is wrong

User chooses weird parameters, but at first glance the plot looks OK, since it does not show correct labels, all labels are decimal numbers at weird positions.

```python
st.plot(type='dayplot', number_of_ticks=10)
```

**Before:**
Figure_2
**After:** ticks are still placed oddly due to user explicit choice, but now it is clear that ticks are off
Figure_2a

Got contacted in private mail with an example of this problem

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/imaging/tests/test_waveform.py
+++ b/obspy/imaging/tests/test_waveform.py
@@ -11,63 +11,64 @@ from obspy.core.util import AttribDict
 import pytest
 
 
+def _create_stream(starttime, endtime, sampling_rate):
+    """
+    Helper method to create a Stream object that can be used for testing
+    waveform plotting.
+
+    Takes the time frame of the Stream to be created and a sampling rate.
+    Any other header information will have to be adjusted on a case by case
+    basis. Please remember to use the same sampling rate for one Trace as
+    merging and plotting will not work otherwise.
+
+    This method will create a single sine curve to a first approximation
+    with superimposed 10 smaller sine curves on it.
+
+    :return: Stream object
+    """
+    time_delta = endtime - starttime
+    number_of_samples = int(time_delta * sampling_rate) + 1
+    # Calculate first sine wave.
+    curve = np.linspace(0, 2 * np.pi, number_of_samples // 2)
+    # Superimpose it with a smaller but shorter wavelength sine wave.
+    curve = np.sin(curve) + 0.2 * np.sin(10 * curve)
+    # To get a thick curve alternate between two curves.
+    data = np.empty(number_of_samples)
+    # Check if even number and adjust if necessary.
+    if number_of_samples % 2 == 0:
+        data[0::2] = curve
+        data[1::2] = curve + 0.2
+    else:
+        data[-1] = 0.0
+        data[0:-1][0::2] = curve
+        data[0:-1][1::2] = curve + 0.2
+    tr = Trace()
+    tr.stats.starttime = starttime
+    tr.stats.sampling_rate = float(sampling_rate)
+    # Fill dummy header.
+    tr.stats.network = 'BW'
+    tr.stats.station = 'OBSPY'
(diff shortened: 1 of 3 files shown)


megies:
Test fail is unrelated, should be good to go

obspy/imaging/waveform.py
```diff
@@ -1011,24 +1011,43 @@ def __dayplot_set_x_ticks(self, *args, **kwargs):  # @UnusedVariable
...
             else:
-                count = 10
                 for _i in range(15, 1, -1):
                     if time_value % _i == 0:
-                        count = _i
+                        count = _i + 1
                         break
+                else:
+                    # these are really weird cases with interval being
+                    # relactively large primes like "interval=17" and will
```
QuLogic, line 1024:
```suggestion
                    # relatively large primes like "interval=17" and will
```

obspy/imaging/tests/test_waveform.py
```diff
@@ -11,63 +11,64 @@
...
+    # To get a thick curve alternate between two curves.
+    data = np.empty(number_of_samples)
+    # Check if even number and adjust if necessary.
+    if number_of_samples % 2 == 0:
+        data[0::2] = curve
+        data[1::2] = curve + 0.2
+    else:
+        data[-1] = 0.0
+        data[0:-1][0::2] = curve
+        data[0:-1][1::2] = curve + 0.2
```
QuLogic, line 44: I know this is just the original code, but isn't it basically:
```suggestion
    # To get a thick curve alternate between two curves.
    curve[1::2] += 0.2
    # Check if even number and adjust if necessary.
    if number_of_samples % 2 == 1:
        curve[-1] = 0.0
```
(and then `data == curve`.)

QuLogic (approved):

obspy/imaging/tests/test_waveform.py
megies: looks like data is twice as long as curve and then curve gets squashed in twice there? in any case, not worth spending time on beautifying this piece of super old test code, I'd say. :)

obspy/imaging/waveform.py
megies: done, thanks!

PR #3370, opened 2023-11-15
Support for reading CyberShake seismograms.

### LINKED ISSUE #3359: Reading CyberShake seismograms in obspy
### Problem

Hi, is there any interest to add an obspy reader for CyberShake seismogram data? These are synthetic seismograms used for seismic hazard analysis. The documentation for the format is here: 

https://strike.scec.org/scecpedia/Accessing_CyberShake_Seismograms#Seismogram_Format

and that page contains some example Python code used to read the data. It'd be nice to be able to load these seismograms directly into traces using obspy `read`

### Proposed solution

_No response_

jrekoske:
### What does this PR do?

This PR adds a new submodule, `obspy.io.cybershake`, to provide support for reading the CyberShake seismogram format which is described here: https://strike.scec.org/scecpedia/Accessing_CyberShake_Seismograms#Seismogram_Format.
I also added a test and some documentation for the new feature. This is my first contribution to obspy so I added my name to the `CONTRIBUTORS.txt` and `CODEOWNERS` and I'm making this a draft pull request first to check that the documentation builds correctly.

This closes #3359 

### PR Checklist
- [x] All tests still pass.


jrekoske:
I can't add the build_docs label to this PR since I don't have write access. Could this be added to the PR? Thanks.

obspy/io/cybershake/README.txt
```diff
@@ -0,0 +1,27 @@
+package obspy.io.cybershake
+======================
```
megies, line 2: needs length adjusted

obspy/io/cybershake/__init__.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+"""
+obspy.io.cybershake - CyberShake read support for ObsPy
+==========================================================
```
megies, line 4: needs length adjusted

obspy/io/cybershake/core.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+import struct
+import numpy as np
+from obspy import Trace, Stream
+
+
+NET_CODE = 'CS'  # Assign a network code for CyberShake data
+BAND_INST_CODE = 'MX'  # M indicates mid-period and X indicates synthetic data
+LOCATION_CODE = '00'
```
megies, line 13: Are these values good for all data that might be in this file format? Another option would be to add these as kwargs to `_read_cybershake()`, then users could pass custom values through via a `read()` call.
(Even though people could of course always change that later on modifying the traces Stats attributes)

Also there is no real reason to save space in the length of these variable names here, so it would be good to have fully explicit names `NETWORK_CODE` and `BAND_INSTRUMENT_CODE` to not leave any room for ambiguity. Explicit is better than implicit, see Python Zen :)

obspy/io/cybershake/core.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    try:
+        fp_in = open(filename, "rb")
+        header_str = fp_in.read(56)
+        split_point = header_str[0:8].find(b'\x00')
+        version = header_str[0:split_point].decode()
+        assert version == "12.10"
+        return True
+    except Exception:
+        return False
```
megies, line 33: This needs a (see [our test runner that checks for warnings throwing an exception](https://tests.obspy.org/136240/#7)).. 

```python
finally:
    fp_in.close()
```

..or alternatively, which is easier to read, probably with context manager

```python
    try:
        with open(filename, "rb") as fp_in:
            header_str = fp_in.read(56)
        split_point = header_str[0:8].find(b'\x00')
        version = header_str[0:split_point].decode()
        assert version == "12.10"
        return True
```

As a second thought, this will only work with a filename provided as a string. Ideally, the `_is_...` and `_read_...` function also accept things like open file handles or `io.BytesIO` objects, which gives users more flexibility. You can look at e.g. `obspy/io/sac/core.py` for an example with the `_is_sac` and `_internal_is_sac` routines. (In case of the file type checker routine that involves returning the file pointer to it's original state at the end)
But I'll leave it up to you if you wanna go the distance here.

(ideally we should unify this at some point for all plugins, maybe through a decorator, but nobody got to that yet)

obspy/io/cybershake/tests/test_core_cybershake.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Testing reading of CyberShake seismograms.
+    """
+    def test_read_cybershake(self, testdata):
+        valid_file = testdata['test.grm']
+        invalid_file = testdata['not_cybershake.grm']
+        # Check that two files are/are not valid CyberShake data
+        assert _is_cybershake(valid_file)
+        assert not _is_cybershake(invalid_file)
+
```
megies, line 20: please separate the test function into two here (`test_is_cybershake()` and `test_read_cybershake()`) for more granularity in tests, testing "is()" and "read()" is two separate functionalities.

obspy/io/cybershake/tests/test_core_cybershake.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
megies, line 1: usually we just call this file `test_core.py` it being clear it's about cybershake by the folder structure. just cosmetics but might as well stay in line.

megies (changes requested):
I have some comments, most of them very minor and a few of them with potential improvements.. up to you if you wanna go any further here or leave it like this.

Oh and a one-liner in changelog is missing still.

Overall, I gotta say insanely good job on this PR, this is one of the cleanest PR submissions ever and everything in real good shape already. :rocket: :exploding_head:

megies:
Oh and one last comment: Might be good to just copy the format info from that linked wiki and dump it into a text file under `io/cybershake/docs/` (the "data organization" part talking about the whole file format and the block on the header part). So that it's easier to look up later directly in our source in case that wiki moves or goes away.

obspy/io/cybershake/core.py
jrekoske: It is possible one could have data using a different period band, so I changed it to allow for a user-specified `BAND_CODE` but the default is still 'M'. I also spelled out the full variable names now.

obspy/io/cybershake/core.py
jrekoske: Okay I changed this now to allow for file-like objects following the `obspy/io/sac/core.py` example

jrekoske:
Thanks for the review @megies, I addressed your comments in these last three commits.

CHANGELOG.txt
```diff
@@ -15,6 +15,8 @@ Changes:
    * Natural Resources Canada (NRCAN) added to list of known clients
  - obspy.io:
    * add read and write support for CSV, EVENTTXT and CSZ formats (see #3285)
+ - obspy.io.cybershake:
+   * add read support for CyberShake seismogram format
```
megies, line 19: please add PR number in brackets

megies:
One last thing: The `_is_..` format checker needs to reposition the pointer to the original position if a file like object is provided by the user, compare SAC looks for `.tell()` at top and and `.seek(...)` at the end. Otherwise e.g. multiple back-to-back automatic format checks can't be done. In the SAC case it is in `_internal_is_..` but it could also be in the `_is_...` since it really only applies to the case of a file like object already provided and in the case of a filename the file gets closed at the end anyway.

megies:
Thanks for the contribution, very clean first PR! :rocket:

PR #3379, opened 2023-12-01
Respect `method='full'` for `Stream.plot(type='section')`

### LINKED ISSUE #3375: `Stream.plot(type='section')` does not respect `method='full'`
### Avoid duplicates

### Bug Summary

When using the Stream plotting method with `type='section'`, no matter the `method` used the data are resampled if they exceed 10,000 samples, see here in the `__init__()` method for `WaveformPlotting`:
https://github.com/obspy/obspy/blob/00122ed909bf6a488ed25d279a83dbea10a6599a/obspy/imaging/waveform.py#L125-L129
and in `__sect_init_traces()`:
https://github.com/obspy/obspy/blob/00122ed909bf6a488ed25d279a83dbea10a6599a/obspy/imaging/waveform.py#L1235-L1239

It seems to me that the latter bit of code should be checking if `method='full'` and if so, setting `tmp_data = tr.data` — rather than resampling. Currently that kwarg is not checked. Therefore, as it stands, users have no way of fully plotting the data without resampling if they're using `type='section'`.

I'm happy to make a PR to address this.

### Code to Reproduce

_No response_

### Error Traceback

_No response_

### ObsPy Version?

1.4.0

### Operating System?

macOS (M1)

### Python Version?

3.11.5

### Installation Method?

conda

liamtoney:
### What does this PR do?

For section-type plots: Ensures that the full Trace is plotted when a user requests it to be (i.e., when specifying `method='full'`). Parallels the [logic implemented in the `WaveformPlotting.plot()` method](https://github.com/obspy/obspy/blob/maintenance_1.4.x/obspy/imaging/waveform.py#L392-L406).

Fixes #3375.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -36,6 +36,8 @@ Changes:
      reduced number of points plotted with matplotlib >=3.6 (see #3318)
    * fix some issues with bad x-tick positioning and labeling in dayplots
      (see #3361)
+   * fix a bug in Stream.plot() where the plotting method argument (e.g.,
+     'fast', 'full') was ignored for section-type plots
  - obspy.io.gcf:
    * Fixed an issue in algorithm to split and encode last few data into GCF 
      blocks (see #3252)
--- a/obspy/imaging/tests/test_waveform.py
+++ b/obspy/imaging/tests/test_waveform.py
@@ -479,8 +479,8 @@ class TestWaveformPlot:
 
     def test_section_max_npts(self):
         """
-        Check that plotting with method='full' is respected for type='section' if points
-        are greater than max_npts.
+        Check that plotting with method='full' is respected for type='section'
+        if points are greater than max_npts.
         """
         starttime = UTCDateTime(0)
         endtime = starttime + 10009
--- a/obspy/imaging/tests/test_waveform.py
+++ b/obspy/imaging/tests/test_waveform.py
@@ -476,3 +476,15 @@ class TestWaveformPlot:
         st.plot(outfile=image_path, type='dayplot',
                 timezone='EST', time_offset=-5,
                 events=cat)
+
+    def test_section_max_npts(self):
+        """
+        Check that plotting with method='full' is respected for type='section' if points
+        are greater than max_npts.
+        """
+        starttime = UTCDateTime(0)
+        endtime = starttime + 10009
+        st = _create_stream(starttime, endtime, 1)  # st[0].stats.npts = 10010
+        st[0].stats.distance = 0  # So that the section plotting works
+        fig = st.plot(type='section', method='full')
+        assert fig.axes[0].lines[0].get_xdata().size == 10010
--- a/obspy/imaging/waveform.py
+++ b/obspy/imaging/waveform.py
@@ -1252,10 +1252,20 @@ class WaveformPlotting(object):
         self._tr_delta = np.empty(self._tr_num)
         # TODO dynamic DATA_MAXLENGTH according to dpi
         for _i, tr in enumerate(self.stream):
-            if len(tr.data) >= self.max_npts:
+            method_ = self.plotting_method
+            if method_ is None:
+                if len(tr.data) >= self.max_npts:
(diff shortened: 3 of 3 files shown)


liamtoney:
Demo:
```python
from obspy import read
import numpy as np

# Read in example data
st = read()

# Pretend that the three components are actually traces
# from different stations at 10, 20, 30 km from source
for tr, distance in zip(st, [10 * 1000, 20 * 1000, 30 * 1000]):
    tr.stats.distance = distance  # [m]

# Artifically increase the number of points in each trace to exceed max_npts
# (this makes the resampling for 'fast' clear)
for tr in st:
    tr.data = np.tile(tr.data, 20)

# Plot record section
fig = plt.figure()
st.plot(fig=fig, type='section', orientation='horizontal', method='fast', alpha=1, color='red')
st.plot(fig=fig, type='section', orientation='horizontal', method='full', alpha=1)
fig.axes[0].set_xlim(5, 10)  # Zoom in to show effect of the different plotting methods
fig.show()
```
out

liamtoney:
> seems like a nice fix! i think you may have to rebase to the current maintenance branch though

Sorry, I don't understand. The current maintenance branch is [`maintenance_1.4.x`](https://github.com/obspy/obspy/tree/maintenance_1.4.x), no?

filefolder:
> > seems like a nice fix! i think you may have to rebase to the current maintenance branch though
> 
> Sorry, I don't understand. The current maintenance branch is [`maintenance_1.4.x`](https://github.com/obspy/obspy/tree/maintenance_1.4.x), no?

ahh you're right sorry! i have lost track of the passage of time between versions.

liamtoney:
Hi folks, just checking if there is more feedback on this PR. Thanks.

filefolder (approved):
looks good to me and tested fine on my machine

liamtoney:
> looks good to me and tested fine on my machine

Great! Looks like this is ready to be merged, then? I can't do that part :)

megies:
sorry for the delay @liamtoney.. looks good to me, just two comments:

1. would be nice to have a minimal regression test, easy to do with just a few lines using e.g. `np.random.random_integers(-100, 100, 10010)` as a single Trace's data and then plotting it and checking the resulting `Line` object, something like `assert len(ax.lines[0].get_xdata()) == 10010` (to make sure point number is not reduced to `max_npts`.

2. I was wondering why the min-max plot was looking that off from the original data.. turns out it's doing something different completely from the min-max routine on normal wiggle plot. I think it might be good to refactor the `_plot_min_max` routine a little bit to factor out the calculation routine for x/y values and then use the same routine in the section plot.

Would be nice to still get 1. in here, but 2. could be something for a different PR / another time.

liamtoney:
> 2\. I was wondering why the min-max plot was looking that off from the original data.. turns out it's doing something different completely from the min-max routine on normal wiggle plot. I think it might be good to refactor the `_plot_min_max` routine a little bit to factor out the calculation routine for x/y values and then use the same routine in the section plot.

Are you referring to the fact that the section plot resamples the data while the standard plot does the min/max? Maybe a separate PR for that?

liamtoney:
> 1. would be nice to have a minimal regression test, easy to do with just a few lines using e.g. `np.random.random_integers(-100, 100, 10010)` as a single Trace's data and then plotting it and checking the resulting `Line` object, something like `assert len(ax.lines[0].get_xdata()) == 10010` (to make sure point number is not reduced to `max_npts`.

Nice test idea! I implemented that and am ready for re-review, I think.

megies:
> Are you referring to the fact that the section plot resamples the data while the standard plot does the min/max? Maybe a separate PR for that?

Yes, I was assuming it was doing the same min-max thing we have for the regular wiggle plot and was confused why the line was overshooting in some points and stuff. And yes, definitely something for another time, if at all. It's not a big deal really.

megies (approved):
Thanks for the fix! :rocket:

PR #3387, opened 2023-12-21
Ppsd fix processing time info

megies:
### What does this PR do?

Fixes a problem that can lead to every second time slice not getting processed in PPSD. I'll leave this as a draft for now, since the issue mostly seems to come from an issue with the core modules, and still looking into that.

See https://discourse.obspy.org/t/ppsd-plot-temporal-doesnt-work/1846/11

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -49,6 +49,9 @@ Changes:
  - obspy.io.zmap:
    * Fix writing events with origins that have no depth or depth errors given
      (see #3343)
+ - obspy.signal:
+   * PPSD: Fix some time slices getting skipped if sampling rate does not align
+     well with the length of one PSD slice. (see #3387)
  - obspy.taup:
    * bugfix to allow calculations for a model with no discontinuities
      (see #3070, #3244)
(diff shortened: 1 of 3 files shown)


megies:
So, the origin of the problem is actually not a problem of obspy core functionality, but rather of the SAC format it seems, which the data from that discourse post was stored in.

I do not know SAC well, but looking at the manual, the information on how samples are spaced in time is stored as the "delta" between samples as a single-precision 32-bit float. So this is running into floating point accuracy issues quite fast. Maybe our old, original SAC reader implementation was working around that somehow, but currently that seems to be an issue. The data here has a sampling rate of 25 Hz, so delta is `0.04` seconds. This is basically what is happening:

```python
# our sac plugin code readin the SAC float header section, here the sample spacing "delta"
delta = np.frombuffer(b'\n\xd7#=', np.float32)[0]
```

For a continuous trace 24h trace (25Hz, 2mil samples) this leads to the endtime being calculated off by about 15% of the sample spacing.

megies:
So even though the issue came up with data with a weird sampling rate, this still should get fixed.
PR is changing it so the proper start time of each selected time window is inserted in the list of processed data times as opposed to the first sample timing of the actual cut out time slice. This could lead to a mismatch when checking if the next time slice should be processed or not (i.e. being considered "already processed and should be skipped").

Issue #2376, opened 2019-04-10
libmseed 3.0.0 is now available


QuLogic:
It includes breaking changes; there is a [guide for porting from 2 to 3](https://iris-edu.github.io/libmseed/porting-guide.html), though I have only lightly glanced through it.

krischer:
I've written a basic plug-in for this here: https://github.com/krischer/libmseed3_obspy_plugin

I'm not entirely sure how stabilized the format is so I'd wait a bit until we integrate it into ObsPy.

chad-iris:
libmseed 3 now supports reading and writing version 2 of miniSEED so, in general, can be used in place of libmseed 2.x.  The "default" format is version 3 for libmseed, and since that is not yet approved by the FDSN is not recommended to be a "default" format for ObsPy.  That may change soon, but still is likely best for a major version release of ObsPy.

megies:
My feeling is, it's OK to have the current libmseed for the next releases 1.2.0 (py2+3) and 2.0.0 (Py3 only - same functionality as 1.2.0).
We can put libmseed 3.0 update on the roadmap for 2.1

megies:
Not sure what the status on this one is.. @chad-iris

zadeviggers:
The lack of miniSEED v3 support is an issue for us at CRISiSLab. Ideally our researchers should be able to use their usual tools (i.e. obspy) to directly parse and analyse the data generated by our systems without needing to run an extra transform script each time to make the data compatible.

chad-earthscope:
> Not sure what the status on this one is.. @chad-iris

libmseed v3 is now stable, with the new miniSEED v3 format being officially adopted: https://docs.fdsn.org/projects/miniseed3/en/latest/

Releases are here:
https://github.com/EarthScope/libmseed/releases

A guide to porting software from v2 to v3 is here:
https://earthscope.github.io/libmseed/porting-guide.html

There is some advanced functionality added to parse specified files into contiguous traces with associated "Record Lists", which allows a) a coherent determination of traces by libmseed with specifiable tolerances, b) a list of individual records that go into each trace if such details are needed, and c) ability to read the data without unpacking/decompressing and later uncompressing only what is desired, aka a lazy load of sorts, and optionally _to a specified memory buffer_.  I figured this sort of reading would be good for frameworks like ObsPy.

[...]

PR #3411, opened 2024-02-26
Mseed3 read/write support

crotwell:
### What does this PR do?

*Adds read and write support for miniseed3.*

*Read and write the new FDSN miniseed3 format.*

FDSN [Specification](http://docs.fdsn.org/projects/miniseed3/en/latest/)

I set this up to use my [simplemseed](https://pypi.org/project/simplemseed/) pypi package. Most other obspy.io subpackages seem to not have external dependencies, so was unsure how to handle this. Simplemseed has a dependency of crc32c, so not as simple as just copying files in, although I guess it is possible. 

Also likely worth bringing @chad-earthscope into the conversation as there may be advantages to using mseed3 from new version of his libmseed C library. Personally, I like having a pure python implementation as it makes long term maintenance easier than via a C library, but opinions may vary.

Because the new json "extra headers" in miniseed3 are very flexible, it may be worth thinking about automatically copying more from the Trace Stats object. I have followed other io formats by having "mseed3"
inside of the stats, but that may be an extra layer that doesn't actually help anyone. It would be relative easy to just copy anything from stats that is not a standard part of the header into the extra headers. Its just json after all.

Code for parsing [FDSN source ids](http://docs.fdsn.org/projects/source-identifiers/en/v1.0/) is used internally, see [code here](https://github.com/crotwell/simplemseed/blob/main/src/simplemseed/fdsnsourceid.py). Not sure if it would be helpful to expose this further within obspy for dealing with FDSN source ids outside of miniseed3?

This code also creates longer network, station, etc in stats, which may break code that assumes the seed length limits. For example a SID of `FDSN:XX2024_REALFAKE_01234567_H_HRQ_Z` becomes:
```
         network: XX2024
         station: REALFAKE
        location: 012345
         channel: H_HRQ_Z
       starttime: 2024-01-02T15:13:55.123456Z
         endtime: 2024-01-02T15:13:58.123456Z
   sampling_rate: 1.0
           delta: 1.0
            npts: 4
           calib: 1.0
         _format: MSEED3
          mseed3: AttribDict({'pubVer': 0, 'eh': AttribDict({})})
``` 

Believe I have checked all the items for submission, but happy to revise if missed anything. 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/mseed3/core.py
+++ b/obspy/io/mseed3/core.py
@@ -23,6 +23,7 @@ from simplemseed import (
 )
 import simplemseed
 
+OBSPY_FORMAT_MSEED3 = "MSEED3"
 MSEED_STATS_KEY = "mseed3"
 PUB_VER_KEY = "publication_version"
 EX_HEAD_KEY = "eh"
@@ -347,7 +348,7 @@ def mseed3_to_obspy_header(ms3):
         stats["location"] = nslc.locationCode
         stats["channel"] = nslc.channelCode
     except:
-        print(f"Oops, can't extract FDSN Sid: ${ms3.identifier}")
+        #Oops, can't extract FDSN Sid, put all in channel???
         stats["channel"] = ms3.identifier
     micros = int(round(h.nanosecond / 1000))
     stats["starttime"] = UTCDateTime(
--- a/obspy/io/mseed3/tests/test_mseed3_read_write.py
+++ b/obspy/io/mseed3/tests/test_mseed3_read_write.py
@@ -1,12 +1,17 @@
 import json
 from tempfile import NamedTemporaryFile
 
-from simplemseed import FDSNSourceId, MSeed3Record, MSeed3Header
+from simplemseed import (
+    FDSNSourceId,
+    MSeed3Record,
+    MSeed3Header,
+    )
 from obspy.io.mseed3.core import  (
     ObsPyMSEED3DataOverflowError,
     MSEED_STATS_KEY,
     PUB_VER_KEY,
     UNKNOWN_PUBLICATION_VERSION,
+    OBSPY_FORMAT_MSEED3
 )
 
 from obspy import read, Stream, Trace
@@ -28,7 +33,7 @@ class TestMSEED3ReadingAndWriting:
                 # 0 not decompressible, can't return empty stream so???
                 continue
             jsonpath = path.parent.joinpath(path.name[:-6] + "json")
-            stream = read(path, format="MSEED3")
+            stream = read(path, format=OBSPY_FORMAT_MSEED3)
             assert len(stream) == 1
             trace = stream[0]
             with open(jsonpath, "r") as injson:
@@ -59,7 +64,7 @@ class TestMSEED3ReadingAndWriting:
         Read file test via L{obspy.core.Stream}.
         """
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         assert len(stream) == 6
         assert stream[0].stats.network == "CO"
         assert stream[0].stats.station == "BIRD"
@@ -74,52 +79,52 @@ class TestMSEED3ReadingAndWriting:
 
     def test_write_int32(self, datapath):
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         with NamedTemporaryFile() as tf:
             outfile = tf.name
-            stream.write(outfile, format="MSEED3", encoding="INT32")
-            redo = read(outfile, format="MSEED3")
+            stream.write(outfile, format=OBSPY_FORMAT_MSEED3, encoding="INT32")
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
             self._check_bird_jsc(stream, redo)
             assert stream[0].data.dtype == np.int32
             assert redo[0].data.dtype == np.int32
 
     def test_write_float32(self, datapath):
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         with NamedTemporaryFile() as tf:
             outfile = tf.name
-            stream.write(outfile, format="MSEED3", encoding="FLOAT32")
-            redo = read(outfile, format="MSEED3")
+            stream.write(outfile, format=OBSPY_FORMAT_MSEED3, encoding="FLOAT32")
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
         self._check_bird_jsc(stream, redo)
         assert redo[0].data.dtype == np.float32
 
     def test_write_float64(self, datapath):
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         with NamedTemporaryFile() as tf:
             outfile = tf.name
-            stream.write(outfile, format="MSEED3", encoding="FLOAT64")
-            redo = read(outfile, format="MSEED3")
+            stream.write(outfile, format=OBSPY_FORMAT_MSEED3, encoding="FLOAT64")
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
         self._check_bird_jsc(stream, redo)
         assert redo[0].data.dtype == np.float64
 
     def test_write_steim1(self, datapath):
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         with NamedTemporaryFile() as tf:
             outfile = tf.name
-            stream.write(outfile, format="MSEED3", encoding="STEIM1")
-            redo = read(outfile, format="MSEED3")
+            stream.write(outfile, format=OBSPY_FORMAT_MSEED3, encoding="STEIM1")
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
         self._check_bird_jsc(stream, redo)
         assert redo[0].data.dtype == np.int32
 
     def test_write_steim2(self, datapath):
         testfile = datapath / "bird_jsc.ms3"
-        stream = read(testfile, format="MSEED3")
+        stream = read(testfile, format=OBSPY_FORMAT_MSEED3)
         with NamedTemporaryFile() as tf:
             outfile = tf.name
-            stream.write(outfile, format="MSEED3", encoding="STEIM2")
-            redo = read(outfile, format="MSEED3")
+            stream.write(outfile, format=OBSPY_FORMAT_MSEED3, encoding="STEIM2")
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
         self._check_bird_jsc(stream, redo)
         assert redo[0].data.dtype == np.int32
 
@@ -157,15 +162,15 @@ class TestMSEED3ReadingAndWriting:
             outfile = tf.name
             with open(outfile, 'wb') as out:
                 out.write(recordBytes)
-            st = read(outfile, format="MSEED3")
+            st = read(outfile, format=OBSPY_FORMAT_MSEED3)
             assert len(st) == 1
             tr = st[0]
             assert tr.stats.network == net
             assert tr.stats.station == sta
             assert tr.stats.location == loc
             assert tr.stats.channel == f"{band}_{s}_{subs}"
-            st.write(outfile, format="MSEED3")
-            redo = read(outfile, format="MSEED3")
+            st.write(outfile, format=OBSPY_FORMAT_MSEED3)
+            redo = read(outfile, format=OBSPY_FORMAT_MSEED3)
             assert len(redo) == 1
             redotr = redo[0]
             assert redotr.stats.network == net
@@ -190,7 +195,7 @@ class TestMSEED3ReadingAndWriting:
             outfile = tf.name
             with open(outfile, 'wb') as out:
                 out.write(recordBytes)
-            st = read(outfile, format="MSEED3")
+            st = read(outfile, format=OBSPY_FORMAT_MSEED3)
             assert len(st) == 1
             tr = st[0]
             assert tr.stats.network == ""
@@ -217,17 +222,17 @@ class TestMSEED3ReadingAndWriting:
         stream = Stream([tr])
         with pytest.raises(Exception):
             with NamedTemporaryFile() as tf:
-                stream.write(tf.name, format="MSEED3", encoding="STEIM1")
+                stream.write(tf.name, format=OBSPY_FORMAT_MSEED3, encoding="STEIM1")
         with pytest.raises(Exception):
             with NamedTemporaryFile() as tf:
-                stream.write(tf.name, format="MSEED3", encoding="STEIM2")
+                stream.write(tf.name, format=OBSPY_FORMAT_MSEED3, encoding="STEIM2")
 
     def test_guess_encoding(self, datapath):
         tr = Trace(np.array([1, 2, 3, -17], dtype="int64"))
         stream = Stream([tr])
         # guess output encoding
         with BytesIO() as buf:
-            stream.write(buf, format="MSEED3")
+            stream.write(buf, format=OBSPY_FORMAT_MSEED3)
             buf.seek(0)
             in_stream = read(buf)
         assert in_stream[0].data.dtype == np.int32
@@ -236,7 +241,7 @@ class TestMSEED3ReadingAndWriting:
         stream = Stream([tr])
         # guess output encoding
         with BytesIO() as buf:
-            stream.write(buf, format="MSEED3")
+            stream.write(buf, format=OBSPY_FORMAT_MSEED3)
             buf.seek(0)
             in_stream = read(buf)
         assert in_stream[0].data.dtype == np.float32
@@ -245,7 +250,7 @@ class TestMSEED3ReadingAndWriting:
         stream = Stream([tr])
         # guess output encoding
         with BytesIO() as buf:
-            stream.write(buf, format="MSEED3")
+            stream.write(buf, format=OBSPY_FORMAT_MSEED3)
             buf.seek(0)
             in_stream = read(buf)
         assert in_stream[0].data.dtype == np.float64
@@ -269,4 +274,4 @@ class TestMSEED3ReadingAndWriting:
         stream = Stream([tr])
         with pytest.raises(ObsPyMSEED3DataOverflowError):
             with NamedTemporaryFile() as tf:
-                stream.write(tf.name, format="MSEED3", encoding="INT32")
+                stream.write(tf.name, format=OBSPY_FORMAT_MSEED3, encoding="INT32")


megies:
Let it be known.. I have been feeling real bad for a really long time for not having commented on this one yet.

Was kind of trying to get 1.4.1 out and now trying to get a good 1.5.0 out.

I am a bit wary of adding a dependency and we had a prototype for incorporating libmseed v3 but it's not fully finished and to me alone I can't really judge how much work it is to finish.

I am kinda thinking to keep pushing 1.5.0 out fast so that all these additions that have been in master for a long time not get stalled in master further, and then maybe do a full freeze on everything else and just figure out mseed v3 for a 1.6.0 (or 2.0) release..

crotwell:
No worries, understand you are shouldering a big load and trying to get releases out the door makes sense. I am advocating for mseed3 as a format, but understand it will take time. I joked that it would probably take a decade to get it out there when we meet in Utrecht in 2016, and it is looking like that estimate maybe was too short! :)

megies:
Here's the options we are looking at right now..

 1. vendor libmseed v3 like we used to do for libmseed v2, basically finalizing https://github.com/obspy/obspy/issues/2376#issuecomment-481757805 and integrating into obspy, potentially for now simply alongside the existing libmseed v2 (this is what I feel like might be ideal)
 2. pursue this PR #3411, which would introduce new hard dependencies for any obspy installation whether it needs mseed 3 or not

But maybe there's a third option we/I didn't consider yet..

 3. outsource this pull request into a separate package that can be installed additionally to an existing obspy installation and that hooks in via the plugin system we have (similar to https://github.com/trichter/obspyh5)

I'm still not sure what the best way to go ahead is short to mid term. I'm probably just a bit too paranoid about adding new hard dependencies.
In any case, I don't think there's any reason we can't have both libmseed v3 and @crotwell's python implementation alongside each other as two separate read/write plugins.

megies:
I mean, I don't know this crc32c package, but it looks to be widely used and well maintained and it's packaged for conda-forge and it has zero dependencies itself, so probably there is no reason why it should mean trouble for us, considering we pretty much exclusively focus on conda packaging as our main installation path.

megies:
@crotwell is there a way to select whether STEIM1 oder STEIM2 is used as encoding (with this obspy integration)? Can't seem to find it so far. I was considering running write operations across some large data amounts to see if we run into any more numpy exceptions on numpy v2

crotwell:
Humm, looks like I did not add that, currently it writes as one of the primitive seed types, based on the numpy array type. Probably should add an "encoding" arg to be more like existing mseed writer. More soon...

crotwell:
future reference, now with encoding:
```
stream.write(filename, format="MSEED3", encoding="STEIM2")
```

megies:
@crotwell I can just repeat myself and say sorry for how long this has been left dangling. @ThomasLecocq will be visiting here in early March and we'll be working on this then, looking at all the tiny details and hoping to get it fully ready and merged and hoping to release a new obspy version with it too

megies:
> Also, just to be clear, if you believe that using a libmseed based solution for mseed3 is better for obspy, then do it. I won't be offended if you think that is a better way to go.

That was one thing that was bugging me, but to be honest, having thought about it a lot I think its best to go with your Python module. I don't see how it could be problematic in terms of dependencies etc since we its all in the conda forge ecosystem and honestly even if integrating the libmseed v3 C code might be cleaner, we don't have anybody left who would develop/maintain the interface layer we'd need anyway.

megies:
I just wanna make sure we look at all the details that might be able to break backwards compatibility (like the SEED codes lengths etc) and make sure we get all of these tiny details right but then go with the simplemseed module basically

chad-earthscope:
Despite @crotwell sending me a link to simplemseed some time ago I hadn't looked at it very closely and just now took a look and have a few thoughts contrasting it with libmseed that might be useful as you consider choices.  Philip please correct any assertions I make that are wrong, and take this as a wish list for simplemseed.

- It appears that the reader must know the miniSEED version before reading.  This will likely lead to some decoding  logic that detects the version using the first record or some such, which will at some point be wrong.  libmseed will read mixed version data and present it agnostically to the caller.  I strongly believe data users will not care which version of miniSEED is being read and when it gets mixed will be a unnecessary barrier to use.
- It looks like there is support for uncompressed data and Steim 1 and 2.  While not common, older codecs are not supported. 
- There are some non-obvious, hard-coded limitations, e.g. [only 512-4096 byte records](https://github.com/crotwell/simplemseed/blob/0c63100e8ed3e4e1a875781df522ee297efa5161/src/simplemseed/miniseed.py#L715-L721).
- Speed when compared to the libmseed-based [mseedlib Python module](https://github.com/EarthScope/mseedlib).  The below script is a really simple and naive test of reading through records with simplemseed and mseedlib, with the following results in seconds of runtime:

_Parsing records_
```
simplemseed record read time: 22.28845716700016
   mseedlib record read time: 5.140508125001361
```
_Parsing records, printing some details_
```
simplemseed record read time: 44.225525958001526
   mseedlib record read time: 16.899378582998906
```

_Parsing records, decompressing data samples_
```
simplemseed record read time: 546.8377198329981
   mseedlib record read time: 11.890655750001315
```

I'm sure some of these things can be fixed with time or I might have misunderstood.

P.S. The "Trace" family of functions in libmseed are also supported by mseedlib, which are the library's functions for reconstructing time series from records (and where libmseed shines).  I belive this is the 99% use case for miniSEED, but I believe ObsPy is still "rolling it's own" in this regard.

P.P.S.

The test code:
```Python
import timeit
import numpy

input_file = '../data/IU_COLA_00_BHZ' # 2396984320 bytes, aka 2.2G

import simplemseed

def simplemseed_records():
    with open(input_file, "rb") as inms2:
        for ms2rec in simplemseed.readMiniseed2Records(inms2):

            # Print summary
            #print(ms2rec.summary())

            # Decode data to numpy array
            #dataArray = ms2rec.decompressed()

            pass

from mseedlib import MS3RecordReader, TimeFormat

nptype = {'i': numpy.int32, 'f': numpy.float32, 'd': numpy.float64, 't': numpy.char}

def mseedlib_records():
    with MS3RecordReader(input_file, unpack_data=False) as msreader:
        for msr in msreader:

            # Print summary
            #print(f'{msr.sourceid}, {msr.reclen}, {msr.samplecnt} samples, {msr.starttime_str(timeformat=TimeFormat.ISOMONTHDAY_SPACE_Z)}')

            # Decode data to numpy array, must set unpack_data=True in MS3RecordReader
            #dataArray = numpy.frombuffer(msr.datasamples, dtype=nptype[msr.sampletype])

            pass

# Run timing tests
simplemseed_seconds = timeit.timeit("simplemseed_records()", globals=globals(), number=1)

mseedlib_seconds = timeit.timeit("mseedlib_records()", globals=globals(), number=1)


print (f"simplemseed record read time: {simplemseed_seconds}")
print (f"   mseedlib record read time: {mseedlib_seconds}")
```

crotwell:
To be clear, I was **NOT** proposing to replace the existing obspy miniseed2 reading, and although my library can read miniseed2 for most data, I don't pretend it is anywhere near as robust for all the edge cases. And to be honest I have zero interest in maintaining a miniseed2 reading code or implementing a bunch of old compression styles. My goal here is to be able to read/write miniseed3 in obspy. I had already created some python code that did it and so I put in a PR to get the ball rolling. I consider it a success since we are at least having a conversation about it! :)

Not sure I agree that mixing miniseed2 and 3 in one file is a good idea. But if obspy wants to be able to do that than my library is probably not the right choice.

I also did not worry much about optimizing speed in my code as it seemed fast enough for my uses, but that is a valid concern. On the other hand, as @megies mentioned, maintaining a mixed C-Python codebase is harder than just a Python one, so there are tradeoffs.

@chad-earthscope you have any interest in creating a PR for obspy based on mseedlib? I am happy to step aside if your solution ends up being judged the better one.

krischer:
Hi all,

for the sake of completeness there is also this really old prototype: https://github.com/krischer/libmseed3_obspy_plugin - Please ignore the actual code - it would likely take longer to bring it up to snuff compared to rewriting it.

The point I want to make is that ObsPy has a plugin system for file formats and third-party packages like `simplemseed` and `mseedlib` could hook into it. `obspy.read()` and `obspy.Stream.write()` would directly work and use the third-party packages without any required change to ObsPy. Once that is stable one could make the choice to make one of the third-party packages a default dependency or even vendor it. An additional benefit of that approach would be to split the long-term maintenance burden.

In the aforementined prototype the plugin hooks are here: https://github.com/krischer/libmseed3_obspy_plugin/blob/master/setup.py#L34-L42 - it is only a tad bit different with the more modern `pyproject.toml` files.

This should not be hard to implement for both, `simplemseed` and `mseedlib`. An interesting ObsPy change would be the introduction of a priority order for plugins.

crotwell:
Think I have a plugin of this here (stealing much from @krischer example):
https://github.com/crotwell/simplemseed_obspy_plugin

Is there a mechanism to register plugins? Or is it just publish to pypi?

megies:
> Is there a mechanism to register plugins? Or is it just publish to pypi?

We do this in the installation step using `setuptools` (look for `ENTRY_POINTS` in obspy's `setup.py`) and then these entry points get stored in a well defined manner for every installed module in the Python environment (see [here](https://packaging.python.org/en/latest/specifications/entry-points/#file-format)). Basically they get stored in ini format alongside the installed module e.g. `lib/python3.1/site-packages/obspy-1.4.1.dist-info/entry_points.txt`.

We used to (and still do) parse them with `pkg_resources` module but this module advertizes itself as deprecated nowadays and is supposed to get replaced with [Python stdlib `importlib`](https://docs.python.org/3/library/importlib.metadata.html) which has seen more and more changes recently on that matter, see [open PR on this matter too](https://github.com/obspy/obspy/pull/3333)

megies:
The main difference of integrating into obspy directly / vendoring or keeping it separate is that in the first case it is very stable (which not only has positives), i.e. if users have some code that does a specific thing on a specific obspy version, it will do the exact same thing 5 years later on that same obspy version. But at the same time updates and fixes might take a long time (or forever) to make their way to the user (via a full new major or minor obspy release).
On the other hand keeping it separate means that obspy potentially gives up some control of backwards compatibility and results for users depend on a wider variety of version numbers of packages in their environment.

What Lion is hinting at is that nobody will *or even can* keep you from setting up a plugin that acts as an I/O format (a fully fleshed out [example from @trichter here](https://github.com/trichter/obspyh5)), but for something as important as MiniSEED v3 support eventually we probably want to have at least one plugin in obspy anyway.
In terms of `simplemseed` I think having that plugin layer inside or outside of obspy makes a rather small difference anyway and I we just need to see whether we have to adjust some core routines aynway (e.g. in terms of the longer SEED codes etc).

chad-earthscope:
> Not sure I agree that mixing miniseed2 and 3 in one file is a good idea. 

FWIW I think having mixed version miniSEED in a file, object, stream, memory buffer, etc. is inevitable.  Folks will run software that gets data from a data center (e.g. all waveforms for P phase for this event within X degrees) and this will eventually include mixed versions.  Needing to separate the versions to read the data is the unnecessary barrier I referred to.

crotwell:
@megies if you are working on this now...

> Because the new json "extra headers" in miniseed3 are very flexible, it may be worth thinking about automatically copying more from the Trace Stats object. I have followed other io formats by having "mseed3"
inside of the stats, but that may be an extra layer that doesn't actually help anyone. It would be relative easy to just copy anything from stats that is not a standard part of the header into the extra headers. Its just json after all.

The more I think about this, the more I think that putting extra headers inside "mseed3" in the Stats is wrong and probably better to go 1:1 between Stats and extra headers, perhaps removing any autogenerated items in Stats. How do you feel about this idea?

crotwell:
OK, will defer to your judgement on that.

Also, I ran into problems with numpy 2 "weak_and_warn" checks under recent python (3.13) and so removed them from simplemseed. Not sure when this changed in numpy, but it causes differences in behavior and that will cause test to fail under 3.13 that pass under 3.9.

Have you figured out how to deal with this in the rest of obspy?

```
AttributeError: module 'numpy' has no attribute '_set_promotion_state'
```

megies:
We've started looking into some details and there are some things that need more thought. In any case, I think we should go ahead with this PR after some more changes to details and thus add `simplemseed` as a hard dependency (since it should not be problematic as everything is on pypi and in cond-forge). We are also looking at `mseedlib` and  at least in the mid-term (maybe even for 1.5.0???? maybe...) I think we should add that too, potentially eventually even replacing the old `io.mseed` completely, (if we can accomodate most utility stuff hardcore users might rely on?). I might give packaging `mseedlib` for conda-forge a shot and then we could get that one started too.
That would mean potentially ending up with two plugins for the same file format, which should not be a problem though.

some thoughts so far.. (sorry if a bit cobbled together..)

 - read/write plugin name might have to be something else than `MSEED3`, to avoid clashes with a potential `mseedlib` integration, maybe `SIMPLEMSEED3`?
 - same for the module, probably rename to `io.simplemseed` (or `io.simplemseed3` but the former is fine I think, doesn't have to indicate the version we want to use it for but can match the naming of the separate package)
 - even with these changes, probably the dictionary naming in Stats should stay `mseed3` since that would be interchangeable no matter which plugin is used to read/write, handling of the json stuff should be easy to keep the same
 - do we want some logic added to `Stats` to facilitate the FDSN Source ID definitions? Maybe we could even have two subclasses to `Stats`, one that is based on traditional N/S/L/C attributes and one that uses the FDSN source ID granularity and both making sure to behave mostly the same and try to be nice to everybody as best as possible. Basically elevating some of the utility stuff in this PR (with changes) to the core part, as its a FDSN definition and should be potentially prevalent in real world eventually. Some more thought needed here.. Maybe I'll draft a short PR to showcase this tomorrow, separate from this PR
 - some minor naming things. I think `Stats.mseed3.pubVer` should be changed to `publication_version` to be explicit. Having the rest of the extra headers in a dictionary `eh` seems redundant to me. I think that extra level could be flattened out. Only reason I can see is if somebody were to put something non JSON-serializable into `Stats.mseed3` but I think its fair enough to have that raise an exception.
 - not a single test file I can see anywhere across simplemseed, mseedlib and libmseed has any SEED ID / Source ID that is not adhering to the old NSLC 2/5/2/3 scheme.. (or I'm blind) which is kinda funny and not ideal I think. This needs changing for sure. ;)
 - how to handle the `FDSN:` part of the source ID needs some thought. Should other values be allowed or considered? FDSN standard does not speak of any other values being allowed (which is kinda weird.. why have that part in the first place then?) But `mseedlib` seems to be flexible here, so maybe this PR should be flexile with it too (and then again, see above, maybe we want to extend `Stats` to store it)
 - tests in this PR have no test file with empty extra headers I believe, should be changed
 - potentially we could think about adding validation of FDSN reserved headers (similar to schema validation on StationXML read/write) since there is a schema for it

crotwell:
> That would mean potentially ending up with two plugins for the same file format, which should not be a problem though.

Just wondering if having two for the same format inside obspy makes sense. I am assuming that users mostly just use the `read()` function, which means that they always use the first matching plugin? I guess I worry that the maintenance effort is doubled with two codes for little gain?

If you think a mseedlib reading/writing is advantageous, then perhaps it should the single reader for miniseed3, and as a bonus you get miniseed2 in the same package? What I am saying is I don't want to increase your workload via this PR. Mostly I just care about miniseed3 working in obspy _somehow_. 

But if you think my PR can be a bridge to a mseedlib implementation eventually, and then it will be deprecated, then that is fine with me too.

crotwell:
> how to handle the FDSN: part of the source ID needs some thought. Should other values be allowed or considered? FDSN standard does not speak of any other values being allowed (which is kinda weird.. why have that part in the first place then?) But mseedlib seems to be flexible here, so maybe this PR should be flexile with it too (and then again, see above, maybe we want to extend Stats to store it)

If memory serves, the thought was that always including it means you never have to decide when you should include it. Easy to imagine two mseed3 records one with source id of `FDSN:CO_JSC_00_L_H_Z` and another with `CO_JSC_00_L_H_Z` and having to decide if they are the same. And I think some flexibility in allowing the mseed3 record format to be used for non-seismology type things at some point in the future. 

Basically not putting it in saves 5 bytes, but gives up some benefits and flexibility. The spec should probably have said more, but it does say:

> Source identifier. A unique identifier of the source of the data contained in the record. Recommended to use URI-based identfiers. Commonly an [FDSN Source Identifier](http://docs.fdsn.org/projects/source-identifiers/).

so using a `FDSN:` style id is not required for mseed3. I suppose one could create mseed3 with `MYOWN:blabla_bla` for non-seismic uses?

crotwell:
> not a single test file I can see anywhere across simplemseed, mseedlib and libmseed has any SEED ID / Source ID that is not adhering to the old NSLC 2/5/2/3 scheme.. (or I'm blind) which is kinda funny and not ideal I think. This needs changing for sure. ;)

Yep, added a test (which found an issue). In addition to being some parts being long, the subsource can be empty, so a channel like `S_XYZ_` ie a channel without an orientation like maybe for temperature. I have pushed for simplemseed but not yet created a new pypi release to fix this.
 
> tests in this PR have no test file with empty extra headers I believe, should be changed

I think the reference data, like reference-sinusoid-int16 has extra headers length == 0. But maybe the test cases are not checking the Stats conversion for those? I thought they were but maybe not.

> some minor naming things. I think Stats.mseed3.pubVer should be changed to publication_version to be explicit. Having the rest of the extra headers in a dictionary eh seems redundant to me. I think that extra level could be flattened out. Only reason I can see is if somebody were to put something non JSON-serializable into Stats.mseed3 but I think its fair enough to have that raise an exception.

So should publication_version be inside mseed3, or higher in the Stats? If it is in mseed3, then it needs to be pulled out before write as the extra headers in a record should not have publication_version as it is part of the fixed header. I think I was just trying to avoid a remove-write-reinsert step in the code while keeping everything inside stats.mseed3, but maybe that is the right way to go. 

Perhaps publication_version directly directly in Stats and Stats.mseed3 == extra headers?
Pushed this as a change, so hopefully ok.

megies:
> Add: How would obspy deal with a timeseries id that doesn't correspond to nslc somehow? Is there a convention to put the identifier into something besides Stats.network, etc?

That's exactly what I meant with it being probably a good idea to extend Stats in that regard. I'll put up a PR against master to showcase what I mean today.



> I have pushed for simplemseed but not yet created a new pypi release to fix this.

No hurry really from obspy point of view, if a (simplemseed) release is needed we can quickly make one, but we can always just easily install from source while working on things right now.



> So should publication_version be inside mseed3, or higher in the Stats?

Nah, I'd say it belongs right where it is. Sorry I just realized my comment was a bit vague here, I was really only referring to the naming "pubVer" vs. "publication_version". To me it feels like it's kind of "old coding habits" to abbreviate variables and things like that, while nowadays and especially in Python it's generally the paradigm to rather be explicit and self explaining instead of saving some space in a code line. =)



> If it is in mseed3, then it needs to be pulled out before write as the extra headers in a record should not have publication_version as it is part of the fixed header

Yes, I think that sounds like a good way here. Grab publication version first, then write everything else as extra headers. Just inside obspy in the Stats it just feels like unneeded nesting that serves no purpose for users. Normal users won't care if some specific thing goes into a fixed header or a variable length extra header in the format implementation, and the people that are aware of these things will be able to make sense of it anyway.



> Perhaps publication_version directly directly in Stats and Stats.mseed3 == extra headers?
> Pushed this as a change, so hopefully ok.

I'd say definitely inside `Stats.mseed3` since its format specific and makes no sense for other file formats. And no worries, can always revert the change quickly. =)

crotwell:
I moved pub ver into mseed3.

>>    Add: How would obspy deal with a timeseries id that doesn't correspond to nslc somehow? Is there a convention to put the identifier into something besides Stats.network, etc?

> That's exactly what I meant with it being probably a good idea to extend Stats in that regard. I'll put up a PR against master to showcase what I mean today.

Pushed in a test case for non-FDSN: source id, I put the whole source id into channel, but that is unlikely to be the correct way to go. Can modify based on your idea.

PR #3418, opened 2024-03-11
Trace/Stream: add methods to switch byteorder

### LINKED ISSUE #3417: Convert byte order to native on reading data or merging traces?
### Problem

First posted on the forum: https://discourse.obspy.org/t/the-big-endian-and-little-endian-confilict-for-merging-seismogram/1890/3

Currently, we raise an exception when attempting to merge traces that would otherwise be mergeable but only differ on the byteorder of the underlying data array, e.g. coming from different files with data in differing byteorder.
Probably we should fix the byte order and enable merging instead.

```python
import numpy as np                                                              
from obspy import read                                                          
                                                                                
                                                                                
# make different endian sacfile                                                 
tr = read()[0]                                                                  
starttime = tr.stats.starttime                                                  
                                                                                
first_part = tr.copy()                                                          
first_part = first_part.trim(starttime, starttime + 5)                          
first_part.write("/tmp/first.sac", format="SAC", byteorder="<")                 
                                                                                
second_part = tr.copy()                                                         
second_part = second_part.trim(starttime + 5, starttime + 10)                   
second_part.write("/tmp/second.sac", format="SAC", byteorder=">")               
                                                                                
# merge them                                                                    
st = read("/tmp/first.sac")                                                     
st += read("/tmp/second.sac")                                                   
st.plot(automerge=False)                                                        
                                                                                
try:                                                                            
    st.merge()                                                                  
except Exception as e:                                                          
    print(str(e))                                                               
                                                                                
for tr in st:                                                                   
    dtype = tr.data.dtype.newbyteorder('native')                                
    tr.data = np.require(tr.data, dtype=dtype)                                  
                                                                                
st.merge()                                                                      
st.plot()                                                                       
```

### Proposed solution

Ideally we should probably simply convert byteorder to native on reading any data, but then we might also have to avoid keeping the original byteorder around in metadata (like we do when reading MiniSEED) or we might run into weird issues further down the road (e.g. when writing data again) and there is a mismatch between the data byteorder in the current object and the target byteorder when writing data out to file.

A more conservative approach might be to only convert byte order to native when actually performing operations that work on combining multiple traces, so probably only when merging (aka calling `Trace.__(i)add__`)

Anybody have any thoughts on this? Might be easy to overlook some implications when changing things in this regard..

megies:
### What does this PR do?

Show up some ways how to work around exceptions getting raised when trying to merge traces with different byteorder but otherwise being mergable.
Still not sure if we want to automatically adjust byteorder when needed as this might have side effects we don't think of right now maybe.. so maybe a more conservative option would be to show a meaningful error message if a merge is failing solely due to a byteorder mismatch that points at the option to manually force byteorder to a common one.

Kinda would like to hear thoughts.

fixes #3417 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,9 @@ Changes:
      (see #3261)
    * allow to specify Trace and Stream filter frequencies as arguments
      (see #3294)
+   * improve error messages when traces fail to get merged due to differing
+     data type, sampling rate etc (see #3418)
+   * add helper methods to change/unify byteorder on trace/stream (see #3418)
  - obspy.clients.filesystem:
    * update syntax for SQLAlchemy 2.0 compatibility (see #3269)
  - obspy.clients.fdsn
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3639,6 +3639,17 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
             self.traces += traces
         return self
 
+    def newbyteorder(self, byteorder='native'):
+        """
+        Change byteorder of the data
+
+        For details see
+        :meth:`Trace.newbyteorder <obspy.core.trace.Trace.newbyteorder>`.
+        """
+        for tr in self:
+            tr.newbyteorder(byteorder)
+        return self
+
 
 def _is_pickle(filename):  # @UnusedVariable
     """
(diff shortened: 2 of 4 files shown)


d-chambers (approved):
This seems like a straight-forward solution and a simple implementation. The only thing I can think of that might go wrong could be in writing the stream/traces to some binary format that assumes a certain byte order. However, since this behavior is something the user has to opt in to use I don't see how it could affect any existing codes.

megies:
I added some improved error messages on failed merges (e.g. due to dtype mismatch) and also added tests. Should be good to go now

trichter (approved):

PR #3442, opened 2024-05-13
try building docset in github actions

### LINKED ISSUE #3307: docs: re-enable building docsets (for dash/zeal)
Would be good to have docsets built again (for macOS "dash" / linux "zeal"), looks like that got scrapped during the transition to github actions.

https://docsets.obspy.org/

Alternatively could see if we can get obspy into the dash build chain, maybe.

megies:
### What does this PR do?

Tries to build docsets for dash/Zeal again, in github actions now

closes #3307 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/docs_conda_env.yml
+++ b/.github/docs_conda_env.yml
@@ -27,3 +27,4 @@ dependencies:
   - sphinx_rtd_theme
   # deps used in tutorial but not working anymore
   #- mlpy
+  - doc2dash  # for building docset for dash/zeal
--- a/.github/workflows/docs_building.yml
+++ b/.github/workflows/docs_building.yml
@@ -45,7 +45,9 @@ jobs:
         run: |
           cd misc/docs/build
           python -c 'import obspy; print(obspy.__version__)'
-          echo "DOCSETVERSION=$(shell head -3 html/objects.inv | grep ' ObsPy Documentation ' | sed 's#.*(\(.*\)).*#\1#')" >> "$GITHUB_ENV"
+          echo "DOCSETVERSION=$(head -3 html/objects.inv | grep ' ObsPy Documentation ' | sed 's#.*(\(.*\)).*#\1#')" >> "$GITHUB_ENV"
+          ls -l *
+          ls -l */*
           doc2dash --verbose --name "ObsPy ${DOCSETVERSION}" --icon html/_static/obspy-logo_32x32.png --online-redirect-url https://docs.obspy.org/ html/
           ls -l *
           sed 's#\(<string>[Oo]bs[Pp]y\) .*\(</string>\)#\1 ${DOCSETVERSION}\2#' --in-place "ObsPy $(DOCSETVERSION).docset/Contents/Info.plist"
--- a/.github/workflows/docs_building.yml
+++ b/.github/workflows/docs_building.yml
@@ -43,13 +43,8 @@ jobs:
       - name: build docset
         shell: bash -l {0}
         run: |
-          cd misc/docs/build
-          python -c 'import obspy; print(obspy.__version__)'
-          echo "DOCSETVERSION=$(head -3 html/objects.inv | grep ' ObsPy Documentation ' | sed 's#.*(\(.*\)).*#\1#')" >> "$GITHUB_ENV"
-          doc2dash --verbose --name "ObsPy ${DOCSETVERSION}" --icon html/_static/obspy_logo_no_text_32x32.png --online-redirect-url https://docs.obspy.org/ html/
-          sed 's#\(<string>[Oo]bs[Pp]y\) .*\(</string>\)#\1 ${DOCSETVERSION}\2#' --in-place "ObsPy $(DOCSETVERSION).docset/Contents/Info.plist"
-          cat ../docset_css_fixes.css >> "ObsPy ${DOCSETVERSION}.docset/Contents/Resources/Documents/_static/css/custom.css"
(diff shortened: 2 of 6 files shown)


megies:
Looks like docs building in github actions does not kick off if adding the "build_docs" label right when opening the PR. But that's a minor issue

megies:
This is ready pretty much, only thing missing is grabbing the artifacts on our server. I had a very quick look at it @trichter but couldn't immediately figure out how to add grabbing the docset artifacts as well

megies:
The deploy isn't working anymore because it only expects one artifact. Should be easy to fix, we could just filter by artifact name:

```python
r = requests.get(run['artifacts_url'], params={'name': 'obspydoc'}, **RKW)
```

Like in browser: https://api.github.com/repos/obspy/obspy/actions/runs/9064790999/artifacts?name=obspydoc

@trichter the deploybot code isn't under version control looks like? Mind if I make it a git repo out of it and push it to github? Looks like the only thing we want to avoid pushing is that file with the github token in it?

trichter:
> The deploy isn't working anymore because it only expects one artifact. Should be easy to fix, we could just [filter by artifact name](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28)

Perfect, you already solved it.

The deploy code is in folder misc/scripts/docs_deploybot

trichter (approved):
I finished with the deploy code on the server.

Doc for this pr is [here](https://docs.obspy.org/pr/docsets/).
Docset is [here](https://docs.obspy.org/pr/docsets/obspy-master.docset.tgz) and [here](https://docs.obspy.org/pr/docsets/Obspy.docset).

Let's merge this PR and see if it works with the push to master. If everything works I can add the modified server code in a separate PR.

megies:
> The deploy code is in folder misc/scripts/docs_deploybot

Ah, OK, I didn't see it there :+1: 

> Let's merge this PR and see if it works with the push to master. If everything works I can add the modified server code in a separate PR.

Sounds good :+1: We mostly want the `master` build docsets to keep updating, I guess, and the stable version builds to get stored, but I guess that's probably built in already similar to how it is done with the docs

trichter:
> Sounds good 👍 We mostly want the master build docsets to keep updating, I guess, and the stable version builds to get stored, but I guess that's probably built in already similar to how it is done with the docs

Yes, I changed this server-side.

@megies Can you please have a look at the new master docset: https://docsets.obspy.org/

The name of the master tgz file is a bit different than before. Does it matter? Do I have to adapt it? Also, there is a xml file. I guess this has a special purpose. Do wee need to update it automatically or can we delete it?

megies:
> @megies Can you please have a look at the new master docset: https://docsets.obspy.org/

docset looks ok in zeal, even though now it has the navigation column on every page which isn't ideal but I don't see an easy option to toggle that in `doc2dash` and I don't feel like looking into how to change that..



> The name of the master tgz file is a bit different than before. Does it matter? Do I have to adapt it?

I guess it can stay, or actually we can make the old name a symlink to the new one maybe. The xml file is basically what you add in Zeal as a "feed" and then it checks the version number in there regularly and gives an option to update the local docset copy.

Then we could have another xml file "stable" that links to the current stable docset tgz. We'd just need some routine to rewrite the xml files with new version number every time they get deployed. The version number we'd have to scrape from the `index.html` probably, I fear..

trichter:
OK, so you need the full version number for master:

```
<entry>
    <version>1.2.1.post0+1.g219b24d2ee.obspy.master</version>
    <url>http://docsets.obspy.org/obspy-master.tgz</url>
</entry>
```

It is not enough to do 


```
<entry>
    <version>dev</version>
    <url>http://docsets.obspy.org/obspy-master.tgz</url>
</entry>
```

megies:
I think so, yes. I believe zeal/dash is looking at the xml you provide as "feed URL" to decide if a local docset needs updating. Users can always manually go to the folder where those docsets are locally stored and delete and replace them manually, but otherwise (at least in the Linux "zeal" version) the re is no option to manually supply a docset, it's either through a feed URL or selecting from the official dash repository menu.

megies:
If it is too much hassle though, I think just having them docsets up there on our server is fine

megies:
I could also try to be better at injecting the version number into the "Info.plist" which seems the main file for stuff like this and could be parsed with an xml parser

trichter:
I think it is OK, I can write some ugly code to create these two XML files, obspy-master.xml and obspy-stable.xml

megies:
@trichter I think the docs deploy broke: https://docs.obspy.org/pr/fix_tsindex_feb2024/

Looks like only the docsets got uploaded (which we don't even really need for PRs)

trichter:
Ups, sorry, will fix it.

trichter:
Done. The docset for this PR is available at https://docs.obspy.org/pr/fix_tsindex_feb2024/docsets
We can later change the workflow to not upload the docset artifact for PRs.
Will still need this function to test the xml file.

trichter:
@megies, one more question. Do we need the uncompressed docset (`.docset` folders), or just the `.tgz`, or both?
I ask, because I saw that the xml refers to the `.tgz` archive.

megies:
I think just the tgz is fine. I don't know dash (the original Mac app) and how you can add docsets there (I think mainly through the official dash repository), but in zeal (the Linux clone) you can either enter a "feed URL" (link to that xml with the remote version and tgz location specified) or you can manually put the uncompressed docset in the location where the app locally stores its application data. So probably no use to have it uncompressed on the server, I believe. Not sure why we did it that way.

trichter:
I removed the uncompressed docsets.

PR #3461, opened 2024-05-24
numpy 2.0 compatibility

### LINKED ISSUE #3460: Numpy 2.0(.0rc2)
Numpy 2.0 will release with some changes in behavior in June. We need to ideally adjust our code to work with it like before or pin `numpy<2`.

https://numpy.org/devdocs/numpy_2_0_migration_guide.html

https://mail.python.org/archives/list/numpy-discussion@python.org/thread/AACXABTJ5T44T275ISQWKGL3IIOZAYTB/

https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-0-specific-advice

https://tests.obspy.org/140045/

Numpy 2.0.0rc2 can be installed with conda like this (or using the new env file from the related pull request)
```shell
(base)$ conda create -n np2 python numpy==2.0.0rc2 conda-forge/label/numpy_rc::_numpy_rc obspy pytest pytest-json-report
# optionally switch to dev obspy
(base)$ conda activate np2
(np2)$ conda remove --offline --force obspy
(np2)$ pip install -e /path/to/obspy/git
```

megies:
### What does this PR do?

Tries to get ourselves ready for upcoming numpy 2.0 in June 2024

fixes #3460 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,6 +122,16 @@ jobs:
           - os: ubuntu-latest
             python-version: '3.12'
             options: default warnings
+          # numpy 2 specific builds
+          - os: ubuntu-latest
+            python-version: '3.12'
+            options: default warnings np2
+          - os: macos-latest
+            python-version: '3.12'
+            options: default warnings np2
+          - os: windows-latest
+            python-version: '3.12'
+            options: default warnings np2
           - os: ubuntu-latest
             python-version: '3.12'
             options: ${{ (github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test_network')) && 'network warnings' || 'default' }}
@@ -136,7 +146,7 @@ jobs:
       # set individual cache key
       cache_key: ${{ matrix.os }}-py${{ matrix.python-version }}-${{ contains(matrix.options, 'mindep') }}-${{ contains(matrix.options, 'mindepversion') }}
       # set conda environment file with dependencies
-      env_file: .github/test${{ contains(matrix.options, 'mindepversion') && '_mindepversion' || contains(matrix.options, 'mindep') && '_mindep' || '' }}_conda_env.yml
+      env_file: .github/test${{ contains(matrix.options, 'np2') && '_np2' || contains(matrix.options, 'mindepversion') && '_mindepversion' || contains(matrix.options, 'mindep') && '_mindep' || '' }}_conda_env.yml
       # set additional runtest options (--keep-images, --network, -W error)
       runtest_options: ${{ github.event_name == 'pull_request' && '--keep-images' || '' }} ${{ contains(matrix.options, 'network') && '--network' || '' }} ${{ contains(matrix.options, 'warnings') && '-W error' || '' }}
     steps:
--- a/.github/test_mindep_conda_env.yml
+++ b/.github/test_mindep_conda_env.yml
@@ -7,7 +7,7 @@ dependencies:
   - decorator
   - lxml
   - matplotlib
-  - numpy
+  - numpy < 2
   - scipy
   - requests
   - setuptools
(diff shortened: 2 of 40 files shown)


trichter:
Looks good, maybe also add test cases without turning warnings into errors and network test cases?

megies:
> Looks good, maybe also add test cases without turning warnings into errors and network test cases?

Yea, maybe change it later.. only concern is how that would blow up the size of our test matrix. Working on the numpy stuff rn

megies:
It's looking quite good. Not a lot more to figure out, I think, based on local tests. Let's see CI. Need to fix some bitshift thing in mseed and some precision loss issue in realtime that needs some upcasting in places probably.

Most dangerous stuff I've seen is when things like `np.uint8` or `np.float32` go into calculations without upcasting which runs into precision loss sometimes. And some weird stuff going on with bitshifts that used to upcast in weird fashion, feels like that was abusing bitshifts, not sure though, no bitshift magician, really.

megies:
The [test fail in trace preview creation](https://tests.obspy.org/140223/#1) seems to come from a numpy bug (or at least changed behavior, kind of) in the `ptp()` method-now-function, see numpy/numpy#26530

megies:
see numpy/numpy#13718 for a deprecation warning shown in `io.nordic`

.github/workflows/tests.yml
```diff
@@ -136,7 +146,7 @@ jobs:
       # set individual cache key
       cache_key: ${{ matrix.os }}-py${{ matrix.python-version }}-${{ contains(matrix.options, 'mindep') }}-${{ contains(matrix.options, 'mindepversion') }}
       # set conda environment file with dependencies
-      env_file: .github/test${{ contains(matrix.options, 'mindepversion') && '_mindepversion' || contains(matrix.options, 'mindep') && '_mindep' || '' }}_conda_env.yml
+      env_file: .github/test${{ contains(matrix.options, 'np2') && '_np2' || contains(matrix.options, 'mindepversion') && '_mindepversion' || contains(matrix.options, 'mindep') && '_mindep' || '' }}_conda_env.yml
```
ThomasLecocq, line 149: will this remain when the merge is done ? or should the mindep still be tested in the future?

ThomasLecocq (approved):
looks all good to me!

megies:
I think this is ready. Most changes for numpy v2.0 are quite trivial, mostly replacing some removed function with its drop in replacement.

The most concern is numpy from v2 not automatically upcasting to higher datatypes wherever needed, which can lead to e.g. `np.int8` and similar overflowing if not manually upcast or precision loss in single precision floats that used to automatically get upcast to double precision whenever needed.
I'm fairly sure, that our tests might not catch all cases where this potentially happens, but there isn't really much we can do about it now. I added manual upcasts wherever our tests started failing (usually just at the x-th digit behind the comma..).

This also happens in some bitwise operations. I'm not really used to using bitwise operations a lot, so if somebody could double check these, would be great. In any case tests still pass..
 - https://github.com/obspy/obspy/pull/3461/commits/ffb326b63e083983e5150265dbf1402f6c681f18
 - https://github.com/obspy/obspy/pull/3461/commits/46231a42fe0f1e473cd2b128673e9ca39ed2e433

Network test fails are unrelated, something broke with EarthScope FDSN WS

That one failing test on windows is unrelated too, looks like an issue with removing read permission from a test file (so we can check it is skipped by our Scanner utility)

.github/workflows/tests.yml
megies: > mindep still be tested in the future?

mindep is still being tested, matrix was just made bigger with additional runners for np2. this line only sets up the filename for the conda env file used in the respective runner, depending on the "flags"

megies: After numpy releases v2.0.0 we can probably kick out these additional runs again, since our "normal" current dependency version runners will eventually pick up numpy v2. It's just for the time while there is only the v2.0.0rc2 prerelease to test with we need some extra specs in the env file so that the pre release gets picked up

jkmacc-LANL (approved):

PR #3488, opened 2024-07-04
fdsn client : raise a FDSNNoServiceException rather than a ValueError…

### LINKED ISSUE #3483: Raise a FDSNException rather than a ValueError when a service is not present in Client.services
### Avoid duplicates

### Bug Summary

This morning I stumble on a ValueError raised by obspy.clients.fdsn.Client.get_waveforms
It appears that during client initialization, the request for services "application.wadl" returned a (transient) 50X error.
Later, when executing get_waveforms method, "dataselect" was not present in Client.services and a ValueError was raised.

Should obspy raise a FDSNException (or more specific like FDSNNoServiceException) rather than this ValueError ?
It would be easier later on to handle this kind of exception in our programs.

```python
try:
    client.get_waveforms(....)
except FDSNNoDataException:
    pass
except FDSNException:
    handle(...)
```
Would be cleaner than :
```python
try:
    client.get_waveforms(....)
except FDSNNoDataException:
    pass
except FDSNException, ValueError:
    handle(...)
```

### Code to Reproduce

_No response_

### Error Traceback

```Python traceback
ValueError: The current client does not have a dataselect service.
  ...
  File "obspy/clients/fdsn/client.py", line 858, in get_waveforms
    raise ValueError(msg)
```

### ObsPy Version?

1.4.1

### Operating System?

Ubuntu

### Python Version?

3.11.9

### Installation Method?

pip

simonpanay:
… if service is not present in Client.services

### What does this PR do?

This PR replace ValueError raised with FDSNNoServiceException when dataselect|station|event is not present in Client.services and get_waveforms|stations|events[_bulk] is called.

fixes #3483 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -27,6 +27,8 @@ Changes:
      current msindex (see #3403)
  - obspy.clients.fdsn
    * Natural Resources Canada (NRCAN) added to list of known clients
+   * A FDSNNoServiceException is now raised instead of a ValueError when
+     querying a webservice not present in Client.services  (see #3483)
  - obspy.clients.seedlink:
    * fix a bug in basic client `get_info()` which leaded to exceptions when
      querying with `level="channel"` in presence of stations with no current
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -86,6 +86,7 @@ Muñoz Mesa, Jesús A.
 Murray-Bergquist, Louisa
 Nesbitt, Ian M.
 Nof, Ran Novitsky
+Panay, Simon
 Panning, Mark P.
 Parker, Tom
 Pestourie, Romain
--- a/obspy/clients/fdsn/client.py
+++ b/obspy/clients/fdsn/client.py
@@ -546,7 +546,7 @@ class Client(object):
         """
         if "event" not in self.services:
             msg = "The current client does not have an event service."
-            raise ValueError(msg)
+            raise FDSNNoServiceException(msg)
 
         locs = locals()
         setup_query_dict('event', locs, kwargs)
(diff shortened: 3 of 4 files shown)


simonpanay:
@megies 
I wrote a first test to assert that the right exception is raised when no service is present in Client instance.
Should I repeat this test in all other methods or try refactor this behaviour in client.fdsn.client ?

megies:
No, that one test looks perfect and is all we need. :+1:

megies:
A couple of small things but this is almost good to go already
 - one flake8 fail for a too long file [you can see in the diff](https://github.com/obspy/obspy/pull/3488/files#diff-8847d239f7b06bfe16caf1535be21cc2bf21e24bec709390cbe82b06fef0e91dR565)
 - you could actually improve the test, make a for loop to check all three services for the proper exception getting raised.
 - add a short one-liner in changelog
 - move the test in that `test_client.py` file into the new separate class for non-networked tests (I recently made that change in a PR and you merged it in), since it will run without internet connection

Then one last thing, we want to rebase on recent master to get rid of that "merge master" commit. I can do it at the end if needed, just let me know.

simonpanay:
All done
And I rebased my branch on master (I hope I did it correctly)

megies:
That rebase went bad, I fixed it and made some last adjustments to the test. Actually, loop wasn't the best idea actually, because I think it might be better to test one by one. Also, the client you used there is only setup once on the test class, so removing those services would have affected other test routines.

megies (approved):
Should be good to merge once CI is green

megies:
Thanks for the PR @simonpanay :rocket:

simonpanay:
Great ! Thanks @megies

PR #3515, opened 2024-11-03
Read the origin-specific data from the Phase Block in the ISF Bulletin

zacherle:
### What does this PR do?

This PR parses phase blocks from the ISC/IMS1.0:SHORT bulletin into the Catalog, including origin-specific data.

In `obspy/io/iaspei/core.py`:
1) The `specify_preferred_origin` subroutine searches for an `origin` marked explicitly as `#PRIME` or one that can implicitly be considered preferred.
2) The `_read_phases` method, for each block of seismic phases, checks for the presence of the `#OrigID` tag, which specifies the `origin` associated with the seismic phases.
3) The `_parse_phase` method parses origin-specific data (e.g., epicentral distance and time residual) in the phase block into the Arrival object.
4) Completes the missing `Amplitude.to_pick` link.
5) Stores the list of Arrival items in the specified `origin` object.

Minor changes: 
- Adjusted parsing to be less case-sensitive for certain bulletin keywords (e.g., `Event`, `EVENT`). 
- Added tolerance for extra lines before the `DATA_TYPE` bulletin header (e.g., AutoDRM message envelope: `BEGIN IMS1.0 ...`). 
- Fixed units for hypocenter depth and uncertainties to meters, consistent with QuakeML.
- Previously, if the `_parse_phase` subroutine could not reference an `origin`, the origin-specific data was stored as a comment. I suggest removing this feature; it is now commented out.

The current processing approach for the `ISF/IMS1.0:SHORT` bulletin phase block, where origin-specific data is stored as a textual comment, appears to be a workaround.

This PR aims to align with the ISC/IMS1.0 format specification: http://isc-mirror.iris.washington.edu/standards/isf/
The documentation specifies that a preferred-origin (`#PRIME` origin) or an `#OrigID` tag should be specified for each phase block, enabling the processing of origin-specific data (Arrival object).

Additionally, there is currently no `event.magnitude.Amplitude.to_pick` link, making it impossible to associate amplitude data with the correct stations and phases.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/iaspei/core.py
+++ b/obspy/io/iaspei/core.py
@@ -16,7 +16,8 @@ from obspy import UTCDateTime
 from obspy.core.event import (
     Catalog, Event, Origin, Comment, EventDescription, OriginUncertainty,
     QuantityError, OriginQuality, CreationInfo, Magnitude, ResourceIdentifier,
-    Pick, StationMagnitude, WaveformStreamID, Amplitude)
+    Pick, StationMagnitude, WaveformStreamID, Amplitude, Arrival)
+from obspy.core.event.header import AmplitudeUnit
 from obspy.core.util.obspy_types import ObsPyReadingError
 from .util import (
     float_or_none, int_or_none, fixed_flag, evaluation_mode_and_status,
@@ -77,8 +78,16 @@ class ISFReader(object):
     def deserialize(self):
         if not self.lines:
             raise ObsPyReadingError()
-        line = self._get_next_line()
-        if not line.startswith('DATA_TYPE BULLETIN IMS1.0:short'):
+        header_flag = False
+        while not self._next_line_type():
+            # Skip the lines preceding the header
+            # (if the bulletin is enveloped as an AutoDRM response message)
+            # (BEGIN IMS1.0, MSG_TYPE DATA, ...)
+            line = self._get_next_line()
+            if line.upper().startswith('DATA_TYPE BULLETIN IMS1.0:SHORT'):
+                header_flag = True
+                break
+        if not header_flag:
             raise ObsPyReadingError()
         try:
             self._deserialize()
@@ -90,7 +99,7 @@ class ISFReader(object):
         line = self._get_next_line()
         catalog_description = line.strip()
         self.cat.description = catalog_description
-        if not self.lines[0].startswith('Event'):
+        if not self.lines[0].lower().startswith('event'):
             raise ObsPyReadingError()
         # get next line stops the loop eventually, raising a controlled
         # exception
@@ -152,6 +161,7 @@ class ISFReader(object):
         # read origins block
         if block_type == 'origins':
             self._read_origins()
+            self._specify_preferred_origin()
         # read publications block
         elif block_type == 'bibliography':
             self._read_bibliography()
@@ -169,13 +179,27 @@ class ISFReader(object):
 
     def _read_phases(self):
         event = self.cat[-1]
+        origin = event.preferred_origin()
+        line = self.lines[0].strip()
+        # OrigID Phase and Phase Information Comment
+        if line.startswith('(#OrigID'):
+            origin_id = ResourceIdentifier(
+                self._construct_id(['origin', line[8:].rstrip(") ").strip()]))
+            origin = origin_id.get_referred_object()
+        if origin:
+            origin_id = origin.resource_id
+        else:
+            origin_id = None
+
         while not self._next_line_type():
             line = self._get_next_line()
             if line.strip().startswith('('):
                 comment = self._parse_generic_comment(line)
-                event.picks[-1].comments.append(comment)
+                if len(event.picks):
+                    event.picks[-1].comments.append(comment)
                 continue
-            pick, amplitude, station_magnitude = self._parse_phase(line)
+            pick, amplitude, station_magnitude, arrival = self._parse_phase(
+                                                             line, origin_id)
             if (pick, amplitude, station_magnitude) == (None, None, None):
                 continue
             event.picks.append(pick)
@@ -183,6 +207,10 @@ class ISFReader(object):
                 event.amplitudes.append(amplitude)
             if station_magnitude:
                 event.station_magnitudes.append(station_magnitude)
+            if origin:
+                # The list of Arrival items is stored in the specified origin
+                if arrival:
+                    origin.arrivals.append(arrival)
             continue
 
     def _read_origins(self):
@@ -225,6 +253,23 @@ class ISFReader(object):
             event.origins.extend(origins)
             event.event_type = event_type
             event.event_type_certainty = event_type_certainty
+
+    def _specify_preferred_origin(self):
+        '''
+        Find the preferred origin in two simple cases:
+          - only one origin for an event
+          - tagged #PRIME
+        Affects: event.preferred_origin_id
+        '''
+        event = self.cat[-1]
+        event.preferred_origin_id = None
+        if len(event.origins) == 1:
+            event.preferred_origin_id = event.origins[0].resource_id.id
+            return
+        for origin in event.origins:
+            for comment in origin.comments:
+                if "#PRIME" in comment.text.upper():
+                    event.preferred_origin_id = origin.resource_id.id
 
     def _read_magnitudes(self):
         event = self.cat[-1]
@@ -276,20 +321,20 @@ class ISFReader(object):
         #                           a fixed epicenter solution)
         epicenter_fixed = fixed_flag(line[54])
         # 56-60   f5.1  semi-major axis of 90% ellipse or its estimate
-        #               (km, blank if fixed epicenter)
+        #               ([m], blank if fixed epicenter)
         _uncertainty_major_m = float_or_none(line[55:60], multiplier=1e3)
         # 62-66   f5.1  semi-minor axis of 90% ellipse or its estimate
-        #               (km, blank if fixed epicenter)
+        #               ([m], blank if fixed epicenter)
         _uncertainty_minor_m = float_or_none(line[61:66], multiplier=1e3)
         # 68-70   i3    strike (0 <= x <= 360) of error ellipse clock-wise from
         #                       North (degrees)
         _uncertainty_major_azimuth = float_or_none(line[67:70])
-        # 72-76   f5.1  depth (km)
+        # 72-76   f5.1  depth [m]
         depth = float_or_none(line[71:76], multiplier=1e3)
         # 77      a1    fixed flag (f = fixed depth station, d = depth phases,
         #                           blank if not a fixed depth)
-        epicenter_fixed = fixed_flag(line[76])
-        # 79-82   f4.1  depth error 90% (km; blank if fixed depth)
+        depth_fixed = fixed_flag(line[76])
+        # 79-82   f4.1  depth error 90% ([m]; blank if fixed depth)
         depth_error = float_or_none(line[78:82], multiplier=1e3)
         # 84-87   i4    number of defining phases
         used_phase_count = int_or_none(line[83:87])
@@ -356,7 +401,8 @@ class ISFReader(object):
             time=time, resource_id=origin_id, longitude=longitude,
             latitude=latitude, depth=depth, depth_errors=depth_error,
             origin_uncertainty=origin_uncertainty, time_fixed=time_fixed,
-            epicenter_fixed=epicenter_fixed, origin_quality=origin_quality,
+            epicenter_fixed=epicenter_fixed or depth_fixed,
+            origin_quality=origin_quality,
             comments=comments, creation_info=creation_info)
         # event init always sets an empty QuantityError, even when specifying
         # None, which is strange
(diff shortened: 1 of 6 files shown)


zacherle:
> One comment now already:
> 
> > Additionally, there is currently no `event.magnitude.Amplitude.to_pick` link, making it impossible to associate amplitude data with the correct stations and phases.
> 
> Should be possible, we follow QuakeML schema to link via resource IDs:
> 
> ```python
> from obspy.core.event import Amplitude, Pick
> p = Pick()
> a = Amplitude(pick_id=p.resource_id)
> a.pick_id.get_referred_object() is p  # True
> ```
My patch is exactly what you state. I just simply added the missing argument `pick_id=pick.resource_id` when calling the Amplitude constructor.

ThomasLecocq:
Hi @zacherle ; I'm going through the idle PRs one by one.

I did a first pass on this one, and as first observation, I'm wondering if it'd make sense to _keep_ the `comments` populated in case current users would use them, but think of maybe adding a warning that the `.comment()` method for getting the specific information should be replaced by a direct access to <the object> using <the attribute>... ?

To move forward, could you:

- [ ] rebase on current master
- [ ] check the code style (lint - flake8)
- [ ] add a test file (or fake an existing one) to include values for all fields (SNR, etc) & test them in test_core.py
- [ ] edit CHANGELOG.txt
- [ ] make sure the documentation is complete
  - [ ] check docstrings/doctests to public function
  - [ ] add an example usage for the main core documentation?
- [ ] check the CONTRIBUTORS lists for your name

zacherle:
@ThomasLecocq  Thank you for looking at it.

On the question of whether or not to keep comments for their possible current users:

As @megies commented in his first commit, using Pick.comments is temporary:

```python
def _parse_phase(self, line):
    # since we can not identify which origin a phase line corresponds to,
    # we can not use any of the included information that would go in the
    # Arrival object, as that would have to be attached to the appropriate
    # origin..
    # for now, just append all of these items as comments to the pick
```

It was probably only intended for debugging purposes.
I believe that users understood it the same way and do not use this part.
Then let's not be afraid to delete these Pick.comments.

In the last PR commit, processing of the "orphan" phases is controlled by two
flags: skip_orphan and origin_specific_to_comments.
Default is to skip the orphan phases (skip_orphan=True).
But user can set to process the orphan phases and store the origin-specific
data in the Pick.comments (origin_specific_to_comments=True) almost the same as before.

We have another change in this PR that may appear to be a backward compatibility break and may be noticed by more users.
But this is not a new feature, more of a bug fix.

Earthquake depth and earthquake location uncertainties are now in meters, not \[km\] as before.
This is in line with the QakeML specification and the obspy.core.event.Event class.
In the original code there is a `multiplier=1e3` to convert depth from \[km\] to \[m\], but it was not working due to the missing `**kwargs` argument when calling the `type_or_none` function in io.iaspei.util.py:

```python
 def float_or_none(my_string, **kwargs):
-    return type_or_none(my_string, float)
+    return type_or_none(my_string, float, **kwargs)
```

The change of depth units is so noticeable that hopefully there will be no mistakes.

- [ ] rebase on current master
- [x] check the code style (lint - flake8)
- [x] add a test file (or fake an existing one) to include values for all fields (SNR, etc) & test them in test_core.py
- [x] edit CHANGELOG.txt
- [ ] make sure the documentation is complete
  - [x] check docstrings/doctests to public function
  - [ ] add an example usage for the main core documentation?
- [x] check the CONTRIBUTORS lists for your name

Please set the `build_docs` label to test the _read_ims10_bulletin method docstring.

ThomasLecocq:
>Please set the build_docs label to test the _read_ims10_bulletin method docstring.

Done, sorry, didn't see the message

zacherle:
Thanks, modified docstring is here
https://docs.obspy.org/pr/feat_io_iaspei_phase_block/packages/autogen/obspy.io.iaspei.core._read_ims10_bulletin.html#obspy.io.iaspei.core._read_ims10_bulletin

ThomasLecocq:
doc looks OK: maybe a bit dense & I see a few typos/missing words "(fce args)" etc, could you check those?

zacherle:
Are we not spending too much effort on handling cases where an orphaned phase block occurs in a bulletin? (No origin can be assigned to an orphaned phase block.)
Such cases can be easily and correctly handled by skipping this phase block (and issuing a warning).
Such a case can occur:
1) Intentionally
The ISF format allows multiple phase blocks for one event. If any of them is not tagged with the #OrigID tag, it is excluded from further processing.
2) This is a bug
In this case, a fix must be made in the ISF bulletin. Fixing data in legacy formats, is that a job for obspy?

The `skip_orphane` and `origin_specific_to_comments` options should be removed, including the related code. Should I do this in the next commit, or will someone else do it in the next ObsPy version?
I believe that backward compatibility will not be an issue, since the feature of writing values to Pick.comments has not been documented and is not used anywhere else in ObsPy. 

Alternatively, as a simple compromise solution, the skip_orphan and origin_specific_to_comments options could be left as hidden, undocumented options, as is currently the case with the _no_uuid_hashes option. The code would be the same, but users would not be encouraged to use these non-standard solutions.

ThomasLecocq:
>Alternatively, as a simple compromise solution, the skip_orphan and origin_specific_to_comments options could be left as hidden, undocumented options, as is currently the case with the _no_uuid_hashes option. The code would be the same, but users would not be encouraged to use these non-standard solutions.

kind of agree with this. But maybe your comments/thinking to the code base of those methods so the logic/rationale is preserved?

zacherle:
I have retained the current implementation, which includes the two optional keyword
arguments 'skip_orphan' and 'origin_specific_to_comments'.
However, I have relocated the documentation for these arguments to a less prominent
location. Instead of placing it in the interface function _read_isf10_bulletin,
I have moved it to the internal method ISFReader.__init__, where these arguments
are actually utilized.
This approach allows us to keep the options available for now,
while discouraging new users from employing them.

The code should be further refined, but this can be addressed in future versions
of ObsPy by those who have a broader understanding of the various needs of ISF
format users.
For now, it meets our requirements and serves its purpose:
1. It reads bulletins containing phase data, which we provide to ISC and share with others.
2. The results of reading GSE bulletins using _read_gse2 and ISF bulletins using _read_isf10_bulletin are comparable.
3. The original capability to read summary ISF bulletins that do not contain detailed phase data is preserved.

obspy/io/iaspei/core.py
```diff
@@ -226,6 +294,25 @@ def _read_origins(self):
...
+    def _specify_preferred_origin(self):
+        """
+        Find the preferred origin
+
+        in two simple cases:
+          - only one origin for an event
+          - tagged #PRIME in the origin comments
+        Affects: event.preferred_origin_id
+        """
+        event = self.cat[-1]
```
megies, line 312: Why last event in Catalog? Or is there ever only one Event?

obspy/io/iaspei/core.py
```diff
@@ -226,6 +294,25 @@ def _read_origins(self):
...
+        event.preferred_origin_id = None
+        if len(event.origins) == 1:
+            event.preferred_origin_id = event.origins[0].resource_id.id
+            return
+        for origin in event.origins:
+            for comment in origin.comments:
+                if "#PRIME" in comment.text.upper():
+                    event.preferred_origin_id = origin.resource_id.id
```
megies, line 323: I feel like there could be either a "return" after hitting the condition or a check if multiple origins fit it and issue a warning about the selection being ambiguous

obspy/io/iaspei/core.py
```diff
@@ -356,7 +443,8 @@ def _parse_origin(self, line):
             time=time, resource_id=origin_id, longitude=longitude,
             latitude=latitude, depth=depth, depth_errors=depth_error,
             origin_uncertainty=origin_uncertainty, time_fixed=time_fixed,
-            epicenter_fixed=epicenter_fixed, origin_quality=origin_quality,
+            epicenter_fixed=epicenter_fixed or depth_fixed,
```
megies: so if I understand correctly, the original naming "epicenter_fixed" was bad and it really should be "epicenter_or_depth_fixed=epicenter_fixed or depth_fixed"?

since it's a private routine, we could fix the naming if we wanted to and just rename it. But up to you.

megies: Ah nevermind, that's the Origin constructor and apparently QuakeML doesn't have a flag for depth fixed which is weird..

obspy/io/iaspei/core.py
```diff
@@ -627,8 +865,56 @@ def _read_ims10_bulletin(filename_or_buf, **kwargs):
...
+        multiple origins are given and the #PRIME or #OrigID tag is not present
+        ), a warning is issued and the parsing of this phase block is skipped.
+
+    :type file: str or :class:`~pathlib.Path` or
+        file-like object( e.g. `ByesIO`)
+    :param filename_or_buf: bulletin file to read
+    :raises: :class:`~obspy.core.ObsPyReadingError` on problems to read file
+    :rtype: :class:`~obspy.core.event.Catalog`
+    :return: Catalog object
+    :meta public:
```
megies: Please try to avoid adding automatic edits from custom IDEs.

obspy/io/iaspei/core.py
```diff
@@ -627,8 +865,56 @@ def _read_ims10_bulletin(filename_or_buf, **kwargs):
...
+    #     If False, orphan phase blocks will be skipped
+    #     (i.e. if no origin could be assigned to the phase block).
+    # :param origin_specific_to_comments: Default False.
+    #     If True, origin-specific data in orphan phase blocks
+    #     will be stored as text in the Pick.comments
+    # Use of these two options is not recommended,
+    # as the resulting Catalog object will not be fully
+    # in accordance with the QuakeML standard.
+    # They are only provided for backward compatibility.
+    # See the ISFReader.__init__ method docstring for details.
```
megies: If this is valuable information for users / specific use cases, which it seems it is.. this should not be a comment that nobody will ever see but should be in the docstring in the function so that people can see it if needing to look at this low level function for details

obspy/io/iaspei/core.py
megies: I removed those.

megies:
A couple of minor questions.

Otherwise good to go, I think, if CI is OK with it.. hard to say with such big edits, but it seems old tests are mostly unchanged and there's tests for new functionality, so there's that. :+1:

obspy/io/iaspei/core.py
zacherle: self.cat[-1] is the last event loaded.
It is one-pass sequential processing of the bulletin, so I don't know why the Event objects are stored in a list.
I kept the original programming style and do not consider myself competent for a more extensive refactoring.

obspy/io/iaspei/core.py
zacherle: Only one origin should be tagged as #PRIME.
In case of multiple origins and multiple phase blocks, the #OriginID tag should be used to specify the origin for each phase block.
However, if multiple origins are marked as #PRIME, the last one will be selected. Added counter and warning message.

obspy/io/iaspei/core.py
zacherle: I understand.
I tried to distinguish private and public for Sphinx.

obspy/io/iaspei/core.py
zacherle: These comments are repeated in the `ISFReader.__init__` docstring.
Redundant comments deleted.
I don't want to emphasize this option.
We're just keeping this option as a precaution until we're sure we can delete it.

obspy/io/iaspei/core.py
megies: Which might be a good thing, but better keep it consistent since we don't have that anywhere else. And if anybody feels that's an important thing we should do it across the whole code base.. but sphinx groups by public/private anyway (based on `_` prefix). Or am I missing something?

obspy/io/iaspei/core.py
megies: > We're just keeping this option as a precaution until we're sure we can delete it.

Ok that's fine. Just.. if its a public function/method and we release it with some args/kwargs, we won't really delete them later. That's kind of our policy to not ever be breaking codes with new releases

zacherle: Then it would be best to remove it right now.
(skip_orphan=False and origin_specific_to_comments=True options and all subsequent code)
I don't think anyone will miss it.
But I don't know all the users, so I'll go with your opinion.

obspy/io/iaspei/core.py
zacherle: I just learned that the Origin class has a `depth_type` argument that can indicate whether the depth is fixed or not.
And that the constructor call does not match the current implementation of the Origin class in obspy (not existing  `origin_quality` argument instead of `quality` argument, nowhere to set depth_type argument).
I will try to fix it.

zacherle: This was fixed in the latest commit aa85cb3
The issue with the fixed depth flag is now resolved by the depth_type argument.
The solution was taken from the io/gse2/bulletin.py.
Furthermore, the origin.time_errors, origin.depth_errors and origin.quality data were added.
Now the obspy catalog obtained by reading the GSE2 format and
the catalog obtained by reading the IMS format are better comparable.

obspy/io/iaspei/tests/test_core.py
```diff
@@ -69,24 +43,75 @@ def _assert_catalog(self, got):
                     origin.origin_uncertainty.confidence_ellipsoid = None
                 # QuakeML reader seems to add empty QuantityError for
                 # pick.horizontal_slowness_errors
-                for key in ['time_errors', 'longitude_errors',
-                            'latitude_errors', 'depth_errors']:
+                for key in ['longitude_errors', 'latitude_errors',
+                            # 'depth_errors', 'time_errors',
```
ThomasLecocq, line 47: Why no longer testing for depth_errors and time_errors ?

zacherle: Originally, these attributes were not subject to testing and were just set to None here.
By commenting this out, I am actually including these attributes in the testing.

megies:
only rebased and force-pushed

PR #3550, opened 2025-03-11
Stats: subclasses for NSLC and Station Identifier

megies:
### What does this PR do?

Try out a way to handle the two ways of identifying stations/channels. The old way being a combination of network, station, location and channel code, often following the definitions of the SEED 2.4 standard. And the new way being the definition of FDSN Source Identifier, used in MiniSEED v3, comprised of namespace, network, station, location, band, source and subsource channels.

One potential way to try and do this cleanly is with having two subclasses to `Stats`. The aim being to try and make it almost behave like before so to make it "easy" for most normal experience users, but silently under the hood being very clear of the two different systems, with clear exceptions being raised in scenarios with IDs that can not be mapped without ambiguity (e.g. a NSLC-type channel code with less than 3 characters) to be flexible and well defined for more experienced "power" users.

This PR is not fully done yet and still has some bugs and still missing the logic for mapping back and forth with exceptions defined for ambiguous mapping actions, But I'd like to hear some feedback, since I feel like we kind of need some change/addition to `Stats` before we can release a new major version with MiniSEED v3 support (from simplemseed or mseedlib or both).

```python
>>> from obspy.core import Stats
>>> header_nslc = {'network': 'BW', 'station': 'MANZ', 'location': '', 'channel': 'BHZ'}
>>> stats = Stats(header=header_nslc)
>>> print(stats)
         network: BW
         station: MANZ
        location: 
         channel: BHZ
       starttime: 1970-01-01T00:00:00.000000Z
         endtime: 1970-01-01T00:00:00.000000Z
   sampling_rate: 1.0
           delta: 1.0
            npts: 0
           calib: 1.0
>>> print(isinstance(stats, Stats))
True
>>> print(type(stats))
<class 'obspy.core.trace.NSLCStats'>
```

```python
>>> from obspy.core import Stats
>>> header_sid = {'namespace': 'FDSN', 'network': 'BW', 'station': 'MANZ', 'location': '',
...               'band': 'B', 'source': 'H', 'subsource': 'Z'}
>>> stats = Stats(header=header_sid)
>>> print(stats)
       namespace: FDSN
         network: BW
         station: MANZ
        location: 
            band: B
          source: H
       subsource: Z
       starttime: 1970-01-01T00:00:00.000000Z
         endtime: 1970-01-01T00:00:00.000000Z
   sampling_rate: 1.0
           delta: 1.0
            npts: 0
           calib: 1.0
>>> print(isinstance(stats, Stats))
True
>>> print(type(stats))
<class 'obspy.core.trace.SourceIdentifierStats'>
```

The two systems (NSLC vs. Source Identifier) are not fully compatible and in both directions there can be [cases where a mapping can not be made](https://docs.fdsn.org/projects/source-identifiers/en/v1.0/definition.html#mapping-of-seed-2-4-codes). This means that there is a need for some kind of logic to handle data coming in from NSLC based data formats as well as Source Identifier based data formats (MiniSEED v3), because in some cases reading one format and writing to another will not be possible with a fully automatic mapping, so we kind of want well defined exceptions being raised etc, so the users can work around issues in well defined ways. 

This is just one possible way to do it, and it would be good to discuss other options.

The main consideration is, that currently our Stats object is based on NSLC with one field for "channel". Therefore if we leave it like that while already integrating MiniSEED v3 support, we would have to make a mapping from Source Identifier to NSLC right when reading data, which can be problematic and ambiguous and make additional problems in reverse mapping back when writing the same data in MiniSEED v3 again even. So it feels the better way is to somehow make Stats Source-Identifier-ready right away and only do mappings back or forth on the fly whenever needed and not earlier, so that problems in mapping only pop up when there really is a need to map.

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -11,6 +11,7 @@ Module for handling ObsPy :class:`~obspy.core.trace.Trace` and
 """
 import inspect
 import math
+import re
 import warnings
 from copy import copy, deepcopy
 
@@ -24,6 +25,23 @@ from obspy.core.util.base import _get_function_from_entry_point
 from obspy.core.util.decorator import raise_if_masked, skip_if_no_data
 from obspy.core.util.misc import (flat_not_masked_contiguous, get_window_times,
                                   limit_numpy_fft_cache)
+
+
+# list of keys we handle that are common to NSLC and FDSN Source Identifier
+BASE_STATS_KEYS = (
+    'starttime', 'endtime', 'sampling_rate', 'delta', 'npts', 'calib')
+# complete list of all parts of a NSLC
+NSLC_KEYS = ('network', 'station', 'location', 'channel')
+# name for the FDSN Source Identifier composed of all of its parts
+SOURCE_ID_KEY = 'source_identifier'
+# complete list of all parts of a FDSN Source Identifier
+SOURCE_ID_KEYS = ('namespace', 'network', 'station', 'location', 'band',
+                  'source', 'subsource')
+# these are specific to FDSN Source Identifier, so if these appear it can not
+# be a plain NSLC description
+SOURCE_ID_SPECIFIC_KEYS = ('namespace', 'band', 'source', 'subsource')
+SOURCE_ID_REGEX = re.compile(
+    r'(.*):([^_]*)_([^_]*)_([^_]*)_([^_]*)_([^_]*)_(.*)')
 
 
 class Stats(AttribDict):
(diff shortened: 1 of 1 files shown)


megies:
Actually, maybe it's a better and more lightweight / less changes approach to rather create a `ChannelCode` object and use that in existing `Stats` instead of the approach of subclassing `Stats`..

megies:
As this PR becomes more of a favorite for me, I've added more details and improved the original description for this PR.

PR #3562, opened 2025-04-10
inventory.py: "where" indices helper

filefolder:
UPDATE FEB 28: 

2nd review, fixed some logic. `select` does in fact return a shallow copy. 

UPDATE:: I've done an overhaul to the previous version of this PR which  added a much simpler "get_indices" helper.

Changes:

1) Minor edit of the "shallow copy" warning in `select` since it doesn't seem to be returning a shallow copy (**this is not so)
2) New `where` function in inventory.py that returns a tuple or list of tuples of array indices similar to numpy. Search can be seed_id or standard `select` kwargs or a combination of both
3) _getitem_ modified to accept normal [1][2][3] indexing OR tuples (e.g. what `where` returns). A channel-level 3-tuple will return a list of channels which are **actual** shallow copies of the original inventory that can be used to update the original inventory easily.

I tried to avoid changing `select`, rather just add a new function that anyone not involved in inventory editing won't ever need. A lot of this could of course be done in net/sta/cha do loops but it does come in handy when trying to surgically (or quickly!) edit or insert certain things. 

Will test for the next few weeks...

examples:

test inventory
```
inv = obspy.read_inventory()
```

find indicies in various ways
```
inv.where(channel='BH?')
>> [(0, 0, 3), (0, 0, 4), (0, 0, 5), (0, 1, 3), (0, 1, 4), (0, 1, 5)]

inv.where('*.*.*.BH?') #or use seed_id
>> [(0, 0, 3), (0, 0, 4), (0, 0, 5), (0, 1, 3), (0, 1, 4), (0, 1, 5)]

inv.where('*.*.*.BH?',station='XYZ') #can do both, but seed_id takes precedence (throws warning)
>> UserWarning: station filter params will be superseded by seed_id filter
>> [(0, 0, 3), (0, 0, 4), (0, 0, 5), (0, 1, 3), (0, 1, 4), (0, 1, 5)]

inv.where('BW') #just network level
>> [(1,)]

inv.where('GR',station='WET') #or both, no warning now because no station in seed_id
>> [(0,1)]

inv.where('GR',channel='*Z') 
>> [(0, 0, 0), (0, 0, 3), (0, 0, 6), (0, 0, 9), (0, 1, 0), (0, 1, 3), (0, 1, 6)]

inv.where('*.FUR..BH?')
>> [(0, 0, 3), (0, 0, 4), (0, 0, 5)]
```

rename all BH? to XX
```
for i,idx in enumerate(inv.where(channel='BH?')):
    inv[idx].code = 'XX'+str(i)

# print(inv)
```

confirm changes edited original inventory and now show up when searching XX?
```
inv.where(channel='BH?')
>> []

inv.where(channel='XX?')

>> [(0, 0, 3), (0, 0, 4), (0, 0, 5), (0, 1, 3), (0, 1, 4), (0, 1, 5)]


inv[inv.where(channel='XX?')][0:2]

>> [Channel 'XX0', Location '' 
        Time range: 2006-12-16T00:00:00.000000Z - --
        Latitude: 48.1629, Longitude: 11.2752, Elevation: 565.0 m, Local Depth: 0.0 m
        Azimuth: 0.00 degrees from north, clockwise
        Dip: -90.00 degrees down from horizontal
        Channel types: TRIGGERED, GEOPHYSICAL
        Sampling Rate: 20.00 Hz
        Sensor (Description): Streckeisen STS-2/N seismometer (None)
        Response information available,
 Channel 'XX1', Location '' 
        Time range: 2006-12-16T00:00:00.000000Z - --
        Latitude: 48.1629, Longitude: 11.2752, Elevation: 565.0 m, Local Depth: 0.0 m
        Azimuth: 0.00 degrees from north, clockwise
        Dip: 0.00 degrees down from horizontal
        Channel types: TRIGGERED, GEOPHYSICAL
        Sampling Rate: 20.00 Hz
        Sensor (Description): Streckeisen STS-2/N seismometer (None)
        Response information available]

```

pull a channel from 2020 FUR and insert it RJOB around 2006 (a normal thing people do) 
```
 # a 3 component channel-level index
A = inv.where('GR.FUR..XX1',time=obspy.UTCDateTime(2020,1,1))

# a 2 component (station level) index
B = inv.where('*.RJOB',time=obspy.UTCDateTime(2006,1,1))

# adding a single channel at index A (GR.FUR..XX1) into the channels of BW.RJOB
inv[B][0].channels += inv[A] 
```
copy the response from GR.WET..LHZ and use it to replace all the responses in the HH? channels

```
r = inv[inv.where('GR.WET..LHZ')][0].response
for c in inv[inv.where('GR.WET..HH?')]:
   c.response = r
```


ThomasLecocq:
hey @filefolder sorry I didn't check this one earlier. Seems a very good idea & implementation. Would it make sense to generalize it to add the "3 parts" (NET.STA.LOC) version and list the channels ? Some users use a single NET.STA acquisition with LOC being remote sensors, I bet that'd help. Also check the https://github.com/obspy/obspy/pull/3562/files "Files Changed" tab, it lists lines not covered by tests (seems) ?

ThomasLecocq:
Hi @filefolder , going back to this.

1. I see the potential usage of it, but for the regular user, it's not really clear from the doctrings (what do you do with the indices)
2. this new method should be covered by tests (at least doctests (so no +SKIP)

related to 1.: I suppose the idea is to do `inv[i][j][k]` to access a specific channel ? Why not make a general `get_object` (or other name) that would return the desired object (Network, Station, Channel) ?

filefolder:
Appreciate your attention to this! Sorry I phased out for a bit there. 

The problem I was trying to address is that, as far as I know, a user can't "work with" an object without knowing its indices, I suppose mostly this sort of searching would be via `select`

For example, 

```
In [27]: inv.select(station='SWNNC')[0][0].channels
Out[27]: 
[Channel 'CHZ', Location '' 
	Time range: 2022-02-01T00:00:00.000000Z - 2022-02-21T11:26:22.000000Z
	Latitude: -33.1939, Longitude: 117.2890, Elevation: 250.0 m, Local Depth: 0.0 m
	Azimuth: 0.00 degrees from north, clockwise
	Dip: -90.00 degrees down from horizontal
	Sampling Rate: 250.00 Hz
	Sensor (Description): BB (DT-SOLO BD3C-5 0.2 Hz node (BB))
	Response information available]

In [28]: inv.select(station='SWNNC')[0][0].channels = []

In [29]: inv.select(station='SWNNC')[0][0].channels
Out[29]: 
[Channel 'CHZ', Location '' 
	Time range: 2022-02-01T00:00:00.000000Z - 2022-02-21T11:26:22.000000Z
	Latitude: -33.1939, Longitude: 117.2890, Elevation: 250.0 m, Local Depth: 0.0 m
	Azimuth: 0.00 degrees from north, clockwise
	Dip: -90.00 degrees down from horizontal
	Sampling Rate: 250.00 Hz
	Sensor (Description): BB (DT-SOLO BD3C-5 0.2 Hz node (BB))
	Response information available]
```

The channels are still there, so one can't make any easy search modifications in this way (again as far as I know..)

but if you know the indices it works as expected.

```
In [30]: inv[0][2].channels = []

In [31]: inv[0][2].channels
Out[31]: []
```

It sounds like a type of object retriever that follows through with actually inserting said indices is probably a bit more useful but would have to consider that a bit more fully. Since I assume this is sort of a esoteric thing no one will notice I have been a bit lazy.

ThomasLecocq:
would a simple solution be to add a `return_copy=True` by default to the `select()` method, and you could override it setting it to `False` and you'll get a "view" on the original objects ?

ThomasLecocq:
edit, hummmm, select should be a simple shallow copy : https://docs.obspy.org/packages/autogen/obspy.core.inventory.inventory.Inventory.select.html

edit: YES, but of the objects themselves, not the containers. What happens when you :

```
for chan in inv.select(station='SWNNC')[0][0].channels):
   del chan
   
```

ThomasLecocq:
That looks fun !

megies:
Please don't be mad at us, but together with @ThomasLecocq we thought to bump this one, hopin to finish a RC today still :/

PR #3570, opened 2025-05-06
Decide on a fixed scheme for dependency versions for the future like SPEC0, bump dependency versions for next major release, update CI and docs building accordingly

megies:
### What does this PR do?

**1.** Adjusts minimum dependency versions for next major release 1.5.0 and also adjusts CI build matrix accordingly. For python+numpy we decided to follow [NEP029](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) which leaves scipy and matplotlib minimum versions open for discussion, but probably could just go by release date of scipy+matplotlib and allow maybe 2-3 years backwards (like [numpy recommends](https://numpy.org/neps/nep-0029-deprecation_policy.html#implementation) in terms of clarity of wording and policy)

from NEP029:
```
On Apr 05, 2024 drop support for Python 3.9 (initially released on Oct 05, 2020)
On Jun 22, 2024 drop support for NumPy 1.23 (initially released on Jun 22, 2022)
On Dec 18, 2024 drop support for NumPy 1.24 (initially released on Dec 18, 2022)
On Apr 04, 2025 drop support for Python 3.10 (initially released on Oct 04, 2021)
<--  May 2025  -->
On Jun 17, 2025 drop support for NumPy 1.25 (initially released on Jun 17, 2023)
On Sep 16, 2025 drop support for NumPy 1.26 (initially released on Sep 16, 2023)
<--  Mar 2026  -->
On Apr 24, 2026 drop support for Python 3.11 (initially released on Oct 24, 2022)
On Jun 16, 2026 drop support for NumPy 2.0 (initially released on Jun 15, 2024)
```

scipy release dates:
```
1.9  -- Jul 29, 2022
1.10 -- Jan 03, 2023    <--  May 2025
1.11 -- Jun 25, 2023
1.12 -- Jan 20, 2024
1.13 -- Apr 02, 2024
1.14 -- Jun 24, 2024
1.15 -- Jan 02, 2025
1.16 -- Jun 22, 2025
1.17 -- Jan 10, 2026    <-- latest
```

matplotlib release dates:
```
3.6  -- Sep 16, 2022
3.7  -- Feb 13, 2023     <--  May 2025
3.8  -- Sep 15, 2023
3.9  -- May 16, 2024
3.10 -- Dec 14, 2024     <-- latest
```

--

**2.** Adjust some CI envs, build matrix and workflows accordingly

**3.** Update outdated `plot_directive` sphinx extension

see #3228 

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/.github/docs_conda_env.yml
+++ b/.github/docs_conda_env.yml
@@ -21,9 +21,8 @@ dependencies:
   # docs
   - cartopy>=0.20
   - pip
-  - pip:
-    - pybtex
-    - m2r2 >= 0.3.2  # latest version which is not yet included in conda-forge
+  - pybtex
+  - m2r2 >= 0.3.2
   - sphinx>=4.2.0
   - sphinx_rtd_theme
   # deps used in tutorial but not working anymore
--- a/.github/test_mindep_conda_env.yml
+++ b/.github/test_mindep_conda_env.yml
@@ -13,7 +13,7 @@ dependencies:
   # see pypa/setuptools#4480
   - setuptools !=71.0.1
   # see #3258
-  - sqlalchemy < 2.0
+  - sqlalchemy
   # tests
   - pytest
   - pytest-cov
--- a/.github/test_mindepversion_conda_env.yml
+++ b/.github/test_mindepversion_conda_env.yml
@@ -3,16 +3,16 @@ channels:
   - conda-forge
 dependencies:
   # obspy
-  - numpy=1.25
-  - scipy=1.11
-  - matplotlib=3.8
+  - numpy=2.2
+  - scipy=1.15
+  - matplotlib=3.10
   - decorator
   - lxml
   - requests
   # see pypa/setuptools#4480
   - setuptools !=71.0.1
   # see #3258
-  - sqlalchemy < 2.0
+  - sqlalchemy
   # soft dependencies
   - cartopy
   - geographiclib
@@ -20,7 +20,7 @@ dependencies:
   - pyshp != 3.0.0
   # tests
   - packaging
-  - pyproj < 3.5
+  - pyproj
   - pytest
   - pytest-cov
   - pytest-json-report >= 1.4
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -30,18 +30,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        build: [cp38, cp39, cp310, cp311, cp312, cp313, cp314]
+        build: [cp312, cp313, cp314]
         exclude-flag:
           - ${{ github.ref_name == 'master' && github.event_name == 'push' }}
         exclude:
-          - build: cp38
-            exclude-flag: true
-          - build: cp39
-            exclude-flag: true
-          - build: cp310
-            exclude-flag: true
-          - build: cp311
-            exclude-flag: true
           - build: cp312
             exclude-flag: true
           - build: cp313
--- a/.github/workflows/docs_building.yml
+++ b/.github/workflows/docs_building.yml
@@ -32,7 +32,11 @@ jobs:
           python-version: '3.12'
           activate-environment: obspydoc
           environment-file: .github/docs_conda_env.yml
-
+      - name: print package info
+        shell: bash -l {0}
+        run: |
+          conda info -a
+          conda list
       - name: install obspy
         shell: bash -l {0}
         run: |
--- a/.github/workflows/docs_building.yml
+++ b/.github/workflows/docs_building.yml
@@ -29,7 +29,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v4
         with:
           miniconda-version: 'latest'
-          python-version: '3.11'
+          python-version: '3.12'
           activate-environment: obspydoc
           environment-file: .github/docs_conda_env.yml
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,27 +88,27 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.12', '3.13', '3.14']
         options: [default]
         include:
           # The mindep option indicates optional dependencies should not be installed.
           - os: ubuntu-latest
-            python-version: '3.10'
+            python-version: '3.12'
             options: default mindep
           # The mindepversion option indicates dependencies should be installed with minimal supported version.
           - os: ubuntu-latest
-            python-version: '3.10'
+            python-version: '3.12'
             options: default mindepversion
           - os: macos-latest
-            python-version: '3.10'
+            python-version: '3.12'
             options: default mindepversion
           - os: windows-latest
-            python-version: '3.10'
+            python-version: '3.12'
             options: default mindepversion
           # For push events and for test_network labels, options is set to 'network' and runs for network tests are added.
           # Otherwise these include statements do nothing, because options='default' is already present in matrix.
           - os: ubuntu-latest
-            python-version: '3.10'
+            python-version: '3.12'
             options: ${{ (github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test_network')) && 'network mindepversion' || 'default' }}
           - os: ubuntu-latest
             python-version: '3.14'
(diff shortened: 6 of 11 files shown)


megies:
Looks good to me, also the update of the docs building env and plot directive sphinx extension. I'll leave it open for discussion until we really close in on the release though, mostly for the minimum version numbers we want for scipy and matplotlib.

megies:
Probably will be superseded by #3662

megies:
see https://github.com/obspy/obspy/issues/3228#issuecomment-4010370695

ThomasLecocq:
also see https://scientific-python.org/specs/spec-0000/ ?

ThomasLecocq:
ClaudeAI + SPEC0000 suggestion, if we plan e.g. a release once a year Q1:


ObsPy | Release | EOL | Min Python | Min NumPy | Min SciPy | Min Matplotlib
-- | -- | -- | -- | -- | -- | --
1.5.0 | 2026 Q1 | 2027 Q4 | 3.8 (we know why) | 1.21 | 1.9 | 3.7
1.6.0 | 2027 Q1 | 2028 Q4 | 3.11 | 2.0 | 1.13 | 3.9
1.7.0 | 2028 Q1 | 2029 Q4 | 3.12 | 2.2 | 1.15 | 3.10
1.8.0 | 2029 Q1 | 2030 Q4 | 3.13 | 2.4 | 1.17 | 3.11*

in our case 1.6 could come much faster, we'll have to adapt then:

ObsPy | Release | EOL | Min Python | Min NumPy | Min SciPy | Min Matplotlib
-- | -- | -- | -- | -- | -- | --
1.5.0 | 2026 Q1 | 2027 Q4 | 3.8 | 1.21 | 1.9 | 3.7
1.6.0 | 2026 Q3 | 2028 Q2 | 3.11 | 2.0 | 1.13 | 3.9
1.7.0 | 2027 Q1 | 2028 Q4 | 3.12 | 2.2 | 1.15 | 3.10
1.8.0 | 2028 Q1 | 2029 Q4 | 3.13 | 2.4 | 1.17 | 3.11*
1.9.0 | 2029 Q1 | 2030 Q4 | 3.14 | 2.6* | 1.19* | 3.13*

filefolder:
Not necessarily advocating, but a more forgiving scheme could be something following the 5-year Ubuntu LTS schedule which probably tracks a lot closer with the average seismologist's computer than a yearly version bump.

Version 22.04 ships with 3.10 and is supported until April 2027. 3.8 support just ended last year. I think these are essentially the limits on what's reasonable. 

Or just make it up as we go...

megies:
@ThomasLecocq and I will work on obspy this week, so this one is one of the first things to nail down. We talked about it a lot today, and in the end it kind of boils down to

 - we need a *clear fixed rule* for what we do now and in the future about the version bumps, mainly to avoid wasting more time on discussions around this topic in the future
 - all the *big Python packages unanimously agree* (i.e. all our main dependencies) to adhere to [`SPEC 0`](https://scientific-python.org/specs/spec-0000/) and actively endorse it (numpy, scipy, matplotlib, ipython, ..)
 - if we do a custom rule (follow *nix distro XYZ LTS), then the next person might show up and argue, "but I'm using distro ABC LTS and that's even more important here, so we should support scipy version X.Y.Z" etc.

So @ThomasLecocq and I are leaning towards just applying `SPEC 0`.

Regarding the argument of "but average seismologist might use Python X (5 years old)", we would argue, but then why not use the stable obspy from 5 years ago too, without the new features? It just seems unrealistic and unfair, to expect a small package like obspy (with very few core people behind it) can support a very long backlog of workarounds for old dependencies, when the big packages with many more people working on them refuse to do that exact same thing [for good reasons](https://scientific-python.org/specs/spec-0000/#motivation).

Its the same thing with LTS distros like e.g. Debian. Yes, you get a very stable system, but at the same time, you accept not getting any new features in the lifespan of that system. And if you want more current software, you have to do it in your user space, separately from the system packages.

The main counter argument I see is..
 - numpy etc need to be so strict, since they are so far down at the basis of the whole ecosystem, so if they are too loose about the supported version numbers, they blow up the whole "what needs to be supported" for all packages higher up the food chain. And obspy is not in that exact same position and could be more loose..?

And then, following this argument, I could see a point that we could instead base our decision on what we *could* still support given the main build systems we use, which is..
 - github CI for pypi wheels
 - conda-forge for conda packages

But then situations could arise that obspy version 1.x.0 supports certain Python versions but when we release 1.x.1, one of your build systems might have dropped a Python version in between and we can not build and offer packages for the same set of minimum dependencies that the initial release was supporting.

But this leads back to the original problem. Do we want to spend a lot of time on custom decisions around this or just pick the gold standard and not spend more time on it any longer.

The ["background" of `SPEC 0`](https://scientific-python.org/specs/spec-0000/#background) kind of sums it up perfectly, really:

"In the past, longer support cycles were common. There were several reasons for this, including the Python 2 / 3 transition, difficulties installing packages, and users needing to use old, operating-system provided versions of Python. The situation has since improved due to improved installations via binary wheels, virtual environments becoming commonplace, and support for Python 2 being dropped."

Other thoughts?

ThomasLecocq:
Applying SPEC0 , and with an example of 2 coming releases (1.6.0 now and 1.7.0 in March), that would look like this:
image

On this graph, we see the enormous version support we did for 1.5.0 :-) and we can already see where 1.6.0 and 1.7.0 would land. For fast-release packages, it would mean we support 4 versions (numpy, scipy) - for slower (matplotlib) the risk is to only have 1 version: e.g. in March 2027, if 3.12 is not out, only 3.11 would be supported. So we should have a rule to have "2 years, minimum 2 versions" as a safety net.

megies:
Just rebased for now to get it able to merge, version numbers still to be decided together with the overall new strategy

megies:
One thing @ThomasLecocq and I discussed as well: Once again, we want to try to keep up a more regular major release schedule (aiming for at least once per year in March) and with the major release, fix the next version's dependency versions based on the planned release date and adhere to it even if the release gets delayed (by lets half a year or up to a year)

ThomasLecocq:
@filefolder @claudiodsf @d-chambers @QuLogic @chad-earthscope @raciner @jschaeff @jkmacc-LANL @trichter 
pinging you all to grab your attention on this matter 😇🔥

jschaeff:
Hello,

Nice point to define a policy for dropping support of old dependencies. 

In general, users should try to stay up to date in their obspy versions and python. Nowadays, relying on the Linux distrib's python is definitely NOT the recommended way to work. 

I'm wondering how UV could help out  to make sure each release pins the right dep version for each obspy version (including suported python versions of course) ? 
UV has made those dependencies nightmare way more manageable and more importantly, way more reproducible.
Adding installation instructions with UV would be nice.

Did you discuss this option ?

raciner:
> @filefolder @claudiodsf @d-chambers @QuLogic @chad-earthscope @raciner @jschaeff @jkmacc-LANL @trichter pinging you all to grab your attention on this matter 😇🔥

I think we shouldn't make our own life difficult by supporting many combinations of libraries, as we obviously don't have a big team keeping track of this and I guess most of us would prefer to contribute new stuff over running after legacy compatibility issues to make people happy who don't want to upgrade their default installation from 2022. Also, most of our users probably are more knowledgeable than a standard Ubuntu user. Most people probably use some of the tools around to create their own environment and to fulfill library dependencies instead of going with 5 year old system defaults.

filefolder:
IIRC the whole issue for the last release was that we had already maintained support for 3.8, so why not just keep it in? I think it's great we accomplished that, sort of an ObsPy LTS.

For future releases I'm happy to adhere to whatever schedule so long as we plan it out ahead of time.

nota bene I checked today's unique IPs & user agents on our server

Python 3.7 - 7.1% (respect!)
Python 3.8 - 4.7%
Python 3.9 - 9.5%
Python 3.10 - 9.5%
Python 3.11 - 40.4%
Python 3.12 - 14.4% 
Python 3.13 - 11.9%

Might be cute to track this over time but it's worth keeping an open mind that a good proportion of people downloading waveform data may not enjoy using the computer

chad-earthscope:
my 2 cents: in general software is moving to faster cycles (beyond seismology and python), faster change is inevitable.  Users do not like the disruption from changes (understandably).  The incompatibility of these two realities is mostly felt in projects like this that directly serve non-technologists, but we cannot afford to keep users protected from change.  A reasonable approach is a) retaining old versions for old environments (done) and b) do not create new software for old environments.  Said another way, new software releases are for the future with no dependencies that are EOL within 6 months, ideally 1 year.  Maintenance and back-porting are separate work and can better target highest-impact issues.

ThomasLecocq:
syngine tests are linked to #3790

ThomasLecocq:
SPEC0 script added to the misc/scripts . When run, it generates the figure above, and reports a table (* are projected releases based on the dependencies' regular rythm)

| ObsPy | date | python | numpy | scipy | matplotlib |
|---|---|---|---|---|---|
| 1.5.0 | 2026-03-01 | 3.12 – 3.14 | 2.0 – 2.4 | 1.13 – 1.17 | 3.9 – 3.10 |
| 1.6.0 | 2026-09-01 | 3.12 – 3.14 | 2.2 – 2.5 | 1.15 – 1.18 | 3.10 – 3.11 |
| 1.7.0 | 2027-03-01 | 3.13 – 3.15 | 2.3 – 2.6* | 1.16 – 1.19* | 3.11 – 3.12* |

megies:
The new fixed strategy is mentioned in the wiki in the dev corner and under install instructions
https://github.com/obspy/obspy/wiki/Minimum-Supported-Dependencies-%28SPEC0%29

PR #3580, opened 2025-06-24
Add normalization factor calculation method to `PolesZerosResponseStage`

### LINKED ISSUE #2574: Add calc_normalization_factor to PolesZeros

liamtoney:
### What does this PR do?

Adds a method `calculate_normalization_factor()` to the `PolesZerosResponseStage` class. Adds option to calculate the normalization factor from the poles and zeros when a `PolesZerosResponseStage` object is created via `normalization_factor=None`.

The content of this PR is included in the much more involved PR #2592 — the purpose of this PR is to make a bite-size added feature easily merge-able. Closes #2574.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -203,8 +203,9 @@ class PolesZerosResponseStage(ResponseStage):
     :param zeros: All zeros of the stage.
     :type poles: list[complex]
     :param poles: All poles of the stage.
-    :type normalization_factor: float, optional
-    :param normalization_factor:
+    :type normalization_factor: float or None, optional
+    :param normalization_factor: Normalization factor. If None, it will be calculated
+        from the poles and zeros.
     """
     def __init__(self, stage_sequence_number, stage_gain,
                  stage_gain_frequency, input_units, output_units,
@@ -220,9 +221,12 @@ class PolesZerosResponseStage(ResponseStage):
         # handled by properties.
         self.pz_transfer_function_type = pz_transfer_function_type
         self.normalization_frequency = normalization_frequency
-        self.normalization_factor = float(normalization_factor)
         self.zeros = zeros
         self.poles = poles
+        # If the user sets normalization_factor to None, we calculate it
+        if normalization_factor is None:
+            normalization_factor = self.calculate_normalization_factor()
+        self.normalization_factor = float(normalization_factor)
         super(PolesZerosResponseStage, self).__init__(
             stage_sequence_number=stage_sequence_number,
             input_units=input_units,
@@ -356,6 +360,8 @@ class PolesZerosResponseStage(ResponseStage):
         More reading here:
         https://docs.fdsn.org/projects/stationxml/en/latest/response.html#poles-and-zeros
         """
+        if not self.normalization_frequency:
+            return None
         if self.pz_transfer_function_type == 'LAPLACE (RADIANS/SECOND)':
             s = 2j * np.pi * self.normalization_frequency
         elif self.pz_transfer_function_type == 'LAPLACE (HERTZ)':
--- a/obspy/core/inventory/response.py
+++ b/obspy/core/inventory/response.py
@@ -348,6 +348,26 @@ class PolesZerosResponseStage(ResponseStage):
                    f"'LAPLACE (RADIANS/SECOND)'")
             raise ValueError(msg)
 
+    def calculate_normalization_factor(self):
+        """
+        Calculate the normalization factor, which normalizes the pole–zero expansion to
+        unity at the normalization frequency, from the response stage poles and zeros.
+
+        More reading here:
+        https://docs.fdsn.org/projects/stationxml/en/latest/response.html#poles-and-zeros
+        """
+        if self.pz_transfer_function_type == 'LAPLACE (RADIANS/SECOND)':
+            s = 2j * np.pi * self.normalization_frequency
+        elif self.pz_transfer_function_type == 'LAPLACE (HERTZ)':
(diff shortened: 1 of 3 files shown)


liamtoney:
Note a difference from how the initialization of the `PolesZerosResponseStage` is done in this PR versus in PR #2592. In that PR, the default value for `normalization_factor` was set to `None`, so that **by default** the normalization factor was automatically calculated from the poles and zeros. This PR favors maintaining the existing default value of `1.0` to ensure complete backwards compatibility; users can elect to specify `normalization_factor=None` if they do want the normalization factor calculated at initialization.

liamtoney:
> Just add the `build_docs` tag to this PR.
> Docs will be served at [docs.obspy.org/pr/{branch_name}](https://docs.obspy.org/pr/) (do not use master branch).
> Please post a link to the relevant piece of documentation.

Please add the "build_docs" label to this PR since the new method has a docstring. Looks like I cannot do it myself.

liamtoney:
Thanks @ThomasLecocq — documentation for the new method is viewable [here](https://docs.obspy.org/pr/calculate-a0/packages/autogen/obspy.core.inventory.response.PolesZerosResponseStage.calculate_normalization_factor.html).

liamtoney:
This PR is now ready for review.

ThomasLecocq:
Looks good to me code & doc & test-wise , I'm not an expert in PAZ :-) Would it make sense to add an example usage in the doc ?

liamtoney:
> Would it make sense to add an example usage in the doc ?

I added a simple example to the method docstring:

Screenshot 2025-07-29 at 12 26 03 AM

obspy/core/inventory/response.py
```diff
@@ -220,9 +221,12 @@ def __init__(self, stage_sequence_number, stage_gain,
         # handled by properties.
         self.pz_transfer_function_type = pz_transfer_function_type
         self.normalization_frequency = normalization_frequency
-        self.normalization_factor = float(normalization_factor)
         self.zeros = zeros
         self.poles = poles
+        # If the user sets normalization_factor to None, we calculate it
+        if normalization_factor is None:
+            normalization_factor = self.calculate_normalization_factor()
```
ThomasLecocq, line 228: as in your test case, if the information provided is PAZ without normalization_factor, does this line get called? If yes, it'll raise, correct?

liamtoney: The `__init__()` method default is `normalization_factor=1.0`, so in that case the normalization factor is not calculated here. It's just left at 1. If the user specifies `normalization_factor=None`, then this line gets called. The other option for this PR would be to set the default to `None` in the method definition (`normalization_factor=None`), but that has the potential to be a breaking change if folks have code assuming an initialized PAZ stage will have normalization factor of 1. Hence the choice here.

ThomasLecocq:
merged master into the PR to trigger the tests & see if we get a 💚 light

liamtoney:
> merged master into the PR to trigger the tests & see if we get a 💚 light

Looks like just codecov project failed?

ThomasLecocq (approved):

liamtoney:
Hi there, just following up to see if this can be merged now!

obspy/core/inventory/response.py
```diff
@@ -348,6 +352,42 @@ def to_radians_per_second(self):
...
+
+        .. rubric:: Example
+
+        >>> from obspy import read
+        >>> paz_stage = read()[0].stats.response.response_stages[0]
+        >>> paz_stage.calculate_normalization_factor()  # doctest: +ELLIPSIS
+        59206129.7610747...
+        """
+        if not self.normalization_frequency:
+            return None
```
megies, line 375: Might make sense maybe to add `calculate_normalization_factor(..., normalization_frequency=None)` to the function call, so that users could use a value of choice, but not important right now and could be added later if needed

megies (approved):
Looks OK 👍  I rebased on current master to get rid of merge commits and then good to go on green CI I think

megies:
force pushed and resolved CHANGELOG merge conflict but CI was already green so merging :rocket:

PR #3589, opened 2025-07-24
PPSD - Preloading evalresp responses from metadata

ThomasLecocq:
### What does this PR do?

This PR adds a new private method ``_preload_responses`` that is called at the end of the ``__init__`` of the PPSD object. This will parse _whatever_ ``metadata`` and return a list of responses matching the ``PPSD.id`` (seed id), with start and end dates, and the evalresp response already calculated.

When computing lots of PPSD for lots of windows, lots of days, e.g. for a single channel that has NEVER changed responses throughout the years, it is definitely useless to compute the evalresp for EACH time window slice of the added Stream/Traces. This implementation only computes it once.

Still to check: although all tests pass, I don't really know what happens when the trace.stats.sampling_rate is not integer, e.g. 100.1 Hz - there is a test for that, and it passes, but does evalresp produce a different response shape for a 100.0 Hz or 100.1 Hz sampling? 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -165,7 +165,6 @@ Changes:
  - obspy.signal.spectral_estimation.PPSD:
    * performance improvement in __init__: faster frequency array calculation
      (see #3644)
- - obspy.signal.spectral_estimation.PPSD:
    * Changed the moment instrument responses are calculated by evalresp: they
      are now preloaded at PPSD instantiation time, and no longer calculated
      for each time window! (see #3589)


ThomasLecocq:
@FMassin you're probably a good candidate to test this PR :-)

obspy/signal/spectral_estimation.py
```diff
@@ -630,6 +638,78 @@ def current_times_used(self):
...
+        if isinstance(self.metadata, (Inventory, str, Parser)):
+            if isinstance(self.metadata, str):
+                inv = read_inventory(self.metadata)
+            elif isinstance(self.metadata, Parser):
+                inv = read_inventory(BytesIO(self.metadata.get_seed()))
+            else:
+                inv = self.metadata
+            for network in inv.networks:
+                for station in network.stations:
+                    for channel in station.channels:
```
ThomasLecocq, line 652: should probably replace the nested loop by 

for seed_id in inv.get_contents()['channels']:
 ....

filefolder:
this one looks super useful, will try to run some tests in the next week

filefolder:
Tested this on a 16 station multi-year network (7200 sec windows), worked great and it nearly halved the processing time. Nice!

megies:
Absolutely makes sense to avoid repeat calls to evalresp to get the exact same if it yields a significant speedup. :+1: 

Looking at the PR it introduces quite significant changes, I feel like it would be possible to keep the footprint of the changes smaller and I feel like the PR is introducing a massive if/elif/else clause replacing more fine grained subroutine structure.

Also, this PR made me think, instead of doing this here for PPSD exclusively, we could maybe keep this response cache bookkeeping e.g. at the `Response` objects, so that it immediately would speed up all scenarios with repeated response requests also from repeat calls to `remove_response()`. One issue I can see is with Response objects getting changed manually which would need to invalidate the cache somehow.

ThomasLecocq:
@raciner you might want to have a look/check at this one maybe ?

raciner:
> @raciner you might want to have a look/check at this one maybe ?

I will have a look as soon as I have time to do so.

raciner:
Hi Thomas, I would suggest to add a caching logic to _get_response and change it to something like (pseudocode)
```
def _get_response(self,tr):
    if response_already_computed:
        return response
    elif isinstance(...)
        response = compute_response()
        cache(response)
        return response
    elif ...
```
Doing this on the level of a PPSD object would avoid issues we would face with caching on the Response level with users changing the response while the original version is still in the cache.

raciner:
After giving it some more thought, I would maybe add a cache self._response_cache = dict() in __init__. After this, I would use it in the varius _get_response_from_xxx methods. Example:

```
    def _get_response_from_inventory(self, tr):
        inventory = self.metadata
        response = inventory.get_response(self.id, tr.stats.starttime)
        
        # inventory always is the same and is immutable over the life time of PPSD()
        # using the id() as a hashing key is valid
        
        if id(response) in self._response_cache:
            return self._response_cache[id(response)]
        resp, _ = response.get_evalresp_response(
            t_samp=self.delta, nfft=self.nfft, output="VEL")
        self._response_cache[id(response)] = resp
        return resp

```
One could probably come up with similar caching strategies for the other functions. By the way, the speedup is impressive. Processing two days of 200sps data now takes half of the time (not taking into account loading the data and metadata). Considering that this function will be used heavily in several projects, thinking about what to do with remove_response at other places seems to be a good idea.

raciner:
I see that this is somehow stuck. This speedup would be a huge benefit for my group as we get our responses exclusively from inventory. This probably is the case for many others who use obspy together with an fdsn web service or a plain inventory file. I don't know whether there is a general consensus here, but if not, maybe one could consider at least to go forward with the optimisation I suggested?

filefolder:
it would indeed be great to get this in master so people can use and test it thoroughly. it appears to work fine on my end.

PR #3548, opened 2025-03-05
[WIP] Replace io.mseed (and our vendored libmseed) by mseedlib

ThomasLecocq:
### What does this PR do?

First attempt to get rid of the vendored libmseed and custom codes and make use of @chad-earthscope 's [mseedlib](https://github.com/EarthScope/mseedlib), as part of the efforts to get MSEED3 read/write support for obspy (see also #3411)

First sets of main tests pass, some issues left with gaps, endianess, etc...

WIP :-)

### Why was it initiated?  Any relevant Issues?

*Please fill in (link relevant issues with "see #123456", mark issues that get resolved with "fixes #12345")*

### PR Checklist
- [x] Correct base branch selected? `master` for new features, `maintenance_...` for bug fixes
- [ ] While the PR is still work-in-progress, the `no_ci` label can be added to skip CI builds
- [ ] If the PR is making changes to documentation, docs pages can be built automatically.
      Just add the `build_docs` tag to this PR.
      Docs will be served at [docs.obspy.org/pr/{branch_name}](https://docs.obspy.org/pr/) (do not use master branch).
      Please post a link to the relevant piece of documentation.
- [ ] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
      just add the `test_network` tag to this PR.
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are covered via new tests.
- [ ] Any new or changed features are fully documented.
- [ ] Significant changes have been added to `CHANGELOG.txt` .
- [ ] First time contributors have added your name to `CONTRIBUTORS.txt` .
- [ ] If the changes affect any plotting functions you have checked that the plots
      from all the CI builds look correct. Add the "upload_plots" tag so that plotting 
      outputs are attached as artifacts. 
- [ ] New modules, add the module to `CODEOWNERS` with your github handle
- [ ] Add the yellow `ready for review` label when you are ready for the PR to be reviewed.

Issue #3559, opened 2025-03-26
Stats: How to facilitate FDSN Source Identifier?


megies:
Alongside MiniSEED version 3, FDSN has introduced a new standard for handling how measurement setups are uniquely identified, called [FDSN Source Identifier](https://docs.fdsn.org/projects/source-identifiers/en/v1.0/#). It is likely that this new system will become more and more prevalent over time.
In principle the new system seems mostly like a clarification / clearer separation of multiple pieces of information combined in the channel code, but still there are implications for practical use and conversions between the different systems.

Currently, obspy kind of sits somewhere in between, with using the generally prevalent `NSLC` system of having separate network, station, location and channel codes, also used by the dominant `StationXML` standard (which does not restrict length of either code fields).
In contrast to SEED 2.4, obspy in principle does not impose restrictions on the length of these fields (same as StationXML) and leaves it mostly up to the user and the file formats interacted with in read/write operations to make it make sense (e.g. MiniSEED v2 can not fit network code with more than two characters).

Trying to push support for MiniSEED v3 finally over the doorstep into obspy (with [simplemseed integration](#3411) pretty much ready and up for grabs) leaves us with a decision to make how to go forward with our `Stats` object.

As always, one of the main aspects to consider is how to make changes in such a way to minimize immediate impact on users' existing code bases, especially considering how this is at the very core of things and variable access like `trace.stats.channel` (or even assignments like `trace.stats.channel = ...`) is likely to be in a vast majority of codes. So if a change happens that would mean interacting with `trace.stats` in such ways would change in a backwards incompatible fashion, this would mean a major disruption to users.

The following is mostly a brainstorming of some possible approaches, followed by an overview of mapping issues between SEED24/SID and lax NSLC/SID.

## Strategies for Stats Object

Here's a list of options I can think of to go about this, for sure there are other ways and some that I didn't think of that might be better..

### 1. Leave Stats as is for now, keep NSLC system, integrate MiniSEEDv3 and release a new major version

Positives:
- no changes in obspy core
- no changes for users right now
- get to new obspy release faster?

Negatives:
- forced to map SID to NSLC for every read/write interaction with MiniSEED3, which potentially is not round trip safe or even plain not possible in the first place
- users (of MiniSEED v3) potentially have to adjust twice if we change `Stats` later

[...]

PR #3604, opened 2025-08-20
Add io.mseed3 using pymseed (R/W support) - (supersedes #3548)

ThomasLecocq:
Second attempt (continuation of #3548) to bring **read and write**  support for miniseed version 3 into obspy.

The principle is simple: create an `io.mseed3` module to allow reading and writing mseed3 files using pymseed's top level objects.

### Todo :

- [ ] Sanitize code
- [x] Bring in mseed 3 test cases
- [ ] Sync the logic with @megies 's #3558 (for me the simplest approach)
- [x] Add more miniseed2 test case ?
- [x] Have @chad-earthscope 's help for the byteorder stuff 😇 
- [x] Evaluate the complexity to add mseed3 write support using pymseed

### PR Checklist
- [ ] All tests still pass.


ThomasLecocq:
@chad-earthscope here is a first pass on using pymseed instead of mseedlib, I added some write support based on the example in the repo. Looking forward to discussing with you.

chad-earthscope:
Hi @ThomasLecocq 

Some comments/questions after a quick perusal of obspy/io/mseed3/core.py

* Reading _is_mseed3() makes me want to expose the population of a MS3TraceList from a buffer: https://github.com/EarthScope/pymseed/issues/1 to avoid needing to write temporary files like that, at least when the data is in a buffer.  This still wouldn't support all file-like objects.
* I suggest doing MS3TraceList.from_file(filename, ...) instead of without the ".from_file".
* There is code for checking encodings (`UNSUPPORTED_ENCODINGS`) independently and throwing ValueError when it's not supported.  Is this not something to leave to the library?
* Also code checking for valid record lengths.  Is this not something to leave to the library?  I suppose you may want to do this so that you can provide a clear reason for what would be an error (because the C library does not have specific returns for those cases).  Alternatively, it might be possible to configure the logging system to capture the library logs and then print them on errors, I'll consider how this might be done.
* byte order determination: beyond informational for the human user I do not see how this detail is needed by ObsPy.  ObsPy should not be reading anything directly from the records and therefore shouldn't need to know anything about the byte order since it's all handled for you.  Regarding "information for the human user", I'm dubious that's much use either.  It's only meaningful for individual records, even those can have mixed orders where header is different from payload.  Information about whether byte swapping was needed for reading is available at MS3Record.swapflag or more usefully at MS3Record.swapflag_dict(); this could be used with information about the system order to determine the order of the data.
* Mapping of publication version value of 0 to "X" data quality is non-standard, is that really needed?
* `hasattr(record, 'pubtime')` pymseed does not set pubtime FWIW, maybe that's done in ObsPy?
* Handling v2 versus v3 header bit flags will require custom parsing of records.  pymseed/libmseed automatically normalizes the 3 flags that matter into MS3Record.flags/MS3Record.flags_dict()
*  I'm suspicious the parsing of FDSN Source IDs <-> SEED NSLC is not right.  I recommend using either pymseed's mapping functions (`sourceid2nslc()` and `nslc2sourceid()`) or the Python module here:https://github.com/FDSN/source-identifiers/blob/main/fdsn_source_identifiers.py
For example, this is incorrect:
```python
                # For MSEED3, use FDSN format: FDSN:NET_STA_LOC_CHAN_SUB_SUBSUB
                sourceid = f"FDSN:{network}_{station}_{location}_{channel}__"
```
* "# Default to big-endian for MSEED".  miniSEED v3 is little-endian headers always, most encodings little endian, except Steim[12] which are big endian.  A good example of why ObsPy should offload and ignore this detail.
* `byteorder`, `Blkt1000`, `word_order` attributes, and `rec.msr.Blkt1000` are not going to exist.
* "# Convert numpy array to list for pymseed" that is more data movement than needed.  pymseed will directly use data from a memoryview if the object supports it, and will convert to a list internally if needed.   You should provide it with numpy arrays.

We were working in parallel and just today I released pymseed v0.0.2 with some API changes (easily adapted), which are an improvement in my humble opinion.

When I get more time I'll write up how I think you may want to use the record list functionality, which allows not only lazy-loading decompression of samples, but also decompression directly into a buffer of your choice.  It does require a 2nd pass through the data but can eliminate a lot of data movement.

Great start!

ThomasLecocq:
> * Reading _is_mseed3() makes me want to expose the population of a MS3TraceList from a buffer: [Implement MS3TraceList.from_buffer() EarthScope/pymseed#1](https://github.com/EarthScope/pymseed/issues/1) to avoid needing to write temporary files like that, at least when the data is in a buffer.  This still wouldn't support all file-like objects.

Perfect, we'll adapt the thin-layer for the non supported cases

> * I suggest doing MS3TraceList.from_file(filename, ...) instead of without the ".from_file".

OK will do

- [ ] TODO use .from_file()

> * There is code for checking encodings (`UNSUPPORTED_ENCODINGS`) independently and throwing ValueError when it's not supported.  Is this not something to leave to the library?

Totally, I tried to mimic what io.mseed was doing, but I believe it's useless now when using pymseed

> * Also code checking for valid record lengths.  Is this not something to leave to the library?  I suppose you may want to do this so that you can provide a clear reason for what would be an error (because the C library does not have specific returns for those cases).  Alternatively, it might be possible to configure the logging system to capture the library logs and then print them on errors, I'll consider how this might be done.

Again, it was to mimic io.mseed with the idea of trying to get mseed2 test files PASS on io.mseed3 too, not sure it's essential now.

> * byte order determination: beyond informational for the human user I do not see how this detail is needed by ObsPy.  ObsPy should not be reading anything directly from the records and therefore shouldn't need to know anything about the byte order since it's all handled for you.  Regarding "information for the human user", I'm dubious that's much use either.  It's only meaningful for individual records, even those can have mixed orders where header is different from payload.  Information about whether byte swapping was needed for reading is available at MS3Record.swapflag or more usefully at MS3Record.swapflag_dict(); this could be used with information about the system order to determine the order of the data.

Again, that comes from io.mseed, having to deal with those. Agree it's simpler to leave it to the library and report in the headers

> * Mapping of publication version value of 0 to "X" data quality is non-standard, is that really needed?

Humm, dunno where I found this, probably in an old io.mseed test file - will check

- [ ] TODO check

> * `hasattr(record, 'pubtime')` pymseed does not set pubtime FWIW, maybe that's done in ObsPy?

Dunno, will check

- [ ] TODO check

> * Handling v2 versus v3 header bit flags will require custom parsing of records.  pymseed/libmseed automatically normalizes the 3 flags that matter into MS3Record.flags/MS3Record.flags_dict()

OK

> * I'm suspicious the parsing of FDSN Source IDs <-> SEED NSLC is not right.  I recommend using either pymseed's mapping functions (`sourceid2nslc()` and `nslc2sourceid()`) or the Python module here:https://github.com/FDSN/source-identifiers/blob/main/fdsn_source_identifiers.py
>   For example, this is incorrect:
> 
> ```python
>                 # For MSEED3, use FDSN format: FDSN:NET_STA_LOC_CHAN_SUB_SUBSUB
>                 sourceid = f"FDSN:{network}_{station}_{location}_{channel}__"
> ```
> 

It was definitely late yesterday... My bad. Indeed, so much better to use the mapping functions

- [ ] TODO replace name mapping with pymseed mapping functions

> * "# Default to big-endian for MSEED".  miniSEED v3 is little-endian headers always, most encodings little endian, except Steim[12] which are big endian.  A good example of why ObsPy should offload and ignore this detail.

Again, mimic io.mseed ... and agree :-)

> * `byteorder`, `Blkt1000`, `word_order` attributes, and `rec.msr.Blkt1000` are not going to exist.

Dunno what implications this could have (if we don't test the old mseed2 files against io.mseed3 at least). Seems fair.

> * "# Convert numpy array to list for pymseed" that is more data movement than needed.  pymseed will directly use data from a memoryview if the object supports it, and will convert to a list internally if needed.   You should provide it with numpy arrays.

- [ ] TODO avoid converting arrays to list - check

> 
> We were working in parallel and just today I released pymseed v0.0.2 with some API changes (easily adapted), which are an improvement in my humble opinion.

Happy to see move forward all together !! Thanks for all the work in pymseed, it's sooooo helpful !

> 
> When I get more time I'll write up how I think you may want to use the record list functionality, which allows not only lazy-loading decompression of samples, but also decompression directly into a buffer of your choice. It does require a 2nd pass through the data but can eliminate a lot of data movement.

Oki, looking forward !

> 
> Great start!

Cheers,
Thom

chad-earthscope:
Fixes and new `MS3TraceList.from_buffer()`  and `MS3TraceList.add_buffer()` in pymseed v0.0.3 available at  https://pypi.org/project/pymseed/

ThomasLecocq:
@chad-earthscope I'll probably spend some of my XMAS free time on that one

chad-earthscope:
> When I get more time I'll write up how I think you may want to use the record list functionality, which allows not only lazy-loading decompression of samples, but also decompression directly into a buffer of your choice. It does require a 2nd pass through the data but can eliminate a lot of data movement.

See https://github.com/EarthScope/pymseed/blob/main/examples/read_numpy.py for an example of reading miniSEED and explicitly decompressing the samples into a pre-allocated array in python-managed memory (as a numpy array).

The `read_traces_to_numpy()` method pre-allocates a numpy array of the correct size and type and then uses the more fundamental `unpack_recordlist()` that unpacks data samples into buffer-protocol storage of your choosing.

The downside of this pattern: reading the data twice.  Not a big deal if it's a buffer in memory, but the double I/O is not great on slow storage systems.  Luckily the CPU load is pretty light for reading the data without decompressing data in the first pass.

The upside of this pattern: decompressed data do not need to be copied and can be decompressed to a python-managed buffer of your choice, which may be a big reduction in data movement.  Furthermore this utilizes libmseed's trace determination so there is no need to figure out how the input records "fit" together, the result is pre-determined traces.

I do not have a good feel for the trade offs regarding when this is worth is and when copying data is acceptable.

chad-earthscope:
@ThomasLecocq
If you are interested I can help with this code, but I could use some help knowing how best to do that.
How to run relevant tests?
I don't think I can push to this branch.  I could make a fork of this branch but this is more difficult to coordinate probably?

megies:
@chad-earthscope I sent you an invite that should let you push to branches on the main repo

ThomasLecocq:
@chad-earthscope yes yes yes please ! probably need to get the branch up to date with master, but that should be easy

chad-earthscope:
OK, MSEED3 reading and writing are in place in this PR using pymseed.
@ThomasLecocq, apologies I took a heavy approach and replaced much of the existing code.

Notes:

1. pymseed added in EXTRAS_REQUIRES, seems like it'd be core but I left it there for you all to decide.

2. I focussed on my notion of the major use case: "reading time series for processing", which omits caring about record-level details.  I think it would be worthwhile to support record-focussed inspection,
but I recommend doing so, even in this module, in code separate from the major use-case.

3. The _is_mseed3() quandary, a failing test in test_waveform_plugins.py:
I think obspy/core/tests/test_waveform_plugins.py tests fail for MSEED3 because the module says it can read the same miniSEED v2 files that the MSEED module can read. Message like: `Exception: False positives for isFormat:` with a list of readable files. Seems like it should be allowed for more than one module to read the same data file, and a preference is maintained.  I didn't see how to specify such a preference or tell the test it's acceptable so that it can _also_ read these data.  Otherwise all other tests pass.

4. ~Some miniSEED v3 should be added for tests.  I didn't know the best approach for that, in some common area or in the module area.  Recommendations welcome, or even better, add them (some example data in [libmseed source](https://github.com/EarthScope/libmseed/tree/main/test/data)).~ Edit: tests added.

I generally tried to match the MSEED io module, but there are deviations as follows.

New functionality:

* read: `sourceid` matching on read. `sourcename` matching implemeneted with a few limitations.
* ~read: `twopass` boolean, just-in-time decompression to minimize memory use decompressing directly into numpy arrays, but requires two reads of the input.  Default: False~
* write: `overwrite` boolean to control overwriting of files.  Default: True

Omitted functionality:

* ~read: `details` is not supported.  It was a lot of detail that mostly belongs in a different calling pattern or as attributes but not at the Trace-level.  It's challangin to convey record-level details in a meaningful way at the Trace level, even more complex when there are two versions of records.~  Edit: `details` arg added.
* write: `flush=bool` not implemented, this could be implemented but for that to make sense it would need to modify the Traces to remove the data packed in order to leave the data that was not packed in the Stream.
* write: anything "byteorder", it's hard to imagine how anyone cares, let the libraries deal with it.

chad-earthscope:
> read: details is not supported.

A `details` argument has been added.  It results in `timing_qualities`, `publication_versions`, and `encodings` being added to the mseed3 stats, e.g.:

```
mseed3: AttribDict({
    'source_id': 'FDSN:US_BW06__B_H_N', 
    'number_of_records': 11,
    'timing_qualities': [74, 13, 39, 0, 11, 19, 0, 40, 46, 44],
    'publication_versions': [3],
    'encodings': ['STEIM-2 integer compression']})
```
These are run-length-deduplicated lists, so they retain the order they appear in the input data, but do not include adjacent duplicated values.

chad-earthscope:
On the topic of FDSN Source IDs I've added a description of the approach taken in this PR to #3559

chad-earthscope:
@ThomasLecocq I hope you're OK that I added the "ready for review" label to get more eyes on this.

chad-earthscope:
The git history in this branch was on a stale master, w/ duplicated commits, resulting in a messy state far behind master.  It was too much to detangle.  I've flattened all the commits for a clean and up to date relationship to master (`--force-with-lease`).  @ThomasLecocq a branch named `wip_pymseed_io_mseed3.oldstate` retains the old state and your original code in case you want to reference it.  Sorry for the noise in this branch.

ThomasLecocq:
Hi Chad,

Sorry for not replying earlier, am travelling in Scotland and didn't see the notifications!! Thank you so much for attacking this PR!! And all the work done here! I'll have a look in the next couple of days too. Cheers 🏴󠁧󠁢󠁳󠁣󠁴󠁿

chad-earthscope:
I've added what I think are some of the final pieces of integration needed.  Two major issues remain:

* pymseed added in EXTRAS_REQUIRES, seems like it'd be core but pymseed requires python >= 3.10, which is EOL in a few months.  If the next release moves up, no problem; but if not, then some strategy is needed.

* The `_is_mseed3()` quandary, the MSEED and MSEED3 modules can _both_ read miniSEED v2 so the tests fail with a bunch of "false positives" that are expected and a good thing in my opinion.  I could work on fixing this, I'd prefer a system that allows more than one module to read a format, but would like to know this is OK with the dev group before proceeding.

ThomasLecocq:
Hi @chad-earthscope , we're on it this week with @megies :-)

PR #3621, opened 2025-09-23
[QC] optimize numpy operations

raciner:
to speed up computation and reduce memory consumption

### What does this PR do?

Some existing code is replaced with numpy calls. This speeds up computation and reduces memory consumption.

The speed was slow and the memory consumption high for large input files.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,10 @@ maintenance_1.4.x
 Changes:
  - obspy.io.shapefile:
    * add support for pyshp v3 (see #3599)
+ - obspy.signal
+   * MSEEDMetadata: Computation of sample metrics replaced with vectorised
+     numpy code. Speedup of roughly factor 10 for integer data and 1.5 for
+     floating point data. (see #3621)
 
 1.4.2 (doi: 10.5281/zenodo.15309143)
 ====================================


obspy/signal/quality_control.py
```diff
@@ -447,31 +447,30 @@ def _compute_sample_metrics(self):
...
+            [tr.data.astype(np.float64, copy=False) for tr in self.data])
+
         # Manually implement these as they have to work across a list of
         # arrays.
-        self.meta['sample_mean'] = \
-            sum(tr.data.sum() for tr in self.data) / npts
+        self.meta['sample_mean'] = full_samples.mean()
 
-        full_samples = np.concatenate([tr.data for tr in self.data])
         self.meta['sample_median'] = np.median(full_samples)
```
ThomasLecocq: could this and the three following two lines be replaced by np.percentile(full_samples, [25,50,75]) ? Is that faster (I would suppose it'd be computing the distribution only once?)

obspy/signal/quality_control.py
```diff
@@ -447,31 +447,30 @@ def _compute_sample_metrics(self):
...
-        self.meta['sample_rms'] = \
-            np.sqrt(sum((tr.data.astype(object) ** 2).sum()
-                        for tr in self.data) / npts)
+        self.meta['sample_rms'] = np.sqrt(np.mean(full_samples ** 2))
 
-        self.meta['sample_stdev'] = np.sqrt(sum(
-            ((tr.data - self.meta["sample_mean"]) ** 2).sum()
-            for tr in self.data) / npts)
+        # Sample standard deviation
+        self.meta['sample_stdev'] = full_samples.std(ddof=0)
```
ThomasLecocq: didn't even know about the ddof thing :-) learning every day (default is 0 but I guess you wanted to be fix it?)

ThomasLecocq:
oh and for the near future, please branch & PR against master, not maintenance - we'll get rid of this branch soon & release directly from master.

obspy/signal/quality_control.py
raciner: Benchmarks with an array of 100_000_000 elements suggest a speedup of a factor of 2 when using your suggestion, I would definitely go for this.

I also wonder, whether we should make use of the fact that `np.mean(full_samples**2)` is already calculated and can be used instead of computing np.std from scratch:

Instead of calling np.std(...), 
```
squared = numpy.mean(full_samples**2) is stored and then 
self.meta['sample_stdev'] = np.sqrt(squared - np.mean(full_samples)**2)
```
as this will potentially avoid computing `np.mean(full_samples**2)` twice. 

My benchmarks suggest that this gives a factor of ~2 as well for computing the standard deviation. I suggest to make this change as well before merging the code.

obspy/signal/quality_control.py
raciner: Yes, the changes are covered by the test test_quality_control.py -> test_overall_sample_metrics.

obspy/signal/quality_control.py
ThomasLecocq: Agree !

raciner: Ok, I just updated the pull request. The updated code should reflect today's discussion.

raciner:
Update to further improve performance as discussed.

ThomasLecocq:
looks good, could you add a line in the changelog too, it's always nice to report on performance improvements when we release :-)

ThomasLecocq (approved):

raciner:
Done.

PR #3622, opened 2025-09-24
Faster read of large miniseed files.

raciner:
### What does this PR do?

This PR can cut down processing times by anorder of magnitude if start and end time are used (e.g. for SDS reading with time window).

It uses memory mapped data representation and interpolation search to find the place which actually must be read instead of fully reading the file.

Original code is slow for time window reads of large files and doesn't make use of the internal order of miniseed files (records sorted in ascending order). 

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -138,6 +138,7 @@ def __is_mseed(fp, file_size):  # NOQA
         fp.seek(record_length - 1, 1)
     return False
 
+
 def __read_header(buffer, index):
     """
         Reads the first record of the given buffer and returns
@@ -152,6 +153,7 @@ def __read_header(buffer, index):
     header = _read_mseed(io.BytesIO(buffer[index:index+4096].tobytes()),
                          headonly=True)
     return header[0].stats
+
 
 def __bisect_mseed(buffer, timestamp, before=True):
     """
@@ -233,8 +235,6 @@ def __bisect_mseed(buffer, timestamp, before=True):
         return low
     else:
         return high
-
-
 
 
 def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
--- a/obspy/io/mseed/core.py
+++ b/obspy/io/mseed/core.py
@@ -138,6 +138,104 @@ def __is_mseed(fp, file_size):  # NOQA
         fp.seek(record_length - 1, 1)
     return False
 
+def __read_header(buffer, index):
+    """
+        Reads the first record of the given buffer and returns
+        header information.
+
+        :param buffer: numpy.array(dtype=int8) or compatible containing
+                       raw miniseed data.
+        :param index:  position in the buffer which should be read
+
+        returns obspy.core.trace.Stats object
+    """
+    header = _read_mseed(io.BytesIO(buffer[index:index+4096].tobytes()),
+                         headonly=True)
+    return header[0].stats
+
+def __bisect_mseed(buffer, timestamp, before=True):
+    """
+         Quickly finds a timestamp in a miniseed stream using interpolation
+         search.
+
+         The code is a helper for _read_mseed and is written to be efficient,
+         a record size of 4096 is assumed. The code stops when the area
+         searched is less than 1MB in size. This code is intended to fasten
+         up the code when trying to cut out a time window of data from a
+         large input file. Using this, the code potentially only needs to
+         actually read a very small portion of a file. Cutting off at 1MB
+         also eliminates the need of treating boundary conditions.
+
+         The algorithm used is "interplation search" as described by W.W.
+         Peterson 1957: doi:10.1147/rd.12.0130.
+
+         This code only works when the data is ordered (i.e. the time
+         stamps are ordered in ascending order). Otherwise the code
+         may fail, in this case, None is returned.
+
+         :param buffer: numpy.array(dtype=int8) or compatible containing
+                        raw miniseed data.
+         :param timestamp: obspy.core.UTCDateTime() timestamp to look for
+                           in buffer
+         :param before: If True, return the index of the record containing
+                        the timestamp, if False, return the index of the
+                        next record.
+
+         returns index of the buffer either including the timestamp or
+                 the next one or None if the time stamp cannot be found.
+    """
+    low = 0
+    high = len(buffer)
+
+    # cut off here, linear search will be fast enough, avoid treating
+    # boundary cases
+    stopat = 1024*1024
+
+    if high - low <= stopat:
+        if before:
+            return low
+        else:
+            return high
+
+    starttime = __read_header(buffer, low).starttime
+    endtime = __read_header(buffer, high-4096).endtime
+
+    # timestamp not in this file
+    if timestamp < starttime or timestamp > endtime:
+        return None
+
+    while high-low > stopat:
+
+        # Cannot find timestamp
+        if timestamp < starttime or timestamp > endtime:
+            return None
+
+        # classical interpolation search
+        informed_guess_percentage = (timestamp - starttime) / \
+                                    (endtime - starttime)
+        buffer_raw_guess = informed_guess_percentage * (high-low)
+        informed_guess = low + int(buffer_raw_guess // 4096) * 4096
+
+        # algorithm stuck due to rounding to 4096 boundaries
+        if informed_guess <= low or informed_guess >= high:
+            informed_guess = (high - low) / 2
+            informed_guess = low + int(informed_guess // 4096) * 4096
+
+        guess_time = __read_header(buffer, informed_guess).starttime
+
+        if guess_time > timestamp:
+            high = informed_guess
+            endtime = guess_time
+        else:
+            low = informed_guess
+            starttime = guess_time
+    if before:
+        return low
+    else:
+        return high
+
+
+
 
 def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
                 sourcename=None, reclen=None, details=False,
@@ -306,7 +404,10 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
     # If it's a file name just read it.
     if isinstance(mseed_object, str):
         # Read to NumPy array which is used as a buffer.
-        bfr_np = np.fromfile(mseed_object, dtype=np.int8)
+        # use memmap, faster than fromfile, same functionality
+        bfr_np = np.memmap(mseed_object, dtype=np.int8, mode="c")
+        # bfr_np = np.fromfile(mseed_object, dtype=np.int8)
+
     elif hasattr(mseed_object, 'read'):
         bfr_np = from_buffer(mseed_object.read(), dtype=np.int8)
 
@@ -333,6 +434,8 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
         break
     bfr_np = bfr_np[offset:]
     buflen = len(bfr_np)
+    bufptr_low = 0
+    bufptr_high = buflen
 
     # If no selection is given pass None to the C function.
     if starttime is None and endtime is None and sourcename is None:
@@ -347,6 +450,12 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
                 raise ValueError(msg)
             selections.timewindows.contents.starttime = \
                 util._convert_datetime_to_mstime(starttime)
+
+            # bisect to the actual starttime
+            # avoids parsing of the whole file
+            bufptr_low = __bisect_mseed(bfr_np, starttime, before=True) \
+                or 0
+
         else:
             # HPTERROR results in no starttime.
             selections.timewindows.contents.starttime = HPTERROR
@@ -356,6 +465,9 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
                 raise ValueError(msg)
             selections.timewindows.contents.endtime = \
                 util._convert_datetime_to_mstime(endtime)
+            bufptr_high = __bisect_mseed(bfr_np, endtime, before=False) \
+                or buflen
+
         else:
             # HPTERROR results in no starttime.
             selections.timewindows.contents.endtime = HPTERROR
@@ -396,9 +508,10 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
     clibmseed.verbose = bool(verbose)
     try:
         lil = clibmseed.readMSEEDBuffer(
-            bfr_np, buflen, selections, C.c_int8(unpack_data),
-            reclen, C.c_int8(verbose), C.c_int8(details), header_byteorder,
-            alloc_data)
+            bfr_np[bufptr_low:bufptr_high], bufptr_high - bufptr_low,
+            selections, C.c_int8(unpack_data), reclen, C.c_int8(verbose),
+            C.c_int8(details), header_byteorder, alloc_data)
+
     except InternalMSEEDError as e:
         msg = e.args[0]
         if offset and offset in str(e):


Jollyfant:
Would this fail for mSEED files with record sizes larger than 4096? I imagine the algorithm may bissect into the middle of a mSEED record and be unable to read the header.

ThomasLecocq:
Hey all,

Honestly I wouldn't work on that, my free time is dedicated to making the mseed3 wrapper, not to adapt the current version.

raciner:
> Would this fail for mSEED files with record sizes larger than 4096? I imagine the algorithm may bissect into the middle of a mSEED record and be unable to read the header.

Yes, it would. I just checked the standard. It says, that as of now (2005) larger record sizes (> 4096) have not been written, but doesn't explicitly forbid them. I understand that incorporating this code as it is might break some applications. However, at least I still need a solution right now for large files, as we have ongoing experiments and it will take ages until manufacturers incorporate mseed v3. Typically, they files in an SDS structure and only a few seconds of data are needed at a time. Reading a chunk of a few seconds of data currently fails, if the whole file is larger than 2GB, even if the chunk itself would just be a few MB. I would be happy to collaborate to get a cleaner solution which could be used by others or otherwise just create a local subclass which solves the problems for me.

ThomasLecocq:
Hi Roman,
Sorry for my sharp message above, unnecessarily short.

The approach you propose if great, I wonder if it could be generalized using "any" record length (found in the header)  ? It'd indeed be a good approach ahead of mseed2+3 support (even if e.g. my pymseed wrapper works for mseed2>2GB too - but it'll take more time to be ready & released).

Tom

raciner:
Hi Thomas,

mseed3 comes with a different header structure which will make quick searches difficult, much to my displeasure... For mseed3, my personal assumption is that some operators will still use fixed length records to allow quick bisectional searches (what e.g. seiscomp currently does on mseed2). For mseed3, probably an optional parameter would be needed, e.g. obspy.core.read(..., fixed_header=True) to enable bisectional search, by default it will have to be switched off.

I guess we will have to solve this issue for mseed3 separately. For the time being, I suggest to alter the code I suggest in this pull request to read the record length from the first record and adjust it, should it be larger than 4096.

ThomasLecocq:
>I guess we will have to solve this issue for mseed3 separately. For the time being, I suggest to alter the code I suggest in this pull request to read the record length from the first record and adjust it, should it be larger than 4096.

Totally agree with this, larger _or smaller_ than 4096 ?

raciner:
I have looked a little bit more into this. In order to make this clean and work in all cases, we will have to determine the record length first. I would write some code to look for blockette 1000 in the first record and read the record length from there. If this fails, or if any of the subsequent optimised code fails, the code should revert to the non-optimised version. Any suggestions on this?

megies:
Hi @raciner sorry for chiming in late. I think it will be great to have this in obspy. :+1: 

Here's some thoughts:

 - like mentioned before, we should be flexible with record length and it's not hard to add. I guess the whole approach assumes and only makes sense if the file is somewhat "regular", like a file saved directly by a digitizer or slarchive or similar, with records strictly ordered and usually that means record length is the same for all records. So yes, I would also go with simply looking up the record length of the first record and using that to move around
 - I agree, there should be the option to use this or not, since there will be failures in unsorted files or multiplexed mseed data
   - either we keep the "safe" slower approach as default and add a kwarg to opt-in to fast mode
   - or we default to fast mode but then we need to carefully catch problems and fall back to slow
   - either way, there should be a kwarg switch, no matter if it's opt-in or opt-out
   - in any case, probably most people would likely not get impacted too much, since most people will liekly not specify starttime/endtime during reading.. but still need to be careful to not hurt people
 - we have a small object oriented tool that can move through records, get header info etc and do things like hop to certain records (assuming same record length). that might make some of the tasks you do simpler potentially (like checking record start time) and maybe it could also be a bit cleaner to add some methods for the bisect there. but I'm not totally opposed to having it in the code directly like so far. see class [RecordAnalyser](https://github.com/obspy/obspy/blob/master/obspy/io/mseed/scripts/recordanalyzer.py#L31) which is used in the command line tool `obspy-mseed-recordanalyzer`

So something like `RecordAnalyzer.bisect(..)` could also be where the fast search logic is. But like I said, food for thought, not opposed to current approach.

raciner:
I have worked on this today. I have addressed the limit of 2GB and something, that bothered me for some time, that I cannot pass raw input data (output data from proprietary software) directly to obspy without using io.BytesIO which will result in a copy of the data being made. 

The biggest benefit in our use case for bisection would be obspy.clients.filesystem.sds. Any software reading a five minute window from a whole day will read all full day files in question at the moment. 

I'm aware that this code will need some more reviews and careful consideration. I have also adjusted the unit tests, however, I couldn't come up with something useful for demonstrating and testing the new ability of reading large inputs (> 2GB). Entirely mocking it will not demonstrate it, actually creating a large input file in memory will likely exceed the memory limits...

obspy/io/mseed/core.py
```diff
@@ -147,44 +265,46 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
+
+    :param mseed_object: Filename, numpy.ndarray with dtype=numpy.int8 or
+        open file like object that contains the binary Mini-SEED data or
+        anything, that can be read into np.nadarray with np.frombuffer.
+        Any object that provides a read() method will be considered to
+        be a file like object.
+    :type starttime: :class:~obspy.core.utcdatetime.UTCDateTime
     :param starttime: Only read data samples after or at the start time.
-    :type endtime: :class:`~obspy.core.utcdatetime.UTCDateTime`
+    :type endtime: :class:~obspy.core.utcdatetime.UTCDateTime
```
megies: There's quite few edits that look like automated IDE edits, that I'm pretty sure aren't according to Sphinx syntax. I'm triggering a docs build, but I'm pretty sure those need a revert (ideally squashed in to not have it clutter git history)

raciner: This is likely caused by me using another tool chain while working from home rather than in the office. I will take a look.

megies:
> I'm aware that this code will need some more reviews and careful consideration

Like mentioned before, we definitely want a switch in form of a kwarg on `_read_mseed()` which will get passed through from `read()`, likely opt-in to begin with. In that case, the changes would only affect (power) users deliberately opting in, which means they will have read the documentation for that kwarg, where we can put information on potential problems with the approach. Also it only affects people using `starttime`/`endtime` already which is probably a small percentage.
So I'm not really too worried about it.



> I couldn't come up with something useful for demonstrating and testing the new ability of reading large inputs (> 2GB). Entirely mocking it will not demonstrate it, actually creating a large input file in memory will likely exceed the memory limits...

Some things that could potentially be tested..
 - mocking large file size >2GB
   - read not doing something (shows error) without opting in to bisect
   - read doing something with start/endtime and opt in to bisect
 - some test that shows working fallback solution, maybe loading a test file and inversely sorting the records into a BytesIO object and trying to read with bisect which should fail and fallback to normal reading ideally

raciner:
I will work on this as soon as I have time to do so. I thought of having the feature opt-in (available but switched off by default) and having the sds client having it switched on by default, as sds file structures should be well behaved. I will make sure that the code reverts to its default behaviour in case a file is not suitable for bisection (unordered, multiplexed) and make sure this is clearly documented in the code. 

read should work for files larger than 2GB, as the other modification of the code I made will be able to handle this.

raciner:
I cleaned up the code and added proper unit tests. Editing the change log is still missing. I cannot make any sense of the failed tests, they don't look related to my changes in the code. @megies: Could you have a look?

ThomasLecocq:
looks like a series of pyproj issues ? https://tests.obspy.org/147552/ aka unrelated
but this one is numpy version related: https://github.com/obspy/obspy/actions/runs/19267874178/job/55088107698?pr=3622

raciner:
> but this one is numpy version related: https://github.com/obspy/obspy/actions/runs/19267874178/job/55088107698?pr=3622

Doesn't make sense to me, I didn't touch the file where it fails or anything in the entire module. I wonder why it only fails for version 3.8.

ThomasLecocq:
Hi
Sorry my message wasn't clear: it's not you, it's linked to dependencies and their own minimal dependencies...

raciner:
I documented the changes in the changelog, there is not much I can do to fix the failing unit tests, I think. I suggest that I add a kwargs option to obspy.clients.filesystem.SDS to make use of the new fast read ability and consider the code ready for review afterwards.

ThomasLecocq:
the two tests are indeed related to changes in PIL and probably pyproj, not your PR.

raciner:
The changes can be used with code like

```
import obspy.clients.filesystem.sds
import obspy.core
client = obspy.clients.filesystem.sds.Client("/miniseed")
st = client.get_waveforms(network="8R",station="*",location="*",channel="FHZ",
    starttime=obspy.core.UTCDateTime(2025,11,21,1,0,0),
    endtime=obspy.core.UTCDateTime(2025,11,21,1,10,0),
    use_bisection=True)
```

So far, my objectives are fulfilled. I will add the ready for review label again.

ThomasLecocq:
@raciner could you rebase this (or cherry pick) on master ? And if possible reduce the size of the new test files ?

ThomasLecocq:
@raciner I'll go ahead and rebase + check the last typos, except if you have time for this?

raciner:
@ThomasLecocq: I am back from holidays now, I will have a look asap.

ThomasLecocq:
@raciner I tried to trigger Copilot to see what this AI buddy says... Not sure all is applicable, but there are actually a few interesting points, e.g. the warnings, the memmap link on tr.data etc:

[wed_jan_07_2026_pr_analysis_and_quality_assessment.md](https://github.com/user-attachments/files/24478224/wed_jan_07_2026_pr_analysis_and_quality_assessment.md)

raciner:
> @raciner I tried to trigger Copilot to see what this AI buddy says... Not sure all is applicable, but there are actually a few interesting points, e.g. the warnings, the memmap link on tr.data etc:
> 
> [wed_jan_07_2026_pr_analysis_and_quality_assessment.md](https://github.com/user-attachments/files/24478224/wed_jan_07_2026_pr_analysis_and_quality_assessment.md)

I disagree with the potential memory usage and performance issues it raises, as these are the same issues that already exist in the current code. The code I introduced will use memmap if the parameter given to the library is a path to a file, and one is able to pass a memory like object to the code which contains the content of a miniseed file. In all other cases, the performance will stay the same as it currently is, including the potential of large memory consumption. However, there is one difference in behaviour. In case of a memory mapped file, deleting it while the file is memory mapped may cause issues with non Unix-style operating systems. At least, this should be documented. Regarding 3: read_header is cheap enough the way it is implemented... Regarding 4) ok, documentation should be improved.

Formatting problems: will be addressed...

raciner:
@ThomasLecocq apparently I did not do the right thing when trying to rebase. Maybe I will need some help to get this fixed...

ThomasLecocq:
@raciner I changed the base here on the web GUI too, looks better already 😇 
image

raciner:
The documentation is updated. Are there any dangling threads?

obspy/io/mseed/core.py
```diff
@@ -258,29 +407,41 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
         bo = None
 
-    # Determine total size. Either its a file-like object.
-    if hasattr(mseed_object, "tell") and hasattr(mseed_object, "seek"):
-        cur_pos = mseed_object.tell()
-        mseed_object.seek(0, 2)
-        length = mseed_object.tell() - cur_pos
-        mseed_object.seek(cur_pos, 0)
-    # Or a file name.
+    use_bisection = 'use_bisection' in kwargs and kwargs['use_bisection']
```
claudiodsf: Maybe use here: `use_bisection = kwargs.get('use_bisection', False)`

obspy/io/mseed/core.py
```diff
@@ -371,6 +538,84 @@ def isdigit(x):
                 encode('ascii', 'ignore')
         else:
             selections.srcname = b'*'
+    if bufptr_low > bufptr_high:
+        warnings.warn("File is not ordered, not using bisection.")
+        bufptr_low = 0
+        bufptr_high = buflen
+    # workaround for files which are larger than 2**31
```
claudiodsf, line 576: Replace `2**31` in comments with `LIBMSEED_MAX`

obspy/io/mseed/core.py
```diff
@@ -348,6 +504,13 @@ def isdigit(x):
                 raise ValueError(msg)
             selections.timewindows.contents.starttime = \
                 util._convert_datetime_to_mstime(starttime)
+
+            # bisect to the actual starttime
+            # avoids parsing of the whole file
+            bufptr_low = use_bisection and \
+                __bisect_mseed(bfr_np, starttime, record_length,
+                               before=True) or 0
```
claudiodsf: Looks like this expression returns `0` when `__bisect_mseed()` returns `None`. So, what is the point of returning `None` (instead of  `0`) in `__bisect_mseed()`?

obspy/io/mseed/core.py
```diff
@@ -371,6 +538,84 @@ def isdigit(x):
...
+    # workaround for files which are larger than 2**31
+    #
+    # Basic layout
+    # - split the memory mapped data into chuncks of size 2**31 - record_length
+    # for each chunk:
+    #    - process chunk
+    #    - merge it with predecessing chunck if possible
+    # - Return stream object.
+    _max = obspy.io.mseed.headers.LIBMSEED_MAX
+    if bufptr_high - bufptr_low > _max - record_length:
```
claudiodsf: The code after this `if` clause is a lot indented. Would it be possible (and more readable) to move it to a helper function?

claudiodsf:
Hi @raciner. Thanks for this PR! 

@ThomasLecocq asked a few persons (including myself) to review the remaining PRs before 1.5.0 can be released.

So, I decided to tackle this one with a few comments.

Please, let me know,
Claudio

obspy/io/mseed/core.py
```diff
@@ -258,29 +407,41 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
-        length = mseed_object.tell() - cur_pos
-        mseed_object.seek(cur_pos, 0)
-    # Or a file name.
+    use_bisection = 'use_bisection' in kwargs and kwargs['use_bisection']
+    # If it's a file name just read it.
+    if isinstance(mseed_object, str):
+        # Read to NumPy array which is used as a buffer.
+        # use memmap, faster than fromfile, same functionality
+        bfr_np = np.memmap(mseed_object, dtype=np.int8, mode="c")
+        # bfr_np = np.fromfile(mseed_object, dtype=np.int8)
```
claudiodsf: Remove this commented code

obspy/io/mseed/core.py
```diff
@@ -206,21 +355,21 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
     1 Trace(s) in Stream:
     BW.UH3..EHZ | 2010-06-20T00:00:00.999999Z - ... | 200.0 Hz, 242 samples
 
-    Read with ``details=True`` to read more details of the file if present.
+    Read with `details=True to read more details of the file if present.
```
claudiodsf: This looks like an accidental change that should be reverted. I think that using double quotes here is correct.

obspy/io/mseed/core.py
```diff
@@ -206,21 +355,21 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
 
     >>> st = read("/path/to/timingquality.mseed", details=True)
     >>> print(st[0].stats.mseed.blkt1001.timing_quality)
     55
 
-    ``False`` means that the necessary information could not be found in the
+    `False means that the necessary information could not be found in the
```
claudiodsf: Same here

obspy/io/mseed/core.py
```diff
@@ -206,21 +355,21 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
     file.
 
     >>> print(st[0].stats.mseed.calibration_type)
     False
 
     Note that each change in timing quality from record to record may trigger a
     new Trace object to be created so the Stream object may contain many Trace
-    objects if ``details=True`` is used.
+    objects if `details=True is used.
```
claudiodsf: Same here

obspy/io/mseed/core.py
```diff
@@ -140,6 +142,147 @@ def __is_mseed(fp, file_size):  # NOQA
...
+            guess_header = __read_header(buf, informed_guess,
+                                         record_length)
+            if guess_header.id != startid:
+                warnings.warn("Found two different stream ids in input data, \
+                    reverting to default algorithm.")
+                return None
+            guess_time = guess_header.stats.starttime
+
+            # time stamps not in ascending order
+            if guess_time < starttime or guess_time > endtime:
```
claudiodsf: More pythonic: `if not (starttime <= guess_time <= endtime):`

obspy/io/mseed/core.py
```diff
@@ -15,11 +15,13 @@
...
-               ObsPyMSEEDFilesizeTooLargeError, ObsPyMSEEDError)
+               ObsPyMSEEDError)
 from .headers import (DATATYPES, ENCODINGS, HPTERROR, HPTMODULUS, SAMPLETYPE,
                       SEED_CONTROL_HEADERS, UNSUPPORTED_ENCODINGS,
-                      VALID_CONTROL_HEADERS, VALID_RECORD_LENGTHS, Selections,
-                      SelectTime, Blkt100S, Blkt1001S, clibmseed)
+                      VALID_CONTROL_HEADERS, VALID_RECORD_LENGTHS,
+                      Selections, SelectTime, Blkt100S,
+                      Blkt1001S, clibmseed)
+import obspy.io.mseed.headers
```
claudiodsf: Can't you just import `LIBMSEED_MAX` in the import above? (i.e., `from .headers import LIBMSEED_MAX`)

obspy/io/mseed/core.py
```diff
@@ -140,6 +142,147 @@ def __is_mseed(fp, file_size):  # NOQA
     return False
 
 
+def __read_header(buf, index, record_length):
+    """
+        Reads the first record of the given buffer and returns
```
claudiodsf: The docstring text is indented one level too much

obspy/io/mseed/core.py
```diff
@@ -140,6 +142,147 @@ def __is_mseed(fp, file_size):  # NOQA
...
+    """
+
+    header = _read_mseed(buf[index:index+record_length], reclen=record_length,
+                         headonly=True)
+    return header[0]
+
+
+def __bisect_mseed(buf, timestamp, record_length, before=True):
+    """
+         Quickly finds a timestamp in a miniseed stream using interpolation
```
claudiodsf: The docstring text is indented one level too much

obspy/io/mseed/core.py
```diff
@@ -140,6 +142,147 @@ def __is_mseed(fp, file_size):  # NOQA
...
+    step_size = 4096
+    if record_length > 4096:
+        step_size = record_length
+    stopat = 16 * step_size
+
+    if high - low <= stopat:
+        if before:
+            return low
+        else:
+            return high
```
claudiodsf, line 217: More pythonic: `return low if before else high`

obspy/io/mseed/core.py
```diff
@@ -140,6 +142,147 @@ def __is_mseed(fp, file_size):  # NOQA
...
+            warnings.warn(e)
+            return None
+
+        if guess_time > timestamp:
+            high = informed_guess
+            endtime = guess_time
+        else:
+            low = informed_guess
+            starttime = guess_time
+    if before:
```
claudiodsf, line 279: More pythonic: `return low if before else high`

obspy/io/mseed/core.py
raciner: Although it doesn't make a difference in this logic, I think, there may be reasons in the future to distinguish between None (algorithm failed) and 0 (algorithm terminated with byte position 0 as result).

obspy/io/mseed/core.py
raciner: The current style is clearer to me.

obspy/io/mseed/core.py
raciner: Same as above.

raciner:
I think I addressed the raised concerns and suggestions. I adopted most of the suggestions and gave a reason why I didn't in a few cases.

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -5,11 +5,11 @@
 import os
 import platform
 import random
-import re
+# import re
```
claudiodsf: There are a few commented code lines in this file that should be removed

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -5,11 +5,11 @@
...
 import signal
 import sys
 import warnings
-from unittest import mock
+# from unittest import mock
```
claudiodsf: See above

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -19,8 +19,8 @@
 from obspy.core.util.attribdict import AttribDict
 from obspy.core.util.base import CatchAndAssertWarnings
 from obspy.io.mseed import (InternalMSEEDError, InternalMSEEDWarning,
-                            ObsPyMSEEDFilesizeTooSmallError,
-                            ObsPyMSEEDFilesizeTooLargeError)
+                            ObsPyMSEEDFilesizeTooSmallError)
+#                            ObsPyMSEEDFilesizeTooLargeError)
```
claudiodsf: See above

obspy/io/mseed/tests/test_mseed_special_issues.py
```diff
@@ -1132,23 +1132,24 @@ def test_reading_less_than_128_bytes(self, testdata):
...
+#        getsize_mock.return_value = 2 ** 31 + 1
+#        filename = testdata['BW.BGLD.__.EHE.D.2008.001.first_10_records']
+#        msg = (
+#            "ObsPy can currently not directly read mini-SEED files that are "
+#            "larger than 2^31 bytes (2048 MiB). To still read it, please "
+#            "read the file in chunks as documented here: "
+#            "https://github.com/obspy/obspy/pull/1419#issuecomment-221582369")
+#        msg = re.escape(msg)
+#        with pytest.raises(ObsPyMSEEDFilesizeTooLargeError, match=msg):
+#            _read_mseed(filename)
```
claudiodsf: See above

obspy/io/mseed/core.py
```diff
@@ -140,6 +141,177 @@ def __is_mseed(fp, file_size):  # NOQA
...
+    """
+
+    header = _read_mseed(buf[index:index+record_length], reclen=record_length,
+                         headonly=True)
+    return header[0]
+
+
+def __bisect_mseed(buf, timestamp, record_length, before=True):
+    """
+     Quickly finds a timestamp in a miniseed stream using interpolation
```
claudiodsf: Sorry for being picky, but there is still an extra indent space 🙄

obspy/io/mseed/core.py
claudiodsf: Fair enough ;-)

obspy/io/mseed/core.py
claudiodsf: Fair enough ;-)

obspy/io/mseed/core.py
```diff
@@ -140,6 +141,177 @@ def __is_mseed(fp, file_size):  # NOQA
...
+            starttime = guess_time
+    if before:
+        return low
+    else:
+        return high
+
+
+def _can_merge(prev, curr, details):
+    """
+     check whether two obspy.core.trace objects are
```
claudiodsf: Extra indented space here as well

obspy/io/mseed/core.py
```diff
@@ -140,6 +141,177 @@ def __is_mseed(fp, file_size):  # NOQA
     return False
 
 
+def __read_header(buf, index, record_length):
```
claudiodsf: Not sure about the ObsPy code conventions here: has a double underscore a particular meaning (`__read_header()`) ?

obspy/io/mseed/core.py
```diff
@@ -140,6 +141,177 @@ def __is_mseed(fp, file_size):  # NOQA
...
+
+    returns obspy.core.trace.Trace object without data
+    """
+
+    header = _read_mseed(buf[index:index+record_length], reclen=record_length,
+                         headonly=True)
+    return header[0]
+
+
+def __bisect_mseed(buf, timestamp, record_length, before=True):
```
claudiodsf: Same comment on double underscore

claudiodsf:
Thanks @raciner for addressing my comments! I have a few extra ones, before this is good to go 😉

raciner:
I fixed everything pointed out, I think.

ThomasLecocq:
@raciner what happens if I don't pass starttime&endtime to read a >2GB file , would that work now ?

raciner:
> @raciner what happens if I don't pass starttime&endtime to read a >2GB file , would that work now ?

Yes, it will. The code will split up the data in to suitable chunks and reconnect them at the end.

obspy/io/mseed/core.py
claudiodsf: @ThomasLecocq is this stile valid for ObsPy code conventions? Shouldn't it be just one underscore?

obspy/io/mseed/core.py
claudiodsf: @raciner this has still to be fixed 🙄

claudiodsf:
Thanks @raciner! Just a couple of things missing (see comments). One needs input from @ThomasLecocq

ThomasLecocq:
@claudiodsf I don't see open comments ?

claudiodsf:
@raciner @ThomasLecocq The comments are now published. Sorry for the confusion

claudiodsf:
Pinging again @raciner and @ThomasLecocq on this: only a few minor details remaining!

raciner:
> Pinging again @raciner and @ThomasLecocq on this: only a few minor details remaining!

Sorry for the waiting time, I completely forgot. The remaining issues should now be addressed.

claudiodsf:
> Sorry for the waiting time, I completely forgot. The remaining issues should now be addressed.

Great! Let's wait for the tests, then I would say that this is good to go.

claudiodsf:
This is ok for me!

@ThomasLecocq, is it up to you to merge this PR?

obspy/io/mseed/core.py
```diff
@@ -258,29 +436,40 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
         bo = None
 
-    # Determine total size. Either its a file-like object.
-    if hasattr(mseed_object, "tell") and hasattr(mseed_object, "seek"):
-        cur_pos = mseed_object.tell()
-        mseed_object.seek(0, 2)
-        length = mseed_object.tell() - cur_pos
-        mseed_object.seek(cur_pos, 0)
-    # Or a file name.
+    use_bisection = kwargs.get('use_bisection', False)
```
megies: Was there a specific reason to have the `use_bisect` inside `kwargs`? I think it would be clearer to just have it as another argument. That's how we do across the code base usually (except some really messy cases like the waveform plotting 🙈 ).
I see that we had `**kwargs` before already, but I don't think it was used for anything, at least not that I can see. I think the reason we usually always have it in the low-level routines as well is because all these internal routines get usually called from top-level `read` which accepts some custom arguments for other I/O routines too and then we usually always have `**kwargs` at the end to just keep a container for everything that might not even be relevant.

obspy/io/mseed/core.py
```diff
@@ -258,29 +436,40 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
-        msg = ("ObsPy can currently not directly read mini-SEED files that "
-               "are larger than 2^31 bytes (2048 MiB). To still read it, "
-               "please read the file in chunks as documented here: "
-               "https://github.com/obspy/obspy/pull/1419"
-               "#issuecomment-221582369")
-        raise ObsPyMSEEDFilesizeTooLargeError(msg)
 
-    info = util.get_record_information(mseed_object, endian=bo)
+    # get_record_information expects a file like object
+    first_record = io.BytesIO(bfr_np[0:1048576])
```
megies, line 473: Could change this to `VALID_RECORD_LENGTHS[-1]` I guess, but then it might get changed unintentionally if that variable ever gets changed, so I guess maybe it's better to keep as is. 👍

megies:
Sorry about the wait, I looked through it in detail and I think we can merge on green CI.
Thanks @raciner, that's great additions. Personally I don't use DAS data much, but I understand it might have been a big pain point not to have this in obspy faster for people that do.

The only thing left that I think should be changed is `use_bisection` left in the `kwargs`. I think we should just have it as a new argument after the others.

obspy/io/mseed/core.py
raciner: Ok, I made it an optional argument.

obspy/io/mseed/core.py
raciner: Ok, we will leave it...

ThomasLecocq:
MergeTimeGotGIF

obspy/io/mseed/core.py
```diff
@@ -176,19 +355,20 @@ def _read_mseed(mseed_object, starttime=None, endtime=None, headonly=False,
...
         - ``4`` : Generic Calibration  (Blockette 390)
         - ``-2`` : Calibration Abort (Blockette 395)
-
     :type header_byteorder: int or str, optional
     :param header_byteorder: Must be either ``0`` or ``'<'`` for LSBF or
         little-endian, ``1`` or ``'>'`` for MBF or big-endian. ``'='`` is the
         native byte order. Used to enforce the header byte order. Useful in
         some rare cases where the automatic byte order detection fails.
-
+    :param use_bisection: If ``True`` use bisection for start and end time.
```
claudiodsf: Last, last thing. Maybe we should better explain why this is useful?
Here's an example generated by ChatGPT 😉 

```
:param use_bisection: If ``True``, use bisection to locate the start and end
    times directly in the file, avoiding a full read. This can reduce
    processing time by up to an order of magnitude when a time window is
    specified, and is particularly important for large miniSEED files
    (e.g., > 2 GB). Only applies to strictly ordered, single-channel files;
    the code falls back to the default behavior otherwise.
```

claudiodsf:
@raciner @ThomasLecocq 

Last, last comment regarding the docstring (see above ⬆️ )

obspy/io/mseed/core.py
raciner: Ok, I have adopted this.

claudiodsf:
Great! Good to merge for me if CI tests pass

megies:
Really nice PR, thanks again! :rocket:

megies (approved):

ThomasLecocq:
WOOOOOOOOOOOOOOOOOOOW :-) 🥳

claudiodsf:
👏 👏 👏

Issue #3617, opened 2025-09-16
Add IGN to obspy.clients.fdsn.header URL_MAPPINGS


JBarcoIGN:
### Problem

Dear developers,

On behalf of the Spanish National Geographic Institute, which oversees the National Seismic Network of Spain, we would like to request the inclusion of the following information in the predefined services in obspy.clients.fdsn.header.URL_MAPPINGS:

Id: IGN
URL: http://fdsnws.sismologia.ign.es

Thank you very much in advance,

Jaime Barco IT Department / National Geographic Institute

ThomasLecocq:
Hi @JBarcoIGN , it's HTTP not HTTPS, correct ?

ThomasLecocq:
I see it's properly indexed there too:

https://www.fdsn.org/ws/datacenters/1/query

ThomasLecocq:
linked to works in https://github.com/obspy/obspy/issues/3556 - if the URL_MAPPINGS gets updated, IGN will be there.

JBarcoIGN:
Hi @ThomasLecocq,

Yes, it is correct.

Now is HTTP, although HTTPS is planned.

Issue #3556, opened 2025-03-20
FDSN datacenter URL mapping: switch to https?


petrrr:
I would like to suggest that we switch to protocoll `https://` for the URL mapping if the datacentre supports  that protocol, and above all if it supports authentication. This is the case for most EIDA datacentres. 

#### Motivation:
The main motivation is indeed related to authenticated requests. For good practice we should discourage the transmission of secrets over an unencrypted connections. Also, some datacentres may choose to enable authentication only over https. 

Currently, if the direct client is used with the abbreviation it will translate to `http://` protocol and auth tokens or user/secret is transmitted unencrypted. If the datacentre does not support authenticated requests via `http://` the request will fail. 

https://github.com/obspy/obspy/blob/28472163f63efeb4a60de6afca64e16e461275f7/obspy/clients/fdsn/header.py#L102-L136

This is a simple change. 
Just let me how to processed. Does this require a PR or is a simple commit to the repo sufficient.

filefolder:
Just tested connecting to these with https. The below only seem to work with http

http://eida-federator.ethz.ch
http://batsws.earth.sinica.edu.tw
http://rtserve.beg.utexas.edu

I guess could just change everything but those?

megies:
Totally 👍 
Feel free to make a PR against master branch

megies:
> For good practice we should discourage the transmission of secrets over an unencrypted connections

Sounds good, we could issue a warning if doing an authenticated request over http

petrrr:
Hi @megies! 

I realise there is already some provisions in the code to avoid passing tokens over an unsecure http connection:

https://github.com/obspy/obspy/blob/master/obspy/clients/fdsn/client.py#L413-L420

An https connection is enforced for the `auth` method, which is good. 

This still might have some disadvantages:

1. This approach assumes the `https://` protocoll has been correctly implemented, which in EIDA should be the case. But ... 
2. Also the `queryauth` should better not use `http` either, because the user:pw provided ?
3. Service discovery (`application.wadl` parsing) is still done over `http`. This might result in inconsistencies. I would actually consider it legitimate to exclude `auth` and `queryauth` methods for http protocol.

Maybe ...

1. the protocol replacement this protocol replacement should occur immediately on init and
2. a warning should be emitted if this replacement occurs

megies:
[...]

PR #3628, opened 2025-10-30
manual refresh of URL_MAPPINGS / change all IRIS links to httpS

filefolder:
simple manual update adding some new datacenters and updating http --> https as needed (e.g. #3556, #3617)

note that not all datacenters are listed in the service at `https://www.fdsn.org/ws/datacenters/1/query`, so we may have to continue to do this manually for some time.

based against master, all links tested

edit: also had to update some test references to http://iris

### DIFF
--- a/obspy/__init__.py
+++ b/obspy/__init__.py
@@ -68,24 +68,6 @@ _add_format_plugin_table(Stream.write, "waveform", "write", numspaces=8)
 _add_format_plugin_table(Catalog.write, "event", "write", numspaces=8)
 _add_format_plugin_table(Inventory.write, "inventory", "write", numspaces=8)
 
-if int(sys.version[0]) < 3:
-    raise ImportError("""You are running ObsPy >= 1.3 on Python 2
-
-ObsPy version 1.3 and above is not compatible with Python 2, and you still
-ended up with this version installed. This should not have happened.
-Make sure you have pip >= 9.0 and setuptools >= 24.2:
-
- $ pip install pip setuptools --upgrade
-
-Your choices:
-
-- Upgrade to Python 3.
-
-- Install an older version of ObsPy:
-
- $ pip install 'obspy<1.3'
-""")
-
 
 if __name__ == '__main__':
     import doctest
--- a/obspy/clients/fdsn/tests/test_mass_downloader.py
+++ b/obspy/clients/fdsn/tests/test_mass_downloader.py
@@ -13,18 +13,12 @@ import copy
 import logging
 import os
 import shutil
-import sys
 import tempfile
+from http.client import HTTPException
 from socket import timeout as socket_timeout
 from unittest import mock
 
 import pytest
-
-if sys.version_info.major == 2:
-    from httplib import HTTPException
-else:
-    from http.client import HTTPException
-
 import numpy as np
 
 import obspy


ThomasLecocq:
looks good, I added the test_network tag, it should run the network-related tests (dunno how to trigger them except by editing a space somewhere)?

ThomasLecocq:
edit, I manually re-triggered the action

ThomasLecocq:
I think there are a number of other mentions of e.g. the HTTP iris stuff (fdsn client, evalresp, etc) and maybe in the doc too ? Did you check (I can do it later this afternoon if needed)

https://docs.obspy.org/search.html?q=http%3A%2F%2Fservice.iris.edu&check_keywords=yes&area=default

filefolder:
no, i only tended to the ones causing errors so there are some likely stragglers. I can do a full sed on them tomorrow.

ThomasLecocq:
thanks !

filefolder:
nuclear option seems to have broken something else unforeseen! who would have guessed! will investigate

ThomasLecocq:
looks like the earthworm stuff is not related (a change in the network codes available) - only the iris credential thingy?

It's not your changes that broke the tests, it's "new" tests compared to the initial set: e.g. https://tests.obspy.org/146601/

filefolder:
Fixed these issues with master:

- EW issue in `test_get_waveform_no_cleanup`, the example data request (which is always ~1 day before now, and it's only cached on the server for a relatively short period) no longer returns multiple traces. I have skipped that particular test for now until the last running ew server is turned off or someone can find a better test station (currently `AV.AKV..BHE`), whichever comes first. 

- `fdsn/test_client.test_authentication_exceptions`... I think `download_url_mock` needed to be reset and turning off `discover_services` in some of Client calls seemed to have fixed it. Also changed the test sever to GFZ for the eida token stuff since I don't believe those are in the IRIS wadl anymore.

But there were still two new ubuntu-specific network test issues...

```
FAILED clients/earthworm/__init__.py::obspy.clients.earthworm - ExceptionGroup: multiple unraisable exception warnings (3 sub-exceptions)
FAILED clients/syngine/tests/test_client.py::TestClient::test_reading_saczip_files - UserWarning: Sample spacing read from SAC file (2.500000000 when rounded to nanoseconds) was rounded of to microsecond precision (2.500000000) to avoid floating point issues when converting to sampling rate (see #3408)
```

for the EW one, I added a safety try/finally socket close in `read_wave_server_v` (which maybe was a small bug in that original function), and for the saczip test I just suppressed that particular warning. Seems like that solved those.

Now there looks like there may be one or two transient timeout issues with various servers since they are passing on my machine at home... re-running those. EDIT: all good!

ThomasLecocq:
Do you think we could get some Earthworm crew to discuss this ?

filefolder:
I am not sure who might know.. 

here's the function:

```
    @skip_on_network_error
    @pytest.mark.skip(reason='Example data AV.AKV..BHE no longer valid')
    def test_get_waveform_no_cleanup(self, ew_client, ew_stream_no_cleanup):
        """
        Tests get_waveforms method again, 1 channel no cleanup.
        """
        # example 2 -- 1 channel, no cleanup
        ew_stream = ew_stream_no_cleanup
        delta = ew_stream[0].stats.delta
        assert len(ew_stream) >= 2
        summed_length = sum(len(tr) for tr in ew_stream)
        assert summed_length in (50, 51)
        assert ew_stream[0].stats.starttime >= self.start - delta
        assert ew_stream[0].stats.starttime <= self.start + delta
        assert ew_stream[-1].stats.endtime >= self.end - delta
        assert ew_stream[-1].stats.endtime <= self.end + delta
        for trace in ew_stream:
            assert trace.stats.network == 'AV'
            assert trace.stats.station == 'AKV'
            assert trace.stats.location in ('--', '')
            assert trace.stats.channel == 'BHE'
```

the returned stream  in `ew_stream_no_cleanup` is:

```
1 Trace(s) in Stream:
AV.AKV..BHE | 2025-10-29T20:42:08.160000Z - 2025-10-29T20:42:09.160000Z | 50.0 Hz, 51 samples
```

It seems like it used to return a stream with two traces which had been seamlessly split. I don't know why that would have been or if there are other channels currently available that would have that issue (or how long they might continue to!). AK.AKV doesn't seem like a real station and may exist purely for testing (?). 

The cleanup kwarg here is not necessarily the same as in obspy.core.stream, it seems to be specific to the earthworm buffer (e.g. in `waveserver.read_wave_server_v`) . 

IMO for now the function is flagged with a warning, we should raise an issue and come back to this once we can recover or synthesize some test data to save and read directly.

filefolder:
oops, i mixed up AK and AV above

it looks like swapping to AV.ACH might do for now:

2 Trace(s) in Stream:
AV.ACH..BHZ | 2025-10-29T20:42:08.160000Z - 2025-10-29T20:42:08.880000Z | 50.0 Hz, 37 samples
AV.ACH..BHZ | 2025-10-29T20:42:08.900000Z - 2025-10-29T20:42:09.160000Z | 50.0 Hz, 14 samples

e.g.

from obspy.clients.earthworm.client import Client

start = UTCDateTime() - 24 * 3600 * 4
end = start+1
ew_client = Client("pubavo1.wr.usgs.gov", 16022, timeout=30.0)
st = ew_client.get_waveforms(starttime=start,endtime=end,network='AV',station='ACH',,location='',channel='BHZ',cleanup=False)

It would be good to save this while it's still broken but not exactly sure how to save the buffer

filefolder:
by the time I uploaded a new version the new test data on AV.ACH stopped returning split streams. this isn't really going to be tenable. 

edit: as an aside, the NRL tests need to be remodeled as well at some point, a lot of the code still points to v1 and they randomly timeout half the time

e.g. `FAILED clients/nrl/tests/test_nrl.py::TestNRLRemote::test_nrl_type - requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))`

edit 2: the above 104 error (only on ubuntu-network for some reason) took about 10 re-tests to finally work

ThomasLecocq:
wow... those network tests are soooo unstable sometimes ... I guess we should live with it... and either SKIP or issue Warning when they fail, with an informative message that it "could" be code-related but we suspect it's server related? With a warning at least it's "tested" and infromative?

obspy/clients/earthworm/waveserver.py
```diff
@@ -258,70 +258,75 @@ def read_wave_server_v(server, port, scnl, start, end, timeout=None,
...
+            current_tb.data = bufs[0]
 
-    current_tb.ndata = len(current_tb.data)
+        current_tb.ndata = len(current_tb.data)
 
-    return tbl
+        return tbl
+    # ensure socket closes regardless
+    finally:
+        sock.close()
```
megies, line 329: git(hub) diffs sadly are super messy when indentation comes into play.. I'm assuming you added

```python
try:
    ...
finally:
    sock.close()
```

and everything in the block is just the same without changes, indented?

obspy/clients/fdsn/__init__.py
```diff
@@ -34,38 +34,42 @@
...
+RESIF       https://ws.resif.fr
 RESIFPH5    http://ph5ws.resif.fr
-SCEDC       http://service.scedc.caltech.edu
+SCEDC       https://service.scedc.caltech.edu
 TEXNET      http://rtserve.beg.utexas.edu
-UIB-NORSAR  http://eida.geo.uib.no
-USGS        http://earthquake.usgs.gov
+UIB-NORSAR  https://eida.geo.uib.no
+USGS        https://earthquake.usgs.gov
 USP         http://sismo.iag.usp.br
```
megies: Some of the ones that are left as "http" seem to also support "https".. like e.g. https://sismo.iag.usp.br/fdsnws/dataselect/1

Note that not all implementations of FDSNWS bother to show nice info pages on any or all sub-routes, like sometimes just doing "http.../fdsnws/" will show an empty page

megies:
> Now there looks like there may be one or two transient timeout issues with various servers since they are passing on my machine at home... re-running those. EDIT: all good!

Might be an automatic IP blacklist or whatever else. E.g. I got our University IP blocked by `clients.neic` test server, running obspy testsuite ages ago :rofl:

obspy/clients/earthworm/waveserver.py
filefolder: yes correct!

obspy/clients/fdsn/__init__.py
filefolder: good call on both accounts.. let me be a bit more rigorous and update where both work. I feel comfortable editorializing in this case regardless of what's at FDSN.

KAGSR is the only 404, but is still listed at FDSN at that address. In this case I am inclined to trust the FDSN listing and keep it in unless anyone can confirm. There are a lot of regional RU data centers now that I look at them...

filefolder:
re: network issues, only thing I can add is that it's only the ubuntu tests. agree not a pressing issue at the moment, just FYI you may have to manually re-run some tests for now and we may have to remove any lingering NRLv1 stuff soon.

PR #3639, opened 2025-11-20
address ubuntu-latest, py3.8 test issues

filefolder:
i think this should fix the two currently failing 3.8 tests

1) min numpy 1.20 > 1.21 to fix the mindepversion ubuntu 3.8 test
2) hack a libtiff.so.5 symlink explicitly for the ubuntu-3.8 test

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,8 +5,9 @@ Changes:
  - General
    * Fix documentation for the command line scripts (see #1208 #3569)
    * Fix code coverage reporting (see #3646)
-   * Git: update min numpy version to 1.21 and symlink libtiff.so.5
-     for ubuntu 3.8 tests
+   * Github CI: update min numpy version to 1.21 and symlink libtiff.so.5
+     for ubuntu 3.8 tests (see #3639)
+   * Minimum Dependencies: numpy 1.21 (was 1.20) (see #3639)
  - obspy.core:
    * response: allow recalculating overall sensitivity even if unit type is not
      one of VEL/ACC/DISP (see #3235)


filefolder:
i don't know what codecov's problem is. and frankly i don't want to know

ThomasLecocq (changes requested):
a tiny line in the changelog (and check in the doc if nothing need to be bumped?) and it's good to go!

filefolder:
OK, wasn't sure if git workflow hacks go in the changelog. do we know what the issue with codecov is?

ThomasLecocq:
>codecov

baaah no, it's not really talking much ... I propose we solve codecov later, it's been failing/complaining quite a lot recently, maybe some work is needed for it...

> changelog

Mjah, probably good to put a section as general where we'll update all the min dependencies before hitting the release button ?

ThomasLecocq (approved):

ThomasLecocq:
re numpy 1.21 - shouldn't we also pin it for the setup/pyproject.toml ?

filefolder:
makes sense! i think it was just in setup.py

ThomasLecocq:
what happened ?

filefolder:
ugh git got messy and i think i did something wrong. have rebuilt without conflicts

filefolder:
all righty

ThomasLecocq:
Beauty :)

ThomasLecocq:
added the test_wheels to check if all wheels build OK (py3.12 only is tested when pushed to master) -- this might require a last commit from your side to trigger (which you can do by adapting the changelog to the changes (setup)? )

filefolder:
i can re-run ALL jobs but i can't seem to just hit the wheels. is that correct?

ThomasLecocq:
> i can re-run ALL jobs but i can't seem to just hit the wheels. is that correct?

indeed, because the flag was added AFTER the commit, I think :)

filefolder:
ok i sent a dummy commit through, i guess that did it whereas running all the tests again didn't.

ThomasLecocq:
Excellent, all the build jobs pass !! Congratulations:-)

filefolder:
what's the good word on flushing this badboy

ThomasLecocq (approved):

PR #3646, opened 2025-12-02
fixing codecov uploads, splitting default or network tests reports (fixes #3645)

### LINKED ISSUE #3645: codecov reporting -2% all the time
Codecov is marking 🔴 on CI and the online report shows a -2.x% reduction .

Image

According to (clever) AIs:

>Codecov is not really “complaining” about an error here; it is just flagging a coverage comparison detail.

>Coverage data is based on HEAD 933aa53 (20 uploads) compared to BASE d2c4aa1 (25 uploads)” with the warning “Commits have different number of coverage report uploads”.
>This means the base commit (on master/main) has 25 coverage uploads aggregated into its coverage baseline, while your PR head commit only has 20. Because Codecov aggregates multiple uploads (e.g., different CI jobs, OS/Python versions), the mismatch can make the comparison slightly noisy or less precise.

>The “Change -2.70%” reflects that, with the currently uploaded reports, total coverage on the PR head is about 2.7 percentage points lower than on the base commit. That drop plus the unequal number of uploads is what triggers the little warning icon.

>In practical terms: tests passed, coverage data was uploaded, but Codecov is warning that it is comparing apples to slightly different apples (20 vs 25 reports), so the reported -2.70% change may partly come from missing or different uploads rather than real coverage loss. If you care about strict comparability, you would ensure the same set of CI jobs (and thus the same number of coverage uploads) run on both base and PR; otherwise you can mostly ignore this as an informational warning.

What are the differences in concatenated uploads ? weird

ThomasLecocq:
Fixes #3645

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,7 +4,7 @@ master
 Changes:
  - General
    * Fix documentation for the command line scripts (see #1208 #3569)
-   * Fix code coverage reporting (see #3645)
+   * Fix code coverage reporting (see #3646)
  - obspy.core:
    * response: allow recalculating overall sensitivity even if unit type is not
      one of VEL/ACC/DISP (see #3235)


ThomasLecocq:
the change worked, no uploaded artifact fails now:
image

Still have to find a way for codecov to be happy when non-network tests are reporting: currently it's complaining that there is a -2.73% change, but it's only because it's comparing pears and apples...

ThomasLecocq:
tests fail, but codecov gets all the XML reports :-)

me happy 😄 🎉 

image

filefolder (approved):
i guess you've just changed the root where the CC tests are done? seems sensible and you can't argue with results

ThomasLecocq:
> i guess you've just changed the root where the CC tests are done? seems sensible and you can't argue with results

yes indeed, and made sure the LST files are not captured by the "search" - now the files sent to codecov contain the file list and the XML, which seems to make codecov happy.

ThomasLecocq:
talking to my AI best friend:

>using codecov on github CI, how to handle the fact that some tests are skipped (e.g. the core tests are always made, but the "network" tests are only made when the PR is tagged "network_test") which leads codecov to report on LOWER coverage ?

--

Codecov has built‑in support for this use case: you should split coverage by test suite using “flags” and enable “carryforward flags” so that coverage from infrequently‑run suites (like your network tests) is reused on later commits where those suites are skipped.[1]

## Core idea

- Tag each coverage upload with a flag that identifies the suite, for example `core` and `network`.[2]
- Configure those flags in `codecov.yml` with `carryforward: true` so that when a PR only runs `core` tests, Codecov will combine the new `core` coverage with the most recent `network` coverage from a previous commit, instead of treating the network‑related lines as untested.[1]

## Minimal configuration sketch

In your GitHub Actions jobs:

```yaml
# Core tests job
- name: Run core tests
  run: pytest tests/core --cov=my_pkg --cov-report=xml:coverage-core.xml

- name: Upload coverage (core)
  uses: codecov/codecov-action@v4
  with:
    files: coverage-core.xml
    flags: core
```

```yaml
# Network tests job (only runs on PRs tagged/network)
- name: Run network tests
  run: pytest tests/network --cov=my_pkg --cov-report=xml:coverage-network.xml

- name: Upload coverage (network)
  uses: codecov/codecov-action@v4
  with:
    files: coverage-network.xml
    flags: network
```

Then in `codecov.yml` at repo root:

```yaml
flags:
  core:
    carryforward: true
  network:
    carryforward: true
```

With `carryforward: true`, if the current PR only uploads `core` coverage, Codecov will “carry forward” the latest known `network` coverage from earlier commits and compute overall project coverage from the union, so you do not see an artificial drop when `network` tests are skipped.[1]

## Additional tips

- Ensure both suites run at least once on your main branch after enabling flags, so Codecov has baseline coverage for each flag to carry forward.[1]
- If you want separate coverage gates (e.g. “core must be ≥90%, network ≥70%”), you can also define status checks per flag in `codecov.yml` while still using carryforward for total coverage.[2]

[1](https://[about.codecov](https://about.codecov.io/blog/new-product-test-only-what-you-change-with-carryforward-flags/).io/blog/new-product-test-only-what-you-change-with-carryforward-flags/)
[2](https://about.codecov.io/blog/codecov-components-breaking-down-coverage-by-filters/)
[3](https://discourse.julialang.org/t/ignoring-some-files-in-coverage-testing-when-github-actions-is-taking-care-of-continuous-integration/118847)
[4](https://[about.codecov](https://about.codecov.io/blog/codecov-components-breaking-down-coverage-by-filters/).io/blog/should-i-include-test-files-in-code-coverage-calculations/)
[5](https://club.ministryoftesting.com/t/how-to-ignore-some-tests-in-ci-with-github-actions/54167)
[6](https://community.codecov.com/t/sudden-significant-coverage-drop/1610)
[7](https://blog.fossasia.org/tag/codecov/)
[8](https://community.codecov.com/t/i-would-like-to-skip-the-codecov-when-source-code-when-files-unrelated-to-the-tests-are-modified/4433)
[9](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1473)
[10](https://codefresh.io/blog/code-coverage-codecov/)

ThomasLecocq:
OK looks like the flag has been added correctly when it's a network test:

https://github.com/obspy/obspy/actions/runs/19887261502/job/56997214465#step:10:350

> ./codecov  upload-coverage -t <redacted> --fail-on-error --git-service github --sha 1372d53fe9484c562fad23aa479321238545ffdf --disable-search --file /Users/runner/work/obspy/obspy/coverage.xml --flag unittests_network --gcov-executable gcov --name codecov-umbrella

vs default workflow:

https://github.com/obspy/obspy/actions/runs/19887261502/job/56997214426#step:10:349

> ./codecov  upload-coverage -t <redacted> --fail-on-error --git-service github --sha 1372d53fe9484c562fad23aa479321238545ffdf --disable-search --file /Users/runner/work/obspy/obspy/coverage.xml --flag unittests --gcov-executable gcov --name codecov-umbrella

which codecov seems to interpret correctly:

image


The -2.75% is because in the reference, unittests actually contain the network tests too - but when merged & created new reference set, it'll be OK

PR #3648, opened 2025-12-03
change CI to trigger when adding test_network or test_wheel labels to PRs

ThomasLecocq:
This PR modifies the classic tests & wheel CI workflow to make sure they trigger when a label is **added** to the PR, similarly as when adding `build_docs` - The current workflows will restart when `test_network` or `test_wheels` is added

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,33 +29,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Check if workflow should run - always runs quickly to provide immediate feedback
-  check_run_conditions:
-    name: check run conditions
-    runs-on: ubuntu-latest
-    outputs:
-      should_run: ${{ steps.check.outputs.should_run }}
-    steps:
-      - name: Check if workflow should run
-        id: check
-        run: |
-          if [[ "${{ github.event_name }}" == "push" ]]; then
-            echo "should_run=true" >> $GITHUB_OUTPUT
-          elif [[ "${{ github.event.action }}" != "labeled" ]]; then
-            echo "should_run=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


ThomasLecocq:
step 1: waiting for the classical tests to finish, then I'll add :

1. the network tests
2. the wheel building tests

ThomasLecocq:
ok, so classic run shows the wheels skipped & the network tests NOT done:
image

now adding the label for networks

ThomasLecocq:
adding the label indeed retriggered all normal tests (default + network):

image

now adding the wheels label

ThomasLecocq:
adding the test_wheels label re-triggered ALL jobs, including the default/network (because the condition is satisfied), but at least, this doens't require dummy commits anymore - the test_wheels, test_network work like build_docs

🥳 🎉 
image

ThomasLecocq:
@megies while in this PR, would it make sense to add a new label:

`test_network_all` : when added to a PR, tells the build matrix to run the **network** tests across the whole OS/python version matrix ? Currently, we only run the network on the latest python version of our test matrix, which is fine in general, but since Clients use sockets etc, there could be behaviour changes between python versions?

PS: dah, adding the "ready for review" label triggered all jobs ... not sure it's desired

ThomasLecocq:
adding a random label doesn't stop the current jobs, but the window at the bottom doesn't show the latest real runs - it just shows that jobs were skipped because the label isn't one of the protected ones. 

With this, it means that for example:
1. propose a PR
2. push, test, debug, push, push, push: all tests are green light
3. tag "ready for review" : the bottom window will get gray of skipped tests, and the reviewer has to go to the Checks tab:

image

ThomasLecocq:
stopping here... it's not what I wanted :'( now the tests report green on the general PR page, while of course it's only the latest skipped ones and not the complete set before that are RED ...

PR #3651, opened 2025-12-05
Add Seiscomp scml0.14 compatibility, migrate code from sc3ml to scml

filefolder:
### What does this PR do?

Adds read/write capability for the new Seiscomp 7.0.0  `0.14` schema

There was a big shift on this release from using "sc3ml" to just "scml" which required a bit of juggling. Accordingly re-wrote a lot of code references from "sc3ml" to "scml" including adding the option to use `SCML` as a format option, keeping `SC3ML` in for backwards compatibility (should add a depreciation warning though).

Refreshes the published XSL files for all versions which usually is a great improvement on reading and writing. However there was a change/bugfix in their published XSL on 26.07.2024 regarding the confidence ellipsoids were calculated, seems they were off by 1000x due to a kilometer/meter mix-up. This was fixed from 0.10 on (which I believe was SC 3 or 4?) but not in 0.9 and below, so I had to skip testing on those. Some test data also had to be updated to suit. The older "wrong" versions of the XSL remain for legacy reasons (matching what is still included in SC) but I did drop 0.6 support as it was causing additional problems and is thoroughly decrepit. Users can always dig up the XLS files and do this manually if needed.

Another feature that I think was missing is the option to save SCML to whichever version you like. By default it is 0.12 which from memory is the last sort of significant update and should be compatible with versions 5+. 

I have added a few 0.14 tests but nothing that tests any of the more esoteric event types. As the changes from 0.13 are fairly small I don't envision anything weird happening.

I think important to have out for 1.5 ...

### DIFF
--- a/obspy/io/seiscomp/core.py
+++ b/obspy/io/seiscomp/core.py
@@ -15,6 +15,7 @@ import re
 
 from lxml import etree
 
+from obspy.core.util.decorator import deprecated
 from obspy.io.quakeml.core import _xml_doc_from_anything
 
 
@@ -23,6 +24,13 @@ SCHEMA_VERSION = ['0.7', '0.8', '0.9', '0.10',
                   '0.11', '0.12', '0.13', '0.14']
 # from version 0.14 onwards "sc3ml" is dropped
 NEW_SCHEMA_VERSION = ['0.14']
+
+
+# deprecated with 1.5.0, remove a few main releases later
+@deprecated(
+    'Deprecated and will be removed next release, use _is_scml() instead')
+def _is_sc3ml(*args, **kwargs):
+    return _is_scml(*args, **kwargs)
 
 
 def _is_scml(path_or_file_object):
--- a/obspy/io/seiscomp/event.py
+++ b/obspy/io/seiscomp/event.py
@@ -16,6 +16,7 @@ import re
 
 from lxml import etree
 
+from obspy.core.util.decorator import deprecated
 from obspy.io.quakeml.core import Pickler, Unpickler, _xml_doc_from_anything
 from obspy.io.seiscomp.core import validate as validate_scml
 
@@ -32,6 +33,14 @@ SCHEMA_VERSION = ['0.7', '0.8', '0.9', '0.10',
                   '0.11', '0.12', '0.13', '0.14']
 # from version 0.14 onwards "sc3ml" is dropped
 NEW_SCHEMA_VERSION = ['0.14']
+
+
+# deprecated with 1.5.0, remove a few main releases later
+@deprecated(
+    'Deprecated and will be removed next release, use high level '
+    'read_events(.., format="SCML") or if really needed _read_scml() instead')
+def _read_sc3ml(*args, **kwargs):
+    return _read_scml(*args, **kwargs)
 
 
 def _read_scml(filename, id_prefix='smi:org.gfz.de/geofon/'):
@@ -97,6 +106,15 @@ def _read_scml(filename, id_prefix='smi:org.gfz.de/geofon/'):
                             ID_PREFIX=etree.XSLT.strparam(id_prefix))
 
     return Unpickler().load(io.BytesIO(quakeml_doc))
+
+
+# deprecated with 1.5.0, remove a few main releases later
+@deprecated(
+    'Deprecated and will be removed next release, use high level '
+    'Catalog.write(.., format="SCML") or if really needed _write_scml() '
+    'instead')
+def _write_sc3ml(*args, **kwargs):
+    return _write_scml(*args, **kwargs)
 
 
 def _write_scml(catalog, filename, validate=False, verbose=False,
--- a/obspy/io/seiscomp/inventory.py
+++ b/obspy/io/seiscomp/inventory.py
@@ -22,6 +22,7 @@ import numpy as np
 from scipy.signal import tf2zpk
 
 import obspy
+from obspy.core.util.decorator import deprecated
 from obspy.core.util.obspy_types import (ComplexWithUncertainties,
                                          FloatWithUncertaintiesAndUnit)
 from obspy.core.inventory import (Azimuth, ClockDrift, Dip,
@@ -90,6 +91,15 @@ def _parse_list_of_complex_string(complex_string):
                                                     numbers)
         raise ValueError(msg)
     return numbers
+
+
+# deprecated with 1.5.0, remove a few main releases later
+@deprecated(
+    'Deprecated and will be removed next release, use high level '
+    'read_inventory(.., format="SCML") or if really needed _read_scml() '
+    'instead')
+def _read_sc3ml(*args, **kwargs):
+    return _read_scml(*args, **kwargs)
 
 
 def _read_scml(path_or_file_object, **kwargs):
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -28,6 +28,7 @@ from pathlib import PurePath
 import numpy as np
 
 from obspy.core.util.attribdict import (AttribDict)
+from obspy.core.util.deprecation_helpers import ObsPyDeprecationWarning
 from obspy.core.util.misc import to_int_or_zero, buffered_load_entry_point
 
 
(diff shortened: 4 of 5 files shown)


ThomasLecocq:
Solid work again @filefolder ! Thanks! and indeed, needed for 1.5.0.

We should get some seiscomp guys check this PR too:  @gempa-jabe @gempa-dirk could you have a look here ? It'd be cool to have your input ! 🥳

obspy/clients/filesystem/sds.py
```diff
@@ -27,7 +27,7 @@
     LOC    :  Location identifier, up to 8 characters, no spaces
     DAY    :  3 digit day of year, padded with zeros
 
-See https://www.seiscomp3.org/wiki/doc/applications/slarchive/SDS.
+See https://www.seiscomp.de/seiscomp3/doc/applications/slarchive/SDS.html
```
gempa-dirk: A more up-to-date link is:
https://www.seiscomp.de/doc/apps/slarchive.html#slarchive

obspy/io/seiscomp/core.py
```diff
@@ -18,13 +18,16 @@
...
+                  '0.11', '0.12', '0.13', '0.14']
+# from version 0.14 onwards "sc3ml" is officially dropped
+NEW_SCHEMA_VERSION = ['0.14']
 
 
-def _is_sc3ml(path_or_file_object):
+def _is_scml(path_or_file_object):
     """
-    Simple function checking if the passed object contains a valid scxml file
+    Simple function checking if the passed object contains a valid scml file
```
gempa-dirk: SCML

gempa-dirk (changes requested):
For style sheet conversion you may simply consider the files in https://github.com/SeisComP/common/blob/master/libs/xml/ in order to drop a line-by-line review.

obspy/clients/filesystem/sds.py
filefolder: thanks Dirk! these are the same XML from the master branch (as of a week ago at least), but the way it was uploaded doesn't seem to have translated well. I could do another refresh by deleting/re-adding for clarity. will fix the link also.

obspy/io/seiscomp/inventory.py
```diff
@@ -1000,7 +1000,7 @@ def _read_float_var(elem, cls, unit=False, datum=False, additional_mapping={}):
     normally ObsPy would read this directly from a tag, but with different
     tag names this is no longer possible; instead we just pass the value
     and not the tag name. We always set the unit/datum/uncertainties to None
-    because they are not provided by scml ?
+    because they are not provided by SCML ?
```
gempa-dirk: Question mark is confusing. Is this actually a question? Is so, it should be answered before providing code. See also above.

"provided by SCML" seems misleading. Either "provided by files in SCML" While, the statement is correct for unit, times in SCML do know value and uncertainty.

filefolder: that was a legacy comment... agree about the ?s, have removed. 

not knowing how accurate this still is, I would change to: `We set the unit/datum/uncertainties to None because they are not present in SCML XSD.`

ThomasLecocq:
should there be some kind of DeprecationWarning to the SC3ML -> SCML format ?

filefolder:
That was my intent but I am actually not too sure how to implement. Simple way could be (core/event/catalog.py and core/inventory/inventory.py) 

```
        if format == 'SC3ML':
            msg = "'SC3ML' will be depreciated soon, use 'SCML'"
            print(msg)
```
I am aware of `from obspy.core.util.deprecation_helpers import ObsPyDeprecationWarning` but I don't understand how or if it works.

filefolder:
OK, have rebased and added a deprecation warning (which actually manually changes format instances of `SC3ML` directly to `SCML` and removed it as an option)

```
cat = obspy.read_events('cat.scxml',format='SC3ML')
/usr/local/lib/python3.13/site-packages/obspy/core/util/decorator.py:209: ObsPyDeprecationWarning: Format 'SC3ML' is deprecated since ObsPy 1.5.0 and will be removed in a future release. Use 'SCML' instead.
```

gempa-dirk:
> For style sheet conversion you may simply consider the files in https://github.com/SeisComP/common/blob/master/libs/xml/ in order to drop a line-by-line review.

If  you have considered my previous recommendation, then I have no further comments.

filefolder:
> > For style sheet conversion you may simply consider the files in https://github.com/SeisComP/common/blob/master/libs/xml/ in order to drop a line-by-line review.
> 
> If you have considered my previous recommendation, then I have no further comments.

Thanks Dirk, yes they're the same files. I think I replied to one of the code comments.

filefolder:
also fixes #3689

megies:
Rebased on current master and fixed merge conflict to see CI

megies:
@filefolder I was confused why some commits don't show your icon as author, but I see some commits have 'filefolder@github' as email address. Maybe you did some commits through the github webpage and the address to use in that case might not be set in your account

filefolder:
thanks for the pushes. yes i didn't have my email set for the longest time... hopefully fixed as of last week.

obspy/io/seiscomp/tests/data/iris_events.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<q:quakeml xmlns:q="http://quakeml.org/xmlns/quakeml/1.2" xmlns="http://quakeml.org/xmlns/bed/1.2">
```
megies, line 2: was this file copied from another test data folder? can we not reuse them from there?

obspy/io/seiscomp/tests/test_core.py
```diff
@@ -15,32 +15,33 @@
 
 import pytest
 
-from obspy.io.seiscomp.core import _is_sc3ml, validate
+from obspy.io.seiscomp.core import _is_scml, validate
```
megies, line 18: I wonder if we should keep the old one with deprecation warning and returning result of the renamed one for one release at least, even though it is a private function.. what do you think? Would be just in case somebody uses it to check if some file is a SC(3)ML file.

Makes me think.. I don't think we have a public routine that can act as a file format detection, do we? Maybe we should have something like `def is_format(filename, format)` in `core`..

obspy/io/seiscomp/event.py
```diff
@@ -56,38 +67,45 @@ def _read_sc3ml(filename, id_prefix='smi:org.gfz.de/geofon/'):
...
-    quakeml_doc = transform(scxml_doc,
+    quakeml_doc = transform(scml_doc,
                             ID_PREFIX=etree.XSLT.strparam(id_prefix))
 
     return Unpickler().load(io.BytesIO(quakeml_doc))
 
 
-def _write_sc3ml(catalog, filename, validate=False, verbose=False,
-                 event_removal=False, **kwargs):  # @UnusedVariable
+def _write_scml(catalog, filename, validate=False, verbose=False,
```
megies, line 120: Same as above, maybe we could think of having the old one alias for the new one, but here it's less warranted I think, since there really shouldn't be a reason to use this method directly.

megies:
I went through everything and it looks good to me in general. Great job, such a big changeset here.. :+1: 

The only thing is maybe we could keep short functions with the old names issuing deprecation warnings, even though it is private functions but still to be nice if somebody uses them. And for the `_is_sc3ml()` I think if somebody wanted to check if a file is of that format, there is no public function currently for that job that can be used.

Aside from that, since tests seem good, should be good to merge :+1:

obspy/io/seiscomp/tests/data/iris_events.xml
filefolder: good question-- there are three of this file. 

`./core/tests/data/iris_events.xml` & `./io/quakeml/tests/data/iris_events.xml` are identical but out of scope for this PR (todo?)

`./io/seiscomp/tests/data/iris_events.xml` is some specific SC test version that had to be converted/reconverted/washed so has the same info but structured differently.

obspy/io/seiscomp/tests/test_core.py
filefolder: i think having this in core is probably a good idea, though I would probably just 'head' the file looking for 'quakeml' or 'gfz'. it could also return the schema version which is often useful info.

in my weakly held opinion probably stick with the `scml` version though. sc3ml is already pretty out of date

obspy/io/seiscomp/tests/data/iris_events.xml
megies: Ah well, we can just leave like this for now :+1:

obspy/io/seiscomp/tests/test_core.py
megies: Hmm, I don't really understand, I think maybe it wasn't clear what I meant but I'll make a short commit with what I meant =)

megies:
@filefolder I listed you in code owners file for `io.seiscomp`, basically we use it to keep track of who is/was working on specific parts and also you'll get pinged if a PR makes changes in that part. If you don't want that just let me know and I can remove it.

I added one commit with some deprecation helpers for the old private functions, hope I didn't screw up, let's see CI, but then I think this should be good to merge

Thanks again @filefolder for all the work in here :rocket:

ThomasLecocq:
GreenRocketGIF

PR #3653, opened 2025-12-09
core: Optimisation for `Trace`, `Stats` and `UTCDateTime`

miili:
### What does this PR do?

This PR introduces a micro-optimisation for heavily-used objects in the obspy stack.

#### UTCDateTime:
* Avoid property lookup calls in `_operate`, speeding up any comparison of `UTCDateTime` objects.
* Introducing `__slots__` for faster attribute access.
* Changed internal property getters to direct attribute access.
* Changed internal property setters to use `_set_ns` and `_set_precision`.
* Removed `__setattr__`, which in special cases issued a warning. Moved warning to dedicated method.

#### Stats:
* Faster `UTCDateTime(ns=endtime_ns)` initialisation when `endtime` is updated.
* Low-overhead `self[key] = value` assigmemts in `__setattr__`.

#### Trace:
* Avoid unnecessary header copies during `Trace` initialisation when a `Stats` header is provided.

These optimisations deliver significant performance gains when creating millions of `Trace` instances.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -22,6 +22,8 @@ Changes:
    * window 'slepian' was removed in Trace/Stream.taper() as it was removed
      from scipy with version 1.6, add new window 'dpss' that scipy promotes as
      replacement but might need other parameters passed in (see #3331)
+   * UTCDateTime: Using `__slots__`, faster comparisons and initialisation (see #3653)
+   * Trace: faster initialisation and stats updates (see #3653)
  - obspy.clients.filesystem:
    * tsindex: update syntax for SQLAlchemy 2.0 compatibility (see #3269)
    * tsindex: leap second handling was deactivated as it is not needed with


obspy/core/trace.py
```diff
@@ -202,7 +202,8 @@ def __setitem__(self, key, value):
                 timediff = 0
             else:
                 timediff = float(self.npts - 1) * delta
-            self.__dict__['endtime'] = self.starttime + timediff
+            endtime = UTCDateTime(ns=self.starttime.ns + timediff * 1e9)
+            self.__dict__['endtime'] = endtime
```
miili: This avoids a costly `UTCDateTime` initialisation through `UTCDateTime.__add__`

obspy/core/trace.py
```diff
@@ -202,7 +202,11 @@ def __setitem__(self, key, value):
                 timediff = 0
             else:
                 timediff = float(self.npts - 1) * delta
-            self.__dict__['endtime'] = self.starttime + timediff
+
+            endtime_ns = int(self.starttime.ns + timediff * 1e9)
+            current_endtime = self.__dict__.get('endtime', None)
+            if current_endtime and current_endtime.ns != endtime_ns:
+                self.__dict__['endtime'] = UTCDateTime(ns=endtime_ns)
```
miili: Again, this is costly and avoidable.

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1312,6 +1312,7 @@ def test_zero_precision_doesnt_print_dot(self):
         utc_str = str(utc)
         assert '.' not in utc_str
 
+    @pytest.mark.skip(reason="Skipping until decision on immutability.")
```
ThomasLecocq: Why? And this is not explained in the description of the PR?

obspy/core/tests/test_utcdatetime.py
```diff
@@ -1327,6 +1328,7 @@ def test_change_time_attr_raises_warning(self):
         assert 'will raise an Exception' in str(warn.message)
         assert isinstance(warn.message, ObsPyDeprecationWarning)
 
+    @pytest.mark.skip(reason="Skipping until decision on immutability.")
```
ThomasLecocq: Why?

obspy/core/trace.py
```diff
@@ -177,32 +177,31 @@ def __setitem__(self, key, value):
...
+                delta = 0.0
             self.__dict__['delta'] = delta
             # set derived value: endtime
-            if self.npts == 0:
-                timediff = 0
-            else:
-                timediff = float(self.npts - 1) * delta
-            self.__dict__['endtime'] = self.starttime + timediff
+            timediff = 0 if self.npts == 0 else float(self.npts - 1) * delta
+            endtime_ns = self.starttime.ns + int(round(timediff * 1e9))
```
ThomasLecocq: What is the impact of rounding here, compared to before?

obspy/core/utcdatetime.py
```diff
@@ -270,21 +270,24 @@ class UTCDateTime(object):
     .. _ISO8601:2004: https://en.wikipedia.org/wiki/ISO_8601
     """
     DEFAULT_PRECISION = 6
-    _initialized = False
-    _has_warned = False  # this is a temporary, it will be removed soon
+    __slots__ = ('__ns', '__precision', '_initialized', '_has_warned',
```
ThomasLecocq, line 273: Could you explain the advantage of using this, in the PR description?

ThomasLecocq (changes requested):
Hi, thanks for the PR.

This looks interesting, but I'm wondering what the overall improvement is, and what "significant" means in that regard. Is it saving a few microseconds per instance?

Thanks for providing more information regarding the changes and, especially, the pytest changes linked to something related to immutability?

obspy/core/tests/test_utcdatetime.py
miili: done.

obspy/core/trace.py
miili: There is no difference. I just elevated the logic away to avoid the `UTCDateTime` overhead.

obspy/core/utcdatetime.py
miili: done.

obspy/core/tests/test_utcdatetime.py
ThomasLecocq: i don't see any information on that in the PR description

obspy/core/tests/test_utcdatetime.py
ThomasLecocq: i don't see any information on that in the PR description

obspy/core/utcdatetime.py
ThomasLecocq: >__slots__ has a mixed reputation in the Python community. On the one hand, they are considered to be popular. Lots of people like using them. Others say that they are badly understood, tricky to get right, and don't have much of an effect unless there are many instances of objects that use them. This article will explain what they are, how, and why to use them, and when not to use them.
https://wiki.python.org/moin/UsingSlots

What's our position on that ? (= for me, it was the first time seeing this AND had to search for the info)

obspy/core/tests/test_utcdatetime.py
miili: oopsi, re-enabled the warning and tests but moved logic from `__setattr__` to a dedicated method.

obspy/core/tests/test_utcdatetime.py
miili: re-enabled the warning and test

obspy/core/utcdatetime.py
miili: Slots reduce the object's size and enable faster attribute access. `UTCDateTime` is an atomic object, so this is absolutely feasible; it has two attributes: `ns` and `precision`.

> The short answer is slots are more efficient in terms of memory space and speed of access

Slots are also used in the Python standard library for date and datetime
https://github.com/python/cpython/blob/main/Lib/_pydatetime.py#L988

miili:
@ThomasLecocq, with millions of traces, we eliminate quite some overhead. In SeisBench, we see a ~20% performance increase when annotating 3000 channels of five-minute length.

```
SeisBench annotating 3000 channels on GPU
Old Average time 6.2663388
New Average time 5.3470239
```

This is one profiling example of `UTCDateTime.__init__` showing the performance improvement. However, direct attribute access for all comparison methods that call `_operate` likely yields the most significant benefit, in combination with `__slots__`.

image

All tests pass.

ThomasLecocq:
I've removed the tag ready for review since you're still batch (force? why?) pushing changes. Also, please address my comments/questions above.

miili:
I have changed `UTCDateTime`'s signature to use keyword arguments instead of opaque `**kwargs`. The initialisation of `UTCDateTime` is very hot: at least 4x per trace creation and also during comparisons through `UTCDateTime._operation`. The exploding initialisations and excessive `kwargs.pop` calls can cause a significant overhead.

New benchmark
```
SeisBench annotating 3000 channels on GPU
Old Average time 6.2663388
New Average time 4.7455874
```

All tests pass, there is no change to the user-exposed API.

What are your open questions? I amended and forced onto this branch to keep the commit log small.
Do you want me to materialize all used `**kwargs`?

ThomasLecocq:
Thanks for the explanation. I guess my question:
> Thanks for providing more information regarding the changes and, especially, the pytest changes linked to something related to immutability?

is no longer relevant since you removed the changes you made to the testing suite.

miili:
@ThomasLecocq, what is missing to merge this PR? Thanks for your review!

ThomasLecocq:
Hi,
aside from "time" for us to review, you can already add your name to the contributors (if you want) and a line in the changelog referring to this PR. Thanks.

miili:
Done.

ThomasLecocq:
Hey @miili , just to explain why I moved back the PR into "Needs Discussion/Time" : this PR contains "new" ways (compared to the code base), the ``__slots__`` are certainly nice for UTCDateTime, but it also comes with limitations, e.g. not allowing to add a parameter to UTCDateTime on the fly. We don't do it in obspy, but we need to make sure it's not something users would have done in the past (for whatever reason) , as this could break their code. Added to this, when "new" ways are introduced (getters, setters, etc), it also means *we* will have to maintain them for the future, so better we (I, in this very case, didn't even know slots existed until your PR...) understand it thoroughly.

miili:
Hi @ThomasLecocq , I see your concerns. What you sketch would be a heavy abuse of the object which shouldn't be encouraged nor can be expected to be supported.

I hope this article gives an interesting perspective.

https://refactoring.guru/smells/speculative-generality 

`UTCDatetime` is very hot code which should be as lean as possible. Most importantly the internal getters calls have to be removed.

ThomasLecocq:
interesting article (and website in general, didn't know it!)

megies:
In general, I don't see any immediate reason why we can not incorporate changes like this.

Out of curiosity, what kind of scenario is that with 20% time reduction you mentioned? Is that an actual real world workflow that includes reading actual data (which presumably would still take up the larger chunk of CPU time)? Or is that more of a toy example of generating synthetic data sets?

If I understand correctly, the main (and only?) implication for users is that no arbitrary attributes can be set on `UTCDateTime` anymore? I think we can live with that, especially given that since v1.2.0 we seem to already have shown a warning that this may/will not be possible anymore in the future (do not remember for what exact reason that warning was added).

There is some cosmetics/refactoring in here as well, I think it would've been nice to keep that separate/out to make review easier, especially since this PR is touching the very core of obspy and thus needs quite thorough review, even though we should be able to rely on the quite extensive tests touching core functionality.

I'll make a few more detailed/targeted comments soon.

miili:
Hey @megies , that 20% speedup is measured in a real-world scenario: SeisBench annotating like 24h of waveforms.

The main speedup stems from replacing the internal property getters with the actual slotted attributes. I'd have to look into the profiling again though.

Test coverage of the core library is good and extensive. All tests still pass without any modifications.

ThomasLecocq:
Hello @miili , 

I checked back on this and came up with a plan in order to avoid potential issues with older UTCDateTime objects being unpickled into the new structure, missing the slots:

- [ ] in current master (via a clean PR): add UTCDateTime pickiling & unpickling tests (and save a few test data pickles, make sure all test pass
- [ ] merge master into you PR, see if all test pass (I believe as such, they wont; the new UTCDateTime internals will not be satisfied)
- [ ] provide a backward compatibility when loading pre-1.5.0 UTCDateTime pickles
- [ ] check that all tests pass :-)

Copilot suggests something like this for the backward compat:

```python
# In obspy/core/utcdatetime.py

class UTCDateTime:
    # ... existing code ...
    
    def __setstate__(self, state):
        """
        Handle unpickling of old UTCDateTime objects.
        
        This ensures that pickles created before this PR (which may have
        different internal structure) can still be loaded.
        """
        # Handle old pickles that may have different attributes
        if '_ns' not in state and '__ns' not in state:
            # Very old format: try to reconstruct from timestamp
            if 'timestamp' in state:
                self._from_timestamp(state['timestamp'])
            else:
                raise ValueError("Cannot reconstruct UTCDateTime from old pickle format")
        else:
            # Ensure critical attributes are set
            self.__dict__.update(state)
        
        # Ensure precision is set (may not exist in very old pickles)
        if 'precision' not in self.__dict__:
            self.precision = self.DEFAULT_PRECISION
        
        # Ensure initialization flag is set
        if '_initialized' not in self.__dict__:
            self._initialized = True
```

ThomasLecocq:
@miili could you merge master into this PR to check if the UTCDateTime tests pass properly?

miili:
Will do, hope I find some time next week. Thanks for your review and input, Thomas. I will add the pickle compat tests.

ThomasLecocq:
Great thanks. Master contains some changes of the private variables names etc, hopefully not too much of a pain to merge. I'll give it a go if I have time too ! Thanks!

ThomasLecocq:
superseded by https://github.com/obspy/obspy/pull/3695

PR #3661, opened 2025-12-17
Fix io.win 24bits samples (fixes #3652)

### LINKED ISSUE #3652: obspy.io.win: Incorrect reading of 24-bit data due to operator precedence in _read_win
### Avoid duplicates

- [x] I searched existing issues

###  Description
I encountered an issue where the `read()` function in `obspy.io.win` incorrectly reads WIN formatted data sampled in 24-bit.

The restored waveform data shows anomalous behavior and incorrect values. I have attached a plot below demonstrating the incorrectly restored data.

Image

This is the sample file used to produce the image. (sorry for large file)
[25112618.zip](https://github.com/user-attachments/files/24027455/25112618.zip)

###  Root Cause Analysis
I investigated the source code and identified the issue in `obspy.io.win.core._read_win()`.
The problem lies in lines 161-163 (depending on the version):

```python
idata2 = output[chanum][-1] +\
    from_buffer(sdata[3 * i:3 * (i + 1)] + b' ',
                '>i')[0] >> 8
```
This code is intended to calculate the new value by adding the differential value (` from_buffer(sdata[3 * i:3 * (i + 1)] + b' ','>i')[0] >> 8`) to the previous value `(output[chanum][-1])`. T

However, in Python, the addition operator + has higher precedence than the bitwise shift operator >> (Reference: 6.17. Operator precedence https://docs.python.org/3/reference/expressions.html)
Therefore, the current code is evaluated as
```python
(output[chanum][-1] + from_buffer(...)[0]) >> 8
```

### suggested Fix
Explicitly fix the priority of the operations by adding parentheses to ensure the bitshift occurs before the addition.
```python


idata2 = output[chanum][-1] +\
    (from_buffer(sdata[3 * i:3 * (i + 1)] + b' ',
                 '>i')[0] >> 8)
```

### Code to Reproduce

```python
from obspy import read

# Path to the sample file causing the issue
# Please note: The attached sample file might be large.
filename = r"path/to/your/25112618" 

st = read(filename, format="WIN")

# Plot the waveform
t = st[0].times()
d = st[0].data

plt.plot(t, d)
plt.xlim(0, 100)
plt.show()
```

### Error Traceback

```Python traceback
No Error Traceback
```

### ObsPy Version?

1.4.2

### Operating System?

Windows

### Python Version?

3.13.9

### Installation Method?

pip

ThomasLecocq:
Corrects the reading of 24 bits WIN files - Fixes #3652)

### PR Checklist
- [ ] All tests still pass.

### DIFF
--- a/obspy/io/win/tests/test_core.py
+++ b/obspy/io/win/tests/test_core.py
@@ -90,7 +90,7 @@ class TestCore():
         assert st[0].stats.sampling_rate == 200.0
 
         expected_stds = {
-            "...0000": 62491.2874698,}
+            "...0000": 62491.2874698}
 
         for tr in st:
             assert round(expected_stds[tr.id], 4) == round(tr.data.std(), 4)


QuLogic:
Do we need all 30K for the test file? I don't see why we need all 6 channels at least. I feel like this could fit in 1K or less.

ThomasLecocq:
>only 1 channel : Agree

@Subaru617 could you provide that only 1 chan (and maybe a little shorter too) ?

Regarding the dicussion in #3652 I have the feeling the trace starts with 1 ZERO and then goes to the sample, is that correct?

ThomasLecocq:
@Subaru617 could you check here ?

ThomasLecocq:
I used AI to cut the binary file... this world is crazy.

Issue #3649, opened 2025-12-03
ValueError: Cannot use capturing groups in re.Scanner under Python 3.13.10 and 3.14.1 (seismic_phase tokenizer)


k-katfish:
### Avoid duplicates

- [x] I searched existing issues

### Problem

When running ObsPy with Python 3.13, importing obspy.taup fails with:

```python
ValueError: Cannot use capturing groups in re.Scanner
```

This originates from obspy/taup/seismic_phase.py where the tokenizer (used by leg_puller to "Tokenize a phase name into legs") uses a regex pattern containing capturing groups, which Python 3.13 no longer allows in re.Scanner.

The offending pattern:

```python
tokenizer = re.Scanner([
    # Surface wave phase velocity "phases"
    (r"\.?\d+\.?\d*kmps", self_tokenizer),
    # Composite legs.
    (r"Pn|Sn|Pg|Sg|Pb|Sb|Pdiff|Sdiff|Kdiff|Ped|Sed", self_tokenizer),
    # Reflections.
    (r"([\^v])([mci]|\.?\d+\.?\d*)", self_tokenizer),    # <-- Broken line
    # Invalid phases.
    (r"[PS][ps]", wrong_phase),
    # Single legs.
    (r"[KkIiJmcPpSs]", self_tokenizer),
    # Single numerical value
    (r"\.?\d+\.?\d*", self_tokenizer)
])
```

**Regex capture groups `()`:** Python 3.13 changed re.Scanner to forbid all capturing groups, previous versions of Python silently permitted these.

### Proposed Fix

Replace capturing groups (()) with a non-capturing group ((?:...)) and flatten the leading character class.

New version:

```python
(r"[\^v](?:[mci]|\.?\d+\.?\d*)", self_tokenizer),
```

Functionally identical, avoids capturing groups, compatible across supported Python versions.

### Code to Reproduce

```python
from obspy.taup import TauPyModel
```

### Error Traceback

```Python traceback
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    from obspy.taup import TauPyModel
  File "/usr/local/lib/python3.13/site-packages/obspy/taup/__init__.py", line 449, in <module>
    from .tau import TauPyModel  # NOQA
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/obspy/taup/tau.py", line 18, in <module>
    from .taup_path import TauPPath
[...]

PR #3662, opened 2025-12-17
Python 3.13 & 3.14 - futureproof signal.polarization for ODR - complex numbers - codec removal

ThomasLecocq:
### What does this PR do?

It tests the default tests on py 3.13 & 3.14 , and the network tests on 3.14 (for now)

Let's use this temporarily to check future problems/incompatibilities.

Linked to #3649 for example.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/tests/test_client.py
+++ b/obspy/clients/fdsn/tests/test_client.py
@@ -656,19 +656,17 @@ class TestClient():
                 ("IU", "ANMO", "*", "HHZ",
                  UTCDateTime("2010-03-25T00:00:00"),
                  UTCDateTime("2010-03-25T00:00:08")))
-        # As of 03 December 2018, it looks like EARTHSCOPE is ignoring
-        # minimumlength?
-        params = dict(quality="B", longestonly=False, minimumlength=5)
+        # removed the params (min length etc, the test will fail)
         for client in clients:
             # test output to stream
-            got = client.get_waveforms_bulk(bulk, **params)
+            got = client.get_waveforms_bulk(bulk)
             # Remove fdsnws URL as it is not in the data from the disc.
             for tr in got:
                 del tr.stats._fdsnws_dataselect_url
             assert got == expected, failmsg(got, expected)
             # test output to file
             with NamedTemporaryFile() as tf:
-                client.get_waveforms_bulk(bulk, filename=tf.name, **params)
+                client.get_waveforms_bulk(bulk, filename=tf.name)
                 got = read(tf.name)
             assert got == expected, failmsg(got, expected)
         # test cases for providing a request string


ThomasLecocq:
@filefolder your new sandbox :-)

filefolder:
quick notes on 3.14 since 3.13 seems fine. both look perfectly usable outside of the tests

1) `complex(0, 0j)` needs to be `complex(0, 0)` (both give 0j)

(which is the source of the issue in `test_reconstructing_stage_0_from_other_blockettes`)

2) this seems like an excessive volume of warnings / needs syntax update
```
clients/nrl/tests/test_nrl.py: 1072 warnings
  /home/pickle/.local/lib/python3.14/site-packages/obspy/clients/nrl/client.py:562: DeprecationWarning: codecs.open() is deprecated. Use open() instead.
    with codecs.open(path, mode='r', encoding='UTF-8') as f:
```

3) this function in `test_mseed_special_issues` seems to be causing a stink re: signal.SIGKILL for reasons not especially clear

```
    def test_infinite_loop(self, testdata):
        """
        Tests that libmseed doesn't enter an infinite loop on buggy files.
        """
        filename = testdata['infinite-loop.mseed']

        process = multiprocessing.Process(target=_test_function,
                                          args=(filename, ))
        process.start()
        process.join(60)

        fail = process.is_alive()
        process.terminate()
        if process.is_alive():
            if platform.system() == 'Windows':
                os.kill(process.pid, signal.CTRL_BREAK_EVENT)
            else:
                os.kill(process.pid, signal.SIGKILL)
        assert not fail

```
```
  File "/opt/cluster/software/python3/3.14/lib/python3.14/multiprocessing/forkserver.py", line 16, in <module>
    from . import resource_tracker
  File "/opt/cluster/software/python3/3.14/lib/python3.14/multiprocessing/resource_tracker.py", line 30, in <module>
    _IGNORED_SIGNALS = (signal.SIGINT, signal.SIGTERM)
                        ^^^^^^^^^^^^^
AttributeError: module 'signal' has no attribute 'SIGINT' (consider renaming '/home/pickle/.local/lib/python3.14/site-packages/obspy/signal/__init__.py' since it has the same name as the standard library module named 'signal' and prevents importing that standard library module)
```

filefolder:
some network timeouts and the usual codecov headache but that should do it. for last issue there was a bit of an overlap between python's `signal` and obspy's `signal` but workaround works

filefolder:
note that ubuntu-latest py3.12 + will now fail due to #3676

ThomasLecocq:
@filefolder : now that I'm preparing 1.5.0, just noticing we need to check/get your changes here into master (and make sure they pass on 3.8+ too :-) )

filefolder:
Held up by that scipy.odr thing. I think for expediency probably best to just hide the warnings for now and figure out a proper solution for 1.6.

ThomasLecocq:
I think we can afford to put the import inside the method, and if it fails for both the old and new package name, raise ?

filefolder:
I guess what I am trying to avoid for the moment is adding a new dependency since there are probably angles to this I am not considering and time is short.

Should we simply _advise_ users install the package themselves if they're using scipy >= 0.19?

ThomasLecocq:
Right, yes, I think that was the idea of @megies  ; but on the long run we can still check for scipy or the other package, without making it a fixed dependency

filefolder:
gave it a shot, was a bit more complex than I was planning on naturally

ThomasLecocq:
I'm re-adding the 3.8-3.12 tests too, so we can check if all the proposed changes are safe for ALL versions

obspy/signal/polarization.py
```diff
@@ -328,6 +327,28 @@ def particle_motion_odr(stream, noise_thres=0):
     :type noise_thres: float
     :returns: azimuth, incidence, error of azimuth, error of incidence
     """
+
+    # scipy dropping ODR in 1.19, new package not a perfect replacment
+    from importlib.metadata import version
```
ThomasLecocq: maybe opening a can of worms here: but should we make a beauuuuuutiful big PR on using `packaging` and `packaging.Version` for version comparisons ?

ThomasLecocq: @QuLogic you suggested that back in the importlib PR

ThomasLecocq:
congratulations @filefolder , looks like most of the tests pass :-)

obspy/signal/polarization.py
filefolder: packaging is probably best practice, in this case i figured importlib had a bit less overhead and we can safely assume the scipy versioning structure. happy to change it

ThomasLecocq:
I merged the github action from master into this, the test matrix is going to be huge, but at least we'll know if we can merge this into master and move on! good night!

d-chambers:
Hey @ThomasLecocq, for multiple reasons, not the least that it will be kinder to the CI, shouldn't we drop the python versions that are passed EoL (eg 3.8 and 3.9, see [here](https://devguide.python.org/versions/))

ThomasLecocq:
Agree and don't want to agree. This version is (surprisingly) very compatible... So refusing older version might sound rude. But allowing them might sound like an encouragement to keep going... And force us to bugfix/backport? No. I feel we can release 1.5.0 with max tolerance, then plan 1.6.0 fast with reduced tolerance (3 or 4 max versions) ... We need to be AND tolerant AND pragmatic with the releases and the concept of long term versions. If we release minor/major versions often, we can manage.

filefolder:
i'm very happy we managed to keep 3.8 in.. and it wasn't too difficult. maybe not intuitive but there are a lot of old data servers out there (and at least a few in our building) that are still very much in use. 

if possible would be great to even keep <3.10 for 1.6 but just drop testing/official support. anyhoo..

obspy/signal/polarization.py
filefolder: i just changed it to packaging.Version to be done with it. plus i misspelled replacement and it was bugging me

claudiodsf: A new typo? `ccipy` 😜

filefolder: ay yi yi

ThomasLecocq:
is `packaging` a default builtin py package ? or needs to be added to the requirements ?

claudiodsf:
> is `packaging` a default builtin py package ? or needs to be added to the requirements ?

It should be added as dependency: https://packaging.pypa.io/

claudiodsf:
> > is `packaging` a default builtin py package ? or needs to be added to the requirements ?
> 
> It should be added as dependency: https://packaging.pypa.io/

But, in practice, it's always installed for us, since it's a dependency of Matplotlib: https://github.com/matplotlib/matplotlib/blob/main/pyproject.toml

obspy/io/mseed/tests/test_mseed_special_issues.py
ThomasLecocq, line 1: this is the only scary/complex change (to me) - Not sure it's desirable to force spawn/fork ; this should be handled at the OS level, not within the code?

obspy/signal/polarization.py
ThomasLecocq: typo still pending

obspy/signal/polarization.py
```diff
@@ -340,28 +362,46 @@ def particle_motion_odr(stream, noise_thres=0):
...
     r = np.sqrt(n ** 2 + e ** 2)
 
-    data = scipy.odr.Data(r, abs(z))
-    model = scipy.odr.Model(fit_func)
-    odr = scipy.odr.ODR(data, model, beta0=[1.0])
-    out = odr.run()
-    in_slope = out.beta[0]
-    in_error = out.sd_beta[0]
+    def fit_func(beta, x):
+        # XXX: Eventually this is correct: return beta[0] * x + beta[1]
```
ThomasLecocq: what does this mean ?

CHANGELOG.txt
```diff
@@ -128,6 +128,8 @@ Changes:
      numpy 2.4.0 deprecation expiration. (see #3668)
    * Add warnings.catch_warnings for spectral_estimation tests, linked to
      a new numpy 2.4.0 deprecation. (see #3668)
+   * Prep SciPy dropping ODR for polarization.partical_motion_odr, prompt user
```
ThomasLecocq, line 157: reference to 3662 too ?

obspy/core/util/obspy_types.py
ThomasLecocq, line 1: I don't understand what this does? There is no check for py>=3.14 , so it's always applied? 

Side note: is this code part covered with tests?

obspy/signal/polarization.py
filefolder: beats me, that was in there before

obspy/io/mseed/tests/test_mseed_special_issues.py
filefolder: open to suggestions

ThomasLecocq: I mean , it's just a test case, it doesn't really matter - If it were in the main code and calls, I'd be more scared. Doesn't really matter here.

obspy/core/util/obspy_types.py
filefolder: tbh i don't know if this is tested anywhere explicitly, but there were certainly a lot of failures in 3.14 before that catch. 

the catch doesn't need a version check, just a safety catch for 3.14 explicitly. the comment is too vague, i will expand on it

obspy/signal/polarization.py
```diff
@@ -328,6 +327,29 @@ def particle_motion_odr(stream, noise_thres=0):
     :type noise_thres: float
     :returns: azimuth, incidence, error of azimuth, error of incidence
     """
+
+    # scipy removing ODR in 1.19.0, new package not a drop-in replacement
+    from importlib.metadata import version
+    from packaging.version import Version
+    scipy_ver = Version(version("scipy"))
```
megies: we have helpers for the main deps: `from obspy.core.util.base import SCIPY_VERSION`

megies: these are simple lists of three items '[major, minor, bugfix version]' and I think all 'post' version stuff is stripped, which should work fine for most comparisons and make them easy

ThomasLecocq: on it

filefolder: guess you're already seeing it but:

`FAILED signal/tests/test_polarization.py::TestPolarization::test_polarization_pm - TypeError: '<' not supported between instances of 'list' and 'tuple'`

ThomasLecocq:
Happy with this one :-)

.github/workflows/build_wheels.yml
```diff
@@ -30,18 +30,13 @@ jobs:
...
         exclude:
-          - build: cp38
-            exclude-flag: true
-          - build: cp39
-            exclude-flag: true
-          - build: cp310
-            exclude-flag: true
           - build: cp311
             exclude-flag: true
+          # Need to repeat for 312, 313, 314?
```
ThomasLecocq: @megies check here

megies: If I understand correctly, these excludes apply to the above mentioned exclude-flag, meaning that if it's a push to master, it will only test building wheels for the latest python (which is not listed in these "excludes"), see comment above under "strategy".
But I'm just assuming that's what's going on.

PR #3665, opened 2025-12-20
FIX: improper y-limit in Stream/Trace plot when 0-value trace exists

Dengda98:
### What does this PR do?

Hi ObsPy !

To explain this bug, see the tiny script below,

``` python
from obspy import *
st = read()
tr2 = st[0].copy()
tr2.data[:] = 0.0    # all zero
tr2.stats.channel += '2'  # just for distinguish
for tr in st:
    tr.data[:] *= 1e-8
st += Stream(tr2)
st.plot()
```
image

In this figure, the y-limit of all axes are set to [-0.05, 0.05], which is probably the default margin used by matplotlib when plotting all-zero data. **If the range of other traces is smaller than [-0.05, 0.05]**, an improper y-limit will be set in the function `WaveformPlotting.__plot_set_y_ticks()`.

To solve this, I just add if-statements to set an extremely small y-limit for this specific issure. This issue might be extremely unlikely to occur in real data, but I encountered it during my synthetic seismogram research, so I submitted this PR.

### PR Checklist

### DIFF
--- a/obspy/imaging/waveform.py
+++ b/obspy/imaging/waveform.py
@@ -710,8 +710,13 @@ class WaveformPlotting(object):
                             date2num(trace.stats.starttime.datetime))
             ax.plot(x_values, trace.data, color=self.color,
                     linewidth=self.linewidth, linestyle=self.linestyle)
-            if np.all(trace.data == 0.0):
-                ax.set_ylim([-1e-200, 1e-200])
+            
+            ymin = np.min(trace.data)
+            if ymin == np.max(trace.data):
+                if ymin == 0.0:
+                    ax.set_ylim([-1e-200, 1e-200])
+                else:
+                    ax.set_ylim((1 + np.array([-1e-15, 1e-15])) * ymin)
             
         # Write to self.ids
         trace = st[0]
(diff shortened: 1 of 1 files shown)


ThomasLecocq:
What happens when you plot equal_scale=False to the plot command?

Dengda98:
@ThomasLecocq 

The plot with equal_scale=False looks fine, probably the default ylim issure only affect this part.

https://github.com/obspy/obspy/blob/1b4da279413ab66e13dceb3224df6f6a53f1a354/obspy/imaging/waveform.py#L822-L830

ThomasLecocq:
@Dengda98 didn't forget you :) Just trying to get my ideas together on how to approach this in the most generalized way (your idea is very valid, trying to think if it could happen e.g. if not "all zeros", but "all the same value", for example, etc?)

Dengda98:
@ThomasLecocq Thanks! 
I test "all the same value", the issue also exists. An improper y-range will cover the normal y-range.
I commit a new solution, but haven't found a generalized or elegant way.

---------
**Bug reproduce** (same value)
``` python
from obspy import *
st = read()
tr2 = st[0].copy()
tr2.data[:] = 1e8    # all the same value, and large enough to cover the normal data
tr2.stats.channel += '2'  # just for distinguish
# for tr in st:
#     tr.data[:] *= 1e-8
st += Stream(tr2)
st.plot()
```

image

Issue #3643, opened 2025-12-02
read_inventory() doesn't work with byte string


waytolive:
### Avoid duplicates

- [x] I searched existing issues

### Bug Summary

if type(pathname_or_url) is byte then in function _generic_reader()
then the condition `if not isinstance(pathname_or_url, str)...`  which causes the error

the code will never be executed:
```
    elif isinstance(pathname_or_url, bytes) and \
            pathname_or_url.strip().startswith(b'<'):
        # XML string
        return callback_func(io.BytesIO(pathname_or_url), **kwargs)
```

### Code to Reproduce

```python

```

### Error Traceback

```Python traceback

```

### ObsPy Version?

1.4.2

### Operating System?

Windows 11

### Python Version?

3.8

### Installation Method?

pip

ThomasLecocq:
Please provide a complete example to reproduce the error

filefolder:
here's the function in q:

```python
def _generic_reader(pathname_or_url=None, callback_func=None,
                    **kwargs):
    # convert pathlib.Path objects to str for compatibility.
    if isinstance(pathname_or_url, PurePath):
        pathname_or_url = str(pathname_or_url)
    if not isinstance(pathname_or_url, str):
        # not a string - we assume a file-like object
        try:
            # first try reading directly
            generic = callback_func(pathname_or_url, **kwargs)
        except TypeError:
            # if this fails, create a temporary file which is read directly
            # from the file system
            pathname_or_url.seek(0)
            with NamedTemporaryFile() as fh:
                fh.write(pathname_or_url.read())
                generic = callback_func(fh.name, **kwargs)
        return generic
    elif isinstance(pathname_or_url, bytes) and \
            pathname_or_url.strip().startswith(b'<'):
        # XML string
        return callback_func(io.BytesIO(pathname_or_url), **kwargs)
    elif "://" in pathname_or_url[:10]:
        # URL
        # extract extension if any
        suffix = Path(Path(pathname_or_url).name).suffix
        if suffix == '':
            suffix = ".tmp"
        with NamedTemporaryFile(suffix=sanitize_filename(suffix)) as fh:
            download_to_file(url=pathname_or_url, filename_or_buffer=fh)
            generic = callback_func(fh.name, **kwargs)
        return generic
    else:
        pathname = pathname_or_url
        # File name(s)
        pathnames = sorted(glob.glob(pathname))
        if not pathnames:
            # try to give more specific information why the stream is empty
[...]

PR #3671, opened 2025-12-24
fix generic_reader (#3643), attempt fix at 3.8 mindepver test

filefolder:
### What does this PR do?

`(core/util/base) _generic_reader` logic was a bit shaky. reordered a bit and also allowed ANY byte object to be read since same as bytestring.

also using this as a testing ground for the ubuntu 3.8 mindepversion test. constraining cartopy and pyproj to older versions. 

fixes issue #3643

### DIFF
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -714,8 +714,9 @@ def _generic_reader(pathname_or_url=None, callback_func=None,
         pathname_or_url = str(pathname_or_url)
     if not isinstance(pathname_or_url, str):
         # first check if bytes
-        if isinstance(pathname_or_url, bytes):
-            # probably XML string, but handle any type bytes object
+        if isinstance(pathname_or_url, bytes) and \
+                pathname_or_url.strip().startswith(b'<'):
+            # XML string
             return callback_func(io.BytesIO(pathname_or_url), **kwargs)
         # not a string OR bytes- we assume a file-like object
         try:


filefolder:
- generic_reader fix: I think it's OK to drop the filter for just byte strings and just apply to all bytes in case people want to attempt something silly.

- the 3.8 mindep test seems like it's working again also.

- codecov

ThomasLecocq:
I don't really understand how the TemporaryFile part can be reached, it's when you pass an object that has a .read method, but e.g. no .seek ? Why not reading the content & passing through io.BytesIO ?

 Good catch for the mindep !

filefolder:
i think I tried getting cute and making it more simple than it could be. that first catch needs to remain specific to bytes AND xml, the original solution in the bug report:

```
def _generic_reader(pathname_or_url=None, callback_func=None,
                    **kwargs):
    # convert pathlib.Path objects to str for compatibility.
    if isinstance(pathname_or_url, PurePath):
        pathname_or_url = str(pathname_or_url)
    if not isinstance(pathname_or_url, str):
        # first check if bytes
        if isinstance(pathname_or_url, bytes) and \
                pathname_or_url.strip().startswith(b'<'):
            # XML string
            return callback_func(io.BytesIO(pathname_or_url), **kwargs)
        # not a string OR bytes- we assume a file-like object
        try:
            # first try reading directly
            generic = callback_func(pathname_or_url, **kwargs)
        except TypeError:
            # if this fails, create a temporary file which is read directly
            # from the file system
            pathname_or_url.seek(0)
            with NamedTemporaryFile() as fh:
                fh.write(pathname_or_url.read())
                generic = callback_func(fh.name, **kwargs)
        return generic
    elif "://" in pathname_or_url[:10]:
        # URL
        # extract extension if any
        suffix = Path(Path(pathname_or_url).name).suffix
        if suffix == '':
            suffix = ".tmp"
        with NamedTemporaryFile(suffix=sanitize_filename(suffix)) as fh:
            download_to_file(url=pathname_or_url, filename_or_buffer=fh)
            generic = callback_func(fh.name, **kwargs)
        return generic
    else:
        pathname = pathname_or_url
        # File name(s)
        pathnames = sorted(glob.glob(pathname))
        if not pathnames:
            # try to give more specific information why the stream is empty
            if glob.has_magic(pathname) and not glob.glob(pathname):
                raise Exception("No file matching file pattern: %s" % pathname)
            elif not glob.has_magic(pathname) and not Path(pathname).is_file():
                raise IOError(2, "No such file or directory", pathname)

        generic = callback_func(pathnames[0], **kwargs)
        if len(pathnames) > 1:
            for filename in pathnames[1:]:
                generic.extend(callback_func(filename, **kwargs))
        return generic
```

now drains from byte xml -> directly readable -> write & read. does that compute?

ThomasLecocq:
yup yup :) forgot we (I) also used the NamedTemporaryFile below. Tests look mega happy :) I'll trigger the network tests just to be sure, but normally it should not be "that network" (=aka the http:// urls are tested with core, I think).

ThomasLecocq:
@filefolder you'd need to force push something here to trigger the network tests 😇

ThomasLecocq:
actually would be good to have a test case for why all this is necessary (e.g. from OP in #3643 )

filefolder:
> actually would be good to have a test case for why all this is necessary (e.g. from OP in #3643 )

yeah I am a bit stumped 

fwiw this now works, but too fried to figure out how to create a non-xml byte object to test, or if it's worth adding a new test for this

```
 In [12]: with open('test.qml','rb') as f:
    ...:     xml_bytes = f.read()
    ...: 

In [13]: xml_bytes
Out[13]: b'<?xml version=\'1.0\' encoding=\'utf-8\'?>\n<q:quakeml xmlns:q="http://quakeml.org/xmlns/quakeml/1.2" xmlns="http://quakeml.org/xmlns/bed/1.2">\n  <eventParameters publicID="smi:org.gfz-potsdam.de/geofon/NA">\n    <event publicID="smi:ch.ethz.sed/event/historical/1165">\n      <origin publicID="smi:ch.ethz.sed/event/historical/1165/origin/1">\n        <time>\n          <value>2011-03-11T05:46:24.120000Z</value>\n        </time>\n        <latitude>\n          <value>38.297</value>\n        </latitude>\n        <longitude>\n          <value>142.373</value>\n        </longitude>\n      </origin>\n      <magnitude publicID="smi:ch.ethz.sed/magnitude/37465">\n        <mag>\n          <value>5.5</value>\n          <uncertainty>0.1</uncertainty>\n        </mag>\n        <type>MS</type>\n        <originID>smi:ch.ethz.sed/event/historical/1165/origin/1</originID>\n        <methodID>smi:ch.ethz.sed/magnitude/generic/surface_wave_magnitude</methodID>\n        <stationCount>8</stationCount>\n        <evaluationStatus>preliminary</evaluationStatus>\n        <comment id="smi:some/comment/id/muh">\n          <text>Some comment</text>\n          <creationInfo>\n            <author>EMSC</author>\n          </creationInfo>\n        </comment>\n        <comment>\n          <text>Another comment</text>\n        </comment>\n        <creationInfo>\n          <author>NEIC</author>\n        </creationInfo>\n      </magnitude>\n    </event>\n  </eventParameters>\n</q:quakeml>\n'

In [14]: hmm = obspy.read_events(xml_bytes)

In [15]: hmm
Out[15]: 
1 Event(s) in Catalog:
2011-03-11T05:46:24.120000Z | +38.297, +142.373 | 5.5  MS
```

ThomasLecocq:
that snippet is good enough for now I believe :) , as a new test (using the testdata)

filefolder:
looks like already exists! in `io/quakeml/tests`

```
    def test_read_string(self, testdata):
        """
        Test reading a QuakeML string/unicode object via read_events.
        """
        with open(testdata['neries_events.xml'], 'rb') as fp:
            data = fp.read()

        catalog = read_events(data)
        assert_no_extras(catalog)
        assert len(catalog) == 3
```

ThomasLecocq (approved):

PR #3677, opened 2026-01-26
faster packet handler for basic client

raciner:
The existing code of obspy.clients.seedlink.basic_client.py calls _packet_handler for every single record.
This will then convert, add and merge every single record, leading to memory resizing every time, resulting
in O(N²) behaviour.

The proposed change will store all miniseed records and convert them in one large batch, resulting in O(N) behaviour.
Downloading five minutes of broadband data from the CH network took 5 minutes with the existing code and takes
30 seconds with the proposed change, resulting in a reduction of about 90%.

I had to adjust obspy.io.core.mseed to make it read numpy.ndarrays directly. This is taken from my changes in github.com/obspy/obspy/pull/3622. I tried to be as minimal as possible, so that these two PRs do not depend on each other.

### What does this PR do?

*Please fill in*

*Please fill in (link relevant issues with "see #123456", mark issues that get resolved with "fixes #12345")*

### PR Checklist
- [X] If all tests including network modules (e.g. `clients.fdsn`) should be tested for the PR,
- [X] All tests still pass.
- [X] Add the yellow `ready for review` label when you are ready for the PR to be reviewed.

### DIFF
--- a/obspy/clients/seedlink/basic_client.py
+++ b/obspy/clients/seedlink/basic_client.py
@@ -11,13 +11,13 @@ SeedLink request client for ObsPy.
 import fnmatch
 import warnings
 
+import numpy as np
 from lxml import etree
 
 from obspy import Stream
 from .slclient import SLClient, SLPacket
 from .client.seedlinkconnection import SeedLinkConnection
 from obspy.io.mseed.core import _read_mseed
-import numpy as np
 
 
 class Client(object):


raciner:
I don't believe the failed unit tests to be related to the code changes I made.

ThomasLecocq:
@raciner the packets received are sequential, no overlap, no gap ?

obspy/clients/seedlink/basic_client.py
filefolder, line 1: minor but should this import numpy `as np` for consistency with the rest of the code

obspy/clients/seedlink/basic_client.py
```diff
@@ -376,6 +387,46 @@ def _packet_handler(self, count, slpack):
...
+        type_ = slpack.get_type()
+        if self.debug:
+            print(type_)
+
+        # process INFO packets here
+        if type_ == SLPacket.TYPE_SLINF:
+            if self.debug:
+                print(SLPacket.TYPE_SLINF)
+            return False
+        elif type_ == SLPacket.TYPE_SLINFT:
```
claudiodsf, line 411: You need just a simple `if` here, since the previous clause returns

raciner:
> @raciner the packets received are sequential, no overlap, no gap ?

They typically are, but this is already taken care of in the C part of the code anyway, in  io/mseed/src/obspy-readbuffer.c starting from line 518. Doing this in python in the step size of a record length (typically 512 bytes) comes with a huge penalty, as memory is allocated every single time and data is copied, while in the C code, pointers are moved around without any reallocation.

obspy/clients/seedlink/basic_client.py
raciner: I have copied this logic from the regular packet handler. Although I agree that `if` would be sufficient, I would like to leave it the same as in the original code.

obspy/clients/seedlink/basic_client.py
raciner: Fixed.

obspy/clients/seedlink/basic_client.py
claudiodsf: Ok

raciner:
Just asking, are we ready to go with this one? I just came across a colleague doing crazy workarounds to avoid using basic_client due to the performance issues, having this fixed would be very beneficial, I think.

ThomasLecocq:
Haven't yet had a look at it, we're indeed trying to push as many as possible in 1.5 😇

obspy/clients/seedlink/basic_client.py
```diff
@@ -17,6 +17,9 @@
 from .slclient import SLClient, SLPacket
 from .client.seedlinkconnection import SeedLinkConnection
 
+import obspy.io.mseed.core
```
ThomasLecocq: this isn't very standard practice in obspy, I think ? We use to from ... import _read_mseed ?

obspy/clients/seedlink/basic_client.py
```diff
@@ -179,11 +182,19 @@ def _multiselect_request(self, multiselect, starttime, endtime):
...
-        stream = self.stream
+
+        # holds all data received by the client
+        self.bulk_data_container = list()
+
+        self._slclient.run(packet_handler=self._fast_packet_handler)
+
+        # process the data all in one go, much more efficient
+        input = np.concatenate(self.bulk_data_container)
+        self.bulk_data_container = list()
```
ThomasLecocq, line 192: this line is to reset/empty the list ?

raciner: Yes, it is.

obspy/clients/seedlink/basic_client.py
megies: I think we're not fully consistent on this one across the code base, not sure though :sweat_smile:

megies:
This almost got overlooked since we didnt have it on the project page yet, but this looks good to me if CI is fine with it :+1:

megies:
Had to do some magic tricks and squash the couple commits together since it got messy with the merging in master and rebasing wasn't cleanly possible. But I think it turned out fine now, to rebase it cleanly

PR #3678, opened 2026-02-01
MassDownloader, use threads_per_client for StationXML downloads

chad-earthscope:
### What does this PR do?

Uses the existing `MassDownloader.download(threads_per_client)` value for StationXML downloads.
Previously this value is used to control miniSEED downloads.  All values defaulted to 3, but the 
StationXML download helper was fixed at 3 with no connection to the caller's desired number 
of threads.

Seems reasonable that the caller can control both the miniSEED and StationXML download concurrency.

The documentation for .download contains a statement that is very general about downloads
being controlled  by this value:
```:param threads_per_client: The number of download threads launched per client.```

So I assumed that it controlled all downloads and only happen to notice when looking at the code.

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/clients/fdsn/mass_downloader/download_helpers.py
+++ b/obspy/clients/fdsn/mass_downloader/download_helpers.py
@@ -648,7 +648,7 @@ class ClientDownloadHelper(object):
                 stationxml_storage=self.stationxml_storage,
                 logger=self.logger)
 
-    def download_stationxml(self, threads_per_client=3):
+    def download_stationxml(self, threads=3):
         """
         Actually download the StationXML files.
 
@@ -690,7 +690,7 @@ class ClientDownloadHelper(object):
 
         # Download it.
         s_time = timeit.default_timer()
-        pool = ThreadPool(min(threads_per_client, len(arguments)))
+        pool = ThreadPool(min(threads, len(arguments)))
         results = pool.map(star_download_station, arguments)
         pool.close()
         e_time = timeit.default_timer()
--- a/obspy/clients/fdsn/mass_downloader/mass_downloader.py
+++ b/obspy/clients/fdsn/mass_downloader/mass_downloader.py
@@ -213,7 +213,7 @@ class MassDownloader(object):
 
             # Download StationXML data.
             helper.prepare_stationxml_download()
-            helper.download_stationxml(threads_per_client=threads_per_client)
+            helper.download_stationxml(threads=threads_per_client)
 
             # Sanitize the downloaded things if desired. Assures that all
             # waveform data also has the corresponding station information.

### CODE REVIEW

**Comment by filefolder** [comment]
Looks fine to me, I guess assuming that servers can handle concurrent XML requests? Never considered this before, how does this work in practice?

Unfortunately I'm actually posting because I had a tangent question you in particular (or anyone else!) might have good intuition with. Does download thread concurrency really speed up the process? On our server (lo-tech non-cloud seiscomp install) it just seems to divide the bandwidth AND clog the queue, but I haven't actually tried measuring it. My feeling is that even if it does make things faster, it's also a bit rude given that there are often finite download slots, so I often wonder if the default threads_per_client should be set to 1 unless a server offers guidance otherwise or I am otherwise very mistaken.

File: CHANGELOG.txt [hunk 1]
```diff
@@ -48,6 +48,8 @@ Changes:
      getting raised as an FDSNException but some low-level exceptions like
      AttributeError, which in turn caused problems in Mass Downloader (see
      #3513)
+ - obspy.clients.fdsn.mass_downloader
+   * Use the `download(threads_per_client)` value for StationXML downloads.
```
ThomasLecocq, line 68: add PR number here to allow tracing back to the explanation

ThomasLecocq (approved):
looks good to me, and reasonable :-)

chad-earthscope:
> Does download thread concurrency really speed up the process? On our server (lo-tech non-cloud seiscomp install) it just seems to divide the bandwidth AND clog the queue, but I haven't actually tried measuring it.

The short answer is yes.  From the network perspective it is well established that multiple streams between two points can generally move data faster than a single stream.  This is the basis of bbcp or GridFTP for moving data fast.  For the server, it is usually running on thread-capable hardware/infrastructure, so if the task is cpu-bound more processes/threads obviously helps.  Of course it all depends on multiple factors of any specific use case.

> My feeling is that even if it does make things faster, it's also a bit rude given that there are often finite download slots, so I often wonder if the default threads_per_client should be set to 1 unless a server offers guidance otherwise or I am

I don't know when this code was created but it was defaulting to 3 (and remains so) and has been used for many years in this state I'm guessing.  So 1 might be a downgrade in speed. FWIW.

obspy/clients/fdsn/mass_downloader/download_helpers.py
```diff
@@ -648,7 +648,7 @@ def prepare_stationxml_download(self):
                 stationxml_storage=self.stationxml_storage,
                 logger=self.logger)
 
-    def download_stationxml(self, threads=3):
+    def download_stationxml(self, threads_per_client=3):
```
megies: I know this naming might be clearer an dbetter, but I'm not sure if the rename is worth potentially breaking old codes that might use `threads` as a kwarg by its name.
I'd rather keep the old name for that reason, even if the name might be worse..

megies: Otherwise fine and good to go

megies (changes requested):

obspy/clients/fdsn/mass_downloader/download_helpers.py
megies: Looks like that naming is coming from docs.. would have to look in detail but then usually we prefer changing the naming in the docs rather, to not break things potentially

chad-earthscope: Hmm, I thought that helper functionality was local to mass downloader.  (In fact there is another, completely different download_stationxml() in mass downloader).

I was trying to make it consistent with user supplied value.

Will revert that when I have a moment.

Issue #2026, opened 2017-12-10
core: remove "attach_response()" ??


megies:
Should we remove `attach_response()` for `1.2.0` already? See #2025

megies:
Punting to some yet-tbd future release due to valid concerns in https://github.com/obspy/obspy/issues/2025#issuecomment-350588517

megies:
This was brought up again, maybe we should opt to print a warning about potential removal in the future to gently steer people away from using this mechanism..

Main reason to do so is that abusing this mechanism (without ill intent) can lead to a very high burden of easily avoidable individual requests to data centers, that can actually lead to valid requests timing out if the amount of concurrent connections/requests gets too high.

ThomasLecocq:
+1 for the Warning !!

petrrr:
+1
I am discovering this old issue, because I was about file one. 
We discussed this with @megies on a different channel.

Apparently, some users use this feature also in the context of massive data processing and in such cases the amount of requests gets huge and repetitive. This created quite some issues within the EIDA federation this summer. 

It is possible to get more complete inventories and reuse them. There is also some support to manipulate inventories (selection parts, or combining into larger ones).  

Anyway, there would be some desire to move relatively fast on this. 

Suggested actions:

0. Review the documentation with the aim to strongly discourage usages of this feature, provide some details on the reason. 

1. Produce a deprecation warning if the feature is used. I guess this could be inserted in the next release (including minor) but would not have an immediate effect.

2. Remove the feature with the next major release? (too ambitious?)


Possible docu changes:

Remove examples with `attach_response=True`:
* https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_waveforms.html

Change the description of the parameter:

```
attach_response ([bool](https://docs.python.org/3/library/functions.html#bool)) – Specify whether the station web service should be used to automatically attach response information to each trace in the result set. A warning will be shown if a response can not be found for a channel. Does nothing if output to a file was specified.
```
* https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_waveforms.html
* https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_waveforms_bulk.html

I have not spotted any occurrences of `attach_response=True`, but maybe `metadata=True` of which I was non aware of and which might be an alias?

petrrr:
Hi @megies, @ThomasLecocq!

[...]

PR #3683, opened 2026-02-11
deprecate attach_response

d-chambers:
## Summary

  Deprecate Trace.attach_response() and Stream.attach_response() (issue #2026), while keeping existing behavior intact for backward compatibility.

  ## What changed

  - Added deprecation warnings:
      - Trace.attach_response() now emits ObsPyDeprecationWarning.
      - Stream.attach_response() now emits ObsPyDeprecationWarning.
  - Updated method docstrings to mark these APIs as deprecated and recommend passing metadata directly via inventory= to:
      - Trace.remove_response()
      - Trace.remove_sensitivity()
      - Stream.remove_response()
  - Removed public attach_response() usage examples from relevant docstrings.
  - Updated internal implementation paths to avoid relying on deprecated public APIs:
      - FDSN client response attachment now uses internal response lookup per-trace.
  - Preserved legacy warning/continue behavior in example stream creation (read() path), so no behavioral regression in tests relying on warning-based flow.
  - Added changelog note under master (obspy.core section).

Also, as mentioned [here](https://github.com/obspy/obspy/issues/3228#issuecomment-3878801383)

### DIFF
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -1753,12 +1753,9 @@ class TestTrace:
         tr.stats.location = '30'
         tr.stats.channel = 'LDO'
         tr.stats.starttime = UTC("2010-07-23T00:00:00")
-        # remove response
-        del tr.stats.response
         inv = read_inventory(
             testdata['stationxml_IU.ANTO.30.LDO.xml'], format='StationXML')
-        tr.stats.response = tr._get_response(inv)
-        tr.remove_response()
+        tr.remove_response(inventory=inv)
 
         # blockette 62, stage 1 + blockette 58, stage 2
         tr = read()[0]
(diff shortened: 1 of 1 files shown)


ThomasLecocq:
thank youuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu ! 🙏

ThomasLecocq:
the scipy ODR warnings are addressed in #3662

d-chambers:
Looks like i forgot to lint.  It really has been awhile... will fix when i get back to a computer

obspy/clients/fdsn/client.py
```diff
@@ -1051,11 +1011,12 @@ def get_waveforms_bulk(self, bulk, quality=None, minimumlength=None,
             instead of being parsed to an ObsPy object. Thus it will contain
             the raw data from the webservices.
         :type attach_response: bool
-        :param attach_response: Specify whether the station web service should
-            be used to automatically attach response information to each trace
-            in the result set. A warning will be shown if a response can not be
-            found for a channel. Does nothing if output to a file was
-            specified.
+        :param attach_response: Convenience option to automatically attach
```
ThomasLecocq: maybe add "Will be deprecated in 1.6.0"

obspy/clients/fdsn/client.py
```diff
@@ -869,11 +847,12 @@ def get_waveforms(self, network, station, location, channel, starttime,
...
```
ThomasLecocq: maybe add "Will be deprecated in 1.6.0"

obspy/core/tests/test_trace.py
```diff
@@ -1756,7 +1757,7 @@ def test_remove_polynomial_response(self, testdata):
         del tr.stats.response
         inv = read_inventory(
             testdata['stationxml_IU.ANTO.30.LDO.xml'], format='StationXML')
-        tr.attach_response(inv)
+        tr.stats.response = tr._get_response(inv)
```
megies: I'll change this to the canonical `tr.remove_response(inv)` which is doing the exact same but without using private methods and the old "magic" that we should've never added in the first place of using `tr.stats.response`

megies (approved):
Thanks @d-chambers! Good to go on green CI 👍

megies:
test fails are unrelated (the scipy ODR issue)

PR #3691, opened 2026-03-02
Speedup `import obspy` on Python 3.10+

claudiodsf:
> ⚠️  Disclaimer: I used GitHub Copilot to help diagnose the issue and to write a working solution. I then reviewed the code manually.

### What does this PR do?

Add module-level caching to reduce `entry_points()` calls from 111 to 1 during import, improving ObsPy import time by ~5x (from ~4s to ~0.74s on my machine).

On current master, import time is much slower than 1.4.2:

```bash
(master) ➜  ~ time python -c "import obspy"
python -c "import obspy"  4.04s user 1.83s system 132% cpu 4.442 total

(obspy142) ➜  ~ time python -c "import obspy"
python -c "import obspy"  0.14s user 0.07s system 55% cpu 0.379 total
```

This is related to the migration from `pkg_resources` to `importlib.metadata` (#3333). While this migration was necessary because `pkg_resources` is deprecated, it introduced a performance regression because `importlib.metadata.entry_points()` was being called 111 times during a single import.

In Python 3.10+, each call to `entry_points()` scans all installed packages in the environment, making the cumulative cost very high. With ~700 entry points installed in a typical Python 3.13 environment (mine 😉), this resulted in import times of 4+ seconds.

This PR addresses the issue by implementing module-level caching of `entry_points()`, reducing the number of calls from 111 to 1 and improving import time by ~20x (from 4.5s to 0.22s on Python 3.13 on my machine). The solution is compatible with Python 3.8-3.13, handling API differences between Python 3.8/3.9 (which return a `dict`) and Python 3.10+ (which return an `EntryPoints` object).

The new code runs much faster:

```bash
➜  ~ time python -c "import obspy"
python -c "import obspy"  1.31s user 0.15s system 371% cpu 0.392 total

# second time is even faster
➜  ~ time python -c "import obspy"
python -c "import obspy"  0.74s user 0.23s system 413% cpu 0.233 total
```

### PR Checklist
- [x] All tests still pass.

### DIFF
--- a/obspy/core/util/base.py
+++ b/obspy/core/util/base.py
@@ -378,8 +378,8 @@ def _get_ordered_entry_points(group, subgroup=None, order_list=[]):
     """
     Gets a ordered dictionary of all available plug-ins of a group or subgroup.
     """
-    # get all available entry points
-    ep_dict = _get_entry_points(group, subgroup)
+    # Copy so pop() below does not modify the lru_cache result.
+    ep_dict = _get_entry_points(group, subgroup).copy()
     # loop through official supported waveform plug-ins and add them to
     # ordered dict of entry points
     entry_points = OrderedDict()


obspy/core/util/base.py
```diff
@@ -318,17 +348,39 @@ def _get_entry_points(group, subgroup=None):
...
+                if ep.group.startswith(f'{group}.'):
+                    parent_name = ep.group[len(group)+1:]  # e.g. "MSEED"
+                    if parent_name not in subgroups:
+                        subgroups[parent_name] = []
+                    subgroups[parent_name].append(ep)
+
+            # Only include parent plugins that have the matching subgroup
             features = {}
-            for ep in eps:
-                sub_eps = tuple(importlib.metadata.entry_points(
```
claudiodsf, line 325: That's the key point: avoid calling many times `importlib.metadata.entry_points()`

claudiodsf:
Linting errors are related to other lines in `base.py` that I didn't touch in this PR.

obspy/core/util/base.py
```diff
@@ -262,6 +262,34 @@ def get_entry_point_dist_name(entry_point):
...
+
+
+def _get_all_entry_points():
+    """
+    Get all entry points, using a module-level cache to avoid repeated
+    expensive scans of all installed python packages.
+
+    Returns a flat list of all EntryPoint objects.
+    """
+    global _CACHED_ENTRY_POINTS
```
QuLogic: You don't need a global; add a `functools.(lru_)cache` decorator on this function.

claudiodsf: I didn't know this trick! Thanks for teaching me something new!

claudiodsf:
> Linting errors are related to other lines in `base.py` that I didn't touch in this PR.

Not true! Linting errors where mine 🙄 

Hopefully, fixed

filefolder:
nice work! i was wondering if it had gotten really slow or I was just getting more impatient

ThomasLecocq:
Aah ! Nice catch! I also noticed the slowdown when calling msnoise CLI (it imports obspy) :-) will review :-)

obspy/core/util/base.py
```diff
@@ -262,6 +263,22 @@ def get_entry_point_dist_name(entry_point):
...
+    """
+    eps = importlib.metadata.entry_points()
+    if not isinstance(eps, dict):
+        # Python 3.10+: convert to list
+        return list(eps)
+    # Python 3.8/3.9: flatten the dict values to a list
+    return [ep for group_eps in eps.values() for ep in group_eps]
+
+
 def _get_entry_points(group, subgroup=None):
```
d-chambers, line 283: Is it worth also caching this?

obspy/core/util/base.py
```diff
@@ -262,6 +263,22 @@ def get_entry_point_dist_name(entry_point):
     return entry_point.value.split('.')[0]
 
 
+@functools.lru_cache(maxsize=None)
```
megies, line 266: If not using size limit, maybe use `@functools.cache`? Slightly shorter and looks even better then? =)

claudiodsf: `@functools.cache` was added in Python 3.9... 😞 

https://docs.python.org/3/library/functools.html#functools.cache

obspy/core/util/base.py
claudiodsf: Done! This makes `import obspy` even slightly faster 👍

claudiodsf:
Trying again. Test were failing because of a cached dict being modified afterwards.

ThomasLecocq:
Looks good to me !

obspy/core/util/base.py
megies: Oh OK!

PR #3714, opened 2026-03-19
MSEED: memory leak fixed, msr_free not called in one case

raciner:
### What does this PR do?

I believe I found a memory leak in obspy-readbuffer.c.
```
        if (OBSPY_ISVALIDBLANK(mseed + offset)) {
            offset += MINRECLEN;
            msr_free(&msr);
            continue;
        }
```
I added msr_free(&msr), otherwise msr_free is not called in the current iteration, leading to msr not being freed.

### Links to other Issues?

No.

### AI used?

Used MS Copilot, which gave a lot of hints, most of them irrelevant. I believe this one is the one which matters in my use case.

### PR Checklist

### Issue labels

### DIFF
--- a/obspy/io/mseed/src/obspy-readbuffer.c
+++ b/obspy/io/mseed/src/obspy-readbuffer.c
@@ -331,11 +331,6 @@ readMSEEDBuffer (char *mseed, int buflen, Selections *selections, flag
 
     // Read all records and save them in a linked list.
     while (offset < buflen) {
-        msr = msr_init(NULL);
-        if ( msr == NULL ) {
-            ms_log (2, "readMSEEDBuffer(): Error initializing msr\n");
-            return NULL;
-        }
         if (verbose > 1) {
             ms_log(0, "readMSEEDBuffer(): calling msr_parse with "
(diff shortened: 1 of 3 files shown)


obspy/io/mseed/src/obspy-readbuffer.c
```diff
@@ -365,6 +365,7 @@ readMSEEDBuffer (char *mseed, int buflen, Selections *selections, flag
         // Skip empty or noise records.
         if (OBSPY_ISVALIDBLANK(mseed + offset)) {
             offset += MINRECLEN;
+            msr_free(&msr);
```
QuLogic: AFAICT, nothing uses `msr` before this. Since these are checks for validity, we should probably move the `msr_init` after this block and avoid having to alloc/free it at all before we're ready to use it.

raciner: This would make sense to me, too. As well, after looking at the code again today, I have the impression there are more such cases. I will look further into this as soon as I have the time to do so.

megies:
rebased to fix changelog conflicts

PR #3716, opened 2026-03-19
Fix #3634: add max_bulk_lines to BaseRoutingClient to prevent 413 errors

### LINKED ISSUE #3634: fdsn client / routing client improvement: better handling for virtual networks
### Problem

The routing client provides support for virtual networks (Vnet) and it is currently the only way Obspy provides to support for Vnets. It uses the routing service to expand the Vnet into a explicit list, then the list is used to request either inventory information (fdsnws/station) or waveform data (fdsnws/dataselect) to the various endpoints (nodes) via a **single** POST requests for each node. 

In particular for large Vnets (AdriaArray or AlpArray) these expanded list can become extremly large, resulting in POST requests with many lines, stations or channels to individual nodes. These requests, due to the size and number of requested streams are prone to fail or to timeout, in particular with the standard timeout.

A repoducable example for the AdriaArray Vnet "_ADARRAY":

``` python
import obspy
from obspy import UTCDateTime as UTC

rclient = RoutingClient("eida-routing", debug=True, credentials=None)

t1 = UTC("2024-01-01T00:00:00")
t2 = t1 + 60*60.0
net = "_ADARRAY"
sta = "*"
cha = "HHE,HHN,HHZ"
loc = "*"

inv = rclient.get_stations(starttime=t1, endtime=t2,
                           network=net, channel=cha, 
                           level="channel")

st = rclient.get_waveforms(network=net, station=sta,
                           channel=cha, location=loc,
                           starttime=t1, endtime=t2)

```

For "_ADARRAY" the EIDA node INGV provides the most channels and takes time to provide the complete dataset. 

### Impact on **station** service:

In the past even the large request to station service were problematic for "_ADARRAY". These large requests to station can now be handled (at node INGV) and to do not cause a problem at the time of writing for this specifc Vnet. However, for other large Vnets or setups such problems could resurface.

### Impact on **dataselect** service:

The dataselect request to various nodes will not complete within the set timeout. It is possible to get a complete reply if the timeout is set significantly larger.

With a time set to 600s (10 min) I was able to get reliable responses. 

```
rclient = RoutingClient("eida-routing", debug=True, timeout=600, credentials=None)
```

### User feedback

The your might not get a accurate feedback if any problems occured during download, eg. if single requests failed or timeout.

### Additional observation

From the debug outputs it becomes evident that `get_waveform` requests are always anticipated with a large request to station service. It is not evident why this is implement this way and these request to station might not be necessary at least not always.

### Proposed solution

The suggestion is to enhance the client with some additional features or improvements.

### User transparency

The user should be correctly informed about an problem in a way the user can understand why data received are not complete. 

There was also a suggestion to collect the processing status in a data structure with similar information as provided by the debug output. This status data structure could then have different usages, eg. it could provided a detailed report on the data which have been received or not.

### Splitting in smaller POST list

It should be possible to limit the number of lines in one single request, if this limit is execed the request would be split into multiple requests. The target should be that each request can be concluded within the set timeout. The limit should be a users configurable parameter.

A consideration would be to implement a mechanism which reacts appropriately to error conditions. In particular, the client should be able to react to status code indicating too large requests or if a timeout occurs.

Splitting requests should be possible to both station and dataselect, but the limits which are necessary for reliable functioning might be different.

### Simplify dataselect request

We should check if the two step request station then dataselect can be avoided. In principle, it should be possible to use the dataselect parameters with routing as well and then use the provided list directly.

NikolaosSokos:
Split oversized bulk requests into chunks before dispatching to the ThreadPool, avoiding "Request Entity Too Large" responses from FDSNWS endpoints. Default limit is 100 lines, matching fdsnws_fetch behaviour.

### What does this PR do?

Adds a `max_bulk_lines` parameter (default: 100) to `BaseRoutingClient.__init__`. When a routed bulk request exceeds this limit, `_download_parallel` splits it into smaller chunks before dispatching to the `ThreadPool`. This prevents HTTP 413 "Request Entity Too Large" errors and long-tail timeouts when downloading waveforms or station metadata for large virtual networks via `EIDAWSRoutingClient` or `FederatorRoutingClient`.

The default of 100 matches the behaviour of `fdsnws_fetch`.

### Links to other Issues?

fixes #3634 

### AI used?

Claude Sonnet 4.6 (Claude Code) was used to review the implementation for correctness, identify a redundant variable, and validate the unit test logic. All design decisions and code were authored by the contributor.

### PR Checklist

- [x] Correct **base branch** selected? [`master`](https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model)
- [x] **Tests**: Added `test_bulk_line_splitting` in `test_base_routing_client.py` using `unittest.mock`

Rare actions items:

- [x] First time contributors: Feel free to add your name to `CONTRIBUTORS.txt`

### Issue labels

### DIFF
--- a/obspy/CONTRIBUTORS.txt
+++ b/obspy/CONTRIBUTORS.txt
@@ -113,6 +113,7 @@ Schneider, Simon
 Shimizu, Toshiya
 Sippl, Christian
 Snoke, Arthur
+Sokos, Nikos
 Stähler, Simon C.
 Stange, Stefan
 Strutz, Dominik
--- a/obspy/clients/fdsn/routing/routing_client.py
+++ b/obspy/clients/fdsn/routing/routing_client.py
@@ -15,9 +15,7 @@ from multiprocessing.dummy import Pool as ThreadPool
 
 import decorator
 import io
-import logging
 import sys
-import time
 import traceback
 import warnings
 from urllib.parse import urlparse
@@ -29,9 +27,6 @@ from ...base import HTTPClient
 from .. import client
 from ..client import raise_on_error
 from ..header import FDSNException, URL_MAPPINGS, FDSNNoDataException
-
-
-logger = logging.getLogger(__name__)
 
 
 def RoutingClient(routing_type, *args, **kwargs):  # NOQA
@@ -303,7 +298,6 @@ class BaseRoutingClient(HTTPClient):
                     new_dl_requests.append(new_req)
             dl_requests = new_dl_requests
 
-        t0 = time.time()
         pool = ThreadPool(processes=len(dl_requests))
         results = pool.map(_try_download_bulk, dl_requests)
 
@@ -325,15 +319,6 @@ class BaseRoutingClient(HTTPClient):
         # Explitly close the thread pool as somehow this does not work
         # automatically under linux. See #2342.
         pool.close()
-
-        t1 = time.time()
-        failed_count = len([r for r in results if r is None])
-        status = "INCOMPLETE" if failed_count > 0 else "COMPLETE"
-        logger.info("RoutingClient: Download completed in %.2f seconds. "
-                    "Total chunks: %i, Success: %i, Failed: %i (%s)",
-                    t1 - t0, len(dl_requests),
-                    len([r for r in results if r is not None]),
-                    failed_count, status)
 
         return collection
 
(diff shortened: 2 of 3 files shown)


megies:
Looks good in general..
 - needs to be rebased it looks like (should be based on `master` as a new feature)
 - i would rather remove the logger again, we generally try to not be verbose unless its explicit warnings and here the code is just doing what it is supposed to do

NikolaosSokos:
I've completely removed the logger and the timing output.
I also rebased the commits against the newest master, changed the PR's base branch to target master directly. Let me know if anything else is needed!

PR #3718, opened 2026-03-22
add publicID to station.extra when reading SCML (supercedes #3194)

### LINKED ISSUE #3193: Adding publicID of seiscomp inventories to station object

filefolder:
### What does this PR do?

This is a remake of the derelict #3194 which reads and logs the station-level SCML publicID attribute. I added extra tests recommended previously and based on master.

Closes #3193 

Also updates the generic "scml import" `source` attribute with "ObsPy SCML import" and adds an additional SCML 0.14 inventory test file which will come in handy later.

### AI used?

yes, to remake some re compiles in seiscomp/tests/test_inventory/test_compare_xml since I will never understand how those work

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,7 +2,9 @@ master
 ======
 
 Changes:
- - ...
+ - General
+ - obspy.io.seiscomp:
+   * Add publicID to station.extra for select (see #3193 #3194)
 
 maintenance_1.5.x
 =================
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,7 +4,7 @@ master
 Changes:
  - General
  - obspy.io.seiscomp:
-   * Add publicID to station.extra for select (see #3193 #3194)
+   * Add publicID to station.extra for select (see #3718 #3193 #3194)
 
 maintenance_1.5.x
 =================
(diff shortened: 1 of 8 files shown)


filefolder:
ubuntu-3.14 fail related to #3719

megies:
I just restarted that failing test, hoping it might go away and a specific version of a dependency responsible, and now there is also a failing test in the offline parts of the FDSN client tests: https://tests.obspy.org/156296/#1

Maybe there is a more general problem, not something to be fixed in tests? But I'm failing to reproduce it locally. I tried to use similar dependency versions like in that Ubuntu CI runner but no luck.

ThomasLecocq:
@megies should this be 1.5.1 or master?

megies:
> @megies should this be 1.5.1 or master?

Definitely a feature change, so not for a bugfix release. (and it's correctly based on master already)

PR #3719, opened 2026-03-22
3.14 pytest issues: suppress sqlite3 warnings, add context managers to mp calls

filefolder:
python3.14 causing various test issues, some can be fixed on our end, others probably have to be ignored

1) multiprocessing causing issues in pytest, solution is mostly to wrap in context managers and/or use threadpool.

2) there are some places where SQLite databases aren't closing which 3.14 no longer tolerates. seems to only be a problem in pytest, solution is probably just to ignore these warnings and/or update pytest-cov.

3) increased timeout buffer for py3.10 mindep ubuntu to 10 minutes (5 wasn't cutting it), force use of the faster libmamba solver which it didn't appear to be using.

based on master


filefolder:
OK this should be working now. I think it may be an issue with PROJ-- have just suppressed the sqlite warnings specific to the cartopy tests in the `image_path` function in pytest

obspy/conftest.py
```diff
@@ -136,8 +136,15 @@ def image_path(request, save_image_directory):
...
     # finally close all figs created by this test
+    import gc
+    import warnings
     from matplotlib.pyplot import close
     close('all')
+    # hide these from PROJ unclosed sqlite handles
+    # which python 3.14 no longer tolerates
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", ResourceWarning)
+        gc.collect()
```
megies: this looks so weird.. calling `gc.collect()` and then it is emitting warnings too. I've seen some weird errors in pytest, so I assume it's one of those, do you have a link to an error on https://tests.obspy.org that gets fixed by this?

filefolder: yes here: https://github.com/obspy/obspy/pull/3718

i am not sure gc.collect() actually needed there in hindsight.

seems like fundamentally it is out of our control so just filtering the warnings may be required, but i honestly couldn't figure it out

filefolder: does appear that the gc was needed afterall

megies: link to actual test run: https://tests.obspy.org/155382/

megies:
see https://github.com/obspy/obspy/pull/3718#issuecomment-4289297241

filefolder:
suspect the new FDSN bug is 3.14 being stricter about a missing pool.close() or pool.terminate() in mass downloader. may have to go through all these.

megies:
I've tried locally with 3.14 though and could not replicate any of these.. have you been able to get these fails locally?

filefolder:
> I've tried locally with 3.14 though and could not replicate any of these.. have you been able to get these fails locally?

I have not.. but I think the issue could be here:

```
    from multiprocessing.resource_tracker import main;main(14)
  File "/home/runner/miniconda3/envs/test/lib/python3.14/multiprocessing/resource_tracker.py", line 34, in <module>
    _IGNORED_SIGNALS = (signal.SIGINT, signal.SIGTERM)
                        ^^^^^^^^^^^^^
AttributeError: module 'signal' has no attribute 'SIGINT' (consider renaming '/home/runner/miniconda3/envs/test/lib/python3.14/site-packages/obspy/signal/__init__.py' since it has the same name as the standard library module named 'signal' and prevents importing that standard library module)
```

Which leaves the mp stuff hanging specifically in pytest due to the overlapping "signal" imports. I already had to change something else regarding this for the 1.5 release using some creative importing. Fortunately by wrapping the 3-4 multiprocessing calls in a context manager (now done here) the problem doesn't seem to fail in pytest anymore and is probably a bit more robust anyway. 

Have re-shuffled a bit and mostly am now just suppressing the specific sqlite3 pytest warnings in pytest.ini.

There was also a new issue with py3.10 ubuntu mindep, the 5 minute timeout didn't seem to be enough. Have bumped to 10 and made the libmamba solver explicitly used. On that tangent maybe we should turn off the 3.8 and 3.9 tests now?

megies:
Looking at this again, I think it should go into maintenance branch? I've merged maintenance into master, so the sqlalchemy fixes are in there now and I will retrigger tests here now to see what the status is.

ThomasLecocq:
so the thing with "signal" being shadowed by obspy.signal is solved in #3789 - leaving this open for the other issues, we need to check if they still come up.

Issue #3549, opened 2025-03-08
UnicodeDecodeError in massdownloader


zhong-yy:
### Avoid duplicates

- [x] I searched existing issues

### Bug Summary

`UnicodeDecodeError` only occurs for some starttimes and endtimes.  At first, I suspect that some data files might be corrupted, and I tried to the find out those corrupted files by shrinking the time range for downloading.  Then I was surprised to find that I couldn't reproduce the error if I split the time range into two period. Thus, I think this should be a bug.

In the following example, if I set `starttime=obspy.UTCDateTime(2018, 7, 1, 0, 0, 0)` and `endtime=obspy.UTCDateTime(2022, 10, 1, 0, 0, 0)`, the `UnicodeDecodeError` will occur. However, no error is raised for the time ranges `2018-07-01 to 2020-10-01` and `2018-10-01 to 2022-10-01`

### Code to Reproduce

```python
import obspy
from obspy.clients.fdsn.mass_downloader import (
    RectangularDomain,
    Restrictions,
    MassDownloader,
)

domain = RectangularDomain(
    minlatitude=60.93,
    maxlatitude=61.68,
    minlongitude=-153.04,
    maxlongitude=-151.49,
)

restrictions = Restrictions(
    starttime=obspy.UTCDateTime(2018, 7, 1, 0, 0, 0),
    endtime=obspy.UTCDateTime(2022, 10, 1, 0, 0, 0),
    chunklength_in_sec=86400, # Chunk it to have one file per day.
    network="AV",
    station="SPBL",
    location=None,
    channel="BHZ",
    reject_channels_with_gaps=False,
    minimum_length=0,
    minimum_interstation_distance_in_m=100,
)

mdl = MassDownloader(providers=["IRIS"],debug=True)
mdl.download(
    domain,
    restrictions,
    download_chunk_size_in_mb=100,
    threads_per_client=16,
    mseed_storage="waveforms",
    stationxml_storage="stations",
)
```

### Error Traceback

```Python traceback
Traceback (most recent call last):
  File "/home/zyy/miniconda3/envs/test/lib/python3.10/site-packages/obspy/io/mseed/util.py", line 608, in _get_record_information
    _t = file_object.read(120).decode().strip()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 0: invalid start byte

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/zyy/test_mdl/download.py", line 50, in <module>
[...]

PR #3721, opened 2026-03-24
more distinct miniseed read errors, print zero size instead of unknown format

filefolder:
### What does this PR do?

the "Invalid MiniSEED file" error messages are often a bit too vague to determine what is happening (e.g. #3549). Now expanded somewhat.

Also catches the read in base.py-- after all attempts to determine format it will finally check if the file is zero before printing the "unknown format".

on master

### AI used?

for sanity washing

### DIFF
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -435,16 +435,16 @@ class TestWaveformPlugins:
         st2 = read(ascii_path / 'slist.ascii')
         assert st1 == st2
 
-    def test_raise_on_unknown_format(self):
+    def test_raise_on_empty_file(self):
         """
-        Test case for issue #338:
+        Test reading an empty file.
         """
         with NamedTemporaryFile() as tf:
             tmpfile = tf.name
             # create empty file
             open(tmpfile, 'wb').close()
             # using format keyword
-            with pytest.raises(TypeError):
+            with pytest.raises(ValueError, match="File is empty"):
                 read(tmpfile)
 
     def test_deepcopy(self):
--- a/obspy/core/tests/test_waveform_plugins.py
+++ b/obspy/core/tests/test_waveform_plugins.py
@@ -435,16 +435,16 @@ class TestWaveformPlugins:
         st2 = read(ascii_path / 'slist.ascii')
         assert st1 == st2
 
-    def test_raise_on_empty_file(self):
+    def test_raise_on_unknown_format(self):
         """
-        Test reading an empty file.
+        Writes and reads a small non-empty file.
         """
         with NamedTemporaryFile() as tf:
             tmpfile = tf.name
-            # create empty file
-            open(tmpfile, 'wb').close()
+            # create small non-empty file
+            open(tmpfile, 'wb').write(b'\x82\x00\x00')
             # using format keyword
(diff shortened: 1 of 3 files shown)


obspy/core/tests/test_waveform_plugins.py
```diff
@@ -437,12 +437,12 @@ def test_read_zip_archive(self, datapath, ascii_path):
...
         """
-        Test case for issue #338:
+        Writes and reads a small non-empty file.
         """
         with NamedTemporaryFile() as tf:
             tmpfile = tf.name
-            # create empty file
-            open(tmpfile, 'wb').close()
+            # create small non-empty file
+            open(tmpfile, 'wb').write(b'\x82\x00\x00')
```
ThomasLecocq, line 445: @filefolder you're changing the behaviour here ? I see the change in base.py is raising a ValueError when the file is empty, but the pytest below only checks TypeError, no?

obspy/io/mseed/util.py
```diff
@@ -607,14 +607,16 @@ def _get_record_information(file_object, offset=0, endian=None):
...
+                             "unable to decode header bytes.")
         if not _t:
             info = _get_record_information(file_object=file_object,
                                            endian=endian)
             file_object.seek(initial_position, 0)
             return info
         else:
-            raise ValueError("Invalid MiniSEED file.")
+            raise ValueError("Invalid MiniSEED file: "
+                             "unexpected content in noise record.")
```
ThomasLecocq, line 619: noise?

filefolder: ha I don't actually know. the comment for that block is `# Might be a noise record or completely empty.` and I continued on with that just to better isolate where the failure was happening. IIRC the whole point to this PR is just to narrow down what's actually causing the error.

obspy/core/tests/test_waveform_plugins.py
megies: Yeah, has to be reverted, this is an intentional test for empty files clearly.

obspy/core/util/base.py
```diff
@@ -536,6 +536,9 @@ def _read_from_plugin(plugin_type, filename, format=None, **kwargs):
             if is_format:
                 break
         else:
+            if isinstance(filename, str) and \
+                    Path(filename).stat().st_size == 0:
+                raise ValueError("File is empty: {}".format(filename))
```
megies, line 541: In principle something like this is nice, but if we do it, I think we should have a proper function that goes through all the options a read input can have (str, BytesIO, ...) and checks for empty file.

obspy/io/mseed/util.py
megies: But, you clearly did this for an unclear error you encountered right? So what was the actual problem in your case? Was it a noise record?

megies:
I feel like let's wait for feedback. It seems a bit unclear what the actual use case for this was and if it helped make the error clearer.
I think it could be nice to have a better message for empty file in general, but then we should try to do it for all flavors the incoming read input can come in as.
That test change would need to be reverted too. All in all, maybe not quite ready yet and needs more discussion I feel?

megies (changes requested):

filefolder:
TBH this PR can probably be flushed; it was mainly just to provide a clue for issues like #3549

PR #3722, opened 2026-03-24
slice() now working on reference instead of copy

raciner:
…aking a full copy.

### What does this PR do?

The current situation is as follows:
```
"""
[...]
        :return: New :class:`~obspy.core.trace.Trace` object. Does not copy
            data but just passes a reference to it.
[...]
"""
tr = copy(self)

```
The proposed change replaces the current code with code which works on references, reducing time and memory consumption in cases a program uses a lot of slices. In theory, this could cause trouble for people relying on the fact, that the current code would make a deep copy of the data, although it is clearly stated differently in the manual. I am not sure whether pushing this to maintenance is appropriate.

### AI used?

Copilot used to make the code look pythonic.

### PR Checklist

- [X] **Tests**: Added new tests for any new features or fixed regressions
- [X] **Changelog**: Added a short note in `CHANGELOG.txt` (only obsolete if fixing a bug introduced *after* the last release)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,8 @@ Changes:
      (see #3709)
  - obspy.clients:
    * SDS client: fix handling of empty files (see #3709)
+ - obpsy.core.trace: 
+   * slice now works on references instead of making a copy of the trace
 
 1.5.0 (doi: 10.5281/zenodo.19005357)
 ====================================
--- a/obspy/core/tests/test_stream.py
+++ b/obspy/core/tests/test_stream.py
@@ -2281,7 +2281,7 @@ class TestStream:
         assert "filter" in pr[4] and "lowpass" in pr[4]
         assert "differentiate" in pr[5]
         assert "integrate" in pr[6]
-        assert "trim" in pr[7]
+        assert "slice" in pr[7]
         assert "detrend" in pr[8]
         assert "taper" in pr[9]
         assert "normalize" in pr[10]
--- a/obspy/core/tests/test_trace.py
+++ b/obspy/core/tests/test_trace.py
@@ -602,7 +602,7 @@ class TestTrace:
         tr2 = tr.slice(tr.stats.starttime)
         assert "processing" not in tr.stats
         assert "processing" in tr2.stats
-        assert "trim" in tr2.stats.processing[0]
+        assert "slice" in tr2.stats.processing[0]
 
     def test_slice_no_starttime_or_endtime(self):
         """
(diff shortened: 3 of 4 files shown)


megies:
Current code for `Trace.slice()` is pretty simple and straightforward. First we make a shallow copy of the whole trace and then we replace the trace header with a deepcopy of the original header (to not potentially backpropagate header changes to the original trace).

> The proposed change replaces the current code with code which works on references, reducing time and memory consumption in cases a program uses a lot of slices.

Current code indeed works on a reference, when it comes to the data part, which should have the most impact on time/memory.

> that the current code would make a deep copy of the data, although it is clearly stated differently in the manual

Data part does not get deepcopied.

---

Is this from an AI code review of the code base? Potentially you or an AI got mislead by `tr = copy(self)` which uses `copy.copy` and ends up with a shallow copy of the data array.

There is also a test for slice that breaks if you internally replace slice with a deepcopy..

```diff
diff --git a/obspy/core/trace.py b/obspy/core/trace.py
index f42170877..adac3719f 100644
--- a/obspy/core/trace.py
+++ b/obspy/core/trace.py
@@ -1233,8 +1233,8 @@ class Trace(object):
         >>> tr2.data
         array([2, 3, 4, 5, 6, 7, 8])
         """
-        tr = copy(self)
-        tr.stats = deepcopy(self.stats)
+        tr = deepcopy(self)
+        #tr.stats = deepcopy(self.stats)
         tr.trim(starttime=starttime, endtime=endtime,
                 nearest_sample=nearest_sample)
         return tr
```

```
  File "/home/megies/git/obspy/obspy/core/tests/test_trace.py", line 593, in test_slice
    assert tr.data[2] == 10
AssertionError: assert 2 == 10
========================================================= short test summary info ==========================================================
FAILED core/tests/test_trace.py::TestTrace::test_slice - assert 2 == 10
```

in this test:

```python
 584     def test_slice(self):
 585         """
 586         Tests the slicing of trace objects.
 587         """
 588         tr = Trace(data=np.arange(10, dtype=np.int32))
 589         mempos = tr.data.ctypes.data
 590         t = tr.stats.starttime
 591         tr1 = tr.slice(t + 2, t + 8)
 592         tr1.data[0] = 10
 593         assert tr.data[2] == 10
 594         assert tr.data.ctypes.data == mempos
 595         assert tr.data[2:9].ctypes.data == tr1.data.ctypes.data
 596         assert tr1.data.ctypes.data - 8 == mempos
 597 
 598         # Test the processing information for the slicing. The sliced trace
 599         # should have a processing information showing that it has been
 600         # trimmed. The original trace should have nothing.
 601         tr = Trace(data=np.arange(10, dtype=np.int32))
 602         tr2 = tr.slice(tr.stats.starttime)
 603         assert "processing" not in tr.stats
 604         assert "processing" in tr2.stats
 605         assert "trim" in tr2.stats.processing[0]
```

Also here's a quick visual check:

```python
from obspy import read
tr = read()[0]
tr2 = tr.slice(tr.stats.starttime+5, tr.stats.endtime-5)
tr2.data[:] = 0
tr.plot()
```

Figure_1

---

In terms of speedup/memory use.. Did you benchmark this change? If it makes it significantly faster we could look more into this and fine tune the slicing (even though I feel like the speedup needs to be really significant because the code would be much harder to understand with this change.

raciner:
I made this discovery myself without AI when working with huge data sets. However, I dropped following it up. I will check again whether my claim holds or not...

megies:
Ok, let's bump to future release for now, let us now if you have more insight what your issue was :+1:

PR #3736, opened 2026-05-05
tsindex: fix unclosed sqlite objects causing fails in CI tests

megies:
### What does this PR do?

Tries to fix current issues with failing tests due to unclosed sqlite objects. See e.g...
 - https://github.com/obspy/obspy/actions/runs/25368414681/job/74395099463?pr=3735#step:10:795
 - https://tests.obspy.org/156570/#1

Hard to tell where exactly the problem is, since it seems pytest is failing in another unrelated test finally, when the problem originates in some other test??

### Links to other Issues?

--

### AI used?

no

### PR Checklist

Rare actions items:

- [ ] First time contributors: Feel free to add your name to `CONTRIBUTORS.txt`

### Issue labels

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,6 +2,9 @@ maintenance_1.5.x
 =================
 
 Changes:
+ - obspy.clients.filesystem:
+   * tsindex: improve handling of sql sessions, less connections staying open
+     (see #3736)
  - obspy.io.mseed:
    * fix the type of exception raised when trying to read an empty file
      (see #3709)
--- a/obspy/clients/filesystem/tests/test_tsindex.py
+++ b/obspy/clients/filesystem/tests/test_tsindex.py
@@ -9,7 +9,7 @@ from unittest import mock
 
 import pytest
 import sqlalchemy as sa
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import sessionmaker, close_all_sessions
 
 from obspy import read, UTCDateTime
 from obspy.clients.filesystem.tsindex import Client, Indexer, \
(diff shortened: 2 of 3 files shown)


ThomasLecocq:
I have the same issues of sqlite non closing for msnoise. I wonder if it has to do with the gpfs (or whatever cloud filesystem of the CI builder) , that takes longer to free file handles etc?

megies:
@ThomasLecocq apparently it's "only" Python 3.13+ issuing a warning on unclosed sqlite resources (DBs, connections, ...) which wasn't issued before. So, from what I gather is, nothing really changed, just these warnings now expose dirty code and ultimately then also trip up pytest in a quite weird way.

It's quite weird to debug, because the failures don't happen in the problematic test methods itself, but in a different test method when pytest finally thinks that this left over unraised warnings (only happens with `-W error`) should be issued now.
So it's a lot of backtracking where exactly the dirty handles are happening.

I've seen people that say they just added ignore rules for these sqlite warnings in the `pytest.ini` but it feels like maybe this should just be handled more cleanly in the code..?

First step here was to get rid of the pytest warnings by closing all stuff where needed in the tests, but I feel like it should now be actually done in the code (in the `tsindex` stuff). No expert on sqlalchemy, but from what I see it recommends using stuff like `with Session() as session: ...` which closes sessions at the end of the block. I feel like in the code, there just is no management of sessions/connections at all and everthing is just ultimately left to gc.

:shrug:

megies:
In any case, in terms of the test suite. Definitely there should be no interaction between tests. And this caused interactions (other tests failing from previous tests' issues)

megies:
I wrapped any opening sessions I saw in the code in context managers, but still our tests had issues if not deploying a `close_all_sessions()` nuke after each test method to clean up. I also tried adding destructors to clean up in the code itself but that didn't make a difference it seemed, still needed the cleanup in the test code afterwards. So I reverted the destructors to not break things I'm not aware of and not covered by the tests for no reason. I left it in with a revert for future reference though.

megies:
remaining fail is unrelated

Issue #3731, opened 2026-04-22
mseed reader fails on Emscripten/WASM system: np.memmap not available


MMesch:
### Avoid duplicates

- [x] I searched existing issues

### Bug Summary

Introducing np.memmap as default file reader for mseed in [this commit](https://github.com/obspy/obspy/commit/19c97c9c9bccfc869f0e6d7c694d960ee2a2ccfa#diff-0a262d8ab34d16672e6a952b75f55be2025a31442e6f82fe9bb1b5b27f3c4e59R381-R387), unfortunately broke out of the box emscripten use of obspy. I guess it's not a "real" bug because that platform isn't officially supported yet but flagging it up.

Meanwhile a monkey patch workaround exists:

```
def fake_memmap(filename, dtype=np.uint8, mode=None, offset=0, shape=None, order=None):
    with open(filename, "rb") as f:
        f.seek(offset)
        data = np.fromfile(f, dtype=dtype)
    if shape is not None:
        data = data.reshape(shape)
    return data

np.memmap = fake_memmap
```

Rather than fixing this one by one, it may make sense to do a global pass over obspy and add some guards that ensure emscripten compatibility and make it officially supported. I don't think many changes are required. Other thoughts? I can give this a go.

### Code to Reproduce

```python
from obspy import read
st = read("000000000_0036EE80.mseed") # standard example file

on an emscripten deployment (e.g. https://mmesch.com/jupyterlite-obspy/lab/index.html)
```

### Error Traceback

```Python traceback
File /lib/python3.13/site-packages/obspy/core/util/base.py:559, in _read_from_plugin(plugin_type, filename, format, **kwargs)
    557     raise TypeError(msg % (format_ep.name, ', '.join(eps)))
    558 # read
--> 559 list_obj = read_format(filename, **kwargs)
    560 return list_obj, format_ep.name

File /lib/python3.13/site-packages/obspy/io/mseed/core.py:447, in _read_mseed(mseed_object, starttime, endtime, headonly, sourcename, reclen, details, header_byteorder, verbose, use_bisection, **kwargs)
    443 # If it's a file name just read it.
    444 if isinstance(mseed_object, str):
    445     # Read to NumPy array which is used as a buffer.
    446     # use memmap, faster than fromfile, same functionality
--> 447     bfr_np = np.memmap(mseed_object, dtype=np.int8, mode="c")
    449 elif hasattr(mseed_object, 'read'):
[...]

PR #3741, opened 2026-05-22
MSEED: fallback from np.memmap to np.frombuffer in case memmap is not implemented

raciner:
### What does this PR do?

This PR fixes the issue that reading of miniseed files fails in case numpy.memmap is not implemented. In this case, the code will now read the file into a buffer the classic way. Also, the tests are updated to include a test of the fallback solution.

### Links to other Issues?

See #3731

### AI used?

No

### PR Checklist

- [X] **Tests**: Added new tests for any new features or fixed regressions

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -20,6 +20,7 @@ Changes:
    * fix the type of exception raised when trying to read an empty file
      (see #3709)
    * fix a void return type in write mseed ctypes wrapper (see #3735)
+   * fix reading in case when numpy.memmap is not implemented (see #3741)
  - obspy.clients:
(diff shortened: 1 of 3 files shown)


ThomasLecocq (approved):

megies:
I think we could also have this targeting `maintenance_1.5.x` branch and use it for 1.5.1 bugfix release.. it should not change any functionality, since it won't change anything if memmap can be used and only kicks in in super rare edge cases when it would have just crashed anyway.

megies:
Cherry picked your commit onto maintenance, think this classifies as a bug fix for 1.5.1

PR #3743, opened 2026-05-29
Fix reftek130 malformed eh header

athulpalliath:
Old REFTEK DAS firmware  sometimes wrote blank or scrambled 
fields in the EH packet binary header, causing ObsPy to raise a 
ValueError even though the DT packets containing actual waveform data 
are intact.

This PR makes two changes:
1. Adds fallback_sampling_rate parameter to _read_reftek130() and 
   to_stream() so users can specify the sampling rate when the EH 
   packet sampling_rate field cannot be parsed
2. Adds graceful handling in EHPacket.__init__ so malformed fields 
   are set to None with a warning instead of crashing

Tested on real data from REFTEK units deployed in the Koyna-Warna 
seismic network (India) circa 2006-2008.


megies:
This seems like curious edge cases, but then again, I don't think it will hurt much to add anyway.

Can you add a *small* example file, just a handful of reftek packets long would be ideal, that showcases the problem, and then also a short test case for the new functionality would be needed. Also changelog. All of this is part of the pull request template we put in place.

obspy/io/reftek/core.py
```diff
@@ -90,6 +91,15 @@ def _read_reftek130(filename, network="", location="", component_codes=None,
...
         permuted package sequence (showing the related warning message).
+    :type fallback_sampling_rate: float, optional
+    :param fallback_sampling_rate: Sampling rate to use if the sampling rate
+        field in the EH packet cannot be parsed (e.g. due to old firmware
+        writing malformed header fields). If None and the sampling rate cannot
+        be parsed, a Reftek130Exception is raised. Example usage::
+
+            >>> from obspy import read
+            >>> st = read("file.rt130", format="REFTEK130",
+            ...           fallback_sampling_rate=100.0)  # doctest: +SKIP
```
megies: The part with "Example usage: ... " and the code block can be removed. No need to show how to use a simple kwarg. Should be clear from the description. The builtin keyword None needs to go into double backticks for sphinx:
```
``None``
```

obspy/io/reftek/core.py
```diff
@@ -207,11 +218,15 @@ def drop_not_implemented_packet_types(self):
...
-                  sort_permuted_package_sequence=False):
+                  sort_permuted_package_sequence=False,
+                  fallback_sampling_rate=None):
         """
         :type headonly: bool
         :param headonly: Determines whether or not to unpack the data or just
             read the headers.
+        :type fallback_sampling_rate: float, optional
+        :param fallback_sampling_rate: Sampling rate to use if the sampling
+            rate field in the EH packet cannot be parsed.
```
megies: There is more detail info in the other docstring, so I would either link to there like...
```
More detail, see :func:`_read_reftek130`
```
.. or simply put the same detail info here too

megies (changes requested):
Needs a *small* test file and test case testing both behaviors (raising exception vs. successful reading with warning), then also changelog and some other minor changes

athulpalliath:
Added the test fixtures, test cases (covering both the raise-without-fallback and read-with-fallback behaviors, plus a case for other malformed EH fields), changelog entry, and cleaned up the docstrings as requested. Ready for another look.

PR #3745, opened 2026-06-01
Fix freqattributes.spectrum to honour n1/n2 sub-window (fixes #3491)

### LINKED ISSUE #3491: Incorrect behavior of obspy.signal.freqattributes.spectrum
### Avoid duplicates

### Bug Summary

The spectrum() function accepts inputs n1 and n2 claimed to be start and end indices (presumably for the input time series, but I don't think it's 100% clear that it's the intent). However, this behavior is not implemented; changing n1 and n2 does not change the output spectrum length or values at all. 

Additionally, spectrum() does not scale its output to match either power spectral density or energy spectral density (it matches the time series energy only for the case of dt=1 in whatever units are used, presumably seconds). spectrum() doesn't claim to use any specific scaling so this isn't necessarily a big problem per se. But, because welch() claims to return power spectral estimates, and calculates them using spectrum(), its output is not properly scaled.

### Code to Reproduce

```python
from obspy.signal.freqattributes import spectrum
from obspy.signal.invsim import cosine_taper
import numpy as np

## generate noise
l = 2**15
x = np.random.normal(0, 1, l)

## standardize x
x -= np.mean(x)
x /= np.sqrt(np.mean(x**2))

## define window
w = cosine_taper(l)

## calculate spectrum without specifying n1 and n2, and check length & Parseval's relation
s = spectrum(x, w, l)
print([len(s), l, "spectrum is the same length as the input signal"])
print([np.sum(s), np.sum(x**2), "sum of spectrum matches energy of time series assuming dt = 1"])
print([np.sum(s)/l, np.mean(x**2),"dividing by l would make it match power of time series always"])

## calculate spectrum while defining n1 and n2, and check length & Parseval's relation
n1 = 30
n2 = 500
s = spectrum(x, w, l, n1, n2)
print([len(s), n2-n1, "spectrum is still the same length as the original input signal, not n2-n1"])
print([np.sum(s), np.sum(x**2), "sum of spectrum matches energy of time series assuming dt = 1"])
```

### Error Traceback

_No response_

### ObsPy Version?

1.4.0

### Operating System?

Ubuntu 20.04

### Python Version?

3.11.6

### Installation Method?

pip

gaoflow:
## What

Fixes #3491.

`obspy.signal.freqattributes.spectrum(data, win, nfft, n1, n2)` documents `n1`/`n2` as start/end indices into the input series (*"If n1 and n2 are not specified the periodogram of the entire sequence is returned"*), but they were only used to form the normalisation scalar `n = n2 - n1` — the FFT was always taken over the **whole** `data` array (`xw = data * win`). So changing `n1`/`n2` had no effect on the output.

As a direct consequence, `welch()` — which advances `n1`/`n2` to average periodograms of sliding sections — actually averaged the spectrum of the whole array with itself (returning the single-section result), and raised `ValueError: operands could not be broadcast together` whenever the section length `l` was shorter than the data, because the (float) section indices were never applied to `data`.

## Fix

- `spectrum()`: compute the FFT on the requested sub-window, `xw = data[n1:n2] * win`.
- `welch()`: rewrite the loop to step over fixed-length, integer-indexed sections (`step = round((1 - over) * l)`, guarding `step >= 1` so `over = 1` no longer divides by zero), so each section is exactly `l` samples and matches `win`.

## Backwards compatibility

Default behaviour is unchanged:
- `spectrum(data, win, nfft)` → `n1 = 0, n2 = len(data)` → periodogram of the whole array (identical to before).
- A full-length window collapses `welch()` to a single section equal to `spectrum()` of the whole array — this is exactly how `welch()` is used internally by `central_frequency_unwindowed`, so that path is unaffected.

Only previously-broken cases change: passing non-default `n1`/`n2`, or `welch()` with a section shorter than the data (which used to raise).

## Note on scaling

The issue also notes that `spectrum()`/`welch()` are not scaled to a true PSD/ESD. That is a separate, output-magnitude-changing concern with backwards-compatibility implications, so it is intentionally left out of this PR, which targets the indexing bug.

## Tests

Existing `test_freqattributes.py` passes unchanged. Added two regression tests (`test_spectrum_respects_n1_n2`, `test_welch_averages_distinct_sections`) that fail on `master` and pass with the fix. flake8 clean; CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,7 +2,12 @@ master
 ======
 
 Changes:
- - ...
+ - obspy.signal:
+   * freqattributes.spectrum now honours the n1/n2 sub-window arguments (they
+     were previously ignored), which also fixes freqattributes.welch so that it
+     averages periodograms of distinct, correctly-sized sections instead of
+     repeatedly spectrum-ing the whole array (or raising a broadcasting error
+     for section lengths shorter than the data) (see #3491)
 
 maintenance_1.5.x
 =================
--- a/obspy/signal/freqattributes.py
+++ b/obspy/signal/freqattributes.py
@@ -48,7 +48,7 @@ def spectrum(data, win, nfft, n1=0, n2=0):
         n2 = len(data)
     n = n2 - n1
     u = pow(np.linalg.norm([win]), 2) / n
-    xw = data * win
+    xw = data[n1:n2] * win
     px = pow(abs(fftpack.fft(xw, nfft)), 2) / (n * u)
     px[0] = px[1]
     return px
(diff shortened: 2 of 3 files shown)


binggao1230:

PR #3746, opened 2026-06-01
Keep stream unchanged when ->ZNE rotation fails (fixes #2692)

### LINKED ISSUE #2692: Failed rotation modifies stream inplace
Obspy version: 1.2.2 (installed from source)
Python version: 3.7.7
OS: Ubuntu 18.04

When a rotation failes between lines 3492 and 3527 of stream.py (see below), all traces corresponding to the failed rotation are removed from the original stream. This can be problematic, e.g. when trying to output debug information about a stream with failed rotation. I'd suggest modifing the rotation function to keep the original stream unchanged unless the rotation succeeds.

 https://github.com/obspy/obspy/blob/88687a146a7c3ca8f35c608db2243cf5fed6813c/obspy/core/stream.py#L3492-L3527

Code to reproduce issue:
```
import obspy

client = Client('IRIS')
inv = client.get_stations(net='IU', station='TEIG', location='10', channel='HH*', level='response')

stream = obspy.read('inplace_modification.mseed')
print(f'Stream length before: {len(stream)}')

try:
    stream.rotate('->ZNE', inventory=inv)
except ValueError:
    print('Rotation failed')

print(f'Stream length after: {len(stream)}')
```
Files: 
[inplace_modification.mseed.zip](https://github.com/obspy/obspy/files/5007899/inplace_modification.mseed.zip)

Output:
```
Stream length before: 3
Rotation failed
Stream length after: 0
```

gaoflow:
## What

Fixes #2692.

`Stream.rotate(method="->ZNE", inventory=...)` modifies the stream in place *before* the rotation can fail. In `_rotate_specific_channels_to_zne()` the original traces are removed from the stream (and `_trim_common_channels()` trims them in place) up front, but the metadata lookup (`metadata_getter`) and `rotate2zne()` happen afterwards. If any of those fails — most commonly because the inventory has no orientation metadata for one of the channels — the traces have already been removed/trimmed, leaving the stream **corrupted or completely empty**.

### Reproduce

```python
import numpy as np
from obspy import Trace, Stream, UTCDateTime
from obspy.core.inventory import Inventory

st = Stream([Trace(np.arange(100.), header={
    'network': 'BW', 'station': 'FFB1', 'location': '',
    'channel': 'HH' + c, 'sampling_rate': 100.0,
    'starttime': UTCDateTime(0)}) for c in '12Z'])

inv = Inventory(networks=[], source='test')  # no metadata -> rotation fails
try:
    st.rotate('->ZNE', inventory=inv)
except Exception:
    pass
print(len(st))   # master: 0  (stream wiped!)  -> expected: 3
```

## Fix

Do all of the work (trim, sort, metadata lookup, rotation) on a **copy**, and only commit the result back to the stream — removing the original traces and appending the rotated ones — once the rotation has fully succeeded.

## Backwards compatibility

The successful path is unchanged: verified against the existing `->ZNE` test data (`ffbx_unrotated_gaps.mseed` + `ffbx.stationxml`), which still rotates BH1/BH2/BHZ and HH1/HH2/HHZ (with gaps) to ZNE exactly as before. All existing rotation tests pass (`test_stream.py` + `test_rotate.py`).

## Tests

Added `test_rotate_to_zne_failure_keeps_stream_unchanged`, which asserts the stream is intact after a failed rotation. It fails on `master` (stream becomes empty) and passes with the fix. flake8 clean; CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,7 +2,11 @@ master
 ======
 
 Changes:
- - ...
+ - obspy.core:
+   * Stream.rotate(method="->ZNE", ...) no longer modifies the stream in place
+     when the rotation fails (e.g. because orientation metadata is missing from
+     the inventory); the original traces are now kept until the rotation has
+     fully succeeded (see #2692)
 
 maintenance_1.5.x
 =================
--- a/obspy/core/stream.py
+++ b/obspy/core/stream.py
@@ -3606,9 +3606,12 @@ seismometer_correction_simulation.html#using-a-resp-file>`_.
         st = self.select(network=network, station=station, location=location)
         st = (st.select(channel=channels[0]) + st.select(channel=channels[1]) +
               st.select(channel=channels[2]))
-        # remove the original unrotated traces from the stream
-        for tr in st.traces:
-            self.remove(tr)
+        # Remember the original traces but do not modify the stream yet: a
+        # failing rotation must leave the input stream unchanged (see #2692).
+        # All work is therefore done on a copy and only committed back to the
+        # stream once the rotation has fully succeeded.
+        original_traces = list(st.traces)
+        st = st.copy()
         # cut data so that we end up with a set of matching pieces for the tree
         # components (i.e. cut away any parts where one of the three components
         # has no data)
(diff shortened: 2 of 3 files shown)


binggao1230:

PR #3747, opened 2026-06-01
Fix NDK format autodetection for negative hypocenter depth (fixes #2823)

### LINKED ISSUE #2823: read_events fails to determine ndk format automatically for some events
```python
import obspy as op
cat = op.read_events('https://www.ldeo.columbia.edu/~gcmt/projects/CMT/catalog/NEW_QUICK/2018/E201807010051A.ndk')
```
`ERROR: Unknown format for file` raises

If specifying the format, it works.

gaoflow:
## What

Fixes #2823.

`read_events()` fails to autodetect the NDK format for some valid GCMT events (it raises `Unknown format for file`), even though reading works fine when the format is specified explicitly.

`obspy.io.ndk.core._is_ndk()` validates the first-record hypocenter depth with `(0 <= depth <= 800000)`. As @ThomasLecocq diagnosed in the issue, this assumes the depth is in **metres** and forbids negative values — but per the [NDK format documentation](https://www.ldeo.columbia.edu/~gcmt/projects/CMT/catalog/allorder.ndk_explained) the depth is in **kilometres** and may legitimately be negative (e.g. an event located above sea level). The reporter's file has a depth of `-1.1`, so autodetection rejected it while `_read_ndk()` parses it without any problem.

## Fix

Relax the check to `(-10 <= depth <= 800)` (depth in km), exactly as suggested by the maintainer in the issue and confirmed by the reporter.

## Tests

Added `test_is_ndk_negative_depth`, which builds a record with a negative depth and asserts `_is_ndk()` accepts it. It fails on `master` and passes with the fix. The full `test_ndk.py` suite (20 tests) passes; flake8 clean; CHANGELOG updated.

Credit to @ThomasLecocq for diagnosing the root cause and proposing the bound.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -2,7 +2,11 @@ master
 ======
 
 Changes:
- - ...
+ - obspy.io.ndk:
+   * fix format autodetection (_is_ndk) for NDK files whose hypocenter depth is
+     negative; depths are in kilometres and may be negative (e.g. events above
+     sea level), but the check assumed metres and rejected negative values
+     (see #2823)
 
 maintenance_1.5.x
 =================
--- a/obspy/io/ndk/core.py
+++ b/obspy/io/ndk/core.py
@@ -126,9 +126,13 @@ def _is_ndk(filename):
     except ValueError:
         return False
 
+    # The depth is given in kilometres and, according to the NDK format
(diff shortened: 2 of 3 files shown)


binggao1230:

PR #3748, opened 2026-06-01
Fix output units when unit-bearing blockette is not first in a stage

### LINKED ISSUE #2830: Inventory unit issue for multiple blockettes in the same stage
* Please check whether the bug was already reported or fixed.
* Please provide the following information:
  -  ObsPy version, Python version and Platform (Windows, OSX, Linux ...)
  -  How did you install ObsPy and Python (pip, anaconda, from source ...)
  -  If possible please supply a [short, self contained, correct example](http://sscce.org/) that
     demonstrates the issue.
  -  If this is a regression (used to work in an earlier version of ObsPy),
     please note when it used to work.
* Please take the time to [format your ticket appropriately](https://guides.github.com/features/mastering-markdown/)
  (e.g. put error tracebacks into code blocks)

Obspy version 1.2.2, OSX, installed via anaconda.

@jholland-usgs found an issue when there are multiple blockettes in the same stage the units can be wrong.

```
inv = read_inventory("explicit_example_.RESP.GS.OK001.01.HNZ")
# The sensitivity unit ends up being the output from Stage 3 instead of from stage 4
print(inv.networks[0].stations[0].channels[0].response.instrument_sensitivity.output_units)
```

The above produces an output of `WRONGUNIT` while the last blockette in stage 4 has units of counts:

```
B061F03     Stage sequence number:                 4
B061F05     Symmetry type:                         B
B061F06     Response in units lookup:              WRONGUNIT - Wrong Units
B061F07     Response out units lookup:             COUNT - Digital counts
B061F08     Number of numerators:                  146
#		Numerator coefficients:
#		  i, coefficient
```

Here is the RESP file that produces this issue: 
[explicit_example_.RESP.GS.OK001.01.HNZ.TXT](https://github.com/obspy/obspy/files/6392162/explicit_example_.RESP.GS.OK001.01.HNZ.TXT)

When I get a chance to look into this I will make a pull-request, but I wanted to document this here so that it isn't forgotten.

Please close this if it has already been identified.

gaoflow:
## What & why

Fixes #2830.

`read_inventory` (RESP / XSEED Parser) reported the wrong overall **output units** for a response when the final stage does not begin with a unit-bearing blockette.

`Parser._parse_blkt_response` determines the output units by looping over the stages in reverse order, but it only ever inspected the **first** blockette of each stage:

```python
for _s in reversed(sorted(stages.keys())):
    _s = stages[_s][0]          # <- first blockette of the stage only
    for attr in dir(_s):
        if "unit" in attr and "out" in attr and "__" not in attr:
            ...
```

When the last stage starts with a blockette that carries no units (e.g. a decimation blockette, B057) followed by the response blockette that actually holds the output units (B061), the inner `for…else: continue` skipped the **entire** final stage and the units fell back to the previous stage.

## Reproduction

Using the RESP file attached to #2830 (stage 3 out = `WRONGUNIT`, stage 4 out = `COUNT`):

```python
>>> from obspy import read_inventory
>>> inv = read_inventory("resp_2830.txt")
>>> inv[0][0][0].response.instrument_sensitivity.output_units
'WRONGUNIT'        # before  (should be COUNT)
'COUNT'            # after
```

The individual stages were always parsed correctly (`[(1,'V'),(2,'V'),(3,'WRONGUNIT'),(4,'COUNT')]`); only the overall instrument-sensitivity output units were wrong.

## Fix

Inspect **every** blockette of the stage instead of just the first one, iterating in reverse so the last unit-bearing blockette (the stage's final output) wins. Stages whose response blockette is the only/first one — the common case — are unaffected.

## Verification

- New regression test `test_output_units_from_later_blockette_in_stage` (with the #2830 RESP file as test data) — fails on master (`WRONGUNIT`), passes with the fix (`COUNT`).
- Full `obspy/io/xseed/tests/` suite passes (83 passed), so existing single-response-blockette stages are unchanged.
- `flake8` clean; CHANGELOG updated.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -178,6 +178,11 @@ Changes:
  - obspy.io.xseed:
    * Improve error message when trying to read a local path to a file that does
      not exist with XSEED Parser (see #2686)
+   * Determine the overall output units from the final stage even when the
+     unit-bearing blockette is not the first one in that stage (e.g. a stage
+     starting with a decimation blockette). Previously only the first blockette
+     of each stage was inspected, so such stages were skipped and the units
+     fell back to a previous stage (see #2830)
  - obspy.signal:
    * PPSD: performance improvement in __init__: faster frequency array
      calculation (see #3644)
--- a/obspy/io/xseed/parser.py
+++ b/obspy/io/xseed/parser.py
@@ -1216,15 +1216,22 @@ class Parser(object):
                 input_units = None
 
         # Find the output units by looping over the stages in reverse order
-        # and finding the first stage whose first blockette has an output
-        # unit set.
+        # and finding the highest-numbered stage that has a blockette with an
+        # output unit set. A single stage can contain several blockettes (e.g.
+        # a decimation blockette without units followed by a response blockette
+        # that carries them), so every blockette of the stage is inspected -
+        # in reverse, so the last unit-bearing blockette (the stage's final
+        # output) wins - rather than only the first one (see #2830).
         unit_lookup_key = None
         for _s in reversed(sorted(stages.keys())):
-            _s = stages[_s][0]
-            for attr in dir(_s):
-                if "unit" in attr and "out" in attr and "__" not in attr:
-                    unit_lookup_key = getattr(_s, attr)
-                    break
+            for _blkt in reversed(stages[_s]):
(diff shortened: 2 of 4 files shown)


binggao1230:

PR #3767, opened 2026-07-14
seedlink client: fix seedlink client not respecting user set timeout

megies:
### What does this PR do?

Currently when making the initial low-level socket connection for the seedlink client, no timeout is set, even if the client has a 20s default timeout and users can specify an even shorter one on `Client()` initialization. This leads to socket connection in some cases (unreachable server, wrong address, ...) timing out with a presumable default, which I found to be around 120s.
Fix this by also setting the client's stored timeout during the socket connection, so that shorter timeout on bad connections can be achieved.

Also remove a hard coded timeout inside the routine, which looks like it never did anything (since the blocking call is above it anyway) and replace it with the client's timeout (probably doesn't change anything, but for good measure..).

Not really sure how to properly test this, so no new tests for this were added.
The timeout happens in `socket.socket.connect()` which also includes DNS lookup and all kinds of things. In any case this change fixes the issue for me when trying to connect to a known seedlink server internet address which has a connection issue (meaning DNS lookup is likely working and done but the located IP address is not answering).

### Links to other Issues?

--

### AI used?

no

### PR Checklist

Rare actions items:

- [ ] First time contributors: Feel free to add your name to `CONTRIBUTORS.txt`

### Issue labels

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -25,6 +25,12 @@ Changes:
    * deprecate flinnengdahl web service, which EarthScope announced will be
      retired in July 2026. Users should use obspy.geodetics instead.
      (see #3756)
+ - obspy.clients.seedlink:
+   * respect user specified timeout also in making the initial socket
+     connection. Currently, when establishing the initial low-level socket
+     connection no timeout is used, which in practice leads to a default 2min
+     timeout even if user sets a shorter timeout when establishing the client
+     (see #3767)
  - obspy.imaging:
    * fix a bug in dayplot when data is only available for part of the time
      window (see #3780)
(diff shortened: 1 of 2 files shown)


ThomasLecocq (approved):

ThomasLecocq:
shouldn't this go into maintenance ?

megies:
I guess it could be seen either way.. it kind of alters behavior (faster timeout), that's why I put it here. Of course, you could also argue that the timeout should always have been properly acknowledged, so it could be seen as a fix too

megies:
rebased on maintenance branch

calum-chamberlain:
This fix seems to be resulting in errors for me instantiating a seedlink client using the EasySeedLinkClient without setting a timeout (thus `timeout=None` as default). This results in an error in `is_connected_impl` where the timeout (`None`) if compared to a float.

Issue #3768, opened 2026-07-15
Broadly useful default FDSN Client data center services mappings


chad-earthscope:
### Problem

The issue: now that EarthScope no longer offers an fdsnws-event service, the default client instantiation fails when events are requested.  This is a bummer from a UX perspective.  Scripts that previous worked now fail.  Coaching usage will only go so far.

### Proposed solution

The goal: make the previously long-term working scripts "just work" and the default usage friendly for broad use cases.

In effect I propose that the following, somehow, becomes the default for `client = client = Client()`:
```python
from obspy.clients.fdsn import Client
client = Client("EARTHSCOPE", service_mappings={
    "event": "https://earthquake.usgs.gov/fdsnws/event/1"})
```

chad-earthscope:
There are a couple of ways to accomplish this:

1. Create a service mapping for "EARTHSCOPE+USGS" that uses EarthScope endpoints and include in the USGS NEIC event endpoint and use that for the default
2. Create some code that parses the data center designation like "EARTHSCOPE+USGS" and creates a service mapping from the list of centers based on left-right priority.  So "EARTHSCOPE+ISC" would result in EarthScope dataselect & stations and ISC event.

I'm happy to create a PR, hopeful to discuss the concept first.

chad-earthscope:
As an experiment I created a PR for accepting `base_url` as lists and then merging services in priority of the specified (short)URLs.
https://github.com/obspy/obspy/pull/3769

I'm not convinced the above the right approach, please treat as food for thought.

megies:
> There are a couple of ways to accomplish this:
> 
>     1. Create a service mapping for "EARTHSCOPE+USGS" that uses EarthScope endpoints and include in the USGS NEIC event endpoint and use that for the default
> 
>     2. Create some code that parses the data center designation like "EARTHSCOPE+USGS" and creates a service mapping from the list of centers based on left-right priority.  So "EARTHSCOPE+ISC" would result in EarthScope dataselect & stations and ISC event.

@chad-earthscope I agree that's an unfortunate situation to have a default client now that covers only 2/3 of the endpoints and something needs to change.

[...]

PR #3769, opened 2026-07-15
FDSN Client accepts a list of base_url entries and merges endpoints.  Default changed to "EARTHSCOPE","USGS"

chad-earthscope:
### What this PR does

The `base_url` kwarg accepts a list of data center IDs and/or base URLs.  Discovery is conducted on each, and service URLs are merged together prioritizing the first entry for each service.  For example with `base_url = ["EARTHSCOPE", "ISC"]`, the `dataselect` and `station` services are discovered from EarthScope (but no `event` service), and the `event` service from ISC is discovered and added to the service mapping.

The default for `base_url` is also changed to `["EARTHSCOPE", "USGS"]`.

The motivation for this is the changed user experience now that EarthScope does not offer an `event` service.  Scripts that used to work now fail, and a default client that can get metadata, waveforms, and events is the experience this is trying to partially restore.

This is effectively an easier, and more dynamic, path to something like this:
```python
client = Client("EARTHSCOPE", service_mappings={
    "event": "https://earthquake.usgs.gov/fdsnws/event/1"})
```
It is also less explicit.

I'm not convinced this is the right approach, please treat as food for thought

### Links to other Issues

See https://github.com/obspy/obspy/issues/3768

### AI used

AI was used for review

### PR Checklist

### Issue labels

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,6 +12,10 @@ Changes:
      'measurement_units' when being initialized (see #3623)
  - obspy.clients.fdsn:
    * update URL endpoint for IGN to https (see #3744)
+   * FDSN Client `base_url` can be a list of datacenter shortcuts/URLs;
+     services missing from earlier entries are filled in from later ones
+     without overwriting. Default is now `["EARTHSCOPE", "USGS"]` so
+     event queries work after EarthScope dropped its event service.
  - obspy.clients.filesystem:
    * tsindex: improve handling of sql sessions, less connections staying open
      (see #3736)
@@ -42,7 +46,7 @@ Changes:
      mindepversion tests
    * Minimum Dependencies: numpy 1.21 (was 1.20) (see #3639)
  - obspy.core:
-   * response: tweaks to allow sane processing of PolynomialResponseStage 
+   * response: tweaks to allow sane processing of PolynomialResponseStage
      responses if they have 2 coefficients or less (see #3632)
    * response: allow recalculating overall sensitivity even if unit type is not
      one of VEL/ACC/DISP (see #3235), add method for calculating the normalization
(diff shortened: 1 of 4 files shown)


chad-earthscope:
```python
>>> from obspy.clients.fdsn import Client

>>> client = Client()
>>> for service in ("dataselect", "station", "event"):
...     if service in client.services:
...         print(service, client._build_url(service, "query"))
... 
dataselect https://service.earthscope.org/fdsnws/dataselect/1/query
station https://service.earthscope.org/fdsnws/station/1/query
event https://earthquake.usgs.gov/fdsnws/event/1/query


>>> client = Client(["EARTHSCOPE", "ISC"])
>>> for service in ("dataselect", "station", "event"):
...     if service in client.services:
...         print(service, client._build_url(service, "query"))
... 
dataselect https://service.earthscope.org/fdsnws/dataselect/1/query
station https://service.earthscope.org/fdsnws/station/1/query
event https://www.isc.ac.uk/fdsnws/event/1/query


>>> client = Client(["EARTHSCOPE", "https://www.seismicportal.eu"])
>>> for service in ("dataselect", "station", "event"):
...     if service in client.services:
...         print(service, client._build_url(service, "query"))
... 
dataselect https://service.earthscope.org/fdsnws/dataselect/1/query
station https://service.earthscope.org/fdsnws/station/1/query
event https://www.seismicportal.eu/fdsnws/event/1/query
```

ThomasLecocq:
@megies should this be in maintenance 1.5.x ?

megies:
> @megies should this be in maintenance 1.5.x ?

I'm torn, but thinking more about it, I think you could argue that even though it's a change in behavior now (making things work again), it's not really a behaviour change from our side, since the default FDSN Client should have never stopped working with events in the first place, so it's more fixing something that should not have broke :+1: 

However, I think we should go with the other approach maybe (but open to opinions of course!), see https://github.com/obspy/obspy/issues/3768#issuecomment-5397294638

ThomasLecocq:
Done in #3784 :-)

PR #3782, opened 2026-08-25
CI: add UV ci workflow for the tests - WIP

ThomasLecocq:
### What does this PR do?

it creates (in //) a new CI workflow using UV for running all tests. 

### Links to other Issues?

#3570 's @jschaeff mentionned UV and I think it's a good idea to test it here, it can later serve as basis to check the setup.py (for [tests] e.g. missing pytest-cov) and writing the documentation

### AI used?

The Claude created the YML file (and tested its workflow down to the obspy-runtests).
The Lecocq talked nicely to the Claude to make it correct the mistakes & bugs & #{|@@#|@#|....

### PR Checklist

### DIFF
--- a/.github/test_mindep_conda_env.yml
+++ b/.github/test_mindep_conda_env.yml
@@ -4,6 +4,8 @@ channels:
   - defaults
 dependencies:
   # obspy
+  # No optional feature deps installed here. Keep the test tooling in sync with
+  # .github/uv_mindep_requirements.txt (the uv counterpart).
   - decorator
   - lxml
   - matplotlib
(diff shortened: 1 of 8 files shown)


megies:
Do we want to add a second pipeline for CI tests parallel to the existing one? It's added maintenance burden.
I feel like if you think that we should test based on pypi (using UV or whatever else) because it's faster or whatever, then I think we should just make a full switch from the conda based environments we use in CI right now. Otherwise we have two parallel CI workflows to maintain that do the same thing.. right?

ThomasLecocq:
aha, funny enough, testing the free-threading version surfaced a little left-over unittest wrong line :)

tests: use pytest.skip in test_issue_2278 (not self.skipTest)

TestResourceIdentifier is a plain pytest class, not a unittest.TestCase, so
it has no skipTest method. test_issue_2278 skips when Python does not reuse
id()s between a freed batch of objects and a fresh one; on GIL CPython that
reuse routinely happens, the skip branch is not taken, and the latent bug
stays hidden. On free-threaded 3.14t the allocator does not reuse those ids,
the branch is taken, and self.skipTest raises AttributeError.

Switch to pytest.skip (pytest already imported). No-op on GIL builds (branch
not taken); on 3.14t the test now skips cleanly instead of erroring.

Surfaced by the tests-uv-experimental canary (3.14t).

PR #3798, opened 2026-09-01
Fix dayplot right-hand cutoff at some widths (Fixes #3466)

### LINKED ISSUE #3466: [Dayplot]: Traces being cut off on right-hand side of plot
### Avoid duplicates

- [x] I searched existing issues

### Bug Summary

Depending on the horizontal size of the output plot (set using the `size` parameter) when using `type='dayplot'`, the traces can become "stretched" to the right, resulting in data being cut off. I've tested this on two different installations of obspy 1.4.1 (Windows and Linux) and the issue is present on both.

I haven't done extensive testing, but it seems like this happens at a horizontal resolution of around 4000 pixels. The exact value seems to change slightly depending on the data, but the issue seems to happen regardless of the input data; I've tested around 5 different data sets from 5 different sources, and this happened when plotting all of them.

I'll attach two images showing the affected output: the first is at 4000 x 4000, the second is at 4050 x 4000. Notice how the tremors on the right-hand side (visible in the 1st image) get completely cut off in the 2nd image (it's more noticeable if you download the images, or open them in new tabs, and quickly flip between them)
Resolution set to 4000 x 4000
Resolution set to 4050 x 4000

Changing other parameters, like `dpi` or the output format, makes no difference.

I'll add the code below, and here is the link for the data used for the above images from SCEDC (~6.6MB): [https://service.scedc.caltech.edu/fdsnws/dataselect/1/query?net=CI&sta=WOR&cha=BHZ&loc=--&start=2019-07-06T00:00:00&end=2019-07-06T23:59:59&format=miniseed&nodata=404](https://service.scedc.caltech.edu/fdsnws/dataselect/1/query?net=CI&sta=WOR&cha=BHZ&loc=--&start=2019-07-06T00:00:00&end=2019-07-06T23:59:59&format=miniseed&nodata=404)

### Code to Reproduce

```python
#
#NOTE: Replace horizontal size with 4000 and 4050 to reproduce exact problem
#

from obspy import read
from obspy.core import UTCDateTime
from obspy.core import Stream

st = UTCDateTime("2019-07-06T00:00:00")
et = UTCDateTime("2019-07-06T23:59:59")

wf = read("path/to/data.miniseed")

wf.plot(type='dayplot', interval=15, right_vertical_labels=False,
        vertical_scaling_range=50000, one_tick_per_line=True,
        color=['white'], linewidth=1, bgcolor='black', show_y_UTC_label=False, 
        starttime=st, endtime=et, 
        dpi=200, size=(xxxx,4000), y_labels_size=4, outfile='outfilename.png',
        subplots_adjust_left=0.02, subplots_adjust_right=0.98, subplots_adjust_top=0.98, 
        subplots_adjust_bottom=0.02)
```

### Error Traceback

_No response_

### ObsPy Version?

1.4.1

### Operating System?

Windows, Linux

### Python Version?

3.10.14

### Installation Method?

pip

ThomasLecocq:
dayplot binned each interval into int(spp) samples per pixel and dumped the leftover into the last pixel, squashing up to ~1/int(spp) of every row's tail into the right-most pixel when spi is not a multiple of the width. Bin evenly via reduceat instead. Regression test added.

Before-after view:

image

Fixes #3466

### AI used?

Claudy & proofread by a human

### PR Checklist

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -8,6 +8,12 @@ Changes:
  - obspy.signal:
    * spectral estimation: add getter functions for rotational noise models
      (see #3732)
+ - obspy.imaging:
+   * dayplot: fix the right-hand side of each interval being cut off / squashed
+     into the last pixel at plot widths where the samples per interval are not
+     an integer multiple of the pixel width. Samples are now distributed evenly
+     across pixels instead of folding the remainder into the last one
+     (see #3466)
 
 1.5.1 (doi: 10.5281/zenodo.22108563)
 ====================================
@@ -53,11 +59,6 @@ Changes:
  - obspy.imaging:
    * fix a bug in dayplot when data is only available for part of the time
      window (see #3780)
-   * dayplot: fix the right-hand side of each interval being cut off / squashed
-     into the last pixel at plot widths where the samples per interval are not
-     an integer multiple of the pixel width. Samples are now distributed evenly
-     across pixels instead of folding the remainder into the last one
-     (see #3466)
  - obspy.io.mseed:
    * fix the type of exception raised when trying to read an empty file
      (see #3709)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -53,6 +53,11 @@ Changes:
  - obspy.imaging:
    * fix a bug in dayplot when data is only available for part of the time
      window (see #3780)
+   * dayplot: fix the right-hand side of each interval being cut off / squashed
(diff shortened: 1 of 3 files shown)


ThomasLecocq:
using the example from #3466 , now the plots visuals are independent of the plot's width:

1800 px wide plot (click to open)
image

1900 px wide plot (click to open)
image

filefolder (approved):
looks good to me

PR #3801, opened 2026-09-01
Fix: Raise explicit exceptions instead of misplaced bare raise (E0704)

ThomasLecocq:
A bare `raise` outside an active exception handler raises `RuntimeError: No active exception to re-raise`, not the intended exception.

io/csv/core.py: three bare `raise` statements sit inside `try` blocks (not their `except`) and were relied on to drop into the surrounding `except Exception`. They only worked because the RuntimeError they actually raised is itself an Exception. Raise an explicit ValueError for the NaN / invalid-value cases instead; still caught by the same handler, but no longer dependent on that accident.

clients/seedlink/tests/test_basic_client.py: the `for/else` bare `raise` was meant to surface the AssertionError from the failed offsets, but runs after the `except` has exited via `continue`, so no exception is active. Capture the AssertionError and re-raise it.

### AI used?

Claudyboy

### PR Checklist

### DIFF
--- a/obspy/clients/seedlink/tests/test_basic_client.py
+++ b/obspy/clients/seedlink/tests/test_basic_client.py
@@ -36,15 +36,17 @@ class TestClient():
         # getting a result depends on two things.. how long backwards the ring
         # buffer stores data and how close to realtime the data is available,
         # so check some different offsets and see if we get some data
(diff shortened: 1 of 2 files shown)


trichter (approved):
Fine with me

PR #3807, opened 2026-09-18
Add request hook for HTTP requests, i.e. to add bearer token for auth and more

chad-earthscope:
Add a `request_hook` mechanism for injecting headers into outgoing HTTP requests

## Problem

EarthScope is soon going to require authentication in order to enforce quotas for data access; exact details are not yet known, but authentication will be needed regardless. This will drive increased use of authentication. To reduce the friction these changes will cause for users, EarthScope would like users to adopt the new token-based authentication pattern (aligned with the FDSN proposal) instead of the deprecated Digest-based user/password pattern. ObsPy is used for the majority of data access to EarthScope's FDSN services, so token-based auth support would go a long way to helping users make this transition more easily.

This PR adds an optional hook to web service Clients allowing HTTP request headers to be modified, such as the injection of bearer tokens for auth.

The management of tokens is more complex than static user/password pairs, as the tokens may need to be refreshed routinely and support a variety of workflows. The mechanisms may vary from data center to data center. To support this without adding new dependencies, the PR adds a callable hook that returns optionally modified headers. In the case of EarthScope, the `earthscope-sdk` is the preferred package for managing users' tokens. Building an appropriate request hook with this SDK is trivial (and used as an example), but the hook could also be included in the SDK itself to make it even easier. This approach keeps the token management separate from ObsPy.

Note that this functionality is motivated by token-base auth, but it is generic and may be used to inject other headers for future use cases.

## Current situation

ObsPy's HTTP clients support `user`/`password` and EIDA-token auth, but there is no way to attach arbitrary headers such as a bearer token to outgoing requests.

## What's added

- **`obspy.clients.hooks`** (new module): a *request hook* is any callable `f(request)` invoked on every outgoing request before it's sent, letting it inspect the request (URL, host, scheme, method, headers) and add headers.
  - `HookRequest`: a stack-neutral view normalizing the two different backends ObsPy's clients use (`urllib.request.Request` for `fdsn.Client`, `requests.PreparedRequest` for everything built on `HTTPClient`, e.g. syngine and the routing clients) into one common interface, so a single hook works against all of them.
  - `set_header()` / `set_secret_header()`: the latter is redirect-safe — it won't get copied onto a cross-host redirect, the right choice for anything sensitive like a token.
  - `BearerTokenHook`: attaches `Authorization: Bearer <token>`, scoped per host (a dict of `host -> token`, where each token may itself be a provider callable for refreshable tokens, or a single callable applied to every host), so one hook instance can serve a multi-data-center query without leaking one DC's token to another. Defaults to HTTPS-only.
  - `LoggingHook`: logs every outgoing request/host; composable with `chain(...)` alongside an authenticating hook.
  - `RequestHookError`: raised (and never silently swallowed by the clients' generic exception handling) when a hook itself fails, so a broken hook fails loudly with its own traceback.

- **Wired into every client that makes HTTP requests:**
  - `obspy.clients.fdsn.Client(..., request_hook=...)`, plus a `set_request_hook()` setter that can be changed post-construction without rebuilding the opener from scratch (so it composes with existing `user`/`password`/EIDA-token auth).
  - `obspy.clients.base.HTTPClient(..., request_hook=...)` (the `requests`-based base class), covering syngine and the routing clients; also has a `set_request_hook()` setter.
  - `RoutingClient(..., request_hook=...)` — applies to the routing/federator query itself and to every data center it discovers.
  - `MassDownloader(..., request_hook=...)` — forwards the hook to every auto-created `Client` for its providers (providers passed in as already-built `Client` instances keep their own hook).

## How it would be used

```python
import threading

from earthscope_sdk import EarthScopeClient
from obspy.clients.fdsn.mass_downloader import MassDownloader
from obspy.clients.hooks import BearerTokenHook

esc = EarthScopeClient()
lock = threading.Lock()

def earthscope_token(request):
    # Serialize refreshes: this hook is called concurrently by the
    # routing client and the mass downloader, and the SDK does not
    # lock its own refresh path.
    with lock:
        esc.ctx.auth_flow.refresh_if_necessary()
        return esc.ctx.auth_flow.access_token

# Use included BearerTokenHook convenience wrapper
# Value can be a callable or literal token
hook = BearerTokenHook({
    "*.earthscope.org": earthscope_token,
    "my.datacenter.net": "my-token",
})

# Single data center
client = Client("EARTHSCOPE+USGS", request_hook=hook)

# Federated query across several data centers with one hook
routed = RoutingClient("eida-routing", request_hook=hook)

# Bulk download across many providers
mdl = MassDownloader(request_hook=hook)
```

The hook resolves what to attach (if any) purely from the request's host, so it works transparently even for data centers the caller never explicitly named (e.g. ones discovered through routing). `earthscope_token` is only consulted for requests actually going to a matching host — it never has to check for itself, and its token cannot reach another data center.

## Notable design details / caveats (documented in the module docstring)

- The two backends behave differently under redirects: urllib re-invokes the hook per hop, `requests` does not (its `auth=` callable isn't re-run by `Session.rebuild_auth()`), so a host-aware hook only fully re-resolves per-hop on the urllib-backed `fdsn.Client`.
- Hooks may be called concurrently (routing clients and mass downloader both use thread pools), so a stateful hook (e.g. token cache) must be thread-safe.
- Hooks can only read requests and add headers — not rewrite URL/body/method.
- Since a hook failure now propagates out of the thread pools used by `RoutingClient` and `MassDownloader`, their `pool.close()` calls were moved into `finally` blocks so a broken hook can't leak the pool.

## Tests / docs

- New `obspy/clients/tests/test_hooks.py` covering both backends, redirect behavior, `BearerTokenHook` (including glob-pattern host matching and token providers), `LoggingHook` redaction, `chain()`, and error wrapping.
- Extended `test_client.py`, `test_base_routing_client.py`, `test_mass_downloader.py` for hook wiring in each client.
- New Sphinx page `misc/docs/source/packages/obspy.clients.hooks.rst`, added to the package index.
- `CHANGELOG.txt` updated under `obspy.clients`.

### PR Checklist

### Issue labels


chad-earthscope:
Any thoughts on broader, general usability by @javiquinte would be appreciated.

PR #3810, opened 2026-09-22
FDSNWS: Add JWT authentication

megies:
### What does this PR do?

This is a preliminary draft to brainstorm how to include the new [json web token (JWT)](https://www.rfc-editor.org/info/rfc7519/) based authentication scheme for FDSNWS, which currently is in proposal/review stage (see https://github.com/FDSN/Review-New-Auth-Proposal).

From a first look, it should be very simple to just include the option for the user to provide an access token in its encoded form (i.e. the header, payload and signature in a base64 encoded string concatenated together). All we would need to do then is basically send the encoded token along as an HTTP header (`Authorization: Bearer <token>`). 
Since access tokens are relatively short lived (on the order of hours), this likely means that workflows would include refreshing the access tokens with separate tools ([eas2cli](https://pypi.org/project/eas2cli/), [earthscope-cli](https://pypi.org/project/earthscope-cli/), ..), e.g. automatically in cronjobs or manually on command line.

Another option would be to wrap around the above helpers to refresh the access tokens in obspy directly (fairly easy with a single POST e.g. in the case of `EAS`). But that might not be needed, really.

### Links to other Issues?

see https://github.com/FDSN/Review-New-Auth-Proposal

### AI used?

no

### PR Checklist

- [ ] Correct **base branch** selected? [`master` for new features, `maintenance_?.?.x` for bug fixes](https://github.com/obspy/obspy/wiki/ObsPy-Git-Branching-Model)

### Issue labels

### Example / Showcase

The following example requires the token to be downloaded from https://geofon.gfz.de/eas2/ or using the tool [eas2cli](https://pypi.org/project/eas2cli/)

```python
import json
import pytest
from obspy import UTCDateTime
from obspy.clients.fdsn.header import FDSNUnauthorizedException


t = UTCDateTime('2026-09-01T12:00:00')

with open('/home/megies/.eidajwt', 'rt') as fh: 
    jwt = json.loads(fh.read())['access_token']


client = Client('GEOFON', jwt=jwt, debug=True)
print(client.get_waveforms('GE', 'IBBN', '', 'LHZ', t, t+5))

# modify token signature. this means the header and payload are still OK, but
# the server should check the signature and detect the mismatch, which should
# invalidate it and make the server reject the request
jwt_invalid = jwt + 'dfsff'
# force discovery of services again
Client._Client__service_discovery_cache = {}
client = Client('GEOFON', jwt=jwt_invalid, debug=True)
with pytest.raises(FDSNUnauthorizedException) as e:
    print(client.get_waveforms('GE', 'IBBN', '', 'LHZ', t, t+5))
assert e.value.status_code == 401 
```


chad-earthscope:
> Since access tokens are relatively short lived (on the order of hours), this likely means that workflows would include refreshing the access tokens with separate tools ([eas2cli](https://pypi.org/project/eas2cli/), [earthscope-cli](https://pypi.org/project/earthscope-cli/), ..), e.g. automatically in cronjobs or manually on command line.

This puts the burden on the user to ensure that the token being added to the request is valid and when it's not interrupts data flow, which I suspect will be awkward and confusing to users trying to resume from the failed request.  It's also more code for the user.  As pointed out some data centers/federations offering tokens have tooling to automatically refresh tokens, and the point of https://github.com/obspy/obspy/pull/3807 was to let those toolkits do their work seamlessly to the user (once initialized).  

An alternative would be to allow the value of jwt to be a callable (in addition to a static string perhaps) and return a valid token or throw an exception.

